PR #544, opened 2020-07-06
CLDR-13900 catch bad superscripts

macchiati:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13900

### DIFF
--- a/tools/java/org/unicode/cldr/test/CheckCLDR.java
+++ b/tools/java/org/unicode/cldr/test/CheckCLDR.java
@@ -717,7 +717,7 @@ abstract public class CheckCLDR {
             invalidPlaceHolder, asciiQuotesNotAllowed, badMinimumGroupingDigits, inconsistentPeriods,
             inheritanceMarkerNotAllowed, invalidDurationUnitPattern, invalidDelimiter, illegalCharactersInPattern,
             badParseLenient, tooManyValues, invalidSymbol, invalidGenderCode,
-            mismatchedUnitComponent
+            mismatchedUnitComponent, longPowerWithSubscripts
             ;
 
             @Override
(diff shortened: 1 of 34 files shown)


tools/java/org/unicode/cldr/test/CheckCLDR.java
```diff
@@ -1392,4 +1392,8 @@ public CLDRFile getEnglishFile() {
     public void setEnglishFile(CLDRFile englishFile) {
         this.englishFile = englishFile;
     }
+
+    public CharSequence fixedValueIfInherited(String value, String path) {
```
macchiati, line 1396: Added utility that will need for other tickets also.

tools/java/org/unicode/cldr/test/CheckUnits.java
```diff
@@ -33,7 +34,7 @@
     public CheckCLDR setCldrFileToCheck(CLDRFile cldrFileToCheck, Options options, List<CheckStatus> possibleErrors) {
         super.setCldrFileToCheck(cldrFileToCheck, options, possibleErrors);
         GrammarInfo grammarInfo = CLDRConfig.getInstance().getSupplementalDataInfo().getGrammarInfo(cldrFileToCheck.getLocaleID());
-        genders = grammarInfo == null ? Collections.emptySet() : grammarInfo.get(GrammaticalTarget.nominal, GrammaticalFeature.grammaticalGender, GrammaticalScope.units);
+        genders = grammarInfo == null ? null : grammarInfo.get(GrammaticalTarget.nominal, GrammaticalFeature.grammaticalGender, GrammaticalScope.units);
```
macchiati, line 37: Changed to null so we can tell difference between "doesn't use gender" and "no info available"

tools/java/org/unicode/cldr/test/CheckUnits.java
```diff
@@ -93,8 +94,21 @@ public CheckCLDR handleCheck(String path, String fullPath, String value, Options
             }
             String idType;
             switch(pathType) {
+            case power: {
```
macchiati, line 97: This is slightly different than the ticket; cause an error but note to users that they can use gendered variants, eg feminine of 'cubic'

macchiati:
It was failing before because it was (correctly) detecting errors in the trunk data. So I went through and fixed them. Almost all were incorrectly inheriting from root:
1. Removed the count-less long power units (those were left over and should all be removed, but that's for later)
2. Where everything inherited from root, deleted the items.
3. Where some of the counts inherited from root, replaced those with other count values and marked as provisional.

PR #545, opened 2020-07-06
CLDR-13857 shallow copy in cloneAsThawed if instance is frozen

AlintaLu:
- Old cloneAsThawed is an expensive operation as it copies everything over to a mutable structure. New cloneAsThawed makes shallow copy of elements if XPathParts is frozen, and makes an element mutable unless it is necessary.
- Collections.unmodifiableList/unmodifiableMap move to ImmutableList/ImmutableMap

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13857

### DIFF
--- a/tools/java/org/unicode/cldr/util/XPathParts.java
+++ b/tools/java/org/unicode/cldr/util/XPathParts.java
@@ -10,7 +10,6 @@ import java.io.File;
 import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -21,6 +20,8 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSet.Builder;
 import com.ibm.icu.impl.Utility;
@@ -439,7 +440,18 @@ public final class XPathParts implements Freezable<XPathParts>, Comparable<XPath
     }
 
     public void putAttributeValue(int elementIndex, String attribute, String value) {
-        elements.get(elementIndex >= 0 ? elementIndex : elementIndex + size()).putAttribute(attribute, value);
+        elementIndex = elementIndex >= 0 ? elementIndex : elementIndex + size();
+        Map<String, String> ea = elements.get(elementIndex).attributes;
+        if (value == null && (ea == null || !ea.containsKey(attribute))) {
+            return;
+        }
+        if (value != null && ea != null && ea.containsKey(attribute) && ea.get(attribute).equals(value)) {
+            return;
+        }
+        makeElementsMutable();
+        makeElementMutable(elementIndex);
+        // make mutable may change elements.get(elementIndex), so we have to use elements.get(elementIndex) after calling
+        elements.get(elementIndex).putAttribute(attribute, value);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tools/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -439,7 +440,18 @@ public String getAttributeValue(int elementIndex, String attribute) {
     }
 
     public void putAttributeValue(int elementIndex, String attribute, String value) {
-        elements.get(elementIndex >= 0 ? elementIndex : elementIndex + size()).putAttribute(attribute, value);
+        elementIndex = elementIndex >= 0 ? elementIndex : elementIndex + size();
+        Map<String, String> ea = elements.get(elementIndex).attributes;
+        if (value == null && (ea == null || !ea.containsKey(attribute))) {
+            return;
+        }
+        if (value != null && ea != null && ea.containsKey(attribute) && ea.get(attribute).equals(value)) {
```
macchiati, line 448: ea.containsKey(attribute) && ea.get(attribute).equals(value)

should be optimized (since value is at this point guaranteed not to be null) as:

value.equals(ea.get(attribute))

tools/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -527,7 +562,14 @@ public XPathParts removeAttributes(String elementName, Collection<String> attrib
     }
 
     public XPathParts removeAttributes(int elementIndex, Collection<String> attributeNames) {
-        elements.get(elementIndex >= 0 ? elementIndex : elementIndex + size()).removeAttributes(attributeNames);
+        elementIndex = elementIndex >= 0 ? elementIndex : elementIndex + size();
+        if (elements.get(elementIndex).attributes == null || attributeNames == null) {
```
macchiati, line 566: change to:

Map<...> ea = elements.get(elementIndex).attributes;

if (temp == null || attributeNames == null || attributeNames.isEmpty() || Collections.disjoint(attributeNames, ea.keySet())) {

That verifies that we have to make the element mutable before going further.

tools/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -712,8 +754,7 @@ private XPathParts parseError(String s, int i) {
     public static final int XPATH_STYLE = 0, XML_OPEN = 1, XML_CLOSE = 2, XML_NO_VALUE = 3;
     public static final String NEWLINE = "\n";
 
-    private final class Element implements Cloneable, Freezable<Element> {
```
macchiati, line 715: Might not need Cloneable.

tools/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1049,16 +1070,17 @@ public XPathParts set(XPathParts parts) {
...
-            for (Element element : parts.elements) {
-                elements.add((Element) element.clone());
-            }
-            return this;
-        } catch (CloneNotSupportedException e) {
-            throw (InternalError) new InternalError().initCause(e);
+
+        dtdData = parts.dtdData;
+        elements.clear();
+        if (parts.elements != null && parts.elements.size() > 0) {
```
macchiati, line 1076: parts.elements.size() > 0

better to do:

!parts.elements.isEmpty()

tools/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1049,16 +1070,17 @@ public XPathParts set(XPathParts parts) {
...
+            makeElementsMutable();
         }
+        for (Element element : parts.elements) {
+            elements.add(element.cloneAsThawed());
```
macchiati, line 1080: If what we are copying is frozen, we can copy all the elements without cloning.

tools/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1049,16 +1070,17 @@ public XPathParts set(XPathParts parts) {
...
-            elements.clear();
-            for (Element element : parts.elements) {
-                elements.add((Element) element.clone());
-            }
-            return this;
-        } catch (CloneNotSupportedException e) {
-            throw (InternalError) new InternalError().initCause(e);
+
+        dtdData = parts.dtdData;
+        elements.clear();
```
macchiati, line 1075: Note that an unfrozen XPathParts might have an immutable list — so you have to do:

if (!elements.isEmpty()) {
  makeMutable();
  elements.clear();
}

AlintaLu:
See https://github.com/unicode-org/cldr/pull/547

PR #546, opened 2020-07-06
CLDR-13905 bogus link in error for dot unit

btangmu:
-Do not delete required count attribute from unitPattern paths

-New TestCheckDisplayCollisions.testCountNormalization, checkUnknown

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13905

### DIFF
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
@@ -23,20 +23,20 @@ import com.google.common.collect.ImmutableList;
 public class TestCheckDisplayCollisions extends TestFmwkPlus {
     private static final String ukRegion =  "//ldml/localeDisplayNames/territories/territory[@type=\"GB\"]";
     private static final String englandSubdivision = "//ldml/localeDisplayNames/subdivisions/subdivision[@type=\"gbeng\"]";
-    
+
     private static final String scorpioEmoji = "//ldml/annotations/annotation[@cp=\"♏\"][@type=\"tts\"]";
     private static final String scorpionEmoji = "//ldml/annotations/annotation[@cp=\"🦂\"][@type=\"tts\"]";
-    
+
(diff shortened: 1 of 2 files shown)


tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
```diff
@@ -78,15 +78,46 @@ public void testInheritance() {
...
         CLDRFile frCaResolved = factory.make("fr_CA", false);
-        checkFile(cdc, frCA, frCaResolved, 
+        checkFile(cdc, frCA, frCaResolved,
             scorpioEmoji, ukRegion);
     }
-    
+
+    public void testCountNormalization() {
+        final String testLocale = "pt";
+        final String unitPattern1 = "//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"graphics-dot\"]/unitPattern[@count=\"one\"]";
```
macchiati, line 92: Must test the three cases:
 "//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"graphics-dot\"]/unitPattern[@count=\"one\"]"
 "//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"graphics-dot\"]/unitPattern[@count=\"other\"]" // different count as # 1. MUST NOT COLLIDE WITH #1
"//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"length-point\"]/unitPattern[@count=\"one\"]"; // different unit as # 1. MUST COLLIDE WITH #1

macchiati: And that is with all 3 having the same value.

macchiati: Also check to make sure that case="nominative" and case="accusative" not collide when they paths are otherwise identical and have the same value.

btangmu: OK, I'll add those tests and make another PR.

PR #548, opened 2020-07-07
CLDR-13882 38 extended symbols are not available

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13882


macchiati:
Note: I didn't have time to split the pages, so the Symbols2 page will be long at the comprehensive level.
Kristi should check the characters; Tom, if you wouldn't mind a sanity check of the code.

macchiati:
I wasn't able to test it in my local ST yet (separate issue), so I won't
merge until then.

Mark

macchiati:
BTW, when reviewing I almost always hit the settings option to hide
whitespace changes; usually makes it easier to read.

[image: Screen Shot 2020-07-07 at 17.10.16.png]


Mark

PR #549, opened 2020-07-07
CLDR-13900 Mistakenly caused the unit tests to be skipped for the current phase

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13900


macchiati:
I'd incorrectly copied some code from some other Check classes, which caused the tests to be disabled for this phase; not what was desired!

PR #554, opened 2020-07-08
CLDR-13905 bogus link in error for dot unit

btangmu:
-Do not delete required count attribute from unitPattern paths

-New TestCheckDisplayCollisions.testUnitPatternCollisions, mustCollide, checkUnknown

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13905

### DIFF
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
@@ -23,20 +23,20 @@ import com.google.common.collect.ImmutableList;
 public class TestCheckDisplayCollisions extends TestFmwkPlus {
     private static final String ukRegion =  "//ldml/localeDisplayNames/territories/territory[@type=\"GB\"]";
     private static final String englandSubdivision = "//ldml/localeDisplayNames/subdivisions/subdivision[@type=\"gbeng\"]";
-    
+
     private static final String scorpioEmoji = "//ldml/annotations/annotation[@cp=\"♏\"][@type=\"tts\"]";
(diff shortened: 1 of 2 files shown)


btangmu:
This has more extensive testing; otherwise it's the same as https://github.com/unicode-org/cldr/pull/546

btangmu:
> Looks great, thanks! I still don't quite get **why** it is working, because it doesn't seem to be normalizing away the count and case values, but I don't want to stop this from going in.

I think the reason is that neither the UNITS regex nor the more specific tests care whether "count" matches. The UNITS regex is:
        UNITS("//ldml/units/unitLength.*/(displayName|unitPattern|perUnitPattern)", MatchType.REGEX),
Paths that match that, and have matching values, are guilty of collisions until proven innocent. The "more specific tests" I mean are around the line "boolean isDuration = myUnit.startsWith("duration");".

macchiati:
That's what I don't understand.

The code has to know that:

   1. a difference in unit (point vs dot) DOES matter (and should cause an
   error if the values are identical)
   2. a difference in the count (one vs two) or case (nominative vs
   accusative) does NOT matter (and should *not* cause an error if the
   values are identical)

Your test appears to be testing for it correctly, and it is working, so
that's why I approved it. But I'm not seeing that difference expressed in
the code, which leaves me uneasy. Will look it over later when I have time.

Mark

btangmu:
> That's what I don't understand. The code has to know that: 1. a difference in unit (point vs dot) DOES matter...

"myUnit.equals(unit)" is one of the conditions in the "more specific tests" I referred to. It can cause a suspected collision to be treated as a non-collision. There's no such test for "count".

PR #555, opened 2020-07-08
CLDR-13945 OPEN is a reserved SQL keyword, breaks on Derby

btangmu:
-Change table column open to is_open for all platforms (mysql/derby/etc)

-A one-time manual db modification will be required on production and smoketest

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13945


btangmu:
I'm making this a DRAFT PR, since I want to be sure that the reviewers understand we need to be careful about merging to smoketest and pushing to production. The db needs a simple one-time change on both those servers, coordinated with the merger and the push to production. It's just this one-liner in mysql: ALTER TABLE cldr_forum_posts CHANGE "open" "is_open" tinyint(1);

btangmu:
I think the one-time change in mysql is preferable in the long term to ongoing different column names between mysql and derby.

btangmu:
See https://github.com/unicode-org/cldr/pull/724

PR #556, opened 2020-07-09
CLDR-13838 adjust root currencySpacing to avoid double space

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13838

Adjust root currencySpacing to avoid inserting additional space if the currency symbol text already includes space  (e.g. "EUR ", which some clients have done).

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -3657,12 +3657,12 @@ for derived annotations.
 		<currencyFormats numberSystem="latn">
 			<currencySpacing>
 				<beforeCurrency>
-					<currencyMatch>[:^S:]</currencyMatch>
+					<currencyMatch>[[:^S:]&amp;[:^Z:]]</currencyMatch>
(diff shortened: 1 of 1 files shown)


sffc:
Seems harmless, but we should do our best to make sure that stock CLDR symbols don't have the redundant space to begin with. If clients are inserting their own space, that's not really our problem. However, the proposed change is fine and harmless.

pedberg-icu:
>  If clients are inserting their own space, that's not really our problem. However, the proposed change is fine and harmless.

No, it is our problem. They were doing this before ICU started honoring currencySpacing to get the right behavior. The fact that ICU behavior changed should not change the behavior of the apps that are doing this. This fix supports such legacy apps but still does the right thing for new clients.

Apps do not all suddenly change when ICU does.

PR #558, opened 2020-07-09
CLDR-13857 use singleton to get CLDR directory

AlintaLu:
Optimize CLDRConfig.getCldrBaseDirectory()

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13857

### DIFF
--- a/tools/java/org/unicode/cldr/util/CLDRConfig.java
+++ b/tools/java/org/unicode/cldr/util/CLDRConfig.java
@@ -533,7 +533,7 @@ public class CLDRConfig extends Properties {
         }
     }
 
-    static class FileWrapper {
+    private static class FileWrapper {
         private File cldrDir = null;
         private FileWrapper() {
             String dir = getInstance().getProperty("CLDR_DIR", null);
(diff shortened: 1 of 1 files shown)


tools/java/org/unicode/cldr/util/CLDRConfig.java
```diff
@@ -533,15 +533,37 @@ public int getProperty(String k, int defVal) {
...
-    public File getCldrBaseDirectory() {
-        String dir = getProperty("CLDR_DIR", null);
-        if (dir != null) {
-            return new File(dir);
-        } else {
-            return null;
+    static class FileWrapper {
+        private File cldrDir = null;
+        private FileWrapper() {
+            String dir = getInstance().getProperty("CLDR_DIR", null);
```
macchiati, line 539: good work here, using the getInstance to allow the class to be static.

tools/java/org/unicode/cldr/util/CLDRConfig.java
```diff
@@ -533,15 +533,37 @@ public int getProperty(String k, int defVal) {
...
+        public static FileWrapper getFileWrapperInstance() {
+            if (fileWrapperInstance == null) {
+                synchronized(FileWrapper.class) {
+                    if (fileWrapperInstance == null) {
+                        fileWrapperInstance = new FileWrapper();
+                    }
+                }
+            }
+            return fileWrapperInstance;
         }
```
macchiati, line 560: I think these can be replaced by the following, avoiding the need for the test and synchronization (which is part of the reason for using the Singleton model).

        private static FileWrapper fileWrapperInstance = new FileWrapper();
        private static FileWrapper getFileWrapperInstance() {
            return fileWrapperInstance;
        }

PR #561, opened 2020-07-13
CLDR-13954 Make sure child SEED locales are included.

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13954

### DIFF
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCLDRFile.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCLDRFile.java
@@ -50,6 +50,7 @@ import org.unicode.cldr.util.SupplementalDataInfo.PluralType;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
 import com.google.common.collect.TreeMultimap;
 import com.ibm.icu.dev.test.TestFmwk;
 import com.ibm.icu.impl.Relation;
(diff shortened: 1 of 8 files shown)


macchiati:
This looks bigger than it is, because I added a parameter with refactoring. The files to focus on are CLDRFile, where we replaced a test of SEED, and on SupplementalDataInfo, where there is a logic change to consider SEED, and the test file.

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestCLDRFile.java
```diff
@@ -844,4 +844,13 @@ public void TestExtraPaths() {
         assertTrue("Should be empty", missing.isEmpty());
     }
 
+    public void TestExtraPaths13954() {
+        CLDRFile es = cldrFactory.make("es", true);
+        CLDRFile es_US = cldrFactory.make("es_US", true);
+        if (!es_US.getRawExtraPaths().containsAll(es.getRawExtraPaths())) {
+            TreeSet<String> missing = new TreeSet<>(es.getRawExtraPaths());
```
hagbard, line 851: Sets.difference(es.getRawExtraPaths(), es_US.getRawExtraPaths()) perhaps?
It's use the ordering of the first set, so maybe not ordered the same.

tools/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -4494,14 +4494,17 @@ public UnitPreferences getUnitPreferences() {
 
     /**
      * Grammar info for locales, with inheritance
+     * @param seedOnly TODO
      */
-    public GrammarInfo getGrammarInfo(String locale) {
-        while (locale != null) {
+    public GrammarInfo getGrammarInfo(String locale, boolean seedOnly) {
```
hagbard, line 4499: Maybe use an enum over a boolean here (readability).

supplementalData.getGrammarInfo(cldrFile.getLocaleID(), ALL_LOCALES)
seems to read better than:
supplementalData.getGrammarInfo(cldrFile.getLocaleID(), false)

macchiati: This is temporary, since we'll be removing SEED next release. Did make a few other changes for simplicity

macchiati:
Thanks.. should be ready to check in 20min or so.

PR #566, opened 2020-07-16
CLDR-13857 Fix BufferedInputStream

AlintaLu:
- Remove FilterBomInputStream and use existing StripUTF8BOMInputStream as they do the same thing
- Faster XML reading by using BufferedInputStream in StripUTF8BOMInputStream
- Remove XML read code in CLDRFile.loadFromInputStream and use existing XMLFileReader.read

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13857


macchiati:
Yiqing, if you are ready with this, you'll need to hit the "Ready for review" button.

PR #567, opened 2020-07-17
CLDR-13720 shorten narrow display names for digital units

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13720

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -8015,52 +8015,52 @@ annotations.
 				<unitPattern count="other">{0}m/gUK</unitPattern>
 			</unit>
 			<unit type="digital-petabyte">
-				<displayName>PByte</displayName>
+				<displayName>PB</displayName>
 				<unitPattern count="one">{0}PB</unitPattern>
 				<unitPattern count="other">{0}PB</unitPattern>
 			</unit>
 			<unit type="digital-terabyte">
-				<displayName>TByte</displayName>
+				<displayName>TB</displayName>
 				<unitPattern count="one">{0}TB</unitPattern>
 				<unitPattern count="other">{0}TB</unitPattern>
 			</unit>
 			<unit type="digital-terabit">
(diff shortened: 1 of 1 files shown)


macchiati:
Note: all of them were shortened except for "bit".

common/main/en.xml
```diff
@@ -8015,52 +8015,52 @@ annotations.
...
 			</unit>
 			<unit type="digital-kilobit">
-				<displayName>kbit</displayName>
+				<displayName>kb</displayName>
 				<unitPattern count="one">{0}kb</unitPattern>
 				<unitPattern count="other">{0}kb</unitPattern>
 			</unit>
 			<unit type="digital-byte">
-				<displayName>byte</displayName>
+				<displayName>B</displayName>
```
sffc, line 8063: Why did you change `byte` to `B` but not `bit` to `b`?

robertgmelo: Maybe because narrow values on bit type are `bit`, so I assume Mark wanted to keep the logic same as other units. In any case, it's a good question. Maybe if we change _displayName_ from `bit` to `b`, we would also need to change the values, but Mark can confirm.
https://github.com/unicode-org/cldr/blob/master/common/main/en.xml#L8069-L8070

macchiati: The question is whether "b" would be recognizable in isolation: "3 b" or "3b". The kb or Mb is distinctive enough, but I'm a bit concerned about the singleton 'b'. But maybe I shouldn't be?

pedberg-icu: I think you should be. I do not think b is distinctive. I am not even sure that B by itself is distinctive.

macchiati:
We need to decide this before Wednesday if we want to get into 38.

macchiati:
Ok, I'll move the ticket out of design and merge

PR #568, opened 2020-07-17
CLDR-13969 Make ST faster by avoiding needless DOM manipulation

btangmu:
-New cldrChecksum to see whether row data has changed; skip DOM update if unchanged

-New unit tests for cldrChecksum

-Remove unused valueToItem

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13969
- [x] Updated PR title and link in previous line to include Issue number

PR #569, opened 2020-07-20
CLDR-13969 Make ST faster by avoiding needless DOM manipulation

btangmu:
-New cldrChecksum to see whether row data has changed; skip DOM update if unchanged

-New unit tests for cldrChecksum

-Remove unused valueToItem

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13969


btangmu:
This is like https://github.com/unicode-org/cldr/pull/568 but improved per Mark's suggestions

PR #525, opened 2020-06-22
CLDR-13306 Generate disruptive data churn report

btangmu:
-Work around bug for grammaticalState in DtdData.java

-ChartDelta: report when finished and report file-creation directories

-Remove dead/unused code from ChartDelta.java and Chart.java, make private where possible

-Remove duplicate new PathDescription from ExampleGenerator.java

-New ChartChurn.java based on copy of ChartDelta.java

-Filter paths based on design doc (in progress)

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13306
- [x] Updated PR title and link in previous line to include Issue number

PR #573, opened 2020-07-22

btangmu:
-ChartDelta: report when finished and report file-creation directories

-Avoid deprecation warning for DtdData.getInstance; cache getCldrBaseDirectory as CLDR_BASE_DIR

-Prevent exception if e is null in DtdData.isMetadata (mysterious transient grammaticalState bug)

-Refactor to obviate single-line getValueSplitter; call DtdData.getValueSplitter directly instead

-Remove dead/unused code from ChartDelta.java and Chart.java, make private where possible

-Remove duplicate new PathDescription from ExampleGenerator.java

-New ChartChurn.java based on copy of ChartDelta.java

-Filter paths based on design doc (in progress): pathCountsForChurn

##### Checklist

### DIFF
--- a/tools/java/org/unicode/cldr/tool/ChartChurn.java
+++ b/tools/java/org/unicode/cldr/tool/ChartChurn.java
@@ -59,10 +59,10 @@ import com.ibm.icu.text.NumberFormat;
 import com.ibm.icu.util.ICUUncheckedIOException;
 import com.ibm.icu.util.Output;
 
-public class ChartDelta extends Chart {
+public class ChartChurn extends Chart {
     private static final boolean verbose_skipping = false;
 
-    private static final String DIR_NAME = "delta";
+    private static final String DIR_NAME = "churn";
 
     private static final boolean SKIP_REFORMAT_ANNOTATIONS = ToolConstants.PREV_CHART_VERSION.compareTo("30") >= 0;
 
@@ -103,8 +103,16 @@ public class ChartDelta extends Chart {
     private final Level minimumPathCoverage;
     private final boolean verbose;
 
+    /**
+     * Generate charts describing CLDR data "churn", i.e., changes in paths of particular
+     * importance for stability; for example, "//ldml/characters/exemplarCharacters".
+     *
+     * This main() can be run directly or through GenerateAllCharts.main().
+     *
+     * @param args
+     */
     public static void main(String[] args) {
-        System.out.println("use -DCHART_VERSION=36.0 -DPREV_CHART_VERSION=34.0 to generate the differences between v36 and v34.");
+        System.out.println("use -DCHART_VERSION=37.0 -DPREV_CHART_VERSION=36.0 to generate the churn between v36 and v37.");
         MyOptions.parse(args);
         Matcher fileFilter = !MyOptions.fileFilter.option.doesOccur() ? null : PatternCache.get(MyOptions.fileFilter.option.getValue()).matcher("");
         if (MyOptions.orgFilter.option.doesOccur()) {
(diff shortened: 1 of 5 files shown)


btangmu:
This is a DRAFT PR for reference on work in progress. It takes into account changes suggested by Mark and Steven in earlier https://github.com/unicode-org/cldr/pull/525 .

So far ChartChurn.java is essentially just a copy of ChartDelta.java, with "delta" changed to "churn", and with the addition of pathCountsForChurn which filters the set of paths to check only paths of special interest.

If we decide this is roughly the right approach, the next steps should include: (1) enhance pathCountsForChurn to use regex, maybe RegexLookup, to select the exact set if paths; and (2) refactor ChartDelta and ChartChurn to share code instead of duplicating it.

tools/java/org/unicode/cldr/tool/ChartDelta.java
```diff
@@ -69,11 +69,8 @@
     private static final PageId DEBUG_PAGE_ID = PageId.DayPeriod;
```
macchiati, line 69: It makes it much longer to read when you do a bunch of mechanical cleanups at the same time, since I have to sift through everything to make sure that it isn't relevant. There has to be some way to communicate that only X is expected to be a material change, and all else is cleanup. Maybe add your own review comments?

tools/java/org/unicode/cldr/tool/ChartChurn.java
```diff
@@ -0,0 +1,1116 @@
+package org.unicode.cldr.tool;
```
macchiati, line 1: This is doing a lot of unnecessary work, I think because it was copied from the delta chart; that chart is much more complicated because it is looking at changes in various attributes, etc. and also looking at more than just main+annotations. We don't need to bother with the non-ldml pages, for example.

It might be the wrong format also, because we want to have one big chart that we can review, including all of the locales in one chart instead of being split into hundreds of locales. More like the Flagged chart in that respect. (You might have changed that but hard to tell because so much else is going on.)

macchiati:
Tom, I think it would be much simpler to start with something like ChartDayPeriods. You can see that it basically gathers some information, stuffs it into a table, and that's it.

tools/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -1440,7 +1440,7 @@ public boolean hasValue(String elementName) {
     public boolean isMetadata(XPathParts pathPlain) {
         for (String s : pathPlain.getElements()) {
             Element e = getElementFromName().get(s);
-            if (e.elementStatus == ElementStatus.metadata) {
```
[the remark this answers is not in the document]
srl295: Agreed, I don't think there should be a null path element.

srl295:
Also from a usage model,
1. I noticed that the `@CLDRTool` annotation doesn't seem to be added on these tools, that is no doubt due to a lack of documentation. I filed CLDR-13987 to track this.

2. The tool complained, but pressed on, when the 'cldr-archive' directory wasn't present.

```
Couldn't open factory: /Users/srl/src/cldr-archive/cldr-37.0/common/main
Will compare: main              /Users/srl/src/cldr/common/main         null
```

This seems like it should be a hard failure.

 I was able to create such an archive using the following (given CLDR_DIR):

```shell
mkdir -p ${CLDR_DIR}/../cldr-archive/
git clone --reference  ${CLDR_DIR} --depth 1  https://github.com/unicode-org/cldr -b release-37 ${CLDR_DIR}/../cldr-archive/cldr-37.0
```
(this is the most efficient way to fetch this content using git… or, is the production data needed?)

srl295:
also, I'm not sure why include-version and include-date are iframes, might be better served with inline substitution or JavaScript.

When I tried to run the tool locally, the include-version turned into mojibake because it didn't declare an encoding.

```patch
diff --git a/tools/java/org/unicode/cldr/tool/include-date.html b/tools/java/org/unicode/cldr/tool/include-date.html
index 8bb08ff830..5ea2430032 100644
--- a/tools/java/org/unicode/cldr/tool/include-date.html
+++ b/tools/java/org/unicode/cldr/tool/include-date.html
@@ -1 +1,2 @@
+<meta charset="utf-8"/>
 <p style="text-align: right">%date%</p>
diff --git a/tools/java/org/unicode/cldr/tool/include-version.html b/tools/java/org/unicode/cldr/tool/include-version.html
index 85909eb810..6afacb4164 100644
--- a/tools/java/org/unicode/cldr/tool/include-version.html
+++ b/tools/java/org/unicode/cldr/tool/include-version.html
@@ -1 +1,2 @@
+<meta charset="utf-8"/>
 CLDR v<span class='version'>%version%</span>
```

macchiati:
Those are so there is no different for identical content. Made y management
of charts far easier

btangmu:
I'm closing this and I've opened a new PR: https://github.com/unicode-org/cldr/pull/585

PR #577, opened 2020-07-24
CLDR-13948 add caching for downloading Tomcat

echeran:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13948

This PR caches the Tomcat binary tarball so that the build process is more resilient to network/server issues.  Also, the workflow trigger condition is fixed to run on PRs from any branch.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -16,7 +16,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: '*'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


echeran:
[First test run](https://github.com/echeran/cldr/pull/1/checks?check_run_id=904841383) - cache key not present, so download is performed.
[Second test run](https://github.com/echeran/cldr/pull/1/checks?check_run_id=904889585) - cache key present, so download is skipped.

@srl295 I don't have permissions to assign you as a reviewer, but feel free to assign yourself.

PR #580, opened 2020-07-25
CLDR-13994 stop propagation of events from the .hide-review button to body

srl295:
CLDR-13994

- To fix a performance issue
- Semantic change: you can now hide unrelated review items without dismissing the popover.

### DIFF
--- a/tools/cldr-apps/WebContent/js/review.js
+++ b/tools/cldr-apps/WebContent/js/review.js
@@ -377,9 +377,9 @@ function toggleFix(event) {
 /**
  * Hide or show line for the review (Dashboard) page
  *
- * Called only by the Startup function at the top of this file
+ * This function is the click handler for the .hide-review button.
  */
-function toggleReviewLine() {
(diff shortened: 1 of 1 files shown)


macchiati:
I'm marking the ticket as Accepted so that we can get this tested on
smoketest asap. Though this was a joint effort, marking Steven as the
Assignee since he has the PR.

Mark

srl295:
Merged. Mark can you check?

Need to do some errands but will check in later

btangmu:
On smoketest, I logged in as the user with id 2283 and accessed the Sindhi Dashboard successfully. That's something at least, though we hadn't checked whether the problem occurred on smoketest for that user. The smoketest db isn't up to date with production -- but neither is my localhost db, and I was able to repro the problem on localhost.

srl295:
sounds good, let me know if it affects the local test

Kristi27:
Repoduced by Sindhi particular vetter's account only

PR #582, opened 2020-07-25
CLDR-13980 Add kk_Arab-CN locale

MuratKaribai:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13980


MuratKaribai:
@macchiati I created a file with minimal data. Is it possible to add them to the Survey Tool? I would like to make it to the CLDR 38 version

PR #585, opened 2020-07-27

btangmu:
-This PR has only preliminary minor improvements and clean-up

-ChartDelta: report when finished and report file-creation directories

-Remove duplicate new PathDescription from ExampleGenerator.java

-Avoid deprecation warning for DtdData.getInstance; cache getCldrBaseDirectory as CLDR_BASE_DIR

-Refactor to obviate single-line getValueSplitter; call DtdData.getValueSplitter directly instead

-Remove dead/unused code from ChartDelta.java and Chart.java, make private where possible

##### Checklist

### DIFF
--- a/tools/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -2091,8 +2091,6 @@ public class ExampleGenerator {
             this.pathDescription = new PathDescription(supplementalDataInfo, englishFile, extras, starredPaths,
                 PathDescription.ErrorHandling.CONTINUE);
 
-            this.pathDescription = new PathDescription(supplementalDataInfo, englishFile, extras, starredPaths,
-                PathDescription.ErrorHandling.CONTINUE);
             if (helpMessages == null) {
                 helpMessages = new HelpMessages("test_help_messages.html");
             }
--- a/tools/java/org/unicode/cldr/tool/Chart.java
+++ b/tools/java/org/unicode/cldr/tool/Chart.java
@@ -100,7 +100,7 @@ public abstract class Chart {
         standardFooter(pw, AnalyticsID.CLDR);
     }
 
-    enum AnalyticsID {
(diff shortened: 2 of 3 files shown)


btangmu:
This PR has only preliminary minor improvements and clean-up. It takes into account the discussions in https://github.com/unicode-org/cldr/pull/573

btangmu:
> > This PR has only preliminary minor improvements and clean-up. It takes into account the discussions in #573
> 
> Cleanup LGTM but might be good to change the commit/PR title to reflect its state, since there's no "disruptive data churn report' generated by this commit…

I almost always use the title of the issue as the title of the PR, which is the default in github. Then I add details in the subsequent lines. This seems helpful for mentally associating the PR with the ticket. Is there a set of guidelines or best practices we should follow for this kind of thing?

srl295:
> > > This PR has only preliminary minor improvements and clean-up. It takes into account the discussions in #573
> > 
> > 
> > Cleanup LGTM but might be good to change the commit/PR title to reflect its state, since there's no "disruptive data churn report' generated by this commit…
> 
> I almost always use the title of the issue as the title of the PR, which is the default in github. Then I add details in the subsequent lines. This seems helpful for mentally associating the PR with the ticket. Is there a set of guidelines or best practices we should follow for this kind of thing?

That's generally a good practice, my only point is that if the commit does NOT actually generate a disruptive churn report, the commit (and PR) could reflect that. Otherwise we will have this commit `CLDR-13306 Generate disruptive data churn report` that doesn't actually generate such a report. It's minor, but will add confusion later if we are trying to find which commit actually added that report.

I would suggest that the commit be named something like "cleanup ChartDelta and ExampleGenerator"?

btangmu:
> I would suggest that the commit be named something like "cleanup ChartDelta and ExampleGenerator"?

OK, I'll do that. It means omitting the ticket name, so I have mixed feelings about it. Normally I have the ticket name as the first line of the commit (and PR) message, and leave it to subsequent lines to say what the commit actually does.

PR #590, opened 2020-07-29
CLDR-13616 Monitor Forum Participation

btangmu:
-New column: Needing action

-Since slower now, revise implementation to use ajax and show Loading... message

-For server, use SurveyAjax instead of SurveyMain; return json, build table on client

-New special: forum_participation

-New special/forum_participation.js and CldrStForumParticipation.js

-More unit testing

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13616


tools/cldr-apps/WebContent/js/special/forum_participation.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+/**
+ * Based on specials/forum.js
+ *
+ * TODO: use modern JavaScript -- import/export, not define
```
srl295, line 6: 👍

tools/cldr-apps/src/org/unicode/cldr/web/SurveyForumParticipation.java
```diff
@@ -138,6 +132,133 @@ private int queryForumDb(Cell cell, String loc) throws SQLException {
...
+     * @param conn the Connection
+     * @param forumRootId the id of the root (initial) forum post of the thread
+     * @return true or false
+     * @throws SQLException
+     */
+    private boolean ourOrgPostedLast(int forumRootId) throws SQLException {
+        boolean result = false;
+        PreparedStatement ps = null;
+        String sql = "SELECT " + userTable + ".org "
+            + " FROM " + forumTable
```
srl295, line 211: Ideally these would be `?` params. But ok

btangmu: I'm not sure what benefit that would have. The table names don't vary for these queries. Would there be a performance benefit in spite of that?

srl295: @btangmu sorry, correct these are table names, not parameters.

PR #593, opened 2020-07-29
CLDR-13771 Ansible playbook and docs for setup of Survey Tool.

srl295:
CLDR-13771
- For now, this targets cldr-ref which is a test machine.
- add a lint step

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -1,0 +1,30 @@
+name: Ansible Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        lfs: false # not needed for this job, as we don’t currently do a Java build
+    - name: Lint Ansible Playbook
+      uses: ansible/ansible-lint-action@master
+      with:
+        targets: |
+          tools/scripts/ansible/setup-playbook.yml
+        override-deps: |
+          ansible==2.9
+          ansible-lint==4.2.0
+        args: ""
+  build:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        lfs: false # not needed for this job, as we don’t currently do a Java build
+    - name: Install Vagrant and Ansible
+      run: brew install ansible
+    - name: Try the provision
+      run: cd tools/scripts/ansible && ansible-galaxy install -r roles.yml && ln -sv test-local-vars local-vars && vagrant up
--- a/tools/scripts/ansible/.gitignore
+++ b/tools/scripts/ansible/.gitignore
@@ -1,0 +1,3 @@
+/local-*
+/.vagrant
+/*.log
--- a/tools/scripts/ansible/README.md
+++ b/tools/scripts/ansible/README.md
@@ -1,0 +1,87 @@
+# Ansible for Survey Tool
+
+These are ansible scripts for setup and maintenance of the Survey Tool.
+
+## Scope
+
+Right now, the test setup mostly controls Tomcat, but not the nginx proxy
+due to public port issues (https).
+
+## Setup
+
+### Setup: Control system
+
+This is your local system, where you control the others from.
+
+- Install Ansible <https://ansible.com>
+- Install some prereqs:
+
+```shell
+ansible-galaxy install -r roles.yml
+```
+
+- Make sure you can `ssh` into all of the needed systems. For example, `ssh cldr-ref.unicode.org` should succeed without needing a password.
+
+- You should be able to run `ansible all -m ping` and get something back like the following:
+
+```shell
+cldr-ref.unicode.org | SUCCESS => {
+    "ansible_facts": {
+        "discovered_interpreter_python": "/usr/bin/python"
+    },
+    "changed": false,
+    "ping": "pong"
+}
+```
+
+### Setup: Managed systems
+
+- Install python3. Make sure `python --version` returns "Python 3…"
+
+### Setup: Config file
+
+- Create a file `local-vars/local.yml` with the following (but with a secure password instead of `hunter42`.)
+
+```yaml
+mysql_users:
+  - name: surveytool
+    host: localhost
+    password: hunter42
+    priv: 'cldrdb.*:ALL'
+surveytooldeploy:
+  password: hunter43
+  vap: hunter44
+  testpw: hunter45
+  oldversion: 36
+  newversion: 37
+```
+
+## Configure
+
+Run the setup playbook.
+
+```shell
+ansible-playbook --check setup-playbook.yml
+```
+
+This is in dry run mode. When it looks good to you, take the `--check` out and run it again.
+
+## Local Test
+
+- install vagrant and some provider such as virtualbox or libvirt
+
+```shell
+vagrant up
+```
+
+- To log into the new host, run `vagrant ssh`
+
+- To iterate, trying to reapply ansible, run `vagrant provision --provision-with=ansible`
+
+- to deploy ST to this, use:
+
+```shell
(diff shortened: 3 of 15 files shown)


srl295:
So what I have working is a completely automated setup & deployment of an Ubuntu system, from scratch, to hosting the Survey Tool.
  http://cldr-ref.unicode.org:8080/cldr-apps/

HTTPS/nginx is not part of this yet, it's a little harder to test due to letsencrypt.
What this means is we will be able to:- spin up new SurveyTools quickly if needed…- run tests against the same environment used in production- setup a production-like environment, right on your own desktop:  ( vagrant up && curl )

macchiati:
Great!

Mark

srl295:
No impact on production. Ready to review!

btangmu:
Looks good! I understand there should be no effect on the production server.

What's the expected effect of the new file in .github/workflows on our regular github PR process? Could it cause a slowdown or interruption if anything goes wrong running it, for example if cldr-ref is down? When there are more than one file in .github/workflows, do they run sequentially or in parallel?

Although I'm not experienced with ansible or yml or j2, I'll be happy to approve the PR if we're clear that the risk is minimal. I guess we could always quickly disable the new workflow file if needed.

.github/workflows/ansible-lint.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Lint Ansible Playbook
+      uses: ansible/ansible-lint-action@master
+      with:
+        targets: |
+          tools/scripts/ansible/setup-playbook.yml
+        override-deps: |
+          ansible==2.9
+          ansible-lint==4.2.0
+        args: ""
+  build:
```
echeran, line 21: Does the build job depend on the lint job succeeding, or are they independent?  Asking b/c I don't know Ansible, so I can't say if linting is more like validation / error-prone checking or if it is more like formatting style.

If it linting is more than just strongly recommended (ex: in the ICU4X case, we treat warnings like errors), then you could make the build job depend on the lint job.

But if the the build job or lint job take a long time, and/or one job's input does not necessarily depend on the success of output of the other job, then it makes sense to keep them un-linked as deps.

srl295:
> Looks good! I understand there should be no effect on the production server.
>
> What's the expected effect of the new file in .github/workflows on our regular github PR process? Could it cause a slowdown or interruption if anything goes wrong running it, for example if cldr-ref is down? 

None. They aren't required to pass in order to merge. In fact, if you expand the "checks" on a commit, only "jira-ticket" and "license/cla" are required, not even travis. 

It doesn't deploy to cldr-ref, it deploys to a temporary VM owned by github and then shuts it down. Nothing here actually deploys to our resources. 

> When there are more than one file in .github/workflows, do they run sequentially or in parallel?

parallel.
 
> Although I'm not experienced with ansible or yml or j2, I'll be happy to approve the PR if we're clear that the risk is minimal. I guess we could always quickly disable the new workflow file if needed.

thanks.  .j2 is just a template format like jsp, you will recognize the .xml and .properties files that they generate.

.github/workflows/ansible-lint.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        lfs: false # not needed currently
```
echeran, line 11: Would the comment still make sense if it were reworded as "not needed for this job"?  If so, that might be clearer, and then ditto for line 26.

.github/workflows/ansible-lint.yml
srl295: > Does the build job depend on the lint job succeeding, or are they independent? Asking b/c I don't know Ansible, so I can't say if linting is more like validation / error-prone checking or if it is more like formatting style.

They are independent.   Amusingly, only the _mac_ Github VMs have VT-x enabled to sublaunch an Ubuntu VM with vagrant. (Go figure, that's what Github support said somewhere.)  But the ansible-lint action is only Linux.

The linting is validation/error prone and deprecation checking. It “made” me rewrite a 'notifier' to become a 'handler'.  The build step _could_ pass if lint fails, but at this point I'd like to keep the linter happy since i'm newer to ansible. (I have been using it personally for a couple of years, but not as part of CD.)

>  If it linting is more than just strongly recommended (ex: in the ICU4X case, we treat warnings like errors), then you could make the build job depend on the lint job.

Does that mean running them in series?

> But if the the build job or lint job take a long time, and/or one job's input does not necessarily depend on the success of output of the other job, then it makes sense to keep them un-linked as deps.

The build job takes a while because it spins up a VM.  Actually, one way I will want to refactor this, and I'm not sure entirely how, is to make the 'build' step deploy the `cldr-apps.war` artifact from the 'ant' workflow. It seems you [cannot share artifacts across workflows](https://stackoverflow.com/questions/60355925/share-artifacts-between-workflows-github-actions), so this 'build' job may want to move to be a later step in the 'ant' workflow (which you reviewed the other day).  However, it would be good to make the 'ant' workflow more speedy before that, perhaps by making all of the long running tests a parallel step in 'ant'.

I'll do this. I'll draw out a simple flowchart of where I think the build steps and artifacts need to go. One of the tensions here is premature optimization vs. caching.  I wince at all of those GB of data, downloaded anew each time, but I think the reason is because the buildbots are of course stateless independent throwaway units.  So this is kind of the 'new normal' for distributed cloud based building.

.github/workflows/ansible-lint.yml
srl295: Done.  `lfs: false` is the default, but I wanted to make it explicit that this wasn't an oversight here.

.github/workflows/ansible-lint.yml
srl295: @echeran I'm inclined to try to merge this as is if there aren't issues, because it does bring an improvement, and then to iterate.

echeran: Sounds good to me.

.github/workflows/ansible-lint.yml
```diff
@@ -0,0 +1,30 @@
+name: Ansible Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
```
mihnita, line 7: **Nitpick:** do we really want *-latest?

Sure, I can see pros and cons...
But on the other side the ansible version is fixed (in override-deps)
So do we want to be on the latest and greatest everywhere, or nowhere?
I would make everything latest, or everything stable, but not a mixture.

**Note:** in my comments I tag with "Nitpick" stuff that I don't feel strongly about.
Nothing is wrong, it is just a personal opinion.
Implement if you like it, ignore if you don't (there is no need to explain why not :-)

**Disclaimer:** I know nothing about Ansible

mihnita: One more thing: if all my comments are nitpicks, then I usually approve. You can implement or not, your call :-)

srl295: I think latest is probably fine, but i'm not sure.  The actual ansible version is fixed.

echeran: Yeah, I think latest is fine, given it's GH Actions.  For applications, there's the balance between updating versions too frequently (instability and deps hell) and too infrequently (security holes, new features inaccessible).  I think the reason why the model GH Actions is quite different is because each job ("task") is run on a fresh VM, and only for a short amount of time.  The actions are written independent of the OS's bundled libs/deps, and thus actions are independent of each other.  So I think that's why the [selection of OS versions](https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources) are so few.  Slightly unrelated, but interestingly, for paid situations, there's a [1:2:10 ratio](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions) of cost among the different OSes.

srl295:
@btangmu @macchiati can I get an official review on the strength of the other ones here?

PR #594, opened 2020-07-30
CLDR-14016 fix alias rules

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14016


macchiati:
Note to reviewers: will add comments to the ticket about the changes in the data.

common/supplemental/supplementalMetadata.xml
```diff
@@ -1572,11 +1585,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 						
 			
             <!-- variants -->
-            <variantAlias type="AALAND" replacement="AX" reason="deprecated"/>
-            <variantAlias type="POLYTONI" replacement="POLYTON" reason="deprecated"/>
-            <variantAlias type="HEPLOC" replacement="ALALC97" reason="deprecated"/>
-			<variantAlias type="arevela" replacement="hy" reason="deprecated"/> <!-- arevela -->
-			<variantAlias type="arevmda" replacement="hyw" reason="deprecated"/> <!-- arevmda -->
+            <variantAlias type="aaland" replacement="AX" reason="deprecated"/>
```
FrankYFTang, line 1588: Should we keep the VARIANT in this file always in UPPER case? since we already use lower case for language code, UPPER case for territory code, and Capital case for Script code in this file already.

common/testData/localeIdentifiers/localeCanonicalization.txt
```diff
@@ -0,0 +1,1647 @@
...
+che	;	ce
+chi	;	zh
+chu	;	cu
+chv	;	cv
+cjr	;	mom
+cka	;	cmr
+cld	;	syr
+cmk	;	xch
+cmn	;	zh
+cnr	;	sr_ME
```
FrankYFTang, line 103: I believe this should be cnr ; sr_Latn_ME because we also have
    		<languageAlias type="sr_ME" replacement="sr_Latn_ME" reason="legacy"/>
in the data.

common/supplemental/supplementalMetadata.xml
FrankYFTang, line 1588: About <variantAlias type="aaland" replacement="AX" reason="deprecated"/>
how could the code figure the AX not not a language code "ax" but a region code "AX" here. 
Should this be <variantAlias type="aaland" replacement="und-AX" reason="deprecated"/>
instead?

common/testData/localeIdentifiers/localeCanonicalization.txt
```diff
@@ -0,0 +1,1647 @@
...
+en_aaland	;	en_AX
+en_arevela	;	en
+en_arevmda_arevela	;	en
+en_lojban	;	en
+hy_arevela	;	hy
+hy_arevmda	;	hyw
+hy_arevmda_arevela	;	hyw
+hye_arevmda	;	hyw
+no_bokmal_nynorsk	;	nb
+no_nynorsk_bokmal	;	nb
```
FrankYFTang, line 39: This imply, in the algorithm, we need to sort the variant first before look up , so we will first search for no_bokmal before searching no_nynorsk, right?

common/supplemental/supplementalMetadata.xml
```diff
@@ -1572,11 +1585,8 @@ For terms of use, see http://www.unicode.org/copyright.html
...
+            <variantAlias type="polytoni" replacement="polyton" reason="deprecated"/>
```
FrankYFTang, line 1589: if the type in variantAlias is containing not just variant code, but any of l,s,r,v combination, then these should be
<variantAlias type="aaland" replacement="und-AX" reason="deprecated"/>
<variantAlias type="polytoni" replacement="und-polyton" reason="deprecated"/>
<variantAlias type="heploc" replacement="und-alalc97" reason="deprecated"/>

common/testData/localeIdentifiers/localeCanonicalization.txt
```diff
@@ -0,0 +1,1647 @@
...
+bod_Adlm_AC_fonipa	;	bo_Adlm_AC_fonipa
+bos_Adlm_AC_fonipa	;	bs_Adlm_AC_fonipa
+bre_Adlm_AC_fonipa	;	br_Adlm_AC_fonipa
+bul_Adlm_AC_fonipa	;	bg_Adlm_AC_fonipa
+bur_Adlm_AC_fonipa	;	my_Adlm_AC_fonipa
+bxk_Adlm_AC_fonipa	;	luy_Adlm_AC_fonipa
+bxr_Adlm_AC_fonipa	;	bua_Adlm_AC_fonipa
+cat_Adlm_AC_fonipa	;	ca_Adlm_AC_fonipa
+ccq_Adlm_AC_fonipa	;	rki_Adlm_AC_fonipa
+cel_Adlm_AC_fonipa_gaulish	;	xtg_Adlm_AC_fonipa
```
FrankYFTang, line 1240: Also, could you explain to me why cel_Adlm_AC_fonipa_gaulish	will be canonicalized into xtg_Adlm_AC_fonipa ?

common/supplemental/supplementalMetadata.xml
```diff
@@ -167,6 +167,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <languageAlias type="cnr" replacement="sr_ME" reason="legacy"/> <!-- Montenegrin -->
```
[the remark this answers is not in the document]
macchiati: got these, thanks!

common/supplemental/supplementalMetadata.xml
macchiati: Yes, that would be better.

common/supplemental/supplementalMetadata.xml
macchiati: Maybe the best approach is that languageAlias has language *tags* as both type and replacement. Others must never change type. That is, instead of 
<variantAlias type="aaland" replacement="und-AX" reason="deprecated"/>
we should have:
<languageAlias type="und_aaland" replacement="und_AX" reason="deprecated"/>

common/testData/localeIdentifiers/localeCanonicalization.txt
macchiati: Good catch. The code doesn't yet handle cases where items are split off with _x_. All the other cases are have type values that are not well-formed, so they aren't tested.

common/testData/localeIdentifiers/localeCanonicalization.txt
macchiati: Yes, that's in the notes. You have to sort first by the number of fields then alphabetically. Of course, the implementation can use a much faster algorithm than CLDRs with a speedier lookup; this is to be a simple reference implementation.

common/testData/localeIdentifiers/localeCanonicalization.txt
```diff
@@ -0,0 +1,1647 @@
...
+bod	;	bo
+bos	;	bs
+bre	;	br
+bul	;	bg
+bur	;	my
+bxk	;	luy
+bxr	;	bua
+cat	;	ca
+ccq	;	rki
+cel_gaulish	;	xtg
```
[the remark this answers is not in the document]
macchiati: See above. I think the best solution is just change to map to xtg. The cel_gualish doesn't carry useful information

common/testData/localeIdentifiers/localeCanonicalization.txt
macchiati: sr_Latn_ME ~ sr_ME, since sr in ME is written in Latn.

FrankYFTang:
Mark- I change my https://github.com/unicode-org/icu/pull/1254 to match your data and test cases and I am able to make the C++ parts passed all the tests in this PR. The java part is not done yet and I still plan to do some major refactoring on the C++ code to make it cleaner, but please take a look at the tests file I changed, in particular those test  data which was there before but I need to change to match your data change. Would those make sense? (Ignore anything in the Java part, I was half way during the weekend and switch to work on the C++ to match this PR)

macchiati:
Whoa, you are fast!

Your changes look good, both code and data.

Mark

FrankYFTang:
So.. with your change of the data, now we only have the following combination of fields in the type of languageAlias:
language_Script
language_COUNTRY
language_variant
language
und_variant

I wonder could we LIMIT the UTS35 algorithm to only looking at these 5 combinations while we look into languageAlias?

FrankYFTang:
Mark, I also have problem to figure out how to match the following entries in the file. I wonder should we remove them:

```
            <languageAlias type="sgn_BE_FR" replacement="sfb" reason="deprecated"/> <!-- Belgian-French Sign Language -->
            <languageAlias type="sgn_BE_NL" replacement="vgt" reason="deprecated"/> <!-- Belgian-Flemish Sign Language -->
            <languageAlias type="sgn_CH_DE" replacement="sgg" reason="deprecated"/> <!-- Swiss German Sign Language -->

````
If we got a sgn_BE_FR, then we have sgn as language and BE as region, but then what is FR? 
similar question to 
```
<languageAlias type="zh_min_nan" replacement="nan" reason="deprecated"/>
```

File: tools/cldr-unittest/src/org/unicode/cldr/unittest/TestLsrvCanonicalizer.java [hunk 8]
```diff
@@ -0,0 +1,90 @@
...
+//            if (!rule.regular) {
+//                System.out.println(
+//                    rule.lstrType + "\t➕\t" + rule.typeParts + "\t⇒\t"+rule.replacementParts + "\t➕\t" + rule.reason + "\t" + diff(rule.typeParts, rule.replacementParts));
+//            }
+//        }
+//    }
+
+
+
+//    private static String diff(XLanguageTag tagParts, XLanguageTag replacementParts) {
```
hagbard, line 74: Adding immediately commented out code without any comments declaring its intent is weird.
I mean yes it's sort of clear it's a debug thing, but there's a DEBUG flag above *and* this is test code anyway, so why not leave it in guarded by the debug flag? At least that way it won't rot over time with successive refactorings of the code around it.

tools/java/org/unicode/cldr/util/LanguageTagCanonicalizer.java
```diff
@@ -52,6 +52,9 @@ public LanguageTagCanonicalizer(LstrType lstrType) {
     // TODO, handle variants
     @Override
     public synchronized String transform(String locale) {
+        return transform (locale, OutputOption.ICU_LCVARIANT);
+    }
+    public synchronized String transform(String locale, OutputOption oo) {
```
hagbard, line 57: "oo" doesn't seem like a particularly clear variable name to me. Maybe "outputOpts" ?

tools/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -1073,6 +1075,17 @@ boolean isWellFormed(String candidate) {
         public String toCompatString() {
             return this == region ? "territory" : toString();
         }
+
+        public static LstrType fromString(String rawType) {
+            try {
+                return valueOf(rawType);
+            } catch (Exception e) {
```
hagbard, line 1082: Do the if check first and just don't have a try-catch block.
valueOf() is defined to only throw IllegalArgumentException or NullPointerException and I think you want to complain if you're given null anyway (it's a very different type of failure mode).

tools/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -1073,6 +1075,17 @@ boolean isWellFormed(String candidate) {
         public String toCompatString() {
             return this == region ? "territory" : toString();
         }
+
+        public static LstrType fromString(String rawType) {
```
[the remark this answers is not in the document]
hagbard: Please comment this to say how it conflates "territory" as an alternate identifier for region.

tools/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -1073,6 +1075,17 @@ boolean isWellFormed(String candidate) {
         public String toCompatString() {
             return this == region ? "territory" : toString();
```
hagbard, line 1076: Another way to do this is to store the compat string as a field, but either way it would be nice to see a comment on this public method explaining its *not* the enum name.

tools/java/org/unicode/cldr/tool/GenerateLocaleIDTestData.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map.Entry;
+
+import org.unicode.cldr.util.CLDRPaths;
+import org.unicode.cldr.util.CldrUtility;
+import org.unicode.cldr.util.LsrvCanonicalizer;
+import org.unicode.cldr.util.LsrvCanonicalizer.TestDataTypes;
+import org.unicode.cldr.util.StandardCodes.LstrType;
+import org.unicode.cldr.util.TempPrintWriter;
+
+public class GenerateLocaleIDTestData {
```
hagbard, line 14: JavaDoc?

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestLsrvCanonicalizer.java
```diff
@@ -0,0 +1,90 @@
+package org.unicode.cldr.unittest;
```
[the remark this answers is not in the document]
hagbard: JavaDoc or link to a design doc?

aphillips:
@FrankYFTang 

> If we got a sgn_BE_FR, then we have sgn as language and BE as region, but then what is FR?
> similar question to
> 
> ```
> <languageAlias type="zh_min_nan" replacement="nan" reason="deprecated"/>
> ```

These are all grandfathered tags. The `sgn` ones are part of the `irregular` production in BCP47: these are tags that would not otherwise be legal according to the language tag syntax and their meaning as language tags is assigned at a tag level. The subtags should not be taken to have meaning. `zh-min-nan` is "syntactically legal", as it uses subtags that could be (an undefined) multiple-extlang sequence--but which are not actually *extlang* subtags at all. Tag meaning again is at the tag level. All grandfathered tags are deprecated in favor of modern equivalents.

FrankYFTang:
> @FrankYFTang
> 
> > If we got a sgn_BE_FR, then we have sgn as language and BE as region, but then what is FR?
> > similar question to
> > ```
> > <languageAlias type="zh_min_nan" replacement="nan" reason="deprecated"/>
> > ```
> 
> These are all grandfathered tags. The `sgn` ones are part of the `irregular` production in BCP47: these are tags that would not otherwise be legal according to the language tag syntax and their meaning as language tags is assigned at a tag level. 

I know what these are for, the problem is, in order to look up this data set (all the languageAlias), the algorithm need to use combination of fields of locale to match the type attribute, but there are no "combinations of locale fields" of any locale that will be sgn_BE_FR. And therefore it will never be able to be matched in the algorithm. If the algorithm will never match these entries, then there are no reason to keep these around. Of course, this means IF we want to keep these entries around, we have to specify the text of the algorithm in a way that will be possible to match them, but currently from my understand of what the algorithm Mark is going to put down, it won't. 

>The subtags should not be taken to have meaning. `zh-min-nan` is "syntactically legal", as it uses subtags that could be (an undefined) multiple-extlang sequence--but which are not actually _extlang_ subtags at all. Tag meaning again is at the tag level. All grandfathered tags are deprecated in favor of modern equivalents.

1. It is "syntactically legal" as Language-Tag in BCP47 as stated in https://tools.ietf.org/html/bcp47#section-2.1
2. It is NOT "syntactically legal" as unicode_locale_id in UTS35 as stated in https://unicode-org.github.io/cldr/ldml/tr35.html#Unicode_language_identifier

The algorithm in https://unicode-org.github.io/cldr/ldml/tr35.html#Unicode_language_identifier
(in BOTH [3.2.1 Canonical Unicode Locale Identifiers](https://unicode-org.github.io/cldr/ldml/tr35.html#Canonical_Unicode_Locale_Identifiers) and [BCP 47 Language Tag to Unicode BCP 47 Locale Identifier](https://unicode-org.github.io/cldr/ldml/tr35.html#Language_Tag_to_Locale_Identifier) are operated under the syntax of unicode_locale_id in UTS35, NOT the syntax of Language-Tag in BCP47 . Therefore, in the eyes of the algorithm, zh-min-nan is "syntactically illegal" since it is illegal under unicode_locale_id.
Notice in
 [3.2.1 Canonical Unicode Locale Identifiers](https://unicode-org.github.io/cldr/ldml/tr35.html#Canonical_Unicode_Locale_Identifiers) 

```
A unicode_locale_id can be transformed into canonical form in the following way:
...
Replace aliases in the unicode_language_id and tlang (if any) using the following process:
```
so `zh-min-nan` in this part is "syntactically ILLEGAL" 

and in [BCP 47 Language Tag to Unicode BCP 47 Locale Identifier](https://unicode-org.github.io/cldr/ldml/tr35.html#Language_Tag_to_Locale_Identifier)
```
1. Canonicalize the syntax of the language tag (afterwards, there will be no extlang subtags) as per 3.2.1 Canonical Unicode Locale Identifiers.
2. If the BCP 47 primary language subtag matches the type attribute of a languageAlias element in Supplemental Data, replace the language subtag with the replacement value.
```
Notice the locale identifier is ALREADY Canonicalize the syntax of the language tag per 3.2.1 Canonical Unicode Locale Identifiers in step 1, and therefore when it reach step 2 "If the BCP 47 primary language subtag matches the type attribute of a languageAlias element in Supplemental Data..." we should follow the syntax of unicode_locale_id already and therefore `zh-min-nan`  is also "syntactically ILLEGAL"  in this part.

Be honest,  I think the problem is these two part of text in the spec have major flaw now and won't solve the issues of these 5 special cases. This actually also apply to those entries of type='i_..." in  data because these also are  "syntactically ILLEGAL" under the definition of unicode_locale_id. (although they are "syntactically legal" under BCP47 2.1

FrankYFTang:
Mark, as I stated early, I think any entries which the algorithm in UTS35 won't be able to match should be removed from here (or we should change the algorithm in a away that it will be possible to match):

Therefore, please EITHER
a. remove the following entries which the value in type is illegal per unicode_locale_id (I know they are legal per Language-Tag of https://tools.ietf.org/html/bcp47#section-2.1 but we are operating under unicode_locale_id now, right?), OR
b. point out how could the algorithm create a combination of fields from the locale match the type attribute of them. 

```
            <languageAlias type="i_ami" replacement="ami" reason="deprecated"/> <!-- Amis -->
            <languageAlias type="i_bnn" replacement="bnn" reason="deprecated"/> <!-- Bunun -->
            <languageAlias type="i_hak" replacement="hak" reason="deprecated"/> <!-- Hakka -->
            <languageAlias type="i_klingon" replacement="tlh" reason="deprecated"/> <!-- Klingon -->
            <languageAlias type="i_lux" replacement="lb" reason="deprecated"/> <!-- Luxembourgish -->
            <languageAlias type="i_navajo" replacement="nv" reason="deprecated"/> <!-- Navajo -->
            <languageAlias type="i_pwn" replacement="pwn" reason="deprecated"/> <!-- Paiwan -->
            <languageAlias type="i_tao" replacement="tao" reason="deprecated"/> <!-- Tao -->
            <languageAlias type="i_tay" replacement="tay" reason="deprecated"/> <!-- Tayal -->
            <languageAlias type="i_tsu" replacement="tsu" reason="deprecated"/> <!-- Tsou -->
            <languageAlias type="no_bok" replacement="nb" reason="deprecated"/> <!-- Norwegian Bokmal -->
            <languageAlias type="no_nyn" replacement="nn" reason="deprecated"/> <!-- Norwegian Nynorsk -->
            <languageAlias type="sgn_BE_FR" replacement="sfb" reason="deprecated"/> <!-- Belgian-French Sign Language -->
            <languageAlias type="sgn_BE_NL" replacement="vgt" reason="deprecated"/> <!-- Belgian-Flemish Sign Language -->
            <languageAlias type="sgn_CH_DE" replacement="sgg" reason="deprecated"/> <!-- Swiss German Sign Language -->
...
            <languageAlias type="zh_min_nan" replacement="nan" reason="deprecated"/> <!-- Minnan, Hokkien, Amoy, Taiwanese, Southern Min, Southern Fujian, Hoklo, Southern Fukien, Ho-lo -->
```

Either one of the two actions need to be done, otherwise, these entries won't be able to be used in the algorithm anyway.

common/supplemental/supplementalMetadata.xml
macchiati: fixed.

common/supplemental/supplementalMetadata.xml
```diff
@@ -167,6 +167,7 @@ For terms of use, see http://www.unicode.org/copyright.html
...
             <languageAlias type="no" replacement="nb" reason="legacy"/> <!-- Norwegian -->
             <languageAlias type="tl" replacement="fil" reason="legacy"/> <!-- Tagalog -->
+            <!--  
     		<languageAlias type="az_AZ" replacement="az_Latn_AZ" reason="legacy"/>
```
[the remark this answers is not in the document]
macchiati: Ok, doing that.

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestLsrvCanonicalizer.java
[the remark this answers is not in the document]
macchiati: Added to javadoc

tools/java/org/unicode/cldr/util/StandardCodes.java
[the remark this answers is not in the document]
macchiati: done

macchiati:
> So.. with your change of the data, now we only have the following combination of fields in the type of languageAlias:
> language_COUNTRY
> language_variant
> language
> und_variant
> 
> I wonder could we LIMIT the UTS35 algorithm to only looking at these 5 combinations while we look into languageAlias?

I don't think we want to limit that in the spec, because we might need to change it in the future.

FrankYFTang:
> > So.. with your change of the data, now we only have the following combination of fields in the type of languageAlias:
> > language_COUNTRY
> > language_variant
> > language
> > und_variant
> > I wonder could we LIMIT the UTS35 algorithm to only looking at these 5 combinations while we look into languageAlias?
> 
> I don't think we want to limit that in the spec, because we might need to change it in the future.

If we limit that in the spec now, we could change the spec in the future when we need to, right?
The other possibility is we do not limit the spec as I suggested above, but in ICU we implement only the lookup for these 4 combinations plus assertion to ensure no other combination would be show up in the data. And in case the data show up, we change the implementation to look for the newly arrive combination. That will keep both the spec "open" and the implementation "close" but efficient but still conform to the spec.

FrankYFTang:
IF the purpose (or at least PART OF the purpose) of the algorithm, is to transform locale FROM what is syntactical legal under the definition of Language-Tag in BCP47 as stated in https://tools.ietf.org/html/bcp47#section-2.1 but syntactical illegal under the definition  unicode_locale_id in UTS35 as stated in https://unicode-org.github.io/cldr/ldml/tr35.html#Unicode_language_identifier to legal under the definition  unicode_locale_id in UTS35, then the operation NEED to change to follow under the syntax of Language-Tag in BCP47 instead of unicode_locale_id in UTS35. Otherwise, we won't be able to pick up those differences in the algorithm.

macchiati:
> > @FrankYFTang
> > > If we got a sgn_BE_FR, then we have sgn as language and BE as region, but then what is FR?
> > > similar question to
> > > ```
> > > <languageAlias type="zh_min_nan" replacement="nan" reason="deprecated"/>
> > > ```
> > 
> > 
> > These are all grandfathered tags. The `sgn` ones are part of the `irregular` production in BCP47: these are tags that would not otherwise be legal according to the language tag syntax and their meaning as language tags is assigned at a tag level.
> 
> I know what these are for, the problem is, in order to look up this data set (all the languageAlias), the algorithm need to use combination of fields of locale to match the type attribute, but there are no "combinations of locale fields" of any locale that will be sgn_BE_FR. And therefore it will never be able to be matched in the algorithm. If the algorithm will never match these entries, then there are no reason to keep these around. Of course, this means IF we want to keep these entries around, we have to specify the text of the algorithm in a way that will be possible to match them, but currently from my understand of what the algorithm Mark is going to put down, it won't.

The alias information is intended to also encompass tags and subtags of BCP47 that cannot occur in Unicode language/locale identifiers. That includes deprecated-and-replaced tags in BCP47 such as i-default, so that a BCP47 language tag can be converted into a Unicode BCP 47 locale identifier (http://unicode.org/reports/tr35/#Language_Tag_to_Locale_Identifier).

However, that is a somewhat different process than canonicalizing an already valid Unicode BCP 47 locale identifier, since certain forms like "i-default" won't occur. For the latter process, some items could be filtered out.

Anyway, we haven't yet done the spec changes, which will be part of this, and need to explain the situation.

FrankYFTang:
> > > @FrankYFTang
> > > > If we got a sgn_BE_FR, then we have sgn as language and BE as region, but then what is FR?
> > > > similar question to
> > > > ```
> > > > <languageAlias type="zh_min_nan" replacement="nan" reason="deprecated"/>
> > > > ```
> > > 
> > > 
> > > These are all grandfathered tags. The `sgn` ones are part of the `irregular` production in BCP47: these are tags that would not otherwise be legal according to the language tag syntax and their meaning as language tags is assigned at a tag level.
> > 
> > 
> > I know what these are for, the problem is, in order to look up this data set (all the languageAlias), the algorithm need to use combination of fields of locale to match the type attribute, but there are no "combinations of locale fields" of any locale that will be sgn_BE_FR. And therefore it will never be able to be matched in the algorithm. If the algorithm will never match these entries, then there are no reason to keep these around. Of course, this means IF we want to keep these entries around, we have to specify the text of the algorithm in a way that will be possible to match them, but currently from my understand of what the algorithm Mark is going to put down, it won't.
> 
> The alias information is intended to also encompass tags and subtags of BCP47 that cannot occur in Unicode language/locale identifiers. That includes deprecated-and-replaced tags in BCP47 such as i-default, so that a BCP47 language tag can be converted into a Unicode BCP 47 locale identifier (http://unicode.org/reports/tr35/#Language_Tag_to_Locale_Identifier).


Mark, I understand that. But the current text in the spec in THAT session states:
"1. Canonicalize the syntax of the language tag (afterwards, there will be no extlang subtags) as per 3.2.1 Canonical Unicode Locale Identifiers."
so.... in that text, UNTIL WE CHANGE IT, there are no way we will have "extlang subtags" as it clearly stated "afterwards, there will be no extlang subtags". And this is part of the issue we have to address here.  


> 
> However, that is a somewhat different process than canonicalizing an already valid Unicode BCP 47 locale identifier, since certain forms like "i-default" won't occur. For the latter process, some items could be filtered out.
> 
> Anyway, we haven't yet done the spec changes, which will be part of this, and need to explain the situation.

macchiati:
I'd like to go ahead and merge this; can you take another look?

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestLsrvCanonicalizer.java
macchiati: It's gone.

tools/java/org/unicode/cldr/tool/GenerateLocaleIDTestData.java
macchiati: done

tools/java/org/unicode/cldr/util/LanguageTagCanonicalizer.java
macchiati: I'd like to clean this up later. To much to do right now, and I don't want another cycle.

tools/java/org/unicode/cldr/util/StandardCodes.java
macchiati: done (the doc, and some other changes)

srl295:
@macchiati will need a squash (and i will need to come back to review it)

macchiati:
> LGTM under the assumption that the UTS 35 spec text will address the issues I mentioned in this PR.

Yes, definitely!

macchiati:
> @macchiati will need a squash (and i will need to come back to review it)

Yes, that'd be great. (The PR had gotten into a weird state where every time I made a commit, it would complain about an old commit message, I'd use the line line tool to squash, but then the same thing would happen on the next commit.)

macchiati:
Great, thanks.

And that was fast with the ICU code!

We can also chat about how to optimize the ICU code, if you want.

Mark

PR #595, opened 2020-07-30
CLDR-13654 add missing units (not for translation)

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13654


srl295:
@macchiati  ticket here is marked as done?

PR #596, opened 2020-07-30
CLDR-14000 fix tests for examples

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14000


macchiati:
The change so far checks that we have examples for the new grammar features. It (correctly) fails because we don't yet.
TODO
<… für {0} …>	"//ldml/numbers/minimalPairs/caseMinimalPairs[@case=\"([^\"]*+)\"]",
<Die {0} ist …>	"//ldml/numbers/minimalPairs/genderMinimalPairs[@gender=\"([^\"]*+)\"]",
<feminine>	"//ldml/units/unitLength[@type=\"([^\"]*+)\"]/unit[@type=\"([^\"]*+)\"]/gender",

PR #598, opened 2020-07-30
CLDR-13306 Data churn report: Fix ChartDelta CLDR_BASE_DIR bug

btangmu:
-CLDR_BASE_DIR must be File not String to prevent mixup between two versions of DtdData.getInstance

##### Checklist


btangmu:
Something wrong with this PR due to parentheses after issue number

PR #599, opened 2020-07-30
CLDR-13306 Data churn: Fix ChartDelta CLDR_BASE_DIR bug

btangmu:
CLDR-13306

-CLDR_BASE_DIR must be File not String to prevent mixup between two versions of DtdData.getInstance

##### Checklist


macchiati:
Question: are the different versions of getInstance functionally different?

btangmu:
> Question: are the different versions of getInstance functionally different?

Yes. In the String version, the string is supposed to be a version number, like "37.0" in "cldr-37.0".

PR #601, opened 2020-07-31
CLDR-13948 deploy to cldr-smoke2 on merge

srl295:
CLDR-13948

- restructure the 'ant' workflow, for parallel building
- on push to master, deploy to cldr-ref.

(This is on the main fork because of secrets, so i could try out deployment. )

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -3,8 +3,6 @@
 # TODO:
 # - [ ] parameterize unittest so you can pass specific options (i.e. exhaustive) on manual run
 # - [ ] parameterize datacheck so you can pass specific options (i.e. -z) on manual run
-# - [ ] parallelize build for great speed
-# - [ ] add deployment workflow
 
 # Docs: https://docs.github.com/en/actions/language-and-framework-guides/building-and-testing-java-with-ant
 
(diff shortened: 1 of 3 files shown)


srl295:
When this is merged, it will update https://cldr-ref.unicode.org/cldr-apps/

.github/workflows/ant.yml
```diff
@@ -60,10 +57,105 @@ jobs:
...
+      - test
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download cldr-apps.war
+      uses: actions/download-artifact@v2
+      with:
+        name: cldr-apps
+    - name: Deploy to cldr-ref (experimental)
+      # Only on push.
+      if: ${{ github.event_name == 'push' }}
```
echeran, line 133: Note: this workflow is configured to execute on push events only on the master branch (which is what gets triggered when merging a PR).  Which makes sense, as it does for certain other things (ex: we have our API docs re-generate only on merges to master).

The comment "Only on push" kinda of obfuscates that fact, so maybe reword?

srl295: @echeran the point is that PRs don't cause a deployment, only a merge. The workflow is triggered on pr OR push.

srl295: @echeran updated the wording of the comment.

srl295:
⚠️ the Branch Protection Rules will need to change when this merges, because it reorganizes some of the checks.  For example, `cldr-ant/test` is a new check, which is not (as I write this) required.

I don't think  removed any checks.

srl295:
@macchiati now approved by @echeran , can i get a rubber stamp?

macchiati:
Glad to, but I think you need to add me as a reviewer.

PR #602, opened 2020-07-31

btangmu:
-Add boolean DO_CHURN to ChartDelta; if true do Churn instead of Delta

-Catch NullPointerException thrown by DtdData.isMetadata

-For now, use a simple list of high-level paths rather than a RegexLookup

-Comments

##### Checklist


btangmu:
The zip file attached to the ticket has sample output of "delta" and "churn" between versions 36 and 37.

btangmu:
> code LGTM. Ideally DO_CHURN would be a parameter, etc. as a future step

Agreed.

PR #604, opened 2020-08-01
CLDR-6746 TestPaths to include seed and exemplars

srl295:
CLDR-6746

- TestPaths needed to include seed and exemplars
- two files had deprecate formats, fixed.

I moved the normalization mode out of the deprecated options, and removed the deprecated draftiness on RBNF.

- [ ] Open Q for @macchiati in Jira, should the test enforce deprecation on the deprecated postalcode elements? (could be a separate PR)

### DIFF
--- a/tools/java/org/unicode/cldr/util/DtdData.java
+++ b/tools/java/org/unicode/cldr/util/DtdData.java
@@ -99,7 +99,7 @@ public class DtdData extends XMLFileReader.SimpleHandler {
         CDATA, ID, IDREF, IDREFS, ENTITY, ENTITIES, NMTOKEN, NMTOKENS, ENUMERATED_TYPE
     }
 
-    static final Set<String> DRAFT_ON_NON_LEAF_ALLOWED = ImmutableSet.of("collation", "transform", "unitPreferenceData");
+    static final Set<String> DRAFT_ON_NON_LEAF_ALLOWED = ImmutableSet.of("collation", "transform", "unitPreferenceData", "rulesetGrouping");
 
     public static class Attribute implements Named {
         public final String name;
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3089,7 +3089,6 @@ $Revision$
     <!--@MATCH:literal/NumberingSystemRules, OrdinalRules, SpelloutRules-->
 <!ATTLIST rulesetGrouping draft (approved | contributed | provisional | unconfirmed | true | false) #IMPLIED >
     <!--@METADATA-->
-    <!--@DEPRECATED-->
 
 <!ELEMENT ruleset ( alias | ( rbnfrule*, special* ) ) >
     <!--@ORDERED-->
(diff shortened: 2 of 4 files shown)


seed/rbnf/nci.xml
```diff
@@ -11,7 +11,7 @@ For terms of use, see http://www.unicode.org/copyright.html
         <language type="nci"/>
     </identity>
     <rbnf>
-        <rulesetGrouping type="SpelloutRules" draft="unconfirmed">
```
macchiati, line 14: The better choice is to add rulesetGrouping to DtdData.DRAFT_ON_NON_LEAF_ALLOWED

macchiati: Wherever the test message is surfacing, we should add a pointer to DtdData.DRAFT_ON_NON_LEAF_ALLOWED, but with a caution that it should only be expanded if we are certain that it is ok to have draft on a non-leaf (should be rare).

srl295: > The better choice is to add rulesetGrouping to DtdData.DRAFT_ON_NON_LEAF_ALLOWED

Ok great thanks. I had assumed the behavior was intended, but will fix.

macchiati:
It's normally intended, but there are exceptions for entities that behave
as a(complex) unit.

srl295:
@macchiati new error message (will be fixed by updating DtdData):

> Error: (TestPaths.java:446) Deprecated attribute in data: ldml:rulesetGrouping:draft 	;/Users/srl/src/cldr/seed/rbnf/nci.xml - consider adding to `DtdData.DRAFT_ON_NON_LEAF_ALLOWED` if you are sure this is ok.

But, 3e1fdaae575c42c1f30bf8e51676bd32e1bd1cd9 in CLDR-9228 _explicitly_ set deprecated `draft` status on `rulesetGrouping`. 

I've changed this to respect `DRAFT_ON_NON_LEAF_ALLOWED` in the DTD comment handling, so that `rulesetGrouping`'s draft attribute is deprecated if there are no child elements.

seed/rbnf/nci.xml
srl295: see message below

tools/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -224,7 +224,10 @@ public void addComment(String commentIn) {
                     attributeStatus = AttributeStatus.value;
                     break;
                 case "@DEPRECATED":
-                    isDeprecatedAttribute = true;
+                    if (!name.equals("draft")
```
macchiati, line 227: For the others ("collation", "transform", "unitPreferenceData") I think we explicitly removed the @Deprecated on the element, which is cleaner than having this special test. 

(There is no particular value to distinguishing the "with-children" case from the "without children" case.)

srl295:
Ok so in short: fix the DTD to remove the depreciation from rulesetGrouping?

PR #606, opened 2020-08-04

btangmu:
-Replace DO_CHURN with highLevelOnly in MyOptions

-Encapsulate Delta vs Churn per highLevelOnly: dirName, chartNameCap

-New inner class HighLevelPaths determines which paths are high-level

-Match some paths exactly

-Match other paths with special functions like isHighLevelTerritoryName

-Report on usage of highLevelPaths, with highLevelPathMatched

-Comments

##### Checklist


btangmu:
The file 37.0-2020-08-04.zip attached to the ticket has output churn and delta charts per this PR

btangmu:
See https://github.com/unicode-org/cldr/pull/608

PR #609, opened 2020-08-05
CLDR-14059 cldr-unittest/build.xml -z final_testing instead of -z build

btangmu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14059

### DIFF
--- a/tools/cldr-unittest/build.xml
+++ b/tools/cldr-unittest/build.xml
@@ -189,7 +189,7 @@
 	<target name="datacheck" description="Run the standard data tests"
 		depends="init,build">
 		<antcall target="_dataCheck">
-			<param name="datacheck.arg" value="-S common,seed -e -z BUILD" />
+			<param name="datacheck.arg" value="-S common,seed -e -z FINAL_TESTING" />
 			<param name="datacheck.jvmarg" value="${jvm_options}" />
 		</antcall>
 	</target>
@@ -197,7 +197,7 @@
 	<target name="cldr.datacheck.logged" description="Run the standard data tests"
 		depends="init,build">
 		<antcall target="_dataCheck_logged">
-			<param name="datacheck.arg" value="-S common,seed -e -z BUILD" />
+			<param name="datacheck.arg" value="-S common,seed -e -z FINAL_TESTING" />
 			<param name="datacheck.jvmarg" value="${jvm_options}" />
 		</antcall>
 	</target>


btangmu:
[java] an [Aragonese]	error	▸Date_&_Time|Fields|Relative_Week|Period◂	〈the week of {0}〉	【】	〈la semana de〉	«=»	【】	⁅missing placeholders⁆	❮Error: Need at least 1 placeholder(s), but only have 0. Placeholders are: {{0}={NUMBER_OF_WEEKS}, e.g. “3”}; see <a href='http://cldr.unicode.org/translation/error-codes#missingPlaceholders'  target='cldr_error_codes'>missing placeholders</a>.❯	http://st.unicode.org/cldr-apps/v#/an//65c3a9ea9b7882df

     [java] an [Aragonese]	error	▸Characters|Symbols2|math|≥_-name◂	〈null〉	【】	〈E13.1-200〉	«=»	【】	⁅display collision⁆	❮Error: Can't have same translation as [other-symbol: ‾ -name]. Please change either this name or the other one. See <a target='doc' href='http://cldr.unicode.org/translation/short-names-and-keywords#TOC-Unique-Names'>Unique-Names</a>.❯	http://st.unicode.org/cldr-apps/v#/an//7c95d7975dec07f8	root
...

macchiati:
I think we need to change the order, because otherwise remaining data
errors prevent us from merging necessary code / data changes.

So here's what I think we need for the task order:

1. Set ST to read-only.
2. Export to XML.
3. Resolve remaining errors in tests in XML. Once we are completely
error-free, we:
4. Turn on -z final_testing on the build machine. At this point, any *new*
data errors will prevent checkins.

Mark

srl295:
@macchiati I assume `-z FINAL_TESTING` errors are already a warning? I'm wondering if we should do something like this:

0. (have FINAL_TESTING always enforced for `master`, but `-z build` for any branch named `staging-*` (with `-z FINAL_TESTING` being run as an additional side build)- this can all be done with github actions 
1. Set ST to read-only.
2. Export to XML. — check into branch `staging-38` and NOT to `master`
3. Resolve remaining errors in tests in XML, on the `staging-38`. We can make multiple PRs  Once we are completely
error-free, we:
4. open a new PR (allowing 'multiple commits') for merging the `staging-38` branch back into `master`.  At this point, any *new* data errors will prevent checkins, because builds on `master` are enforced.

macchiati:
That sounds reasonable. Let's run by everyone on Wednesday.

Mark

macchiati:
Steven, can you file a ticket for your comment on Aug 17?

PR #612, opened 2020-08-06
CLDR-14061 Copy from VXML for version 38

btangmu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14061
- [x] Updated PR title and link in previous line to include Issue number

PR #613, opened 2020-08-06
CLDR-14061 Generate VXML for version 38

btangmu:
-Skip non-xml files like .DS_Store

-Do not generate en.xml or root.xml

-Log when finished

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14061

### DIFF
--- a/tools/cldr-apps/src/org/unicode/cldr/web/OutputFileManager.java
+++ b/tools/cldr-apps/src/org/unicode/cldr/web/OutputFileManager.java
@@ -8,6 +8,7 @@
 package org.unicode.cldr.web;
 
 import java.io.File;
+import java.io.FileFilter;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -104,6 +105,13 @@ public class OutputFileManager {
     public static final String PXML_PREFIX = "/pxml/main";
     public static final String TXML_PREFIX = "/txml/main";
     public static final String RXML_PREFIX = "/rxml/main";
+    private static final FileFilter xmlFileFilter = new FileFilter() {
+        @Override
+        public boolean accept(File file) {
+            String s = file.getName().toLowerCase();
+            return s.endsWith(XML_SUFFIX) && !"en.xml".equals(s) && !"root.xml".equals(s);
+        }
+    };
 
     /**
      * Names of directories
@@ -205,6 +213,7 @@ public class OutputFileManager {
                     ofm.verifyAllFiles(out, vxmlDir);
                 }
             }
+            System.out.println("outputAndVerifyAllFiles finished");
         } catch (Exception e) {
             System.err.println("Exception in outputAndVerifyAllFiles: " + e);
             e.printStackTrace();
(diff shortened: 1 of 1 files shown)


btangmu:
Running on localhost with these changes, I get "VXML verification succeeded", and the resulting files (in vetdata-2020-08-06T18-19-15-605Z) are identical with those in https://github.com/unicode-org/cldr/pull/612 (from vetdata-2020-08-05T19-56-49-323Z) except that en.xml and root.xml are not generated:

$ diff -r vetdata-2020-08-05T19-56-49-323Z vetdata-2020-08-06T18-19-15-605Z
Only in vetdata-2020-08-05T19-56-49-323Z/pxml/common/main: en.xml
Only in vetdata-2020-08-05T19-56-49-323Z/pxml/common/main: root.xml
Only in vetdata-2020-08-05T19-56-49-323Z/vxml/common/annotations: en.xml
Only in vetdata-2020-08-05T19-56-49-323Z/vxml/common/annotations: root.xml
Only in vetdata-2020-08-05T19-56-49-323Z/vxml/common/main: en.xml
Only in vetdata-2020-08-05T19-56-49-323Z/vxml/common/main: root.xml

tools/cldr-apps/src/org/unicode/cldr/web/OutputFileManager.java
```diff
@@ -635,9 +654,8 @@ private boolean verifyParentChildSameDirectory(Writer out, File vxmlDir)
                     }
                     CLDRLocale parLoc = childLoc.getParent();
                     if (parLoc != null) {
-                        // String parentName = fileName.replaceFirst("_[a-zA-Z]+\\.xml$", "\\.xml");
                         String parentName = parLoc.toString() + XML_SUFFIX;
-                        if (!childName.equals(parentName) && !"root.xml".equals(parentName)) {
+                        if (!childName.equals(parentName) && !"en.xml".equals(parentName) && !"root.xml".equals(parentName)) {
```
macchiati, line 658: Not sure about this. Can you explain more what is happening?

btangmu: When I first made a change to prevent the generation of en.xml and root.xml, I got this:

Verification failure, child without parent
Child, present: /Users/tbishop/Documents/WenlinDocs/Organizations/Unicode/CLDR_job/vetdata-2020-08-06T17-49-10-961Z/vxml/common/main/en_PR.xml
Parent, absent: /Users/tbishop/Documents/WenlinDocs/Organizations/Unicode/CLDR_job/vetdata-2020-08-06T17-49-10-961Z/vxml/common/main/en.xml

This change prevents that spurious failure, making it OK to have en_PR.xml without en.xml.

macchiati:
On the server you probably need to have everything for testing. Options are:

   1. Instead of generating the en.xml & root.xml files, just use baseline
   to avoid losing comments, OR
   2. Generate the zip file without en.xml & root.xml, OR
   3. When you create a PR to put vxml files into master, don't override
   en.xml & root.xml.

Any of those work, but #3 requires the person doing it to remember to avoid
that step, so is a bit more fragile.

Mark

btangmu:
> On the server you probably need to have everything for testing.

I don't think so, unless you mean some later stage of testing. For generating the vxml, the verification succeeds with this PR. If en.xml or root.xml are needed for any testing, either the ones in baseline (cldr/common) could be used directly, or they could be copied from baseline into vxml when necessary. Why waste time generating superfluous and problematic variants of en.xml and root.xml?

macchiati:
> On the server you probably need to have everything for testing.
I misunderstood. I was trying to outline all the options for fixing the
problem, but your approach works, so approving.
Mark

PR #614, opened 2020-08-06
CLDR-14060 fix errors from xml

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14060


macchiati:
See email

macchiati:
Kristi, Tom and Steven are both OOO. Can you approve this so that subsequent error fixes can build on it?

macchiati:
NOTE: this does not fix all the errors, but gives us a base for continuing to fix the others.

macchiati:
Yes, those were some mistakes that happened when adding the characters. The
codes for emoji are arbitrary, as long as they don't collide... but these
did, causing the errors.

Thanks!

Mark

macchiati:
Hmm. had to squash to fix a commit message. Can you re-review?

PR #616, opened 2020-08-07
CLDR-14060 fix chart generation (churn)

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14060
- [ ] Updated PR title and link in previous line to include Issue number

PR #617, opened 2020-08-10

btangmu:
-Add main with additional boolean arg for highLevelOnly

-Call from GenerateAllCharts for both delta and churn

##### Checklist


btangmu:
Mark, I imitated your https://github.com/unicode-org/cldr/pull/616 for this.

I'm still in the middle of running GenerateAllCharts -- it's taking a long time -- so I can't confirm yet that it works, but this is very close to your changes in PR 61 and should be fine.

PR #618, opened 2020-08-10
CLDR-14066 update charts

macchiati:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14066


macchiati:
This fixes some chart problems, and I also added a patch for the problem with ia.xml (which unblocks other tickets)

macchiati:
Ready for review. Mostly charts work, but a couple of patches to get tests to pass.

macchiati:
BTW, Krisiti, I was able to get all the tsvs to show up as tables except when they were too big, or in one case that I can fix:

https://github.com/unicode-org/cldr-staging/blob/master/docs/charts/38/tsv/locale-missing.tsv

srl295:
@macchiati LGTM but needs squash

srl295:
> i approve but needs to be squashed

i should say "probably" — did you intend for 8 commits? Since there are a lot of things going here

srl295:
FWIW master started failing with 229e26539e8bddba715314cad08defa827607de5 (final testing)

PR #620, opened 2020-08-10
CLDR-14071 BRS A05.7 fix en 001

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14071

### DIFF
--- a/common/main/en_001.xml
+++ b/common/main/en_001.xml
@@ -17,6 +17,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<languages>
 			<language type="mus" draft="contributed">Creek</language>
 			<language type="nds_NL">West Low German</language>
+			<language type="sah">Yakut</language>
 		</languages>
 		<territories>
 			<territory type="BL">St Barthélemy</territory>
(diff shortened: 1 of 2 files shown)


common/main/en_001.xml
```diff
@@ -26,7 +27,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<territory type="PM">St Pierre &amp; Miquelon</territory>
 			<territory type="SH">St Helena</territory>
 			<territory type="UM">US Outlying Islands</territory>
-			<territory type="VC">St Vincent &amp; Grenadines</territory>
+			<territory type="VC">St Vincent &amp; the Grenadines</territory>
```
macchiati, line 30: Change this back by deleting "the "

common/main/en_001.xml
```diff
@@ -1159,6 +1290,15 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<unitPattern count="other">{0} secs</unitPattern>
 				<perUnitPattern>{0}/s</perUnitPattern>
 			</unit>
+			<unit type="duration-microsecond">
+				<displayName>μsecs</displayName>
+				<unitPattern count="one">{0} μs</unitPattern>
+				<unitPattern count="other">{0} μs</unitPattern>
+			</unit>
+			<unit type="graphics-dot">
+				<unitPattern count="one">{0} dots</unitPattern>
```
macchiati, line 1299: "one" must be singular.

macchiati:
By design, we have to move everything new in en_GB up to en_001, otherwise
the other "world English" versions will inherit from en(-US).

There were some other changes that we could back out if you are uneasy. I
think those are mostly in date formats. However, we would also need to back
them out of en_GB, since that is where they are coming form. Alternatively,
we could wait on that part until Peter gets back, since he knows the most
about the date patterns.

Mark

macchiati:
Thanks for the careful review.
Mark

PR #622, opened 2020-08-11
CLDR-13396 change the usage information for zh in MY

macchiati:
This should cause the likely subtags to be correctly generated in a later phase

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13396

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -3559,7 +3559,7 @@ XXX Code for transations where no currency is involved
 		<territory type="MY" gdp="933300000000" literacyPercent="93.1" population="31809700">	<!--Malaysia-->
 			<languagePopulation type="ms" populationPercent="75" officialStatus="official"/>	<!--Malay-->
 			<languagePopulation type="en" populationPercent="21"/>	<!--English-->
-			<languagePopulation type="zh_Hant" populationPercent="17"/>	<!--Chinese (Traditional)-->
+			<languagePopulation type="zh_Hans" populationPercent="17"/>	<!--Chinese (Traditional)-->
 			<languagePopulation type="ta" populationPercent="4.2"/>	<!--Tamil-->
 			<languagePopulation type="iba" populationPercent="2.5"/>	<!--Iban-->
 			<languagePopulation type="jv" populationPercent="1.2"/>	<!--Javanese-->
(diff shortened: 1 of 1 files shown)


common/supplemental/supplementalData.xml
```diff
@@ -3559,7 +3559,7 @@ XXX Code for transations where no currency is involved
 		<territory type="MY" gdp="933300000000" literacyPercent="93.1" population="31809700">	<!--Malaysia-->
 			<languagePopulation type="ms" populationPercent="75" officialStatus="official"/>	<!--Malay-->
 			<languagePopulation type="en" populationPercent="21"/>	<!--English-->
-			<languagePopulation type="zh_Hant" populationPercent="17"/>	<!--Chinese (Traditional)-->
+			<languagePopulation type="zh_Hans" populationPercent="17"/>	<!--Chinese (Traditional)-->
```
markusicu, line 3562: Please change the comment to match. I would also add something like "See CLDR-13396 for Hans vs. Hant."

You can use git commit --amend to avoid adding a separate commit.

macchiati:
> In ICU we require to squash before merging, so that the commit message is checked, and is reviewable. We did have problems early on with people messing up the commit message when we initially allowed "squash and merge".

Ah, we've been using squash and merge for a bit, and hadn't noticed problems. How was the commit message getting messed up?

markusicu:
> Ah, we've been using squash and merge for a bit, and hadn't noticed problems. How was the commit message getting messed up?

"Squash and merge" gives the user a free-form edit box for the commit message. It's pre-populated from the PR, so *should* have the ticket prefix, but anyone can delete that and write whatever they want. And that's what goes into the repo, bypassing automated checks and human review.

In the early ICU-on-GitHub days we got a few commits with bad commit messages. I don't remember whether we cleaned them up -- doing so changes the commit hashes of all following commits and thus invalidates everyone's repo clones and work-in-progress. So we removed the "squash and merge" option.

Squashing before committing is just a couple of clicks away using Shane's tool linked into the PR UI.

In ICU we care about the cross-linking of tickets and commits. I thought CLDR cared, too.

macchiati:
> I thought CLDR cared, too.

We do, just hadn't realized the problem.

The issue with squashing separately is that I thought you need to
re-request review. If that isn't the case, then it looks clean to not allow
squash & merge and I can bring it up in the TC.

Mark

markusicu:
> The issue with squashing separately is that I thought you need to re-request review. If that isn't the case, then it looks clean to not allow squash & merge and I can bring it up in the TC.

You do need approval after squashing. I send a comment with "lgtm please squash", and after the author squashed and there are no red flags (no new changes, commit message looks good), then I approve the PR. That is, I don't bother formally approving before the squashing.

The whole point is to have an opportunity to double-check the commit message and to have the automated tool complain about it if something went wrong -- before it goes in.

PR #623, opened 2020-08-13
CLDR-13382 Fix unicode.org/cldr link

jwilk:
CLDR-13382 (it can go under this ticket)


jwilk:
CLA is not a hoop I'm willing to jump through.

srl295:
ok

>

PR #624, opened 2020-08-13
CLDR-14085 sv script translation

stenshamn:
sv: simplified and traditional translated out of context - is plural, should be singular

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14085


srl295:
@stenshamn for some reason I can't make Travis recheck, but the build should pass now that #618 is merged.

PR #625, opened 2020-08-14
CLDR-14044 fix debug output in OutputFileManager

srl295:
CLDR-14044

a log statement wasn't behind the right check, nor did it call debug() as the other  instances did. Check context for related similar  calls.


srl295:
rebased to avoid spurious errs

PR #627, opened 2020-08-17
CLDR-14055 Add annotations for new locales

andrew-z:
CLDR-14055 

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-_____

### DIFF
--- a/common/annotations/mt.xml
+++ b/common/annotations/mt.xml
@@ -1,0 +1,2541 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2020 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+
+Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/ for derived annotations.
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="mt"/>
+	</identity>
+	<annotations>
+		<annotation cp="🏻">ton tal-ġilda | tip 1–2</annotation>
+		<annotation cp="🏻" type="tts">ton ċar tal-ġilda</annotation>
+		<annotation cp="🏼">ton tal-ġilda | tip 3</annotation>
+		<annotation cp="🏼" type="tts">ton medju-ċar tal-ġilda</annotation>
+		<annotation cp="🏽">ton tal-ġilda | tip 4</annotation>
+		<annotation cp="🏽" type="tts">ton medju tal-ġilda</annotation>
+		<annotation cp="🏾">ton tal-ġilda | tip 5</annotation>
+		<annotation cp="🏾" type="tts">ton medju-skur tal-ġilda</annotation>
+		<annotation cp="🏿">ton tal-ġilda | tip 6</annotation>
+		<annotation cp="🏿" type="tts">ton skur tal-ġilda</annotation>
+		<annotation cp="😀">wiċċ | jidħaq ħafna</annotation>
+		<annotation cp="😀" type="tts">wiċċ jidħak ħafna</annotation>
+		<annotation cp="😃">ħalq | jitbissem | wiċċ | miftuħ</annotation>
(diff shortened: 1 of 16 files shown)


andrew-z:
Several unit tests are failing - need to learn how to mitigate

andrew-z:
@macchiati @Kristi27 
I've prepared a first draft of new locale annotations. I solved my issues with Java so I could run tests locally, but doing so lead to a problem: for 3 locales (nso, quc, tn) tests fail because these locales are not present in common\main folder. They are. however, present in seed\main, but I don't understand what I should do to make the test happy with that.

Could you please advise?

andrew-z:
> the files look good. However, some of them are in the wrong place. If the main locale is in seed/main/, such as nso.xml, then the annotations have to be in seed/annotations

Thanks Mark.
I didn't realize there is annotations folder under seed as well... 
Would putting those seed files from me there be enough? I don't understand what's the difference between "seed" and the rest.

macchiati:
Practically speaking, you don't need to know the difference; just that if
you see an XXX.xml in seed/main, then the annotation file XXX.xml must go
into seed/annotations, not in main/annotations.

Mark

Kristi27:
@macchiati Do the  changes look good? 
Are we going to merge this for v38?

common/annotations/mt.xml
```diff
@@ -0,0 +1,2543 @@
...
+		<annotation cp="💪" type="tts">biceps bil-muskoli</annotation>
+		<annotation cp="👂">ġisem</annotation>
+		<annotation cp="👂" type="tts">widna</annotation>
+		<annotation cp="👃">ġisem</annotation>
+		<annotation cp="👃" type="tts">imnieħer</annotation>
+		<annotation cp="🧠">intelliġenti</annotation>
+		<annotation cp="🧠" type="tts">moħħ</annotation>
+		<annotation cp="👀">wiċċ | għajn</annotation>
+		<annotation cp="👀" type="tts">għajnejn</annotation>
+		<annotation cp="👁">ġisem</annotation>
```
srl295, line 380: should be għajn not ġisem (body)

common/annotations/mt.xml
```diff
@@ -0,0 +1,2543 @@
...
+		<annotation cp="👁" type="tts">għajn</annotation>
+		<annotation cp="👅">ġisem</annotation>
```
srl295, line 382: ilsien (tongue) not ġisem (body)

common/annotations/mt.xml
srl295, line 380:
```suggestion
		<annotation cp="👁">għajn</annotation>

```

common/annotations/mt.xml
srl295, line 382:
```suggestion
		<annotation cp="👅">ilsien</annotation>

```

as per the tts

common/annotations/mt.xml
```diff
@@ -0,0 +1,2543 @@
...
+		<annotation cp="👅" type="tts">ilsien</annotation>
+		<annotation cp="👄">xufftejn</annotation>
+		<annotation cp="👄" type="tts">ħalq</annotation>
+		<annotation cp="👶">żagħżugħ | tarbija</annotation>
```
srl295, line 386:
```suggestion
		<annotation cp="👶">żagħżugħ | tarbija</annotation>

```

this is more of a tarbija (baby) than a żagħżugħ (youth/youngster)

andrew-z:
@srl295 I've fixed Maltese as you suggested. 
By the way, "single-commit" check encourages me to do force push every time, so I don't understand how can I accept proposed changes (since that would create a second commit). Or maybe I'm using GitHub wrong? 🤔 (my experience is mostly with AzureDevOps...)

andrew-z:
@srl295 As for uniqueness check - I am not confident I can fix all of that 😐 Most of the duplicates (leaving out quc for now) are because of 🆙 and 🔼 having same tts name, which is a very known issue I see (even pointed out in [this page](http://cldr.unicode.org/translation/characters-emoji-symbols/short-names-and-keywords#TOC-Unique-Names)). I am trying to use my linguistic intuition for some cases by replacing "button up" in 🆙 with just "up", but of course I have no confidence I'm doing it correctly. What would be the way to proceed with this in general? Is it still okay to check in with some questionable translations and then get them corrected?

Kristi27:
@andrew-z for timing purposes (we are creating the maintenance branch shortly), @srl295 will do the following:
1. Merge 
2. Run DAIP
3. Generate Derived

srl295:
CLDRModify (DAIP) done in #691

PR #631, opened 2020-08-18
CLDR-14093 BRS Task A05.5 : CLDRModify pass 1 common/subdivisions

btangmu:
-Differences in Copyright year

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14093


btangmu:
I just noticed the instructions (http://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-passes) say "import the Wikidata first". I didn't do that, and don't know what it means.

macchiati:
Deleted that note (we do this process twice, once after we bring in the
data, and then once later on).

Mark

btangmu:
> Deleted that note (we do this process twice, once after we bring in the data, and then once later on). Mark

I see. Thanks!

PR #638, opened 2020-08-19
CLDR-12010 add plural operand.

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-12010

### DIFF
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestSupplementalInfo.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestSupplementalInfo.java
@@ -107,7 +107,7 @@ public class TestSupplementalInfo extends TestFmwkPlus {
     }
 
     public void TestPluralSampleOrder() {
-        HashSet<PluralInfo> seen = new HashSet<PluralInfo>();
+        HashSet<PluralInfo> seen = new HashSet<>();
         for (String locale : SUPPLEMENTAL.getPluralLocales()) {
             if (locale.equals("root")) {
                 continue;
@@ -148,7 +148,7 @@ public class TestSupplementalInfo extends TestFmwkPlus {
 
     public void TestPluralRanges() {
         PluralRulesFactory prf = PluralRulesFactory.getInstance(SUPPLEMENTAL);
-        Set<String> localesToTest = new TreeSet<String>(
+        Set<String> localesToTest = new TreeSet<>(
             SUPPLEMENTAL.getPluralRangesLocales());
         for (String locale : StandardCodes.make().getLocaleCoverageLocales(
             "google")) { // superset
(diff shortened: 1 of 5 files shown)


macchiati:
draft for now! need to add code to exclude fr - many from logical group tests, etc.

macchiati:
Note: the new plural operand for French is special. It will fall back to 'other' for translations, and thus has the LogicalGroup isOptional. It is handled specially in ICU, with a new rule variable ('e')

If there are any problems with it in integration, please let me know.

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -107,7 +107,7 @@ public static void main(String[] args) {
     }
 
     public void TestPluralSampleOrder() {
-        HashSet<PluralInfo> seen = new HashSet<PluralInfo>();
```
macchiati, line 110: Eclipse made some format cleanup. The only real change is marked below.

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -561,9 +562,10 @@ public void TestDigitPluralCompleteness() {
             // % 100 = 1
             {"kw", "many", "00,000,0000"},  // n != 1 and n % 100 = 1,21,41,61,81
             {"kw", "zero", "0"},    // n = 0
+            {"fr", "many", ""},    // e is special
```
macchiati, line 565: This is the only real change. It says that fr doesn't have any real samples for 'many'

tools/java/org/unicode/cldr/util/LogicalGrouping.java
```diff
@@ -199,6 +199,13 @@ public static boolean isOptional(CLDRFile cldrFile, String path) {
         if (pluralType.equals("0") || pluralType.equals("1")) return true;
         if (!pluralType.equals("zero") && !pluralType.equals("one")) return false;
 
+        if (pluralType.equals("many")) { // special case for french
+            String localeId = cldrFile.getLocaleID();
+            if (localeId.startsWith("fr")
+                && (localeId.length() == 2 || localeId.charAt(2) == '_'))
```
srl295, line 205: using CLDRLocale this is `if(localeId.childOf(fr))`… where fr is a singleton. Should revisit someday.

srl295:
right, errs are incomplete logical group. If French is excluded from check, will there be real errs?

macchiati:
I'll have to work on it. There is a status where something is optional: if
it is present the others miss be, bit it can be missing. But somehow that
isn't working.

macchiati:
Fixed the test failures. The logic for isOptional was faulty, so fixed that.

PR #639, opened 2020-08-19
CLDR-13141 set szl plurals to same as pl

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13141


macchiati:
Needs more work; extra plural cases are causing logical group failures. Setting back to Draft for now.

PR #642, opened 2020-08-19
CLDR-13936 add yue exemplar characters

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13936


macchiati:
> `[凸 灘 雀]` are in the [CLDR-13936](https://unicode-org.atlassian.net/browse/CLDR-13936)'s description but missing here.

Those were already in the AUX set, and the sets need to be disjoint or the tests fail.

PR #644, opened 2020-08-19
CLDR-13771 playbook to include letsencrypt and deploy script

srl295:
CLDR-13771

this does the letsencrypt/nginx/proxy setup via ansible

- use cldr-smoke2 but leave cldr-ref out
- the autodeployment on merge is in PR#601 [CLDR-13948]

[CLDR-13948]: https://unicode-org.atlassian.net/browse/CLDR-13948

### DIFF
--- a/tools/scripts/ansible/README.md
+++ b/tools/scripts/ansible/README.md
@@ -20,9 +20,11 @@ This is your local system, where you control the others from.
 ansible-galaxy install -r roles.yml
 ```
 
-- Make sure you can `ssh` into all of the needed systems. For example, `ssh cldr-ref.unicode.org` should succeed without needing a password.
+- Make sure you can `ssh` into all of the needed systems. For example,
+`ssh cldr-ref.unicode.org` should succeed without needing a password.
 
-- You should be able to run `ansible all -m ping` and get something back like the following:
+- You should be able to run `ansible all -m ping` and get something back
+like the following:
 
 ```shell
 cldr-ref.unicode.org | SUCCESS => {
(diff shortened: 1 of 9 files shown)


tools/scripts/ansible/setup-playbook.yml
```diff
@@ -81,3 +133,19 @@
...
+  vars_files:
+    - vars/main.yml
+    - local-vars/local.yml
+  tasks:
+    - name: Install certbot packages
+      apt:
+        pkg:
+          - python3-certbot-nginx
+    - name: setup certbot
+      command: sudo certbot --nginx  --agree-tos -m {{ certbot_admin_email }} -d {{ inventory_hostname }} --non-interactive --keep
```
sffc, line 147: All this does is create the nginx config file.  Where do you generate the actual certificate, and how do you point the nginx config file to that certificate?

btangmu: For certbot config, cf. https://unicode-org.atlassian.net/browse/CLDR-13567

btangmu: The current production server uses /etc/nginx/sites-enabled/stproxy

I think certbot automatically generates the certificate if installed right.

srl295: @sffc this does create the certificate and updates the config file, that's why the `python3-certbot-nginx` plugin was installed. https://cldr-smoke2.unicode.org was setup using only this playbook.

srl295:
This now should pass the ansible-lint (vagrant) phase. Please recheck. I think it's ready to go.

tools/scripts/ansible/setup-playbook.yml
```diff
@@ -52,25 +61,69 @@
...
+    - name: Setup surveytool user for deploy
+      user:
+        name: surveytool
+        shell: /bin/bash
+    - name: Give acess to surveytool user
+      file:
+        path: /var/lib/tomcat8/cldr/cldr-trunk
+        owner: surveytool
+        recurse: yes
+    - name: Setup surveytool auth
```
echeran, line 106: The GH Action for Ansible build is currently failing with an error saying that there's an invalid key name in the YAML around somewhere here.

srl295: @echeran please recheck… think i fixed it

tools/scripts/ansible/README.md
```diff
@@ -38,6 +38,28 @@ cldr-ref.unicode.org | SUCCESS => {
...
+```
+
+### Setup: surveytool keypair
+
+```shell
+mkdir -p ./local-vars
+ssh-keygen -t rsa -b 4096 -f ./local-surveytool -P '' -C 'surveytool deploy'
+```
+
+The contents of the `local-vars/surveytool.pub` file is used for the `key:` parameter below in `local.yml`.
```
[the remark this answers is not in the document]
srl295: @echeran fixed in https://github.com/unicode-org/cldr/commit/bc5742a530ed06892b6d456e8b877e5e551cebd7

srl295:
@macchiati @sffc these are now approved by @echeran , can i get a rubber stamp?

PR #646, opened 2020-08-19
CLDR-13918 Switching file/path operations to using normalized rather than canonical paths.

hagbard:
Details in CLDR-13918, but this PR tries to catch every file operation and piece of code where file "canonicalization" was being done, and change it to go via a single utility function (which uses path normalization rather than canonicalization).

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13918


tools/java/org/unicode/cldr/tool/SearchCLDR.java
```diff
@@ -364,16 +365,12 @@ public static void main(String[] args) {
...
         int item = 0;
         for (File s : sourceDirs) {
-            try {
-                String path = s.getCanonicalPath();
-                int baseLoc = path.lastIndexOf("/cldr/");
-                if (baseLoc < 0) {
-                    throw new ICUUncheckedIOException("source doesn't contain /cldr/");
-                }
-                newDirs[item++] = new File(base, path.substring(baseLoc + 5));
-            } catch (IOException e) {
```
hagbard, line 374: No IOException possible using new utilities file.
Here and a few other places in this PR.

tools/java/org/unicode/cldr/util/LDMLUtilities.java
```diff
@@ -1962,15 +1962,7 @@ public static String filenameToURL(String filename) {
...
 
         File f = new File(filename);
-        String tmp = null;
-        try {
-            // This normally gives a better path
-            tmp = f.getCanonicalPath();
-        } catch (IOException ioe) {
-            // But this can be used as a backup, for cases
-            // where the file does not exist, etc.
-            tmp = f.getAbsolutePath();
```
hagbard, line 1972: No longer reachable/applicable since failure cannot occur.

macchiati (dismissed):

hagbard:
Should I wait for other reviewers?

macchiati:
> Should I wait for other reviewers?

No need. You can hit [Squash and merge]

PR #650, opened 2020-08-19
CLDR-13622 fix yoruba exemplars

macchiati:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13622

### DIFF
--- a/common/main/yo.xml
+++ b/common/main/yo.xml
@@ -716,7 +716,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		</codePatterns>
 	</localeDisplayNames>
 	<characters>
-		<exemplarCharacters>[a á à b d e é è ẹ {ẹ\u0301} {ẹ\u0300} f g {gb} h i í ì j k l m n o ó ò ọ {ọ\u0301} {ọ\u0300} p r s ṣ t u ú ù w y]</exemplarCharacters>
+		<exemplarCharacters>[a á à b d e é è ẹ f g gb h i í ì j k l m n ń o ó ò ọ p r s ṣ t u ú ù w y]</exemplarCharacters>
 		<exemplarCharacters type="auxiliary">[c q v x z]</exemplarCharacters>
 		<exemplarCharacters type="index">[A B D E F G H I J K L M N O P R S T U W Y]</exemplarCharacters>
 		<exemplarCharacters type="numbers" draft="contributed">↑↑↑</exemplarCharacters>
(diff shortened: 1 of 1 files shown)


srl295:
@macchiati needs squash

srl295:
### Analysis

Removed: `[{ẹ̀}{ẹ́}{ọ̀}{ọ́}]`
Added: `[ń]`

common/main/yo.xml
```diff
@@ -716,7 +716,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		</codePatterns>
 	</localeDisplayNames>
 	<characters>
-		<exemplarCharacters>[a á à b d e é è ẹ {ẹ\u0301} {ẹ\u0300} f g {gb} h i í ì j k l m n o ó ò ọ {ọ\u0301} {ọ\u0300} p r s ṣ t u ú ù w y]</exemplarCharacters>
+		<exemplarCharacters>[a á à b d e é è ẹ f g {gb} h i í ì j k l m n ń o ó ò ọ p r s ṣ t u ú ù w y]</exemplarCharacters>
```
srl295, line 719:
```suggestion
		<exemplarCharacters>[a á à b d e é è ẹ {ẹ\u0301} {ẹ\u0300} f g {gb} h i í ì j k l m n ń o ó ò ọ {ọ\u0301} {ọ\u0300} p r s ṣ t u ú ù w y]</exemplarCharacters>
```

srl295: per @Moyogo https://unicode-org.atlassian.net/browse/CLDR-13622?focusedCommentId=157872

macchiati: Annemarie was tasked with getting this reviewed, and the linguist feedback was in https://unicode-org.atlassian.net/browse/CLDR-13622?focusedCommentId=157640. That is:
> (3) The nasal n is also marked with a high tone mark in syllabic nasal context (i.e ń.kọ́, ń.lọ) or with a mid tone mark - unmarked as said before (i.e n.kò) in the negative form.
However, it does look like there is good evidence tor retaining the 0300/1 combinations.

AEApple: I'm happy to follow up to get feedback based on the latest comments. I agree though that if there are examples of the characters being used in multiple Yoruba documents and websites we should definitely consider retaining those combinations.

AEApple:
So do we just squash to first two commits? I reviewed and commented on the final commit.

common/main/yo.xml
srl295: @macchiati right, and the feedback uses one of the characters (formerly) removed:

> `(i.e ń.k**ọ́**, ń.lọ)`

I think this PR is now a step forward including the ń and we can update again as there's more input.

PR #651, opened 2020-08-19
CLDR-14080 move 3 locales (plus children) to seed

macchiati:
Also remove Guarani, Caddo from CLDR target locales

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14080


macchiati:
The file changes are straightforward, except that a number of files related to tests also need fixing.

PR #636, opened 2020-08-19
CLDR-13822 Add collation for breton language

sven-oly:
https://unicode-org.atlassian.net/browse/CLDR-13822

PR #652, opened 2020-08-19
CLDR-13822 Add collation for breton language

sven-oly:
https://unicode-org.atlassian.net/browse/CLDR-13822

Same changes as #636, now to master.


srl295:
OK, you fixed #636 here.

PR #655, opened 2020-08-20
CLDR-14112 ISO 4217 Currency Update

srl295:
CLDR-14112

- No new codes


srl295:
YUD is now 890, so added to the suppementalData

srl295:
@macchiati I also updated https://sites.google.com/site/cldr/development/updating-codes/update-currency-codes but was not sure about where to record the new metadata.

PR #665, opened 2020-08-25
Containment a14 1 14116bis

srl295:
CLDR-14116 A14.1 (#664  is A14, UN codes)

PR #666, opened 2020-08-25
CLDR-14116 Include TestLanguageGroup in TestAll, improve infra

srl295:
CLDR-14116

- move TestLanguageGroup.ISOLATES to isolates.json, improve err messages
- add Gson to unittest classpath (oversight)

Won't pass tests as is. See also #665


srl295:
@macchiati can you take a look at this?  I moved the ISOLATES to a .json file for maintenance (would like to do this for other hardcoded lists, so some machinery for that)

But, TestLanguageGroup did not pass by itself.  Adding `pcm` and `und` to `isolates.json` quells a couple of items, but there are other issues. I wonder if `LanguageGroup.LANGUAGE_GROUP` needs a tweak (it should also be data!)

Can you pick this up?

srl295:
@macchiati Oops, I didn't mean to put this on the head fork! But actually, that makes it easier for you to add commits to this branch.

macchiati:
Question, can we change the author in github? That way I'll see it on my plate.

srl295:
> Question, can we change the author in github? That way I'll see it on my plate.

You are now the only owner of the PR.  The `fix-grouptest` branch is in the head fork (unicode-org) and so you can push any commits you want to it.

tools/java/org/unicode/cldr/util/data/languages/isolates.json
```diff
@@ -0,0 +1,3 @@
+[
+	"ko", "qu", "root"
```
macchiati, line 2: Question: what is this for?

macchiati (approved):

srl295:
It was a static member in code, ISOLATES

tools/java/org/unicode/cldr/util/data/languages/isolates.json
srl295: Delayed answer:  it's to move `static final Set<String> ISOLATES` out to a data file.

srl295:
@macchiati I think we've moved on and can close this.

PR #669, opened 2020-08-26
CLDR-14119 A21 Algorithmic Locales

srl295:
CLDR-14119 A21 Algorithmic locales

Two commits here:

- algorithmic locales
- also, make SpecialLocales based off the algorithmic list (it was out of sync). This makes sure the Survey Tool does not allow votes against generated locales!

ALLOW_MANY_COMMITS=true


srl295:
OK so here's the deal.

in yue (Hant) `蟲` is worm  🪱 , `毛蟲` bug 🐛

but going through the T>C transliterator they both turn into `虫` in yue_Hans which is a duplicate.

srl295:
I think this is the right fix below, and matches ICU's T-S transform with the same input data.

So, I'm not sure how this got broken, still investigating.

```patch
-               <annotation cp="🐛" type="tts">虫</annotation>
+               <annotation cp="🐛" type="tts">毛虫</annotation>
```

macchiati (dismissed):

srl295:
@macchiati sorry, missed your review comments?

For short, the cldrmodify dropped the triple-up-arrow entries. Restored here.

For narrow (2nd set of diffs) `sr` is the equivalent as `sr_Latn` here, not sure why `sr` did not get these errors.

```diff
index a9ec946e7d..139c6ec9e5 100644
--- a/common/main/sr_Latn.xml
+++ b/common/main/sr_Latn.xml
@@ -9335,12 +9335,15 @@ CLDR data files are interpreted according to the LDML specification (http://unic
                        </unit>
                        <unit type="length-furlong">
                                <displayName>furlonzi</displayName>
+                               <unitPattern count="one">↑↑↑</unitPattern>
                                <unitPattern count="few">{0} fur</unitPattern>
+                               <unitPattern count="other">↑↑↑</unitPattern>
                        </unit>
                        <unit type="length-fathom">
                                <displayName>hv</displayName>
                                <unitPattern count="one">{0} fth</unitPattern>
                                <unitPattern count="few">{0} fth</unitPattern>
+                               <unitPattern count="other">↑↑↑</unitPattern>
                        </unit>
                        <unit type="length-nautical-mile">
                                <displayName>nmi</displayName>
@@ -9876,6 +9879,18 @@ CLDR data files are interpreted according to the LDML specification (http://unic
                                <unitPattern count="few">{0} cm</unitPattern>
                                <unitPattern count="other">{0} cm</unitPattern>
                        </unit>
+                       <unit type="length-furlong">
+                               <displayName>furlonzi</displayName>
+                               <unitPattern count="one">↑↑↑</unitPattern>
+                               <unitPattern count="few">{0} fur</unitPattern>
+                               <unitPattern count="other">↑↑↑</unitPattern>
+                       </unit>
+                       <unit type="length-fathom">
+                               <displayName>hv</displayName>
+                               <unitPattern count="one">{0} fth</unitPattern>
+                               <unitPattern count="few">{0} fth</unitPattern>
+                               <unitPattern count="other">↑↑↑</unitPattern>
+                       </unit>
                        <unit type="length-millimeter">
                                <displayName>mm</displayName>
                                <unitPattern count="one">{0} mm</unitPattern>
```

macchiati:
Sgtm

macchiati:
Those looks right.

I think the uparrow treatment predates the change to cldr staging. Not a
serious problem for this release if you can fix a small number of entries
by hand, but can you file a ticket to fix for next time?

srl295:
There are more of these, so debuggging a bit. Somehow I thought I had 0 errors locally.

Looks like uparrow is explicitly dropped, I think I can change that to preserve uparrow from source.

srl295:
OK, so I am debugging the 'policy' part. I attempted to document: 

```jave
    enum PolicyIfExisting {
        RETAIN,  // Do not transliterate if existing output has locale content
        DISCARD, // Replace existing output locale content
        MINIMIZE // RETAIN, plus drop values if translit is a no-op.
    }
```

…
```java
        de_CH("de", "[ß] Casefold", Transliterator.FORWARD, "[ß]", PolicyIfExisting.MINIMIZE), //
        yue_Hans("yue", "Simplified-Traditional.xml", Transliterator.REVERSE, "[:script=Hant:]", PolicyIfExisting.RETAIN), //
```

So, is `de_CH` and `yue_Hans` NOT supposed to be readonly? i.e. vetters can vote, but anything "missing" is a candidate for transliteration?    Then I need to change the SpecialLocales code to only make READONLY for the policy `DISCARD`.

I think what I am chasing in yue_Hans is not transliterator failures but existing (possibly) bad data in `yue_Hans.xml`… however it got there (?), with the policy RETAIN anything in yue_Hans stays put, values are only added, never changed by the transform.

However, the red-questionmark and white-questionmark had annotations that just say "questionmark", and now that ascii `?` has an annotation, there's a collision.

macchiati:
Yes, de_CH and yue preserve all old data; they just get an infusion of new
data where there wasn't any. Collisions can happen if a new item appears,
and the work is to resolve.
Mark

srl295:
So we kind of have (in an uneducated guess) a problem in the source Cantonese (yue.xml):

`感歎號` = `!` but
`感嘆號` = `❗`

the middle ideographs are T variants of each other. I'm not sure why `紅色` (red?) wasn't used to distinguish the 2nd one.  There may well be a good reason for it. 

But, both of those strings above simplify to `感叹号` which is a collision.

srl295:
`yue.xml` also uses the same term, `橫線` for several of the hyphen/dash type characters. For some reason, CheckDisplayCollissions does not give an error for `yue` but does for the simplified version.

https://github.com/unicode-org/cldr/blob/master/common/annotations/yue.xml#L28

https://github.com/unicode-org/cldr/blob/master/common/annotations/yue.xml#L30

https://github.com/unicode-org/cldr/blob/master/common/annotations/yue.xml#L36

srl295:
I think I've corrected the remaining issue in Yoruba.  With transliteration, the short dessert spoon value was too wide.

```
2020-08-26T23:16:06.8330819Z      [java] yo_BJ [Yoruba (Benin)]	error	▸Units|Volume|dessert-spoon|short-other-nominative◂	〈{0} dstspn〉	【】	〈{0} àmì shíbí oúnjɛ́ kékeré〉	«=»	【】	⁅value too wide⁆	❮Error: Too wide by about 140% (with common fonts).❯	http://st.unicode.org/cldr-apps/v#/yo_BJ//3eecaafe459d5283
```

It turned out that the 'long' was shorter than the 'short', so I swapped them, also in `yo.xml` for consistency.

srl295:
> Those looks right. I think the uparrow treatment predates the change to cldr staging. Not a serious problem for this release if you can fix a small number of entries by hand, but can you file a ticket to fix for next time?

Done, 

https://unicode-org.atlassian.net/browse/CLDR-14127

macchiati:
Sounds good. In the worst case, we drop bad items.

Running the tools are the easy part; figuring out how to resolve the
resulting errors can be tricky.

Mark

macchiati:
Good, thanks.

Mark

srl295:
I'm impressed that all of this used to be done without git…

srl295:
@macchiati Think we're ready for review and merge. Merge if it seems ok after review?

PR #675, opened 2020-08-27
CLDR-14134 plural update, also throw a better NPE

srl295:
CLDR-14134

- fix Hausa (ha.xml) : add 'ranaku {0}' for plural-other.

##### Checklist


srl295:
Hausa updated chart
image

PR #680, opened 2020-08-28
CLDR-14142 derived annotations

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14142


macchiati:
Haven't checked if tests passed yet...

PR #681, opened 2020-08-28
CLDR-13654 fix blood glucose measures

macchiati:
Still needs work, and looks like it won't make v38. But capturing state here.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13654


tools/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -421,7 +421,7 @@ private void addToSourceToTarget(String source, String target, ConversionInfo in
         sourceToTargetInfo.put(source, new TargetInfo(target, info, inputParameters));
         String targetQuantity = baseUnitToQuantity.get(target);
         if (targetQuantity == null) {
-            throw new IllegalArgumentException();
+            throw new IllegalArgumentException("Can find quantity for baseUnit: " + target);
```
srl295, line 424:
```suggestion
            throw new IllegalArgumentException("Can't find quantity for baseUnit: " + target);
```

PR #686, opened 2020-09-01
CLDR-7798 Prometheus exporter for the Survey Tool

srl295:
CLDR-7798 

- reads ST JSON, re-exports as Prometheus

Seems to be ready to go. This is what brings in the surveytool_* data into prometheus.

- [ ] Operationally, need to arrange for this to stay running on corp.unicode.org (fyi @rickmcgowan )

### DIFF
--- a/tools/st-prom-exporter/.gitignore
+++ b/tools/st-prom-exporter/.gitignore
@@ -1,0 +1,2 @@
+/node_modules
+/config.json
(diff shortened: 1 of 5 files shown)


srl295:
It turns out that changing string labels isn't good, all labels are basically 'distinguishing attributes' in CLDR parlance  <https://github.com/prometheus/prometheus/issues/2227#issuecomment-263201106>

So, I have removed the `err: "message"` type labels. If an err is noted, login to the ST to retrieve it.

macchiati (approved):

PR #688, opened 2020-09-01
CLDR-13948 manual push to production

srl295:
CLDR-13948

Looks like it works. The scripting could be a little simpler, but the basic deploy to production seems to be functional.

You can compare this to #601 it is very similar code, but with the hostname variables being different.
(It's too similar, and should be refactored, but I'd like to make sure it's working first.)

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,0 +1,86 @@
+# Deploy to Production
+
+name: production
+
+# Triggered on push to master,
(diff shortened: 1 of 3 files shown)


srl295:
- [ ] Do not allow going 'back' in the history unless a special checkbox is checked.

srl295:
@macchiati Let's get this in and try it out. Can add more err checking later.

PR #578, opened 2020-07-24
CLDR-10102 Seed latin

doublex:
CLDR-10102

Classical Roman numbers could only go as far as a million. Here we have had to use Neo-Latin term for billion. Minuta and secunda are Neo-Latin, too. Classical Latin could not measure more precisely than in hours. There is also no word for quarter in the sense used here, so we have used a neologism.

PR #691, opened 2020-09-02
CLDR-14138 CLDRModify -fp

srl295:
CLDR-14138

- Also, no longer generate empty /diff/log.txt files.
This log was no longer used, so it was removed from the
code.

Modifications of:
- CLDR-14055 Add annotations for new locales (#627)
- CLDR-10102 Seed latin (#578)

##### Checklist


srl295:
- added derived annotations

common/annotations/my.xml
```diff
@@ -1381,7 +1381,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🐡" type="tts">ငါးပူဖောင်း</annotation>
 		<annotation cp="🦈">ငါး | ငါးမန်း</annotation>
 		<annotation cp="🦈" type="tts">ငါးမန်း</annotation>
-		<annotation cp="🐙">ဘဝှဲ ရဘဝှေဲ ဘဝှဲ ဘဝွဲ ဘဝြဲ ဘဝဲ</annotation>
+		<annotation cp="🐙">ဘဝှဲ ရဘဝှေဲ ဘဝှဲ ဘဝှဲ ဘဝွဲ ဘဝဲ</annotation>
```
macchiati, line 1384: I think you have to revert this change in my.xml. (There was an earlier problem with the Zwagyi "fixer" in DAIP)

common/annotationsDerived/af.xml
```diff
@@ -462,6 +462,86 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="👱🏾" type="tts">persoon met blonde hare: medium-donker velkleur</annotation>
 		<annotation cp="👱🏿">blond | donker velkleur | hare | persoon met blonde hare</annotation>
 		<annotation cp="👱🏿" type="tts">persoon met blonde hare: donker velkleur</annotation>
+		<annotation cp="🧑🏻‍❤‍💋‍🧑🏼">ligte velkleur | medium-ligte velkleur | paartjie | romanse | soen | volwassene</annotation>
```
macchiati, line 465: I think these new ones happened as a result of the updates for Emoji 13.1. My fault for not rerunning; glad you did this.

macchiati (changes requested):
Just one fix necessary; reverting my.xml

srl295:
Understood, will do tomorrow.

Any more steps needed? -fQ? Transformed locales?

macchiati:
None of the new ones are transformed, so that's fine. But, doooh:


   - I should have said you need -fQ also, but just for the annotations
   directory.
   - You don't need to do -fp afterwards, but you do need to do the
   derivedAnnotations.


Mark

srl295:
@pedberg-icu this also modifies la.xml and I think passes tests.

srl295:
> But, doooh: 

From a process improvement perspective, we'll call this a success :)

macchiati:
Yup!

srl295:
@macchiati seems `-fQ` does not work for seed/annotations (`Annotations.LOCALES` is built explicitly from `common/annotations`). Is this a bug, or just skip -fQ for seed?

srl295:
ok, it seems there aren't any derivedAnnotations for seed, for similar reasons…

srl295:
whoops. forgot to exclude `my.xml` doing this now:

- removing my.xml from the commit
- re-running `modify -fQ` but not `modify -fp`
- re-running `GenerateDerivedAnnotations`


And for the record, I did similar for `seed/main/la.xml` - removed it, re-ran `modify -fp`.

common/annotations/my.xml
srl295: fixed in c7b1ea206bbfc2ef399019aafd4350858803a7cc

srl295:
@macchiati 
- `common/main/my.xml` did not seem to be affected by -fp
- reverting `common/annotations/my.xml` as above did not seem to produce a new derived annotation.

srl295:
> @macchiati
> 
> * `common/main/my.xml` did not seem to be affected by -fp
> * reverting `common/annotations/my.xml` as above did not seem to produce a new derived annotation.

@macchiati think we are good to go. https://github.com/unicode-org/cldr/pull/691/files#diff-a874a048457c728fe7277edcbffdb107 does not show any scrambled Myanmar anymore.

macchiati (approved):

PR #697, opened 2020-09-03
CLDR-14157 es_419, fix patterns for yMMM, yMMMd (MMMM -> MMM)

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14157

In es_419, fixed the patterns for gregorian availableFormats skeletons yMMM, yMMMd to use MMM instead of MMMM. All other availableFormats and intervalFormats patterns in all calendars in es_419 already had MMM or MMMM matching the skeleton.

In es_MX,es_US made a similar fix for gregorian availableFormats yMMMEd.

A separate ticket will be filed (for a future release) to adjust use of 'de in these patterns (generally patterns with MMM should not have 'de', patterns with MMMM should).

### DIFF
--- a/common/main/es_419.xml
+++ b/common/main/es_419.xml
@@ -1795,8 +1795,8 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="yMd">d/M/y</dateFormatItem>
 						<dateFormatItem id="yMEd">E d/M/y</dateFormatItem>
 						<dateFormatItem id="yMM">M/y</dateFormatItem>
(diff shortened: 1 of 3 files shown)


macchiati (dismissed):
It also occurred in other children of es.xml; as per discussion in meeting. Those should be fixed too.

pedberg-icu:
> It also occurred in other children of es.xml; as per discussion in meeting. Those should be fixed too.

Oh right, thanks, also fixed for yMMMEd in es_MX, es_US.

srl295 (approved):
No formal issues, lgtm

PR #698, opened 2020-09-03
CLDR-13948 Push to Production Improvements

srl295:
CLDR-13948

Note: I put the branch on the 'head fork', so that it can be tried out (see below)

- “we only go forward”:  for all deployments (smoketest and production), display a list of what changes are about to be incorporated.  Allow a `--override` option if we really want to roll backward. If we roll backward, show what we are reverting!

- updates the deploy script deployed by ansible.  This is where the core implementation is done.

- Add a UI for the above.

-----
## How to try it out
1. go to [⏯️ Actions](https://github.com/unicode-org/cldr/actions?query=workflow%3Aproduction), to the "production" workflow
2.  find the Run Workflow button
3. change the branch to **srl-deploy-13948** (Note: once this is merged, use `master` !) 
4. Should appear as below. Enter in a SHA (hash), or a tag, or `master`.  
5. If you need to roll back (i.e. go to a commit PRIOR to the current deployment) add `--override` in the 2nd box.
6. click **Run Workflow**
7. Watch the build status, and then check out https://st2.unicode.org/cldr-apps/ for the result.

image

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -146,4 +146,4 @@ jobs:
       run: |
         echo "${RSA_KEY_SURVEYTOOL}" > .key && chmod go= .key
         echo "${SMOKETEST_KNOWNHOSTS}" > .knownhosts && chmod go= .knownhosts
-        ssh -C -o UserKnownHostsFile=.knownhosts -i .key -p ${SMOKETEST_PORT} surveytool@${SMOKETEST_HOST} sh /usr/local/bin/deploy-to-tomcat.sh < cldr-apps.war ${GITHUB_SHA}
+        ssh -C -o UserKnownHostsFile=.knownhosts -i .key -p ${SMOKETEST_PORT} surveytool@${SMOKETEST_HOST} bash /usr/local/bin/deploy-to-tomcat.sh < cldr-apps.war ${GITHUB_SHA} --override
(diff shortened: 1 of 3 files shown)


macchiati (dismissed):

macchiati:
Nice

srl295:
- [ ] In case of failure, cldr-apps gets undeployed first. 

Todo: all GitHub checks Should happen first. So we don’t bail out with cldr-apps undeployed.

srl295:
Where you're attempting to go backward, the logs show this:

```
cldr-trunk was at  : 082e2cf33b
you want to move to:1d75dd4595
Deploy will include these new items:
---
---
Note, 1d75dd4595e30f34b45c240112087e0d2105d2f7 is not ahead of HEAD 082e2cf33b
Checking for items that would be REVERTED if we proceed:
---
082e2cf33b CLDR-14138 CLDRModify -fp / derived annotations (#691)
ef1f082e26 CLDR-13576 Add Ligurian core data (#496)
e9dc841f24 CLDR-14157 fix patterns (MMMM->MMM) for es_419 yMMM[d], es_MX/es_US yMMMEd (#697)
6865e7f361 CLDR-14160 Revert ms addition of japanese era names, ti addition of English metazone names (#695)
---
STOP. Check the override box if you really want to do this
```

srl295:
Works. Ready for review and merge.

macchiati (approved):

PR #700, opened 2020-09-03
CLDR-10338 add u locale extension -dx, disable dictionary break for listed scripts

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-10338

Add u locale extension key -dx to disable dictionary breaks for one or more script codes listed as arguments.
- Add BCP47 info
- Add key name in English
- Update LDML spec
- Update tests; this is a bit incomplete for now.

### DIFF
--- a/docs/ldml/tr35.html
+++ b/docs/ldml/tr35.html
@@ -102,7 +102,7 @@
       </tr>
       <tr>
         <td>Date</td>
-        <td class="changed">2020-05-13</td>
+        <td class="changed">2020-09-03</td>
       </tr>
       <tr>
         <!-- This link must be made live when posting the final version but is disabled during proposed update stage. -->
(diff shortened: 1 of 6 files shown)


docs/ldml/tr35.html
```diff
@@ -2369,6 +2369,29 @@ <h4><a href="#Key_And_Type_Definitions_" name=
...
+        "https://github.com/unicode-org/cldr/tree/latest/common/bcp47/segmentation.xml">segmentation.xml</a>.</strong></td>
+      </tr>
+      <tr class="changed">
+        <td>"dx"</td>
+        <td>Dictionary break script exclusions</td>
+        <td>
+          <i>ISO 15924 script code</i>
+        </td>
+        <td>
+          <p>Items of type SCRIPT_CODE, which are codes consisting of 4 ASCII letters
```
macchiati, line 2390: Change this to reference a [valid Unicode script subtag](http://unicode.org/reports/tr35/#unicode_script_subtag)

(maybe also other places if you have time)

macchiati (changes requested):
One small item.

Also, should update release page, at least with a TBD

docs/ldml/tr35.html
pedberg-icu: Thanks, I did make that change

docs/ldml/tr35.html
```diff
@@ -2369,6 +2369,29 @@ <h4><a href="#Key_And_Type_Definitions_" name=
           <em>Unknown or Invalid Currency</em>.</p>
         </td>
       </tr>
+      <tr class="changed">
+        <td colspan="4"><strong>A <a href=
+        "#UnicodeDictionaryBreakExclusionIdentifier" name=
+        "UnicodeDictionaryBreakExclusionIdentifier" id=
```
markusicu, line 2375: The anchor names (id=...) in the spec use a variety of styles, but underscores seem common. Do we care?

docs/ldml/tr35.html
```diff
@@ -2369,6 +2369,29 @@ <h4><a href="#Key_And_Type_Definitions_" name=
...
+      <tr class="changed">
+        <td>"dx"</td>
+        <td>Dictionary break script exclusions</td>
+        <td>
+          <i><code><a href="#unicode_script_subtag">unicode_script_subtag</a></code> values</i>
+        </td>
+        <td>
+          <p>One or more items of type SCRIPT_CODE, which are valid
+          <code><a href="#unicode_script_subtag">unicode_script_subtag</a></code> values.</p>
+          <p>The code Zyyy (Common) can be specified to exclude all scripts.</p>
```
markusicu, line 2392: Add that in that case it needs to be the only code?

docs/ldml/tr35.html
pedberg-icu: At least in this table, id values specified as a category are italicized, while explicit values are quoted and non-italic. So this follows that style.

docs/ldml/tr35.html
pedberg-icu: Good point, I will update to indicate this.

pedberg-icu: Done...

docs/ldml/tr35.html
markusicu: I am talking about the anchor text id="UnicodeDictionaryBreakExclusionIdentifier", not the CSS style of the presentation.
Other anchors use underscores (id="Unicode_Dictionary_Break_Exclusion_Identifier"), lowercase, etc.

pedberg-icu: The camel-case style seems to be used for all of the other id anchors in this table, so I just followed that... e.g.
```
#UnicodeCurrencyFormatIdentifier
#UnicodeCollationIdentifier
#UnicodeCurrencyIdentifier
#UnicodeEmojiPresentationStyleIdentifier
#UnicodeFirstDayIdentifier
...
```

pedberg-icu:
Does this seem OK now? I would like to merge for the CLDR-ICU integration if possible

pedberg-icu:
> Also, should update release page, at least with a TBD

I did add add least  skeleton entry for this in https://sites.google.com/site/cldr/index/downloads/cldr-38

markusicu (approved):
lgtm tnx

PR #701, opened 2020-09-04
CLDR-12010 fr plural rules: many should not include 0, fix samples for other

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-12010

Adjustment to new French plural rules from ICU integration testing:
- Rules for "many" should exclude value 0
- Samples for "other" should not include 1000000 (that is for "many")

### DIFF
--- a/common/supplemental/plurals.xml
+++ b/common/supplemental/plurals.xml
@@ -28,8 +28,8 @@ For terms of use, see http://www.unicode.org/copyright.html
         </pluralRules>
         <pluralRules locales="fr">
             <pluralRule count="one">i = 0,1 @integer 0, 1 @decimal 0.0~1.5</pluralRule>
-			<pluralRule count="many">e = 0 and i % 1000000 = 0 and v = 0 or e != 0 .. 5 @integer 1000000</pluralRule>
-            <pluralRule count="other"> @integer 2~17, 100, 1000, 10000, 100000, 1000000, … @decimal 2.0~3.5, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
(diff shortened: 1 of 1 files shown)


sffc (approved):

common/supplemental/plurals.xml
```diff
@@ -28,8 +28,8 @@ For terms of use, see http://www.unicode.org/copyright.html
         </pluralRules>
         <pluralRules locales="fr">
             <pluralRule count="one">i = 0,1 @integer 0, 1 @decimal 0.0~1.5</pluralRule>
-			<pluralRule count="many">e = 0 and i % 1000000 = 0 and v = 0 or e != 0 .. 5 @integer 1000000</pluralRule>
-            <pluralRule count="other"> @integer 2~17, 100, 1000, 10000, 100000, 1000000, … @decimal 2.0~3.5, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
+            <pluralRule count="many">e = 0 and i != 0 and i % 1000000 = 0 and v = 0 or e != 0 .. 5 @integer 1000000</pluralRule>
```
sffc, line 31: I'm not a plural rules expert, but doesn't the "or e != 0 .. 5" mean that this rule matches all numbers with e >= 6?  Like all of 1M, 1.1M, 1.2M?

macchiati: Yes. I suspect we'll need to tweak this further.

macchiati (approved):

pedberg-icu:
Though it may need further tweaks, this clearly seems like an improvement over the previous version, so I will go ahead and merge.

PR #702, opened 2020-09-08
CLDR-14128 Update ICU4J libs to Sep 5; LocaleMatcher update => test fails, skipped

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14128
- [x] Updated PR title and link in previous line to include Issue number

Update the ICU4J libs in CLDR to match the current ICU trunk as of 2020-Sep-05. This includes an updated LocaleMatcher that is incompatible with usage in some of the CLDR unit tests (which are based on CLDR tools/java/com/ibm/icu/util/LocaleMatcher.java).
I have filed https://unicode-org.atlassian.net/browse/CLDR-14166 about that and am currently skipping the failing tests with a logKnownIssue on that ticket.

PR #704, opened 2020-09-08
CLDR-11585 mavenize (2020 parallel version)

srl295:
CLDR-11585

So, it turns out this can be done a little bit in parallel after all.

- add pom.xml and use the _current_ directory structure. Once this is working, then switch over.
- no tests yet (this needs code refactoring)
- cldr-apps builds and runs OK
- it's possible some of the libs are slightly out of date, but that's easily fixable.
- using a temporary repo for ICU4J/utilities until https://github.com/unicode-org/icu/pull/1270 gets merged.

### DIFF
--- a/.github/workflows/mvn-settings.xml
+++ b/.github/workflows/mvn-settings.xml
@@ -1,0 +1,53 @@
+<!--
+Copyright © 1991 and later Unicode, Inc.
+All rights reserved.
+License and terms of use: http://www.unicode.org/copyright.html
+
+This file is used during builds to be able to access a temporary copy
+of icu4j.jar and utilities.jar. See CLDR-11585.
+TODO: Remove this when ICU-21251 is completed.
+-->
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                      http://maven.apache.org/xsd/settings-1.0.0.xsd">
+  <activeProfiles>
+    <activeProfile>github</activeProfile>
+  </activeProfiles>
+  <profiles>
+    <profile>
+      <id>github</id>
+      <repositories>
+        <repository>
+          <id>central</id>
+          <url>https://repo1.maven.org/maven2</url>
+          <releases><enabled>true</enabled></releases>
+          <snapshots><enabled>true</enabled></snapshots>
+        </repository>
+        <repository>
+          <id>githubsrl295</id>
+          <name>GitHub srl295/icu Apache Maven Packages</name>
+          <url>https://maven.pkg.github.com/srl295/icu</url>
+        </repository>
+        <repository>
+          <id>github</id>
+          <name>GitHub unicode-org/cldr Apache Maven Packages</name>
+          <url>https://maven.pkg.github.com/unicode-org/cldr</url>
+        </repository>
+      </repositories>
+    </profile>
+  </profiles>
+
+  <servers>
+    <server>
+      <id>github</id>
+      <username>${GITHUB_ACTOR}</username>
+      <password>${GITHUB_TOKEN}</password>
+    </server>
+    <server>
+      <id>githubsrl295</id>
+      <username>${GITHUB_ACTOR}</username>
+      <password>${GITHUB_TOKEN}</password>
+    </server>
+  </servers>
+</settings>
--- a/tools/cldr-apps/pom.xml
+++ b/tools/cldr-apps/pom.xml
@@ -1,0 +1,243 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>cldr-apps</artifactId>
+	<packaging>war</packaging>
+	<name>CLDR Survey Tool</name>
+
+	<url>https://unicode.org/cldr</url>
+
+	<scm>
+		<connection>scm:git:https://github.com/unicode-org/cldr.git</connection>
+	</scm>
+
+	<parent>
+		<groupId>org.unicode.cldr</groupId>
+		<artifactId>cldr-all</artifactId>
+		<version>38.0-SNAPSHOT</version>
+	</parent>
+
+	<dependencies>
+		<!-- test -->
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-api</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<!-- icu -->
+		<dependency>
+			<groupId>com.ibm.icu</groupId>
+			<artifactId>icu4j</artifactId>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>com.ibm.icu</groupId>
+			<artifactId>utilities</artifactId>
+			<scope>provided</scope>
+		</dependency>
+		<!-- cldr -->
+		<dependency>
+			<groupId>org.unicode.cldr</groupId>
+			<artifactId>cldr</artifactId>
+		</dependency>
+
+		<!-- API -->
+		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>javax.servlet-api</artifactId>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>javax.servlet.jsp</groupId>
+			<artifactId>javax.servlet.jsp-api</artifactId>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>javax.xml.bind</groupId>
+			<artifactId>jaxb-api</artifactId>
+			<scope>provided</scope>
+		</dependency>
+
+		<!-- mail / rss -->
+		<dependency>
+			<groupId>rome</groupId>
+			<artifactId>rome</artifactId>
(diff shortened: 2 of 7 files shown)


srl295:
blocked on https://github.com/unicode-org/icu/pull/1270

srl295:
So, this isn't great, and does not include icu4j libs from #702 yet.

But it's a step forward. Can we get this merged, with a parallel, non-blocking maven build?

tools/cldr-apps/pom.xml
```diff
@@ -0,0 +1,309 @@
...
+      <!-- <scope>test</scope> -->
+    </dependency>
+
+    <dependency>
+    <groupId>mysql</groupId>
+    <artifactId>mysql-connector-java</artifactId>
+    <version>8.0.21</version>
+    </dependency>
+
+<!-- https://mvnrepository.com/artifact/com.google.guava/guava -->
```
hagbard, line 173: Obviously build file indentation etc.
And it would be nice if any commented out dependencies were either comment as to why, to removed.

tools/cldr-apps/pom.xml
```diff
@@ -0,0 +1,309 @@
...
+    <version>2.12.0</version>
+</dependency>
+
+    
+    
+  </dependencies>
+
+  <build>
+    <finalName>cldr-apps</finalName>
+    <sourceDirectory>src</sourceDirectory> <!-- TODO: HACK! until we can pivot src dirs -->
```
hagbard, line 206: Can you reference a CLDR ticket number in the TODO ?

tools/cldr-apps/pom.xml
```diff
@@ -0,0 +1,309 @@
...
+            </systemPropertyVariables>
+          </configuration>
+        </plugin>
+        <plugin>
+          <artifactId>maven-war-plugin</artifactId>
+          <version>3.2.2</version>
+        <configuration>
+          <webResources>
+            <resource>
+              <!-- this is relative to the pom.xml directory -->
```
hagbard, line 246: "the pom.xml directory" --> "the directory this POM file is in" ?

tools/java/pom.xml
```diff
@@ -0,0 +1,219 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <sourceDirectory>.</sourceDirectory> <!-- TODO: HACK! until we can pivot src dirs -->
+   <resources>
+     <resource>
+       <directory>.</directory>
+        <includes>
+          <include>org/unicode/cldr/util/data/**/*</include>
```
hagbard, line 92: Can this be ordered alphabetically ? It nearly is already.

tools/java/pom.xml
```diff
@@ -0,0 +1,219 @@
...
+        <!-- default lifecycle, jar packaging: see https://maven.apache.org/ref/current/maven-core/default-bindings.html#Plugin_bindings_for_jar_packaging -->
+        <plugin>
+          <artifactId>maven-resources-plugin</artifactId>
+          <version>3.0.2</version>
+        </plugin>
+        <plugin>
+          <artifactId>maven-compiler-plugin</artifactId>
+          <version>3.8.0</version>
+          <configuration>
+            <!-- TODO: pivot this when we can restructure src dirs -->
```
hagbard, line 127: It sounds like there's a vaguely structured plan behind some of these TODOs. Having a CLDR ticket for the discussion would be great.

tools/java/pom.xml
```diff
@@ -0,0 +1,219 @@
...
+        <plugin>
+          <artifactId>maven-jar-plugin</artifactId>
+          <version>3.0.2</version>
+          <configuration>
+            <archive>
+              <manifest>
+                <mainClass>org.unicode.cldr.tool.Main</mainClass>
+              </manifest>
+              <manifestEntries>
+                              <Built-By>${user.name}</Built-By>
```
hagbard, line 156: I assume you can auto-indent everything easily enough before merging.

.github/workflows/mvn-settings.xml
```diff
@@ -0,0 +1,48 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
```
hagbard, line 1: Copyright?
Meta-comment on the purpose of these files?

.github/workflows/mvn-settings.xml
```diff
@@ -0,0 +1,48 @@
...
+    </server>
+    <!-- TODO: ARGH. What I'd really like to do is to just append to settings.xml 
+                but that's some extra code. -->
+    <server>
+      <id>githubsrl295</id>
+      <username>${GITHUB_ACTOR}</username>
+      <password>${GITHUB_TOKEN}</password>
+    </server>
+  </servers>
+</settings>
```
hagbard, line 48: Newline at eof?

hagbard (commented):
Just a few quick comments.

tools/pom.xml
```diff
@@ -0,0 +1,17 @@
...
+  <groupId>org.unicode.cldr</groupId>
+  <artifactId>cldr-all</artifactId>
+  <version>38.0-SNAPSHOT</version>
+  <name>CLDR Top Directory</name>
+  <packaging>pom</packaging>
+  
+  <modules>
+    <module>java</module>
+    <module>cldr-apps</module>
+  </modules>
```
echeran, line 16: Since this is the parent (root?) pom.xml, this is where you can "refactor" (de-duplicate) specifying the version numbers of commonly-used dependencies and plugins via [`dependencyManagement`](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management).

Similarly, using `<dependencyManagement>` in `tools/cldr-apps/pom.xml` probably isn't doing much for you, as it is a child/leaf pom.xml.

Also, if you refactor `./java/pom.xml` and `./cldr-apps/pom.xml` to de-duplicate using the parent-child inheritance, then you can run one of the "child" pom.xmls (ex: `./java/pom.xml`) with the proper inerhitance/interpolation in effect by running this command in the directory of the root pom.xml:  `mvn -am -pl java/pom.xml`.

I think the only catch with Maven is that the version number of artifact that the pom.xml defines can't be inherited, and it must be hard-coded.

I don't know if this is useful, but I know that there are [Maven plugins](http://maven.apache.org/maven-release/maven-release-plugin/index.html) out there to manage the incrementing of versions from N to (N+1)-SNAPSHOT to N+1, etc.  They can can manage the version value in the submodule pom.xml's as well.

tools/cldr-apps/pom.xml
srl295: I'll run everything through Eclipse's cleanup and fix the dependency docs, thanks.

tools/cldr-apps/pom.xml
srl295: It's this same one, CLDR-11585. To solve this I need to rearrange the source files which breaks everyone. That's why I want To do this in stages.

tools/java/pom.xml
srl295: I will update CLDR-11585

tools/pom.xml
srl295: > Similarly, using <dependencyManagement> in tools/cldr-apps/pom.xml probably isn't doing much for you, as it is a child/leaf pom.xml.

This turned out to help the build be able to find it as a sibling dependency.

Will do the rest, I quickly scrambled my workspace trying to do this refactoring. 

I'd really like to rename 'java' to something, perhaps 'cldr4j' or 'cldr-tools'. Otherwise it's misnamed.

srl295: corrected… push coming

tools/cldr-apps/pom.xml
srl295: I ended up removing/replacing this comment. It's more important to note that the point of this is to temporarily avoid renaming all of the source files.

tools/cldr-apps/pom.xml
```diff
@@ -0,0 +1,243 @@
...
+
+		<dependency>
+			<groupId>xml-apis</groupId>
+			<artifactId>xml-apis</artifactId>
+		</dependency>
+	</dependencies>
+	<build>
+		<finalName>cldr-apps</finalName>
+		<sourceDirectory>src</sourceDirectory> <!-- TODO: HACK! until we can pivot src dirs -->
+		<pluginManagement><!-- lock down plugins versions to avoid using Maven
```
echeran, line 151: Now that you have `<dependencyManagement>` and `<pluginManagement>` in an ancestor pom.xml with duplicated version info specified there, is it now possible here to remove the tags `<pluginManagement>`, `</pluginManagement>`, and elements of the form `<version>...</version>` (leaving every other tag nested inside them as-is)?  This should not change the behavior in any way, of course.

tools/cldr-apps/pom.xml
```diff
@@ -0,0 +1,243 @@
...
+				</plugin>
+				<plugin>
+					<artifactId>maven-deploy-plugin</artifactId>
+					<version>2.8.2</version>
+				</plugin>
+				<plugin>
+					<groupId>org.codehaus.mojo</groupId>
+					<artifactId>buildnumber-maven-plugin</artifactId>
+					<version>1.4</version>
+					<executions>
```
echeran, line 220: I didn't remember / realize this, but pluginManagement and dependencyManagement propagate the configuration and other information for their respective plugin and dependency declarations, not just the version numbers.  These configuration details are duplicated in the root pom.xml, which `tools/java/pom.xml` is inheriting.  You could remove these details here, similarly, to clean up and allow inheritance.  Of course, verify that what I said is correct and the behavior is unchanged.

Or, alternatively, if you prefer, you could not use the inheritance mechanism for the configuration details, leaving it to be explicitly configured only in leaf pom.xml files and not in the root pom.xml.

I don't have a strong opinion on that, either way.

tools/cldr-apps/pom.xml
srl295: Yes, I will make sure the *Management tags are rolled up into the parent.

tools/cldr-apps/pom.xml
srl295: @echeran
- I've moved *all* `*Management` into the parent pom
- the only common configuration that makes sense to make common (and I have) is the githash stuff, everything else is per-project.

echeran (approved):
LGTM

srl295:
@macchiati @yumaoka any thoughts on this?

macchiati:
I'm not particularly up on maven, so I would go by what Elango says. I'm glad to rubberstamp unless Yoshito has time to review.

macchiati (approved):

PR #705, opened 2020-09-09
CLDR-8383 CI against mysql, disrecommend derby

srl295:
CLDR-8383 

- allow web tests to take a JDBC URL, org.unicode.cldr.unittest.web.jdbcurl
- start mysql in github actions, set it up before test run


srl295:
Travis will still test against derby…for now.

srl295:
when derby support is removed, travis will no longer build, which will be fine per CLDR-11585 which was to obviate it anyway.

srl295:
I've also updated the surveytool-is-broken messages to recommend mysql, and to have the Mysql Configurator button right there.

srl295:
also, docs at .http://cldr.unicode.org/development/running-survey-tool/setup have been updated.

I think this is all done for the most part:  will be a followup PR when we actually remove Derby.

macchiati (approved):

PR #706, opened 2020-09-09
CLDR-14167 Update subdivisions code

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14167


macchiati:
As it turns out, there were no updates to subdivisions that affected CLDR. However, I had to make some code fixes to get that far. The problem was that the tool needs to look at past versions, and there was a bug in how the UnitConverter was constructed that showed up when reading supplemental data. The bug was that the UnitConverter for an old version was using the Validity data for the new version. So I had to restructure a bit to deal with that.

srl295 (approved):
needs squash otherwise LGTM

PR #707, opened 2020-09-10
CLDR-13771 ansible: fix proxy so users don't all appear as 127.0.0.1

srl295:
CLDR-13771 ansible: fix proxy so users don't all appear as 127.0.0.1

A proxy header was set incorrectly, but it's been resolved now.  Yoshito might recognize this issue

thanks for reporting this @btangmu


srl295:
I've already deployed this.

btangmu (approved):

PR #709, opened 2020-09-15
CLDR-8654 parallelize VettingViewer - 25-50% speed improvement

srl295:
CLDR-8654

- use a ForkJoinPool to parallelize the VV generation
- write to separate Appendable buffers and then join them all at once
- add CLDRConfig.get(String, int default) for fetching numeric config parameters
- DefaultErrorStatus errorStatus was moved into getFileInfo() as it did not
have any long running state, and produced concurrency issues.

parameters:

CLDR_VETTINGVIEWER_CHUNKSIZE=1
- number of locales each thread should process before getting
more work

CLDR_VETTINGVIEWER_PARALLEL=0
- number of threads (0 means 'one per CPU')

### DIFF
--- a/tools/java/org/unicode/cldr/util/VettingViewer.java
+++ b/tools/java/org/unicode/cldr/util/VettingViewer.java
@@ -4,7 +4,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -19,6 +18,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.RecursiveAction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -71,6 +72,8 @@ public class VettingViewer<T> {
     private static final boolean TESTING = CldrUtility.getProperty("TEST", false);
 
     public static final Pattern ALT_PROPOSED = PatternCache.get("\\[@alt=\"[^\"]*proposed");
+
+    private static final boolean DEBUG_THREADS = false;
 
     public static Set<CheckCLDR.CheckStatus.Subtype> OK_IF_VOTED = EnumSet.of(Subtype.sameAsEnglish);
 
@@ -415,7 +418,6 @@ public class VettingViewer<T> {
     private final SupplementalDataInfo supplementalDataInfo;
     private final String baselineTitle = "Baseline";
     private final String currentWinningTitle;
-    private ErrorChecker errorChecker;
 
     private final Set<String> defaultContentLocales;
 
@@ -442,7 +444,6 @@ public class VettingViewer<T> {
 
         this.currentWinningTitle = currentWinningTitle;
         reasonsToPaths = Relation.of(new HashMap<String, Set<String>>(), HashSet.class);
-        errorChecker = new DefaultErrorStatus(cldrFactory);
     }
 
     public class WritingInfo implements Comparable<WritingInfo> {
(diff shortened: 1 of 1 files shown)


tools/java/org/unicode/cldr/util/CLDRConfig.java
```diff
@@ -794,4 +794,23 @@ public static VersionInfo getVersion(int... versionInput) {
...
     }
+
+    /**
+     * Get a config value as an integer.
+     * @param key key name
+     * @param defaultValue default value
+     * @return the value, or defaultValue if missing or unparseable.
+     * @see CLDRConfig#get(Object)
+     */
+    public int get(String key, int defaultValue) {
```
macchiati, line 805: The name of this is very obscure. It appears that this calls get(key) which calls getProperty() — which does have a meaningful name. The get() methods should be renamed to getProperty()

tools/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -805,6 +807,161 @@ public void generateSummaryHtmlErrorTables(Appendable output, EnumSet<Choice> ch
...
+                  output.append(this.header);
+                  lastChar = nextChar;
+              }
+            }
+        }
+
+        public int size() {
+            return names.size();
+        }
+
```
macchiati, line 883: It seems like it would be simpler and more understandable to use parallel streams. Or is there a problem with that?

tools/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -805,6 +807,161 @@ public void generateSummaryHtmlErrorTables(Appendable output, EnumSet<Choice> ch
...
+            this.configParallel = Math.max(config.get("CLDR_VETTINGVIEWER_PARALLEL", 0), 0);
+            if(this.configParallel < 1) {
+                this.configParallel = java.lang.Runtime.getRuntime().availableProcessors(); // matches ForkJoinPool() behavior
+            }
+            this.configChunkSize = Math.max(config.get("CLDR_VETTINGVIEWER_CHUNKSIZE", 1), 1);
+            System.err.println("vv: CLDR_VETTINGVIEWER_PARALLEL="+configParallel+", CLDR_VETTINGVIEWER_CHUNKSIZE="+configChunkSize);
+        }
+
+        public void append(Appendable output) throws IOException {
```
macchiati, line 863: This name is misleading. It is appending _this_ to output, rather than the reverse, so should be _appendto_.

tools/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -805,6 +807,161 @@ public void generateSummaryHtmlErrorTables(Appendable output, EnumSet<Choice> ch
 
     }
 
+    private class WriteContext {
+
+        private List<String> names = new ArrayList<>();
```
macchiati, line 812: I'm guessing these are _localeNames_ and _localeIds_

I know our code has lots of bad names, but for new code we should be clearer.

macchiati (changes requested):

tools/java/org/unicode/cldr/util/VettingViewer.java
srl295: They are parallel streams? And then they are appended together at the end, including header-inclusion logic.

tools/java/org/unicode/cldr/util/CLDRConfig.java
srl295: So this is actually inherited from the fact that it extends `Properties` which extends `Hashtable`. 
But I will change this to `getProperty() which is clearer.

srl295: It's worse than that, I actually already added a `getProperty(String, Integer)` some (years) ago.  Deleted the duplicate implementation and changing code to use that...

tools/java/org/unicode/cldr/util/VettingViewer.java
srl295: streams outputs[0…n] are written in parallel by separate threads, using the ForkJoin mechanism's algorithms for efficient scheduling.  Then, all of the streams are written one after another to the requested output.

tools/java/org/unicode/cldr/util/CLDRConfig.java
srl295: Done

macchiati (approved):

tools/java/org/unicode/cldr/util/VettingViewer.java
macchiati: Here is a tutorial on them. https://docs.oracle.com/javase/tutorial/collections/streams/parallelism.html

There are a few instances in the unitTests, but we haven't changed older code.

srl295: OK, that's a different use of 'parallel stream' but, I think it's worth looking into.

PR #712, opened 2020-09-18
CLDR-14176 change plural rule regeneration to have exponent samples

macchiati:
A test failure in ICU exposed that the samples were being generated incorrectly for the new French 'many' plural rule. This should fix that problem, although ICU may now need some changes to recognize syntax like "1.2e6", and as being different than 1200000.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14176

### DIFF
--- a/common/supplemental/ordinals.xml
+++ b/common/supplemental/ordinals.xml
@@ -61,7 +61,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <pluralRule count="other"> @integer 0~7, 9, 10, 12~17, 100, 1000, 10000, 100000, 1000000, …</pluralRule>
         </pluralRules>
         <pluralRules locales="lij">
-            <pluralRule count="many">n = 11,8,80..89,800..899 @integer 8, 11, 80~89, 800~899</pluralRule>
+            <pluralRule count="many">n = 11,8,80..89,800..899 @integer 8, 11, 80~89, 800~803</pluralRule>
             <pluralRule count="other"> @integer 0~7, 9, 10, 12~17, 100, 1000, 10000, 100000, 1000000, …</pluralRule>
         </pluralRules>
 
--- a/common/supplemental/plurals.xml
+++ b/common/supplemental/plurals.xml
@@ -26,11 +26,6 @@ For terms of use, see http://www.unicode.org/copyright.html
             <pluralRule count="one">i = 0,1 @integer 0, 1 @decimal 0.0~1.5</pluralRule>
             <pluralRule count="other"> @integer 2~17, 100, 1000, 10000, 100000, 1000000, … @decimal 2.0~3.5, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
         </pluralRules>
-        <pluralRules locales="fr">
-            <pluralRule count="one">i = 0,1 @integer 0, 1 @decimal 0.0~1.5</pluralRule>
-            <pluralRule count="many">e = 0 and i != 0 and i % 1000000 = 0 and v = 0 or e != 0 .. 5 @integer 1000000</pluralRule>
-            <pluralRule count="other"> @integer 2~17, 100, 1000, 10000, 100000, … @decimal 2.0~3.5, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
-        </pluralRules>
         <pluralRules locales="pt">
             <pluralRule count="one">i = 0..1 @integer 0, 1 @decimal 0.0~1.5</pluralRule>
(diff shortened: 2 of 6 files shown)


common/supplemental/ordinals.xml
```diff
@@ -61,7 +61,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <pluralRule count="other"> @integer 0~7, 9, 10, 12~17, 100, 1000, 10000, 100000, 1000000, …</pluralRule>
         </pluralRules>
         <pluralRules locales="lij">
```
macchiati, line 63: The sample generation changed, this is just a minor byproduct

common/supplemental/plurals.xml
```diff
@@ -116,6 +111,14 @@ For terms of use, see http://www.unicode.org/copyright.html
             <pluralRule count="other"> @integer 0, 5~19, 100, 1000, 10000, 100000, 1000000, … @decimal 0.0, 0.5~1.0, 1.5~2.0, 2.5~2.7, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
         </pluralRules>
 
+        <!-- 3: one,many,other -->
```
macchiati, line 114: the rules changed position (more forms), and adds exponent samples

common/supplemental/plurals.xml
```diff
@@ -191,7 +194,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <pluralRule count="one">n % 10 = 1 and n % 100 != 11,71,91 @integer 1, 21, 31, 41, 51, 61, 81, 101, 1001, … @decimal 1.0, 21.0, 31.0, 41.0, 51.0, 61.0, 81.0, 101.0, 1001.0, …</pluralRule>
             <pluralRule count="two">n % 10 = 2 and n % 100 != 12,72,92 @integer 2, 22, 32, 42, 52, 62, 82, 102, 1002, … @decimal 2.0, 22.0, 32.0, 42.0, 52.0, 62.0, 82.0, 102.0, 1002.0, …</pluralRule>
             <pluralRule count="few">n % 10 = 3..4,9 and n % 100 != 10..19,70..79,90..99 @integer 3, 4, 9, 23, 24, 29, 33, 34, 39, 43, 44, 49, 103, 1003, … @decimal 3.0, 4.0, 9.0, 23.0, 24.0, 29.0, 33.0, 34.0, 103.0, 1003.0, …</pluralRule>
```
macchiati, line 196: another harmless byproduct

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestPluralRuleGeneration.java
```diff
@@ -89,4 +93,52 @@ public void TestAtoms() {
         }
 
     }
```
macchiati, line 95: Introduced a new EFixedDecimal for generating exponent samples; here are tests.

tools/java/org/unicode/cldr/tool/EFixedDecimal.java
```diff
@@ -0,0 +1,82 @@
+package org.unicode.cldr.tool;
+
+import com.ibm.icu.text.PluralRules.FixedDecimal;
+import com.ibm.icu.text.PluralRules.IFixedDecimal;
+import com.ibm.icu.text.PluralRules.Operand;
+import com.ibm.icu.text.UnicodeSet;
+
```
macchiati, line 7: Introduced a new EFixedDecimal for generating exponent samples. Basically just adds the 'e' feature.

tools/java/org/unicode/cldr/util/TempPrintWriter.java
```diff
@@ -22,6 +22,7 @@
     final PrintWriter tempPrintWriter;
     final String tempName;
     final String filename;
+    boolean noReplace = false;
 
 
     public static TempPrintWriter openUTF8Writer(String filename) {
```
macchiati, line 28: Modified this so that the temp files can be removed if there is an exception. Avoids muddying the directories.

tools/java/org/unicode/cldr/tool/GeneratedPluralSamples.java
```diff
@@ -28,7 +28,6 @@
 import com.google.common.base.Joiner;
```
macchiati, line 28: Here are the main changes, moving to EFixedPoint. Had to break out the samples into different groups. I first tried to merge with the range notation, but that got to be really ugly.

pedberg-icu (approved):
Looks good. Do we have an ICU ticket for dealing with the new samples?

macchiati:
Yes, Elango has that ticket:
https://unicode-org.atlassian.net/browse/ICU-21270.

Elango, the samples can now have format like 1e27, so you may need to
change the parsing. But maybe not: I don't know if ICU looks at the samples
other than the ellipsis.

Mark

echeran:
@macchiati @pedberg-icu Okay, I can start taking a look at that on the ICU side now.  I haven't worked directly with plural samples or rule parser logic before, but I'll try to go as fast as I can.

PR #718, opened 2020-09-19
CLDR-13948 only deploy to smoketest in head fork

srl295:
CLDR-13948

- otherwise, any fork with a master branch will attempt (and fail) to deploy.

-----
Specifically locks the smoketest deployment to https://github.com/unicode-org/cldr.git
I don't think there's a reason to lock down "push to production" because it's manually triggered. 

- example where the repo doesn't match (i.e. fork): https://github.com/srl295/cldr/actions/runs/262743821
- example where the repo matches (simulating 'head fork'): https://github.com/srl295/cldr/actions/runs/262751278 (i've changed the workflow to check for `srl295/cldr`)

This brings up a question, though. The maven (and ant) builds currently run on:
 - branch 'master' (any fork)
 - pull request from any branch (any fork)

We may want to change this to actually `push: branches: '*'` so that builds happen on any branch, any fork, and only restrict the automatic deployment step to be branch master, and the head fork.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -13,7 +13,7 @@ name: cldr-ant
 on:
   push:
     branches:
-      - master
+      - '*'
   pull_request:
     branches: '*'
 
@@ -124,6 +124,7 @@ jobs:
     - name: CLDR Data Check
       run: ant -noinput datacheck -f tools/cldr-unittest/build.xml -DCLDR_DIR=$(pwd)
   deploy:
+    if: github.repository == 'unicode-org/cldr' && github.event_name == 'push' && github.ref == 'refs/heads/master'
     needs:
       - build
       - surveytest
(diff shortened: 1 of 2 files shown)


srl295:
@echeran I see that icu4x's workflow is similar. Would it be better to build all branches? This way if someone forks, they get tests happening (any downsides to this?) And, in the head fork, a random feature branch will get built even before there's a PR.

https://github.com/unicode-org/icu4x/blob/master/.github/workflows/build-test.yml#L3-L7

For example in the `keyboards` branch, it didn't get built by workflows until PRs were opened.

https://github.com/unicode-org/cldr/actions?query=branch%3Akeyboards

srl295:
OK, i've changed it so maven and ant build all branches. 

See https://github.com/srl295/cldr/actions/runs/262776401 for a simulated run matching `srl295/cldr` to show that auto deploy to master should still work.

srl295:
- fixed another issue, deploy didn't restrict to one branch!

macchiati (approved):

srl295:
Worked! 

https://github.com/unicode-org/cldr/runs/1139214768?check_suite_focus=true#step:3:18

PR #720, opened 2020-09-21
CLDR-11283 Make regular automated backups of Survey Tool database

btangmu:
-Use Ansible to automate not only the backups but the setup of the backups

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11283

### DIFF
--- a/tools/scripts/ansible/setup-playbook.yml
+++ b/tools/scripts/ansible/setup-playbook.yml
@@ -196,3 +196,5 @@
         --keep --redirect --uir --hsts --staple-ocsp --must-staple
       args:
         creates: /etc/letsencrypt/renewal/{{ inventory_hostname }}.conf
+
+- import_playbook: backup-db-playbook.yml
--- a/tools/scripts/ansible/templates/cldrbackup/backup_sh.j2
+++ b/tools/scripts/ansible/templates/cldrbackup/backup_sh.j2
@@ -1,0 +1,12 @@
+#!/bin/bash
+#
+# A cron job may call this with a line such as:
+#     37 7 * * * sh /home/cldrbackup/backup.sh >/dev/null 2>&1
+
+DATABASE_NAME=cldrdb
+FNAME=cldr-DUMP-`date +%Y-%m-%d`
+
+cd /home/cldrbackup
+mysqldump ${DATABASE_NAME} --add-drop-table --create-options --default-character-set=utf8mb4 > ${FNAME}.sql
+xz ${FNAME}.sql
+rsync -q ${FNAME}.sql.xz corp.unicode.org:/home/users/cldrbackup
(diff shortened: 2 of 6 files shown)


srl295 (approved):
this looks great!

tools/scripts/ansible/templates/cldrbackup/backup_sh.j2
```diff
@@ -0,0 +1,12 @@
...
+# A cron job may call this with a line such as:
+#     37 7 * * * sh /home/cldrbackup/backup.sh >/dev/null 2>&1
+
+DATABASE_NAME=cldrdb
+FNAME=cldr-DUMP-`date +%Y-%m-%d`
+
+cd /home/cldrbackup
+mysqldump ${DATABASE_NAME} --add-drop-table --create-options --default-character-set=utf8mb4 > ${FNAME}.sql
+xz ${FNAME}.sql
+rsync -q ${FNAME}.sql.xz corp.unicode.org:/home/users/cldrbackup
```
srl295, line 12: should `corp.unicode.org` be a variable?

tools/scripts/ansible/templates/cldrbackup/config.j2
```diff
@@ -0,0 +1,3 @@
+Host corp.unicode.org
+ User cldrbackup
```
srl295, line 2: same here, this `cldrbackup` (on corp. not on the local system) could be a variable

tools/scripts/ansible/templates/cldrbackup/backup_sh.j2
btangmu: Good idea!

tools/scripts/ansible/templates/cldrbackup/config.j2
btangmu: Yes. There's probably a lot of refactoring that should be done, for better encapsulation, etc.

* In Ansible, when should we use yml vs j2 vs  raw files?
* How should we organize setup-playbook.yml and other files to keep the code units short, simple, and loosely coupled?

tools/scripts/ansible/setup-playbook.yml
```diff
@@ -196,3 +196,5 @@
         --keep --redirect --uir --hsts --staple-ocsp --must-staple
       args:
         creates: /etc/letsencrypt/renewal/{{ inventory_hostname }}.conf
+
+- import_playbook: backup-db-playbook.yml
```
btangmu, line 200: Is this a good approach to modularizing and separing of concerns? I partly just wanted to avoid turning setup-playbook.yml into a huge unstructured file. Also it may be convenient to be able to run backup-db-playbook.yml separately. Maybe the same applies to other sections of setup-playbook.yml. What's left there could move to separate playbooks, leaving only "import" lines in the most general playbook?

btangmu:
See https://github.com/unicode-org/cldr/pull/725

PR #722, opened 2020-09-22
CLDR-8753 reduce duplicate Collation construction

srl295:
CLDR-8753

Did not measure this, but should cut down on Collator creation in a couple of hot spots.

### DIFF
--- a/tools/java/org/unicode/cldr/util/MapComparator.java
+++ b/tools/java/org/unicode/cldr/util/MapComparator.java
@@ -24,10 +24,19 @@ import com.ibm.icu.util.Freezable;
 import com.ibm.icu.util.ULocale;
 
 public class MapComparator<K> implements Comparator<K>, Freezable<MapComparator<K>> {
-    private RuleBasedCollator uca = (RuleBasedCollator) Collator.getInstance(ULocale.ROOT);
-    {
-        uca.setNumericCollation(true);
+    private static final class CollatorHelper {
+        public static final Collator UCA = getUCA();
+        /**
+         * This does not change, so we can create one and freeze it.
+         * @return
+         */
+        private static Collator getUCA() {
+            final RuleBasedCollator newUca = (RuleBasedCollator) Collator.getInstance(ULocale.ROOT);
+            newUca.setNumericCollation(true);
+            return newUca.freeze();
+        }
     }
+    // initialize this once
     private Map<K, Integer> ordering = new TreeMap<>(); // maps from name to rank
     private List<K> rankToName = new ArrayList<>();
     private boolean errorOnMissing = true;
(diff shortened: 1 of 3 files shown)


tools/java/org/unicode/cldr/util/MapComparator.java
```diff
@@ -24,9 +24,17 @@
 import com.ibm.icu.util.ULocale;
 
 public class MapComparator<K> implements Comparator<K>, Freezable<MapComparator<K>> {
```
srl295, line 26: this class is used in many other spots!

tools/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1005,7 +1004,7 @@ void computeOne(int n) {
     private void writeSummaryTable(Appendable output, String header, Level desiredLevel,
         EnumSet<Choice> choices, T organization) throws IOException {
 
-        Map<String, String> sortedNames = new TreeMap<>(Collator.getInstance());
+        Map<String, String> sortedNames = new TreeMap<>(CLDRConfig.getInstance().getCollator());
```
srl295, line 1007: this is an Emoji-aware English collator, rather than the default (!!) collator.

srl295:
need to do some more tests

tools/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -683,12 +683,15 @@ public T getOrgVoteRaw(Organization orgOfUser) {
...
     private boolean usingKeywordAnnotationVoting = false;
 
-    private final Comparator<T> ucaCollator = new Comparator<T>() {
-        Collator col = Collator.getInstance(ULocale.ENGLISH);
+    private static final Collator englishCollator = Collator.getInstance(ULocale.ENGLISH).freeze();
 
+    /**
+     * Used for comparing objects of type T
+     */
+    private final Comparator<T> objectCollator = new Comparator<T>() {
```
srl295, line 691: I did end up retaining this, just to call String.valueOf…

tools/java/org/unicode/cldr/util/MapComparator.java
```diff
@@ -24,9 +24,17 @@
...
-    {
-        uca.setNumericCollation(true);
+    // initialize this once
+    private static final Collator uca = getUCA();
+
+    /**
+     * This does not change, so we can create one and freeze it.
+     * @return
+     */
+    private static Collator getUCA() {
```
macchiati, line 34: If we are sure this will be called at least ones by all users of MapComparator, this is fine. Otherwise would be good to use the Singleton thread-safe pattern (see other ticket).

macchiati (changes requested):
I can change to approve depending on your answer.

tools/java/org/unicode/cldr/util/MapComparator.java
srl295: it's used exactly once, in initializing the `uca` static member. Singleton pattern would be fine here too.

srl295: I'll check the other ticket and update this tomorrow…good point. let's start to propagate that pattern.

srl295: CLDR-14180

srl295: updated.

macchiati (approved):

PR #725, opened 2020-09-22
CLDR-11283 Make regular automated backups of Survey Tool database

btangmu:
-Use Ansible to automate not only the backups but the setup of the backups

-Variables cldr_database_name, cldr_db_backup_host, cldr_db_backup_user, cldr_db_backup_destination

-Name backups like cldrdb-DUMP... instead of cldr-DUMP..., distinguish st2 from st during overlap

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11283

### DIFF
--- a/tools/scripts/ansible/setup-playbook.yml
+++ b/tools/scripts/ansible/setup-playbook.yml
@@ -196,3 +196,5 @@
         --keep --redirect --uir --hsts --staple-ocsp --must-staple
       args:
         creates: /etc/letsencrypt/renewal/{{ inventory_hostname }}.conf
+
+- import_playbook: backup-db-playbook.yml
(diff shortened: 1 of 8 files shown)


tools/scripts/ansible/setup-playbook.yml
```diff
@@ -196,3 +196,5 @@
         --keep --redirect --uir --hsts --staple-ocsp --must-staple
       args:
         creates: /etc/letsencrypt/renewal/{{ inventory_hostname }}.conf
+
+- import_playbook: backup-db-playbook.yml
```
srl295, line 200: Hm, I don't think this import is needed here. You can just run the backup-db-playbook by itself.

srl295 (approved):
changes LGTM

PR #729, opened 2020-09-23
CLDR-13771 workaround ansible-lint failing

srl295:
CLDR-13771
- see https://github.com/ansible-collections/community.kubernetes/issues/135#issuecomment-661189940
- workaround by not setting a version

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -10,13 +10,12 @@ jobs:
       with:
         lfs: false # not needed for this job, as we don’t currently do a Java build
     - name: Lint Ansible Playbook
-      uses: ansible/ansible-lint-action@master
+      uses: ansible/ansible-lint-action@6c8c14186662e43effbe39a678950896a46799ad
       with:
         targets: |
           tools/scripts/ansible/setup-playbook.yml
-        override-deps: |
-          ansible==2.9
-          ansible-lint==4.2.0
+          tools/scripts/ansible/upgrade-playbook.yml
+          tools/scripts/ansible/backup-db-playbook.yml
         args: ""
     - name: Lint Markdown
       # Warn, don't fail yet


srl295:
- also, lint all playbooks
- also, use a specific hash of the ansible action, since it's not otherwise versioned.

srl295:
^ seems to work.  This ansible lint job passed, where others have been failing for the past few hours.

FYI @btangmu @macchiati in case you see 'ansible lint' failures, this fixes it.

srl295:
failure was:

```
+ override_python_packages
+ [[ -n ansible==2.9
ansible-lint==4.2.0
 ]]
+ pip install ansible==2.9 ansible-lint==4.2.0
WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
Collecting ansible==2.9
  Downloading ansible-2.9.0.tar.gz (14.1 MB)
Collecting ansible-lint==4.2.0
  Downloading ansible_lint-4.2.0-py2.py3-none-any.whl (57 kB)
```

echeran (approved):
LGTM.  Seems straightforward enough, and this is a good change -- just using a snapshot of the action that effectively fixes in place the version of Ansible being used.  FWIW, it looks like they pushed [tags for the GH action](https://github.com/ansible/ansible-lint-action/tags).

srl295:
> LGTM. Seems straightforward enough, and this is a good change -- just using a snapshot of the action that effectively fixes in place the version of Ansible being used. FWIW, it looks like they pushed [tags for the GH action](https://github.com/ansible/ansible-lint-action/tags).

yes, but 83 commits since that tag https://github.com/ansible/ansible-lint-action/compare/v4.1.0...master

btangmu (approved):
Ideally there might be a comment here clarifying the magic number, linking to https://github.com/ansible/ansible-lint-action/tree/6c8c14186662e43effbe39a678950896a46799ad

Maybe add that later, if this isn't replaced soon

PR #733, opened 2020-09-23

btangmu:
-Change filename like `CldrStAjax._b7a33e9fe_.js` instead of query string

-Change filename only if using reverse proxy, detected by X-Real-IP in request.getHeader

-Use Ansible to set up nginx.conf rewrite rule

-Revise unit test TestGitHash for cldr-apps TestAll; TODO: currently it fails

##### Checklist


btangmu:
Currently TestGitHash fails

srl295 (approved):
LGTM on a read through. Not sure about the hash within the tests.

btangmu:
> LGTM on a read through. Not sure about the hash within the tests.

I'll make another PR without the regression test for now. @srl295 could we consult soon on how to implement the regression test? Evidently cldr-apps TestAll needs to do more to simulate a real run of Survey Tool...

btangmu:
See https://github.com/unicode-org/cldr/pull/735

PR #734, opened 2020-09-23
CLDR-11585 Maven: Add tests, simplify pom setup

srl295:
CLDR-11585

after this, then https://sites.google.com/site/cldr/development/maven instructions should work to use maven (from the command line, at least).

-----

Rescoping this one to also include adding (via a shim) all tests (but not consolecheck).

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -24,3 +24,8 @@ jobs:
         run: mvn -s .github/workflows/mvn-settings.xml -B package --file tools/pom.xml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Test with maven
+        run: mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
+        env:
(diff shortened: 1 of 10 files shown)


srl295:
the `githubicu` URL was defined at the wrong level.

srl295:
without this, users will have to add something like this to ~/.m2/settings.xml:

```xml
      <repositories>
        <repository>
          <id>githubicu</id>
          <name>GitHub unicode-org/icu Apache Maven Packages</name>
          <url>https://maven.pkg.github.com/unicode-org/icu</url>
        </repository>
      </repositories>
```

srl295:
I'll do a more complete test integration, this simply sublaunches the entire CLDR test suite as a single junit test.
It's also not configurable.

But, while 'ant' is still in the picture, i wanted to keep it simple.

srl295:
Oops. Need to exclude TestShim from ant!

srl295:
> Oops. Need to exclude TestShim from ant!

fixed

macchiati (approved):

PR #735, opened 2020-09-23

btangmu:
-Change filename like `CldrStAjax._b7a33e9fe_.js` instead of query string

-Change filename only if using reverse proxy, detected by X-Real-IP in request.getHeader

-Use Ansible to set up nginx.conf rewrite rule

-Comments in unit test TestGitHash about what it currently fails to test

##### Checklist


btangmu:
This is the same as https://github.com/unicode-org/cldr/pull/733 except without trying to implement the regression test yet.

srl295 (approved):

PR #738, opened 2020-09-24
CLDR-11585 cache maven dependencies

srl295:
the maven build is timing out sometimes… let's see if caching will help.

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -10,6 +10,7 @@ jobs:
       with:
         lfs: false # not needed for this job, as we don’t currently do a Java build
     - name: Lint Ansible Playbook
+      # the hash here is a specific version of that repo, since there is not a later commit.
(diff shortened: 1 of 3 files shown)


srl295:
results:
 - maven tests only run once, not once at build and once at test (oops)
 - cache+build = 26 seconds,  where the maven build took 43 seconds before tests were added (that was fewer source files being compiled, so the savings are probably even greater)
 - More significantly, those 20 seconds are network traffic, which has been timing out recently.

srl295:
Also:

- do not specify the maven central URL but let it default
- update some action versions

.github/workflows/ansible-lint.yml
```diff
@@ -10,6 +10,7 @@ jobs:
       with:
         lfs: false # not needed for this job, as we don’t currently do a Java build
     - name: Lint Ansible Playbook
+      # the hash here is a specific version of that repo, since there is not a later commit.
```
srl295, line 13: @btangmu from your comment on #729

macchiati (approved):

PR #740, opened 2020-09-25

btangmu:
-Implement a unit test for validating the hash value

-New function CldrUtility.getCldrBaseDirHash for both TestMisc and SurveyAjax to get the hash value

-Use getCldrBaseDirectory (not CLDR_DATA_HASH) for getCldrBaseDirHash

-Move getGitHashForDir to CldrUtility

-Delete some dead code concerning CLDR_WEB_TESTS

##### Checklist

### DIFF
--- a/tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
+++ b/tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
@@ -7,6 +7,7 @@ package org.unicode.cldr.unittest.web;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRConfigImpl;
 import org.unicode.cldr.util.CLDRLocale;
+import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.Factory;
 import org.unicode.cldr.web.STFactory;
 import org.unicode.cldr.web.WebContext;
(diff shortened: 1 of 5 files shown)


tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
```diff
@@ -82,27 +83,26 @@ public void TestIsCoverageOrganization() {
...
-         * Reference: https://unicode-org.atlassian.net/browse/CLDR-13582
-         */
-        // final String hash = CLDRConfig.getInstance().getProperty("CLDR_DATA_HASH");
-        // assertNotNull("getting CLDR_DATA_HASH", hash);
-        // if (hash != null && !hash.matches("[0-9a-f]+")) {
-        //    errln("❌ CLDR_DATA_HASH is not hex: " + hash);
-        // }
+        final String hash = CldrUtility.getCldrBaseDirHash();
+        assertNotNull("getCldrBaseDirHash", hash);
+        if (hash != null) {
```
srl295, line 88: minor: it's already been asserted as not-null on the prior line, so this is an unnecessary ifdef.

tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
```diff
@@ -82,27 +83,26 @@ public void TestIsCoverageOrganization() {
...
+            if (!hash.matches("[0-9a-f]+")) {
+                errln("❌ getCldrBaseDirHash is not hex: " + hash);
```
srl295, line 90: it could be 'unknown' if git is missing or if the code was compiled from a source tarball?maybe this should be a warning?

tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
```diff
@@ -82,27 +83,26 @@ public void TestIsCoverageOrganization() {
...
+                errln("❌ getCldrBaseDirHash is shorter than 8 chars: " + hash);
+            }
+        }
+        final String strings[] = {"CLDR-Apps", "CLDR-Tools"};
+        for (String s : strings) {
+            final String version = CLDRConfigImpl.getGitHashForSlug(s);
+            assertNotNull("getting " + s + " version", version);
+            /*
+             * TODO: fail if CLDRURLS.UNKNOWN_REVISION.equals(version).
+             * Currently they do equal CLDRURLS.UNKNOWN_REVISION, since SurveyMain.init isn't called for TestAll.
```
srl295, line 101: So, in a test environment, you definitely don't have `cldr-apps.war` and thus no manifest to read "CLDR-Apps" from. You also might not have `cldr*.jar` for its tools manifest, if the tests are run in eclipse let's say.  So this might be a warning, but a challenge to get from the test environment.

I don't know if you can use a warfile as if it were a jar on the classpath? If so, we could have a post build step, something like:   `java -DCLDR_APPS_VERSION_TEST -classpath …:…:cldr*.jar:cldr-apps.war TestAll` or something like that?

tools/cldr-apps/src/org/unicode/cldr/util/CLDRConfigImpl.java
```diff
@@ -80,13 +78,6 @@ protected Environment getDefaultEnvironment() {
     }
 
     CLDRConfigImpl() {
-        // TODO remove this after some time- just warn people about the old message
```
srl295, line 83: :+1: "some time" has passed, whatever it was :)

tools/cldr-apps/src/org/unicode/cldr/util/CLDRConfigImpl.java
```diff
@@ -180,7 +171,7 @@ private synchronized void init() {
         // The git version of the cldr.jar file embedded in cldr-apps.
         survprops.put("CLDR_UTILITIES_HASH", getGitHashForSlug("CLDR-Tools"));
         // The git version of the CLDR_DIR currently in use.
-        survprops.put("CLDR_DATA_HASH", getGitHashForDir(survprops.getProperty("CLDR_DIR", null)));
+        survprops.put("CLDR_DATA_HASH", CldrUtility.getGitHashForDir(survprops.getProperty("CLDR_DIR", null)));
```
srl295, line 174: There's no comnon init() for CLDRConfig, but if there were, this line could move there.

srl295 (approved):
I'm going to "LGTM" this one. It's testing a lot more. The one if branch is minor.

tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
btangmu: That's a good point. On the other hand, if anyone runs Survey Tool in a context where the hash is unknown, cache busting will be broken. Warnings aren't always noticed in a timely fashion. It seems more important to prevent a regression in our actual usage, than to accommodate a hypothetical scenario.

tools/cldr-apps/src/org/unicode/cldr/unittest/web/TestMisc.java
btangmu: I considered removing the getGitHashForSlug test entirely, or just the TODO. If you'd prefer either of those changes, that's OK with me.

That's all tangential to the cache-busting issue. I'm just concerned that a lot of code isn't under test at all, and the tight coupling with SurveyMain.init makes a lot of code hard to test. A unit test for getGitHashForSlug could be implemented using some other jar file, maybe a small one just for testing; or some other means of dependency injection. I don't know if you can use a warfile as if it were a jar on the classpath; it's very mysterious territory for me!

tools/cldr-apps/src/org/unicode/cldr/util/CLDRConfigImpl.java
btangmu: At first I tried to call getCldrBaseDirHash there but got "Error: setDir() was already called once." Somehow putting the cart before the horse, I guess -- getCldrBaseDirHash calls getCldrBaseDirectory...

macchiati:
I think there's a ticket to move all possible utility methods (in general
code) out of cldr apps, to where it can be tested

PR #741, opened 2020-09-25
CLDR-11585 improve maven tests 

srl295:
you can now run specific subtests with:

```
mvn test '-Dorg.unicode.cldr.unittest.web.testArgs=-f:TestMisc -n -q' '-Dorg.unicode.cldr.unittest.testArgs=-f:TestUntimedCounter -n -q' 
```

I was unable to get code coverage to work here, but it's just a config file change away.

### DIFF
--- a/tools/cldr-apps/pom.xml
+++ b/tools/cldr-apps/pom.xml
@@ -179,18 +179,6 @@
 					</configuration>
 				</plugin>
 				<plugin>
-					<artifactId>maven-surefire-plugin</artifactId>
-					<version>${maven-surefire-plugin-version}</version>
-					<configuration>
-						<systemPropertyVariables>
-							<CLDR_DIR>${project.basedir}/../../</CLDR_DIR>
-							<CLDR_ENVIRONMENT>UNITTEST</CLDR_ENVIRONMENT>
-							<java.awt.headless>true</java.awt.headless>
-							<!-- <rununittest.arg>-n -q</rununittest.arg> -->
-						</systemPropertyVariables>
-					</configuration>
-				</plugin>
-				<plugin>
 					<artifactId>maven-war-plugin</artifactId>
 					<version>3.2.2</version>
 					<configuration>
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestShim.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestShim.java
@@ -2,8 +2,11 @@ package org.unicode.cldr.unittest;
 
 import java.io.PrintWriter;
 
+import org.unicode.cldr.util.TestShimUtils;
+
 import org.junit.jupiter.api.Test;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 
 /**
  * a JUnit test that calls TestAll.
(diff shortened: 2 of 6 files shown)


tools/cldr-apps/pom.xml
```diff
@@ -178,18 +178,15 @@
 						</includes>
 					</configuration>
 				</plugin>
-				<plugin>
+				<!-- <plugin>
```
btangmu, line 181: What's the rationale for commenting this out rather than just deleting it? Unless there's a really good reason -- like a plan to restore the code in the near future -- I think deletion is better. I've been studying the book Building Maintainable Software by Joost Visser, which includes a rule "Leave No Code in Comments Behind"

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestShim.java
```diff
@@ -2,19 +2,31 @@
...
         // call this instead.
         int errCount = TestAll.runTests(args);
         assertEquals(errCount, 0, "Test had errors");
     }
+
+    @Test
+    public void TestTestShimUtilTest() {
+        // Note: checks the system property corresponding to java.lang,
+        // we expect this to be unset!
+        String args[] = TestShimUtils.getArgs(java.lang.String.class, "-a -b -c");
```
btangmu, line 28: I don't understand the comment at all: "... system property corresponding to java.lang ... unset!"

Could it be clarified?

tools/java/pom.xml
```diff
@@ -118,13 +118,9 @@
 			<plugin>
 				<artifactId>maven-surefire-plugin</artifactId>
 				<configuration>
-					<systemPropertyVariables>
-						<CLDR_DIR>${project.basedir}/../../</CLDR_DIR>
-						<CLDR_ENVIRONMENT>UNITTEST</CLDR_ENVIRONMENT>
-						<java.awt.headless>true</java.awt.headless>
+					<!-- <systemPropertyVariables>
```
btangmu, line 121: Again, "Leave No Code in Comments Behind"?

tools/pom.xml
```diff
@@ -225,7 +225,12 @@
 					<artifactId>maven-surefire-plugin</artifactId>
 					<version>${maven-surefire-plugin-version}</version>
 					<configuration>
-						<argLine>-DCLDR_ENVIRONMENT=UNITTEST -Djava.awt.headless=true  -DCLDR_DIR=../../ -Xmx6g -enableassertions</argLine>
+					<systemPropertyVariables>
```
btangmu, line 228: The indentation seems off. Do we have a linter for that? More generally I've been wishing we had an agreed-on set of tools for normalizing spacing, etc., for all kinds of source files

tools/cldr-apps/pom.xml
srl295: thanks for the recommend! I think with source control, the old version's just a click away. We may want some more parameters here someday.

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestShim.java
srl295: Yes, i'll clarify. To test the mechanism, it assumes `java.lang.testArgs` is unset.

tools/java/pom.xml
srl295: I expect to need this, but will remove it for now.

I thought iot better to comment it out,. than to leave it in but unused.

tools/pom.xml
srl295: yes, i've fixed the indentation. It should be the same as Eclipse's now.

Agreed…

btangmu (approved):

PR #742, opened 2020-09-26
CLDR-12010 plural fixes except for table.

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-12010


macchiati:
This started out as being just fixes to the spec for the new plural operand and French rules. However, it became clear that the code being used in CLDR didn't work correctly, with fixes required to the code and tests, and a revised version of the plural rules for French.

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestPluralRuleGeneration.java
```diff
@@ -102,43 +115,162 @@ public void TestEFixedDecimal() {
             throw new ICUException(e);
         }
         Object[][] checkItems = {
+            // compare each group of items against one another, where a group is bounded by {},
+            // source,  f,      v,  e,  plural (, skipString — if toString doesn't roundtrip)?
```
echeran, line 119: nit: Should this comment be `// source, n, ...` instead of `// source, f, ...` ?

echeran (dismissed):
LGTM.  I left an optional minor comment.

If there isn't already a ticket in ICU and CLDR to replace usages of `FixedDecimal` with  `FormattedNumber`, then we should create them.  I didn't feel great about the constructors I created for `FixedDecimal` because of the logical confusion they create regarding the meanings and values of the operands passed as arguments, but it's only used internally and in only a few places.  Another reason to use `FormattedNumber` instead is because it is public-facing (and also because it is used for plural keyword selection).

I think we need another person to approve as well who has CLDR repo permissions.

srl295 (approved):
i can RSLGTM but it needs to be squashed into 1 commit

macchiati:
Had to squash. Can you reapprove?

srl295 (approved):

PR #736, opened 2020-09-24
CLDR-14175 update JSON converter to handle compoundUnitPattern1

srl295:
CLDR-14175

this was broken for, say, Amharic also.  My first fix is this.
From:

 ```json
          "power2": {
            "compoundUnitPattern1": "{0}²",
            "compoundUnitPattern1": "{0}²",
            "compoundUnitPattern1": "{0}²"
          },
```

 ```json
          "power2": {
            "compoundUnitPattern1": "{0}²",
            "compoundUnitPattern1-count-one": "{0}²",
            "compoundUnitPattern1-count-other": "{0}²"
          },
```

Now, for German a similar issue ensued. My fix outputs this:

```json
          "power2": {
            "compoundUnitPattern1": "Quadrat{0}",
            "compoundUnitPattern1-count-one": "Quadrat{0}",
            "compoundUnitPattern1-count-one-case-accusative": "Quadrat{0}",
            "compoundUnitPattern1-count-one-case-dative": "Quadrat{0}",
            "compoundUnitPattern1-count-one-case-genitive": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-one": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-one-case-accusative": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-one-case-dative": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-one-case-genitive": "Quadrat{0}",
            "compoundUnitPattern1-gender-masculine-count-one": "Quadrat{0}",
            "compoundUnitPattern1-gender-masculine-count-one-case-accusative": "Quadrat{0}",
            "compoundUnitPattern1-gender-masculine-count-one-case-dative": "Quadrat{0}",
            "compoundUnitPattern1-gender-masculine-count-one-case-genitive": "Quadrat{0}",
            "compoundUnitPattern1-count-other": "Quadrat{0}",
            "compoundUnitPattern1-count-other-case-accusative": "Quadrat{0}",
            "compoundUnitPattern1-count-other-case-dative": "Quadrat{0}",
            "compoundUnitPattern1-count-other-case-genitive": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-other": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-other-case-accusative": "Quadrat{0}",
            "compoundUnitPattern1-gender-feminine-count-other-case-dative": "Quadrat{0}",
[...]

PR #744, opened 2020-09-28
CLDR-14202 Add simple version of tool for checking correct conversions from CLDR.

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14202


srl295:
@macchiati is this for ICU data? because #736 was for JSON. should be clear which is which

macchiati:
I think we can use the same program, with some minor modifications (see the
note in the ticket) to handle both.

Mark

PR #745, opened 2020-09-29
CLDR-11585 maven: single jar assembly, mysql, cleanup

srl295:
CLDR-11585

- create an uber-jar for cldr.jar with:

    mvn -pl java package assembly:single -DskipTests=true

- fix a comment in TestShim.java (missed in #741)

- run Maven test against MySQL
(this is an update to CLDR-8383 CI against mysql (#705) and
corrects 6adc82b6b6116bb67e046ec129adb66f4aa6aac4 which introduced
Maven CI against Derby)

- Move more pluginDependencies up from cldr-apps/pom.xml to tools/pom.xml

No plugin versions are specified in sub-poms now.

-----

Note: This unblocks CLDR-8383 from a code perspective.  The only thing holding us to Derby is
user docs / dev environment setup!

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -27,11 +27,21 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: Build with Maven
-        run: mvn -s .github/workflows/mvn-settings.xml -B package --file tools/pom.xml -DskipTests=true
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B package --file tools/pom.xml
+          -DskipTests=true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Setup MySQL
+        run: |
+          sudo systemctl start mysql.service
+          sleep 3
+          mysql --user=root --password=root < tools/cldr-apps/test-setup.sql
       - name: Test with maven
-        run: mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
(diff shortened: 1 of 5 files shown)


.github/workflows/maven.yml
```diff
@@ -27,11 +27,21 @@ jobs:
...
+      - name: Setup MySQL
+        run: |
+          sudo systemctl start mysql.service
+          sleep 3
+          mysql --user=root --password=root < tools/cldr-apps/test-setup.sql
       - name: Test with maven
-        run: mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
+          '-Dorg.unicode.cldr.unittest.web.jdbcurl=jdbc:mysql://cldrtest:VbrB3LFCr6A!@localhost/cldrtest?useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC'
```
btangmu, line 43: What's VbrB3LFCr6A!
?

btangmu: I see now it's in test-setup.sql. Is there any security risk in having that in git? And, could we define it once and name it, rather than repeating it in these two files?

srl295: @btangmu There isn't a risk, because that user is only setup within the git envinronment.

srl295: @btangmu I captured your comment in https://unicode-org.atlassian.net/browse/CLDR-8383 - can we merge this (not a regressin in this respect) and fix in the future?)

tools/pom.xml
```diff
@@ -237,6 +236,10 @@
 					<artifactId>maven-jar-plugin</artifactId>
 					<version>3.0.2</version>
 				</plugin>
+				<plugin>
+					<artifactId>maven-war-plugin</artifactId>
+					<version>3.2.2</version>
```
btangmu, line 241: Here it's version 3.2.2, while in cldr-apps/pom.xml the version is removed above. Why the difference? In general if version is omitted would we get the most recent?

srl295: cldr-apps is a child and inherits from the parent.  this way we only have one place to maintain all the version numbers

btangmu (approved):

PR #746, opened 2020-09-30
CLDR-11585 maven: refactor cldr-apps source

srl295:
- completely refactor cldr-apps source locations to match the
  Maven standard layout
- retain ant as buildable (for now!)

Eclipse:
- update .classpath so that source and build work fine.
- Just ignore errors in TestShim if not on maven.

Tests:
- retains TestShim, with some comments helping with the eclipse warnings
- fix TestAnnotations to support the new source location
- update comments and fix order of assert args
- make sure web app fails properly on failure


srl295:
@btangmu @macchiati — I think (if you import the projects as eclipse projects, not as maven) this won't mess up your eclipse paths at all. 

I haven't tried importing as maven projects yet. It might work a little better.

srl295:
NOTE: if you import these projects into Eclipse as maven, they still don't work well THERE, because tools/java still has the old source layout. I need an entirely different set of .classpath files in order to make that work.  So that will have to wait until we migrate java also

btangmu (approved):

PR #747, opened 2020-09-30
CLDR-14205 Manual trigger for commit checker

srl295:
CLDR-14205 

- new ansible playbook for commit checker setup
- update ansible-lint task to run all playbooks
- update ansible-lint to lint yaml also
- new workflow commit-checker (manually run)

### DIFF
--- a/.github/workflows/commit-checker.yml
+++ b/.github/workflows/commit-checker.yml
@@ -1,0 +1,86 @@
+# Manually run: commit checker
+
+name: commit-checker
+
+# Triggered on push to master,
+# or PR against master.
+on:
+  workflow_dispatch:
+    inputs:
+      end-ref:
+        description: CLDR Commit to check
+        required: true
+        default: 'master'
+      fix-version:
+        description: Fix Version (e.g. 38)
+        required: true
+      last-version:
+        description: Last Version (e.g. 37)
+        required: true
+      tool-repository:
+        description: Commit Checker ICU repo
+        required: true
+        default: 'unicode-org/icu'
+      tool-ref:
+        description: Commit Checker ICU branch/tag
+        required: true
+        default: 'master'
+jobs:
+  build:
+    name: Run Checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: clone CLDR
+        uses: actions/checkout@v2.3.3
+        with:
+          ref: ${{ github.event.inputs.end-ref }}
+          lfs: false # not needed here
+          path: cldr
+          fetch-depth: 0 # expensive, but needed for checker
+      - name: clone ICU for tools
+        uses: actions/checkout@v2.3.3
+        with:
+          repository: ${{ github.event.inputs.tool-repository }}
+          ref: ${{ github.event.inputs.tool-ref }}
+          lfs: false # not needed here
+          path: icu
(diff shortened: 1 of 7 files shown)


srl295:
image

example runs @ https://cldr-smoke2.unicode.org/cldrcc/

.github/workflows/commit-checker.yml
```diff
@@ -0,0 +1,90 @@
...
+          END_REF: ${{ github.event.inputs.end-ref }}
+          FV: ${{ github.event.inputs.fix-version }}
+          LV: ${{ github.event.inputs.last-version }}
+          # the RSA key for connecting
+          RSA_KEY_CCC: ${{ secrets.RSA_KEY_CCC }}
+          # the SSH port
+          CCC_USER: ${{ secrets.CCC_USER }}
+          CCC_PORT: 22
+          # the SSH host
+          CCC_HOST: cldr-smoke2.unicode.org
```
btangmu, line 83: When we switch the DNS and cldr-smoke2 becomes cldr-smoke, code like this will need updating, likewise for st2 and st ...

.github/workflows/commit-checker.yml
```diff
@@ -0,0 +1,90 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      end-ref:
+        description: CLDR Commit to check
+        required: true
+        default: 'master'
+      fix-version:
+        description: Fix Version (e.g. 38)
+        default: 38 # TODO: remove, for testing only
```
btangmu, line 16: When can/should these defaults be changed or removed?

.github/workflows/commit-checker.yml
srl295: Yes.  However, when I made these parameters ('secrets') it obscured them in the logs. So I think it's better this way.

.github/workflows/commit-checker.yml
srl295: What I thought of doing is to leave the defaults in for now to make it easier to test (you can just click and click).   And then when we change the DNS over, then remove those so that you *have* to put in a version number.  There's no reason to keep updating the defaults then.

srl295: Also, right now it points at my branch… while we are still making changes to the tool. when CLDR is happy with the output, then i will attempt to merge my branch into ICU. Then I can update this workflow to point at ICU proper.

srl295: I've removed many of the defaults.

tools/scripts/ansible/redact-db-playbook.yml
```diff
@@ -54,7 +54,8 @@
         group: cldradmin
         mode: '0640'
     - name: fetch and unzip db
-      command: bash /home/cldradmin/fetch-db.sh
```
srl295, line 57: @btangmu https://ansible-lint.readthedocs.io/en/latest/default_rules.html#commands-should-not-change-things-if-nothing-needs-doing normally ansible (it seems) wants to only change state to a fixed status. I think our usage of this playbook is fine, but we need to quell the linter warning.

btangmu (dismissed):

btangmu (approved):

PR #752, opened 2020-10-01
CLDR-13982 button to download flagged items to CSV

srl295:
CLDR-13982

- TC only (because it includes emails)

-----
** demo **
- in right panel:
image

- contents:
image

```csv
LOCALE,LAST_MOD,ID,ORG,EMAIL,XPATH
jv,2019-01-18 06:21:41,1930,Wikimedia,redacted@example.com,"//ldml/characters/exemplarCharacters[@type=""index""]"
so,2019-01-15 04:12:42,2043,Google,redacted@example.com,"//ldml/characters/exemplarCharacters[@type=""index""]"
```

### DIFF
--- a/tools/cldr-apps/src/main/webapp/DataExport.jsp
+++ b/tools/cldr-apps/src/main/webapp/DataExport.jsp
@@ -17,8 +17,12 @@
         response.sendRedirect(request.getContextPath());
         return;
     }
-
-    if (action.equals("list") && UserRegistry.userIsVetter(sess.user)) {
+    if (action.equals("flagged") && UserRegistry.userIsTC(sess.user)) {
+        final String query = "SELECT      flag.locale AS locale,     flag.last_mod AS last_mod,     user.id AS user,     user.org AS org,     user.email AS email,     xpath.xpath AS xpath " +
+        "FROM   " +  DBUtils.Table.VOTE_FLAGGED + " AS flag,     cldr_xpaths AS xpath,     cldr_users AS user WHERE     xpath.id = flag.xpath         AND user.id = flag.submitter ORDER BY last_mod DESC";
+        response.setHeader("content-disposition", "attachment;  filename=\"SurveyTool_flagged.csv" + "\"");
+        DBUtils.writeCsv(query, out);
+    } else if (action.equals("list") && UserRegistry.userIsVetter(sess.user)) {
         if ("ALL".equals(org)) {
             org = null;
         }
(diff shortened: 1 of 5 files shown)


macchiati (commented):
Looks good to me, but Tom should also review.

tools/cldr-apps/src/main/webapp/DataExport.jsp
```diff
@@ -17,8 +17,12 @@
         response.sendRedirect(request.getContextPath());
         return;
     }
-
-    if (action.equals("list") && UserRegistry.userIsVetter(sess.user)) {
+    if (action.equals("flagged") && UserRegistry.userIsTC(sess.user)) {
```
btangmu, line 20: Given the need to migrate away from JSP, it would be more ideal for new code like this to go into a Java file

tools/cldr-apps/src/main/webapp/js/survey.js
```diff
@@ -2144,9 +2145,10 @@ require(["dojo/ready"], function(ready) {
 			setLastShown(hideIfLast);
 
 			/*
-			 * TODO: clarify or rename; this td isn't really a td, is it?
+			 * This is the temporary fragment used for the
+			 * "info panel" contents.
```
btangmu, line 2149: Thanks for updating the comments and names!

btangmu (approved):
LGTM, other than the jsp which I hope we can migrate soon

tools/cldr-apps/src/main/webapp/DataExport.jsp
srl295: @btangmu ideally, but it's parallel to existing code

srl295: @btangmu also, this is just a first change, @macchiati to identify other places where we should do this same download.

macchiati:
If there is an action for me, can you file a ticket?

srl295:
> If there is an action for me, can you file a ticket?

- I set you as the reviewer for CLDR-13982, can you add comments there if there are other areas you would like to see with a download button?

tools/cldr-apps/src/main/webapp/DataExport.jsp
btangmu: OK, maybe while this code is fresh in your mind and you're still working on this ticket would be a good time to move all of DataExport.jsp into a .java file :-)

PR #753, opened 2020-10-01
CLDR-11585 Remove unused CLDR_VET_WEB

btangmu:
-Also remove unused vetweb

-This is to stop spurious warnings after Maven switch

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11585


srl295 (approved):
Good riddance!

I tried to find the history of this, but I failed to.

PR #756, opened 2020-10-02
CLDR-13581 revert und.xml, fix .gitignore

srl295:
CLDR-13581

Partial revert of a6f279f1734de718309487f28f5084c5637d7c2d / #612
- seed/main/und.xml should not be modified. It is a sandbox.

Also. remove common/main/*/und*.xml from .gitignore

This should go into CLDR 38 - otherwise CLDR 38 will have junk (sandbox data) in it.


btangmu (approved):

srl295:
@pedberg-icu this really should go into CLDR 38.

pedberg-icu:
> @pedberg-icu this really should go into CLDR 38. Won't affect ICU.

Thanks, the JIRA ticket does have version 38 so that should happen anyway the next time I update the maint/maint-38 branch and integrate to ICU.

PR #757, opened 2020-10-02
CLDR-12002 move sandbox from und to mul

srl295:
CLDR-12002

- update UserRegistry to check SpecialLocales
- cleanup some hard coded use of und
- use 'root' as default from .js instead of und


tools/cldr-apps/src/main/webapp/js/CldrSurveyVettingLoader.js
```diff
@@ -1876,7 +1876,7 @@ function showV() {
 
 					var theLocale = surveyCurrentLocale;
 					if (surveyCurrentLocale === null || surveyCurrentLocale == '') {
-						theLocale = 'und';
+						theLocale = 'root';
```
srl295, line 1879: this is a fallback, for non-locale pages (such as the home screen)

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
```diff
@@ -1793,8 +1794,8 @@ private static Object userCanAccessForumWhy(User u, CLDRLocale locale) {
             return null; // Admin can modify all
         if (userIsTC(u))
             return null; // TC can modify all
-        if (locale.getLanguage().equals("und")) { // all user accounts can write
-            // to und.
+        if (SpecialLocales.getType(locale) == SpecialLocales.Type.scratch) {
```
srl295, line 1797: we actually could have multiple sandbox locales… i.e. different scripts, etc… without needing to revisit this code.

srl295:
So this actually turned up some architectural issues, which took some digging.

- we have tests that check that the LangTagParser parent of 'root' is `und` and then verifies that `und` exists (yes, the "sandbox" locale)
- LikelySubtags uses `und` to mean (as it does mean) `unknown`.  Yet another reason to disentangle und from  the sandbox.
- `<languageData>` has:
```xml
		<language type="und" territories="AQ BV CP GS HM" alt="secondary"/>
```
and then in `<territoryInfo>`:
```xml
		<territory type="HM" gdp="49010" literacyPercent="99" population="1">	<!--Heard & McDonald Islands-->
			<languagePopulation type="und" literacyPercent="100" populationPercent="100" references="R1064"/>	<!--Unknown language-->
		</territory>
```

- There didn't appear to be any defaultContent reference to `und_ZZ` that I could find. The likely subtag expansion for `und` is `en_Latn_US` .


(I wondered if `mul` should live in a parallel tree to `seed` / `common` (yes, another one,. maybe 'sandbox'?)  owned by the survey tool, so that it is NOT visible to normal CLDR tests, but only appears in the Survey Tool. There are downsides to that, particularly, we would not be able to verify that `mul` would function properly until it gets to the survey tool.)

-----

I think this exercise does show that not using `und` as a test locale is a useful thing to do, it's just a question of how to best handle it. 
 - [ ] Add special case code to SupplementalDataInfo for mul?
 - [ ] Add dummy values to language-pop data for mul?
 - [ ] something else?

macchiati:
⟩ wondered if mul should live in a parallel tree to seed / common (yes,
another one,. maybe 'sandbox'?)

Sounds like a good idea to me

srl295:
Think what I will do is:
- update supplemental data info, etc, to recognize `SpecialLocale.Type.sandbox` as special. Or actually, may be enough to simply exclude such a locale from certain consistency tests.
- moving `mul*` into cldr-apps' own data
- have a special cldr-apps test to verify well-formedness of the mul data
- exclude `mul` from vetting data output (i.e. it will never be part of the check-in process)
- on startup, ST will recreate the sandbox data on disk somewhere so that the `CLDRFile.Factory` etc can be pointed at it.

macchiati:
sounds good

Mark

srl295:
@macchiati PTAL i think it is ready now.

Note that the new JUnit test included won't run from the 'ant' command line, but the cldr-mvn run does pick it up.

srl295:
@pedberg-icu this one should not go into 38.

pedberg-icu:
> @pedberg-icu this one should not go into 38. I'm fine with holding it a bit from being merged if needed.

The JIRA ticket has version 39 so I would not normally merge it to maint/maint-38. The only issue is if this PR is merged and then subsequent 38-targeted PRs depend on these changes (and they might since there is significant churn here).

macchiati (commented):
Looks ok to me, but Tom should review the ST part.

tools/cldr-apps/src/main/java/org/unicode/cldr/util/SandboxLocales.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.UnsupportedEncodingException;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+import org.unicode.cldr.util.XPathParts.Comments;
+
+public class SandboxLocales {
+    File base;
```
btangmu, line 18: Could these be private?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
```diff
@@ -1176,7 +1177,7 @@ String setLocales(WebContext ctx, int theirId, String theirEmail, String newLoca
 
         if (!intLocs) {
             newLocales = normalizeLocaleList(newLocales);
-            if (newLocales.isEmpty()) newLocales = "und";
+            if (newLocales.isEmpty()) newLocales = "mul";
```
btangmu, line 1180: Ideally there might be a constant for the string literal "mul", referring to tr35 "The content contains more than one language or text that is simultaneously in multiple languages (such as brand names)."

common/supplemental/likelySubtags.xml
```diff
@@ -3730,5 +3730,8 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ ?; Yi; ? } => { Sichuan Yi; Yi; China }-->
 		<likelySubtag from="und_Zanb" to="cmg_Zanb_MN"/>
 		<!--{ ?; Zanabazar Square; ? } => { Classical Mongolian; Zanabazar Square; Mongolia }-->
+
+		<!-- testing -->
```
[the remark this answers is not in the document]
btangmu: If so maybe the comment "testing" could be expanded a bit

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
srl295: @btangmu this particular line is only for the user update - it's not related to "multiple languages" per se. Should it instead be instead a constant DEFAULT_LOCALE or something?

tools/cldr-apps/.classpath
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
srl295: I planned to .gitignore them when the maven migration is done, when eclipse will no longer use the project files as is.

common/supplemental/likelySubtags.xml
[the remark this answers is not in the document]
srl295: Hm.  @macchiati I see the "not to be patched by hand". Is there a better solution for this file?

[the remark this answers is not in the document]
macchiati: This file is generated, so the tool needs to be modified for it. However, it would be best to avoid adding "test only" locales to this. Is there a way to avoid adding "mul" here?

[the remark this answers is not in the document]
srl295: @macchiati Now that these locales don't exist when tests are run, we might be able to get away with just not mentioning them in likely subtag.  Is there any other machinery that would break  (in checks, etc) if likely subtags aren't present? Worst case I would add exclusions to any such tests.

I'll try this.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
btangmu: Yes, maybe SpecialLocales.DEFAULT_LOCALE? Anything like that is an improvement for encapsulation

srl295:
might be better to just remove the 'default locale' behavior entirely from
this code. Default Locale might be misleading.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
srl295: removed.

tools/cldr-apps/src/main/java/org/unicode/cldr/util/SandboxLocales.java
srl295: done

srl295:
fixed merge commits

btangmu (approved):

PR #758, opened 2020-10-03
CLDR-12010 Allow for both c and e in the spec

macchiati:
Allow for both c and e in the spec; update code for DecimalQuantity in the test

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-12010

### DIFF
--- a/docs/ldml/tr35-numbers.html
+++ b/docs/ldml/tr35-numbers.html
@@ -2523,7 +2523,7 @@ System Data</a>.</em> ) &gt;<br>
     <p>The xml value for each pluralRule is a <em>condition</em>
     with a boolean result that specifies whether that rule (i.e.
     that plural form) applies to a given numeric value <em>n</em>,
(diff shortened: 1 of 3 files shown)


macchiati:
Can one of you approve this?

echeran (approved):
LGTM.  Does the "future-proofing" in the official docs (potential future disentangling of scientific and compact notation syntax in the plural samples data) and similar on the implementation side (easier potential future switchover from FixedDecimal to DecimalQuantity).

srl295 (approved):

macchiati:
Yes, there is a note (hard to read in the raw HTML) that 'e' is currently a
synonym for 'c', but may be given a separate definition in the future.

(The only reason to keep 'e', as we noted, is because we can't change the
rules because ICU couldn't read them without changes, but we don't need to
go into that.)

Mark

PR #760, opened 2020-10-04
CLDR-13825 Add names for currency characters

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13825


macchiati:
This adds names for the Unicode currency symbols.

- It is limited to symbols that have an ISO currency code, and where there is one main country that uses that symbol.
- It derives the name from that ISO currency code, based on a table in cldr/tools/java/org/unicode/cldr/util/data/codepointToIsoCurrencyCode.tsv
- The code is in ldr/tools/java/org/unicode/cldr/tool/GenerateDerivedAnnotations.java
- In the new generated files, there are some spurious changes due to the reordering of "man/woman with beard": ignore those.

macchiati:
The changes to FixedDecimal are harmless; identical to https://github.com/unicode-org/cldr/pull/758/files#diff-eb8442e1320fc6d267fb7043a7041074

xadxura (approved):

Kristi27 (approved):

srl295:
looks like `mt.xml` got zapped in this one

PR #762, opened 2020-10-05
CLDR-14216 Automate ST db redaction for cldr-smoke

btangmu:
-New redact-db-playbook.yml to set up and perform redaction

-New redact-db.sql, fetch_db_sh.j2 and redact_db_sh.j2

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14216

### DIFF
--- a/tools/scripts/ansible/redact-db-playbook.yml
+++ b/tools/scripts/ansible/redact-db-playbook.yml
@@ -1,0 +1,73 @@
+- hosts: redactdb
+  vars_files:
+    - vars/main.yml
+    - local-vars/local.yml
+  tasks:
+    - name: set up /home/cldradmin/.ssh/config to fetch db backup
+      template:
+        src: templates/cldrbackup/config.j2
+        dest: /home/cldradmin/.ssh/config
+        owner: cldradmin
+        group: cldradmin
+        mode: '0640'
+    - name: set up /home/cldradmin/.ssh/id_rsa
+      no_log: true
+      copy:
+        src: local-vars/cldrbackup/id_rsa
+        dest: /home/cldradmin/.ssh/id_rsa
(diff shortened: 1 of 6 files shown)


tools/scripts/ansible/redact-db-playbook.yml
```diff
@@ -0,0 +1,73 @@
...
+      command: bash /home/cldradmin/fetch-db.sh
+      become: yes
+      become_user: cldradmin
+    - name: stop tomcat
+      become: true
+      service:
+        name: tomcat8
+        state: stopped
+    - name: redact db
+      command: bash /home/cldradmin/redact-db.sh
```
srl295, line 66: good, so this one is a little different than other playbooks in that it is meant to be run only when redact is needed. Sounds good.  It's not included in setup playbook, for example.

srl295 (approved):

tools/scripts/ansible/redact-db-playbook.yml
btangmu: Right. Running redact-db-playbook.yml is a manual operation that combines configuration with performing an actual redaction. Everything is in Ansible. In the long run we might use a cron job and/or a button like https://github.com/unicode-org/cldr/actions?query=workflow%3Aproduction for the redaction itself. Configuration  without immediately doing the redaction itself should be enabled and probably included in setup-playbook.yml. What do you think?

srl295: > Configuration without immediately doing the redaction itself should be enabled and probably included in setup-playbook.yml. 

I think this is fine how it is for now, because we might adjust the parameters in between redaction.

If we were _automating_ the redaction, then yes it would move to setup-playbook.

PR #763, opened 2020-10-06
CLDR-13771 update scripts for azure migration

srl295:
CLDR-13771
- st2 ➡️  st
- cldr-smoke2 ➡️  cldr-smoke


srl295:
this isn't that critical because only the ansible scripts are updated. already tested locally

btangmu (approved):

PR #764, opened 2020-10-07
CLDR-14066 Fix display of constraints in Delta DTD charts

macchiati:
Remove + on attribute if only constraints changed. Suppress identical constraint words for a more concise format. 
Results are in https://unicode-org.github.io/cldr-staging/charts/38/supplemental/dtd_deltas.html

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14066


macchiati:
That is, once they sync, results are in https://unicode-org.github.io/cldr-staging/charts/38/supplemental/dtd_deltas.html

pedberg-icu (approved):

PR #765, opened 2020-10-07
CLDR-13922 in spec, use "legacy" rather than "grandfathered" where possible

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13922

In spec, use "legacy" rather than "grandfathered" where possible. Similar to ICU changes in https://github.com/unicode-org/icu/pull/1243/files

Also fix html validation errors with ampersand in a link


markusicu (approved):
lgtm

PR #768, opened 2020-10-08
CLDR-14225 tz database 2020b updates

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14225


pedberg-icu (approved):

RobThree:
Shouldn't the version have been updated?

PR #769, opened 2020-10-09
CLDR-14226 reduce use of term grandfathered in code

markusicu:
https://unicode-org.atlassian.net/browse/CLDR-14226

For affected enums, I replaced several calls to Enum.valueOf(String) with .from(String) or .fromString(String) and temporarily still handled the old name. Set a breakpoint, ran the unit tests, updated call sites, removed the alias.

For other string comparisons I also set temporary breakpoints to make sure the unit tests don't show use of the old name.

I did not change what looked like data file syntax.

### DIFF
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/LocaleMatcherTest.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/LocaleMatcherTest.java
@@ -313,7 +313,7 @@ public class LocaleMatcherTest extends TestFmwk {
             matchEnHantTw < matchZh);
     }
 
-    public void testMatchGrandfatheredCode() {
+    public void testMatchLegacyCode() {
         if (logKnownIssue("CLDR-14166", "Skip until CLDR updated for new ICU4J LocaleMatcher")) {
             return;
         }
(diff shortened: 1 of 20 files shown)


macchiati (approved):

markusicu:
The ticket is not yet accepted. Should I wait until the TC takes a look?
I wanted to nudge this along but don't want to do the whole thing here. Is someone else willing to own it?

macchiati:
I was only created 2 days ago. I have no objections to its going into 38 as
long as nothing is destabilized. But peter should make the call.

pedberg-icu:
yes, I think this is good for 38, I can own the ticket, and think this should move to Accepted if nobody objects.

Thanks Markus for the PR. I can look at this in more detail Sunday night.
- Peter E

macchiati:
I support Accepted

srl295 (approved):

pedberg-icu (approved):

pedberg-icu:
CLDR TC agreed to accept this for 38. Merging now.

markusicu:
Thanks for approving and merging, but this is just a start. The ticket should not have been closed. I reopened it.

PR #770, opened 2020-10-10
CLDR-14194 describe compound unit localization

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14194


macchiati:
Here is a first cut at the compound name name construction. As before, I think this is easier to review if it is accepted so that the html can be read at github.io rather than trying to make sense of raw HTML.

macchiati:
If you are going to look at the raw HTML, I recommend ⚙️, split, hide whitespace differences.
I'd especially appreciate comments from people who've looked at the compound unit name formatting, like @hugovdm

sffc:
Cool; we will work on implementing this for ICU 69.

srl295 (approved):
lgtm

hugovdm (commented):
[/me hereby ticks this off on /my closed "review requests" list - pardon noise on old threads]

PR #772, opened 2020-10-12
CLDR-13300 fix tests for cldr-staging

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13300

### DIFF
--- a/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestInheritance.java
+++ b/tools/cldr-unittest/src/org/unicode/cldr/unittest/TestInheritance.java
@@ -66,7 +66,7 @@ public class TestInheritance extends TestFmwk {
     public void TestLocalesHaveOfficial() {
         // If we have a language, we have all the region locales where the
         // language is official
-        Set<String> SKIP_TERRITORIES = new HashSet<String>(Arrays.asList("001",
+        Set<String> SKIP_TERRITORIES = new HashSet<>(Arrays.asList("001",
             "150"));
         for (Entry<String, R2<List<String>, String>> s : dataInfo
             .getLocaleAliasInfo().get("territory").entrySet()) {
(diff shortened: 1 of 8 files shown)


macchiati:
Various patches to make the tests work on the cldr-staging data. Mostly due to seed being missing. To run the tests on cldr-staging, you need to do:

-DCLDR_DIR=/Users/markdavis/GitHub/macchiati/cldr-staging/production
-DCLDR_SKIP_SEED

I could have just set "skip seed" based on CLDR_DIR, but thought this was a bit more general. Comments?

macchiati:
Whoever reviews this, if it looks ok please also take care of https://unicode-org.atlassian.net/browse/CLDR-13300 (set yourself as reviewer, and close).

pedberg-icu (approved):
This is very handy, thanks, I was having to redefine env variables and then could only run just a partial set of tests on cldr-staging data.

pedberg-icu:
I am going to go ahead and merge...

macchiati:
Thanks!

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestInheritance.java
```diff
@@ -247,18 +247,18 @@ public void TestLikelyAndDefaultConsistency() {
...
 
-        Set<String> knownMultiScriptLanguages = new HashSet<String>(Arrays.asList("bm", "ha"));
+        Set<String> knownMultiScriptLanguages = new HashSet<>(Arrays.asList("bm", "ha"));
         // get multiscript locales
         for (String localeID : availableLocales) {
             String script = ltp.set(localeID).getScript();
             final String base = ltp.getLanguage();
-            if (!availableLocales.contains(base)) {
-                errln("Missing base locale for: " + localeID);
+            if (!availableLocales.contains(base) && !base.equals("und")) {
```
srl295, line 260: I like your fix, I had to do similar in the 'move sandbox to mul' PR https://github.com/unicode-org/cldr/pull/757

tools/cldr-unittest/src/org/unicode/cldr/unittest/TestInheritance.java
```diff
@@ -247,18 +247,18 @@ public void TestLikelyAndDefaultConsistency() {
...
+                errln("Missing lang-subtag for: " + localeID);
```
srl295, line 261: in https://github.com/unicode-org/cldr/pull/757 I made sure to print out the `base` locale also, because that's actually the one that's missing (not `localeID`)

PR #774, opened 2020-10-12
CLDR-13973 fixing BOM streams and xml validator

srl295:
- remove BOM-removing streams
- add an XMLValidator option --justCheckBom for checking BOM
- make XMLValidator threaded and able to return an error status
- add a maven build step to run XML files through the XMLValidator

NOTE: this is expected to fail, as there's one file with a BOM.


macchiati (approved):

srl295:
see also #795 as a followon, we needed to re-allow BOM input for historical content

PR #775, opened 2020-10-12
CLDR-14180 Bill Pugh singletons

srl295:
CLDR-14180

Multiple classes, notably:
- DBUtils: need special handling for test instances

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/CalendarSortMode.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/CalendarSortMode.java
@@ -220,14 +220,14 @@ public class CalendarSortMode extends SortMode {
     }
 
     @Override
-    Comparator<DataRow> createComparator() {
-        return comparator();
+    Comparator<DataRow> getComparator() {
+        return ComparatorHelper.COMPARATOR;
     }
 
-    public static Comparator<DataRow> comparator() {
-        final int ourKey = SortMode.SortKeyType.SORTKEY_CALENDAR.ordinal();
-        final Comparator<DataRow> codeComparator = CodeSortMode.comparator();
-        return new Comparator<DataRow>() {
+    private static final class ComparatorHelper {
+        static final Comparator<DataRow> COMPARATOR = new  Comparator<DataRow>() {
+            final int ourKey = SortMode.SortKeyType.SORTKEY_CALENDAR.ordinal();
+            final Comparator<DataRow> codeComparator = CodeSortMode.internalGetComparator();
             @Override
             public int compare(DataRow p1, DataRow p2) {
                 if (p1 == p2) {
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/CodeSortMode.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/CodeSortMode.java
@@ -36,13 +36,21 @@ public class CodeSortMode extends SortMode {
     }
 
     @Override
-    Comparator<DataRow> createComparator() {
-        return comparator();
+    Comparator<DataRow> getComparator() {
+        return ComparatorHelper.COMPARATOR;
     }
 
-    public static Comparator<DataRow> comparator() {
-        return new Comparator<DataRow>() {
-            final Collator myCollator = createCollator();
+    static final private class CollatorHelper {
+        private static Collator createCollator() {
+            RuleBasedCollator rbc = ((RuleBasedCollator) Collator.getInstance());
+            rbc.setNumericCollation(true);
+            return rbc.freeze();
+        }
+        static final Collator COLLATOR = createCollator();
+    }
+    static final private class ComparatorHelper {
+        static final Comparator<DataRow> COMPARATOR = new Comparator<DataRow>() {
+            final Collator myCollator = CollatorHelper.COLLATOR;
 
             @Override
             public int compare(DataRow p1, DataRow p2) {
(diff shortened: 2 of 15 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/CalendarSortMode.java
```diff
@@ -220,31 +220,33 @@ public boolean isMember(DataRow p) {
...
 
     @Override
-    Comparator<DataRow> createComparator() {
-        return comparator();
+    Comparator<DataRow> getComparator() {
+        return COMPARATOR;
     }
 
-    public static Comparator<DataRow> comparator() {
+    static final Comparator<DataRow> COMPARATOR = new  Comparator<DataRow>() {
```
macchiati, line 227: This doesn't appear right. The key to the Bill Pugh singletons is that the statics are in their own static class. That keeps them from being created until needed.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/CodeSortMode.java
```diff
@@ -36,13 +36,20 @@ String getName() {
...
-    Comparator<DataRow> createComparator() {
-        return comparator();
+    Comparator<DataRow> getComparator() {
+        return CollatorHelper.COMPARATOR;
     }
 
-    public static Comparator<DataRow> comparator() {
-        return new Comparator<DataRow>() {
-            final Collator myCollator = createCollator();
+    static final private class CollatorHelper {
```
macchiati, line 43: This looks better, except that each static should be in its own helper. Otherwise you create all of them when any one of them is created.

macchiati (changes requested):
Some initial comments: I know this is draft.
I assume just you haven't gotten to some classes like CLDRConfig.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/CalendarSortMode.java
srl295: The `SortMode` subclasses are intertwined, many sibling classes make use of the Comparator instance from another sibling. But, I can make these lazily instantiated.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/CodeSortMode.java
srl295: The Comparator depends on the Collator.

macchiati:
ok
Mark

srl295:
@macchiati I think I've addressed everything.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/CodeSortMode.java
macchiati: ok, makes sense

macchiati (approved):

srl295:
Ok, mostly done.  Note that many of the members on CLDRConfig (how did we get so many?) are really logically `static` because they depend on calling CLDRPaths statically. Ideally these would be instance members.

macchiati:
Looks good. I think there are other cases of "synchronized" that could profit from this technique, like:

    public Factory getCommonAndSeedAndMainAndAnnotationsFactory() {
        synchronized (FULL_FACTORY_SYNC) {

Probably the focus should be on the ones likely to be involved in thread contention though.

srl295:
> Looks good. I think there are other cases of "synchronized" that could profit from this technique, like:
> 
> 
> 
>     public Factory getCommonAndSeedAndMainAndAnnotationsFactory() {
> 
>         synchronized (FULL_FACTORY_SYNC) {

Note that this one is not static for example. Making it static means a call back to CLDRConfig.getInstance(). Which may be ok.

macchiati:
The call back to CLDRConfig.getInstance() only happens once in the lifetime
of the static... so the cost is low.

As to how many... Well, if we ever get rid of SEED that will it down a bit!

Mark

srl295:
> The call back to CLDRConfig.getInstance() only happens once in the lifetime
> of the static... so the cost is low.

Yes, the cost is low, I was just commenting on the API change from non-static to static. In the future we might want multiple CLDRConfigs. But there's a lot that would have to happen to get there.

srl295:
> > Looks good. I think there are other cases of "synchronized" that could profit from this technique, like:
> > ```
> > public Factory getCommonAndSeedAndMainAndAnnotationsFactory() {

I suppose I could leave these getters as non-static (since that's how they are called today) even though the implementation is static, what do you think? Then the API surface would remain the same

srl295:
@macchiati Ready for review, assuming tests pass. (still running)

I did deprecate `CLDRConfig.getStandardCodes()` as `StandardCodes.make()` has been a singleton for a long time. This could be inlined, but should be more performant with this overall change.

macchiati (approved):

macchiati:
> I did deprecate CLDRConfig.getStandardCodes() as StandardCodes.make() has been a singleton for a long time. This could be inlined, but should be more performant with this overall change.

You can do more than that, and just refactor to inline. What about SupplementalDataInfo?

srl295:
> > I did deprecate CLDRConfig.getStandardCodes() as StandardCodes.make() has been a singleton for a long time. This could be inlined, but should be more performant with this overall change.
> 
> You can do more than that, and just refactor to inline. What about SupplementalDataInfo?

I will take a look at SDI and inline this in a followon PR.

macchiati:
Sounds good. Separate PR is better anyway.

Mark

PR #778, opened 2020-10-12
CLDR-14205 commit checker: update

srl295:
CLDR-14205
- improve markdown processing

This is an update to the calling button only. It will solve the "links from the TOC do not work".
The markdown processing wasn't specified properly previously.

### DIFF
--- a/.github/workflows/commit-checker.yml
+++ b/.github/workflows/commit-checker.yml
@@ -8,9 +8,8 @@ on:
   workflow_dispatch:
     inputs:
       end-ref:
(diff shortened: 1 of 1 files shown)


srl295:
Example output:  https://cldr-smoke.unicode.org/cldrcc/CLDR-Report-2020-10-12-v38-4748c79148.html
Example run: https://github.com/unicode-org/cldr/runs/1244514237?check_suite_focus=true

btangmu (approved):

PR #779, opened 2020-10-12
CLDR-14220 Azure migration: log rotation

btangmu:
-Add Ansible setup for /etc/logrotate.d/surveytool-logs

-Rotate /var/lib/tomcat8/cldr/exception.log same as on old server

-Also similarly rotate /var/lib/tomcat8/cldr/cldrmail.log

-Note: catalina.out already uses system default /etc/logrotate.d/tomcat8

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14220

### DIFF
--- a/tools/scripts/ansible/setup-playbook.yml
+++ b/tools/scripts/ansible/setup-playbook.yml
@@ -47,6 +47,23 @@
         owner: tomcat8
         group: tomcat8
         mode: 0775
+    - name: Rotate CLDR logs
+      blockinfile:
+        path: "/etc/logrotate.d/surveytool-logs"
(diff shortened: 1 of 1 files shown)


srl295 (approved):
LGTM, assuming this will create `/etc/logrotate.d/surveytool-logs` if it doesn't exist.

btangmu:
> LGTM, assuming this will create `/etc/logrotate.d/surveytool-logs` if it doesn't exist.

Yes, it will. And it already did, on st; I'll run it for cldr-smoke as well.
Thanks!

PR #780, opened 2020-10-13
CLDR-14128 document approximatelySign,sub/superscript, fix modifications and validation errs

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14128

1. Document:
- new number symbol approximatelySign (was already there, but not marked as a change)
- characterLabelPattern type atttribute values subscript and superscript
2. Repair the Modifications section, was somewhat mangled
3. Fix various pre-existing validation errors


macchiati (approved):
Peter, looks like you also did a bunch of cleanup on the HTML — Thanks!

PR #781, opened 2020-10-13
CLDR-14226 change more uses of "grandfathered", and of "blacklist", "crazy"

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14226

This changes more "uses" of "grandfathered" to "legacy", and also replaces uses of "blacklist" and "crazy".

There are remaining uses of "grandfathered" specifically in reference to BCP47, and remaining uses of "master" specifically referring to a github branch (and URLs for it), but there is not much we can do about those for the time being. I think this PR represents the final chnages in this area for CLDR v38. We can investigate more for v39 (with a new ticket).

Note that some sensitive terms are present just as (1) terms for which we have IPA transcription data, and (2) localized name of language "den".

### DIFF
--- a/common/supplemental/supplementalMetadata.xml
+++ b/common/supplemental/supplementalMetadata.xml
@@ -30,7 +30,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <attributes element="weekendStart" attribute="time" attributeValue="00:00"/>
         </suppress>
         <alias>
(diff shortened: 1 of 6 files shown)


markusicu (approved):
lgtm tnx

macchiati (approved):

PR #782, opened 2020-10-13
CLDR-13396 Redo the fix for MY & zh, but permanently

macchiati:
Update the underlying txt file and rerun the generation programs for likelySubtags and supplementalData. The process is fragile so filed https://unicode-org.atlassian.net/browse/CLDR-14231 for a longer-term fix.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13396


markusicu (approved):
lgtm tnx!

PR #784, opened 2020-10-14
CLDR-14226 replace a few more sensitive terms

markusicu:
https://unicode-org.atlassian.net/browse/CLDR-14226


docs/ldml/tr35.html
```diff
@@ -1147,7 +1147,9 @@ <h4><a name="Canonical_Unicode_Locale_Identifiers" href="#Canonical_Unicode_Loca
               <li>If there are additional subtags in the <i>replacement</i> value, add them to the result, but
                 only if there is no corresponding subtag already in the
                 tag.</li>
-              <li>Five special deprecated grandfathered codes (such as <em>i-default</em>) are in  type attributes, and are also replaced.</li>
+              <li>Five special deprecated <span class="removed">grandfathered codes (</span>
```
pedberg-icu, line 1150: This is unnecessary, the whole block that includes this is marked for removal. You should look at the html in a browser to see this.

docs/ldml/tr35.html
```diff
@@ -1283,7 +1285,9 @@ <h5><a name="Language_Tag_to_Locale_Identifier" href=
           <i>replacement</i> value, add them to the result, but
           only if there is no corresponding subtag already in the
           tag.</li>
-          <li>Five special deprecated grandfathered codes (such as <em>i-default</em>) are in  type attributes, and are also replaced.</li>
+          <li>Five special deprecated <span class="removed">grandfathered codes (</span>
```
pedberg-icu, line 1288: Same, this whole block is being removed.

docs/ldml/tr35.html
```diff
@@ -1312,7 +1316,7 @@ <h5><a name="Language_Tag_to_Locale_Identifier" href=
       "tr35-info.html#Supplemental_Data">Supplemental Data</a>,
       replace the script subtag with the <i>replacement</i>
       value. </li>
-      <li>If the tag is a deprecated grandfathered tag
+      <li>If the tag is a deprecated <span class="removed">grandfathered</span><span class="changed">legacy language</span> tag
```
pedberg-icu, line 1319: This is in a block that is already being removed.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
```diff
@@ -1664,7 +1664,7 @@ public String setSession() {
         if (password.isEmpty()) {
             password = field(SurveyMain.QUERY_PASSWORD_ALT);
         }
-        boolean letmein = SurveyMain.vap.equals(field("letmein")); // back door, using master password
+        boolean letmein = SurveyMain.vap.equals(field("letmein")); // back door, using main password
```
pedberg-icu, line 1667: Not sure whether master has special significance in this context, that is why I left this for further investigation.

tools/java/org/unicode/cldr/util/data/langtagRegex.txt
```diff
@@ -75,12 +75,12 @@ $langtag = (?: ( $language )
       (?: $s ( $privateUse ) )? 5%);
 
 # Here is the final breakdown, with capturing groups for each of these components
-# The variants, extensions, grandfathered, and private-use may have interior '-'
+# The variants, extensions, legacy, and private-use may have interior '-'
```
pedberg-icu, line 78: I think this is specifically referring to the $grandfathered expression here, not sure we want to change this unless we also change that.

pedberg-icu (changes requested):

pedberg-icu:
Markus, the spec changes you made are in blocks that are already marked for removal (surrounded by <div class="remove"> I think). You can see this if you look at these sections in a browser, don't just check the raw html. So I think these are all unnecessary.

Your other changes also may have other dependencies that require further investigation which is why I did not make them.

tools/java/org/unicode/cldr/util/data/langtagRegex.txt
markusicu: There are two .java files that mention this file name, and neither explicitly looks for or mentions "grandfathered".

tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
markusicu: done tnx

markusicu:
I checked with a fine-tooth comb. Nothing referred to tools/java/org/unicode/cldr/util/data/langtagRegex.txt so I deleted it. Judging from StandardCodes.java, it used to read that preprocessed file at some point, but now it is reading the raw language-subtag-registry file. Probably to save a preprocessing step.

markusicu:
Spec changes reverted. Other changes amended. Unit tests pass locally. PTAL

pedberg-icu (approved):
Thanks for the investigations and update!

markusicu:
FYI lstreg.txt was last updated on 2012-dec-18, so it's way out of date with the real thing.
See https://github.com/unicode-org/cldr/commits/master/tools/java/org/unicode/cldr/util/data/lstreg.txt

macchiati:
I don't think we use lstreg.txt for anything. Probably should be nuked.

Mark

PR #786, opened 2020-10-14
CLDR-13777 Logic for Losing in sublocales is faulty, Part Two

btangmu:
-In VoteResolver.setLocale, use locale (like fr_CA) instead of locale.getLanguageLocale (like fr)

-Fix deprecation warnings for single-arg setLocale

-Remove unused return value of VoteResolver.setLocale (make void)

-Remove dead code

-Some changes made automatically by Eclipse (whitespace; ...new... with angle brackets)

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13777


btangmu:
See https://github.com/unicode-org/cldr/pull/788
Same but no changes to .classpath or .settings

PR #788, opened 2020-10-14
CLDR-13777 Logic for Losing in sublocales is faulty, Part Two

btangmu:
-In VoteResolver.setLocale, use locale (like fr_CA) instead of locale.getLanguageLocale (like fr)

-Fix deprecation warnings for single-arg setLocale

-Remove unused return value of VoteResolver.setLocale (make void)

-Remove dead code

-Some changes made automatically by Eclipse (whitespace; ...new... with angle brackets)

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13777

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
@@ -100,9 +100,6 @@ public class CLDRConfigImpl extends CLDRConfig implements JSONString {
         cldrHomeSet = true;
     }
 
-    /**
-     * Never called for cldr-apps TestAll.java.
-     */
     private synchronized void init() {
         if (isInitted)
             return;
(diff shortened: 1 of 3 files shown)


srl295 (approved):

btangmu:
Thanks, Steven!

I'd like to get Mark's review as well on the essential part: in VoteResolver.setLocale, use locale (like fr_CA) instead of locale.getLanguageLocale (like fr)

This appears to have the desired effect of requiring 4 rather than 8 votes for most sublocales, for typical paths

btangmu:
Actually I'll go ahead and merge this, and follow up with Mark

macchiati:
That looks good to me. We want to make sure that the CLDR Organization
sublocales are treated properly, eg we probably want 8 votes for en_GB, but
for the ones we don't target, I think 4 might be appropriate. Let's put
this on the agenda for discussion,
Mark

PR #790, opened 2020-10-15
CLDR-14202 B Simple tool for checking conversions

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14202

### DIFF
--- a/tools/java/org/unicode/cldr/tool/VerifyConverterResults.java
+++ b/tools/java/org/unicode/cldr/tool/VerifyConverterResults.java
@@ -1,0 +1,340 @@
+package org.unicode.cldr.tool;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.UncheckedIOException;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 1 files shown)


macchiati:
Changing fixVersion to 39, since the code for the tool doesn’t have to be in 38. However, I'd like to go ahead and get it merged. We can enhance further in 39.

tools/java/org/unicode/cldr/tool/VerifyConverterResults.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    /*
+    public static void main(String[] args) throws UnsupportedEncodingException {
+        for (final String baseName : getBaseNames()) {
+            gatherData(baseName);
+        }
+    }
+    // Ugly hack to get base names
```
srl295, line 289: there's not a great way to do this

macchiati:
Yes I know that it's not the right way of reading the resource bundles but I ran out of time to do it. Wanted to focus first on the Jason side. Need to work on it more in 39.

srl295:
> Yes I know that it's not the right way of reading the resource bundles but I ran out of time to do it. Wanted to focus first on the Jason side. Need to work on it more in 39.

Right, what I meant is that I am not convinced there *is* a clean way to do it from where this code sits.  One idea might be to actually read the LDML2ICUConverter config data, however, that makes the verification a little more circular and less clean-slate.

macchiati:
I. Think we can run the tool after integrations, and just examine resource
bundles.

PR #791, opened 2020-10-15
CLDR-14220 Azure migration: set st.unicode.org hostname

btangmu:
-This is required for sending email from the production server

-New Ansible playbook: hostname-playbook, imported from setup-playbook

-Move other details of setup-playbook to new server-playbook, packages-playbook, certbot-playbook

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14220


tools/scripts/ansible/hostname-playbook.yml
```diff
@@ -0,0 +1,6 @@
+- hosts: prod
+  become: yes
+  tasks:
+    - name: Set the hostname
+      hostname:
+         name: st.unicode.org
```
srl295, line 6: should this be `{{ inventory_hostname }}` instead?

tools/scripts/ansible/server-playbook.yml
```diff
@@ -0,0 +1,180 @@
...
+  tasks:
+    - name: Install server packages
+      apt:
+        pkg:
+          - tomcat8
+          - tomcat8-admin # needed for deploy
+          - unzip # needed for deploy
+          # for monitoring
+          - prometheus-mysqld-exporter
+          # - prometheus-nginx-exporter # (not there yet)
```
srl295, line 18: nginx-exporter wasn't in the ubuntu version we have yet.

srl295 (approved):

tools/scripts/ansible/hostname-playbook.yml
btangmu: Yes! I hunted around for a way to get the "current" host, but didn't find this. I'll make a follow-up PR with that change, thanks!

I don't know if there's any reason to do the same for other hosts (like smoketest), so I limited it to "prod"

PR #795, opened 2020-10-16
CLDR-13973 restructure UTF-8, XML Reader

srl295:
CLDR-13973
- remove uses of unnecessary Charset.forName("UTF-8")
- re-enable the ability to take in BOM in XML (for compat)
- use InputSource / InputStream instead of Reader input.
- other simplifications


srl295:
bad import statement, updated.

hagbard:
Which code relates to the BOM stuff? Everything else is standard cleanup, but I cannot see a comment or class that indicates it changes behaviour with respect to BOMs.

I'm guessing it's the InputSource stuff, but that's not 100% clear. Maybe a comment for future maintainers?

markusicu (approved):
lgtm

srl295:
Using InputStream instead of Reader is the change. Simpler and lets the xml
code deal with the encoding rather than fixed “utf-8 without bom”

hagbard:
> Using InputStream instead of Reader is the change. Simpler and lets the xml code deal with the encoding rather than fixed “utf-8 without bom”

Ahh okay, maybe a comment there then (since someone might thing "Reader" is a better thing to use in the future).
Can you add a test that will now fail if the code didn't handle a BOM properly?

macchiati (commented):
My only question is whether the refactoring will mean that the reading of the string will be buffering.

btangmu:
Test failure: TestUserRegistry {
      Error: (TestAll.java:98) java.lang.IllegalArgumentException: Can't read data/TestUserRegistry.xml

That's almost the same as what I've been getting locally in Eclipse, though I get: TestUserRegistryErasing: /Users/tbishop/cldr_db_test/org.unicode.cldr.unittest.web.TestSTFactory_output
 {
      Error: (TestAll.java:101) java.lang.NullPointerException
java.lang.NullPointerException
	at org.unicode.cldr.unittest.web.TestSTFactory.runDataDrivenTest(TestSTFactory.java:723)

srl295:
Yes I’m working on that issue.

srl295:
> My only question is whether the refactoring will mean that the reading of the string will be buffering.

This leaves the exact reading up to the XML Parser. It does internally create a buffered reader.

srl295:
> > Using InputStream instead of Reader is the change. Simpler and lets the xml code deal with the encoding rather than fixed “utf-8 without bom”
> 
> Ahh okay, maybe a comment there then (since someone might thing "Reader" is a better thing to use in the future).

Updated the comment

> Can you add a test that will now fail if the code didn't handle a BOM properly?

We could. Although we are trying to _discourage_ use of the BOM.  I verified that this code is able to read the CLDR v1.7.2 `zh.xml`. Do we actually want that test case?

hagbard:
> > > Using InputStream instead of Reader is the change. Simpler and lets the xml code deal with the encoding rather than fixed “utf-8 without bom”
> > 
> > 
> > Ahh okay, maybe a comment there then (since someone might thing "Reader" is a better thing to use in the future).
> 
> Updated the comment
> 
> > Can you add a test that will now fail if the code didn't handle a BOM properly?
> 
> We could. Although we are trying to _discourage_ use of the BOM. I verified that this code is able to read the CLDR v1.7.2 `zh.xml`. Do we actually want that test case?

If it's not hard to add, I'd add one. Then at some point if/when we turn off BOM support we can see it fail so we know it's done.
I'll leave it up to you.

tools/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -331,9 +329,8 @@ private static CLDRFile load(String fileName, String localeName, InputStream fis
      */
     public CLDRFile loadFromInputStream(String fileName, String localeName, InputStream fis, DraftStatus minimalDraftStatus) {
         CLDRFile cldrFile = this;
-        InputStreamReader reader = new InputStreamReader(fis, Charset.forName("UTF-8"));
         MyDeclHandler DEFAULT_DECLHANDLER = new MyDeclHandler(cldrFile, minimalDraftStatus);
-        XMLFileReader.read(fileName, reader, -1, true, DEFAULT_DECLHANDLER);
+        XMLFileReader.read(fileName, fis, -1, true, DEFAULT_DECLHANDLER);
```
hagbard, line 333: I thought you were going to add a comment here (if I understand what you said previously) since this is an important change.

srl295:
> > We could. Although we are trying to _discourage_ use of the BOM. I verified that this code is able to read the CLDR v1.7.2 `zh.xml`. Do we actually want that test case?
> 
> If it's not hard to add, I'd add one. Then at some point if/when we turn off BOM support we can see it fail so we know it's done.
> I'll leave it up to you.

The way this is coded, we stop dealing with the character encoding at all, and just feed the file to the XML parser as we should have before. The XML API formerly used broke on a BOM and required special workarounds, which this PR removes. So in summary:

- correct use of the XML parser.  There's no BOM related code, or for that matter UTF-8 specificity, in our parsing pipeline anymore.
- plus, a special linter phase to explicitly warn us against including BOMs - as a matter of preference.

srl295:
It's in the commit message

    - use InputSource / InputStream instead of Reader input where possible
    (the Reader interface with explicit UTF-8 cannot be used with
    XML files containing a BOM.)

>

hagbard:
> It's in the commit message - use InputSource / InputStream instead of Reader input where possible (the Reader interface with explicit UTF-8 cannot be used with XML files containing a BOM.)
> […](#)

Ahh, I meant in the code, next to where the changes are. Since it's a visibly small change and Reader is generally preferred to InputStream in most other contexts, it's not beyond the realm of possibility that someone will accidentally change it back at some point. This is why a test is good too (in fact if a test targeted at checking BOM is supported existed, I'd probably not worry about a comment in the code).

btangmu (approved):

hagbard:
> > > We could. Although we are trying to _discourage_ use of the BOM. I verified that this code is able to read the CLDR v1.7.2 `zh.xml`. Do we actually want that test case?
> > 
> > If it's not hard to add, I'd add one. Then at some point if/when we turn off BOM support we can see it fail so we know it's done.
> > I'll leave it up to you.
> 
> The way this is coded, we stop dealing with the character encoding at all, and just feed the file to the XML parser as we should have before. The XML API formerly used broke on a BOM and required special workarounds, which this PR removes. So in summary:
> 
> * correct use of the XML parser.  There's no BOM related code, or for that matter UTF-8 specificity, in our parsing pipeline anymore.
> * plus, a special linter phase to explicitly warn us against including BOMs - as a matter of preference.

Ahh okay. That makes more sense. Providing regression is somethings we'll spot easily.

macchiati:
Steven, can you merge so that I can get the growth chart running?

hagbard (approved):

PR #798, opened 2020-10-18
CLDR-14128 spec misc fixes: TOC, links, change spans

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14128

Misc spec fixes from review of https://unicode-org.atlassian.net/browse/CLDR-14194: TOC entry update, fix some links and anchors, fix some change spans.


macchiati (approved):
Thanks for doing this. I think the Modifications section needs some work too; can discuss this week.

PR #800, opened 2020-10-19
CLDR-14236 zh, revert to CLDR 37 value for <metazone type="Indochina">

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14236

Per TC discussion: In zh, revert to CLDR 37 value for &lt;metazone type="Indochina"&gt;, i.e. with the region portion of the name being "印度支那" (Indochina) instead of "印度中国" (India-China)

### DIFF
--- a/common/main/zh.xml
+++ b/common/main/zh.xml
@@ -7563,7 +7563,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			</metazone>
 			<metazone type="Indochina">
 				<long>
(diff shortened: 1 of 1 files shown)


FrankYFTang (approved):
This is good. It rollback to the CLDR 37 value and align with many usage in the official China gov medial .people.com.cn website.

macchiati (approved):

markusicu (approved):
lgtm tnx (based on Frank's review)

PR #805, opened 2020-10-21
CLDR-14236 zh_Hant, revert to CLDR 37 value for <metazone type="Indochina">

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14236

Per TC discussion: Just as https://github.com/unicode-org/cldr/pull/800 did for zh, this PR reverts zh_Hant to the CLDR 37 value for `<metazone type="Indochina">`, i.e. with the region portion of the name being "印度支那" (Indochina) instead of "印度中國" (India-China)


btangmu (approved):

markusicu (approved):
rslgtm

PR #807, opened 2020-10-21
CLDR-14241 Forum Participation is breaking

btangmu:
-Configure nginx with same generous timeouts as it had on the old server

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14241 

### DIFF
--- a/tools/scripts/ansible/server-playbook.yml
+++ b/tools/scripts/ansible/server-playbook.yml
@@ -94,7 +94,10 @@
       blockinfile:
         path: /etc/nginx/sites-enabled/default
         block: |
-          # proxy /cldr-apps/ to tomcat
+          # proxy /cldr-apps/ to tomcat, with generous timeouts
+          proxy_connect_timeout 60s;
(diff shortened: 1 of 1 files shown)


macchiati (approved):

macchiati:
a question: is it just the DB lookup that takes so much time?

Mark

srl295:
> a question: is it just the DB lookup that takes so much time? Mark

I think the timeouts are pretty short by default.

srl295 (approved):

btangmu:
> > a question: is it just the DB lookup that takes so much time? Mark
> I think the timeouts are pretty short by default. Steven

Yes, in this case it is just the DB lookup that takes so much time. See SurveyForumParticipation.getForumStats().

And yes, the default timeouts are pretty short, like proxy_read_timeout 60s -- however that would be plenty of time if we properly used a modern framework. The server should always respond promptly to say "I'm working on it"

PR #813, opened 2020-10-23
CLDR-14212 Update GitHub actions to @vx vs @vx.y.z

srl295:
CLDR-14212

- Background: https://github.community/t/version-numbering-for-actions/16307/4
- Specific issue: https://github.com/actions/setup-java/pull/104#issuecomment-715489633
actions/setup-java@v1.4.2 was creating warning messages about set-env,
needed to switch to v1.4.3.  Instead, all actions now use @v1 or @v2, which points to a
branch.

### DIFF
--- a/.github/workflows/commit-checker.yml
+++ b/.github/workflows/commit-checker.yml
@@ -30,14 +30,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: clone CLDR
-        uses: actions/checkout@v2.3.3
+        uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


srl295:
This is the cause of the 'warnings' in the setup-java phases.

echeran (approved):
LGTM.  Thanks for the links to the specific details on how GH Actions versioning is supposed to work.  It makes sense that the onus is on the action maintainer to keep the latest major version alias pointed to the latest release, since each version is really defined just as the name of a git ref (branch/tag) in the action's git repo.

macchiati (approved):

PR #814, opened 2020-10-26
CLDR-14259 Final TR35 for CLDR 38, revision 61

rickmcgowan:
##### Checklist


pedberg-icu (approved):
Compared in browser the versions in https://www.unicode.org/reports/tr35/tr35-60/ and https://www.unicode.org/reports/tr35/tr35-61/

PR #815, opened 2020-10-26
CLDR-11125 Formatting preferences for enum

yumaoka:
I found a combination of formatting settings which seem to prevent unexpected enum comment/format corruption: Never join already wrapped lines, and enum contsnts to wrap if necessary. With this setting, large enum with comments like enum PageId is not affected. This PR also include the result of clean up/organize import.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11125


macchiati:
Looks great. At this point we'll have to wait for v39, but we can roll this in very early, thanks!

One question: does it split enum lines apart, eg fix up the lines in https://unicode-org.atlassian.net/browse/CLDR-11125 under NEW to look like OLD, or does it just keep them from being joined once they are fixed?

yumaoka:
> fix up the lines in https://unicode-org.atlassian.net/browse/CLDR-11125 under NEW to look like OLD, or does it just keep them from being joined once they are fixed?

The current source is OLD. After applying format globally, it was kept unchanged, and obviously this is what we want.
The rule used for formatting enum body is - each enum constant is aligned to column with one indentation. It does not join multiple lines. I think if one line exceeds 160 columns, it will wrap only for the line - but most of case, this rule is not applied, and not bother us even it happens.

macchiati:
perfect
Mark

macchiati (approved):

srl295 (approved):
rubber stamp LGTM

srl295:
Should be merged before #822 is merged.

srl295:
@yumaoka please rebase

srl295:
this could also be re-applied _after_ #822

srl295:
@yumaoka you are cleared for take-off! 🚀 
rebase and let's get this in

yumaoka:
When I created this PR, David had another proposal for formatting. In addition to this, we had a problem with in CI build at that time.  At this point, this PR is already stale, and we need to catch up the latest codes even we want to apply equivalent changes. We're not going to use this PR - rebased/updated, because recreating the equivalent PR would be much easier. Closing.

PR #816, opened 2020-10-27
CLDR-14098 Generate disruptive data churn report (part3)

btangmu:
-Implement support for more paths, including some in supplementalData

-Catch disruptive-whitespace-only changes, even for paths not explicitly high-level

-New class SuspiciousChange

-Clean-up, comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14098

### DIFF
--- a/tools/java/org/unicode/cldr/tool/ChartDelta.java
+++ b/tools/java/org/unicode/cldr/tool/ChartDelta.java
@@ -410,9 +410,6 @@ public class ChartDelta extends Chart {
                         Output<Boolean> hasReformattedValue = new Output<>();
 
                         for (String path : paths) {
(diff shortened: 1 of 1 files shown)


tools/java/org/unicode/cldr/tool/ChartDelta.java
```diff
@@ -1422,6 +1519,80 @@ private static boolean isHighLevelCurrencyName(String path, String locale) {
             return false;
         }
 
+        /**
```
macchiati, line 1522: I think this will be more compact and easier to change if we use a regex lookup.

btangmu: OK, I'll use a regex.

btangmu:
See https://github.com/unicode-org/cldr/pull/817

PR #817, opened 2020-10-27
CLDR-14098 Generate disruptive data churn report (part3)

btangmu:
-Implement support for more paths, including some in supplementalData

-Catch disruptive-whitespace-only changes, even for paths not explicitly high-level

-New class SuspiciousChange

-Clean-up, comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14098


btangmu:
Same as https://github.com/unicode-org/cldr/pull/816
except for using a regex currencyPattern for isHighLevelCurrencyName

srl295 (approved):

PR #819, opened 2020-10-28
CLDR-14260 update scripts to generate production .zipfiles

srl295:
CLDR-14260

Ready to review.

Note that I also sign the files with  `gpg --detach-sign -a ${file}`


srl295:
(churn above is due to rebasing from maint-38 to master)

This commit now represents what was used to make the release 38 files, except for the version 38.0 instead of 39.0.

While I would like to update this again later in v39, it also might be worth committing this so we have a baseline for generating the distribution content in the future.

macchiati (dismissed):

srl295:
had to dismiss review since the CI hadn't run. Actually, it's fine if this sits around, I can commit it after the maven PR if needed

srl295:
kicking this back to draft. will come back to it.

macchiati (dismissed):

srl295:
This is now ready to review.  What it does is make `cldr-code.jar` now contain the right license information, and sets the stage for building zipfiles of tools, etc with correct licenses.

Also note that apache-license, etc are not needed anymore. The licenses are included with jars, but no apache code is in our repo.

srl295:
see also https://github.com/unicode-org/cldr-staging/pull/31

.github/workflows/maven.yml
```diff
@@ -37,11 +37,6 @@ jobs:
         with:
           name: cldr-apps
           path: tools/cldr-apps/target/cldr-apps.war
-      - name: Build single cldr-code.jar
-        run: >
-          mvn -s .github/workflows/mvn-settings.xml -pl cldr-code package assembly:single -DskipTests=true  --file tools/pom.xml
```
srl295, line 42: this is now automatically attached to the `package` phase.

tools/cldr-code/pom.xml
```diff
@@ -104,18 +120,27 @@
...
 						<manifest>
 							<mainClass>${mainClass}</mainClass>
 						</manifest>
 					</archive>
 				</configuration>
+				<executions>
+					<execution>
+						<phase>package</phase>
+						<goals>
+							<goal>single</goal>
```
srl295, line 140: automatically create `cldr-code.jar` when `mvn package` is run.

tools/cldr-code/src/license/THIRD-PARTY.properties
```diff
@@ -0,0 +1,6 @@
+# Generated by org.codehaus.mojo.license.AddThirdPartyMojo
+# Please fill the missing licenses for dependencies :
+# this may need to be updated before building final outputs
+# (for now)
+com.ibm.icu--icu4j-for-cldr--68.1-release-68-1=Unicode-DFS-2016
+com.ibm.icu--utilities-for-cldr--68.1-release-68-1=Unicode-DFS-2016
```
srl295, line 6: ideally this would be updated when we bump the dependencies… the best solution is to improve things on the ICU side, ICU-21425

tools/cldr-code/src/main/assembly/cldr-code.xml
```diff
@@ -0,0 +1,41 @@
...
+      <unpack>true</unpack>
+      <scope>runtime</scope>
+    </dependencySet>
+  </dependencySets>
+  <fileSets>
+    <fileSet>
+        <directory>${project.basedir}/target/generated-sources/license/</directory>
+        <outputDirectory>/</outputDirectory>
+        <includes>
+            <include>THIRD-PARTY.txt</include>
```
srl295, line 22: generated file

pedberg-icu (approved):
I don't know enough to review the specifics, but it all seems reasonable and seems to work :-)

PR #821, opened 2020-10-30
CLDR-14122 Bulk close some open forum posts

btangmu:
-New Bulk Close Posts command in gear menu, for Admin only

-New SurveyBulkClosePosts.java

-New CldrStBulkClosePosts.js, special/bulk_close_posts.js

-Move equalsOrInheritsCurrentValue to XMLSource.java

-New method SurveyForum.closeThreads, use also for PostType.CLOSE

-New TestCldrStBulkClosePosts.js, bulk_close_posts_json.js

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14122

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
@@ -1,0 +1,124 @@
+package org.unicode.cldr.web;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+
+import org.json.JSONException;
+import org.unicode.cldr.util.XMLSource;
+import org.unicode.cldr.web.SurveyAjax.JSONWriter;
+
+public class SurveyBulkClosePosts {
+
+    private SurveyMain sm;
+
+    private boolean execute;
+
+    /**
+     * Construct a SurveyBulkClosePosts
+     *
+     * @param sm the SurveyMain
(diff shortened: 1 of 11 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private SurveyMain sm;
+
+    private boolean execute;
+
+    /**
+     * Construct a SurveyBulkClosePosts
+     *
+     * @param sm the SurveyMain
+     * @param execute if true, actually close the posts; else report count and provide button
```
srl295, line 23: this is so close to being a REST API… in my mind I can almost see the @-annotations on each function!

srl295 (approved):
💯 looks great

srl295:
love the tests…

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
btangmu: There really should be a progress bar, since the db queries could take a minute or more, for thousands of posts. Let's decide on a uniform way to implement progress bars for all potentially long-running server tasks. Cf. https://unicode-org.atlassian.net/browse/CLDR-14251

srl295:
Yes, I would be reluctant to do too much engineering on that now, best served by a new front end UI framework

btangmu:
Vue progress bars: https://morioh.com/p/e0339f1f83f5

PR #822, opened 2020-11-02
CLDR-14212 finalize move to Maven

srl295:
CLDR-14212

- get rid of ant build files, .classpath, .project, .settings
- for Eclipse, all of these are synthesized from maven (and .gitignored)
- CLDR no longer requires git-lfs
- deploy and consolecheck are now a part of the maven build
- cldr-tools (cldr.jar) and cldr-apps (cldr-apps.war) artifacts are generated on each build
    
NOTE: #815 should merge before this because of churn!
NOTE: #819 should merge before this because it will otherwise be broken.


.gitattributes
```diff
@@ -1,5 +1,5 @@
-*.jar filter=lfs diff=lfs merge=lfs -text
-*.dat filter=lfs diff=lfs merge=lfs -text
-*.zip filter=lfs diff=lfs merge=lfs -text
-*.gz filter=lfs diff=lfs merge=lfs -text
-*.bz2 filter=lfs diff=lfs merge=lfs -text
+# *.jar filter=lfs diff=lfs merge=lfs -text
```
srl295, line 1: disabling these for now so that no new files are added to lfs.

srl295:
the 'cldr-mvn/build' phase includes unit tests (for tools and also for cldd-apps). As such it takes a bit too long. Might be worth splitting the phases a bit:
- build
- unittest 
- check
- deploy

Then, 'unittest' and 'check' could be run in parallel. The downside is checking out again, moving artifacts between phases, etc.

srl295:
- [ ] Also, not only the 'required checks' but travis should probably just be deactivated before this merge.

srl295:
@yumaoka Hmm. I had .gitignore'd and removed .classpath/.project/.settings - but the settings were modified in #815 

Is it possible to have .settings even if the projects are synthesized from maven?

srl295:
@yumaoka I've preserved the .settings files, everything else should be synthesized from maven.

tools/cldr-apps/.settings/org.eclipse.wst.common.component
```diff
@@ -1,35 +1,48 @@
...
+        
         
         <wb-resource deploy-path="/WEB-INF/classes" source-path="/.apt_generated"/>
                 
-        <wb-resource deploy-path="/WEB-INF/classes" source-path="/src/test/java"/>
-        <wb-resource deploy-path="/WEB-INF/classes" source-path="/src/test/resources"/>
         <wb-resource deploy-path="/WEB-INF/classes" source-path="/src/main/resources"/>
-                        
+        <wb-resource deploy-path="/WEB-INF/classes" source-path="/.apt_generated_tests"/>
+        <dependent-module archiveName="cldr-39.0-SNAPSHOT.jar" deploy-path="/WEB-INF/lib" handle="module:/resource/cldr/cldr">
```
btangmu, line 30: How will this dependency on the version -- here 39.0 -- be handled? Is this set within the Eclipse interface, or ...?

tools/cldr-apps/pom.xml
```diff
@@ -18,7 +18,7 @@
 	<parent>
 		<groupId>org.unicode.cldr</groupId>
 		<artifactId>cldr-all</artifactId>
-		<version>38.0-SNAPSHOT</version>
+		<version>39.0-SNAPSHOT</version>
```
btangmu, line 21: How does the versioning work?

tools/java/.settings/org.eclipse.wst.common.component
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?><project-modules id="moduleCoreId" project-version="1.5.0">
```
btangmu, line 1: What is this about: project-version="1.5.0"?

.gitattributes
btangmu: "for now" -- and how will it be in the long term?

.github/workflows/maven.yml
```diff
@@ -51,4 +64,50 @@ jobs:
...
+    needs:
+      - build
+      - check
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download cldr-apps.war
+      uses: actions/download-artifact@v2
+      with:
+        name: cldr-apps
+    - name: Deploy to cldr-smoke (experimental)
```
btangmu, line 96: How is this triggered and how does it relate to other ways of deployment?

btangmu (commented):
This looks great, it's good to see the reduction in code size. I just added a few questions, trying to understand how it works...

.gitattributes
btangmu: If we no longer need lfs, how about deleting rather than commenting out?

btangmu (approved):

srl295:
- [ ] push to production needs to move from ant to maven

tools/cldr-apps/.settings/org.eclipse.wst.common.component
```diff
@@ -1,35 +1,48 @@
 <?xml version="1.0" encoding="UTF-8"?><project-modules id="moduleCoreId" project-version="1.5.0">
-            
+
```
yumaoka, line 2: Eclipse insert extra blank lines when a configuration file is updated. This is probably a bug in Eclipse. You may get rid of these unnecessary blank lines inserted automatically.

tools/cldr-apps/pom.xml
yumaoka: Use of maven release plugin to drop snapshot, tag and bump a version is a popular way to do this.

Alternatively, we may have parent pom to manage single version across modules.

tools/java/.settings/org.eclipse.wst.common.component
yumaoka: It's managed by Eclipse. This is automatically generated file, so I think we don't need to worry about version much.

yumaoka (dismissed):
Looks good to me.

One comment - do we want to use "java" as component folder name? I prefer to use folder name matching artifact ID.

.gitattributes
srl295: sure, I will just delete it.

tools/cldr-apps/.settings/org.eclipse.wst.common.component
srl295: this should get automatically updated by eclipse.

tools/cldr-apps/pom.xml
srl295: you can't have a single version in the parent AND use the version for submodules to find the parent. 
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] 'parent.version' is missing. @ line 22, column 10
[FATAL] 'parent.version' is missing. @ line 18, column 10
```

srl295: I think we can just bump the version whenever a new release is cut .  actually, when the maint branch is cut it can be manually updated.

.github/workflows/maven.yml
srl295: This is the deploy to smoketest on merge that currently is done with ant.  It's triggered the same way.

tools/java/.settings/org.eclipse.wst.common.component
srl295: > Looks good to me.
> 
> One comment - do we want to use "java" as component folder name? I prefer to use folder name matching artifact ID.

Yes, I'd like to rename it, maybe `cldr-utils`? Not sure.

tools/cldr-apps/.settings/org.eclipse.wst.common.component
srl295: yeah. I was going to ignore these files originally, but I think we should just let them get updated…

srl295:
@yumaoka I sent an email to cldr core proposing the name `cldr-utils` for the jar and the directory (replacing the nondescript `java`).   I might make that change in a separate PR.

.gitattributes
srl295: > [x] push to production needs to move from ant to maven

done

srl295:
@yumaoka @btangmu @macchiati … ready to merge this.  We can rename `java` in a separate PR.

srl295:
I've updated http://cldr.unicode.org/development/maven with instructions in using maven from the command line.

macchiati:
Thanks!

Mark

srl295:
Renaming java  could be a separate PR. If this is acceptable I’d be
inclined to merge it to shake out any issues; and renaming could be a
separate PR right afterwards if we want to wait until we have a TC meeting.
.

macchiati:
Sounds good.

btangmu (approved):

srl295:
I'm going to merge this then!

PR #829, opened 2020-11-05
CLDR-8534 support authenticated mail

srl295:
- drop the CLDR_SMTP parameter, vector to existing Java params

To enable mail:
- CLDR_SENDMAIL=true
- mail.host (set to your host, or leave blank for 'pretend' mode)
- set CLDR_SMTP_USER / CLDR_SMTP_PASSWORD for authentication
- All other parameters are documented at:
<https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/package-summary.html>

also fixes CLDR-13156

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
@@ -25,7 +25,9 @@ import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
+import javax.mail.Authenticator;
 import javax.mail.MessagingException;
+import javax.mail.PasswordAuthentication;
 import javax.mail.Session;
 import javax.mail.Transport;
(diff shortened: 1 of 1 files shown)


srl295:
fyi @rickmcgowan

tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
```diff
@@ -41,7 +43,7 @@
  */
 public class MailSender implements Runnable {
 
-    private long waitTill = 0;
+    // private long waitTill = 0;
```
btangmu, line 46: minor suggestion for next time: delete instead of commenting out

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
srl295: the implementation was already commented out, this just removes the warning

PR #820, opened 2020-10-30
CLDR-13730 json: support unit preferences

srl295:
CLDR-13730

- Also, move path transforms out to a separate file
- Other JSON code improvements

Example files:

### `unitPreferenceData.json`

```json
{
  "supplemental": {
    "version": {
      "_unicodeVersion": "13.0.0",
      "_cldrVersion": "38"
    },
    "unitPreferenceData": {
      "area": {
        "default": {
          "001": [
            {
              "unit": "square-kilometer"
            },
            {
              "unit": "hectare"
            },
            {
              "unit": "square-meter"
            },
            {
              "unit": "square-centimeter"
            }
          ],
…
        "person-height": {
          "001": [
            {
              "unit": "centimeter"
            }
          ],
…
          "CA": [
            {
              "unit": "foot-and-inch",
              "geq": 3.0
            },
            {
              "unit": "inch"
            }
          ],
}
```

- [ ] OK, grammatical features are still missing, will recheck

PR #830, opened 2020-11-05
CLDR-13730 add grammaticalFeatures

srl295:
CLDR-13730

- also general cleanup
- moved grammaticalGender to separate file
- corrected duplicate locale fil in common/supplemental/grammaticalFeatures.xml

### DIFF
--- a/common/supplemental/grammaticalFeatures.xml
+++ b/common/supplemental/grammaticalFeatures.xml
@@ -33,7 +33,7 @@ plurals.xml and ordinals.xml. See those files and the LDML spec for more informa
 <supplementalData>
   <version number="$Revision: 1 $"/>
   <grammaticalData>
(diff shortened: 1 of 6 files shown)


srl295:
Want to review the JSON format? See https://gist.github.com/srl295/2c521b0cc01d904beea439df93b51454

srl295:
Notes:

1. I split out `grammaticalGenderFeatures.json` for some data, because the current converter is a stream based approach and isn't easily able to backup and interleave data without special purpose code (such as was done for unit preferences  #820 ). This isn't ideal, and we may want to restructure the converter to make it easier to produce better output.

2. there was a case where the 'no grammatical features' entry had `fil` listed twice. I have the converter ignore the second `fil`, but when there are unit tests for this feature, it would be good to check for this situation.

sffc (approved):
Looks plausible.

PR #835, opened 2020-11-06
CLDR-14285 cause TestFmwk.sourceLocation() to handle lambdas

srl295:
CLDR-14285

The search code for the relevant test source file was wrong.
This fixes it to skip 'lambda$' if present, otherwise the
error source is obscured.


srl295:
I wasn't able to fix this, but it's broken on jdk 10/11 and works on 12+

I improved the err message.

macchiati (approved):

PR #833, opened 2020-11-06
CLDR-13139 Upgrade Tomcat, more preliminary refactoring

btangmu:
-Move parts of server-playbook to new playbooks for mysql, nginx, tomcat, cldr-apps

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13139
- [x] Updated PR title and link in previous line to include Issue number

PR #839, opened 2020-11-11
CLDR-13139 Upgrade Tomcat, more preliminary refactoring

btangmu:
-Move parts of server-playbook to new playbooks for mysql, nginx, tomcat, cldr-apps

##### Checklist


btangmu:
This is like https://github.com/unicode-org/cldr/pull/833 except for moving "Setup index.html" to the nginx playbook so that /var/www already exists

srl295 (approved):

PR #840, opened 2020-11-11
CLDR-14284 TestAttributeValues check for duplicates in space-separated attributes

srl295:
CLDR-14284 

- also removed a comment on a previously-duplicate grammaticalFeatures element
that was fixed with CLDR-13730

##### Checklist

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
@@ -4,6 +4,7 @@ import java.text.ParseException;
 import java.util.Date;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
@@ -153,6 +154,13 @@ public abstract class MatchValue implements Predicate<String> {
         "Loma", "Maya", "Moon", "Nkgb", "Phlv", "Roro", "Sara", "Syre", "Syrj", "Syrn", "Teng", "Visp", "Wole");
     static final Set<String> VARIANT_HACK = ImmutableSet.of("POSIX", "REVISED", "SAAHO");
 
+    /**
+     * Returns true if ALL items match the predicate
+     * @param <T>
+     * @param predicate predicate to check
+     * @param items items to be tested with the predicate
+     * @return
+     */
     public static <T> boolean and(Predicate<T> predicate, Iterable<T> items) {
         for (T item : items) {
             if (!predicate.is(item)) {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAttributeValues.java
```diff
@@ -261,6 +262,14 @@ public void checkElement(String element, Attributes atts) {
         }
 
         private void checkAttribute(String element, String attribute, String attrValue) {
+            if(LdmlConvertRules.isSplittableAttr(element, attribute)) {
+                final Set<String> asSet = new HashSet<>();
+                SPACE_SPLITTER.splitToList(attrValue).forEach(v -> {
+                    if(asSet.add(v) == false) {
+                        throw new IllegalArgumentException("in <" + element+ " " + attribute + "=…, value " + v + " appears twice.");
```
srl295, line 269: This does print an error and stop the test. There may be a more graceful way to fail.

macchiati (changes requested):
This wasn't using what I was thinking of. Will follow up with details

srl295:
@macchiati let me know if this looks correct

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -616,7 +631,14 @@ public static SetMatchValue of(String key) {
 
         @Override
         public  boolean is(String items) {
-            return and(subtest,SPACE_SPLITTER.split(items));
+            Iterable<String> splitItems = SPACE_SPLITTER.split(items);
```
macchiati, line 634: Looks good. Should add quick unit test.
Also, can be pithier. Split to list, then create set from it. If the sizes are not the same, fail.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
srl295: > Should add quick unit test.

TestAttributeValues already hits this, and is verified to fail on the `"… fil fil …"` case for grammatical features.

> Also, can be pithier… If the sizes are not the same, fail

good point, will fix

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -640,7 +659,6 @@ public String getName() {
         }
 
         public static OrMatchValue of(String key) {
-            IntFunction<MatchValue[]> generator = null;
```
srl295, line 643: just a line of dead code, no side effect

macchiati (approved):

PR #843, opened 2020-11-15
CLDR-14289 clean up coverage levels

macchiati:
##### Checklist


pedberg-icu:
Can we close this? Seems like it has been replaced by other PRs for CLDR-14289 (*for CLDR 41) and its follow-on clone (for CLDR 42)

PR #846, opened 2020-11-17

btangmu:

##### Checklist


btangmu:
See https://github.com/unicode-org/cldr/pull/847

PR #847, opened 2020-11-18

btangmu:

##### Checklist

### DIFF
--- a/tools/scripts/ansible/requirements.yml
+++ b/tools/scripts/ansible/requirements.yml
@@ -1,3 +1,13 @@
-sprylab.tomcat-ubuntu
-geerlingguy.mysql
-derjd.journald
+- src: geerlingguy.mysql
+  version: 3.3.0
+- src: geerlingguy.nginx
+  version: 2.8.0
+- src: derjd.journald
+  version: 0.0.1
+- src: robertdebock.tomcat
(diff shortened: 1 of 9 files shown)


tools/scripts/ansible/requirements.yml
```diff
@@ -1,3 +1,13 @@
...
+- src: geerlingguy.mysql
+  version: 3.3.0
+- src: geerlingguy.nginx
+  version: 2.8.0
+- src: derjd.journald
+  version: 0.0.1
+- src: robertdebock.tomcat
+  version: 4.6.2
+# The following are also required for robertdebock.tomcat
+- src: robertdebock.bootstrap
```
btangmu, line 10: can we avoid adding these manually, or test whether they're actually used/needed as dependencies?

srl295: @btangmu is it a bug in robertdebock.tomcat? maybe file a bug upstream then

srl295 (approved):
lgtm so far

tools/scripts/ansible/requirements.yml
btangmu: probably not a bug just my lack of understanding

btangmu:
See https://github.com/unicode-org/cldr/pull/854

PR #852, opened 2020-11-23
CLDR-13328 remove last bits of svn mention, cleanup urls

srl295:
CLDR-13328
- common constant for repo root
- remove unneeded libs for Subversion
- also remove 'rome' which is an RSS library, not used.

### DIFF
--- a/tools/cldr-apps/src/main/webapp/cldr-setup.jsp
+++ b/tools/cldr-apps/src/main/webapp/cldr-setup.jsp
@@ -2,8 +2,6 @@
 <%@page import="java.sql.SQLException"%>
 <%@page import="org.unicode.cldr.util.VoteResolver.VoterInfo"%>
 <%@page import="com.ibm.icu.util.VersionInfo"%>
-<%@page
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/src/main/webapp/cldr-setup.jsp
```diff
@@ -2,8 +2,6 @@
 <%@page import="java.sql.SQLException"%>
 <%@page import="org.unicode.cldr.util.VoteResolver.VoterInfo"%>
 <%@page import="com.ibm.icu.util.VersionInfo"%>
-<%@page
-	import="org.tmatesoft.sqljet.core.internal.lang.SqlParser.commit_stmt_return"%>
```
btangmu, line 6: How does SqlParser relate to this ticket?

srl295: - the import should not have been in this jsp 
- this PR removes the tmatesoft jars including the one imported here.

macchiati (approved):

PR #854, opened 2020-11-24

btangmu:

-Update README.md: requirements.yml not roles.yml; mvn not ant

-For vagrant only, new vagrant-playbook.yml and self-sign-playbook.yml

-Revise context.j2: com.mysql.cj.jdbc.Driver not com.mysql.jdbc.Driver; cachingAllowed; cacheMaxSize

-Make cldr_tomcat_conf_path end in /conf

##### Checklist


tools/scripts/ansible/cldr-apps-playbook.yml
```diff
@@ -38,6 +38,8 @@
         mode: "0644"
       notify: Restart Tomcat
     - name: Checkout CLDR trunk
+      become: yes
+      become_user: surveytool
```
btangmu, line 42: This is working -- we get an incomplete git repository, and owner is tomcat -- I suspect that instead of become_user, we could wait until after checkout, then use file module to change ownership of CLDR trunk to surveytool

tools/scripts/ansible/Vagrantfile
```diff
@@ -2,6 +2,10 @@
 # vi: set ft=ruby :
 
 Vagrant.configure("2") do |config|
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 1024
+    # v.cpus = 2
```
btangmu, line 7: I got an out-of-memory error unless set v.memory here -- but not sure we want to change v.cpus

btangmu:
See https://github.com/unicode-org/cldr/pull/856

PR #856, opened 2020-11-25

btangmu:

-Make cldr_tomcat_conf_path end in /conf

-Use only /etc/tomcat, never /var/lib/tomcat (which no longer exists)

-Update README.md: requirements.yml not roles.yml; mvn not ant

-Revise context.j2: com.mysql.cj.jdbc.Driver not com.mysql.jdbc.Driver; cachingAllowed; cacheMaxSize

-For vagrant only, new vagrant-playbook.yml; just calls setup-playbook.yml but adds flexibility

-Remove problematic AJP section from server.xml

-In deploy-sh.j2 use only spaces for indentation, rather than mix tabs and spaces

##### Checklist

### DIFF
--- a/tools/scripts/ansible/cldr-apps-playbook.yml
+++ b/tools/scripts/ansible/cldr-apps-playbook.yml
@@ -47,6 +47,13 @@
         # this is deep because we will need to keep updating
         # it with history. It does not include LFS as that
         # is not needed for the surveytool.
+    - name: Change ownership of CLDR trunk to surveytool
+      file:
+        path: "{{ cldr_trunk_path }}"
+        owner: surveytool
+        group: root
+        state: directory
+        recurse: yes
   handlers:
     - name: Restart Tomcat
       service:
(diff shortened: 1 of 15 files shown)


btangmu:
I think this may be good the way it is, I just want to test further, so "draft" for now

tools/scripts/ansible/cldr-apps-playbook.yml
```diff
@@ -47,6 +47,13 @@
         # this is deep because we will need to keep updating
         # it with history. It does not include LFS as that
         # is not needed for the surveytool.
+    - name: Change ownership of CLDR trunk to surveytool
+      file:
+        path: "{{ cldr_trunk_path }}"
+        owner: surveytool
+        group: root
+        state: directory
+        recurse: yes
```
btangmu, line 56: ansible-lint would like "mode: ..." added here -- but that's complicated with "recurse: yes"

btangmu: Actually we might skip the "Change ownership" task and instead add "become_user: surveytool" to the previous task -- I thought it wasn't working when I tried it before, but that may have been due to confusion between /etc/tomcat and /var/lib/tomcat

btangmu: Yes, that works. I'll make a new PR.

btangmu:
See https://github.com/unicode-org/cldr/pull/857

PR #859, opened 2020-11-25
CLDR-8802 revamp abstract cache, update UI

srl295:
CLDR-8802

- no data is stored in the repository anymore
- xpath-to-resource data is cached, will be fetched by ST
CLDR_ABSTRACT_EXPIRE_DAYS defaults to 15
stored in 'abstracts' directory parallel to cldr.properties
- abstract content is fetched by client browser directly


srl295:
This revamps the back end. Now, all the UI requests (including wikipedia link) can be done on the front end. I will update the PR with FE updates, but wanted to get this in DRAFT so that the backend code can be inspected

srl295:
Ready to go. See updated UI here:
- using `overflow: hidden` instead of … - so the browser automatically does the clipping.
- "more" button takes you to Wikipedia

image

macchiati (approved):

PR #860, opened 2020-11-27
CLDR-14212 publish to maven on release

srl295:
CLDR-14212

publish ( using version `0.0.0-SNAPSHOT-`_hash_ where _hash_ is the git hash ) all released versions to maven

### DIFF
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -1,0 +1,41 @@
+name: cldr-mvn-deploy
+
+on:
+  release:
+    types: [created]
+jobs:
+  mvn-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          lfs: false
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('tools/**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Build with Maven
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B compile install package --file tools/pom.xml
+          -DskipTests=true
+        env:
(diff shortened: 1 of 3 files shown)


srl295:
This is to support UnicodeTools and other consumers of CLDR via maven.

So this works this way:
- create a release (could be an official `release-38` or an unofficial, prerelease `snapshot-2020-11-27` (the exact tag name is not used currently).
- this creates maven artifacts under https://github.com/orgs/unicode-org/packages?repo_name=cldr  such as:

```xml
<dependency>
  <groupId>org.unicode.cldr</groupId>
  <artifactId>cldr-code</artifactId>
  <version>0.0.0-SNAPSHOT-51d91227f4</version>
</dependency>
```

- as noted, the tag name is not used, but the act of creating a release (or prerelease) causes the artifact with the specific hash to be uploaded to github maven.

btangmu:
> So this works this way:
> 
>     * create a release (could be an official `release-38` or an unofficial, prerelease `snapshot-2020-11-27` (the exact tag name is not used currently).

How does one actually create a release? Just by pressing a button in the Actions tab, or are other steps needed?

btangmu (approved):

srl295:
> > So this works this way:
> > ```
> > * create a release (could be an official `release-38` or an unofficial, prerelease `snapshot-2020-11-27` (the exact tag name is not used currently).
> > ```
> 
> How does one actually create a release? Just by pressing a button in the Actions tab, or are other steps needed?

under 'releases' then 'new':  https://github.com/unicode-org/cldr/releases/new

srl295:
> > > So this works this way:
> > > ```
> > > * create a release (could be an official `release-38` or an unofficial, prerelease `snapshot-2020-11-27` (the exact tag name is not used currently).
> > > ```
> > 
> > 
> > How does one actually create a release? Just by pressing a button in the Actions tab, or are other steps needed?
> 
> under 'releases' then 'new': https://github.com/unicode-org/cldr/releases/new

all releases create a tag, not all tags are releases.  I would add that the release may be a 'prerelease' which would be correct for a release that isn't a real production release (38, 38.1, 39… )

srl295:
@hagbard you might be interested in this from the tooling perspective.

btangmu:
> > under 'releases' then 'new': https://github.com/unicode-org/cldr/releases/new

I see! And "Releases" is part of the right sidebar at https://github.com/unicode-org/cldr
Thanks!

PR #861, opened 2020-11-30
CLDR-11771 Add unit test for VettingViewer.getMissingStatus

btangmu:
-New method TestUtilities.TestMissingStatus

-Delete some dead code

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11771

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUtilities.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUtilities.java
@@ -44,10 +44,11 @@ import org.unicode.cldr.util.SpecialLocales;
 import org.unicode.cldr.util.StringId;
 import org.unicode.cldr.util.SupplementalDataInfo;
 import org.unicode.cldr.util.SupplementalDataInfo.PluralInfo.Count;
+import org.unicode.cldr.util.VettingViewer;
 import org.unicode.cldr.util.VettingViewer.Choice;
+import org.unicode.cldr.util.VettingViewer.MissingStatus;
 import org.unicode.cldr.util.VettingViewer.VoteStatus;
 import org.unicode.cldr.util.VoteResolver;
-import org.unicode.cldr.util.VoteResolver.CandidateInfo;
 import org.unicode.cldr.util.VoteResolver.Level;
 import org.unicode.cldr.util.VoteResolver.Status;
 import org.unicode.cldr.util.VoteResolver.VoterInfo;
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUtilities.java
```diff
@@ -1409,4 +1372,20 @@ public void TestBulkUploadHtml() {
         assertEquals("writeBulkInfoHtml", expected, out.toString());
     }
 
+    /**
+     * Verify that VettingViewer.getMissingStatus returns MissingStatus.PRESENT
+     * for a typical path in a well-populated locale
+     *
+     * Ideally we should also test for MissingStatus.DISPUTED, etc.; that's more difficult
+     */
```
btangmu, line 1380: How could we test for MissingStatus.DISPUTED? It would require the kind of CLDRFile that reflects voting in progress; do we have a way to set that up in the context of cldr-unittest TestAll?

macchiati: We'd have to set it up so that the callback returned the interesting information, I'm guessing.

srl295: One way would be to leverage the `TestSTFactory` which already sets up vetters, disputes, etc. It actually creates a STFactory though which has a DB backing.

But all you need is a `UsersChoice<Organization>` subclass - `VettingViewer.main()` constructs one which includes disputed values, which looks quite suitable for a test case.

srl295: actually, `getMissingStatus()` will never return `disputed`, but member function `getErrorOnPath()` will.

srl295 (approved):
LGTM, +1 for more tests!

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUtilities.java
btangmu: > actually, `getMissingStatus()` will never return `disputed`, but member function `getErrorOnPath()` will.

Good point! The ancient bug, for which this is a follow-up, was https://unicode-org.atlassian.net/browse/CLDR-11765 and it involved items under Disputed in the Dashboard, hence my confusion...

I think this is the first unit test for any method in VettingViewer. Obviously more tests would be helpful.

I doubt whether getMissingStatus even belongs in VettingViewer -- shouldn't the status be the same regardless of whether an item is viewed in the Dashboard or the DataSection view?

srl295: @btangmu it seems to be more of a utility function that may belong somewhere else.

PR #865, opened 2020-12-01
CLDR-14331 Update pcm-t-k0-windows.xml

xadxura:
Update to add dead key transforms for pcm per:
https://unicode-org.atlassian.net/browse/CLDR-14331

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14331

### DIFF
--- a/keyboards/und/pcm-t-k0-windows.xml
+++ b/keyboards/und/pcm-t-k0-windows.xml
@@ -9,7 +9,7 @@
 	</names>
 	<settings fallback="omit" transformPartial="hide"/>
 	<keyMap>
-		<map iso="E00" to="̀"/>
+		<map iso="E00" to="\u{300}"/>
 		<map iso="E01" to="1"/>
 		<map iso="E02" to="2"/>
 		<map iso="E03" to="3"/>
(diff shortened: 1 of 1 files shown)


macchiati:
FYI, in the header you should change
Issue filed: https://unicode-org.atlassian.net/browse/CLDR-_____
to 
Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14331

macchiati (approved):

sven-oly:
We could add dead key support for the grave accent at E00, similar to the acute dead key transforms here. However, this layout has the _combining grave_ on E00, which I thought could just be typed after the base vowel.

If we change to use a dead key for the grave that precedes vowels, then E00 should probably be changed to the non-combining Unicode Character 'GRAVE ACCENT' (U+0060). That would make it consistent with the use of the apostrophe as a dead key.

Since we have heard that the down tone is optional, [CLDR-14250](https://unicode-org.atlassian.net/browse/CLDR-14250?focusedCommentId=158991), using both as dead keys may be a better approach.

srl295 (dismissed):

sven-oly:
I am unclear as to why we want to use the dead key method when the combining grave and combining acute are on E00 and D01 already, supporting vowel + combining mark.

xadxura:
> I am unclear as to why we want to use the dead key method when the combining grave and combining acute are on E00 and D01 already, supporting vowel + combining mark.

I had missed the 0300 and 0301 assignments in on E00 and D01 since they merged visually with dbl quotes and didn't show up when searching. I've switched these to Unicode escape characters to make it clearer, especially since the escape values are used in the transform section.

I agree now with your comment questioning the utility of the dead key assignments. I've removed them.

sven-oly (approved):
Looks good to me.

srl295:
Closing PR as author has closed the ticket.

PR #868, opened 2020-12-01
CLDR-13498 Language name Root should not be open for data contribution

btangmu:
-Add a HIDE line for root in PathHeader.txt

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13498


macchiati (changes requested):
This is the wrong fix. We should instead not be adding root into the set in the first place. See https://unicode-org.atlassian.net/browse/CLDR-13498?focusedCommentId=156698

btangmu:
See https://github.com/unicode-org/cldr/pull/871

PR #869, opened 2020-12-04
CLDR-8802 add test for RDF, and add Scripts support

srl295:
- documented the process here: http://cldr.unicode.org/development/updating-abstracts


macchiati (approved):

srl295:
Note until CLDR-14323 is fixed we may need to manually delete the `…/abstracts/xpath-to-resource.properties` in order for this to take effect.

PR #871, opened 2020-12-07
CLDR-13498 Language name Root should not be open for data contribution

btangmu:
-Skip ROOT_ID for CLDRFile.LANGUAGE_NAME in the loop that iterates over sc.getSurveyToolDisplayCodes

-Remove dead code, revise outdated comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13498

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
@@ -291,7 +291,6 @@ public abstract class XMLSource implements Freezable<XMLSource>, Iterable<String
             return oldPath + "/" + relativePath.replace('\'', '"');
         }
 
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
```diff
@@ -1452,24 +1376,15 @@ public String getLocaleID() {
                             if (s != null && s.size() == 1) continue;
                         }
                         value = TimezoneFormatter.getFallbackName(value);
+                    } else if (typeNo == CLDRFile.LANGUAGE_NAME) {
+                        if (ROOT_ID.equals(value)) {
+                            continue;
+                        }
```
btangmu, line 1382: This is what fixes the bug; other changes are just clean-up

macchiati (approved):

PR #875, opened 2020-12-09
CLDR-14279 es_419 plural time formatting

stenshamn:
Removing the ungrammatical time formatting so it can fall back on parent locale es.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14279


pedberg-icu (approved):
Yes, this looks good. In 'es_419' the generic calendar data also has date-time combining formats but those already match what is in 'es'. And the medium and short forms in gregorian here either match 'es' or inherit from it.

PR #879, opened 2020-12-11
Google Java formatter, AOSP mode, width=120

hagbard:
Reformatting of CLDR code to show formatter differences. 100% automated.


tools/cldr-code/src/main/java/org/unicode/cldr/util/ApproximateWidth.java
```diff
@@ -25,22 +23,22 @@ public static int getWidth(CharSequence s) {
...
-
-            @Override
-            protected boolean handleLine(int lineCount, int start, int end, String[] items) {
-                data.putAll(start, end, Integer.parseInt(items[1]));
-                return true;
-            }
-
-        };
+        SemiFileReader MyFileHander =
+                new SemiFileReader() {
```
hagbard, line 27: Extra indentation is because this is a wrapped line, and to distinguish wrapped lines from following statements, they get 2 x indentation.

See also:
https://docs.google.com/document/d/1JH4KH3jNMFhEYFb8G85adOMgn98iJtSHqs6jNt21-rs/edit#heading=h.9bojtti372ih

This is going to be by far the biggest change in lines of code if we switch to the Google formatter.

tools/cldr-code/src/main/java/org/unicode/cldr/util/BNF.java
```diff
@@ -82,8 +81,7 @@ public BNF complete() {
             if (msg.length() != 0) msg = "Error: Missing definitions for: " + msg;
             String temp = showDiff(ruleSet, variables);
             if (temp.length() != 0) temp = "Warning: Defined but not used: " + temp;
```
hagbard, line 83: This is the kind of code inconsistency that bugs me in the current code base. With a mix of braced and un-braced blocks it would be so easy to end up with misleading code.

If/when we decide on a formatter and are willing to reformat everything, cleaning some of this stuff up becomes much easier:
1: adopt a style
2: run a tool to apply it everywhere
3: reformat with the standard formatter

This decouples tools which structurally modify code from "simple" formatting (which if we trust the formatter) we can just unconditionally run as a final step in any refactoring. Otherwise we might end up deciding to not use a refactoring tool because we don't like how it formats its output.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Builder.java
```diff
@@ -15,21 +15,23 @@
...
- * You start with Builder.with(...) and end with either .get() or .freeze(). With .freeze, the result is unmodifiable
- * (but its objects may be). Examples:
+ * Convenience class for building collections and maps. Allows them to be built by chaining, making
+ * it simpler to set as parameters and fields. Also supplies some operations that are missing on the
+ * JDK maps and collections, and provides finer control for what happens with equal elements.
+ *
+ * <p>You start with Builder.with(...) and end with either .get() or .freeze(). With .freeze, the
+ * result is unmodifiable (but its objects may be). Examples:
+ *
  * <ul>
```
hagbard, line 25: This JavaDoc is invalid (bare '<' , '>' but the Google formatter formats it okay). Not worse than what's there, and the HTML structure is a bit better.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRPaths.java
```diff
@@ -111,25 +128,36 @@
...
-     *             If it must be used, add a comment next to the usage to explain why it is needed.
+     *     If it must be used, add a comment next to the usage to explain why it is needed.
      */
-    @Deprecated
-    public static final String UTIL_DATA_DIR = FileReaders.getRelativeFileName(CldrUtility.class, "data/");
+    @Deprecated public static final String UTIL_DATA_DIR = FileReaders.getRelativeFileName(CldrUtility.class, "data/");
 
     public enum DIRECTORIES {
-        common_dtd, common_properties, common_uca,
+        common_dtd,
```
hagbard, line 136: Seems like a big win for enum formatting, though this case is clearly busted in the current code.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRTransforms.java
```diff
@@ -61,7 +59,7 @@ public CLDRTransforms setShowProgress(Appendable showProgress) {
     final Set<String> overridden = new HashSet<>();
     final DependencyOrder dependencyOrder = new DependencyOrder();
 
-    static public class RegexFindFilenameFilter implements FilenameFilter {
+    public static class RegexFindFilenameFilter implements FilenameFilter {
```
hagbard, line 62: The Google formatter does do consistent reordering of modifiers - I'd forgotten that.
This does mean IDE settings for editing would want to be set to match (e.g. for code templates etc.)

tools/cldr-code/src/main/java/org/unicode/cldr/util/AttributeValueValidity.java
```diff
@@ -773,45 +808,55 @@ public int hashCode() {
...
-                + "</attributeValues>";
+                    + " dtds='"
+                    + type
+                    + "\'"
+                    + " elements='"
+                    + element
+                    + "\'"
+                    + " attributes='"
+                    + attribute
+                    + "\'"
```
macchiati, line 838: Another example where the human-broken line is semantically more readable.

tools/cldr-code/src/main/java/org/unicode/cldr/util/BNF.java
macchiati: This is odd. I think there is an eclipse setting we weere using to add missing brackets, so I'm surprised to see this.

PR #882, opened 2020-12-11
Eclipse formatter, current settings, reimported XML, rewrapping lines

hagbard:
Reformatting of CLDR code to show formatter differences. 100% automated.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
@@ -90,6 +90,7 @@ public class StandardCodes {
     private static final class StandardCodesHelper {
         static final StandardCodes SINGLETON = new StandardCodes();
     }
+
     /**
      * Get the singleton copy of the standard codes.
      */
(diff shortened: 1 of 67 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -1025,19 +1026,10 @@ private void add(CodeType type, String code, List<String> otherData) {
...
 
     public enum LstrType {
-        language("und", "zxx", "mul", "mis", "root"),
-        script("Zzzz", "Zsym", "Zxxx", "Zmth"),
-        region("ZZ"),
-        variant(),
-        extlang(true, false),
-        legacy(true, false),
-        redundant(true, false),
+        language("und", "zxx", "mul", "mis", "root"), script("Zzzz", "Zsym", "Zxxx", "Zmth"), region("ZZ"), variant(), extlang(true, false), legacy(true,
```
macchiati, line 1029: This is exactly what I **don't** want to happen.

ORIGINAL — far easier to understand
        language("und", "zxx", "mul", "mis", "root"),
        script("Zzzz", "Zsym", "Zxxx", "Zmth"),
        region("ZZ"),
        variant(),
        extlang(true, false),
        legacy(true, false),
        redundant(true, false),

NEW — a mess
        language("und", "zxx", "mul", "mis", "root"), script("Zzzz", "Zsym", "Zxxx", "Zmth"), region("ZZ"), variant(), extlang(true, false), legacy(true,
            false), redundant(true, false),

hagbard: The trick now is to compare with the same file in the other pull requests/branches.

This is also a case where a change to the existing settings might well help (in fact it's one of the cases I brought out in the doc).

hagbard: I just pushed:
https://github.com/hagbard/cldr/pull/new/ecl_candidate_changes
which has tweaks from the doc to the Eclipse formatter, was hoped to address some of these changes.

However, while when I cut & paste that exact enum into the formatter settings test page, it works fine, it isn't formatted correctly in the file. As far as I can tell the formatter is simply failing.

This might be because I have issues downloading the dependencies and can't get the code to compile, but formatters shouldn't care about whether dependencies exist.

This is definitely a good case to play with, and it's odd/frightening that the same code formats fine in the settings test page, but fails in real life.

PR #888, opened 2020-12-16
CLDR-14258 JSON build improvements: add cldr-segments-full

srl295:
CLDR-14258 JSON build improvements: add cldr-segments-full

- add a -V option to specify the version, overriding the CLDR version
- restructuring for more clarity on the package calculation process
- reduce the default text output verbosity
- cause the "other" package to not overwrite if multiple "other" sections.
- make sure all tiered (full/modern) modules have files present in full if they are in modern
(only affected module: segments)
- change output format to spend less time in synchronized block:
One output per input file with a summary.
Percentage progress inspired by ninja-build.

Also changes from:

CLDR-14313 compare JSON config to DTD Data

- JSON config was not easily derivable from DTD, however, this commit now
runs a test comparing especially "@MATCH:set" annotations against items treated as
sets in JSON (LdmlConvertRulesTest#testSplittableAttributes) with some exceptions.

JSON output changes caused by this PR
- These items are now arrays instead of strings:
  - supplemental/calendarPreferenceData
  - supplemental/weekData (now grouped by territory)
- Previously MISSING data now added
  - cldr-core/dayPeriods.json:    dayPeriodRuleSet-type-selection"
  - cldr-core/units.json:         unitConstants, unitQuantities
  - cldr-core/unitsMetadata.json: unit deprecations

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/CldrItem.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/CldrItem.java
@@ -241,7 +241,7 @@ public class CldrItem implements Comparable<CldrItem> {
         XPathParts untransformedxpp = XPathParts.getFrozenInstance(untransformedPath);
         XPathParts untransformedfullxpp = XPathParts.getFrozenInstance(untransformedFullPath);
 
-        for (SplittableAttributeSpec s : LdmlConvertRules.SPLITTABLE_ATTRS) {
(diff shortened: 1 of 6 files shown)


srl295:
I think i might merge my CLDR-14313 work here too since it works on similar files

srl295:
@sffc fun with first time use of Numberformatter percents!

srl295:
☝️ updated the description and changelog

btangmu (approved):
Looks OK though I don't fully understand it

PR #889, opened 2020-12-17
CLDR-13987 improve @CLDRTool annotation, default CLDR_DIR in tools

srl295:
CLDR-13987

- use a default CLDR_DIR if the jar is launched from a CLDR dir
- add docs for SearchXML
- correct cldr.jar to cldr-code.jar

Also see rewrites to http://cldr.unicode.org/tools

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/Main.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/Main.java
@@ -8,27 +8,29 @@ import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRTool;
 
 /**
- *
- * Implement a 'main' for the CLDR jar.
+ * Implement a 'main' for the cldr-code jar.
  */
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/Main.java
```diff
@@ -38,6 +40,17 @@ public static void main(String args[]) throws Throwable {
             System.arraycopy(args, 1, args2, 0, args2.length);
             Class<?> c = null;
 
+            try {
```
btangmu, line 43: quibble: better to make a subroutine; main is too long already

btangmu (approved):

macchiati (approved):
Tom's point about reducing the size of the main is fine, but could also be done in separate PR (and do more "method extraction" than just this).

tools/cldr-code/src/main/java/org/unicode/cldr/tool/Main.java
srl295: makes sense

PR #895, opened 2020-12-23
CLDR-14251 Modernize: start a new version of the front end without dojo

btangmu:
-New boolean SurveyTool.USE_DOJO determines usage of the dojo or non-dojo front end

-Works as before if USE_DOJO is true; does not work yet if USE_DOJO is false

-New folder js/new has js files that do not depend on dojo (directly or indirectly)

-Changed file/object naming conventions; shorter; e.g., cldrAjax instead of cldrStAjax

-Rename some dojo-dependent files to start with CldrDojo; e.g., CldrDojoForum.js

-Fix bug in PR 893: cldrGui.makeVoteCountMenu was incomplete

-Fix bug in PR 893: stopIcon should be cldrStatus.stopIcon() in survey.js

-Fix bug in PR 893: bad js syntax in SurveyTool.serveWaitingPage window.setTimeout

-Remove dead code; clean up

-Comments

##### Checklist

### DIFF
--- a/tools/cldr-apps/src/main/webapp/js/new/cldrLocaleMap.js
+++ b/tools/cldr-apps/src/main/webapp/js/new/cldrLocaleMap.js
@@ -1,0 +1,153 @@
+"use strict";
+
+/**
+ * @param aLocmap the map object from json
+ */
+function LocaleMap(aLocmap) {
+  this.locmap = aLocmap;
+}
+
+/**
+ * Run the locale id through the idmap.
+ *
+ * @param menuMap the map
+ * @param locid or null for cldrStatus.getCurrentLocale()
+ * @return canonicalized id, or unchanged
+ */
+LocaleMap.prototype.canonicalizeLocaleId = function canonicalizeLocaleId(
+  locid
+) {
+  if (locid === null) {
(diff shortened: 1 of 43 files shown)


tools/cldr-apps/src/main/webapp/js/new/cldrLocaleMap.js
```diff
@@ -0,0 +1,153 @@
+"use strict";
+
+/**
+ * @param aLocmap the map object from json
+ */
+function LocaleMap(aLocmap) {
+  this.locmap = aLocmap;
```
btangmu, line 7: Note: "it's no longer possible to reference the window object through _this_ inside a strict mode function" -- so this file needs modernizing. This file isn't used, though, if USE_DOJO is true. There's still a lot that won't work yet if USE_DOJO is false.

btangmu: It should be easy to turn LocaleMap into a class with a constructor(), and simplify the syntax, no more .prototype.

btangmu:
I found a bug, typo "xhrArs" instead of "xhrArgs" in cldrAjax.js. I'll fix that and make another PR.

PR #900, opened 2021-01-01
CLDR-14309 move GenerateLanguageContainment and WikidataSubdivisions to use direct RDF query

srl295:
CLDR-14309

- instead of asking the user to do indididual .tsv downloads, the sparql queries are done by the tool.

This change obviates the instructions in <http://cldr.unicode.org/development/updating-codes/updating-language-groups>

Everything above _This will create… languageGroup.xml_ should be replaced with:

```shell
mvn --file=tools/pom.xml compile -DskipTests=true
mvn --file=tools/pom.xml -DCLDR_DIR=${HOME}/src/cldr  -DmainClass=org.unicode.cldr.tool.GenerateLanguageContainment exec:java -pl cldr-rdf
```


srl295:
Note that I did not check in the generated file, but I did verify that the "old" method produces the same `languageGroup.xml`.  There are issues once that file is generated, see #666

srl295:
For subdivisions, comparing the 'manual' (left) method to the 'new' method (right) is very close.

```diff
$ diff -r  ../common/subdivisions/ ~/src/cldr/common/subdivisions/
diff -r ../common/subdivisions/cy.xml /Users/srl295/src/cldr/common/subdivisions/cy.xml
1921c1921
< 			<subdivision type="si040" draft="contributed">Izola</subdivision>
---
> 			<subdivision type="si040" draft="contributed">Bwrdeistref Izola</subdivision>
diff -r ../common/subdivisions/hr.xml /Users/srl295/src/cldr/common/subdivisions/hr.xml
1058d1057
< 			<subdivision type="mk52" draft="contributed">Općina Gostivar²</subdivision>
```

srl295:
http://cldr.unicode.org/development/updating-codes/updating-subdivision-translations also needs to change, as simply running the tool   `WikiSubdivisionLanguages` is all that is needed.

- I had to move `GenerateSubdivisions` and `SubdivisionNode` into the cldr-rdf project due to interlock
- To recap, neither the LanguageGroup nor the Subdivision tool look like they had been run recently. I did not change any actual data with this change, merely improved the generation phase to not have manual steps.

tools/cldr-rdf/src/main/java/org/unicode/cldr/rdf/QueryClient.java
```diff
@@ -37,29 +47,104 @@ public static final QueryClient getInstance() {
...
     
-    public static final String getStringOrNull(final QuerySolution qs, final String k) {
+    public static final String getLiteralOrNull(final QuerySolution qs, final String k) {
         final Literal l = qs.getLiteral(k);
-        if(l == null) return null;
+        return getLiteralOrNull(l);
+    }
+    
+	private static String getLiteralOrNull(final Literal l) {
+		if(l == null) return null;
```
btangmu, line 56: formatting is off?

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
```diff
@@ -49,37 +52,78 @@
...
+        
+        String getEntityName(String key) {
+            String code = entityToCode.get(key);
+            if (code != null) {
+                try {
+                    String name = NAME.apply(code);
+                    if (name != null) {
+                        return name;
+                    }
+                } catch (Exception e) {
```
btangmu, line 112: why would NAME.apply throw an exception? ideally prevent the exception, or print a warning when it happens

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
```diff
@@ -418,68 +465,72 @@ private static void log(String string) {
...
-        int last = key.lastIndexOf('/');
-        return key.substring(last + 1, key.length() - 1);
-    }
-
-    private static Multimap<String, String> loadTsvPairs(Class<?> class1, String file,
-        Function<String, String> keyMapper, Function<String, String> valueMapper) {
-        String rel = FileUtilities.getRelativeFileName(class1, file);
-        System.out.println(rel);
+//    private static String getBest(Collection<String> parents) {
+//        for (String parent : parents) {
```
btangmu, line 469: better to delete than to comment out

btangmu (commented):
I made a few comments but don't understand the code well enough. We can discuss Tuesday if you'd like me to study it more

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
srl295: I'm not sure— this was existing code from https://github.com/unicode-org/cldr/pull/900/files/89fe206ccd3bc8cfbd4b308148d21e4d190ed14f#diff-f5160be6c0e53da3045f35879f2bab04b51ce622ab6899c90062a6cc1bb3bf44L439 

this calls into a pretty deep stack of code in CLDRFile, it definitely could throw an exception of some sort. 

But for the purpose of this operation, not being able to get a name at that level means that it will fall back to the rest of this function which will try other mechanisms for extracting a name.

This code already needs other work (see #666), the purpose of this PR is to reduce the manual steps  needed to 'turn the crank' on the process.

i'm inclined to not change this code until after the other issues in #666 are addressed.

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
srl295: True, in this case it's used by other code that's also commented out at https://github.com/unicode-org/cldr/pull/900/files/89fe206ccd3bc8cfbd4b308148d21e4d190ed14f#diff-f5160be6c0e53da3045f35879f2bab04b51ce622ab6899c90062a6cc1bb3bf44L387

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
srl295: Perhaps I could add a TODO here?

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
srl295: Perhaps I could add a "TODO: only used by commented out code" here?

srl295:
> I made a few comments but don't understand the code well enough. We can discuss Tuesday if you'd like me to study it more

thanks. As I noted above, the main change here is to replace the .tsv files with using the actual API to run the same query.

tools/cldr-rdf/src/main/java/org/unicode/cldr/rdf/QueryClient.java
srl295: @btangmu let's merge this PR (once. you're happy with other aspects) and I'll open a new one to fix the formatting for cldr-rdf.

srl295:
@btangmu we can discuss today— I've updated, fixed the indentation in QueryClient.java, and I'll make a followon PR that brings in formatting for the `cldr-rdf` project.

srl295:
OK…
- Now also updates tsv files in `cldr/tools/cldr-rdf/external`
- I did NOT check in an actual change to the tsv files for the same reasons above that I didn't update languageGroups.xml nor the subdivision xml data.

In other words, you can run the tool and see the .tsv files change.

macchiati (approved):

srl295:
Missed an ambiguous overload

btangmu (approved):

PR #904, opened 2021-01-04
CLDR-14251 Modernize: more front end without dojo

btangmu:
-Reports, Forum, About, AdminPanel are now mostly working without dojo

-New cldrAbout.js and AboutST.java make the About page instead of about.jsp

-New cldrAdmin.js has Admin Panel code from old survey.js

-New AdminPanel.java replaces AdminAjax.jsp and AdminPanel.jsp

-New RecentActivity.java and cldrRecentActivity.js replace myvotes.jsp

-Move showForumStuff, etc., to new cldrForumPanel.js

-Move showInPop, etc., to new cldrInfo.js

-Move createChunk, etc., to new cldrDom.js

-Move showARIDialog, etc., to new cldrRetry.js

-Move oldvotes code to new cldrOldVotes.js

-Move loadMail code to new cldrMail.js

-Move report code to new cldrReportDash.js, cldrReportDates.js, cldrReportNumbers.js, cldrReportZones.js

-Move loadLocales code to new cldrLocales.js

-New unit tests TestCldrAbout.js, TestCldrAdmin.js, TestCldrCreateLogin.js

-Move cldrText.subVerbose into TestCldrText.js

-Remove Toggle Sidebar from the gear menu

-Show About in the gear menu even if not logged in

-Avoid nested functions; reduce function length by using subroutines

-Take advantage of ES6 Object Property Value Shorthand for IIFE return

-Remove dead code; clean up

-Comments

##### Checklist


tools/cldr-apps/js-unittest/TestCldrText.js
```diff
@@ -98,11 +98,85 @@
...
+  }
+
+  /**
+   * Get the string that results from filling in blanks in the given template
+   *
+   * @param template a string like "Sorry, your vote for '${1}' could not be submitted: ${0}"
+   * @param map an array like ["too goofy", "🤪"]
+   * @return a string like "Sorry, your vote for '🤪' could not be submitted: too goofy"
+   */
+  function fillInBlanksWithArray(template, map) {
```
srl295, line 136: ok for now, would be good to use an i18n library function eventually.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/AdminPanel.java
```diff
@@ -0,0 +1,253 @@
...
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.unicode.cldr.util.CLDRConfig;
+import org.unicode.cldr.util.CLDRConfigImpl;
+import org.unicode.cldr.web.SurveyAjax.JSONWriter;
+
+public class AdminPanel {
+    public void getJson(JSONWriter r, HttpServletRequest request, HttpServletResponse response, SurveyMain sm) throws JSONException, IOException {
```
srl295, line 23: should this be an abstract interface?

srl295 (approved):
Lot of good work here!
Certainly the `someHtml = '<b> + … '` type of HTML synthesis should all end up being a template of some kind. But I think that's further down the road in our migration.  Really like the modular approach.

tools/cldr-apps/js-unittest/TestCldrText.js
btangmu: Yes. It's clear that nls/stui.js was meant for i18n -- maybe someday we can use <i>modern</i> dojo or something similar for i18n and actually provide some alternatives to English for the ST UI!

Note that fillInBlanksWithArray is only in the "verbose" version of sub, which is only for unit testing. The production version is only 12 lines -- I was surprised at how small it got:
```
  function sub(k, map) {
    const template = cldrText.get(k);
    if (template) {
      if (map instanceof Array) {
        return template.replace(/\${(\d)}/g, (blank, i) => map[i]);
      }
      if (map instanceof Object) {
        return template.replace(/\${([^}]+)}/g, (blank, i) => map[i]);
      }
    }
    return "";
  }
```

btangmu:
> Certainly the `someHtml = '<b> + … '` type of HTML synthesis should all end up being a template of some kind.

Definitely that will be better! I'm eager to get started with Vue.

> But I think that's further down the road in our migration. Really like the modular approach.

Soon we should have actual "modules" with import/export. That should coincide more or less with getting nearly everything working again without dojo, without jsp, making all js strict, serving only json on the back end, and getting started with Vue.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/AdminPanel.java
srl295: Hmm i would actually go the other way.   Considering : https://github.com/unicode-org/cldr/pull/904/files#diff-00d9540e776e546a0ea8468e69245839e6f6e168af708ed3f902bf81d64ea90dR384

you could actually have a Map from string to JsonAnswerer (with a better name of course) and replace the switch statement with:

`m.get(what).getJSON(r, request, response)`

and we'd have a little dispatch table.

btangmu: Actually I was wrong: I had <em>not</em> made that change to AdminPanel, I had only thought about it. Looking at existing getJson functions --

* new AdminPanel().getJson(r, request, response, sm);
* new SurveyForumParticipation(org).getJson(r);
* new SurveyVettingParticipation(org, sm).getJson(r);
* new SurveyBulkClosePosts(sm, execute).getJson(r);
* AboutST.getJson(r, sm); 

-- AboutST is exceptional (and exceptionally short). Your JsonAnswerer idea sounds good, and might help with refactoring SurveyAjax to avoid a million "else if (what.equals(WHAT_...)) { ..." blocks. For now, I'm leaving AdminPanel as-is...

PR #905, opened 2021-01-04
CLDR-8802 add Currency abstracts

srl295:
CLDR-8802
- also switch off DEBUG on some of the mappers

image

note that the xpath binding is only on the currency **name**, perhaps we should use a prefix model so that other items (symbol, count-one, count-many, etc) are also handled.

### DIFF
--- a/tools/cldr-apps/src/main/webapp/css/redesign.css
+++ b/tools/cldr-apps/src/main/webapp/css/redesign.css
@@ -676,7 +676,6 @@ div.helpAbstract {
     font-size: smaller;
     overflow: hidden;
     height: 10em;
-    border: 1px solid gray;
     padding: .25em;
     font-style: italic;
     position: relative;
@@ -692,7 +691,7 @@ div.helpAbstract {
     text-align: right;
     width: 100%;
     background-color: #d9edf7;
-    opacity: .7;
+    opacity: 1;
 }
 
 .nav .popover {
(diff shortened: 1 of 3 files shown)


srl295:
guide for this is here http://cldr.unicode.org/development/updating-abstracts

macchiati (changes requested):
The changes look good, but can you fix the presentation of the information? I should have brought this up earlier. Look at:

https://user-images.githubusercontent.com/855219/103574372-69612580-4e95-11eb-86e4-bfe18136e2cd.png

The format looks broken because "more" overlaps with the text. It is more apparent with other examples where the last line is longer.

I suggest you move the "more" line _below_ the box.

srl295:
image

How does this look?
- removed the border
- the 'more' is part of the box, so what I actually did is _replace_ the last line with the 'more' (made it opaque instead of translucent)

btangmu:
> How does this look?

Looks good!

srl295:
> [View Diff Across Force-Push](https://github.com/unicode-org/cldr/compare/unicode-org:d1a4493..unicode-org:e90d981)

CSS change only. Done.

btangmu (approved):

srl295:
@macchiati any thoughts on this? would like to merge before the codebase goes stale…

macchiati (approved):

PR #909, opened 2021-01-06
CLDR-14373 Updating time zone data to 2020f

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14373


common/supplemental/metaZones.xml
```diff
@@ -1390,9 +1390,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 				<usesMetazone from="1999-10-31 01:00" mzone="Europe_Eastern"/>
 			</timezone>
 			<timezone type="Europe/Volgograd">
-				<usesMetazone to="1992-03-28 22:00" mzone="Volgograd"/>
-				<usesMetazone to="2018-10-27 23:00" from="1992-03-28 22:00" mzone="Moscow"/>
```
yumaoka, line 1394: After reviewing 2020e data changes for Volgograd, I was wondering if we should add metazone mapping to Moscow at the end. It is true that Volgograd use same offset with Moscow, and 1 hour ahead back and forth. We already created a metazone dedicated for Volgograd. I also reviewed old tz database, but it assigned abbreviation separated from Moscow for this period. My conclusion is it makes more sense to use metazone Volgograd all the time.

pedberg-icu (approved):
Thanks, looks good, agree with Volgograd decision.

common/supplemental/metaZones.xml
```diff
@@ -404,8 +404,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 				<usesMetazone from="1988-10-30 02:01" mzone="Atlantic"/>
 			</timezone>
 			<timezone type="America/Grand_Turk">
-				<usesMetazone to="2015-11-01 06:00" mzone="America_Eastern"/>
-				<usesMetazone to="2018-03-11 07:00" from="2015-11-01 06:00" mzone="Atlantic"/>
+				<usesMetazone to="2015-03-08 07:00" mzone="America_Eastern"/>
+				<usesMetazone to="2018-03-11 07:00" from="2015-03-08 07:00" mzone="Atlantic"/>
```
mj1856, line 408: @yumaoka - Curious the reason for the change here?  The previous date seems more correct with regard to name changes.
https://www.timeanddate.com/time/zone/@3576993?year=2015

PR #918, opened 2021-01-08
CLDR-14251 make USE_DOJO a server side and a request-time param

srl295:
CLDR-14251
`USE_DOJO` (default true) can now be set in cldr.properties
also, any request can have `?USE_DOJO=false` or `?USE_DOJO=true` to override

It turned out to be very simple to make this switchable. Could be helpful in testing old vs. new behavior.

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
@@ -13,12 +13,52 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.json.JSONException;
+import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.VettingViewer;
 
 public class SurveyTool extends HttpServlet {
+    private static final String USE_DOJO_VAR = "USE_DOJO";
     private static final long serialVersionUID = 1L;
-    public static final boolean USE_DOJO = true;
+    private static final boolean USE_DOJO_DEFAULT = true;
+
+    /**
+     * Is dojo enabled by default?
+     * Better to use the other variants taking a WebContext or an HttpServletRequest
+     * Making this private unless there is a good reason to use the default
+     * from another caller
+     * @deprecated this will be going away
+     * @return
+     */
+    @Deprecated
+    private static final boolean defaultUseDojo() {
+        return CLDRConfig.getInstance().getProperty(USE_DOJO_VAR, USE_DOJO_DEFAULT);
+    }
+
+    /**
+     * Is dojo enabled?
+     * @param ctx request context
+     * @deprecated this will be going away
+     * @return
+     */
+    @Deprecated
+    public static final boolean useDojo(WebContext ctx) {
+        return useDojo(ctx.request);
+    }
+    /**
+     * do we use Dojo?
+     * @param ctx request context
+     * @deprecated this will be going away
+     * @return
+     */
+    @Deprecated
+    public static final boolean useDojo(HttpServletRequest ctx) {
+        String param = ctx.getParameter(USE_DOJO_VAR);
+        if(param != null && !param.isEmpty()) {
+            return Boolean.parseBoolean(param); // override
+        }
+        return defaultUseDojo();
+    }
 
     @Override
     public final void init(final ServletConfig config) throws ServletException {
@@ -134,7 +174,7 @@ public class SurveyTool extends HttpServlet {
             out.write("if (dstatus != null) {\n");
             out.write("  dstatus.appendChild(document.createElement('br'));\n");
             out.write("  dstatus.appendChild(document.createTextNode('.'));\n");
-            if (USE_DOJO) {
+            if (useDojo(request)) {
                 out.write("  require(['dojo/ready'], function(ready) {\n");
                 out.write("    ready(function() {\n");
             } else {
@@ -160,7 +200,7 @@ public class SurveyTool extends HttpServlet {
             out.write("          cldrAjax.sendXhr(xhrArgs);\n");
             out.write("        }, 2000); // two seconds\n");
             out.write("      });\n");
-            if (USE_DOJO) {
+            if (useDojo(request)) {
                 out.write("    }\n");
                 out.write("  });\n");
             }
@@ -268,7 +308,7 @@ public class SurveyTool extends HttpServlet {
         out.write("</head>\n");
         out.write("<body lang='" + lang + "' data-spy='scroll' data-target='#itemInfo'>\n");
         out.write("<p id='st-run-gui'>Loading...</p>\n");
-        if (!USE_DOJO) {
+        if (!useDojo(request)) {
             out.write("<script>cldrGui.run()</script>\n");
         }
         out.write("</body>\n</html>\n");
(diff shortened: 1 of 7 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -13,12 +13,49 @@
...
+    }
+    /**
+     * do we use Dojo?
+     * @param ctx request context
+     * @deprecated this will be going away
+     * @return
+     */
+    @Deprecated
+    public static final boolean useDojo(HttpServletRequest ctx) {
+        String param = ctx.getParameter("USE_DOJO");
```
btangmu, line 53: I'm guessing there's an easy way for the user to add "&USE_DOJO=false" to all the urls? I mean not just in the address bar, but in all the automatic xhr requests, etc? How is that done?

btangmu (commented):
This is awesome! It almost makes me reluctant to remove Dojo entirely! Let's think about how to retain this kind of ability long-term use not just for Dojo

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
srl295: shouldn't affect XHRs? or does it.

srl295:
@btangmu So one thing is that this depends on the inbound request… once we go to webpack CLDR-14400 then the front end will be static, not generated based on parameters. Switching wont' be as easy at that point.

That could be a good reason to defer moving to webpack.

btangmu (commented):
I meant to approve this earlier, and I thought I had. Now it will be slightly in conflict with PR #919 that was merged. I'll try to fix that...

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
btangmu: Yes, it can affect XHRs, such as WHAT_ADMIN_PANEL. I think if USE_DOJO is false, then WHAT_ADMIN_PANEL is going to behave as though USE_DOJO were true, and the front and back ends won't cooperate.

A solution might be to add "&USE_DOJO=false" to all urls in cldrAjax.js, only for the non-dojo version of cldrAjax.js. Currently we use the same cldrAjax.js for dojo and non-dojo. We could clone it as CldrDojoAjax.js, or add a USE_DOJO constant to the front end...

srl295:
Is it possible to have the server side be the same ? I’ll investigate the
call sites some more. And no problem.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
btangmu: SurveyAjax.java has:
```
                    if (what.equals(WHAT_ADMIN_PANEL)) {
...
                            if (SurveyTool.USE_DOJO) {
                                response.sendRedirect(request.getContextPath() + "/AdminPanel.jsp" + "?vap=" + SurveyMain.vap);
                            } else {
...
                                JSONWriter r = newJSONStatus(request, sm);
...
                                new AdminPanel().getJson(r, request, response, sm);
```

the useDojo(request) method can't be used in place of USE_DOJO here unless the request specifies USE_DOJO.

If we really want to make the dojo/non-dojo switch a request param, we'd better make a clean break between dojo and non-dojo versions of all the js files; sharing would be too error-prone. Only these 5 are currently shared:

* new/cldrAjax.js
* new/cldrCsvFromTable.js
* new/cldrForumFilter.js
* new/cldrStatus.js
* new/cldrText.js

Maybe generalize the request syntax: `..&frontend=dojo` or `..&frontend=vue`or `..&frontend=default`

btangmu:
I'm pretty sure the XHR problem is real, so this needs more work, maybe a couple hours -- let's weigh the pros and cons

Pros:
* Convenient for comparing dojo/non-dojo behavior while doing the the remaining non-dojo implementation, could make that development faster, more testable
* Could retain in generalized form for multiple versions of frontend, including debugging (non-minified) versus production (minified) versions as well as versions with different dependencies (dojo/vue/webpack...), or newer/older versions...

Cons:
* Additional complexity on both front and back ends
* Maintenance burden to keep multiple alternative versions working beyond a relatively short transition period before older versions are retired
* Code duplication should be avoided (except maybe when one version is meant to be  treated as essentially read-only, as historical reference ...)

srl295:
Actually I reviewed the sites and it shouldn't be too bad.
- the one thing on the client side that was missing is `const USE_DOJO=true / false;`  to communicate the default without needing to check the query string
- then wherever the xhr is created, USE_DOJO can be communicated.

Let me see if I can do it, if you don't like it then we can try something else or scrap this.

srl295:
@btangmu think I got it… please take a look.

also fixed admin panel under use_dojo=false

tools/cldr-apps/src/main/webapp/AdminPanel.jsp
```diff
@@ -9,6 +9,11 @@ if(vap==null ||
 	response.sendRedirect("http://cldr.unicode.org"); // Get out.
 	return;
 }
+if (SurveyTool.useDojo(request) == false) {
+    // AdminPanel.jsp is the dojo version. Redirect to the new one.
+    response.sendRedirect("./v?USE_DOJO=false#admin///");
```
srl295, line 14: this jsp is in admin.html and other places, so it's worth having this redirect here

tools/cldr-apps/src/main/webapp/js/new/cldrLoad.js
```diff
@@ -606,6 +606,7 @@ const cldrLoad = (function () {
       r_vetting_json: cldrDash,
       recent_activity: cldrRecentActivity,
       retry: cldrRetry,
+      admin: cldrAdmin,
```
srl295, line 609: this seems like it was just an oversight

tools/cldr-apps/src/main/webapp/js/new/cldrAjax.js
```diff
@@ -158,8 +158,12 @@ const cldrAjax = (function () {
     } else {
       options.method = "GET";
     }
-
     const request = new XMLHttpRequest();
+    if(xhrArgs.url.indexOf('?') == -1) {
+      xhrArgs.url += '?USE_DOJO='+USE_DOJO;
+    } else {
+      xhrArgs.url += '&USE_DOJO='+USE_DOJO;
+    }
```
srl295, line 166: adds USE_DOJO to all XHRs. One place, will be easy to remove when we can.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -355,7 +395,9 @@ private static void includeCldrJavaScript(HttpServletRequest request, Writer out
 
         out.write(prefix + "jquery.autosize.min.js" + tail); // exceptional
 
-        if (USE_DOJO) {
+        final Boolean doUseDojo = useDojo(request);
+        out.write(String.format("<script>const %s=%s;</script>\n", USE_DOJO_VAR, doUseDojo.toString()));
```
btangmu, line 399: Global variable? Maybe OK if very temporary??

srl295: It's a local, just because it's used twice.

srl295: Oh, the JavaScript variable, yes, this should be very temporary and removed when DOJO is.

btangmu (approved):
OK as a very temporary scaffolding until we remove dojo, as planned before the end of January. USE_DOJO should be the last js global variable we ever use!

srl295:
@btangmu OK, sometimes the computer IS right.  Please rereview, I had missed merging in the PR that (as you noted) broke this.

btangmu (approved):

PR #919, opened 2021-01-08
CLDR-14251 Modernize: fix localizeAnon bug; for non-dojo, add vue, upgrade jquery

btangmu:
-Fix localizeAnon TRANS_HINT_LANGUAGE_NAME bug affecting both dojo and non-dojo versions

-For non-dojo, add vue.js and upgrade to latest version of jquery and jquery-ui

-Add tiny test for vue.js in cldrAbout.js

-Comments

##### Checklist


srl295 (approved):
I think this used to work because the 'old' interpolator supported more complex formats before…

Anyways, sounds good, this will invalidate #918 so I will update that when it's merged

PR #920, opened 2021-01-11
CLDR-14360 chart fixes

macchiati:
There are various small changes (at this point) needed to fix the charts. The BNF-related changes are in anticipation. There are a few updates in syntax to old code.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14360


macchiati:
This is really a work in progress. Much more to follow

PR #922, opened 2021-01-11
CLDR-14355 remove special zh-Hans↔zh-Hant language distances

markusicu:
https://unicode-org.atlassian.net/browse/CLDR-14355


pedberg-icu (approved):

markusicu:
FYI @pedberg-icu please check with me next time you do an integration into ICU -- this should require some changes in expected test results.

PR #929, opened 2021-01-13
CLDR-11268 Portugal Portuguese should use the long scale and Brazil should use the short scale

grhoten:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11268

In addition to the short scale versus long scale difference, I simplified the rule logic and mirrored the style between the two Portuguese variants to make it easier to compare for differences in the future.

I used https://www.flip.pt/Duvidas-Linguisticas/Duvida-Linguistica/DID/4389 for reference to determine the scale used.


grhoten:
@srl295 @macchiati Please review and merge this pull request since I can not assign reviewers
@iwsfutcmd FYI about these changes.

macchiati (approved):

PR #933, opened 2021-01-13
CLDR-14392 Modernize Bulk Close Posts in gear menu

btangmu:
-For new (non-dojo) front end: include bulk_close_posts in getSpecial; no load params

-Write progress to java log so Admin user can see it

-Fix cldrAjax timeout handling (dojo and non-dojo): err if timeout; default no timeout

-Add new cldrAjax.mediumTimeout to restore old 2-sec timeout for certain requests

-Alphabetize specials array including admin

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14392

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
@@ -1098,25 +1098,36 @@ public class SurveyForum {
     /**
      * Close all posts in the threads with the given root ids
      *
+     * This is a long-running process (several minutes) if the number of posts is large. There is no
+     * progress feedback on the front end, unfortunately. As a work-around, print progress to Java log.
+     * Only Admin can use this feature. Admin should be able to watch log even on production server.
+     *
+     * This feature could probably be made a hundred times faster by using an sql stored procedure!
+     *
      * @param conn the db connection
-     * @param rootIdList the list of post ids
+     * @param rootIdList the list of post ids (typically about 4 thousand)
      * @return the number of posts closed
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
```diff
@@ -1098,25 +1098,36 @@ private int savePostToDb(PostInfo postInfo) throws SurveyException {
     /**
      * Close all posts in the threads with the given root ids
      *
+     * This is a long-running process (several minutes) if the number of posts is large. There is no
+     * progress feedback on the front end, unfortunately. As a work-around, print progress to Java log.
+     * Only Admin can use this feature. Admin should be able to watch log even on production server.
+     *
+     * This feature could probably be made a hundred times faster by using an sql stored procedure!
```
srl295, line 1105: I don't know the whole context, but file a new perf bug for this speed up if needed. 

Also, it could be made async but let's wait until there's a better api substrate. 

It could be just moved to a surveytask if there's no interactive response needed. That would keep it from timing out.

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
btangmu: It's only run likely to be run once per version, so not a high priority for performance optimization. And yes, the only feedback really required is confirmation that it completed successfully -- and that does now occur, after five or ten minutes :-)

PR #921, opened 2021-01-11
CLDR-13418 German cardinal doesn't have spellout-cardinal-m rule

grhoten:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13418
- [x] Updated PR title and link in previous line to include Issue number

For completeness, I added one for the ordinal too. For consistency, I used the same naming scheme of cardinal-m as the other rules. This rule seems to apply to the masculine dative and neuter dative case only in German.

I got a German expert to review the changes. The changes seem fine within the scope of the changes. There were some additional comments about other existing rules, but that should be handled with separate tickets.

PR #935, opened 2021-01-14
CLDR-10823 Fix in German RBNF numbering-year

grhoten:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-10823

This will have to be merged after https://github.com/unicode-org/cldr/pull/921 because there are file conflicts, but I wanted to get this started for a review.

After consulting with @markusicu and our fellow German experts, we determined the following.

- The range of 2000-2099 should be using the thousands (e.g. zweitausendeins for 2001). While the English style "twenty twenty" could be used for 2020-2099, it's perhaps a little too casual. The original ticket didn't mention the range, and we're explicitly stating for future posterity why we didn't go this route. @markusicu did state "1100..1999 are of the form 'eleven hundred' .. 'nineteen hundred ninety nine' in German." Our German experts agree on this point.
- The ein should still precede the tausend and hundert. As one our German expert stated, "it sounds more eloquent." So we're leaving these forms untouched. If people desire to really omit the ein, it should be submitted in a separate ticket for discussion and reviewed for all of the numbers in German and not just the numbering-year.

### DIFF
--- a/common/rbnf/de.xml
+++ b/common/rbnf/de.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
 <!--
-Copyright © 1991-2013 Unicode, Inc.
+Copyright © 1991-2021 Unicode, Inc.
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
 For terms of use, see http://www.unicode.org/copyright.html
 -->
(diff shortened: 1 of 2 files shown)


common/rbnf/de.xml
```diff
@@ -16,7 +16,23 @@ For terms of use, see http://www.unicode.org/copyright.html
                 <rbnfrule value="-x">minus →→;</rbnfrule>
                 <rbnfrule value="x.x">=0.0=;</rbnfrule>
                 <rbnfrule value="0">=%spellout-numbering=;</rbnfrule>
-                <rbnfrule value="1100" radix="100">←%spellout-cardinal-masculine←­hundert[­→→];</rbnfrule>
+                <rbnfrule value="1100" radix="100">←←­hundert[­→→];</rbnfrule>
+                <rbnfrule value="2000">=%spellout-numbering=;</rbnfrule>
+                <rbnfrule value="2100" radix="100">←←­hundert[­→→];</rbnfrule>
```
markusicu, line 21: I am trying to say these as the German version of "twenty-one hundred fifty three" and it sounds weird to me.
I suggest you just use `=%spellout-numbering=` for everything above 1999.

common/rbnf/de_CH.xml
```diff
@@ -1,207 +1,223 @@
...
+    </identity>
+    <rbnf>
+        <rulesetGrouping type="SpelloutRules">
+            <ruleset type="spellout-numbering-year">
+                <rbnfrule value="-x">minus →→;</rbnfrule>
+                <rbnfrule value="x.x">=0.0=;</rbnfrule>
+                <rbnfrule value="0">=%spellout-numbering=;</rbnfrule>
+                <rbnfrule value="1100" radix="100">←←­hundert[­→→];</rbnfrule>
+                <rbnfrule value="2000">=%spellout-numbering=;</rbnfrule>
+                <rbnfrule value="2100" radix="100">←←­hundert[­→→];</rbnfrule>
```
markusicu, line 22: ditto for other de-*

common/rbnf/de.xml
grhoten: I have lowered the range from 10000 to 2000 as you requested. I guess we can change our minds in the next few decades if the prevailing wording changes.

markusicu (approved):
lgtm tnx

PR #937, opened 2021-01-15
CLDR-10920 RBNF for Venetian

grhoten:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-10920


srl295:
for what it's worth, main venetian CLDR-13879 isn't in yet.

srl295 (approved):
structure looks reasonable, would be good to get some linguistic review

PR #939, opened 2021-01-15
CLDR-14343 remove "accordian" from emoji keyword

Kristi27:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14343


srl295:
> commit message for 41e8255 fails validation

The commit message is `Update en.xml` but needs to be something like `CLDR-14343 remove “accordian” from emoji keyword`

andrew-z (approved):

srl295 (approved):

PR #940, opened 2021-01-17
CLDR-11407 Spelt out date past 1999 should use multiples of a thousand

grhoten:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11407

### DIFF
--- a/common/rbnf/fr.xml
+++ b/common/rbnf/fr.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
 <!--
-Copyright © 1991-2013 Unicode, Inc.
+Copyright © 1991-2021 Unicode, Inc.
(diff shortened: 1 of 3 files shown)


srl295 (approved):
does what it says, maybe @macchiati can confirm the French

grhoten:
If we want to remake it to be more like English, which repeats the pattern through 10,000 we can do that, but I've gotten conflicting information on the best approach here due to lack of precedent when referring to distant future years. So I'm just limiting the scope to what was mentioned in the reported issue.

English follows the pattern mentioned in the song "In The Year 2525" by Zager And Evans, which I find reasonable for English. Beyond the year 10,000 is pure speculation on the pronunciation. This song normally isn't applicable outside of English. I'm not going to count Futurama's time travel song for year pronunciation beyond 10,000 into account for English because English probably won't be the same when we need to pronounce such numbers regularly, and it's incredibly rare to refer to years beyond 3000 in the Gregorian calendar. The current English rule range is hard to apply outside of English.

PR #943, opened 2021-01-18
CLDR-14159 DAIP should trim all whitespace

btangmu:
-Revise normalizeWhitespace to call subroutines, including new trimNBSP

-Add new unit test TestDisplayAndInputProcessor.TestWhitespaceNormalization

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14159

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
@@ -81,14 +81,21 @@ public class DisplayAndInputProcessor {
         + "numbers/(decimal|currency|percent|scientific)Formats.+/(decimal|currency|percent|scientific)Format.*)");
     private static final Pattern INTERVAL_FORMAT_PATHS = PatternCache.get("//ldml/dates/.+/intervalFormatItem.*");
     private static final Pattern NON_DECIMAL_PERIOD = PatternCache.get("(?<![0#'])\\.(?![0#'])");
-    private static final Pattern WHITESPACE_NO_NBSP_TO_NORMALIZE = PatternCache.get("\\s+"); // string of whitespace not
-    // including NBSP, i.e. [
-    // \t\n\r]+
-    private static final Pattern WHITESPACE_AND_NBSP_TO_NORMALIZE = PatternCache.get("[\\s\\u00A0]+"); // string of
-    // whitespace
-    // including NBSP,
-    // i.e. [
-    // \u00A0\t\n\r]+
+
+    /**
+     * string of whitespace not including NBSP, i.e. [\t\n\r]+
+     */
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -581,9 +588,37 @@ private String inputUnicodeSet(String path, String value) {
...
+     * Also trim initial/final NBSP, unless the value is only the one character, "\u00A0"
+     *
+     * @param path
+     * @param value
+     * @return the normalized value
+     */
     private String normalizeWhitespace(String path, String value) {
-        // turn all whitespace sequences (including tab and newline, and NBSP for certain paths)
-        // into a single space or a single NBSP depending on path.
+        if (wantsRegularSpace(path)) {
```
macchiati, line 601: The cases are, I think:

path doesn't allow NBSP
path doesn't allow SP
path allows either

I think it would be cleaner to have a function that returns an enum {allowSp, allowNbsp, allowSpOrNbsp}

The code for the first two cases looks ok. However, for the third, we want to result in an A0 if there is one in the whitespace sequence, otherwise result in SP. That will need a different REGEX.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
```diff
@@ -392,4 +392,19 @@ public void TestBackspaceFilter() {
...
     }
+
+    /**
+     * Test whether DisplayAndInputProcessor.processInput normalizes whitespace.
+     * This depends very much on the xpath, since for most xpaths NBSP is normalized to
+     * ordinary space which, if initial or final, is then removed by trim(). But for some
+     * xpaths, NBSP is retained and then the standard trim() function doesn't apply.
+     */
+    public void TestWhitespaceNormalization() {
+        DisplayAndInputProcessor daip = new DisplayAndInputProcessor(info.getEnglish(), false);
```
macchiati, line 403: I think you need a few more cases to test. Each of the 3 types of paths, and for each, samples with

A0 20
20 A0
20 A0 20
20 20 

Also, you should test this by running CLDRModify over the source and collecting the changes. That way we can make sure we don't miss some cases.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
btangmu: I see what you mean now about the third case.

btangmu:
See https://github.com/unicode-org/cldr/pull/948

PR #948, opened 2021-01-19
CLDR-14159 DAIP should trim all whitespace

btangmu:
-Revise normalizeWhitespace to call subroutines, including new trimNBSP

-Add new  enum PathSpaceType

-Add new unit test TestDisplayAndInputProcessor.TestWhitespaceNormalization

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14159


btangmu:
This doesn't include data changes such as will be needed for "value<約 >" -- I'll run CLDRModify

btangmu:
See https://github.com/unicode-org/cldr/pull/949

PR #952, opened 2021-01-21
CLDR-11581 change XOF symbol

DavidLRowe:
This pull request changes the symbol element value for XOF from `CFA` to `F&#x202f;CFA`. This includes 97 files in cldr/common/main plus 1 file in cldr/tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/xmb

NB: No changes are needed for XAF entries nor for any elements other than symbol for XOF entries.

`&#x202f;` is used for a narrow no-break space.
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11581

### DIFF
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -5861,7 +5861,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<displayName>Wes-Afrikaanse CFA-frank</displayName>
 				<displayName count="one">Wes-Afrikaanse CFA-frank</displayName>
 				<displayName count="other">Wes-Afrikaanse CFA-frank</displayName>
-				<symbol draft="contributed">CFA</symbol>
+				<symbol draft="contributed">F&#x202f;CFA</symbol>
 			</currency>
 			<currency type="XPF">
 				<displayName>CFP-frank</displayName>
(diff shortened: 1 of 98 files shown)


adrienverge:
Thanks!

DavidLRowe:
@macchiati There seem to be two problems: 
- cldr-mvn check fails because "FCFA" and "F(space)CFA" are considered the same.
- jira-ticket check fails saying the ticket has status "Done"

adrienverge:
> * jira-ticket check fails saying the ticket has status "Done"

I think the correct Jira ticket is https://unicode-org.atlassian.net/browse/CLDR-11581, not https://unicode-org.atlassian.net/browse/CLDR-11851.

DavidLRowe:
Sorry! I've updated the title and the checklist item. How do I trigger the jira-ticket check?

But the problem with "FCFA" and "F(space)CFA" is still a showstopper.

DavidLRowe:
But I see commit 2a950f9 above still has the wrong 11851 number.

macchiati:
The details button lets you fix that.

Worst case, create a new PR with the files, and then use the right commit
message.

Mark

DavidLRowe:
Will close and resubmit. Thanks for your patience!

PR #949, opened 2021-01-19
CLDR-14159 DAIP should trim all whitespace

btangmu:
-Revise normalizeWhitespace to call subroutines, including new trimNBSP

-Add new  enum PathSpaceType

-Add new unit test TestDisplayAndInputProcessor.TestWhitespaceNormalization

-Update xml with changes from running CLDRModify -fp

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14159
- [x] Updated PR title and link in previous line to include Issue number

PR #954, opened 2021-01-21
CLDR-14159 CLDRModify Pass - Happy New Year!

srl295:
CLDR-14159

CLDRModify -fp on common/main to reduce noise for future PRs such as #949


btangmu (changes requested):
Maybe we'd better fix this "Currie" bug first: https://unicode-org.atlassian.net/browse/CLDR-14428

srl295:
@btangmu could be. You own this PR so you could merge it first or not depending on what you'd like to do

btangmu:
See https://github.com/unicode-org/cldr/pull/974

PR #955, opened 2021-01-22
CLDR-14251 Delete excess } in dojo SurveyTool.serveWaitingPage

btangmu:
-This bug has been there a while; only occurs during startup

##### Checklist


srl295 (approved):
thanks

PR #959, opened 2021-01-25
CLDR-1790 make "no" the parent of "nb", "nn"

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-1790

We want to use "no" as the main version for Norwegian, with "nb" and "nn" inheriting from it. The first tasks, no matter how we do this, are:
- Remove &lt;languageAlias type="no" replacement="nb" reason="legacy"/&gt;
- Move the content from nb.xml files to no.xml files, and update the language tag in those files. Add a stub no_NO.xml file. Update default locales.

Then we have two alternative for how to go from there:
1.  Add &lt;languageAlias type="nb" replacement="no" reason="macrolanguage"/&gt;
2. Add &lt;parentLocale parent="no" locales="nb nn"/&gt;, and add stub locale files for nb.xml and nb_NO

The former is how we handle things like zh/cmn. However:
- This would lead to complete elimination of nb.xml files which I thought might be too big of a compatibility issue.
- This by itself would not solve the problem of wanting nn to inherit (for collation, formats, etc) from no, but maybe w could still add a parent relationship for that.
- We would still need to have no called Norwegian and nb called Norwegian Bokmal. We may be able to handle macrolanguage replacements in a special way for getting names, so we can get separate names for the macrolanguage and the primary child that provides the default content.

This PR is an attempt at #2. I will create a separate one for #1.


macchiati (dismissed):
Either way could have some compatibility issues, but I suspect that many clients switch nb and no already.

My preference would be to try option #1 first, and then shift to the parent locales if we see that that causes problem. But I have no problem trying it the other way first.

pedberg-icu:
> Either way could have some compatibility issues, but I suspect that many clients switch nb and no already.
> 
> My preference would be to try option #1 first, and then shift to the parent locales if we see that that causes problem. But I have no problem trying it the other way first.

OK, I will start a new PR for approach #1.

pedberg-icu:
Closing this in favor of https://github.com/unicode-org/cldr/pull/964 for https://unicode-org.atlassian.net/browse/CLDR-2698

PR #960, opened 2021-01-25
CLDR-14311 use JDK11 for build [v42]

srl295:
CLDR-14311

Bump to JDK 11.  To be landed in CLDR v42

### DIFF
--- a/.github/workflows/build-icu.yml
+++ b/.github/workflows/build-icu.yml
@@ -32,10 +32,10 @@ jobs:
           ref: ${{ github.event.inputs.cldr-ref }}
           lfs: false
           repository: ${{ github.event.inputs.cldr-repo }}
-      - name: Set up JDK 1.8
+      - name: Set up JDK
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: 11
       - name: Cache local Maven repository
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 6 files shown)


srl295:
@btangmu a JDK11 build passed, may be J9 though.

btangmu:
FYI, here's some of what I did to install  adoptopenjdk-openj9 on macOS with homebrew (homebrew must already be installed):

```
brew install jenv
brew install adoptopenjdk-openj9
jenv enable-plugin export
ls -la ~/.jenv/versions
jenv global 15
jenv versions
```

Also I added this to my `.bash_profile`:
```
# jenv -- for switching java versions with brew
export PATH="$HOME/.jenv/bin:$PATH"
eval "$(jenv init -)"
```

srl295:
probably land this in v42

.github/workflows/build-icu.yml
```diff
@@ -32,10 +32,10 @@ jobs:
           ref: ${{ github.event.inputs.cldr-ref }}
           lfs: false
           repository: ${{ github.event.inputs.cldr-repo }}
-      - name: Set up JDK 1.8
+      - name: Set up JDK 11
```
btangmu, line 35: does the name need the version number, or would "Set up JDK" do as well?

btangmu (dismissed):

.github/workflows/build-icu.yml
srl295: done

pedberg-icu:
The ticket still says v41, should that be updated?

pedberg-icu (approved):

btangmu (approved):

PR #961, opened 2021-01-25
CLDR-14422 ru dayPeriods, transition to night should be at 22:00

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14422

Transition from evening1 to night1 should be at 22:00, not 24:00/0:00


AEApple:
I'm happy to approve this, but I am concerned since there seems to be inconsistency between Russian speakers about how they refer to the hours between 10pm and 12am ([example 1](https://russian.stackexchange.com/questions/12611/telling-time-in-russian-when-does-night-morning-afternoon-and-evening-start), [example 2](https://learnrussianstepbystep.com/en/time-in-russian/), [example 3](https://www.mylanguageexchange.com/Teachers/rte/rte/LessonView.asp?Id=985)), so if we make this change I'm concerned that we will receive requests to change the dayPeriods again in the future.

macchiati (approved):

PR #963, opened 2021-01-25
CLDR-13759 copy compact currency formats to compact decimal (without ¤)

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13759

For en_IN, make the compact **decimal** formats match the compact **currency** formats (but without the ¤), this achieves the desired result.

Note that in Apple ICU, en_IN has compact currency and decimal formats equal to the result of this, except that Apple ICU en_IN uses 'K' instead of 'T' to indicate thousands (the result of some internal discussion). But the CLDR vetters chose T for the compact currency formats so in CLDR we use that for the compact decimal formats too.

### DIFF
--- a/common/main/en_IN.xml
+++ b/common/main/en_IN.xml
@@ -5203,30 +5203,29 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			</decimalFormatLength>
 			<decimalFormatLength type="short">
 				<decimalFormat>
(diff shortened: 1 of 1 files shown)


macchiati (approved):

sffc:
There was a typo in this PR that is causing integration problems with downstream clients.  Bug filed:

https://unicode-org.atlassian.net/browse/CLDR-14674

PR #964, opened 2021-01-26
CLDR-2698 make no the macrolanguage replacement for nb

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-2698

This ia an alternative to the incomplete [PR-959](https://github.com/unicode-org/cldr/pull/959/) as a way to have "no" as the main language code for Norwegian (with content from "nb"), and with "nn" inherting certain items from it (e.g. formats, collation). We do still want to be able to get a name for "nb" that is distinct from the name for "no".

The steps here are:
- Remove &lt;languageAlias type="no" replacement="nb" reason="legacy"/&gt;
- Replace it with &lt;languageAlias type="nb" replacement="no" reason="macrolanguage"/&gt; i.e. handle this in the same way that we handle zh/cmn, etc.
- Move the content from nb.xml files to no.xml files, and update the language tag in those files. Update default locales.
- Add &lt;parentLocale parent="no" locales="nn"/&gt;
- Update validity, coverage, etc.
- Update tooling.

Design outline here: https://docs.google.com/document/d/1Hpnhc5CSchlOePakScMd7xL6UHX_GicKuW5KAyGNirs/edit#


tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateTime.java
```diff
@@ -144,7 +144,7 @@ public static boolean isGregorianPattern(String path) {
             }
             // special hack for OpenOffice
             territoryAlias.put("CB", "029");
-            languageAlias.put("no", "nb");
+            languageAlias.put("no", "nb"); // this may no longer be needed with this PR
```
macchiati, line 147: Yes, I think this should be taken out.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ExtractMessages.java
```diff
@@ -271,7 +271,7 @@ public String getLocale() {
         { "in", "id" },
         { "iw", "he" },
         { "jw", "jv" },
-        { "no", "nb" },
+        { "nb", "no" },
```
macchiati, line 274: This also just gets dropped.

macchiati (commented):
A couple of minor items.

macchiati (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateTime.java
pedberg-icu: Thanks, I did that

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ExtractMessages.java
pedberg-icu: Thanks, I did that too

macchiati (approved):

pedberg-icu:
OK, I fixed the last test failure with a little hack in TestCLDRLocaleCoverage.java. This needs re-approval please, thanks.

macchiati (approved):

PR #966, opened 2021-01-26
CLDR-14399 change browse.jsp to use API (CLDR-14388)

srl295:
- see CLDR-14388
- clean up API definitions
- introduce STError API object

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAPI.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAPI.java
@@ -2,26 +2,127 @@ package org.unicode.cldr.web.api;
 
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/src/main/webapp/browse.jsp
```diff
@@ -59,34 +59,72 @@ SurveyTool.includeJavaScript(request, out);
...
+			                	    }
+			                	    if(something != null) {
+			                	    	something.value = txt;
+		//	                	        something.innerHTML = txt;  // TODO shold only use for plain text
+			                	    }
+			                	}
+		
+		                          updateIf('xpathid', 	h.decimalId);
+		                          updateIf('xpath', 	h.xpath);
+		                          updateIf('strid', 	h.hexId);
```
btangmu, line 88: Of course this js-inside-jsp is very temporary and will soon migrate to a js file -- and then should be heavily refactored to avoid functions that are long or deeply nested

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAPI.java
```diff
@@ -2,26 +2,127 @@
...
+        if(xpath == null) {
+            return Response.status(Response.Status.NOT_FOUND)
+                .entity(new STError("XPath #" + decId + " not found"))
+                .build();
+        }
+        return Response.ok(new XPathInfo()
+            .setDecimalId(decId)
+            .setHexId(xpt.getStringIDString(decId))
+            .setXpath(xpath)).build();
+    }
```
btangmu, line 127: This is awesome! A big difference from r.put -- maybe we'll still have some use for r.put -- anyway it's going to take a while to transition

btangmu (approved):
This is some serious modernization

PR #967, opened 2021-01-26
CLDR-14399 minor nginx cleanup

srl295:
- remove unneeded transform from nginx (openapi is not the surveytool)

CLDR-14399

this was the only nit I found so far…


btangmu (approved):
right, we only use the rewrite for cldr-apps, not for openapi -- might there ever be a reason to serve js or css for openapi requests?

PR #968, opened 2021-01-26
CLDR-13485 Fix inconsistent use of MU by always using Greek character.

macchiati:
NOTE: also modifies CldrModify to skip files that have not changed except for copyright.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13485

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -2454,9 +2454,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="America/Santa_Isabel">
 				<exemplarCity>Santa Isabel</exemplarCity>
 			</zone>
-			<zone type="Australia/Currie">
-				<exemplarCity>Currie</exemplarCity>
-			</zone>
 			<zone type="Pacific/Honolulu">
 				<exemplarCity>Honolulu</exemplarCity>
 			</zone>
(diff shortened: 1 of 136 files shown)


btangmu:
As far as the data is concerned, should we try to fix the Currie problem first? I'm working on that: https://unicode-org.atlassian.net/browse/CLDR-14428

macchiati:
If you can finish quickly...

btangmu:
> If you can finish quickly...

OK -- note that this PR would move Australia/Currie between America/Santa_Isabel and Pacific/Honolulu, which is completely wrong, while in master it's correctly between Broken_Hill and Melbourne -- so I aim to preserve the correct position with CLDR-14428

macchiati:
The end result should be the right position. However, we don't want to
block any tickets for this, because there is no semantic difference in the
XML files based on those two positions. So in the worst case scenario, we
can fix the position of Australia/Currie *after the data freeze*.

Mark

btangmu:
Now that https://github.com/unicode-org/cldr/pull/970 is merged maybe CLDRModify should be run again to update the data in this PR before merging it

srl295:
> Now that https://github.com/unicode-org/cldr/pull/970 is merged maybe CLDRModify should be run again to update the data in this PR before merging it

It would be helpful to rerun

srl295 (approved):
LGTM and RSLGTM on data

srl295:
@macchiati looks like this also does CLDR-11394

srl295 (dismissed):

srl295 (approved):

btangmu (approved):
Mark, probably best if this is merged before the DAIP whitespace one?

PR #969, opened 2021-01-26
CLDR-14434 tzdata 2021a updates

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14434


markusicu (approved):
lgtm

PR #970, opened 2021-01-26
CLDR-14428 ZoneParser fails for Australia/Currie

btangmu:
-Hard-code data for Australia/Currie in ZoneParser.java

-Also work around a possible compiler bug affecting TestSupplementalInfo.LOCALES_FIXED

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14428

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ZoneParser.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ZoneParser.java
@@ -550,13 +550,8 @@ public class ZoneParser {
 
         @Override
         public int compare(String s1, String s2) {
-            List<String> data1 = data.get(s1);
-            if (data1 == null)
-                data1 = errorData;
-            List<String> data2 = data.get(s2);
-            if (data2 == null)
-                data2 = errorData;
-
+            List<String> data1 = getData(s1);
+            List<String> data2 = getData(s2);
             int result;
             // country
             String country1 = data1.get(2);
@@ -576,6 +571,30 @@ public class ZoneParser {
                 return result;
             // name
             return s1.compareTo(s2);
+        }
+
+        /**
+         * Get timezone data for the given location
+         * Include work-arounds for missing time zones
+         *
+         * @param s the string like "Australia/Currie"
+         * @return a list of 4 strings for latitude, longitude, country, city
+         *
+         * Reference: https://unicode-org.atlassian.net/browse/CLDR-14428
+         */
+        private List<String> getData(String s) {
+            List<String> d = data.get(s);
+            if (d == null) {
+                if ("Australia/Currie".equals(s)) {
+                    // 39°55′52″S 143°51′02″E per https://en.wikipedia.org/wiki/Currie,_Tasmania
+                    // Converted with https://www.latlong.net/degrees-minutes-seconds-to-decimal-degrees
+                    d = Arrays.asList("-39.93", "143.85", "AU", "Currie");
+                } else {
+                    // TODO: "America/Santa_Isabel", "Pacific/Honolulu"?
+                    d = errorData;
+                }
+            }
+            return d;
         }
     };
 
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
@@ -1866,7 +1866,8 @@ public class TestSupplementalInfo extends TestFmwkPlus {
     }
 
     public void Test9924() {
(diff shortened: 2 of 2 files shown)


btangmu:
When I run CLDRModify -fp the only changes are for copyright year, no changes for Australia/Currie; that's the intention

srl295:
@macchiati @btangmu when i merge this change into the JSON generator (using cldr-staging 38.1. data) it causes Currie to be deleted in a lot of locales, is this expected?

```diff
diff --git a/cldr-json/cldr-dates-full/main/az-Cyrl/timeZoneNames.json b/cldr-json/cldr-dates-full/main/az-Cyrl/timeZoneNames.json
index db907f21..0712b346 100644
--- a/cldr-json/cldr-dates-full/main/az-Cyrl/timeZoneNames.json
+++ b/cldr-json/cldr-dates-full/main/az-Cyrl/timeZoneNames.json
@@ -1137,9 +1137,6 @@
               "Broken_Hill": {
                 "exemplarCity": "Broken Hill"
               },
-              "Currie": {
-                "exemplarCity": "Currie"
-              },
               "Darwin": {
                 "exemplarCity": "Darwin"
               },
```

btangmu:
> @macchiati @btangmu when i merge this change into the JSON generator for #972 (using cldr-staging 38.1. data) it causes Currie to be deleted in a lot of locales, is this expected?

No, I would only expect it to affect the collation of Currie, not its presence/absence

macchiati:
Sounds like something is broken.

srl295:
OK. I can dig into it some more. There's no reason to hold the CLDR JSON build for it, as it is easy enough to just revert all `timeZoneNames.json` files (which otherwise didn't change).

btangmu:
az-Cyrl.xml doesn't contain Currie or any zone elements. It presumably inherits them from az.xml, which does have zone elements including Currie. The json files, unlike the xml, explicitly include inherited values? Is there a spec/outline for how the json and xml are related and how/why they differ?

srl295:
@btangmu that is complex…   

What i'm going to do is to split out the changes, and work on the Currie-in-JSON thing as a separate PR

btangmu:
> @macchiati @btangmu @yumaoka
> I ran a bisect to find out where Currie was disappearing. I found that [d909e3d#diff-76dbe60a9eb9c51233a3174cd4b43ef9a0fe25887970f1460e3206e3412a9423R52](https://github.com/unicode-org/cldr/commit/d909e3deed3a68c217d1c19390df959c85f365fd#diff-76dbe60a9eb9c51233a3174cd4b43ef9a0fe25887970f1460e3206e3412a9423R52) (tz 2020f) shows that Australia/Currie actually _is_ removed, since it's now a link to Hobart.
> 
> So my question is: is this a bug? Perhaps we shouldn't "Hard-code data for Australia/Currie in ZoneParser.java" if the zone is no longer in modern use on its own? Or at least, we should be using the linked data and not hard coding data?

@srl295 We just merged this last night: https://github.com/unicode-org/cldr/pull/988 -- I think that answers part of your question (yes: shouldn't hard-code...)

As for "is this a bug [or a feature]", that's sort of a policy question? Mark has suggested in the follow-up ticket that CLDR translations shouldn't be allowed for deprecated zones -- so removal of Currie seems appropriate but I don't know if there are any negative consequences

macchiati:
I think our goal for this release should be to make the minimal changes to
data so that we can hit the freeze.

I found more issues than just Currie, so I wouldn't just single that
particular item out right now (eg don't remove the current translations
until we have a full design, and do it under the other ticket).

Mark

srl295:
So here's where things stand… 

I wrote a test, and found that there are no time zones that don't have a country. 

Do you want a test that verifies that Currie is present in the list of tz names?￼￼ I'm guessing it's missing in ST also but didn't check yet.

macchiati:
Great. By "no timezones" I assume you mean "no valid zones", according to
bcp47/timezone.xml, which is what we use for validity (though it needs some
work, as per https://unicode-org.atlassian.net/browse/CLDR-14453

You shouldn't need to write a test that the timezone names used in the
following (for example) are valid, since that is done by
TestAttributeValues.java.

<zone type=*"Australia/Currie"*>

<exemplarCity>Currie</exemplarCity>

</zone>

It uses the data in ldml.dtd, which then uses MatchValue

<!ATTLIST zone type CDATA #REQUIRED >
    

Specifically Bcp47MatchValue, which tests for either the name or the
aliases.

I introduced a bug in a name and got the expected failure:

Status DtdType Element Attribute Match expression #Failures Failing values
invalid ldml zone type bcp47/tz 2 Australia/Curry, Pacific/Honolululu

What it doesn't yet test for is the deprecated flag or for non-standard
aliases.

I added that to the items to do in
https://unicode-org.atlassian.net/browse/CLDR-14453?focusedCommentId=159945

Mark

PR #971, opened 2021-01-26
CLDR-14097 Added values to handle road preferences better

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14097


hugovdm (approved):
Neat!

srl295 (approved):

PR #975, opened 2021-01-26
CLDR-11581 update XOF

DavidLRowe:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-11581


macchiati:
Made extra PR to allow 202F: https://github.com/unicode-org/cldr/pull/993. Once that goes through, you could rebase to allow your tests to work.

DavidLRowe:
@macchiati I attempted my first-ever rebase. I now have a bunch of files showing changes I didn't make. Is that normal? (That is, are these the changes that would have been in master if I'd started now.) Or have I done something wrong?

macchiati:
Dave, it looks like it has gotten gummed up (that can happen when there are many intervening changes). I think the simplest option at this point is:

1. Revert to master, 
2. create a new, different, branch, 
3. do the bulk replace that you did earlier to get from ">CFA</symbol>" to ">F&#x202f;CFA</symbol>"
4. check in new PR.
5. close this PR

DavidLRowe:
Replaced by #994

PR #978, opened 2021-01-27
CLDR-10831 megabyte gigabyte fa

Kherishma:
Updated the fa form of gigabyte and megabyte to the non-localized version

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-10831


pedberg-icu (dismissed):

pedberg-icu (commented):
The short form *already* uses GB/MB. This PR changes the long form, which is usually the more textual form; is that really what we want to do?

Kherishma:
Hi Peter, 

Do all the changes look correct? If not, I can close this pull request. 

Thanks, 
Kherishma

PR #980, opened 2021-01-27
CLDR-10831 megabyte gigabyte ka

Kherishma:
Updated the ka form of gigabyte and megabyte to the non-localized version

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-10831


pedberg-icu (changes requested):
This only changes the long form, but not the short form. If anything, I would think it the short form that most needs changing.

Kherishma (commented):
Hi Peter, 

I've now only updated the short forms of GB and MB. Let me know if everything looks good. 

Thank you, 
Kherishma

PR #983, opened 2021-01-27
CLDR-14232 fix MapComparator to not regress w/ 38.1

srl295:
CLDR-14232

Without the DtdData change, CLDR 39 is unable to read CLDR 38.1 staging data,
CLDRFile.iterator() crashes with "Missing Map Comparator Values … concentr-milligram-per-deciliter…"

PR #986, opened 2021-01-27
CLDR-14232 add deprecated units to the DtdData comparator

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14232


srl295 (approved):
obviates #983 

LGTM… is there an issue for making the array data (or metadata) driven? would that have helped solve this?

macchiati:
> obviates #983
> 
> LGTM… is there an issue for making the array data (or metadata) driven? would that have helped solve this?

I filed https://unicode-org.atlassian.net/browse/CLDR-14452 about that.

PR #988, opened 2021-01-27
CLDR-14428 ZoneParser failures

btangmu:
-Use the existing linkold_new in getData in ZoneParser.java

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14428
- [x] Updated PR title and link in previous line to include Issue number

PR #989, opened 2021-01-27
CLDR-14428 ZoneParser failures; use linkold_new, remove SKIP_LINKS

btangmu:
-Use the existing linkold_new in getData in ZoneParser.java

-Remove SKIP_LINKS, so that Santa_Isabel is found in linkold_new and collates with Tijuana

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14428


btangmu:
See https://github.com/unicode-org/cldr/pull/988

PR #992, opened 2021-01-28
CLDR-13890 world

Kherishma:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13890

commits in the same PR
ALLOW_MANY_COMMITS


srl295 (approved):
needs squash otherwise lgtm

PR #993, opened 2021-01-28
CLDR-11581 Allow U+202F to distinguish collisions

macchiati:
The goal is to allow 202F to distinguish collisions. 
- fix the set of characters that are distinguished to add 202F (also removed some CO controls and reformatted for clarity
- add hack to get around the NFKC normalization for this character. 

CLDR-11581

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleXMLSource.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleXMLSource.java
@@ -168,7 +168,7 @@ public class SimpleXMLSource extends XMLSource {
     static final Normalizer2 NFKC = Normalizer2.getNFKCInstance();
 
     // The following includes letters, marks, numbers, currencies, and *selected* symbols/punctuation
-    static final UnicodeSet NON_ALPHANUM = new UnicodeSet("[^[:L:][:M:][:N:][:Sc:][\\u202F\uFFFF _ ¡ « ( ) \\[ \\] \\{ \\} § / \\\\ % ٪ ‰ ؉ ‱-″ ` \\^ ¯ ¨ ° + ¬ | ¦ ~ − ⊕ ⍰ ☉ © ®]]").freeze();
+    static final UnicodeSet NON_ALPHANUM = new UnicodeSet("[^[:L:][:M:][:N:][:Sc:][\\u202F\uFFFF _ ¡ « ( ) \\- \\[ \\] \\{ \\} § / \\\\ % ٪ ‰ ؉ ‱-″ ` \\^ ¯ ¨ ° + ¬ | ¦ ~ − ⊕ ⍰ ☉ © ®]]").freeze();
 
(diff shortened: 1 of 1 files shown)


srl295:
ton of collision errors in the build log- something wrong

tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleXMLSource.java
```diff
@@ -168,14 +168,23 @@ public void getPathsWithValue(String valueToMatch, String pathPrefix, Set<String
     static final Normalizer2 NFKC = Normalizer2.getNFKCInstance();
 
     // The following includes letters, marks, numbers, currencies, and *selected* symbols/punctuation
-    static final UnicodeSet NON_ALPHANUM = new UnicodeSet("[^[:L:][:M:][:N:][:Sc:]/+\\-°′″%‰‱٪؉−⍰()⊕☉]").freeze();
+    static final UnicodeSet NON_ALPHANUM = new UnicodeSet("[^[:L:][:M:][:N:][:Sc:][\\u202F\uFFFF _ ¡ « ( ) \\[ \\] \\{ \\} § / \\\\ % ٪ ‰ ؉ ‱-″ ` \\^ ¯ ¨ ° + ¬ | ¦ ~ − ⊕ ⍰ ☉ © ®]]").freeze();
```
srl295: would be good to document here why U+202F is there

macchiati:
Right, fixing right now.

Mark

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleXMLSource.java
```diff
@@ -184,7 +184,7 @@ public static String normalize2(String valueToMatch, Normalizer2 normalizer2) {
             String result = replace(NON_ALPHANUM, normalizer2.normalize(temp), "");
             return result.replace('\uFFFF','\u202F');
         }
-        return replace(NON_ALPHANUM, normalizer2.normalize(valueToMatch.replace('\u202F', '\u00A0')), "");
+        return replace(NON_ALPHANUM, normalizer2.normalize(valueToMatch), "");
```
srl295, line 187: How did this have an effect? Is `java.lang.String.replace(char)` not  non-BMP safe?

I guess it wasn't 𒀯 (U+1202F).

srl295 (approved):
LGTM if tests pass

DavidLRowe:
@srl295 Is this ready to merge? #975 depends on this PR.

DavidLRowe:
Thanks!

tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleXMLSource.java
macchiati: BMP characters never overlap with high or low surrogates; so not a problem.

PR #998, opened 2021-02-02
CLDR-14462 Add missing mapping for America/Nuuk

mj1856:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14462


mj1856:
Nevermind.  This is invalid.  See my comment in the issue.

PR #1000, opened 2021-02-02
CLDR-14386 Modernize List Users in gear menu

btangmu:
-Add new UserList.java for back end

-Add new cldrListUsers.js and cldrAccount.js (cf. CLDR-14383) for front end

-Remove dead code

-Comments

##### Checklist

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -22,6 +22,8 @@ import java.util.Set;
 import java.util.TreeSet;
 
 import javax.servlet.ServletException;
+import javax.servlet.annotation.MultipartConfig;
+import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -83,7 +85,11 @@ import com.ibm.icu.util.Output;
  * also use a POST instead of the _v parameter ) Note, add the preference to the
  * settablePrefsList
  *
+ * The @MultipartConfig annotation enables getting request parameters from a POST request
+ * that has "multipart/form-data" as its content-type, as needed for WHAT_USER_LIST.
  */
+@WebServlet
+@MultipartConfig
 public class SurveyAjax extends HttpServlet {
     final boolean DEBUG = false; //  || SurveyLog.isDebug();
     public final static String WHAT_MY_LOCALES = "mylocales";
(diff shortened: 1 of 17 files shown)


btangmu:
This is far from finished. I suggest merging it with master anyway since it should cause no problems while USE_DOJO is still true, and the changes are extensive enough that we should merge to avoid merge conflicts later.

srl295:
PR number 1,000 ! 🎉

tools/cldr-apps/src/main/webapp/js/new/cldrAjax.js
```diff
@@ -8,7 +8,7 @@
  * and to hide everything else, minimizing global scope pollution.
  */
 const cldrAjax = (function () {
-  const ST_AJAX_DEBUG = false;
+  const ST_AJAX_DEBUG = true;
```
srl295, line 11: should this stay on by default?

tools/cldr-apps/src/main/webapp/js/new/cldrGear.js
```diff
@@ -6,6 +6,13 @@
  *
  * Use an IIFE pattern to create a namespace for the public functions,
  * and to hide everything else, minimizing global scope pollution.
+ *
+ * TODO: "Avoid dropdown menus": "Usability studies show that dropdown menus are annoying";
+ * "difficult for search engines to crawl" https://www.orbitmedia.com/blog/website-navigation/
+ * Survey Tool's current dropdown gear menu is also problematic in terms of library dependencies,
+ * invalid html, inability to scroll (e.g., when console log is open and screen is small),
+ * and failure to close immediately when some items are clicked on. It would be cleaner to
+ * open a new special page ("#gear"), with a list of items, when the gear menu is clicked on.
```
srl295, line 15: @btangmu could be good, and now that the gear is a module it should be simple..

tools/cldr-apps/src/main/webapp/js/special/users.js
```diff
@@ -69,7 +69,7 @@ define("js/special/users.js", ["js/special/SpecialPage.js"], function (
         var locked = [];
 
         var byEmail = {}; // email -> u:{}
-        var byId = {}; // id -> u:{}
+        var byId = {}; // id -> u:{} -- UNUSED!
```
srl295, line 72: drop if unused

srl295 (commented):
still reviewing…

tools/cldr-apps/src/main/webapp/js/special/users.js
btangmu: that whole file will be unused when we drop dojo -- I just added the comment as a reminder during the transition

tools/cldr-apps/src/main/webapp/js/new/cldrAjax.js
btangmu: it makes the console a bit noisy, so maybe we should just turn it on and off as seems most useful; I didn't notice I had switched it in this PR

tools/cldr-apps/src/main/webapp/js/new/cldrGear.js
btangmu: yes

srl295 (approved):
RSLGTM. would be good to start using the new API mechanisms.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -2861,6 +2880,9 @@ public static void getRow(HttpServletRequest request, HttpServletResponse respon
                 }
 
                 try {
+                    /*
+                     * TODO: why use org.json.JSONWriter here but SurveyAjax.JSONWriter elsewhere?
+                     */
```
btangmu, line 2885: Comment outdated by https://github.com/unicode-org/cldr/pull/1002 -- I'll fix it later

PR #1002, opened 2021-02-02
CLDR-14399 move SurveyAjax.JSONWriter to SurveyJSONWrapper

srl295:
CLDR-14399


btangmu (approved):
good choice of name, and great that it's moved to its own file

srl295:
> good choice of name, and great that it's moved to its own file

I didn't want to do too much improvement on it. but i think this is a good step and will serve well.

PR #1005, opened 2021-02-03
CLDR-14212 fix additional code mentioning ant

srl295:
CLDR-14212

- also bump the `ant` version used in tools


tools/pom.xml
```diff
@@ -78,9 +78,9 @@
 			</dependency>
 
 			<dependency>
-				<groupId>org.apache.ant</groupId>
-				<artifactId>ant</artifactId>
-				<version>1.10.8</version>
+			  <groupId>org.apache.ant</groupId>
+			  <artifactId>ant</artifactId>
+			  <version>[1.10.9,)</version>
```
PeterConstable, line 83: Curious version notation. I haven't seen this before. Is this saying something like, 'compatible with 1.10.9 and up to but not including some not-yet-specified version'?

srl295: good point. it's "≥1.10.9" but it'd be better to just pick a specific version.

PeterConstable (approved):

btangmu (approved):

PR #1008, opened 2021-02-03
CLDR-14054 minimumdigitgrouping for es-419, es-US & es-MX 

AEApple:
Update minimumdigitgrouping for es-419, es-US & es-MX per language specialist feedback that Spanish in Latin America uses minimum digit grouping of 1.

I have updated the two other Latin America Spanish variants since our language specialist confirmed that minimum grouping should be 1 for all Latin American Spanish locales, but let me know if you are concerned and I can update the PR. I will update the ticket with sources for each country to demonstrate usage.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14054

### DIFF
--- a/common/main/es_419.xml
+++ b/common/main/es_419.xml
@@ -5052,7 +5052,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<otherNumberingSystems>
 			<native draft="contributed">latn</native>
 		</otherNumberingSystems>
-		<minimumGroupingDigits draft="contributed">2</minimumGroupingDigits>
+		<minimumGroupingDigits draft="contributed">1</minimumGroupingDigits>
 		<symbols numberSystem="latn">
 			<decimal>.</decimal>
 			<group>,</group>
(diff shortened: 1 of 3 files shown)


stenshamn:
Technically you could have removed the entry entirely in es_MX and es_US and it would have inherited from es_419, but I think this works too. @srl295 , what do you think?

pedberg-icu (approved):

AEApple:
The Maven checks are failing. How do I resolve this?

pedberg-icu:
> The Maven checks are failing. How do I resolve this?

This needs to be (1) rebased onto latest master to pick up a fix from another recent PR, and (2) squashed. The fastest way to do this for now is just to create another PR based on latest master. I have done this in https://github.com/unicode-org/cldr/pull/1009 so I will close this one and we can use that one instead.

pedberg-icu:
> Technically you could have removed the entry entirely in es_MX and es_US and it would have inherited from es_419, but I think this works too. @srl295 , what do you think?

It is fine to leave them here for now, they will likely be removed when we do locale minimization

PR #1012, opened 2021-02-05
CLDR-14400 add webpack

srl295:
CLDR-14400 — enjoy,  @btangmu !  🎁  
- add webpack using Node.js and frontend-maven-plugin
- does not do much yet, but adds virtual path `/dist/bundle.js`
- show hello message in `about.jsp` to verify function


srl295:
I'd like to get this in. It doesn't actually benefit any existing code/pages, but we can easily migrate any modules/etc. into this bundle.

On running `mvn package`, Node.js is automatically invoked to do whatever packaging steps are needed, and then generated artifacts are put into `cldr-apps/src/main/webapp/dist/` ( which is added to .gitignore). 

You can run `npm run build` from the `cldr-apps/src/js/` directory at any time to re-run the bundler.

tools/cldr-apps/js/package-lock.json
```diff
@@ -0,0 +1,1266 @@
...
+      "dev": true
+    },
+    "yocto-queue": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/yocto-queue/-/yocto-queue-0.1.0.tgz",
+      "integrity": "sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q==",
+      "dev": true
+    }
+  }
+}
```
btangmu, line 1266: That's a big file! How is it generated/maintained?

tools/cldr-apps/src/main/webapp/about.jsp
```diff
@@ -117,5 +117,8 @@
 		<a href="./survey">Return to Survey Tool</a> |
 		<a href="./index.jsp">Return to CLDR Applications</a> |
 		<a target="_new" href="http://dev.icu-project.org/cgi-bin/cldrwiki.pl?SurveyToolHelp/About">Help on this page (in a new window)</a>
+
+		<!--  load bundle.js from webpack -->
+		<script src="./dist/bundle.js"></script>
 	</body>
```
btangmu, line 123: What about the modern "About" code in js/new/cldrAbout.js? Have this PR been tested with USE_DOJO_DEFAULT false?

tools/cldr-apps/pom.xml
```diff
@@ -233,6 +233,50 @@
...
+					</execution>
+					<execution>
+						<id>npm run lint</id>
+						<goals>
+							<goal>npm</goal>
+						</goals>
+						<configuration>
+							<arguments>run lint</arguments>
+					</configuration>
+				</execution>
```
btangmu, line 272: Indentation is irregular

Lucky22081991 (dismissed):

tools/cldr-apps/js/package-lock.json
srl295: It's completely managed by `npm`, which updates it whenever anything is added or removed. It specifies the exact hash of the packages being used. https://docs.npmjs.com/cli/v6/configuring-npm/package-locks

tools/cldr-apps/src/main/webapp/about.jsp
srl295: I have tested it with (and use it with) USE_DOJO_DEFAULT, this was just a place to cause the script to be loaded.  This jsp is still usable even with USE_DOJO_DEFAULT=false, all the better since this jsp will be going away.

tools/cldr-apps/pom.xml
srl295: Will fix

srl295: @btangmu fixed https://github.com/unicode-org/cldr/compare/unicode-org:821dcb5..unicode-org:a70effe

btangmu (approved):

tools/cldr-apps/src/main/webapp/about.jsp
btangmu: Shall we move js/new/* to js/src/*?

I had been thinking we'd move js/new/* to js/* as soon as we removed the dojo versions, but now we could move new to src, right? And see what happens with dist/bundle. Anyway can keep running directly from src if the bundle doesn't work yet

srl295: > Shall we move js/new/* to js/src/*?

eventually, yes. I would probably do that component by component.  I was going to add Vue and such to the bundle in a followon PR

srl295:
@btangmu https://cldr-smoke.unicode.org/cldr-apps/about.jsp working in production (red letters at bottom)

tools/cldr-apps/src/main/webapp/about.jsp
btangmu: Actually I didn't notice until now that this PR created a new js folder

* tools/cldr-apps/js

distinct from pre-existing

* tools/cldr-apps/src/main/webapp/js

So our current js files will migrate from

* tools/cldr-apps/src/main/webapp/js/new

to

* tools/cldr-apps/js/src

And the bundle will be created at

* tools/cldr-apps/src/main/webapp/dist

Will there then be no more tools/cldr-apps/src/main/webapp/js at all?

btangmu:
> You can run `npm run build` from the `cldr-apps/src/js/` directory at any time to re-run the bundler.

Actually `cldr-apps/js/src/`

tools/cldr-apps/src/main/webapp/about.jsp
srl295: > Will there then be no more tools/cldr-apps/src/main/webapp/js at all?
 
Eventually. But we need to make sure that components can actually function in that environment.

PR #1017, opened 2021-02-09
CLDR-14400 Vue.js aboutbox, debug webpack

srl295:
CLDR-14400
- run prettier on webpack config
- set webpack to debug mode (for now!)
- add a Vue-based about to the ST about page
- change cldrGui to not wipe out the entire <body>


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -137,7 +137,7 @@ private void serveWaitingPage(HttpServletRequest request, PrintWriter out, Surve
          * the front end could be in charge of the JavaScript.
          */
         String url = request.getContextPath() + request.getServletPath();
-        out.write("<html lang='" + SurveyMain.TRANS_HINT_LOCALE.toLanguageTag() + " class='claro'>\n");
+        out.write("<html lang='" + SurveyMain.TRANS_HINT_LOCALE.toLanguageTag() + "' class='claro'>\n");
```
srl295, line 140: @btangmu this was an old bug, causes an invalid HTML page during loading

srl295:
no real substance in the aboutbox yet. again, an incremental step, of getting the loader in place.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
btangmu: good catch!

tools/cldr-apps/js/src/views/About.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+    <div id="home">
```
btangmu, line 2: one small file, one giant leap for Survey Tool

tools/cldr-apps/js/webpack.config.js
```diff
@@ -1,17 +1,28 @@
-const path = require('path');
+const path = require("path");
+const { VueLoaderPlugin } = require("vue-loader");
```
btangmu, line 2: just curious, would `import` instead of `require` work here?

btangmu (approved):

tools/cldr-apps/js/webpack.config.js
srl295: @btangmu not sure

PR #1020, opened 2021-02-10
CLDR-14485 add top level maven project

srl295:
CLDR-14485
It is optional to use, and for the convenience of Eclipse use

-----

- so with this change, i just run the same maven import at the /cldr level. 
 
ImagenPegada-1

- Seems to work for editing in Eclipse…

ImagenPegada-2

Command line building from the tools directory down is unaffected, no files changed there. (I’m not going to rev all of the docs, but `—file=tools/pom.xml` is now optional.)

I don’t want to *require* building from / using the new top level directory, that would break workspaces, build scripts, processes, etc.  for no real benefit at this point. 

You can build from the top level:

```shell
~/src/cldr:$ mvn package -DskipTests=true
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO] 
[INFO] CLDR All Tools                                                     [pom]
[INFO] CLDR Java Tools                                                    [jar]
[INFO] CLDR RDF Tools                                                     [jar]
[INFO] CLDR Survey Tool                                                   [war]
[INFO] CLDR Top Level and Data                                            [pom]
```

- I named `/pom.xml` **cldr-data** and call it “CLDR Top Level and Data”

- the `/tools/pom.xml` is still called **cldr-all** but I changed the description to “CLDR All Tools” (since it is no longer the top level).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .DS_Store
 
 # Maven output directories
+/target
 /tools/target
 /tools/*/target
 
@@ -84,3 +85,5 @@
 
 
 org.eclipse.wst.common.project.facet.core.xml
+/.settings
+/tools/cldr-rdf/.settings/org.eclipse.m2e.core.prefs
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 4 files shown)


macchiati (commented):
I don't see where the specific directory folders are listed like seed and common and so forth. Could you let me know how this is working.

srl295:
> I don't see where the specific directory folders are listed like seed and common and so forth. Could you let me know how this is working.

Certainly.  Eclipse adds everything under the top level pom.xml 

If you look under the current "cldr-all" you will see other files in there, such as scripts, c, python etc. which are also not listed explicitly anywhere. This is the same , just at one directory level higher.

macchiati (approved):

macchiati:
Ok.

Mark

srl295:
Thanks, give it a try and let me know how it goes￼

PR #1022, opened 2021-02-10
CLDR-14444 tr35-collation markdown changes

DavidLRowe:
In section 2.5, there's a nested bulleted list. The sentence "An implementation..." should not be part of the inner bulleted list level.

Other cleanup changes:
- In section 1.1.5, two spurious empty cells removed from the first line of the table showing Danish and emoji ordering.
- In sections 3 and 3.1, the `<span class="element">...</span>` formatting was changed to use backquotes in three locations (`collation` twice, and `defaultCollation` once).
- Various "note" type items (from elements using class="note") standardized to begin with: `> :point_right: **Note**:`

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14444

### DIFF
--- a/docs/ldml/tr35-collation.md
+++ b/docs/ldml/tr35-collation.md
@@ -112,7 +112,7 @@ For example, when sorting names in a database, a sortable string can be formed w
 
 For backwards secondary level sorting, text _segments_ separated by U+FFFE are processed in forward segment order, and _within_ each segment the secondary weights are compared backwards. This is so that such combined strings are processed consistently with merging their sort keys (for example, by concatenating them level by level with a low separator).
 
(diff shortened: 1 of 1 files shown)


macchiati:
Ah, I think you can't set Reviewers, right?

Mark

DavidLRowe:
I don't seem to be able to set reviewers (nor would I really know whom to assign).

DavidLRowe:
I see that there are a few other whitespace changes that were introduced in the editing process.

macchiati (approved):
Looks great, thanks David

macchiati:
I'd approved and merged

PR #1023, opened 2021-02-10
CLDR-14491 BRS Task A05.5 : CLDRModify pass 1 common/main

btangmu:
-CLDRModify with no args

-Changed ordering of timezones Currie, Honolulu

-Changed hex code for U+202F NARROW NO-BREAK SPACE to UTF-8

-Changed copyright year 2020 to 2021

##### Checklist

### DIFF
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -2454,9 +2454,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="America/Santa_Isabel">
 				<exemplarCity>Santa Isabel</exemplarCity>
 			</zone>
-			<zone type="Pacific/Honolulu">
-				<exemplarCity>Honolulu</exemplarCity>
-			</zone>
 			<zone type="Etc/UTC">
 				<long>
 					<standard>Gekoördineerde universele tyd</standard>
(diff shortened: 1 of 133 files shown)


common/main/af.xml
```diff
@@ -2576,12 +2573,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
 			</zone>
-			<zone type="Australia/Currie">
-				<exemplarCity>Currie</exemplarCity>
-			</zone>
 			<zone type="Australia/Melbourne">
 				<exemplarCity>Melbourne</exemplarCity>
 			</zone>
+			<zone type="Australia/Currie">
+				<exemplarCity>Currie</exemplarCity>
+			</zone>
```
srl295, line 2581: is this expected?

I thought the previous fixes would have included any such change.

srl295 (approved):
I'm puzzled by the Currie changes, given https://github.com/unicode-org/cldr/pull/970 

@macchiati ? 

If this is expected, then fine…

macchiati (changes requested):
We need to modify the Daily code to keep the nnbsp. (Narrow no break space)

macchiati (approved):
At first I was worried that maybe F&#x202f;CFA was converting to F<space>CFA, but I checked and it's ok.

About Currie, I'm not worried about the position.

macchiati (commented):
BTW, I also thought we'd done a "clean" run of cldr modify just to fix the copyrights, but apparently not.

common/main/af.xml
btangmu: We ran CLDRModify -fp when testing the effect of the time zone changes — now this is without -fp — maybe that’s why we’re seeing additional locales affected

macchiati:
That's it!

PR #1025, opened 2021-02-11
CLDR-14476 fa, remove explicit <LRM> or replace with \u200E; ko, remove a bogus <

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14476

In fa, which has a bunch of ASCII &lt;LRM&gt; which were probably intended to be actual \u200E characters:
- If the  &lt;LRM&gt; immediately precedes a strong L character (ASCII letter or µ), then just remove it, it would have zero effect if converted to \u200E
- In the other cases, convert the &lt;LRM&gt; to an actual  \u200E character. Most of these are before a / in a per pattern, which makes sense. Two of them precede a space which is followed by a strong L character; these are probably mistakes but they would have an effect in a R-L context so I did convert them.

In ko, there was one bogus &lt; in Japanese era name, removed it.

I will split off a separate ticket to add tests for a future Survey Tool phase.

### DIFF
--- a/common/main/fa.xml
+++ b/common/main/fa.xml
@@ -8736,13 +8736,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			</compoundUnit>
 			<unit type="acceleration-g-force">
 				<displayName>گرانش</displayName>
-				<unitPattern count="one">{0} &lt;LRM&gt;G</unitPattern>
-				<unitPattern count="other">{0} &lt;LRM&gt;G</unitPattern>
+				<unitPattern count="one">{0} G</unitPattern>
+				<unitPattern count="other">{0} G</unitPattern>
 			</unit>
 			<unit type="acceleration-meter-per-square-second">
 				<displayName>متر/مجذور ثانیه</displayName>
-				<unitPattern count="one">{0}&lt;LRM&gt; m/s²</unitPattern>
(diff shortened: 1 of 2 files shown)


AEApple:
Isn't this actually ticket [CLDR-14476](https://unicode-org.atlassian.net/browse/CLDR-14476)?

FrankYFTang:
Peter, this should be 14476 not 14477

FrankYFTang:
> * https://unicode-org.atlassian.net/browse/CLDR-14477

Please change it to https://unicode-org.atlassian.net/browse/CLDR-14476

FrankYFTang (commented):
gee.. it is hard to review invisible characters but I made it.
Please change the commit subject to CLDR-14476

macchiati (commented):
Only issue is the ticket number.

pedberg-icu:
> gee.. it is hard to review invisible characters but I made it.
> Please change the commit subject to [CLDR-14476](https://unicode-org.atlassian.net/browse/CLDR-14476)

Thanks, I fixed the commit message, and also the ticket ID in this PR, please re-approve

macchiati:
> gee.. it is hard to review invisible characters but I made it.

I filed a ticket for that: https://unicode-org.atlassian.net/browse/CLDR-14496?filter=10033

macchiati (approved):

PR #1029, opened 2021-02-11
CLDR-14491 BRS Task A05.5 : CLDRModify pass 3 -fp common/main

btangmu:
-Changed ordering of exemplarCharacters in nn.xml

##### Checklist


srl295 (approved):
this is why i wanted to run this phase. didn't want this to surprise us later

btangmu:
> this is why i wanted to run this phase. didn't want this to surprise us later

Entropy never sleeps so we must be eternally vigilant

PR #1031, opened 2021-02-12
CLDR-14493 Rework CLDR-2698/PR-964 changes to make nb a child of no, not an alias to it

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14493

The changes in [CLDR-2698](https://unicode-org.atlassian.net/browse/CLDR-2698) / [PR-964](https://github.com/unicode-org/cldr/pull/964) swapped the role of nb and no, making nb a deprecated alias to no, and removing the xml files for nb, while also making no a parent locale with child nn. The first part of this would have caused serious compatibility problems for some CLDR and ICU clients, including Windows.

This PR reworks those changes to make no a parent of both nb and no; nb is not deprecated or aliased. All of the xml files that used to be present for nb and nn are restored in this PR, though in some cases their content has been moved elsewhere (as in the previous PR) so that the nb/nn xml files restored in this PR are all empty stubs. This includes having empty stubs in collation/ for nb and nn; however the "actual" collation content for these is in no.

Some constraints/considerations/changes in this PR:
- In common/main, this restores nb.xml, nb_NO.xml, and nb_SJ.xml; nb is the default content for no, and nb_NO is the default content for nb. Both of the regional sublocales of nb (or no) are empty in any case.
- Since we can only have one default content locale for a parent, this means that we cannot have no_NO (and thus it makes no sense to have no_SJ). That is fine, we did not have them in earlier CLDR versions. So we have no (with no regional locales) as the parent of nb and nn, and then nb as the parent of nb_NO and  nb_SJ (and nn as the parent of nn_NO).
-  LocaleIDParser.getParent is fixed to not assume locales without '_' do not have a non-root parent (Mark I think this is the fix you were mentioning).
- TestInheritance.TestLikelyAndDefaultConsistency is fixed to use LocaleIDParser.getParent, not getSimpleParent.
- ... I may update this to mention other notable changes


common/supplemental/dayPeriods.xml
```diff
@@ -59,7 +59,7 @@
 			<dayPeriodRule type="evening1" from="18:00" before="24:00"/>	<!-- aften -->
 			<dayPeriodRule type="night1" from="00:00" before="05:00"/>	<!-- nat -->
 		</dayPeriodRules>
-		<dayPeriodRules locales="no">
+		<dayPeriodRules locales="nb no">
```
macchiati, line 62: Don't think this is necessary because plurals do inheritance, but doesn't hurt.

macchiati (approved):

tools/cldr-code/src/main/resources/org/unicode/cldr/draft/keyboard/osx/osx-locales.csv
```diff
@@ -85,9 +85,9 @@ Maltese,mt,
 Maori,mi,
 Nepali,ne,
 Northern Sami,se,
-Norwegian,no,
-Norwegian Extended,no,extended
-NorwegianSami-PC,no,extended-var
+Norwegian,nb,
+Norwegian Extended,nb,extended
+NorwegianSami-PC,nb,extended-var
```
srl295, line 90: i don't know how the keyboard files are used, but seems plausible

tools/cldr-rdf/external/entityToCode.tsv
```diff
@@ -1669,7 +1669,7 @@ http://www.wikidata.org/entity/Q23014	pfl
 http://www.wikidata.org/entity/Q23526	sla
 http://www.wikidata.org/entity/Q23855	auq
 http://www.wikidata.org/entity/Q25164	nn
-http://www.wikidata.org/entity/Q25167	no
+http://www.wikidata.org/entity/Q25167	nb
```
srl295, line 1672: This is a generated file. Will have to watch it when its next generated.

srl295 (approved):
LGTM…

tools/cldr-rdf/external/entityToCode.tsv
srl295: This is a snapshot for review purposes (an output, not an input).  The actual content is regenerated in the ST.  So this may not work properly and may need some followon. Shouldn't hold this PR though. s

jefgen (approved):
Thank you Peter! 👍

tools/cldr-rdf/external/entityToCode.tsv
pedberg-icu: I had looked up what that Wikidata entity referred to and it is in fact nb, not no. I had changed it before when we aliasing so no really meant nb.

tools/cldr-code/src/main/resources/org/unicode/cldr/draft/keyboard/osx/osx-locales.csv
pedberg-icu: I was assuming that what was being referred to here was the locale ID used on the platform. But this actually goes back to the code that was used in CLDR 38, I had chnaged in my previous Norwegian PR since nb was no longer valid.

tools/cldr-rdf/external/entityToCode.tsv
srl295: @pedberg-icu right, but nothing _reads_ entityToCode.tsv, it's output to give a historical record of the process.

PR #1032, opened 2021-02-12
CLDR-14491 BRS Task A05.5 : CLDRModify pass 3 -fp seed/main

btangmu:
-Changed U+00B5 MICRO SIGN to U+03BC GREEK SMALL LETTER MU

##### Checklist

### DIFF
--- a/seed/main/hi_Latn.xml
+++ b/seed/main/hi_Latn.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2020 Unicode, Inc.
+<!-- Copyright © 1991-2021 Unicode, Inc.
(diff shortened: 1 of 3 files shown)


btangmu:
This should be the last PR for this task. There were no changes for the remaining -fp runs, through exemplars/main.

srl295 (approved):
looks like seed was missed with CLDR-13485 / #968 / 4b7a5ed5f5271c4be2829766dac0f6e508510156 which this commit could be connected to

btangmu:
> looks like seed was missed with [CLDR-13485](https://unicode-org.atlassian.net/browse/CLDR-13485) / #968 / [4b7a5ed](https://github.com/unicode-org/cldr/commit/4b7a5ed5f5271c4be2829766dac0f6e508510156) which this commit could be connected to

Yes. Having seed and common in separate folders keeps causing problems. Recall that we skipped seed/annotations entirely last August due to conflict with Annotations.java being hard-coded for "common", not "seed", and I skipped it this time for the same reason.

PR #1033, opened 2021-02-15
CLDR-14386 Modernize List Users email/etc; fix SurveyAjax.doPost bug

btangmu:
-Fix SurveyAjax.doPost bug by skipping antiquated request.getReader code

-The doPost bug involved non-empty qs in POST due to USE_DOJO=...

-Add test in SurveyAjax for non-UTF-8 request encoding

-Add UserList constructor; misc. improvements to UserList.java (non-dojo)

-Misc. improvements to cldrAccount.js (non-dojo) for mass email, etc.

-Remove problematic and unneeded `@WebServlet` annotation

-Comments

##### Checklist

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -2,7 +2,6 @@ package org.unicode.cldr.web;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.io.Reader;
 import java.io.Writer;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -22,7 +21,6 @@ import java.util.TreeSet;
 
 import javax.servlet.ServletException;
 import javax.servlet.annotation.MultipartConfig;
-import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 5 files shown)


btangmu:
I had thought that @WebServlet might be required along with @MultipartConfig, but @WebServlet turns out to be problematic as well as unneeded.

If we have simply @WebServlet, we get: "[ERROR   ] SRVE9989E: A @WebServlet annotation on the class org.unicode.cldr.web.SurveyAjax has neither a value nor a urlPatterns attribute.  Either value or urlPatterns must be specified."

If we have @WebServlet("/SurveyAjax"), we get: "[ERROR   ] SRVE9016E: Unable to insert mapping [/SurveyAjax] for servlet named [org.unicode.cldr.web.SurveyAjax]. The URL pattern is already defined for servlet named [SurveyAjax]."

I guess that involves what is "already defined" in web.xml.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -157,31 +155,12 @@ public SurveyAjax() {
...
-            value = sb.toString();
-        } else {
-            value = request.getParameter("value"); // POST based value.
-        }
-        processRequest(request, response, value);
+        /*
+         * TODO: is request.getParameter("value") always sufficient? Maybe add a
+         * check that request.getCharacterEncoding() always returns “UTF-8” (as it does
+         * for me currently on localhost) and throw an exception if it doesn’t
+         */
```
btangmu, line 162: Actually I did add the test for UTF-8, see below.

btangmu:
I just found a mistake. I'll make another PR.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -3801,6 +3801,7 @@ public synchronized void doStartup() {
             }
             progress.update("Setup supplemental..");
             getSupplementalDataInfo();
+            progress.update("Checking if startup completed..");
```
btangmu, line 3805: Oops this got here by mistake

PR #1034, opened 2021-02-15
CLDR-14386 Modernize List Users email/etc; fix SurveyAjax.doPost bug

btangmu:
-Fix SurveyAjax.doPost bug by skipping antiquated request.getReader code

-The doPost bug involved non-empty qs in POST due to USE_DOJO=...

-Remove recently introduced problematic @WebServlet annotation

-Add test in SurveyAjax for non-UTF-8 request encoding

-Add UserList constructor; misc. improvements to UserList.java (non-dojo)

-Misc. improvements to cldrAccount.js (non-dojo) for mass email, etc.

-Comments

##### Checklist

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -2,7 +2,6 @@ package org.unicode.cldr.web;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.io.Reader;
 import java.io.Writer;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -22,7 +21,6 @@ import java.util.TreeSet;
 
 import javax.servlet.ServletException;
 import javax.servlet.annotation.MultipartConfig;
-import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -88,7 +86,6 @@
  * The @MultipartConfig annotation enables getting request parameters from a POST request
  * that has "multipart/form-data" as its content-type, as needed for WHAT_USER_LIST.
  */
-@WebServlet
 @MultipartConfig
```
btangmu, line 89: I had thought that @WebServlet might be required along with @MultipartConfig, but @WebServlet turns out to be problematic as well as unneeded.

If we have simply @WebServlet, we get: "[ERROR ] SRVE9989E: A @WebServlet annotation on the class org.unicode.cldr.web.SurveyAjax has neither a value nor a urlPatterns attribute. Either value or urlPatterns must be specified."

If we have @WebServlet("/SurveyAjax"), we get: "[ERROR ] SRVE9016E: Unable to insert mapping [/SurveyAjax] for servlet named [org.unicode.cldr.web.SurveyAjax]. The URL pattern is already defined for servlet named [SurveyAjax]."

I guess that involves what is "already defined" in web.xml.

srl295: I kind of had thought we would add new parts under teh new 'api' package and move away from new items under SurveyAjax , but this sounds fine

srl295 (approved):

PR #1035, opened 2021-02-15
CLDR-14578 Fix and improve comments in grammaticalFeatures.xml

hugovdm:
Rationale for the first change: "use a number for which nominative and genitive forms are distinct":
to make the example clear, we need a plural falling in the "few" or "one" categories, since these have distinctions between "nominative" and "genitive". The plural rule for "few" looks as follows:

```
    few{
        "v = 0 and i % 10 = 2..4 and i % 100 != 12..14 @integer 2~4, 22~24, 3"
        "2~34, 42~44, 52~54, 62, 102, 1002, …"
    }
```

33% seems like a more reasonable number than 2%, based on my lights for example - but 1% could also work.

The data for "percent" for "ru" shows no distinction between nominative and genitive for the "many" and "other" plural categories:

```
    genitive{
        few{"{0} процентов"}
        many{"{0} процентов"}
        one{"{0} процента"}
        other{"{0} процента"}
    }
```

Data for ru nominative:

```
    few{"{0} процента"}
    many{"{0} процентов"}
    one{"{0} процент"}
    other{"{0} процента"}
```

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14578

### DIFF
--- a/common/supplemental/grammaticalFeatures.xml
+++ b/common/supplemental/grammaticalFeatures.xml
@@ -13,10 +13,10 @@ These are the grammatical features that are used in each locale to mark inflecte
 CLDR textual items may be tagged with grammatical features when these are needed to ensure that the
 correct forms are used in context.
 
-For instance, in Russian "Уменьши яркость света до 10 процентов" ("dim the light to 10 percent"),
+For instance, in Russian "Уменьши яркость света до 33 процентов" ("dim the light to 33 percent"),
 the 'percent' unit must be expressed in plural and in the genitive case, whereas "1%" would be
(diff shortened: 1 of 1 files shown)


macchiati (approved):

macchiati (changes requested):
Looks reasonable, but needs to be filed against maint-40 instead of 39.

hugovdm:
Pardon some messiness in this PR: because the commit was 10 months old, I did a rebase onto the upstream/main branch, and force-pushed. (Two force-pushes in fact, because I first went for upstream/master.)

macchiati (approved):

PR #1036, opened 2021-02-16
CLDR-14400 Complete Vuefication (modernization) of About Box

srl295:
CLDR-14400

This is meant to also be a cookbook showing what we can do.

- use a Vue-based About Box
- Basic for now, but a simple template
- Delete AboutST and code in SurveyAjax, use an /api route
- Also, add "npm run watch" in the /js directory, this will rebuild bundle.js on any change

UI: 
image

it's not necessarily a UI improvement, but shows a modern structure.


tools/cldr-apps/js/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "exit 0",
     "lint": "exit 0",
-    "build": "webpack"
+    "build": "webpack",
+    "watch": "webpack watch"
```
srl295, line 10: this turned out to be really handy!  I have `mvn … liberty:dev` in one window, and `npm run watch` in another. Instant reload of any file, whether .js  (fraction of seconds) or .java (10s of seconds!)

tools/cldr-apps/js/src/views/About.vue
```diff
@@ -1,5 +1,73 @@
 <template>
```
srl295, line 1: comments:
- the CSS is *localized*.  Saves hunting around.
- this code simply enumerates the properties in `aboutData`.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/About.java
```diff
@@ -0,0 +1,103 @@
...
+                    + "  \"java_vm_version\": \"15.0.1+9\",\n"
+                    + "  \"os_arch\": \"x86_64\",\n"
+                    + "  \"os_name\": \"Mac OS X\",\n"
+                    + "  \"os_version\": \"10.16\",\n"
+                    + "  \"serverInfo\": \"IBM WebSphere Liberty/21.0.0.1\",\n"
+                    + "  \"servletVersion\": \"4.0\"\n"
+                    + "}\n"
+                    + "")))})
+    public Response getAbout() {
+        Map<String,String> r = new TreeMap<>();
```
srl295, line 65: for now it's a port (copy and paste really) of `AboutST.java` but could be more complex if needed.

tools/cldr-apps/js-unittest/TestCldrAbout.js
```diff
@@ -30,12 +30,5 @@
         "does contain angle brackets"
       );
     });
-
-    const label = "ICU";
-    const value = json.ICU_VERSION;
-    it("should contain " + label + " and " + value, function () {
-      assert(html.indexOf(label) !== -1, label + " does occur in " + html);
-      assert(html.indexOf(value) !== -1, value + " does occur in " + html);
-    });
```
btangmu, line 39: Would this part of the test not pass anymore?

btangmu (approved):

tools/cldr-apps/js/package.json
btangmu: When running that way, can you set breakpoints and step through the code?

btangmu: And is that with Eclipse, or without?

srl295: No impact on front end experience. It means that the npm build happens automatically. This is just for the generation of bundle.js. Yes you can debug in eclipse or in browser.

tools/cldr-apps/js-unittest/TestCldrAbout.js
srl295: Correct because Gethtml doesn't fetch the old endpoint.

btangmu:
@srl295 I just noticed CLDR_DATA_HASH etc. aren't clickable -- it would be good to restore them as links like https://github.com/unicode-org/cldr/commit/910849be69ff37687a229c6140aa5361d76c381f

srl295:
> @srl295 I just noticed CLDR_DATA_HASH etc. aren't clickable -- it would be good to restore them as links like [910849b](https://github.com/unicode-org/cldr/commit/910849be69ff37687a229c6140aa5361d76c381f)

Certainly, I can do that as Vue side code.  Probably could even pull some of that code forwad.  I wanted to keep this PR more straightforward.

PR #1040, opened 2021-02-16
CLDR-14484 fix typo in server config

srl295:
CLDR-14484

should be capital C per https://openliberty.io/docs/20.0.0.12/reference/feature/localConnector-1.0.html


btangmu (approved):
Capital!

PR #1041, opened 2021-02-17
CLDR-14461 for alpha, update readme, update ICU4J libs to version of today

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14461

Updates for CLDR alpha:
- Update readme (licenses already updated from SBRS)
- Update ICU4J libs to version of today, picking up Norwegian changes, latest grammar support, etc.
- Update tools/cldr-code/src/license/THIRD-PARTY.properties as generated. Not sure whether we need to add an ICU or Unicode license reference somewhere.


tools/cldr-code/src/license/THIRD-PARTY.properties
```diff
@@ -1,6 +1,16 @@
...
 # Please fill the missing licenses for dependencies :
-# this may need to be updated before building final outputs
-# (for now)
-com.ibm.icu--icu4j-for-cldr--68.1-release-68-1=Unicode-DFS-2016
-com.ibm.icu--utilities-for-cldr--68.1-release-68-1=Unicode-DFS-2016
+#
+#
+#Wed Feb 17 10:22:40 PST 2021
+com.ibm.icu--icu4j-for-cldr--69.1-SNAPSHOT-cldr-2021-02-17=
+com.ibm.icu--utilities-for-cldr--69.1-SNAPSHOT-cldr-2021-02-17=
```
pedberg-icu, line 16: Do we need to add a reference to ICU or Unicode license somewhere to remove these lines?

srl295: I think there is probably a way to put the correct license metadata into those files so it can be picked up.  I started a PR to do that on the ICU side.

The license is the same as CLDRs so should be covered.

srl295 (approved):

PR #1042, opened 2021-02-17
CLDR-14511 startup modernization

srl295:
CLDR-14511 work in progress
- quell a warning in FileProcessor
- document parts of CLDR startup
- use `doGet()` in SurveyTool instead of `service()` ( `/v` should only respond to `GET`!)
- add Status data for whether startup is in progress


tools/cldr-code/src/main/java/org/unicode/cldr/util/FileProcessor.java
```diff
@@ -10,6 +10,7 @@
 import com.ibm.icu.util.ICUUncheckedIOException;
 
 public class FileProcessor {
+	private final static boolean DEBUG = false;
     private int lineCount;
```
btangmu, line 14: indentation

srl295: fixed, thanks

tools/cldr-apps/js/src/views/Waiting.vue
```diff
@@ -0,0 +1,56 @@
...
+    methods: {
+        fetchStatus: function() {
+            fetch("SurveyAjax?what=status")
+            .then(r => r.json())
+            .then(data => {
+                this.statusData = data;
+                if(!data.triedToStartUp) {
+                    // attempt to fetch /survey
+                    console.log("Attempting fetch of /cldr-apps/survey");
+                    fetch("survey")
```
srl295, line 35: this works to cause ST to startup. Just need the waiting page to retry the status ajax, and call `reload()` when needed

tools/cldr-apps/js/src/index.js
```diff
@@ -3,12 +3,17 @@ import "./style.css";
...
 }
+function showWaiting(el) {
+  return createApp(Waiting).mount(el);
+}
 
 // The following will show up in the cldrBundle global
 export default {
   showAbout,
+  showWaiting,
 };
```
btangmu, line 19: this list -- and this file -- will be too long if we continue with this approach when everything's in the bundle; let's find a structure with separation of concerns -- it may be a lot easier if we convert all the js files into modules as in https://unicode-org.atlassian.net/browse/CLDR-14401

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -129,92 +124,112 @@ private void serveBustedPage(HttpServletRequest request, PrintWriter out) {
...
-            out.write("        }, 2000); // two seconds\n");
-            out.write("      });\n");
-            if (useDojo(request)) {
-                out.write("  });\n");
+            // fire up the Vue based waiting page
+            out.write("</head>\n<body>\n");
+            writeWaitingNavbarHtml(out);
+            out.write("<div id='app'></div>");
+            out.write("<script>cldrBundle.showWaiting('#app');</script>");
+            out.write("</body>");
```
btangmu, line 144: instead of actual js inside html, how about just including an element with a certain id, whose presence tells cldrBundle.start() to show the waiting page instead of the normally-running page?

tools/cldr-apps/js/src/index.js
btangmu: does "default" do anything helpful here?

tools/cldr-apps/js/src/index.js
```diff
@@ -3,12 +3,17 @@ import "./style.css";
 
 import { createApp } from "vue";
 import About from "./views/About.vue";
+import Waiting from "./views/Waiting.vue";
```
btangmu, line 7: 'Components names should always be multiwords like "AlertModal", "DropdownMenu" or "NavbarLogo"'
https://namingconvention.org/vuejs/components-naming.html
https://v3.vuejs.org/style-guide/#multi-word-component-names-essential

srl295: Good point. 
- [ ] I'll pick some better names here.

tools/cldr-apps/js/src/index.js
srl295: > does "default" do anything helpful here in "export default"? maybe not needed

@btangmu yes, it's the 'default export', and is what turns into the cldrBundle object.

srl295: > this list -- and this file -- will be too long if we continue with this approach when everything's in the bundle; let's find a structure with separation of concerns -- it may be a lot easier if we convert all the js files into modules as in https://unicode-org.atlassian.net/browse/CLDR-14401

well, 'the bundle' is intended to be the entire application. But I think it should have a mere handful of top level functions visible here.  But I will reply on your other comment:

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
srl295: @btangmu a router! Actually, what we can have is one entrypoint, something like `updateVue()` that's _always_ called.  And it will internally have some kind of map like this:

```json
{
  "/v#about": "About.vue",
  "/v#retry": "Waiting.vue"
}
```

Would that work for now? then there would be minimal code change needed for moving something from the "new" JS model to the "vue" JS model, and there'd be no extraneous `cldrBundle.` level exports.

btangmu: That sounds better... "loose coupling" means most of our code shouldn't know or care whether we're using webpack

tools/cldr-apps/js/src/index.js
srl295: done changed to AboutPanel and WaitingPanel

btangmu: great!

srl295:
todo: 
- [ ] loose coupling

srl295:
ready for review

image

btangmu (approved):

btangmu:
@srl295 I'll go ahead and merge this

tools/cldr-apps/js/src/index.js
```diff
@@ -2,13 +2,42 @@ import "./style.css";
...
 import { createApp } from "vue";
-import About from "./views/About.vue";
 
-function showAbout(el) {
-  return createApp(About).mount(el);
+import AboutPanel from "./views/AboutPanel.vue";
+import WaitingPanel from "./views/WaitingPanel.vue";
+
+function show(component, el) {
+  const app = createApp(component);
```
btangmu, line 10: minor quibble: I would move show() down below the function that calls it

old-fashioned languages like Pascal required low-level functions at the top and high-level functions at the bottom, and that's also seen in some old-fashioned ways of writing JavaScript ("var foo = function..." vs modern "function foo") -- for me, at least, code is way easier to read with high-level at the top and subroutines at the bottom

tools/cldr-apps/js/src/index.js
```diff
@@ -2,13 +2,42 @@ import "./style.css";
...
+ */
+function showPanel(type, el) {
+  if (lastMounted) {
+    lastMounted.unmount();
+    lastMounted = null;
+  }
+  if ('about' == type) {
+    lastMounted = show(AboutPanel, el);
+  } else if('retry' == type) {
+    lastMounted = show(WaitingPanel, el);
```
btangmu, line 32: pretty.sh would insert a space after if

srl295: good point

srl295:
> @srl295 I'll go ahead and merge this

how'd it work for you?

btangmu:
> how'd it work for you?

Some problems in browser console during startup:

```
Navigated to http://localhost:9080/cldr-apps/survey?email=heikapupperson.ltsmpzt8p@frkv.gaeilge.example.com&uid=...
This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”. v
You are running the esm-bundler build of Vue. It is recommended to configure your bundler to explicitly replace feature flag globals with boolean literals to get proper tree-shaking in the final bundle. See http://link.vuejs.org/feature-flags for more details. runtime-core.esm-bundler.js:3413
[Vue warn]: Failed to resolve component: tt 
  at <WaitingPanel> 2 runtime-core.esm-bundler.js:38
Attempting fetch of /cldr-apps/survey WaitingPanel.vue:74:29
[Vue warn]: Failed to resolve component: tt 
  at <WaitingPanel> runtime-core.esm-bundler.js:38
downloadable font: download failed (font-family: "Open Sans" style:normal weight:400 stretch:100 src index:0): status=2152398850 source: http://fonts.gstatic.com/s/opensans/v18/mem8YaGs126MiZpBA-UFVZ0bf8pkAg.woff2
downloadable font: download failed (font-family: "Noto Sans" style:normal weight:400 stretch:100 src index:0): status=2152398850 source: http://fonts.gstatic.com/s/notosans/v11/o-0IIpQlx3QUlC5A4PNr5TRASf6M7Q.woff2
[Vue warn]: Failed to resolve component: tt 
  at <WaitingPanel> runtime-core.esm-bundler.js:38
[Vue warn]: Failed to resolve component: tt 
  at <WaitingPanel> runtime-core.esm-bundler.js:38
[Vue warn]: Failed to resolve component: tt 
  at <WaitingPanel> runtime-core.esm-bundler.js:38
...
```

srl295:
> `tt`

TT is deprecated i think, i can remove it.

btangmu:
> > `tt`
> 
> TT is deprecated i think, i can remove it.

kk! other than the console warnings, everything seems ok so far; it's a little hard to test the start-up code since it really seems to depend on the time of when the browser is refreshed

srl295:
> > > `tt`
> > 
> > 
> > TT is deprecated i think, i can remove it.
> 
> kk! other than the console warnings, everything seems ok so far; it's a little hard to test the start-up code since it really seems to depend on the time of when the browser is refreshed

1. you can test by putting `SurveyMain.busted("we are now busted..")` somewhere strategically
2. if you visit the `/cldr-apps/survey` link before startup happens, you should now get redirected to `/v#retry`

PR #1045, opened 2021-02-17
CLDR-14383 Modernize My Account Settings, Activity

btangmu:
-My Account getUserActivityLink goes to recent_activity not myvotes.jsp

-Move some functions from cldrSurvey.js to cldrRecentActivity.js

-Add new function cldrAjax.makeUrl to encapsulate SurveyAjax dependency

-Comments; clean-up; js modernization such as var to let/const

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14383


tools/cldr-apps/src/main/webapp/js/new/cldrRecentActivity.js
```diff
@@ -67,6 +75,224 @@ const cldrRecentActivity = (function () {
 
     return html;
   }
+
+  function showRecent(divName, locale, user) {
+    if (!locale) {
+      locale = "";
+    }
```
btangmu, line 82: locale is always null (then "")
divName is always "submitItems"
we can simplify later

tools/cldr-apps/src/main/webapp/js/new/cldrRecentActivity.js
```diff
@@ -67,6 +75,224 @@ const cldrRecentActivity = (function () {
...
+      }
+    }
+
+    function errorHandler(err) {
+      cldrRetry.handleDisconnect("Error in showrecent: " + err);
+    }
+  }
+
+  function showAllItems(divName, user) {
+    const div = document.getElementById(divName);
```
btangmu, line 198: divName is always "allMyItems"

btangmu:
I've made further changes, will make another PR

PR #1048, opened 2021-02-18
CLDR-14506 Updated Singaporean currency symbol

Kherishma:
Updated Singaporean currency symbol from $ to S$ on line 151

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14506

### DIFF
--- a/common/main/en_SG.xml
+++ b/common/main/en_SG.xml
@@ -148,7 +148,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	<numbers>
 		<currencies>
 			<currency type="SGD">
-				<symbol>$</symbol>
+				<symbol>"S$"</symbol>
 			</currency>
 		</currencies>
 	</numbers>


common/main/en_SG.xml
```diff
@@ -148,7 +148,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	<numbers>
 		<currencies>
 			<currency type="SGD">
-				<symbol>$</symbol>
+				<symbol>"S$"</symbol>
```
pedberg-icu, line 151: The symbol should not have quote marks around it. But a bigger question: We currently have four locales for SG: en_SG, ms_SG, ta_SG, and zh_Hans_SG. All of them currently use just "S" for SGD, and for USD they have (either in the locale or inherited from parent) "US$" or "USD". We need a broader discussion about changing the symbol for SGD in all of the _SG locales.

milnivlek:
I was born and raised in Singapore. Please see my latest comment at https://unicode-org.atlassian.net/browse/CLDR-14506?focusedCommentId=182337. This change is incorrect IMO -- the currency symbol ought to remain "$" in `_SG` locales. Instead, what would probably make more sense is to change it to "S$" in **non**-SG locales. Ideally this should apply to all other locales... but at the very least, we should change it for non-SG English language variants.

AEApple:
Closing, since the commit would need to be modified it will be easier to just make a new commit.

PR #1050, opened 2021-02-18
CLDR-14519 alternative names for keyboard LDML

miloush:
- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14519

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -319,21 +319,34 @@ A single name given to the layout by the platform.
 Syntax
 
 ```xml
-<name value="..">  
+<name value=".." [type="{usage}"] />
 ```
 
 _Attribute:_ `value` (required)
 
 > The name of the layout.
 
+_Attribute:_ `type="full"` (optional)
+
+> This attribute indicates the context of an alternative name. Valid values are: 
+> 
+> `layout` The `value` attribute describes the layout pattern, such as QWERTY, INSCRIPT, etc. typically used to distinguish various layouts of the same language. 
+> 
+> `indicator` The `value` attribute describes a short string to be used in currently selected layout indicator, such as US, SI9 etc. 
+> 
+> `normalization` The `value` attribute describes the intended normalization form of the keyboard layout output, and can be either `C`, `D` or `mixed`.
+
 Example
 
 ```xml
-<keyboard locale="bg-t-k0-windows-phonetic-trad">
+<keyboard locale="el-Latn-t-k0-windows-el220">
     …
-    <names>
-        <name value="Bulgarian (Phonetic Traditional)" />
-    </names>
+    <names> 
+        <name value="Greek (220) Latin" /> 
+        <name type="layout" value="QWERTY" /> 
+        <name type="indicator" value="220L" /> 
+        <name type="normalization" value="C" /> 
+    </names> 
     …
 </keyboard>
 ```
--- a/keyboards/dtd/ldmlKeyboard.dtd
+++ b/keyboards/dtd/ldmlKeyboard.dtd
@@ -34,6 +34,8 @@ Except as contained in this notice, the name of a copyright holder shall not be 
 <!ELEMENT names ( name+ ) >
 
 <!ELEMENT name EMPTY >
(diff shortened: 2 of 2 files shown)


keyboards/dtd/ldmlKeyboard.dtd
```diff
@@ -34,6 +34,8 @@ Except as contained in this notice, the name of a copyright holder shall not be
 <!ELEMENT names ( name+ ) >
 
 <!ELEMENT name EMPTY >
+<!ATTLIST name type CDATA #IMPLIED > 
+    <!--@MATCH:literal/layout, indicator, normalization-->
```
srl295, line 38: I think it would be good to have an implied (but legal) `type=` which matches the current use of `<name>`, something like `<name type="display" value="Greek (220) Latin" />`

docs/ldml/tr35-keyboards.md
```diff
@@ -319,21 +319,33 @@ A single name given to the layout by the platform.
...
+<keyboard locale="el-Latn-t-k0-windows-el220">
     …
-    <names>
-        <name value="Bulgarian (Phonetic Traditional)" />
-    </names>
+    <names> 
+        <name value="Greek (220) Latin" /> 
+        <name type="layout" value="QWERTY" /> 
+        <name type="indicator" value="220L" /> 
+    </names>
```
srl295, line 348: would be good to include an example of `type=normalization` also

srl295:
CI failure:
```
Error: (TestDtdData.java:307) : isDistinguished keyboard: elementName.equals("name") && attribute.equals("type"): expected false, got true
```

@macchiati is there an update step that is needed?

keyboards/dtd/ldmlKeyboard.dtd
xadxura: We discussed having permitted normalization types "NFC", "NFD", "mixed" in [CLDR-13273](https://unicode-org.atlassian.net/browse/CLDR-13273). Per Richard's comment, maybe the third type is "other", or possibly omitted is better if the layout is non-conformant. Richard points to the need for compatibility with USE being a third type, ,so "USE", but I don't want to enshrine "USE" this way.

docs/ldml/tr35-keyboards.md
```diff
@@ -319,21 +319,34 @@ A single name given to the layout by the platform.
...
+<keyboard locale="el-Latn-t-k0-windows-el220">
     …
-    <names>
-        <name value="Bulgarian (Phonetic Traditional)" />
-    </names>
+    <names> 
+        <name value="Greek (220) Latin" /> 
+        <name type="layout" value="QWERTY" /> 
+        <name type="indicator" value="220L" /> 
+        <name type="normalization" value="C" />
```
xadxura, line 348: It think it would be clearer if the normalization type were "NFC"

xadxura (changes requested):
Let's clarify the normalization types.

docs/ldml/tr35-keyboards.md
macchiati: agreed. Is there an explanation of what the values are somewhere? (220L is pretty opaque...)

macchiati: Also, by constructing it this way, we can't easily do DTD validation. I think this would be cleaner as:

<name value="Greek (220) Latin" layout="QWERTY" indicator="220L"  normalization="NFC" />

macchiati (changes requested):

miloush:
@macchiati

> I think this would be cleaner as:

Did you perhaps include code without a code block? I can't see your suggestion.

macchiati:
I hadn't quoted the XML, so it disappeared.

miloush:
I see. I don't think it really matters to me which way this is done, I guess it depends on what purpose multiple `<name>` elements were envisioned to serve, as I imagine at least `@layout` and `@normalization` would be shared across names. Also I believe your suggestion would make it impossible to include `<name>` of specific "type" using `<import>` (if that is something we care about). I will keep this open then and we can take it up in the WG.

macchiati:
Sounds good.

Mark

PR #1053, opened 2021-02-18
CLDR-14523 fix BRS problems

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14523

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -1620,10 +1620,6 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Mofu-Gudur; ?; ? } => { Mofu-Gudur; Latin; Unknown Region }-->
 		<likelySubtag from="min" to="min_Latn_ID"/>
 		<!--{ Minangkabau; ?; ? } => { Minangkabau; Latin; Indonesia }-->
-		<likelySubtag from="mis" to="mis_Hatr_IQ"/>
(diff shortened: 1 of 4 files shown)


yumaoka:
@macchiati There are some test failures with this change. Please take a look.

https://github.com/unicode-org/cldr/pull/1053/checks?check_run_id=1930119403
```
Error summary:
CLDR/LikelySubtagsTest/TestCompleteness
CLDR/LikelySubtagsTest/TestMissingInfoForScript


<< 3 TEST(S) FAILED >>
```

macchiati:
The problem with the first commit is that (1) the mapping is not idempotent (filed XXX also) and (2) Hatr is known to not have a language (the script metadata was hiding that by using 'mis', but we don't want 'mis' showing up in likely subtags.

macchiati:
The XXX is https://unicode-org.atlassian.net/browse/CLDR-14524

yumaoka (approved):
LGTM

btangmu:
Please let me know when it's time for me to run CLDRModify again!

macchiati:
I haven't heard from Steven on the following. So I'm doing the 2nd right
now (the first doesn't affect CLDR Modify), so once the 2nd is done you are
good to start CLDRModify
todo: CLDR-14512 Steven/Mark
todo: CLDR-14513 Steven/Mark
Mark

macchiati:
I did both of those. #1 needed no changes, so closed the ticket.
I have https://github.com/unicode-org/cldr/pull/1055. If one of you can
review that'd be great, and you can merge right after.
Mark

PR #1054, opened 2021-02-18
CLDR-14401 Modernize by using ESM (js modules)

btangmu:
-Enable running ESM versions of cldr*.js directly in browser

-This depends on first running convertToModule.mjs

-To test, change SurveyTool.USE_ESM to true

-Move onclick from html to js (more needed)

##### Checklist

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
@@ -21,6 +21,7 @@ public class SurveyTool extends HttpServlet {
     private static final String USE_DOJO_VAR = "USE_DOJO";
     private static final long serialVersionUID = 1L;
     private static final boolean USE_DOJO_DEFAULT = true;
+    private static final boolean USE_ESM = false;
 
     /**
      * Is dojo enabled by default?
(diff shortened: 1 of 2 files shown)


srl295 (approved):
LGTM.

bundle.js can produce an ESM also.

btangmu:
> bundle.js can produce an ESM also.

Interesting! Could it be a minified ESM? For production I've been assuming we'd use a non-ESM bundle... but we do require modern browsers anyway. 

BTW I'm making this PR ready for review -- it's harmless and maybe the start of something. The onclick change is probably for the best anyway -- do you think that will be an issue with webpack as well? Imported functions don't work in js embedded in html attributes?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -21,6 +21,7 @@
     private static final String USE_DOJO_VAR = "USE_DOJO";
     private static final long serialVersionUID = 1L;
     private static final boolean USE_DOJO_DEFAULT = true;
+    private static final boolean USE_ESM = false;
```
srl295, line 24: should be ideally a config parameter

srl295:
> > bundle.js can produce an ESM also.
> 
> Interesting! Could it be a minified ESM? For production I've been assuming we'd use a non-ESM bundle... but we do require modern browsers anyway.

that i'm not sure but I assume it is minifiable. 

> BTW I'm making this PR ready for review -- it's harmless and maybe the start of something. The onclick change is probably for the best anyway -- do you think that will be an issue with webpack as well? Imported functions don't work in js embedded in html attributes?

not sure how you'd access it, might need to have something else export it.  I think this change is fine to go in as it is ..

PR #1055, opened 2021-02-18
CLDR-14513 algorithmic locales BRS

macchiati:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14513

### DIFF
--- a/common/annotations/sr_Latn.xml
+++ b/common/annotations/sr_Latn.xml
@@ -12,7 +12,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<script type="Latn"/>
 	</identity>
 	<annotations>
-		<annotation cp="{">leva vitičasta zagrada | vitičasta zagrada | zagrada</annotation>
+		<annotation cp="{">vitičasta zagrada | zagrada | leva vitičasta zagrada</annotation>
 		<annotation cp="{" type="tts">leva vitičasta zagrada</annotation>
 		<annotation cp="🏻">boja kože | svetla koža | tip kože 1–2</annotation>
 		<annotation cp="🏻" type="tts">svetla koža</annotation>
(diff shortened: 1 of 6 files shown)


common/annotations/sr_Latn.xml
```diff
@@ -12,7 +12,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<script type="Latn"/>
 	</identity>
 	<annotations>
-		<annotation cp="{">leva vitičasta zagrada | vitičasta zagrada | zagrada</annotation>
+		<annotation cp="{">vitičasta zagrada | zagrada | leva vitičasta zagrada</annotation>
 		<annotation cp="{" type="tts">leva vitičasta zagrada</annotation>
```
btangmu, line 16: Why has the ordering changed from alphabetical to non-alphabetical? Same question for changes below in this file

common/annotations/sr_Latn.xml
```diff
@@ -1704,305 +1704,305 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
 		<annotation cp="🛖">koliba | kuća | okrugla koliba</annotation>
 		<annotation cp="🛖" type="tts">koliba</annotation>
 		<annotation cp="🏘">kuće</annotation>
 		<annotation cp="🏘" type="tts">kuće</annotation>
-		<annotation cp="🏚">kuća | zapuštena kuća | zapušteno</annotation>
+		<annotation cp="🏚">zapušteno | zapuštena kuća | kuća</annotation>
 		<annotation cp="🏚" type="tts">zapuštena kuća</annotation>
-		<annotation cp="🏠">dom | kuća</annotation>
+		<annotation cp="🏠">dom | kuća | kuća</annotation>
 		<annotation cp="🏠" type="tts">kuća</annotation>
```
btangmu, line 1870: Here there's a duplicate -- kuća appears twice
It seems that the old file was normalized (maybe by CLDRModify) and the new file is not normalized
This is a kind of regression; it might be fixed when CLDRModify is run again; it would save us some churn, though, if the normalization were combined with the algorithmic generation

btangmu (approved):
I'm approving under the assumption that the de-normalization is par for the course and is to be corrected by CLDRModify. I'm not sure enough to do the merge myself, though, without confirmation

macchiati:
Yes, that's right. The algorithmic changes need DAIP (CLDRModify) run after
them. In this case they are transliterating sr, which produce results that
are not in the right order for the Latin form.
Mark

btangmu:
OK! Should I go ahead and start with CLDRModify tomorrow morning?

PR #1056, opened 2021-02-19
CLDR-14531 Create _config.yml

btangmu:
Configure markdown; see https://unicode-org.atlassian.net/browse/CLDR-14531

##### Checklist


btangmu:
"Bad commit message" -- adding file this way didn't work; I fixed the PR message but not the git commit
I'll make another PR

PR #1057, opened 2021-02-19
CLDR-14531 BRS spec fix

btangmu:
-Create _config.yml to configure markdown

##### Checklist

### DIFF
--- a/docs/ldml/_config.yml
+++ b/docs/ldml/_config.yml
@@ -1,0 +1,3 @@
+name: CLDR LDML
+
+markdown: GFM


macchiati (approved):
I don't know if it makes a difference, but Jan's original didn't have a blank line. I'd say go ahead with this, and then we'll see if we need to adjust.

macchiati:
I'll go ahead and push so we can see the results and tune if we need.

macchiati:
It doesn't work (or maybe I didn't give it enough time). I suspect that according to the instructions we need to put it at the top level, eg in the docs directory, not the LDML directory.

srl295:
> It doesn't work (or maybe I didn't give it enough time). I suspect that according to the instructions we need to put it at the top level, eg in the docs directory, not the LDML directory.

sounds right.

srl295:
The bug wasn't clear as to what it was trying to fix. I added comments on [CLDR-14531](https://unicode-org.atlassian.net/browse/CLDR-14531?focusedCommentId=160245)

btangmu:
> It doesn't work (or maybe I didn't give it enough time). I suspect that according to the instructions we need to put it at the top level, eg in the docs directory, not the LDML directory.

I'll try that...

PR #1058, opened 2021-02-19
CLDR-14530 CLDRFileTransformer should call DisplayAndInputProcessor

btangmu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14530


macchiati (approved):

macchiati:
BTW, I assume you've tried this out to make sure it works.

Mark

btangmu:
> BTW, I assume you've tried this out to make sure it works

Yes! Though I didn't study the output in full detail, so we should look carefully at the results next time; the good news is there should be a lot less noise in those results

PR #1060, opened 2021-02-19
CLDR-14531 BRS spec fix, 2nd try to configure markdown

btangmu:
-Move _config.yml from docs/ldml to docs

##### Checklist


srl295 (approved):

srl295:
LGTM.

👉  Note that a way to debug this kind of thing is to use your own personal fork, where you can set whatever settings you want. https://srl295.github.io/cldr/ldml/tr35.html for example. Then you don't need to make new PRs to try things.

macchiati (approved):

PR #1061, opened 2021-02-21
CLDR-14401 Modernize with ESM

btangmu:
-Convert `new/*.js` to modules in `esm/*.js` and remove folder `new`

-Convert `js-unittest/*.js` to ESM modules

-Revise `js-unittest/Test.html` as needed

-Add `js-unittest/run.sh` since Test.html now requires a server

-Clone the `new/*.js` used with old dojo such as CldrDojoAjax.js from new/cldrAjax.js

-Move more onclick from html to js

-Delete references to `new/*.js` from SurveyTool.java

##### Checklist


btangmu:
I'm going to close this, fix a few things, and make another pr shortly

PR #1062, opened 2021-02-21
CLDR-14401 Modernize: js/esm replaces js/new

btangmu:
-Convert `new/*.js` to modules in `esm/*.js` and remove folder new

-Convert `js-unittest/*.js` to ESM modules

-Revise js-unittest/Test.html as needed

-Add js-unittest/run.sh since Test.html now requires a server

-Clone the `new/*.js` used with old dojo such as CldrDojoAjax.js from new/cldrAjax.js

-Move more onclick from html to js

-Delete references to `new/*.js` from SurveyTool.java

-Fix bug in cldrAdmin.js: change e.header to entry.header

-Fix bug in UserList.java: allow for justOrg.isEmpty

-Comments

##### Checklist


srl295 (approved):
seems reasonable to me… is there a good path to moving this into webpack?

btangmu:
> seems reasonable to me… is there a good path to moving this into webpack?

yes, if we tell webpack to import only cldrGui.js, then webpack can do what the browser does -- import all the modules linked (directly or indirectly) from cldrGui automatically and make them part of the bundle

or so I imagine...

srl295:
> > seems reasonable to me… is there a good path to moving this into webpack?
> 
> yes, if we tell webpack to import only cldrGui.js, then webpack can do what the browser does -- import all the modules linked (directly or indirectly) from cldrGui automatically and make them part of the bundle
> 
> or so I imagine...

Is it still important to have .js files included directly? Or is there a path towards moving everything through webpack for consistency?

btangmu:
> Is it still important to have .js files included directly? Or is there a path towards moving everything through webpack for consistency?

For development and testing, and separation of concerns, I think it's very valuable to retain the ability to run with the esm files directly without bundling them. Almost certainly we'll use bundling normally for production, for the sake of performance, though it's worth measuring the performance difference from time to time as a reality check. Definitely consistency is important, and we can be consistent by making as much code as possible agnostic about bundling -- it should work with or without bundling and it shouldn't care. Let's avoid tight coupling with any of our dependencies. Most of our code shouldn't know and shouldn't care whether we're using webpack or some other bundler or none.

PR #1069, opened 2021-02-23
CLDR-14401 Modernize with ESM; include esm in webpack bundle

btangmu:
-New function runGui in index.js calls cldrGui.run

-Move subroutine show() below its caller

-New boolean SurveyTool.RUN_ALL_JS_FROM_BUNDLE

##### Checklist

### DIFF
--- a/tools/cldr-apps/js/src/index.js
+++ b/tools/cldr-apps/js/src/index.js
@@ -6,13 +6,13 @@ import { createApp } from "vue";
 import AboutPanel from "./views/AboutPanel.vue";
 import WaitingPanel from "./views/WaitingPanel.vue";
 
-function show(component, el) {
-  const app = createApp(component);
-  app.mount(el);
-  return app;
-}
+import * as cldrGui from "../../src/main/webapp/js/esm/cldrGui.js";
 
 let lastMounted = null;
+
+function runGui() {
+  cldrGui.run();
+}
 
 /**
  * Mount a Vue component.
@@ -37,7 +37,14 @@ function showPanel(type, el) {
   return lastMounted;
 }
 
+function show(component, el) {
+  const app = createApp(component);
+  app.mount(el);
+  return app;
+}
+
 // The following will show up in the cldrBundle global
 export default {
+  runGui,
   showPanel,
 };
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
@@ -19,6 +19,13 @@ public class SurveyTool extends HttpServlet {
     private static final String USE_DOJO_VAR = "USE_DOJO";
     private static final long serialVersionUID = 1L;
     private static final boolean USE_DOJO_DEFAULT = true;
+    /**
+     * Normally we want to run all js from a bundle, for performance,
+     * especially in production. This can made false for testing and
+     * development, in which case the browser will load and run the
+     * individual cldr*.js modules directly.
+     */
+    private static final boolean RUN_ALL_JS_FROM_BUNDLE = true;
 
     /**
      * Is dojo enabled by default?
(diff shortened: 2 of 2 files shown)


tools/cldr-apps/js/src/index.js
```diff
@@ -6,14 +6,14 @@ import { createApp } from "vue";
 import AboutPanel from "./views/AboutPanel.vue";
 import WaitingPanel from "./views/WaitingPanel.vue";
```
btangmu, line 7: I think actually cldrAbout.js should import AboutPanel directly -- this detail is the concern of cldrAbout, not index.js

We can move the `views` folder inside `webapps` -- then just like the js files, the vue files will be pulled into the bundle automatically and will also be usable without bundling

srl295: the views are part of the bundle.  I think we should drop cldrAbout.js eventually and have a new top level router for the single page app.

btangmu: cldrAbout might be a special case given how small it is

in general modules should be empowered to use vue without being tightly coupled with webpack

a new top level router sounds good -- I found it awkward to add the user id as in `v#recent_activity///1234`

srl295: I think we need a phased approach and let's discuss this afternoon.

We're not rewriting all modules in Vue for 39, that seems clear. So let's take the current approach with the single top level entrypoint I'm using for those few items (About, Retry, and maybe Dashboard and Login) that use it, and then in 40 revisit a new top level router/top level SPA , ref CLDR-14398

btangmu: My comment about cldrAbout is really tangential to what this ticket does

I've tested this PR and it works for creating bundle.js that contains *everything* -- that's what we want, right? Is there any problem with this PR?

tools/cldr-apps/js/src/index.js
```diff
@@ -6,14 +6,14 @@ import { createApp } from "vue";
...
 
-function show(component, el) {
-  const app = createApp(component);
-  app.mount(el);
-  return app;
-}
+import * as cldrGui from "../../src/main/webapp/js/esm/cldrGui.js";
```
srl295, line 9: Clever.. I'm ok with this for now though it results in duplicate JavaScript

srl295 (approved):
Lgtm for now and let's discuss

tools/cldr-apps/js/src/index.js
srl295: Sorry i had forgotten the actual review !

srl295: > cldrAbout might be a special case given how small it is
> 
> 
> 
> in general modules should be empowered to use vue without being tightly coupled with webpack
> 
> 
> 
> a new top level router sounds good -- I found it awkward to add the user id as in `v#recent_activity///1234`
> 
> 

Question about the bundler- have you tried using the browsers debugger against something in the bundle?

tools/cldr-apps/js/src/index.js
btangmu: not sure what you mean by duplicate js

srl295: I meant that `cldrGui.js` would be pulled into the webpack bundle in `bundle.js`, but would also be available separately (unbundled)  under `src/main/webapp/js/esm/cldrGui.js`

tools/cldr-apps/js/src/index.js
btangmu: Yes, check out the screenshot:
image

srl295:
That’s great, seems like it is working

tools/cldr-apps/js/src/index.js
btangmu: Yes, all the esm files are available separately, but when they're used in the bundle, they're not used outside the bundle -- in that sense there's no duplication

the browser network tab confirms that

srl295: @btangmu let's come back to this

PR #1070, opened 2021-02-23
CLDR-14491 BRS§A24.4 rerun CLDRModify: pass 3 -fp common/annotations

btangmu:
-Normalized 1 file

##### Checklist

### DIFF
--- a/common/annotations/my.xml
+++ b/common/annotations/my.xml
@@ -1381,7 +1381,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🐡" type="tts">ငါးပူဖောင်း</annotation>
 		<annotation cp="🦈">ငါး | ငါးမန်း</annotation>
 		<annotation cp="🦈" type="tts">ငါးမန်း</annotation>
-		<annotation cp="🐙">ဘဝှဲ ရဘဝှေဲ ဘဝှဲ ဘဝှဲ ဘဝွဲ ဘဝျဲ ဘဝဲ | ဘဝဲ</annotation>
+		<annotation cp="🐙">ဘဝှဲ ရဘဝှေဲ ဘဝှဲ ဘဝှဲ ဘဝှဲ ဘဝြဲ ဘဝျဲ ဘဝဲ</annotation>
 		<annotation cp="🐙" type="tts">ဘဝဲ</annotation>
 		<annotation cp="🐚">ကြောင်လိမ် | ခရုခွံ</annotation>
 		<annotation cp="🐚" type="tts">ခရုခွံ</annotation>
@@ -1933,7 +1933,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🚏" type="tts">ဘတ်စ်ကား မှတ်တိုင်</annotation>
 		<annotation cp="🛣">ကားလမ်း | ဟိုင်းဝေး | အမြန်လမ်း</annotation>
 		<annotation cp="🛣" type="tts">ကားလမ်း</annotation>
-		<annotation cp="🛤">သံလမှး သံလမှး၊ ရထှား သံလမှး သံလမှး သံလမှး သံလမွှး သံလမွျး သံလမျး | သံလမ်း</annotation>
+		<annotation cp="🛤">သံလမှး သံလမှး၊ ရထှား သံလမှး သံလမှး သံလမှး သံလမှး သံလမြှး သံလမြျး သံလမျး</annotation>
 		<annotation cp="🛤" type="tts">သံလမ်း</annotation>
 		<annotation cp="🛢">စည်ပိုင်း | ဆီ</annotation>
 		<annotation cp="🛢" type="tts">ဆီ စည်ပိုင်း</annotation>


macchiati (commented):
I think we might want to hold back on integrating. I've heard that there are some problems with the Zawgyi conversion.

btangmu:
> I think we might want to hold back on integrating. I've heard that there are some problems with the Zawgyi conversion.

OK, I'll postpone further CLDRModify work until getting the go-ahead

srl295:
@sven-oly and @sffc can you take a look at this, and also at my comments on https://unicode-org.atlassian.net/browse/CLDR-13956 ? I am still waiting for your feedback on CDR-13956, related to this.

sffc:
@justinseinlin: does it look like these strings will be properly represented in Unicode after this PR?

common/annotations/my.xml
```diff
@@ -1381,7 +1381,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🐡" type="tts">ငါးပူဖောင်း</annotation>
 		<annotation cp="🦈">ငါး | ငါးမန်း</annotation>
 		<annotation cp="🦈" type="tts">ငါးမန်း</annotation>
-		<annotation cp="🐙">ဘဝှဲ ရဘဝှေဲ ဘဝှဲ ဘဝှဲ ဘဝွဲ ဘဝျဲ ဘဝဲ | ဘဝဲ</annotation>
+		<annotation cp="🐙">ဘဝှဲ ရဘဝှေဲ ဘဝှဲ ဘဝှဲ ဘဝှဲ ဘဝြဲ ဘဝျဲ ဘဝဲ</annotation>
```
justinseinlin, line 1384: These are still bad.  

Especially "ဘဝြဲ ဘဝျဲ ", I don't think I can read that as a Burmese native speaker.

common/annotations/my.xml
```diff
@@ -1933,7 +1933,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🚏" type="tts">ဘတ်စ်ကား မှတ်တိုင်</annotation>
 		<annotation cp="🛣">ကားလမ်း | ဟိုင်းဝေး | အမြန်လမ်း</annotation>
 		<annotation cp="🛣" type="tts">ကားလမ်း</annotation>
-		<annotation cp="🛤">သံလမှး သံလမှး၊ ရထှား သံလမှး သံလမှး သံလမှး သံလမွှး သံလမွျး သံလမျး | သံလမ်း</annotation>
+		<annotation cp="🛤">သံလမှး သံလမှး၊ ရထှား သံလမှး သံလမှး သံလမှး သံလမှး သံလမြှး သံလမြျး သံလမျး</annotation>
```
justinseinlin, line 1936: These are not good either.  All of them are garbled although I think this is more like "a bunch of unicode characters being randomly put together" kind of incorrect.

common/annotations/my.xml
```diff
@@ -1933,7 +1933,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🚏" type="tts">ဘတ်စ်ကား မှတ်တိုင်</annotation>
 		<annotation cp="🛣">ကားလမ်း | ဟိုင်းဝေး | အမြန်လမ်း</annotation>
 		<annotation cp="🛣" type="tts">ကားလမ်း</annotation>
-		<annotation cp="🛤">သံလမှး သံလမှး၊ ရထှား သံလမှး သံလမှး သံလမှး သံလမွှး သံလမွျး သံလမျး | သံလမ်း</annotation>
```
justinseinlin, line 1936: In fact, this is the only correct spelling for the given emoji.  The rest are all incorrect and I believe a lot of them repeating themselves.

macchiati:
I think some of this might be the result of incorrect application of the
Zawgyi converter to the user input. @Shane Carr <sffc@google.com> @Craig
Cornelius <ccornelius@google.com> , should we give up on that converter?

Mark

sffc:
According to Justin, there are two problems:

1. The word "octopus" is repeated several times in a row
2. The text is garbled

The text being garbled is possibly a result of these strings falling into the ambiguous range, and thereby being converted multiple times across several CLDR releases.

Since CLDR involves manual review, though, I would suggest not automatically running the converter, and only using the detector to flag possible errors for human review.

macchiati:
Thanks for the feedback. We can avoid running the converter in the next
release.

Good idea for warnings, but unfortunately we've found that the translators
mostly ignore those. We can have errors (which disallow submission), but
then we have the "non 100% correct detection" problem. ...

Mark

sffc:
I find a good way to deal with the "people ignore warnings" problem is to make the warnings errors but give a way to explicitly suppress the errors.  In code you can do that via code comments like `#[allow(unused_variable)]` to disable an unused-variable lint.

macchiati:
Good suggestion. We considered suppressing errors before, but this may give
us the impetus (although there are details to work out). Filed
https://unicode-org.atlassian.net/browse/CLDR-14549

Mark

btangmu:
I'm closing this PR and adding a link to it in https://unicode-org.atlassian.net/browse/CLDR-14549

macchiati:
Good, thanks

PR #1072, opened 2021-02-24
CLDR-14538 Modernize: improve Vue routing

srl295:
CLDR-14538

- simplify Vue routing
- add an UnknownPanel page
(wip, doesn't fix vsummary yet)


srl295:
@btangmu let me know if this updated approach seems reasonable so far

tools/cldr-apps/src/main/webapp/js/esm/cldrGenericVue.js
```diff
@@ -38,7 +49,7 @@ function errorHandler(err) {
 }
 
 function getHtml(json) {
-  let html = cldrStatus.logoIcon();
+  let html = "";
   return html;
```
btangmu, line 53: delete getHtml if is serves no purpose?

tools/cldr-apps/src/main/webapp/js/esm/cldrLoad.js
```diff
@@ -1,7 +1,7 @@
 /*
  * cldrLoad: encapsulate functions for loading GUI content for Survey Tool
  */
-import * as cldrAbout from "./cldrAbout.js";
+import * as cldrGenericVue from "./cldrGenericVue.js";
 import * as cldrAccount from "./cldrAccount.js";
 import * as cldrAdmin from "./cldrAdmin.js";
```
btangmu, line 6: alphabetize imports? better than random order

tools/cldr-apps/src/main/webapp/js/esm/cldrGenericVue.js
srl295: @btangmu I was trying to keep the churn as small as possible on the cldr*.js files. Removing the 'loader' caused a startup failure.  but, I'll keep cutting away at it.

Also, `cldrStatus.logoIcon()` could actually be accessed via the exports now. I'll try that.

tools/cldr-apps/src/main/webapp/js/esm/cldrGenericVue.js
```diff
@@ -1,33 +1,44 @@
 /*
- * cldrAbout: encapsulate functions for the "About" page of Survey Tool
+ * cldrGenericVue: encapsulate functions for the any other special pages of Survey Tool
+ * which route through Vue
  */
 import * as cldrLoad from "./cldrLoad.js";
 import * as cldrRetry from "./cldrRetry.js";
 import * as cldrStatus from "./cldrStatus.js";
+import * as cldrEvent from "./cldrEvent.js";
 import * as cldrSurvey from "./cldrSurvey.js";
```
btangmu, line 10: alphabetize imports

btangmu (dismissed):
looks like a good improvement; might be improved further by moving `specialToComponentMap` from index.js to a new module such as cldrVue.js -- the essential concern of index.js is to make a bundle, not to map vue components

srl295:
> looks like a good improvement; might be improved further by moving `specialToComponentMap` from index.js to a new module such as cldrVue.js -- the essential concern of index.js is to make a bundle, not to map vue components

This is a good point.

I don't expect that many more "top level" exports though, everything else should be internally imported between parts of the code. But agree that would be an improvement, even if index.js ends up with a small number of imports in it.

srl295:
@btangmu let's go ahead and merge this, and i'll do the actual VV changes separately.   This way these improvements can get out there.

srl295:
> @btangmu let's go ahead and merge this, and i'll do the actual VV changes separately. This way these improvements can get out there.

this will make it easier for me to work on multiple PRs in parallel from here on

btangmu:
> let's go ahead and merge this

It's still "draft" -- would you like me to press the ready-for-review button?

tools/cldr-apps/js/src/specialToComponentMap.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Get a list of available specials
+ * @returns String[] list of specials
+ */
+function listSpecials() {
+  return Object.keys(specialToComponentMap);
+}
+
+export { specialToComponent, listSpecials };
```
btangmu, line 37: could this go in webapp/js/ems like most js files?

btangmu (approved):

tools/cldr-apps/js/src/cldrVueRouter.js
```diff
@@ -0,0 +1,42 @@
+import { specialToComponent } from "./specialToComponentMap";
```
btangmu, line 1: this file likewise could go in ems? there doesn't seem to be a benefit to having some js files in one folder and some in another, except maybe index.js which is specifically for bundling

btangmu (approved):

tools/cldr-apps/js/src/cldrVueRouter.js
srl295: So I'd think where we both want to go is to have the front end be a static single page app. 

Such an app should be hostable, and even testable, completely separate from the .jar

So I'd want to move all JavaScript, CSS, templates into one place.  We wouldn't need webapp/ other than a place to put the bundled output.

Obviously we should discuss this some more.

btangmu: If "static" in "static single page app" means that the back end doesn't dynamically generate html, js, etc., then yes, that's the right direction. It should remain possible and easy for the back end to serve distinct development and production forms of the front end.

For loose coupling, 99.99% of our code, and how it's organized, shouldn't depend on how the code is bundled or whether it's bundled at all.

The ability to bypass the bundler for development is valuable. Let's not lose it. I've already found that the browser debugger doesn't always work quite as well with the bundle as with the source. For example, it sometimes gives the wrong line for an error message. Such problems are likely to intensify with minified/optimized bundles. We should also be prepared for the possibility of bugs that are introduced in the bundling process, and hidden unnecessary dependencies on the bundler.

Putting the source files together in one place is good. For maximum flexibility and loose coupling, it should be a place that's accessible not only by webpack. We've seen that webpack has no problem making the bundle using source files that happen to be located inside webapp/js/ems. So this location seems to have important advantages and zero drawbacks. It seems like putting all the source outside of webapp would make us 100% coupled to webpack, and prevent us from running directly from the source, even for development, unless/until we find some alternative method. I'm totally open to the possibility of finding such methods. Whenever it becomes advantageous, overall, to move files, we can move them. In the meantime, nearly all of our modern js files are now in the same folder, consistently follow a naming convention `cldrFoo.js`, and consistently import from each other with syntax `import * as cldrFoo from "./cldrFoo.js";`. Regardless of how the names, locations, conventions, etc., may change in the future, there doesn't seem to be any need or benefit to making exceptions for some files like `specialToComponentMap.js` -- though of course there are many aspects of the situation that aren't clear to me.

srl295: > If "static" in "static single page app" means that the back end doesn't dynamically generate html, js, etc., then yes, that's the right direction.

:+1:

For testing, it would be good to decouple that same front end so that it can be launched and served, and tested, independent from the dynamic part.

Imagine that static-surveytool.unicode.org  serves the JS and HTML, and st.unicode.org is ONLY the APIs.  Not saying we need to do that or that it's a good idea, but that's the thought process.

What I'd like to avoid is the possibility that the front end includes somethign outside the bundle, when we want everything going through the bundle.

What if there was a 'manual bundler' that just copied the same files over (using `/bin/cp`) to the `/webapp` directory?  Then you could import `index.js` etc. directly. Would that work?

btangmu: maybe that would work -- just an extra step

your concern about avoiding hidden dependencies on code outside the bundle makes sense -- currently we have lots of those in webapp, right? but I don't think ems is one of them: when running with the bundle, I've watched the network tab carefully for any download of the source files in ems and there's none

PR #1075, opened 2021-02-25
CLDR-14461 update unitPreferencesTest.txt for changes in CLDR-14515

pedberg-icu:
### 

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14461

The unitPreferencesTest.txt file, which comes from CLDR but is used for ICU tests, needed to be updated due to the changes for mixed metric units in [CLDR-14515](https://unicode-org.atlassian.net/browse/CLDR-14515).


macchiati (approved):

hugovdm (commented):
Thanks! (Ticking off PRs in my "closed review requests" list.)

PR #1076, opened 2021-03-01
CLDR-14540 Modernize Vetting Participation

btangmu:
-New cldrVettingParticipation.js

-Copy and adapt old code from js/special/vetting_participation.js

-Revise cldrVettingParticipation.js so it works

-SurveyVettingParticipation.java delete dead code, make some items final/private, reformat

-Add cache-busting extension for bundle and css in SurveyTool.java

-Fix bug when user id is present in votes table, absent from users table; related debugging

-Fix locale und bug in SurveyAjax.listLocalesForImportOldVotes

-Enable some html in cldrAccount.js to pass xml unit test

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14540

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -1618,6 +1618,9 @@ public class SurveyAjax extends HttpServlet {
          */
         JSONObject oldVotesNull = null;
         localeCount.forEach((loc, count) -> {
+            if ("und".equals(loc)) {
+                return;
+            }
             /*
              * We may get realCount <= count due to filtering. If we catch an exception here thrown in
              * viewOldVotes, use the value already in count, still enabling the user to select the locale.
--- a/tools/cldr-apps/src/main/webapp/js/esm/cldrVettingParticipation.js
+++ b/tools/cldr-apps/src/main/webapp/js/esm/cldrVettingParticipation.js
@@ -1,0 +1,304 @@
+/*
+ * cldrVettingParticipation: encapsulate Survey Tool Vetting Participation code.
+ */
+import * as cldrAccount from "./cldrAccount.js";
+import * as cldrAjax from "./cldrAjax.js";
+import * as cldrInfo from "./cldrInfo.js";
+import * as cldrLoad from "./cldrLoad.js";
+import * as cldrRetry from "./cldrRetry.js";
+import * as cldrStatus from "./cldrStatus.js";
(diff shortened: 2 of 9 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1618,6 +1618,9 @@ public void listLocalesForImportOldVotes(User user, SurveyMain sm,
          */
         JSONObject oldVotesNull = null;
         localeCount.forEach((loc, count) -> {
+            if ("und".equals(loc)) {
```
srl295, line 1621: why?

tools/cldr-apps/src/main/webapp/js/esm/cldrVettingParticipation.js
```diff
@@ -0,0 +1,304 @@
...
+            // Note: can't use locmap to get a translation here, because locmap only
+            // has extant CLDR locales, and by definition 'loc' is not in CLDR yet.
+            text: `${loc}`, // raw code
+          })
+        );
+      }
+    }
+  }
+
+  // Chapter 1
```
srl295, line 116: Chapter 1?

srl295 (approved):
Generally LGTM with questions

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
btangmu: Without this, CLDRLocale.getInstance throws an exception for "und"

Given:
`private static final ULocale UNKNOWN_LOCALE = new ULocale("und");`
I don't know what it would mean to import old votes for UNKNOWN_LOCALE, so it seems better to skip it rather than throw an exception during import old votes.

tools/cldr-apps/src/main/webapp/js/esm/cldrVettingParticipation.js
btangmu: I copied that along with the rest of the code in special/vetting_participation.js -- your guess is at least as good as mine what it means! If it means nothing to you, I'll delete it...

Maybe it means "here is the start of one of the first main sections of code in this function"? Often a comment in the middle of an excessively long function suggests the need for a subroutine; hopefully a better name could be invented than `chapterOne()`

srl295: Aha, The original had one chapter and an epilogue.  I think it was just a section marker, and yes, it could be a subroutine.

PR #1077, opened 2021-03-01
CLDR-14538 Modernize VV

srl295:
First cut, just get it up and running.


srl295:
like About, this has a basic UI but is a lot easier to tweak in Vue.

image

srl295:
the api is `POST ​/api​/summary`

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -547,6 +547,7 @@ public CheckCLDR handleCheck(String path, String fullPath, String value, Options
                         r.put(SurveyMain.QUERY_VALUE_SUFFIX, mySession.settings().get(pref, null));
                         send(r, out);
                     } else if (what.equals(WHAT_VSUMMARY)) {
+                        // TODO: obviated by org.unicode.cldr.web.api.PriorityItems
```
srl295, line 550: the 'old' JS could probably call the new API and we delete this.  I'm inclined to leave it alone until we remove dojo.

srl295:
> great to see more of the new api!

yes! it's definitely getting easier…

PR #1082, opened 2021-03-03
CLDR-14554 Modernize My Account; maybe feature-complete

btangmu:
-Extensive changes in UserList.java and cldrAccount.js

-Shorten long functions by adding subroutines

-Avoid onsubmit in html: same problem as onclick for ESM

-Change bulk-action button name from Change to Do Action, for consistency

-New getInterestLocales, etc., in cldrAccount.js for setting interest locales

-New api/intlocs endpoint on back end using Liberty Server features

-New api/InterestLocales.java and IntLocsRequest.java

-New UserRegistry.userCanSetInterestLocales

-New version of UserRegistry.setLocales uses CookieSession instead of WebContext

-In cldrAjax.js, for api only, use application/json not application/x-www-form-urlencoded

-Delete unused errorHandler in cldrGenericVue.js

-Comments; clean-up

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14554

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/InterestLocales.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/InterestLocales.java
@@ -1,0 +1,89 @@
+package org.unicode.cldr.web.api;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
+import org.unicode.cldr.web.CookieSession;
+import org.unicode.cldr.web.UserRegistry;
+
+@Path("/intlocs")
+public class InterestLocales {
+
+    @POST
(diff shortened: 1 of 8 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/InterestLocales.java
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
+import org.unicode.cldr.web.CookieSession;
+import org.unicode.cldr.web.UserRegistry;
+
+@Path("/intlocs")
```
srl295, line 22: your first API with jax-rs 🎉

tools/cldr-apps/src/main/webapp/js/esm/cldrAjax.js
```diff
@@ -145,6 +145,9 @@ function sendXhr(xhrArgs) {
   if (xhrArgs.postData || xhrArgs.content) {
     options.method = "POST";
     options.data = xhrArgs.postData ? xhrArgs.postData : xhrArgs.content;
+    if (typeof options.data === "object" && xhrArgs.url.includes("/api/")) {
```
srl295, line 148: is this `/api/` check needed given `options.makeJsonPost` ?

srl295 (approved):
LGTM with a comment, fine to merge…

tools/cldr-apps/src/main/webapp/js/esm/cldrAjax.js
btangmu: if `options.makeJsonPost` were set by the caller in cldrAccount then it wouldn't need to be set here, but currently it's only set here in cldrAjax

maybe it would be better for the caller to set `options.contentType` -- anyway this is a hack and there's a lot of room for improvement in making standardizing/simplifying/encapsulating how we set up ajax calls

PR #1083, opened 2021-03-03
CLDR-12003 updating translation

stenshamn:
##### Checklist

- [x ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-12003
- [x ] Updated PR title and link in previous line to include Issue number


macchiati (approved):

pedberg-icu (approved):
Ahh, much better, thanks!

PR #1086, opened 2021-03-04
CLDR-14569 Improve Windows mappings for UTC

mattjohnsonpint:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14569

UTC should be mapped primarily to `Etc/UTC` so that when fetching display names after mapping one gets "Coordinated Universal Time" instead of "Greenwich Mean Time". 

`America/Danmarkshavn` belongs with the other localities that use GMT without DST, rather than with UTC.  It's display names will still get "Greenwich Mean Time", as all of those entries use the "GMT" metazone.


jefgen (dismissed):
Based on talking with @mattjohnsonpint offline, the change lgtm. 

Thank you @mattjohnsonpint! :)

yumaoka (approved):
LGTM. Please create ICU ticket also.
CLDR-39 maintenance branch is already created. You may want to check with Kristi if it can be merged into CLDR-39. If so, you may need to re-create PR against maint-39 instead of master.

mattjohnsonpint:
@yumaoka - I've created the ICU ticket here: https://unicode-org.atlassian.net/browse/ICU-21528

jefgen (approved):
Thanks @mattjohnsonpint !

mattjohnsonpint:
I rebased and retargeted this to the `maint/maint-39` branch.  Is there an automated process to bring this into `master`, or will I need a separate PR for that after this one?

jefgen:
> Is there an automated process to bring this into `master`, or will I need a separate PR for that after this one?

Good question. From asking others, I believe the new approach that CLDR is using is similar to what ICU does:
After a release is finished the `maint/maint-*` branch is merged into the `master` branch.
So any/all of the individual commits made to the maintenance branch will eventually get brought into `master`, without each individual PR author having to create a separate PR.

PR #1088, opened 2021-03-04
CLDR-14557 Modernize look up (browse); first draft

btangmu:
-New LookupPath.vue partly working but not modern yet, js accesses dom in old way

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14557

### DIFF
--- a/tools/cldr-apps/js/src/views/LookupPath.vue
+++ b/tools/cldr-apps/js/src/views/LookupPath.vue
@@ -1,0 +1,134 @@
+<template>
+  <form action="">
+    <label>Base Locale: <input name="loc" value="en-US" /></label
+    ><input type="submit" value="change" />
+  </form>
+  <br />
+  <hr />
+  <table>
+    <tbody>
+      <tr>
+        <td style="border-right: 1px solid gray">
+          What is...
+          <input
+            id="whatis"
+            v-on:change="lookup_whatis()"
+            style="font-size: x-large"
+          />
+        </td>
+        <td>
+          <b>xpath calculator - </b><br />
+          <label
+            >XPath:<input
+              id="xpath"
+              v-on:change="lookup_xpath('xpath')"
+              size="160" /></label
+          ><br />
+          <label
+            >PathHeader:<input
+              id="pathheader"
+              size="160"
+              disabled="true" /></label
+          ><br />
+          <label
+            >XPath strid:<input
+              id="strid"
+              v-on:change="lookup_xpath('strid')"
+              size="32"
+          /></label>
+          <label
+            >XPathID (dec#)<input
+              id="xpathid"
+              v-on:change="lookup_xpath('xpathid')"
+              size="8"
+          /></label>
+          <div id="xpath_answer" style="font-style: italic">
+            enter a value and hit the tab key to begin
+          </div>
+        </td>
+      </tr>
+    </tbody>
+  </table>
+  <div id="whatis_answer"></div>
+  <hr />
+  <div class="helpHtml" style="margin: 2em">
+    <h4>Instructions:</h4>
+    <b>What Is...</b>: Enter a code or a portion of a name in the "What Is"
+    field, such as "jgo" or "English", and press the Tab key. A list of matching
+    codes will be shown.
+    <p>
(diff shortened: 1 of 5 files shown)


tools/cldr-apps/js/src/views/LookupPath.vue
```diff
@@ -0,0 +1,134 @@
+<template>
+  <form action="">
```
btangmu, line 2: this html is derived from browser, running on production server

main change is "v-on:change"

formatted with npx prettier (default options)

tools/cldr-apps/js/src/views/LookupPath.vue
```diff
@@ -0,0 +1,134 @@
...
+      into the XPath field, and press the Tab key. Or, enter an XPath strid,
+      such as <tt>1d142c4be7841aa7</tt> into the XPath strid field and press the
+      Tab key. The other fields (if applicable) will be filled in.
+    </p>
+  </div>
+</template>
+
+<script>
+import * as cldrAjax from "../../../src/main/webapp/js/esm/cldrAjax.js";
+
```
btangmu, line 71: script is from browser (like the html above), and partly from browse.jsp (which has newer code than what is running on production, for api/xpath)

tools/cldr-apps/js/src/views/LookupPath.vue
```diff
@@ -0,0 +1,134 @@
...
+    },
+
+    lookup_xpath: function (from) {
+      var v = document.getElementById(from).value;
+      if (v.length == 0) {
+        return;
+      }
+      var r = document.getElementById("xpath_answer");
+      r.innerHTML = "<i>Looking up xpath " + v + "...</i>";
+      const apiMap = {
```
btangmu, line 100: using innerHTML like this is probably a no-no in a vue file, though it seems to work

should instead have data items loaded using `{{ ... }}` notation...

tools/cldr-apps/js/src/views/LookupPath.vue
```diff
@@ -0,0 +1,134 @@
...
+      var v = document.getElementById("whatis").value;
+      var r = document.getElementById("whatis_answer");
+      if (v.length == 0) {
+        r.innerHTML = "";
+        return;
+      }
+      r.innerHTML = "<i>Looking up " + v + "...</i>";
+      cldrAjax.sendXhr({
+        url: "/cldr-apps/browse_results.jsp?loc=en_US&q=" + v,
+        load: function (h) {
```
btangmu, line 84: hard-coded en_US is wrong here, should be from the "Base Locale" control above

srl295: see the vsummary component: you can pull it in from `cldrStatus`…

tools/cldr-apps/js/src/views/LookupPath.vue
```diff
@@ -0,0 +1,134 @@
...
+    lookup_whatis: function () {
+      var v = document.getElementById("whatis").value;
+      var r = document.getElementById("whatis_answer");
+      if (v.length == 0) {
+        r.innerHTML = "";
+        return;
+      }
+      r.innerHTML = "<i>Looking up " + v + "...</i>";
+      cldrAjax.sendXhr({
+        url: "/cldr-apps/browse_results.jsp?loc=en_US&q=" + v,
```
srl295, line 83:
```suggestion
        url: "./browse_results.jsp?loc=en_US&q=" + v,
```

shouldn't assume /cldr-apps (remember /smoketest?)

btangmu: anyway the jsp has to go, it's not for ajax

maybe new api/whatis

btangmu: btw api/xpath works for api/xpath/hex or api/xpath/dec but it can't take an xpath string and produce the hex or dec yet, can it?

srl295: it should be able to 

please go ahead and modify XPathApi.java to suit whatever you want, this is the only client of it

btangmu:
See https://github.com/unicode-org/cldr/pull/1089

PR #1091, opened 2021-03-06
CLDR-14576 common/bg.xml: adjust translations of MMM

dilyanpalauzov:
The problem was that three MMM (word) was converted
to two MM (number) which is not what was requested.

Not sure if
 <dateFormatItem id="MMM">MMM</dateFormatItem>
has added value.


pedberg-icu:
This is the intended behavior. There are several locales, including Bulgarian, that prefer not to use abbreviated month names; instead MMM maps to numeric month (and then MMMM maps to full month names)

PR #1092, opened 2021-03-08
CLDR-14573 - Add full data of Kaingang and Nheengatu locales to CLDR

robertgmelo:
CLDR-14573 - Add full data of Kaingang and Nheengatu locales to CLDR
- Include kgp and yrl locale data
- Include supplemental data for kgp and yrl
- Update country/language population data

##### Checklist


srl295:
@robertgmelo might be better to split `yrl` (Nheengatu) to a separate PR, as it's completely new

robertgmelo:
Ok, what about supplemental, language data?

PR #1093, opened 2021-03-08
CLDR-14573 - Add full data of Kaingang and Nheengatu locales to CLDR

robertgmelo:
CLDR-14573 - Add full data of Kaingang and Nheengatu locales to CLDR

- Determine coverageLevels for kgp and yrl locales

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-CLDR-14573


robertgmelo:
Please note I've created a separated PR in order to confirm the proper coverage level for those locales. Actually, I've based on pt, since locale data of those locales were based on Portuguese locale data.

PR #1094, opened 2021-03-08
CLDR-14573 Add full data of Kaingang locales to CLDR

robertgmelo:
CLDR-14573 Add full data of Kaingang  locales to CLDR

- Include kgp full locale data
- Include kgp keyboards

##### Checklist


DavidLRowe:
I'm trying to figure out what is required to add a locale (based on PRs for other added locales), so what I have are questions that will need someone else to answer!

Does common/supplemental/ordinals.xml need an entry for kpg?
Does common/supplemental/plurals.xml need an entry for kpg?
Do we need common/casing/kpg.xml?
Does tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java need kpg added to Pattern 'language100'?

robertgmelo:
Hello David, let me see if I can help you. In terms of what is required to add a new locale,  you can check these two references below, they were very helpful to me:
- http://cldr.unicode.org/development/adding-locales
- http://cldr.unicode.org/index/cldr-spec/minimaldata

I think _ordinals.xml, plurals.xml, casing_ are not really **mandatory**. On kgp/yrl cases, as there are no special rules, it wasn't necessary (it follows root rules).

From my personal experience, fill _dayPeriods.xml_ is necessary to avoid failures on tests.

Add **language code** on _attributeValueValidity.xml_ is required, as described in the reference above.

On https://www.unicode.org/reports/tr35/tr35-info.html#Coverage_Levels, we can check more about the coverageLevels, but from my understanding, CLDR technical commitee is able to define the proper value for the language. 100 is only when everything of the language is covered (rare), so I suppose `TestCoverageLevel` does not require the value 100 is it's on common/main. Of course, depending of the value you include, it has implications on test results, once it will check to make sure the coverage for such level is being respected. It's good to check with committee before. 

**Note:** It's also important to mention these PRs (#1094, #1095, #1096) are exceptional flows to the conventional process (through Survey Tool), so the coverage level value will be proper defined when time to move from **seed** to **common** comes.

I hope it can help.

PR #1095, opened 2021-03-08
CLDR-14573 Add full data of Nheengatu locales to CLDR

robertgmelo:
CLDR-14573 Add full data of Nheengatu locales to CLDR

- Include yrl full locale data
- Include yrl keyboards

##### Checklist

### DIFF
--- a/common/main/yrl.xml
+++ b/common/main/yrl.xml
@@ -1,0 +1,9322 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2021 Unicode, Inc.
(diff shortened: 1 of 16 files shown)


DavidLRowe:
I'm trying to figure out what is required to add a locale (based on PRs for other added locales), so what I have are questions that will need someone else to answer!

Does common/supplemental/ordinals.xml need an entry for yrl?
Does common/supplemental/plurals.xml need an entry for yrl?
Do we need common/casing/yrl.xml?
Does tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java need yrl added to Pattern 'language100'?
It's not clear to me where the decision is made to add a language to the $language variable in attributeValueValidity.xml. Does this automatically include everything in common/main?

PR #1096, opened 2021-03-08
CLDR-14573 Add supplemental data for Kaingang and Nheengatu locales to CLDR

robertgmelo:
CLDR-14573 Add supplemental data of Kaingang and Nheengatu locales to CLDR

- Determine coverageLevels for kgp and yrl locales
- Include supplemental data for kgp and yrl
- Update country/language population data

Also see:
Full data for Kaingang is in #1094 
Full data for Nheengatu is in #1095 

##### Checklist


common/supplemental/coverageLevels.xml
```diff
@@ -92,7 +92,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageVariable key="%keys80" value="(calendar|cf|collation|currency|em|fw|hc|lb|lw|ms|rg|ss|numbers|d0|h0|i0|k0|m0|s0|t0|x0)"/>
 		<coverageVariable key="%keys100" value="(col(Alternate|Backwards|CaseFirst|CaseLevel|HiraganaQuaternary|Normalization|Numeric|Reorder|Strength)|kv|sd|timezone|va|variableTop|x)"/>
 		<coverageVariable key="%language30" value="und"/>
-		<coverageVariable key="%language40" value="(de(_(AT|CH))?|en(_(AU|CA|GB|US))?|es(_(ES|419|MX))?|fr(_(CA|CH))?|it|ja|pt(_(BR|PT))?|ru|zh(_(Hans|Hant))?)"/>
+		<coverageVariable key="%language40" value="(de(_(AT|CH))?|en(_(AU|CA|GB|US))?|es(_(ES|419|MX))?|fr(_(CA|CH))?|it|ja|kgp(_(BR))?|pt(_(BR|PT))?|ru|yrl(_(BR|CO|VE))?|zh(_(Hans|Hant))?)"/>
```
macchiati, line 95: This is not correct: we only translate a small number of languages at level 40. It should go to modern instead.

common/supplemental/supplementalData.xml
```diff
@@ -1734,7 +1734,7 @@ XXX Code for transations where no currency is involved
 		<language type="kg" scripts="Latn"/>
 		<language type="kg" territories="CD" alt="secondary"/>
 		<language type="kge" scripts="Latn"/>
-		<language type="kgp" scripts="Latn"/>
+		<language type="kgp" scripts="Latn" territories="BR"/>
```
macchiati, line 1737: This data will be overwritten when the tools are run. See line 1270. I believe that the tools will add the territories that are in common/main, but that needs to be checked.

macchiati (changes requested):

common/supplemental/coverageLevels.xml
robertgmelo: Perfect, Mark. As I've mentioned in the CR, I've separated the PRs exactly to confirm the exact info. 
I'll move to modern (level 80) as per TR35.
Thanks for confirming.

robertgmelo: Mark, one point here is that moving to modern, the test TestLanguageNameCoverage will fail because we still don't have localized language names for those languages. Based on that, what would it be the recommendation? Proceed with the change?
One note here: to deal with that on Android, internally we've gathered the language names with native translators to avoid display the language code, but of course, it does contemplate a subset of CLDR locales. 

Please let me know your thoughts.

common/supplemental/supplementalData.xml
robertgmelo: I've checked instructions available on 1270. When we run `ConvertLanguageData.java`, three files are generated (country_language_population_raw.txt, language_script_raw.txt, supplementalData.xml). On `supplementalData.xml`, the territories are not added because the new languages do not meet the existing rules (major/official status, at least 20% of whole population, or 1M), so it's not considered as primary/secondary.

Is **ConvertLanguageData** being used to generated supplementalData.xml?

Maybe we should remove this, but want to hear from you.

common/supplemental/coverageLevels.xml
macchiati: I think we need the language names in English, but not the other languages. But I'm kinda swamped this week, so I won't be able to spend any time on answering your (very sensible) questions.

You can ping me Monday by email, if that's ok.

robertgmelo: Noted. I can send an email next Monday. tks

macchiati (dismissed):

macchiati:
@srl295 can you check why the checks are failing?

srl295:
@macchiati 

```

TestCLDRLocaleCoverage {
253
    TestCLDROrganizationPresence (0.000s) Passed
254
    TestCldrSuperset (0.002s) ALERT (1 warning(s))
255
    TestLanguageNameCoverage {
256
      Error: (TestCLDRLocaleCoverage.java:96) : localesForNames.containsAll(coverageLocales): expected true
257
    } (0.066s) FAILED (1 failure(s), 4 warning(s))
258
  } (0.068s) FAILED (1 failure(s), 5 warning(s))
259
```

Also the ticket needs to be accepted I think

robertgmelo:
> @robertgmelo take out the extra " - " in the PR and commit title
> 
> Just make it "[CLDR-14573](https://unicode-org.atlassian.net/browse/CLDR-14573) Add …"

Done.

robertgmelo:
@srl295,

Please note the PRs #1094 and #1095 depend of this PR, due to supplemental data.

Regarding locale coverage name issue, this is about the language name of Kaingang and Nheengatu in other locales. We have this information for some locales, for others we've kept Kaingang, to avoid display kgp/yrl.

What would it be the best approach here? Wait ServeyTool process, or we could include those content via PR.

Please share your thoughts and let me know if I can help.

macchiati:
The test is breaking because it checks for the existence of the locale in the main locale folder. 

We need to make this process much simpler, but as it is, the PR that adds a locale needs to include both the basic locale file plus any supplemental info so that the tests of each of those succeed.

srl295:
Yes, That’s the way it was and I suggested splitting the coverage change.
Should it change back? At least, I think the two languages should each be
in their own pr

robertgmelo:
Initially I created one PR, but I kinda agreed with Steven that having in separate PRs would be more organized, so we would have separate commits for each locale. However, as supplemental data is also part of locale (dayPeriods.xml), it may be difficult to separate them. I'm not sure if structural changes on tests would be beneficial right now.

Indeed, one commit would be simpler and easy to go, but probably because I'm submitting the **whole locale data**, _which is not part of the conventional process, correct?_ As Mark commented in the e-mails, this can also serves as a learning process so we can try to improve the process.

If it's preferable as a unique PR, no problem at all, I can do. 

Please let me know your thoughts.

macchiati:
It's not the conventional process, but we're allowing it in this case.

>

macchiati (dismissed):

macchiati:
One of the tests failed:
    TestLanguageNameCoverage {
      Error: (TestCLDRLocaleCoverage.java:96) : localesForNames.containsAll(coverageLocales): expected true
    } (0.066s) FAILED (1 failure(s), 4 warning(s))

robertgmelo:
Sorry for the delay, Mark.

It happens because of the language names that are present in the other PRs (#1094 #1095). As we decided to separate the PRs from supplemental changes, it's _kinda_ expected to happen :( 

As far as I remember, with the 3 PRs, we do not see failures, but I can double-check again.

DavidLRowe:
A question and a comment (from someone trying to figure out how this works!):

In the changes to coverageLevels.xml, I wonder if the region code is needed, since no other locale includes it.

The changes to common/supplemental/supplementalData.xml occur in a "generated" section of that file. I guess that's not wrong, but presumably that data will be overwritten by new data generated from the changes in country_language_population_raw.txt.

DavidLRowe:
This PR is failing a test: CLDR/TestCLDRLocaleCoverage/TestLanguageNameCoverage

If this PR can be corrected and merged, then (presumably) the supplemental data will be in place for #1094  and #1095 to pass the required tests.

DavidLRowe:
the line in TestCLDRLocaleCoverage.java where the error is detected:
```
// If this fails, it is because of a mismatch between coverage and the getCLDRLanguageCodes.
// Usually a problem with coverage.
boolean showRegex = !assertContains("localesForNames.containsAll(coverageLocales)", localesForNames, coverageLocales);
```

PR #1098, opened 2021-03-09
CLDR-14389 fix RTL display (modernize Dashboard)

srl295:
- added new custom component <cldr-value/> for CLDR value
- some UI cleanup in the Dashboard


btangmu (approved):
rslgtm

PR #1106, opened 2021-03-11
CLDR-14461 exclude "for-cldr" deps from mvn license

srl295:
CLDR-14461
- these are under the same Unicode license, Unicode-DFS-2016
- THIRD_PARTY.properties isn't needed anymore (so won't be autogenerated).

see ICU-21425 and ICU-12811


tools/cldr-code/src/license/THIRD-PARTY.properties
```diff
@@ -1,16 +0,0 @@
...
-# - The MIT License
-# - The SAX License
-# - The W3C License
-#-------------------------------------------------------------------------------
-# Please fill the missing licenses for dependencies :
-#
-#
-#Fri Feb 19 13:26:13 EST 2021
-com.ibm.icu--utilities-for-cldr--69.1-SNAPSHOT-cldr-2021-02-17=
-com.ibm.icu--icu4j-for-cldr--69.1-SNAPSHOT-cldr-2021-02-17=
```
srl295, line 16: no longer needed due to exclusion.

pedberg-icu (approved):
Yay! This will save me from constantly having to recheck out an old version of this file to avoid irrelevant date updates.

PR #1115, opened 2021-03-16
CLDR-14554 Modernize My Account; prevent user changing own level

btangmu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14554


srl295 (approved):
Good, solves the client side ..

srl295:
Need server side restriction but can be a follow up

PR #1116, opened 2021-03-16
CLDR-14562 Fix About panel

srl295:
- linkify URLs
- link git hashes

minor improvement to waitingpanel

This will close out CLDR-14652

### DIFF
--- a/tools/cldr-apps/js/src/views/AboutPanel.vue
+++ b/tools/cldr-apps/js/src/views/AboutPanel.vue
@@ -5,7 +5,9 @@
       <a href="http://www.unicode.org/cldr">unicode.org/cldr</a>.
     </p>
     <h1>About this SurveyTool Installation</h1>
-    <table class="aboutBox">
+    <!-- spinner only shows if there's a delay -->
+    <a-spin tip="Loading" v-if="!aboutData" delay="500"/>
+    <table v-if="aboutData" class="aboutBox">
       <thead>
         <th>Property</th>
         <th>Value</th>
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/js/src/views/AboutPanel.vue
```diff
@@ -15,8 +17,23 @@
...
+                   base URL for the linkification -->
+              <a class="aboutValue" v-bind:href="aboutData.CLDR_COMMIT_BASE + v">
+                <i class="glyphicon glyphicon-cog" />{{ v }}
+              </a>
+            </span>
+            <span v-else class="aboutValue">
+              <!-- some other type -->
+              {{ v }}
+            </span>
           </td>
```
btangmu, line 37: wow!

btangmu (approved):

tools/cldr-apps/js/src/views/AboutPanel.vue
srl295: do you like this kind of a fix?

btangmu: I like it as a fun new way of doing things, and it seems fine in this context. It's good to know we can do this. When it starts getting too complicated, we might put more of the logic in `<script>` and less in `<template>`. The repetition of `span` and `aboutValue` would get cumbersome if we kept adding more `v-else-if`. Maybe a compromise would be like `v-if="shouldBeLink(v)" ... v-bind:href="makeLink(v)"` -- then there would be v-if and v-else but not v-else-if, and the details of `shouldBeLink` and `makeLink` would be in `<script>`

srl295: the td's content could be pulled out as a component, kind of a text-or-link object. 

i agree, it could be pulled out to a script - if it was really going to be reused

PR #1126, opened 2021-03-20
CLDR-14579 document changes in DTDs

macchiati:
Added a general description of @description on units, and unitConstant@description, unitQuantity@description, convertUnit@description, plus some additional values for convertUnit@systems. Did not add grammaticalCase values, since those were added in a previous PR.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14579

### DIFF
--- a/docs/ldml/tr35-info.md
+++ b/docs/ldml/tr35-info.md
@@ -759,6 +759,12 @@ The parentLocales data is supplemental data, but is described in detail in the [
 
 The unit conversion data ([units.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/units.xml)) provides the data for converting all of the cldr unit identifiers to base units, and back. That allows conversion between any two convertible units, such as two units of length. For any two convertible units (such as acre and dunum) the first can be converted to the base unit (square-meter), then that base unit can be converted to the second unit.
 
+Many of the elements allow for a common @description attribute, to disambiguate the main attribute value or to explain the choice of other values. For example:
+```xml
+<unitConstant constant="glucose_molar_mass" value="180.1557" 
+  description="derivation from the mean atomic weights according to STANDARD ATOMIC WEIGHTS 2019 on https://ciaaw.org/atomic-weights.htm"/>
+```
+
 ```xml
 <!ELEMENT unitConstants ( unitConstant* ) >
 
@@ -769,6 +775,8 @@ The unit conversion data ([units.xml](https://github.com/unicode-org/cldr/blob/m
 <!ATTLIST unitConstant value CDATA #REQUIRED >
 
 <!ATTLIST unitConstant status NMTOKEN #IMPLIED >
+
+<!ATTLIST unitConstant description CDATA #IMPLIED >
 ```
 
 ### Constants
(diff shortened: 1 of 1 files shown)


sffc (approved):

docs/ldml/tr35-info.md
```diff
@@ -807,6 +815,8 @@ An implementation need not use rationals directly for conversion; it could use d
 <!ATTLIST convertUnit factor CDATA #IMPLIED >
 
 <!ATTLIST convertUnit offset CDATA #IMPLIED >
+
+<!ATTLIST convertUnit description CDATA #IMPLIED >
```
sffc, line 819: Did you mean systems instead of description here?

docs/ldml/tr35-info.md
```diff
@@ -937,6 +962,8 @@ However, in conversion, it may be necessary to resolve them in order to find a m
 <!ATTLIST unitQuantity quantity NMTOKENS #REQUIRED >
 
 <!ATTLIST unitQuantity status NMTOKEN #IMPLIED >
+
+<!ATTLIST unitQuantity description CDATA #IMPLIED >
```
sffc, line 966: This one isn't documented specifically; do you consider it covered by the docs for description above?

docs/ldml/tr35-info.md
macchiati: I meant the description attribute. The goal here is to add the  <!ATTLIST XXX description ...> lines that were missing, for the applicable XXX's

docs/ldml/tr35-info.md
macchiati: Yes, I thought it was sufficient to description @description as a common attribute for these items.

macchiati: BTW, let me know if you'd still approve after seeing my replies.

docs/ldml/tr35-info.md
sffc: Where is the "systems" attribute though?  I don't see any line that says, as stated in the ticket,

```
<!ATTLIST convertUnit systems NMTOKENS #IMPLIED >
<!--@MATCH:set/literal/ussystem, uksystem, metric, si, other -->
```

macchiati:
https://github.com/unicode-org/cldr/blob/ee6788cf0f30e2141f14d1cd1762ef82b8846219/docs/ldml/tr35-info.md#conversion-data

That is missing, good catch. We should have a DTD line there. I'll clone
this ticket for v40 to fix that.

=====

The change that was made was to text further down the page:

<convertUnit source='fahrenheit' baseUnit='kelvin' factor='5/9'
offset='2298.35/9' systems="ussystem uksystem"/>

The factor and offset can be simple expressions, just like the values in
the unitConstants.

Where a factor is not present, the value is 1; where an offset is not
present, the value is 0.

The systems attribute indicates the measurement system(s). Multiple values
may be given; for example, *minute* is marked as systems="metric ussystem
uksystem"

Attribute ValueDescription
*si* the *International System of Units (SI)*
*metric* a superset of the *si* units, with some non-SI units accepted for
use with the SI or simple multiples of metric units, such as pound-metric
(= ½ kilogram)
*ussystem* the inch-pound system as used in the US, also called *US
Customary Units*
*uksystem* the inch-pound system as used in the UK, also called *British
Imperial Units*, differing mostly in units of volume

Mark

PR #1127, opened 2021-03-21
CLDR-14410 Fix wording in Unit_Identifier_Normalization

macchiati:
Fix the wording for a couple of steps in Unit_Identifier_Normalization

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14410

### DIFF
--- a/docs/ldml/tr35-info.md
+++ b/docs/ldml/tr35-info.md
@@ -977,7 +977,7 @@ There are many possible ways to construct complex units. For comparison of unit 
 4. For each single unit, disregarding prefixes and powers, get the order of the _simple_ unit among the `unitQuantity` elements in the [units.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/units.xml). Sort the single units by that order, using a stable sort. If there are private-use single units, sort them after all the non-private use single units.
    * meter-square-kilogram => square-kilogram-meter
    * meter-square-gram ⇒ square-gram-meter
-5. As an edge case, there could be two adjacent single units with the same _simple_ unit but different SI prefixes, such as _meter-kilometer_. In that case, sort the higher-power SI prefixes first, such as _kilometer-meter_.
+5. As an edge case, there could be two adjacent single units with the same _simple_ unit but different prefixes, such as _meter-kilometer_. In that case, sort the larger prefixes first, such as _kilometer-meter_ or _kibibyte-kilobyte_
 6. Within private-use single units, sort by the simple unit alphabetically.
 
 The examples in #4 are due to the following ordering of the `unitQuantity` elements:
--- a/docs/ldml/tr35-info.md
+++ b/docs/ldml/tr35-info.md
@@ -974,10 +974,10 @@ There are many possible ways to construct complex units. For comparison of unit 
 3. Convert multiple instances of a unit into the appropriate power.
    * foot-per-second-second ⇒ foot-per-square-second
    * kilogram-meter-kilogram ⇒ meter-square-kilogram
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-info.md
```diff
@@ -974,10 +974,10 @@ There are many possible ways to construct complex units. For comparison of unit
...
-4. For each single unit, disregarding prefixes and powers, find its base unit using `<convertUnit>`, then get the order of that base unit among the `unitQuantity` elements in the [units.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/units.xml). Then sort the single units by that order.
+4. For each single unit, disregarding prefixes and powers, get the order of the _simple_ unit among the `unitQuantity` elements in the [units.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/units.xml). Sort the single units by that order, using a stable sort. If there are private-use single units, sort them after all the non-private use single units.
    * meter-square-kilogram => square-kilogram-meter
    * meter-square-gram ⇒ square-gram-meter
-5. If two single units have the same simple unit but different SI prefixes, such as "kilometer-meter", sort the higher-power SI prefixes first.
+5. As an edge case, there could be two adjacent single units with the same _simple_ unit but different SI prefixes, such as _meter-kilometer_. In that case, sort the higher-power SI prefixes first, such as _kilometer-meter_.
```
hugovdm, line 980: Do we want to specify binary prefixes too?

Currently ICU is sorting binary prefixes before SI prefixes - but then we're also sorting from small to large, so I need to change the order anyway. If I just flip the sign, it will be SI prefixes before binary prefixes.

Current test case demonstrating ICU sort order:

input: "megafoot-mebifoot-kibifoot-kilofoot"
output: "kibifoot-mebifoot-kilofoot-megafoot"

hugovdm (commented):
That's a nice improvement!

docs/ldml/tr35-info.md
```diff
@@ -977,7 +977,7 @@ There are many possible ways to construct complex units. For comparison of unit
 4. For each single unit, disregarding prefixes and powers, get the order of the _simple_ unit among the `unitQuantity` elements in the [units.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/units.xml). Sort the single units by that order, using a stable sort. If there are private-use single units, sort them after all the non-private use single units.
    * meter-square-kilogram => square-kilogram-meter
    * meter-square-gram ⇒ square-gram-meter
-5. As an edge case, there could be two adjacent single units with the same _simple_ unit but different SI prefixes, such as _meter-kilometer_. In that case, sort the higher-power SI prefixes first, such as _kilometer-meter_.
+5. As an edge case, there could be two adjacent single units with the same _simple_ unit but different prefixes, such as _meter-kilometer_. In that case, sort the larger prefixes first, such as _kilometer-meter_ or _kibibyte-kilobyte_
```
hugovdm, line 980: And this means _megabyte-kibibyte-kilobyte_, not "binary before SI".

I'll adjust ICU logic - it will be a slightly more complicated calculation (compare binary_power*3 to SI power, and if equal, compare base. Looks like this calculation only starts failing at 1024^98. ;-) )

hugovdm (approved):
Just the examples on their own don't make it obvious that it's not binary-before-SI, but the spec seems clear enough and people should read. ;-) LGTM.

I filed a ticket to update ICU:
https://unicode-org.atlassian.net/browse/ICU-21551

docs/ldml/tr35-info.md
hugovdm: I filed https://github.com/icu-units/icu/issues/154 for the ICU update.

PR #1128, opened 2021-03-21
CLDR-14377 Clarify use of the implicit "true"

macchiati:
Make it clear that canonicalization does not 'fix' invalid key values.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14377


FrankYFTang (approved):
LGTM

PR #1129, opened 2021-03-21
CLDR-14330 only uppercase characters inside unicode_language_id

macchiati:
There should not be any uppercase characters in the canonical form of a unicode locale id except inside unicode_language_id

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14330


FrankYFTang (approved):
LGTM

PR #1137, opened 2021-03-22
CLDR-13221 format for currency symbol at decimal point

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13221

### DIFF
--- a/docs/ldml/tr35-numbers.md
+++ b/docs/ldml/tr35-numbers.md
@@ -578,6 +578,14 @@ If there is an explicit negative subpattern, it serves only to specify the negat
 
 > **Note:** The thousands separator and decimal separator in patterns are always ASCII ',' and '.'. They are substituted by the code with the correct local values according to other fields in CLDR. The same is true of the - (ASCII minus sign) and other special characters listed above.
 
+A currency decimal pattern normally contains a currency symbol placeholder (¤, ¤¤, ¤¤¤, or ¤¤¤¤¤). The currency symbol placeholder may occur before the first digit, after the last digit symbol, or where the decimal symbol would otherwise be placed (for formats such as "12€50", as in "12€50 pour une omelette").
+
+Placement | Examples
+-------|-------
+Before|"¤#,##0.00" "¤ #,##0.00" "¤-#,##0.00" "¤ -#,##0.00" "-¤#,##0.00" "-¤ #,##0.00" …
+After|"#,##0.00¤" "#,##0.00 ¤" "#,##0.00-¤" "#,##0.00- ¤" "#,##0.00¤-" "#,##0.00 ¤-" …
+Decimal|"#,##0¤00"
+
 Below is a sample of patterns, special characters, and results:
 
 ##### <a name="Sample_Patterns_and_Results" href="#Sample_Patterns_and_Results">Sample Patterns and Results</a>


docs/ldml/tr35-numbers.md
```diff
@@ -578,6 +578,20 @@ If there is an explicit negative subpattern, it serves only to specify the negat
 
 > **Note:** The thousands separator and decimal separator in patterns are always ASCII ',' and '.'. They are substituted by the code with the correct local values according to other fields in CLDR. The same is true of the - (ASCII minus sign) and other special characters listed above.
 
+A currency decimal pattern normally contains a currency symbol placeholder (one to four ¤ characters). The currency symbol placeholder may occur before the first digit, after the last digit symbol, or where the decimal symbol would otherwise be placed.
+
+Before | After | Decimal
+-------|-------|-------
+¤#,##0.00|#,##0.00¤|#,##0¤00
```
sffc, line 585: This is the only row with a third column.

docs/ldml/tr35-numbers.md
```diff
@@ -578,6 +578,20 @@ If there is an explicit negative subpattern, it serves only to specify the negat
...
+¤ #,##0.00|#,##0.00 ¤|
+¤-#,##0.00|#,##0.00-¤|
+¤ -#,##0.00|#,##0.00- ¤|
+-¤#,##0.00|#,##0.00¤-|
+-¤ #,##0.00|#,##0.00 ¤-|
+…|…|
+
+> If a decimal pattern is used as a currency pattern, and that pattern does not contain a currency symbol placeholder, then it is interpreted as if the decimal symbol were replaced by a single currency symbol. That is, "#,##0.00" ⟹ "#,##0¤00" and "#,##0" ⟹ "#,##0¤". This facility is used for formats such as "12€50", as in "12€50 pour une omelette".
```
sffc, line 593: To clarify: this is only a spec change, and corresponding data changes would come in CLDR 40. Is that correct?

Or is CLDR 39 already shipping with this data, in which case ICU code needs to be updated?

docs/ldml/tr35-numbers.md
sffc, line 585: Can the pattern `#,##0¤00` ever actually occur in CLDR data?  If it can, then we'll need to change the EBND for number patterns (assuming there is one).

macchiati: Yes, I was doing that deliberately.

docs/ldml/tr35-numbers.md
macchiati: Yes, there is no data change, so ICU doesn't need any change right now.

docs/ldml/tr35-numbers.md
```diff
@@ -578,6 +578,20 @@ If there is an explicit negative subpattern, it serves only to specify the negat
 
 > **Note:** The thousands separator and decimal separator in patterns are always ASCII ',' and '.'. They are substituted by the code with the correct local values according to other fields in CLDR. The same is true of the - (ASCII minus sign) and other special characters listed above.
 
+A currency decimal pattern normally contains a currency symbol placeholder (one to four ¤ characters). The currency symbol placeholder may occur before the first digit, after the last digit symbol, or where the decimal symbol would otherwise be placed.
```
sffc, line 581: one to five, not one to four.  According to the HTML version of the spec, 3 = display name, 4 = undefined, 5 = narrow.  I fixed the Markdown in #1139 to reflect what the HTML said.

docs/ldml/tr35-numbers.md
sffc: I thought we had an EBNF for these, but I can't find one. These number patterns continue becoming more and more complex to parse. I'm going to need to implement a parser in ICU4X. Before I do that, I would really like to have a formal grammar that I can implement.

sffc: Why?

docs/ldml/tr35-numbers.md
sffc, line 593: I don't know if the PR review is the right place to raise this concern, but I'm less and less comfortable with UTS 35 having all these weird edge cases documented in prose. If you don't read this "aside", then you aren't spec-compliant, and you might render monetary values that have no currency symbol! That seems dangerous. I would rather *require* that currency patterns contain `¤`, and permit them to occur at the decimal.

sffc (commented):
Okay, we can

docs/ldml/tr35-numbers.md
sffc: Note: This would be a noisy failure for old implementations that don't know how to handle decimal currencies, but that's by design, and better than the alternative of silently failing to display a currency symbol.

docs/ldml/tr35-numbers.md
macchiati: I wanted to show examples of the three states that you could have.

macchiati: If you think it is too confusing I can try a different approach

macchiati: /[0#]¤[0#]/ does not occur in any CLDR data (current or past), so I don't think there is any real compatibility concern.

I think an EBNF is an excellent idea. Please file a ticket for that and we can work on it (for next release).

docs/ldml/tr35-numbers.md
macchiati: Ok. I was trying to carry out the intent of your ticket, but went too far here. Will fix.

docs/ldml/tr35-numbers.md
macchiati: fixed

sffc (approved):
Thank you! This is much better.

I see now what you were trying to do in the table. I think the new table layout is clearer, even though it loses the vertical alignment of the "before" and "after" patterns.

macchiati:
Need another review because of squash to fix commit message (sigh)

sffc (approved):

PR #1141, opened 2021-03-23
CLDR-14445 Minor improvements from review of first markdown version of Dates

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14445

Just a couple of minor improvements found while reviewing the markdown conversion of the Dates section; these are not actually connected to the conversion itself (missing TOC entry for a table, missing word).

Note that there is a separate issue about table captions, I have not made any changes for those here; it is covered by https://unicode-org.atlassian.net/browse/CLDR-14611


macchiati (approved):

pedberg-icu:
Abandoned due to rebase issues, replaced with https://github.com/unicode-org/cldr/pull/1145 which does the same thing

PR #1142, opened 2021-03-23
CLDR-14253 Spec improvements: compound unit pattern algorithms

hugovdm:
See #1142.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14253


macchiati:
The actual changes look good; but can you make the following changes so we can integrate?

1. The target must not be 'master'; instead it needs to be https://github.com/unicode-org/cldr/tree/maint/maint-39
2. The title and all commits must start with CLDR-14253
3. The Checklist needs ____ replaced by 14253
4. Set me as reviewer

See for example, https://github.com/unicode-org/cldr/pull/1133

hugovdm:
Happy to drive this PR forward. I've fixed it up for the usual workflow, into maint-39, pardon the force-push noise. I can't assign reviewers though.

hugovdm:
I looked over my doc again - I have resolved comments that now make sense to me, and resolved comments that are a part of this PR. Remaining comments I would prefer to deal with separately from these smaller changes.

macchiati:
Most of this is now in another PR, and we need to discuss the rest before any action.

PR #1147, opened 2021-03-24
CLDR-14619 LDML BRS

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14619


macchiati:
Peter, I wasn't able to get the modification section straightened out, but did capture all the explanations of the changes, with the ticket numbers. So in a subsequent pass can fix. 

NOTE: I'm thinking it might be more useful to users to see a functional organization than a Section organization, so I'm thinking we might want to use the current ordering, but replace the links with pointers to sections. What do you think?

docs/ldml/tr35.md
```diff
@@ -3684,9 +3686,33 @@ Other contributors to CLDR are listed on the [CLDR Project Page](https://www.uni
 
 ## <a name="Modifications" href="#Modifications">Modifications</a>
 
-**Revision 61**
+**Revision 62**
```
pedberg-icu, line 3689: Hmm. Other PRs have been making changes to the Modification section assuming it was already for CLDR 39. We may need to move some of those changes here.

pedberg-icu (approved):

macchiati:
Agreed. I sent a message out to cldr-core about that (and I'd seen that
some of the modification section was done properly)

Mark

PR #1149, opened 2021-03-24
CLDR-14221 Add section on using delimiter data for linebreak tailoring

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14221

Add section on using delimiter data for linebreak tailoring.

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -52,6 +52,7 @@ The LDML specification is divided into the following parts:
     *   3.5 [More Information](#Character_More_Info)
     *   3.6 [Parse Lenient](#Character_Parse_Lenient)
 *   4 [Delimiter Elements](#Delimiter_Elements)
+    *   4.1 [Tailoring Linebreak Using Delimiters](#Tailor_Linebreak_With_Delimiters)
 *   5 [Measurement System Data](#Measurement_System_Data)
     *   5.1 [Measurement Elements (deprecated)](#Measurement_Elements)
 *   6 [Unit Elements](#Unit_Elements)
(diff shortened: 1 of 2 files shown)


docs/ldml/tr35-general.md
```diff
@@ -642,6 +643,22 @@ When quotations are nested, the quotation marks and alternate marks are used in
...
+in [[UAX14](https://www.unicode.org/reports/tr41/#UAX14)]. This is an example of
+[tailoring type](https://www.unicode.org/reports/tr14/#Tailoring) 1 (from that same document),
+changing the line breaking class assignment for some characters.
+
+If the values of `<quotationStart>` and `<quotationEnd>` are different, then:
+* if the value of `<quotationStart>` is a single character with linebreak class QU: Quotation, change its class to OP: Open Punctuation.
+* if the value of `<quotationEnd>` is a single character with linebreak class QU: Quotation, change its class to CL: Close Punctuation.
+Similarly for `<alternateQuotationStart>` and `<alternateQuotationEnd>`.
+
+Some characters with multiple uses should generally be excluded from this linebreak class remapping, such as:
```
macchiati, line 659: Wouldn't you instead say that if unmatched then that character would be excluded. And then provide the two possible matching characters.

pedberg-icu: Not sure what you mean. What I am trying to say is we should always leave U+2019 as QU and not remap to CL.

pedberg-icu: OK, expanded the exceptions info.

macchiati (approved):

PR #1150, opened 2021-03-24
CLDR-14612 Add a manual build step for maven

srl295:
CLDR-14612

This allows the usual 'maven' build step to be invoked on an arbitrary branch/commit. For example, we could use this to manually build the maint/maint-39 branch.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -7,14 +7,27 @@ on:
   pull_request:
     branches:
     - '*'
+  workflow_dispatch:
+    inputs:
+      git-ref:
+        description: Git Ref (Optional)
+        required: false
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - name: Clone Repository
+        uses: actions/checkout@v2
         with:
           lfs: false
+        if: github.event.inputs.git-ref == ''
+      - name: Clone Repository (manual ref)
+        uses: actions/checkout@v2
+        with:
+          lfs: false
+          ref: ${{ github.event.inputs.git-ref }}
+        if: github.event.inputs.git-ref != ''
       - name: Set up JDK 1.8
         uses: actions/setup-java@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -7,14 +7,27 @@ on:
   pull_request:
     branches:
     - '*'
+  workflow_dispatch:
+    inputs:
+      git-ref:
+        description: Git Ref (Optional)
+        required: false
```
srl295, line 14: adds a manual dispatch step (ala our commit checker)

.github/workflows/maven.yml
```diff
@@ -7,14 +7,27 @@ on:
...
+        uses: actions/checkout@v2
         with:
           lfs: false
+        if: github.event.inputs.git-ref == ''
+      - name: Clone Repository (manual ref)
+        uses: actions/checkout@v2
+        with:
+          lfs: false
+          ref: ${{ github.event.inputs.git-ref }}
+        if: github.event.inputs.git-ref != ''
```
srl295, line 30: this and below:  different checkout step, if an input is given.

.github/workflows/maven.yml
```diff
@@ -87,9 +108,10 @@ jobs:
           name: cldr-code
           path: tools/cldr-code/target/
       - name: run CLDR console check
-        run: java -DCLDR_DIR=$(pwd) -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -e -z BUILD 
+        run: java -DCLDR_DIR=$(pwd) -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -e -z BUILD
   deploy:
-    if: github.repository == 'unicode-org/cldr' && github.event_name == 'push' && github.ref == 'refs/heads/master'
+    # don't run deploy on manual builds!
+    if: github.repository == 'unicode-org/cldr' && github.event_name == 'push' && github.ref == 'refs/heads/master' && github.event.inputs.git-ref == ''
```
srl295, line 114: what it says, SKIP deployment to smoketest for these manual builds!

we may want a way to manually deploy to smoketest, but it should probably be a separate task/step.

btangmu (approved):

Manishearth (approved):
I don't understand the build system that much so I might not be the best person to ask for review, but the shape of this seems good to me!

PR #1151, opened 2021-03-24
CLDR-14541 Modernize Auto Import

btangmu:
-New AutoImport.vue for new special #auto_import

-Replace cldrLoad.dialogIsOpen with cldrStatus.get/setAutoImportBusy

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14541

### DIFF
--- a/tools/cldr-apps/src/main/webapp/js/esm/cldrMenu.js
+++ b/tools/cldr-apps/src/main/webapp/js/esm/cldrMenu.js
@@ -112,7 +112,7 @@ function loadInitialMenusFromJson(json, theLocale) {
 
   setupCoverageLevels(json, theLocale);
 
-  cldrLoad.continueInitializing(json.canAutoImport);
+  cldrLoad.continueInitializing(json.canAutoImport || false);
 }
 
 function setupCoverageLevels(json, theLocale) {
--- a/tools/cldr-apps/src/main/webapp/js/esm/cldrStatus.js
+++ b/tools/cldr-apps/src/main/webapp/js/esm/cldrStatus.js
@@ -396,7 +396,23 @@ function isDashboard() {
   return getCurrentSpecial() === "r_vetting_json";
 }
 
+/**
+ * Is auto-import of old votes in progress? (Boolean)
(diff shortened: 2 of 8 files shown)


tools/cldr-apps/src/main/webapp/js/esm/cldrMenu.js
```diff
@@ -112,7 +112,7 @@ function loadInitialMenusFromJson(json, theLocale) {
 
   setupCoverageLevels(json, theLocale);
 
-  cldrLoad.continueInitializing(json.canAutoImport);
+  cldrLoad.continueInitializing(json.canAutoImport || false);
```
srl295, line 115: leftover from debugging?

tools/cldr-apps/src/main/webapp/js/esm/cldrStatus.js
```diff
@@ -396,7 +396,23 @@ function isDashboard() {
...
+ * Used for inhibiting $('#left-sidebar').hover (Locale chooser)
+ */
+let isAutoImportBusy = false;
+
+function getAutoImportBusy() {
+  return isAutoImportBusy;
+}
+
+function setAutoImportBusy(busy) {
+  isAutoImportBusy = !!busy;
```
srl295, line 411: nice

srl295 (approved):
lgtm

tools/cldr-apps/src/main/webapp/js/esm/cldrMenu.js
btangmu: no, actually I noticed that json.canAutoImport may be undefined, in which case I'd rather pass false than undefined, since a boolean is expected -- though it wouldn't really matter in this case

tools/cldr-apps/src/main/webapp/js/esm/cldrStatus.js
btangmu: another way to do make sure the result is boolean

PR #1154, opened 2021-03-24
Cldr 13480

sven-oly:
##### Checklist


sven-oly:
Not sure where this could/should be merged.

markusicu:
- Which character is a duplicate?
- What problems does this cause? We seem to build ICU data fine?
- This data is generated from one of the Unicode Tools, so unless we fix the tool, this will re-occur.

PR #1170, opened 2021-03-31
CLDR-14615 Modernize by fixing UI glitches: Prev/Next buttons, etc.

btangmu:
-Fix progress bar (Votes/Abstain)

-Fix highlighting of Prev/Next/Toggle buttons

-Fix Priority Items Summary missing from gear menu (perm.userCanUseVettingSummary)

-Improve various alignments

-Use webpack to bundle the forum css

-Update pretty.sh for new esm, css locations

-Move setToptitleVisibility to cldrGui

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14615


tools/cldr-apps/src/main/webapp/css/redesign.css
```diff
@@ -1079,8 +1070,8 @@ html {
 
 #itemInfo div.xpath {
 	position: fixed;
-	/* TODO: get the right width of the "slider" for left sidebar; cf. beware-left-sidebar */
-	left: 34px;
+	/* The left sidebar "slider" width is 2% ( = 20% x 10%); cf. beware-left-sidebar */
+	left: 2.1%;
```
srl295, line 1074: I haven't done a ton with them, but i know sass/less precompilers lets you put math into CSS.

srl295 (approved):

tools/cldr-apps/src/main/webapp/css/redesign.css
btangmu: less would be nice, also for declaring re-usable constants!

also, flex is good for avoiding the need for one element to know the size of another element -- the left sidebar would be better with flex, and it could push the main window to the right instead of overlapping it...

PR #1179, opened 2021-04-02
CLDR-14615 Modernize: move Info Panel code to cldrInfo.js

btangmu:
-Move some Info Panel functions from cldrTable to cldrInfo

-Move appendIcon to cldrDom; called from both cldrTable and cldrInfo

-Update imports and exports as needed

-Fix recently introduced bug in cldrGui: total not const

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14615

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.js
+++ b/tools/cldr-apps/js/src/esm/cldrTable.js
@@ -393,7 +393,7 @@ function reallyUpdateRow(tr, theRow) {
    * Update the vote info.
    */
   if (theRow.voteResolver) {
-    updateRowVoteInfo(tr, theRow);
+    cldrInfo.updateRowVoteInfo(tr, theRow);
   } else {
     tr.voteDiv = null;
   }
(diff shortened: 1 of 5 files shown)


tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -1759,6 +1270,9 @@ function getValidWinningValue(theRow) {
 }
 
 export {
+  NO_WINNING_VALUE,
```
srl295, line 1273: this was surprising to me, winning value shouldn't be null. (Read some of the related code changes but couldn't find a lot of details) In any event, otherwise lgtm.

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrTable.js
btangmu: > this was surprising to me, winning value shouldn't be null...

Me too. There's something related, very old, in the 2nd item under Future Discussions in the main agenda. NO_WINNING_VALUE is also in VoteResolver.java, and there's a TODO comment there about it

PR #1192, opened 2021-04-07
CLDR-14651 fix charts for /trunk/ in link problem, grammatical forms

macchiati:
NOTE: code changes are in v40, but the data is for v39 and the charts for v39 also.

The biggest number of changes was in ToolConstants, where I put in some extra code to distinguish the two ways that we use the version numbers in folder/file names. eg cldr-39.0 vs cldr-39 

Second biggest was in ChartGrammatical forms. I'd done this earlier, but never committed the changes.

(longer term fix would be to only use one style)

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14651

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
@@ -2,6 +2,7 @@ package org.unicode.cldr.tool;
 
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.unicode.cldr.util.CLDRPaths;
 import org.unicode.cldr.util.CldrUtility;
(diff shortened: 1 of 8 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
```diff
@@ -59,45 +60,64 @@
         "36.1",
         "37.0",
         "38.0",
-        "38.1"
+        "38.1",
+        "39.0"
```
srl295, line 64: should be in a data file, as suggested in CLDR-9149 (perhaps i should dust that off and do it)

srl295 (approved):
LGTM but yes , more to clear up w/ versions

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
macchiati: Yes, good idea.

macchiati:
Ugg. Ran into compiler problem, using a future Java version. Will fix.

macchiati:
I had to make a fix; can you rereview?

srl295 (approved):
Change lgtm

PR #1202, opened 2021-04-20
CLDR-14647 Modern Dashboard Widget

btangmu:
-Display Dashboard notification types like Error as circled E

-Include comments for Dashboard items (errors/warnings/...)

-Add a button for opening Dashboard; hide button when Dash open

-Change gear menu to three-line menu

##### Checklist


btangmu:
I just pushed one correction: `this.data` not just `data`

tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -392,7 +389,7 @@ function getSingleRowUrl(tr, theRow) {
   p.append("xpath", theRow.xpathId);
   p.append("fhash", tr.rowHash);
   p.append("automatic", "t");
-  if (cldrStatus.isDashboard()) {
+  if (true || cldrStatus.isDashboard()) {
```
srl295, line 392: leftover?

tools/cldr-apps/js/src/esm/cldrText.js
```diff
@@ -437,14 +437,14 @@ const strings = {
...
   special_forum_participation: "Forum Participation",
-  special_gear: "⚙",
   special_general:
     "Please hover over the sidebar to choose a section to begin entering data. If you have not already done so, please read the <a target='_blank' href='http://www.unicode.org/cldr/survey_tool.html'>Instructions</a>, particularly the Guide and the Walkthrough. You can also use the Dashboard to see all the errors, warnings, and missing items in one place.",
   special_list_emails: "List Email Addresses",
   special_list_users: "List Users",
   special_locales: "Locale List",
   special_lookup: "Look up a code or xpath",
   special_mail: "Notifications (SMOKETEST ONLY)",
+  special_menu: "☰",
```
srl295, line 447: 🍔 !
Is a trigram the best way to do this? would it be better as `glyphicons-menu`?

tools/cldr-apps/js/src/views/DashboardWidget.vue
```diff
@@ -26,7 +26,7 @@
           title="Close"
           @click="closeDashboard"
         >
-          x
+          X
```
srl295, line 29: better as `glyphicon-remove-sign` ?

srl295 (approved):
lgtm with questions, could be followups

tools/cldr-apps/js/src/esm/cldrTable.js
btangmu: Actually that's the beginning of further work -- updating the dash after the user votes. Currently isDashboard is always false for the dash widget but we need to make the request as though it's true

tools/cldr-apps/js/src/esm/cldrText.js
btangmu: maybe, though glyphicons depends on bootstrap which is currently ancient, so reluctant to add more bootstrap dependencies now

tools/cldr-apps/js/src/views/DashboardWidget.vue
btangmu: maybe -- let's discuss, as it relates to bootstrap, fonts...

PR #1203, opened 2021-04-21
CLDR-14395 Improve ST Login

srl295:
- bump commons-codec version to recommended level
    - add more <a- components to vue router
    - cause cldrGui.run() to erase all <body> elements before starting up
    - add basic login form
    - update mainheader to hide 'user' icon if no user
    - refactor cookie code
    - auth/login: fetch from cookie if needed (still need to support anonymous)
    - login/out seems to work
    - improve error handling in startup
    - anonymous sessions (from API)!
    - bump woodstox version to most recent (2014) and document why it is there
    - prefer 'pw=' for the password query in URL (but still support uid=)
    - allow cldrGui.run() to consume a "s=" session param
    (which gets removed from the query)
    - make cldrGui.run() return a promise
    - cleanup how the "/v#" URL is generated from SurveyMain

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrGui.js
+++ b/tools/cldr-apps/js/src/esm/cldrGui.js
@@ -23,29 +23,99 @@ let dashboardVisible = false;
 
 /**
  * Set up the DOM and start executing Survey Tool as a single page app
+ * @returns Promise
  */
 function run() {
-  if (GUI_DEBUG) {
-    console.log("Hello my name is cldrGui.run");
-    debugParse();
-  }
-  const guiContainer = document.getElementById(runGuiId);
-  if (!guiContainer) {
+  try {
     if (GUI_DEBUG) {
-      console.log(
-        "cldrGui.run doing nothing since '" + runGuiId + "' not found"
-      );
+      console.log("Hello my name is cldrGui.run");
+      debugParse();
     }
-    return;
+    const guiContainer = document.getElementById(runGuiId);
+    if (!guiContainer) {
+      if (GUI_DEBUG) {
+        console.log(
+          "cldrGui.run doing nothing since '" + runGuiId + "' not found"
+        );
+      }
+      return Promise.resolve();
+    }
+    while (guiContainer.firstChild) {
+      guiContainer.removeChild(guiContainer.firstChild);
+    }
+    guiContainer.innerHTML = getBodyHtml();
+    insertHeader();
+    if (GUI_DEBUG) {
+      debugElements();
+    }
+    setOnClicks();
+    window.addEventListener("resize", handleResize);
+  } catch (e) {
+    return Promise.reject(e);
   }
-  guiContainer.innerHTML = getBodyHtml();
-  insertHeader();
-  if (GUI_DEBUG) {
-    debugElements();
-  }
-  setOnClicks();
-  window.addEventListener("resize", handleResize);
+  return ensureSession().then(completeStartupWithSession);
+}
 
+async function ensureSession() {
+  // Check first, because the session may already have been set
+  if (haveSession()) return; // Done!
+
+  // Is there a session in the URL?
+  const params = new URLSearchParams(window.location.search);
+  const s = params.get("s");
+  if (s) {
+    // We want to delete this parameter. It either is used and no longer needed,
+    // or it is invalid.
+    params.delete("s");
+    window.location.search = params.toString();
+
+    const inforesponse = await fetch(
+      `api/auth/info?session=${encodeURIComponent(s)}`,
+      {
+        method: "GET",
+        headers: {
+          Accept: "application/json",
+        },
+      }
+    );
+    if (inforesponse.ok) {
+      const infoinfo = await (await inforesponse).json();
+      if (infoinfo.sessionId) {
+        cldrStatus.setSessionId(inforesponse.sessionId);
+      }
+      return; // Done.
+    }
+    // Else: it wasn't accepted as a session id. That's OK.
+  }
+
+  // see if we can get a session
+  const response = await fetch(`api/auth/login`, {
+    method: "POST",
+    headers: {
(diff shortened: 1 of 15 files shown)


tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -23,29 +23,96 @@ let dashboardVisible = false;
...
   }
-  guiContainer.innerHTML = getBodyHtml();
-  insertHeader();
-  if (GUI_DEBUG) {
-    debugElements();
+  return ensureSession().then(completeStartupWithSession);
+}
+
+async function ensureSession() {
+  // Check first, because the session may already have been set
```
btangmu, line 60: this function is too long, needs subroutines -- but not urgent, could fix later

tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -23,29 +23,96 @@ let dashboardVisible = false;
...
+    }
+  } else {
+    throw Error("SurveyTool did not create a session. Try back later.");
+  }
+}
+
+function haveSession() {
+  return (
+    cldrStatus.getSurveyUser() != null || cldrStatus.getSessionId() != null
+  );
```
btangmu, line 115: this comparison with null looks fragile; what if either of these functions returns undefined, or an empty string? How about requiring truthy return value?

tools/cldr-apps/js/src/esm/cldrStatus.js
```diff
@@ -266,15 +268,7 @@ function setSessionId(i) {
...
-  if (sessionIdChangeCallback) {
-    sessionIdChangeCallback(sessionId);
-    sessionIdChangeCallback = null;
-  }
-}
-
-function setSessionIdChangeCallback(func) {
-  sessionIdChangeCallback = func;
+  statusTarget.dispatchEvent(new Event("sessionId"));
 }
```
btangmu, line 272: I like using dispatchEvent!

tools/cldr-apps/js/src/runGui.js
```diff
@@ -1,7 +1,7 @@
 import * as cldrGui from "./esm/cldrGui.js";
 
 function runGui() {
-  cldrGui.run();
+  return cldrGui.run();
 }
```
btangmu, line 5: A @return comment would make this less mysterious

tools/cldr-apps/js/src/views/LoginButton.vue
```diff
@@ -35,7 +50,34 @@ export default {
         // now reload this page now that we've logged out
         await window.location.reload();
       } else {
-        window.location.replace("./login.jsp");
+        if (this.userName && this.password) {
+          try {
```
btangmu, line 54: subroutine would be better than inserting so much code into existing function

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Auth.java
```diff
@@ -57,25 +58,65 @@ public Response login(
...
-            resp.user = CookieSession.sm.reg.get(request.password,
-                request.email, userIP);
-            if (resp.user == null) {
-                return Response.status(403, "Login failed").build();
-            }
-            CookieSession session = CookieSession.retrieveUser(resp.user);
-            if (session == null) {
+            CookieSession session = null;
+            if (!request.isEmpty()) {
+                resp.user = CookieSession.sm.reg.get(request.password,
```
btangmu, line 77: subroutines would really be good here

btangmu (approved):

tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -23,29 +23,99 @@ let dashboardVisible = false;
...
+          "cldrGui.run doing nothing since '" + runGuiId + "' not found"
+        );
+      }
+      return Promise.resolve();
+    }
+    while (guiContainer.firstChild) {
+      guiContainer.removeChild(guiContainer.firstChild);
+    }
+    guiContainer.innerHTML = getBodyHtml();
+    insertHeader();
```
btangmu, line 47: doesn't setting innerHTML effectively make removeChild superfluous?

tools/cldr-apps/js/src/esm/cldrGui.js
srl295: truthy would be fine

tools/cldr-apps/js/src/runGui.js
srl295: yes, or inline the whole thing into index.js

tools/cldr-apps/js/src/views/LoginButton.vue
srl295: yes, the login stuff should go under esm modules

tools/cldr-apps/js/src/esm/cldrGui.js
srl295: didn't seem to work…

srl295:
^ forced recheck by visiting `https://cla-assistant.io/check/unicode-org/cla?pullRequest=1203`

tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -23,29 +23,99 @@ let dashboardVisible = false;
...
+        headers: {
+          Accept: "application/json",
+        },
+      }
+    );
+    if (inforesponse.ok) {
+      const infoinfo = await (await inforesponse).json();
+      if (infoinfo.sessionId) {
+        cldrStatus.setSessionId(inforesponse.sessionId);
+      }
```
btangmu, line 85: I don't understand this... I'm just wondering about the distinction here between infoinfo.sessionId and inforesponse.sessionId -- looks as though maybe inforesponse.sessionId should be infoinfo.sessionId?

PR #1206, opened 2021-04-22
CLDR-14451 add README.md and LICENSE to cldr-json packages

srl295:
- automatically copy unicode-license.txt into the CLDR dataset
- also, improve Ldml2JsonConverter response to incorrect -t options

This concludes CLDR-14451

PR #1208, opened 2021-04-22
CLDR-14640 cldr-json: don't create empty output dirs

srl295:
CLDR-14640 cldr-json: don't create empty output dirs

- [X] this completes the issue


srl295:
does not include the pom.xml fix in #1206 (to avoid conflict)

btangmu (approved):

PR #1209, opened 2021-04-23
CLDR-14647 Modern Dashboard Widget, updating...

btangmu:
-Make key and id unique for Dashboard rows; allow more than one row per xpath

-New module cldrDash.js for managing the Dashboard data

-Add xpathIndex to index all Dashboard rows by xpath and notification category

-Fix missing parameter tr for singleRowErrHandler in cldrTable.js

##### Checklist

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDash.js
+++ b/tools/cldr-apps/js/src/esm/cldrDash.js
@@ -1,0 +1,220 @@
+/*
+ * cldrDash: encapsulate dashboard data.
+ */
+import * as cldrStatus from "./cldrStatus.js";
+import * as cldrSurvey from "./cldrSurvey.js";
+
+/**
+ * An object whose keys are xpath hex IDs, and whose values are objects whose keys are notification categories
+ * such as "Error" or "English_Changed" and values are "entry" objects with code, english, ..., xpath elements.
+ *
+ * There can be at most one item for an xpath within each category, but the same xpath may have items in
+ * multiple categories. For example, the xpath "db7b4f2df0427e4" might have both "Error" and "Warning" items,
+ * but there can't be more than one "Error" item for the same xpath, since the server will have combined them
+ * into a single "Error" item with multiple error messages.
+ */
+let xpathIndex = {};
+
+/**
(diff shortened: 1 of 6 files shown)


btangmu:
This PR is to update counts/rows in Dashboard when the user votes. I think it accomplishes everything that the old Dashboard was able to do in that respect, plus a little more. If the user votes even on a row that wasn't in the dashboard, and the vote results in new errors/warnings/etc, the dashboard is updated accordingly; however the row may not appear in the right order in the dashboard since that ordering is currently only determined by the back end when it generates the complete dashboard json.

Further improvement of the updating might be facilitated by back-end changes. Note that we still depend on the deprecated SurveyAjax.getRow, which accepts a "dashboard=true" query parameter. The new API version doesn't accept that parameter and doesn't deliver the "json.issues" on which the dashboard-update code relies. Possibly "json.issues" could be reconstructed from the plain getrow data. The code would be cleaner, though, if "json.issues" had all the data needed for updating the dashboard for a single xpath -- just like a miniature version of the full dashboard json, in the same format but limited to a single path. The change here to VettingViewer.java doesn't accomplish anything yet other than to enabling to examine the full FileInfo data in the debugger; it points to where we generate something like a single-xpath set of dashboard data but then throw away everything except the labels like "Error".

tools/cldr-apps/js/src/esm/cldrDash.js
```diff
@@ -0,0 +1,220 @@
...
+    hex: xpstrid,
+  });
+  if (result && result.ph) {
+    // success
+    return result.ph;
+  }
+  // failure; fall back on dummy header, and current section/page
+  return {
+    header: "[header]",
+    page: cldrStatus.getCurrentPage(),
```
srl295, line 215: Log err?

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrDash.js
btangmu: Good idea. I'll add that. The fallback might be superfluous... This all needs more testing...

PR #1212, opened 2021-04-26
CLDR-14647 Modern Dashboard: hiding with preferences

btangmu:
-Connect checkboxes for hiding items with preferences in back end database

-Restore json.hidden in Dashboard response

-Change the table to store hex xpath id string instead of decimal xpath id integer

-Change the table name from cldr_review_hide_40 to cldr_dash_hide_40

-Remove dead code

-Comments

##### Checklist

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDash.js
+++ b/tools/cldr-apps/js/src/esm/cldrDash.js
@@ -1,6 +1,7 @@
 /*
  * cldrDash: encapsulate dashboard data.
  */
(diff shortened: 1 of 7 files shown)


srl295 (approved):
This is good and gives us progress.
- I was trying to make the dashboard output not specific to one user, so it could potentially be a cached query, in other words tried to avoid mixing 'public' and 'personal' data in one query.
- I think it's fine to use hex, it's different than other tables.

PR #1220, opened 2021-05-03
CLDR-14675 improve logging

srl295:
CLDR-14675

- [ ] This PR completes the ticket. (but close… there are more call sites that could be done)

- use java.util.logging where possible
- deprecate many SurveyLog functions
- add in commented out code for adding webContext extensions to logging
- bump default logging settings to 10*50MB = 500MB of logs (up from 60MB total)

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
@@ -14,6 +14,7 @@ import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
+import java.util.logging.Logger;
 
 import javax.servlet.http.HttpServletRequest;
 
@@ -34,7 +35,7 @@ import org.unicode.cldr.web.WebContext;
  * rather than environment variables.
  */
 public class CLDRConfigImpl extends CLDRConfig implements JSONString {
-
+    private static final Logger logger = Logger.getLogger(CLDRConfigImpl.class.getName());
     /**
      * Get an instance and downcast
      * @return
@@ -47,7 +48,7 @@ public class CLDRConfigImpl extends CLDRConfig implements JSONString {
         try {
             return (CLDRConfigImpl) config;
         } catch (ClassCastException cce) {
-            System.err.println("Error: CLDRConfig.getInstance() returned "
+            logger.severe("Error: CLDRConfig.getInstance() returned "
                 + config.getClass().getName()
                 + " initialized at "
                 + config.getInitStack());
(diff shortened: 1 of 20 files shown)


tools/cldr-apps/pom.xml
```diff
@@ -52,6 +52,13 @@
 			<version>4.4.1</version>
 		</dependency>
 
+		<!--  used for logging -->
+		<!-- <dependency>
+			<groupId>com.ibm.websphere.appserver.api</groupId>
+			<artifactId>com.ibm.websphere.appserver.api.hpel</artifactId>
+			<version>2.0.51</version>
+			<scope>provided</scope>
+		</dependency> -->
```
btangmu, line 61: Maybe better (for "clean code") just to delete rather than comment out?

btangmu: ... or is the idea that this is to be added at some point?

tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
```diff
@@ -159,23 +160,23 @@ private synchronized void init() {
                 if (homeParent == null) {
                     homeParent = System.getProperty(prop);
                     if (homeParent != null) {
-                        System.err.println(" cldrHome found, using " + prop + " = " + homeParent);
+                        logger.info(" cldrHome found, using " + prop + " = " + homeParent);
                     } else {
-                        System.err.println(" Unset: " + prop);
+                        logger.warning(" Unset: " + prop);
                     }
```
btangmu, line 166: I like the distinction between errors, warnings, and info!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/AbstractCacheManager.java
```diff
@@ -40,9 +42,9 @@ public void setup() {
         Instant d = c.load();
         // Expire the URL mapping every 90 days.
         final Instant latestGoodInstant = cacheDir.getType().getLatestGoodInstant(DEFAULT_DAYS);
-        // TODO: use SurveyLog.logger.info() instead of System.err.println
+        // TODO: use logger.info() instead of System.err.println
```
btangmu, line 45: No longer a TODO if you've already done it?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyLog.java
```diff
@@ -4,21 +4,15 @@
...
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 
 import org.unicode.cldr.util.CLDRConfig;
-import org.unicode.cldr.util.StackTracker;
+
+// TODO: needs POM entry
+//import com.ibm.websphere.logging.hpel.LogRecordContext;
```
btangmu, line 15: is this "TODO" unfinished work for this ticket?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyLog.java
```diff
@@ -28,35 +22,28 @@
...
     public static void logException(Throwable t) {
         logException(t, "", null);
     }
 
+    @Deprecated
     public static void logException(Throwable t, String what) {
         logException(t, what, null);
     }
 
+    @Deprecated
```
btangmu, line 46: Sometimes deprecation warnings have the drawback of making more important warnings harder to spot. For example, currently Eclipse reports 3211 warnings. Who knows how many of them are actual bugs we're not fixing, and how many are just wishful thinking? It might be good policy if we only add deprecation in the context of an actual plan -- with a ticket and a due date -- for removing the deprecated items. Just my 2 cents worth

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyLog.java
```diff
@@ -82,66 +69,23 @@ public static void setDir(File homeFile) {
...
-            ioe.printStackTrace();
-        }
-
-        getChunkyReader().nudge();
-
-        // then, to screen
-        logger.severe(sb.toString());
+    public static void logException(Logger logger, final Throwable exception, String what, WebContext ctx) {
+        // TODO: needs POM entry
+        // LogRecordContext.addExtension("webContext",ctx.toString());
```
btangmu, line 74: Is this TODO tied to completion of the ticket?

btangmu (approved):
Some "TODO" comments should be addressed before closing the ticket?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyLog.java
srl295: it's a refactoring aid. Ticket completion would be removing all usages and deleting this method.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyLog.java
srl295: yes

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyLog.java
srl295: In my way of thinking I would deprecate SurveyAjax  … the same way JSP is deprecated for you (and for me).  But I can see the value of keeping the error volume down. 

as to getrow, I'd change a callsite to use the new getrow rather than update the old one.

srl295:
@btangmu merge conflicts… took the opportunity to address the TODOs

btangmu (approved):

PR #1221, opened 2021-05-03
CLDR-14668 fix additional coverage/General Info issues

srl295:
- fix orgCoverage missing from data() in MainHeaderVue.js
 - update cldrLoad.js / SurveyAjax.java to not pass the org coverage back from
 client to server, but send "auto" and let the server figure it out. Otherwise we'd
 need to delay load.

CLDR-14668

Note:
- Split off CLDR-14703 to modernize cldrLoad and the General Info page

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrCoverage.js
+++ b/tools/cldr-apps/js/src/esm/cldrCoverage.js
@@ -48,14 +48,14 @@ function covName(lev) {
 }
 
 function effectiveCoverage(locale) {
-  const orgCov = getSurveyOrgCov(locale);
-  if (!orgCov) {
-    throw new Error(`surveyOrgCov(${locale}) not yet initialized`);
-  }
   const userCov = getSurveyUserCov();
   if (userCov) {
     return covValue(userCov);
   } else {
+    const orgCov = getSurveyOrgCov(locale);
+    if (!orgCov) {
+      throw new Error(`surveyOrgCov(${locale}) not yet initialized`);
+    }
     return covValue(orgCov);
   }
 }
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/js/src/esm/cldrCoverage.js
```diff
@@ -48,14 +48,14 @@ function covName(lev) {
...
-    throw new Error(`surveyOrgCov(${locale}) not yet initialized`);
-  }
   const userCov = getSurveyUserCov();
   if (userCov) {
     return covValue(userCov);
   } else {
+    const orgCov = getSurveyOrgCov(locale);
+    if (!orgCov) {
+      throw new Error(`surveyOrgCov(${locale}) not yet initialized`);
+    }
```
btangmu, line 58: Supposing this `throw` happens in spite of not being supposed to happen, do we `catch` it anywhere? If not, I wonder if Survey Tool would recover more gracefully with or without the throw? I'd be tempted to print an error to the console then return a fallback value -- just a thought

tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -715,20 +715,15 @@ function loadGeneral(itemLoadInfo) {
...
     currentLocale +
     "&s=" +
     cldrStatus.getSessionId() +
-    "&eff=" +
-    effectiveCov +
-    "&req=" +
-    requiredCov +
+    "&userCov=" +
+    (userCov||'auto') +
     cldrSurvey.cacheKill();
```
btangmu, line 727: Needs prettier

btangmu (approved):
Needs formatting, otherwise LGTM though it's not entirely clear to me

tools/cldr-apps/js/src/esm/cldrCoverage.js
srl295: yes, it probably needs to be caught… we don't have a catch tree on the vue side.

PR #1225, opened 2021-05-04
CLDR-14358 Add langtags.json data

DavidLRowe:
CLDR-14358

This pull request is an initial step for this ticket, providing data for further development work.
Source of data: https://raw.githubusercontent.com/silnrsi/langtags/master/pub/langtags.json
API documentation: https://github.com/silnrsi/langtags repository, doc/langtags.md (https://github.com/silnrsi/langtags/blob/master/doc/langtags.md)


srl295:
@DavidLRowe This is interesting to be sure, but I actually wonder if it would be better if it was transformed into LDML format. For example,

```json
    {
        "full": "zyp-Latn-MM",
        "iana": [ "Zyphe Chin" ],
        "iso639_3": "zyp",
        "localnames": [ "Zyphe" ],
        "name": "Chin, Zyphe",
        "names": [ "Vawngtu", "Zo-pe", "Zophei", "Zoptei", "Zyphe", "Zyphe Chin" ],
        "region": "MM",
        "regionname": "Myanmar",
        "regions": [ "IN" ],
        "script": "Latn",
        "sldr": false,
        "tag": "zyp",
        "tags": [ "zyp-Latn", "zyp-MM" ],
        "windows": "zyp-Latn"
    }
```

Could be:

```xml
<ldml>
    <identity>….   <language type=zyp <script type=Latn <territory type=MM …</identity>
    <localeDisplayNames
       <languages>
           <language type="zyp">Vawngtu</language>
           <language alt="variant-1" type="zyp">Zo-pe</language>
           <language alt="variant-2" type="zyp">Zoptei</language>
     </languages>
… 
```

DavidLRowe:
Hi @srl295 This is the format that is used by a number of SIL programs (including Keyman starting with version 14, I think) that want to offer a list of language tags. It's published on https://github.com/silnrsi/langtags/tree/master/pub . 

Our main concern is that, since we include CLDR as one source of information, we need to be able to identify what data in the CLDR information came from SIL, so that we can avoid a circular dependency.

srl295:
Maybe we at least want to have some way to convert between the formats.

macchiati:
The purpose of this is purely internal just as a check on the SIL data so
we can see what changes from time to time. So I think Jason format is fine
the next time we get a drop is when we'll have to have a tool that verifies
the Delta.

DavidLRowe:
There are ad hoc corrections throughout the year, but the time of biggest change is roughly March to April. (ISO 639-3 updates in Feb, Ethnologue and IANA update after that.) So, a yearly update as part of the even numbered release might be reasonable.

srl295 (approved):
I'm ok then

PR #1229, opened 2021-05-05
CLDR-14395 ST: don't replace location.search

srl295:
- use pushState to avoid an unintended reload

CLDR-_____

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrGui.js
+++ b/tools/cldr-apps/js/src/esm/cldrGui.js
@@ -67,7 +67,13 @@ async function ensureSession() {
     // We want to delete this parameter. It either is used and no longer needed,
     // or it is invalid.
     params.delete("s");
-    window.location.search = params.toString();
+    let newQuery = '';
+    if (params.toString()) {
+      newQuery = '?' + params.toString();
+    }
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -67,7 +67,13 @@ async function ensureSession() {
     // We want to delete this parameter. It either is used and no longer needed,
     // or it is invalid.
     params.delete("s");
-    window.location.search = params.toString();
+    let newQuery = '';
+    if (params.toString()) {
```
btangmu, line 71: setting window.location.search is problematic for sure -- it causes ensureSession to execute twice, throwing away the good session from the "info" request and replacing it with a bad session (no user) from the "login" request

srl295:
@btangmu is this now obsolete?

btangmu:
> @btangmu is this now obsolete?

Yes. The same changes were in https://github.com/unicode-org/cldr/pull/1231

This is slightly better than setting window.location.search, but cldrGui.run still eventually ends up getting called twice and calls the user to get logged out, it just happens later, as when clicking on a link in a report

PR #1231, opened 2021-05-06
CLDR-14705 Fix log-in bugs

btangmu:
-In setSessionId, call dispatchEvent if and only if id changes

-Call cldrStatus.on earlier so it gets called the first time id changes

-Remove unused LoginResponse.user and LoginResponse.newlyLoggedIn, avoid User serialization errors

-Add console logging for debugging

-Comments

CLDR-14705


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Auth.java
```diff
@@ -200,13 +199,12 @@ public Response info(
         // Response
         LoginResponse resp = new LoginResponse();
         resp.sessionId = session;
-        resp.user = s.user;
+
```
srl295, line 202: omit the user? may not be used at present

srl295 (approved):
makes sense

PR #1236, opened 2021-05-07
CLDR-14555 reduce JDBC warnings

srl295:
- replace getDBConnection() with getAConnection() for locations
where commit() is not called (such as read-only usage)
- property org.unicode.cldr.web.DBUtils.DEBUG_DB_OPEN for tracing
where connections are being opened
- update StackTracker to be able to skip with a Predicate<> (debug)

CLDR-14555

- [X] This PR completes the ticket. hopefully!

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
@@ -55,8 +55,12 @@ import com.ibm.icu.dev.util.ElapsedTimer;
  */
 public class DBUtils {
     private static final String DEFAULT_DATA_SOURCE = "java:comp/DefaultDataSource";
-    private static final boolean DEBUG = false;// CldrUtility.getProperty("TEST",
-    // false);
+    private static final boolean DEBUG = false;
+    /**
+     * Set org.unicode.cldr.web.DBUtils.DEBUG_DB_OPEN=true to trace when DBConnections are opened
+     */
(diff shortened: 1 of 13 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
```diff
@@ -815,6 +819,13 @@ public final Connection getDBConnection() {
      * @return
      */
     public final Connection getAConnection() {
+        if (DEBUG_DB_OPEN) {
+            final String myClass = DBUtils.class.getName();
+            final StackTraceElement callerElement = StackTracker
+                .firstCallerMatching(
+                    (StackTraceElement s) -> !s.getClassName().equals(myClass));
+            System.out.println("DB OPEN: " + callerElement);
+        }
```
srl295, line 828: @btangmu might be able to rewrite this with #1220 to use a log level where this code is only executed if the right level of detail is enabled

srl295:
> I'm still seeing this locally (Eclipse console):
> 
> ```
> [WARNING ] WLTC0033W: Resource dataSource[DefaultDataSource]/connectionManager rolled back in cleanup of LocalTransactionContainment.
> [WARNING ] WLTC0032W: One or more local transaction resources were rolled back during the cleanup of a LocalTransactionContainment.
> ```
> 
> It happens when I log in as a TC and submit a new value. Would it help if I screenshare?

thanks, I thought I had tested that case, but I see it as well now.

PR #1239, opened 2021-05-11
CLDR-14504 Add kWh-per-100km

macchiati:
CLDR-14504


sffc (approved):

macchiati:
The most general would be to say that a unit identifier can also have a
prefix

prefixed_unit (prefix)? simple_unit

=>

prefixed_unit (number '-')? (prefix)? simple_unit
(number being non-negative integer, maybe restricted to powers of 10 also.)

It seems fairly common for people to mix numbers into units, eg $/MBFT would
be usd-per-1000-board-feet

Mark

PR #1204, opened 2021-04-21
CLDR-14555 fix SQL warnings

srl295:
- restructure MailSender code, cleanup, delete dead code
- rollback() if no mail action. Otherwise resources are left to be cleaned up
- turn off DEBUG mode in DBUtils

PR #1245, opened 2021-05-12
CLDR-14555 addXpaths needs to not use autocommit conn

srl295:
CLDR-14555 addXpaths needs to not use autocommit conn


srl295:
cross ref: #1204 #1236 #1243

macchiati (approved):

PR #1248, opened 2021-05-13
CLDR-14755 enable locales for grammar collection

macchiati:
CLDR-14755 enable locales for grammar collection

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
@@ -101,6 +101,8 @@ public class GrammarInfo implements Freezable<GrammarInfo>{
     private Map<GrammaticalTarget, Map<GrammaticalFeature, Map<GrammaticalScope,Set<String>>>> targetToFeatureToUsageToValues = new TreeMap<>();
     private boolean frozen = false;
 
+    /** Only internal */
+    @Deprecated
     public void add(GrammaticalTarget target, GrammaticalFeature feature, GrammaticalScope usage, String value) {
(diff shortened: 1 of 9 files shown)


macchiati:
This is a first pass to enable the grammatical data for locales we added in v39. I expect to narrow down the number of genders/cases as a result of the survey before Shakedown.

srl295:
@macchiati i had to make a sub-PR: https://github.com/macchiati/cldr/pull/7

srl295:
ansible lint is just a timeout, looks like it passed (in "only" 35min!)

pedberg-icu (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
```diff
@@ -470,4 +476,88 @@ private void getSourceCaseAndPluralPolish(String gender, String value,
...
+    }
+
+    /**
+     * Return the locales that have either case or gender info for units (or both).
+     */
+    public static Set<String> getGrammarLocales() {
+        return GrammarLocales.data;
+    }
+
+    static final Set<String> INCLUDE_OTHER = ImmutableSet.of(
```
srl295, line 494: should be data driven

srl295 (approved):
LG so far

srl295:
mem fixed in https://github.com/unicode-org/cldr/pull/1259

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
macchiati: I thought of that, but wanted to work out exactly the best way to do it.

srl295 (approved):
rslgtm

PR #1250, opened 2021-05-13
CLDR-9895 add bo collation rules

eroux:
CLDR-9895


eroux:
The development of these rules happened in https://github.com/eroux/tibetan-collation/ see https://github.com/eroux/tibetan-collation/blob/master/implementations/Unicode/CHANGELOG.md for the difference with the rules from dz.xml

pedberg-icu (approved):

pedberg-icu:
I will file a separate PR to change the draft status for this and dz collation from unconfirmed to contributed

eroux:
@pedberg-icu excellent, thanks! 

We haven't really discussed about the dz rules (I didn't want it to hold back the `bo` ones), and they can stay that way for now, but the rules from these PR would improve the `dz` collation... we can keep that for a future PR though, but one easy thing to do would be to have `dz` import the `bo` rules

pedberg-icu:
OK, I will just update the draft status for bo

eroux:
ok yes, that will be more simple, I'll look at the Dzongkha rules for the next CLDR release. Thanks again!

PR #1256, opened 2021-05-14
CLDR-14736 add bal and bal_Latn to seed

DavidLRowe:
CLDR-14736

It's possible that this PR completes the ticket, but there may be one other item that should change.
- Currently common/supplemental/likelySubtags.xml contains an entry for 'bal', but not one for 'bal_Latn'. But the data in likelySubtags.xml is generated from another source, so potentially another change is needed.


DavidLRowe:
I'll correct the data and try again.

PR #1263, opened 2021-05-19
CLDR-14762 update cldrTable for grammatical units

srl295:
CLDR-14762 update cldrTable for grammatical units

- some XPaths were missed from this function when added
to CLDRFile extra paths

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.js
+++ b/tools/cldr-apps/js/src/esm/cldrTable.js
@@ -19,6 +19,7 @@ import * as cldrStatus from "./cldrStatus.js";
 import * as cldrSurvey from "./cldrSurvey.js";
 import * as cldrText from "./cldrText.js";
 import * as cldrVote from "./cldrVote.js";
+import * as cldrXPathUtils from "./cldrXpathUtils.js";
 
 const CLDR_TABLE_DEBUG = false;
 /*
@@ -617,7 +618,7 @@ function checkRowConsistency(theRow) {
         /*
          * In earlier implementation, essentially the same error was reported as "... there is no Bailey Target item!").
          */
-        if (!extraPathAllowsNullValue(theRow.xpath)) {
+        if (!cldrXPathUtils.extraPathAllowsNullValue(theRow.xpath)) {
           console.error(
             "For " +
               theRow.xpstrid +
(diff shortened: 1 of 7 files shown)


tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -663,7 +663,8 @@ function extraPathAllowsNullValue(path) {
   if (
     path.includes("timeZoneNames/metazone") ||
     path.includes("timeZoneNames/zone") ||
-    path.includes("dayPeriods/dayPeriodContext")
+    path.includes("dayPeriods/dayPeriodContext") ||
+    /\/\/ldml\/units\/unitLength.*\/gender.*/.test(path)
```
srl295, line 667: @macchiati what is the right xpath to check for here?

srl295: `//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"light-lumen\"]/gender` is my sample here

macchiati (changes requested):
See email for others. 

Also, since these are not exact, I don't think we need the expense of a regex check.

srl295:
> See email for others.
> 
> Also, since these are not exact, I don't think we need the expense of a regex check.

the check is only performed on items with "inheritance marker  but no inherited value" so a regex (precompiled and cached by JS) is probably fine.

List for posterity:

```
//ldml/dates/calendars/calendar[@type=\"gregorian\"]/dayPeriods/dayPeriodContext[@
//ldml/dates/timeZoneNames/metazone[@
//ldml/dates/timeZoneNames/zone[@
//ldml/numbers/currencies/currency[@
//ldml/units/unitLength[@type="long"]/unit[@
//ldml/units/unitLength[@type="long"]/compoundUnit[@
//ldml/numbers/minimalPairs/caseMinimalPairs/
```

macchiati:
Great. Please me know when this is in, so I can try it out.

Mark

tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -660,12 +660,18 @@ function checkRowConsistency(theRow) {
...
-    return true;
+  for (prefix of [
+    '//ldml/dates/calendars/calendar[@type="gregorian"]/dayPeriods/dayPeriodContext[@',
+    "//ldml/dates/timeZoneNames/metazone[@",
+    "//ldml/dates/timeZoneNames/zone[@",
+    "//ldml/numbers/currencies/currency[@",
+    '//ldml/units/unitLength[@type="long"]/unit[@',
+    '//ldml/units/unitLength[@type="long"]/compoundUnit[@',
+    "//ldml/numbers/minimalPairs/caseMinimalPairs/",
+  ]) {
```
srl295, line 671: quotes inconsistent, but for the purpose of avoiding `\"`

macchiati (dismissed):

srl295:
previous fn didn't work in browser, added a test case.

This seems to 'kick the can down the road' as there are still warnings. Investigating.

srl295:
@macchiati the assertion goes away with this PR. but the scrolling isn't fixed:

```
Warning could not load id 64ff8e7e13e4b11d - not setup - [object HTMLTableRowElement] pc=[object HTMLTableCellElement] sf = undefined
```

strange because `r@64ff8e7e13e4b11d`  is actually a valid ID at that point.

srl295:
fixed in this pr
image

tools/cldr-apps/js/test/nonbrowser/test-cldrXpathUtils.js
```diff
@@ -0,0 +1,22 @@
+const expect = require("chai").expect;
+const mocha = require("mocha");
+
+import * as cldrXpathUtils from "../../src/esm/cldrXpathUtils";
+
+describe("cldrXpathUtils test", function () {
```
srl295, line 6: @btangmu this lets us have js tests outside of a browser (in the nonbrowser dir). I choose to use chai/mocha since you were already using it. Later we can update the browser tests to use the same chai/mocha package

macchiati:
Great, thanks!

srl295:
@macchiati can you review and merge? need to stop for the day

macchiati (approved):

macchiati:
done

Mark

PR #1230, opened 2021-05-06
CLDR-14610 add Tok Pisin (tpi) to seed

DavidLRowe:
CLDR-14610

- [ X ] This PR completes the ticket.

PR #1264, opened 2021-05-19
CLDR-14769 fix pom.xml for language group tool (RDF)

srl295:
- an incorrect mainClass was there, preventing use of the RDF classes
- in fact, it is better to have no mainClass so that it is given from the command line

CLDR-14769

Discovered this with #1230 

- [x] This PR completes the ticket. (need to update docs also)

### DIFF
--- a/tools/cldr-rdf/pom.xml
+++ b/tools/cldr-rdf/pom.xml
@@ -12,7 +12,6 @@
     <url>https://unicode.org/cldr</url>
 
     <properties>
-        <mainClass>org.unicode.cldr.rdf.tool.FetchAbstracts</mainClass>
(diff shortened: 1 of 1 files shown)


tools/cldr-rdf/pom.xml
```diff
@@ -12,7 +12,6 @@
     <url>https://unicode.org/cldr</url>
 
     <properties>
-        <mainClass>org.unicode.cldr.rdf.tool.FetchAbstracts</mainClass>
```
srl295, line 15: no longer exists

btangmu (approved):

srl295:
Updated docs

PR #1265, opened 2021-05-20
CLDR-14772 suppress French many

macchiati:
CLDR-14772

This is the first fix, for shakedown. There will be additional PRs to generalize the check for superfluous plural values. 

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -3363,7 +3363,7 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         }
         Set<Count> pluralCounts = Collections.emptySet();
         if (plurals != null) {
-            pluralCounts = getAdjustedPluralCounts(plurals);
+            pluralCounts = plurals.getAdjustedCounts();
             if (pluralCounts.size() != 1) {
                 // we get all the root paths with count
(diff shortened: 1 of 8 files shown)


macchiati:
Just completed this; ready for review.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -3363,7 +3363,7 @@ public String getWinningPath(String path) {
         }
         Set<Count> pluralCounts = Collections.emptySet();
         if (plurals != null) {
-            pluralCounts = plurals.getCounts();
+            pluralCounts = plurals.getAdjustedCounts();
```
srl295, line 3366: plurals.getCountsPresentInData()? adjusted is a little nebulous

srl295 (approved):
LGTM with some minor questions…

One question though is, is this the right approach given the stated purpose? 
If the only purpose is to suppress the values in the Survey Tool, should that be done outside of CLDRFile (which affects all tooling)? 

We have the following mechanisms in the API for affecting ST visibility:
- `PathHeader.getSurveyToolStatus(xpath)` - takes an xpath, no locale information (Oriented around low-level path suppression)
- `CoverageLevel2` (gets an xpath and a locale, oriented around coverage level)
- `CheckCLDR.Phase.getShowRowAction()` (gets xpath, locale, user level, error status, oriented around access control)

Should these be unified at all?

macchiati:
You bring up some good points. I want to get this in now, but will continue with the ticket.

PR #1270, opened 2021-05-24
CLDR-14675 reduce MailSender debug output

srl295:
CLDR-14675

integer comparison was backwards. should have been "logger.getValue() < … FINE" not >=. 
Anyway, there's a function that can be used instead.


srl295:
go ahead and merge if you want (i know this bug is an annoyance).  i also moved the ticket back into reviewing since it was recently closed.

btangmu (approved):
isLoggable? FINE!

PR #1275, opened 2021-05-26
CLDR-13425 Add <datetimeSkeleton> for std date/time patterns, CLDRModify -fS to add, Check tests

pedberg-icu:
CLDR-13425

Note: There will likely be additional PRs for this ticket. This one includes the following:

1. Add &lt;datetimeSkeleton&gt; element to be used in &lt;dateFormat&gt; and &lt;timeFormat&gt; elements alongside the &lt;pattern&gt; element. It has the same attributes as &lt;pattern&gt; except that it lacks the _type_ and _count_ attributes.
2. Add a new CLDRModify filter -fS to add the &lt;datetimeSkeleton&gt; in every &lt;dateFormat&gt; and &lt;timeFormat&gt; element. If the &lt;pattern&gt; has "↑↑↑" then the &lt;datetimeSkeleton&gt; has the same; otherwise the &lt;datetimeSkeleton&gt; value is the skeleton derived from the value of &lt;pattern&gt;. Note that this currently was intended as a one-time filter for adding the initial &lt;datetimeSkeleton&gt; data; it may need refinement to use on files that already have &lt;datetimeSkeleton&gt; data.
3. Add the &lt;datetimeSkeleton&gt; data generated by running the above filter on all files in main and seed.
4. Add a test in CheckDates to check whether the skeleton in &lt;datetimeSkeleton&gt; generates the same pattern as in the associated &lt;pattern&gt;. To do this, for each CLDRFile being checked, and for each calendar type that has patterns in that CLDRFile, I had to create a DateTimePatternGenerator by iterating over the winning availableFormats items in the same order that ICU would use to build the generator, first sideways if necessary (e.g. islamic-umalqura → islamic → generic → gregorian), and then upwards from the sideways targets (gregorian, chinese) up through the parent locales to root.
5. The above test produced more than 1800 errors in ConsoleCheckCLDR. I made some adjustments to availableFormats items to fix some of these. More adjustments can be made (in a future PR), but some of the issues cannot be fixed, see example below. Consequently I made the test failure a warning instead of an error.
```
Here is an example from the zh gregorian formats. The medium and short formats are
different but have the same skeleton; we cannot generate both from the skeleton.
    <dateFormatLength type="medium">
        <dateFormat>
            <pattern>y年M月d日</pattern>
            <datetimeSkeleton>yMd</datetimeSkeleton>
        </dateFormat>
    </dateFormatLength>
    <dateFormatLength type="short">
        <dateFormat>
            <pattern>y/M/d</pattern>
            <datetimeSkeleton>yMd</datetimeSkeleton>
        </dateFormat>
    </dateFormatLength>
```
6. Notes
    * In root I added an availableFormats entry for GyMd in gregorian and generic; this will need adding in locales. I also added several availableFormats entries in chinese tp separate the use of "(U)" for only the long- and full-style skeletons.
    * In es, I changed the chinese calendar from using 'y' (makes no sense) to using 'r'.

What I did _not_ do in this PR (possibly for a future PR, along with further availableFormats adjustments to reduce warnings):
1. The &lt;datetimeSkeleton&gt; is not displayed in Survey Tool. Not sure that makes sense yet, so far it just corresponds to the &lt;pattern&gt;.
2. However the &lt;datetimeSkeleton&gt; is not currently being automatically regenerated whenever the &lt;pattern&gt; changes. For this release that is probably something that should be done by a refined version of CLDRModify -fS.
3. There are some alt=variant standard patterns (in en_CA at least) that are intended to be used with alt=variant versions of certain availableFormats items. I don't currently have a way to build up a DateTimePatternGenerator that prefers the alt=variant availableFormats items if they exist, so the alt=variant &lt;datetimeSkeleton&gt;s are not currently being tested.

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -1406,7 +1406,7 @@ $Revision$
     <!--@VALUE-->
     <!--@DEPRECATED-->
 
-<!ELEMENT dateFormat ( alias | ( pattern*, displayName*, special* ) ) >
+<!ELEMENT dateFormat ( alias | ( pattern*, datetimeSkeleton*, displayName*, special* ) ) >
 <!ATTLIST dateFormat type NMTOKEN "standard" >
     <!--@MATCH:literal/standard-->
 <!ATTLIST dateFormat alt NMTOKENS #IMPLIED >
@@ -1438,6 +1438,18 @@ $Revision$
     <!--@METADATA-->
     <!--@DEPRECATED:true, false-->
 <!ATTLIST pattern references CDATA #IMPLIED >
+    <!--@METADATA-->
+
+<!ELEMENT datetimeSkeleton ( #PCDATA ) >
+<!ATTLIST datetimeSkeleton numbers CDATA #IMPLIED >
+    <!-- TODO: generalize this to be any (M=|d=)?<numberSystem> -->
+    <!--@MATCH:literal/M=romanlow, d=hanidays, hanidec, hebr, y=jpanyear-->
+    <!--@VALUE-->
+<!ATTLIST datetimeSkeleton alt NMTOKENS #IMPLIED >
+    <!--@MATCH:literal/variant-->
+<!ATTLIST datetimeSkeleton draft (approved | contributed | provisional | unconfirmed) #IMPLIED >
+    <!--@METADATA-->
+<!ATTLIST datetimeSkeleton references CDATA #IMPLIED >
     <!--@METADATA-->
 
 <!ELEMENT displayName ( #PCDATA ) >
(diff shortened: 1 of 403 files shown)


macchiati (approved):

macchiati:
Whew, a lot of great work in here, thanks!

We can discuss more in the meeting how to deal with the issues you found. Here are some preliminary thoughts:

5. There could be a few ways to solve this. I think the easiest might be to add a new skeleton syntax character (N?) that indicated a desire for the numeric form, if available. Give N a very small weight when matching, so it is easily discarded.

Whether this would work or not would depend at least partially on whether there was a consistency among the collisions.

6.1 I think the long term goal should be to just display the datetimeSkeleton, not the pattern. But we don't have to do that now.
6.2 I agree about the refined version of -fS, until we get to the long-term goal. Thereafter, we would just always generate the pattern from the available formats, until (maybe) some day we discard it.
6.3 We should check to see if those alt=variant patterns are necessary. If so, then fixing DateTimePatternGenerator to handle alts might also be a way to handle #5.

Changes to DateTimePatternGenerator would also be require changes in ICU and other clients, so that would be a migration issue, but not one that clients would immediately need to fix.

sffc (commented):
Very cool; thank you!

pedberg-icu:
One option might be to change the mapping from pattern to skeleton so that a pattern section like M月 actually maps to MMM instead of M.

pedberg-icu:
Forked off https://unicode-org.atlassian.net/browse/CLDR-14993 for the remaining issues.

PR #1277, opened 2021-05-26
CLDR-14782 Better root abbreviated names for Islamic calendar months

pedberg-icu:
CLDR-14782

Better root names for Islamic calendar months. Note that these also change the spaces in 4 of the names from regular space to narrow no-break space U+202F (better layout behavior).

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -1713,16 +1713,16 @@ for derived annotations.
 				<months>
 					<monthContext type="format">
 						<monthWidth type="abbreviated">
-							<month type="1">Muh.</month>
-							<month type="2">Saf.</month>
-							<month type="3">Rab. I</month>
-							<month type="4">Rab. II</month>
-							<month type="5">Jum. I</month>
-							<month type="6">Jum. II</month>
-							<month type="7">Raj.</month>
-							<month type="8">Sha.</month>
-							<month type="9">Ram.</month>
-							<month type="10">Shaw.</month>
+							<month type="1">Mhrm.</month>
+							<month type="2">Safr.</month>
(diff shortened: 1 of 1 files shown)


macchiati (approved):

PeterConstable:
Some concerns are that this will be a breaking change for some parsing implementations, and questions from some reviewers as to whether the original issue warrants a change and the proposed abbreviations are an improvement:

>From Arabic speaker point of view, the old names make more sense and more understandable than the new names. And I am not seeing anything offensive in the old names.

pedberg-icu:
> From Arabic speaker point of view, the old names make more sense and more understandable than the new names. And I am not seeing anything offensive in the old names.

The issue was not that the names were offensive to Arabic speakers or to Muslims; it was that use of the name "Ram" in particular was reported as offensive to other users (apparently Hindus in this case) who saw it on the phone of someone using the Muslim calendar in English.

srl295:
> Neither set of abbreviations are things that users are likely to type into something that would parse input text; so I guess the only issue is parsing text that was previously generated from the earlier names?

similar to the "Sept" vs "Sept." issue.

macchiati:
More completely: CLDR doesn't guarantee stability of the months/days, and
especially the abbreviations. So people shouldn't depend on that for
parsing stability. There is a ticket to improve support for parsing in the
future.

Mark

pedberg-icu:
Per TC meeting, may decide to just change abbreviated name for Ramadan, e.g. to Rmdn (transliteration of Arabic would be “Rmdan”). But wait two weeks for further investigation/discussion.

macchiati (approved):

PR #1279, opened 2021-05-27
CLDR-14791 enable changing vetter email, name, password, and org

btangmu:
-Make client send change_INFO_EMAIL not CHANGE_INFO_EMAIL

-New getSelectedAction retains menu selection so POST includes change_INFO_EMAIL

-On server, after change, call reg.getInfo to update u.user for correct response to client

-On client, after changing email address, update justUser to the new address

-Remove message: click Change again to see changes; not needed and no Change to click

-Wait until finished forming response string before calling u.ua.put

-Comments

CLDR-14791

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
@@ -344,14 +344,13 @@ public class UserList {
         if (s0.equals(s1) && s0.length() > 0) {
             String s = "<h4>Change " + what + " to <tt class='codebox'>" + s0 + "</tt></h4>";
             s += "<div class='fnotebox'>" + reg.updateInfo(ctx, u.user.id, u.user.email, type, s0) + "</div>";
-            s += "<i>click Change again to see changes</i>";
+            u.user = reg.getInfo(u.user.id);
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
```diff
@@ -379,7 +379,6 @@ private void changePassword(UserRegistry.InfoType type, String action, UserSetti
         if (s0.equals(s1) && s0.length() > 0) {
             String s = "<h4>Change " + what + " to <tt class='codebox'>" + s0 + "</tt></h4>";
             s += "<div class='fnotebox'>" + reg.updateInfo(ctx, u.user.id, u.user.email, type, s0) + "</div>";
-            s += "<i>click Change again to see changes</i>";
```
srl295, line 382: Now it just works!

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
btangmu: > Now it just works!

Like a Rube Goldberg machine... :-)

PR #1280, opened 2021-05-27
CLDR-11440 add tibetan-phugpa and tibetan-tsurphu calendars

eroux:
CLDR-11440


eroux:
@pedberg-icu what do you think?

eroux:
changing the codes to `tibetan` and `tibetan-tsurphu`

pedberg-icu:
For "tibetan-phugpa" we probably want to use just "tibetan" since that is the more common form of the calendar, and "tibetan-tsurphu" will inherit most symbols data (e.g. names of months, weekdays, etc.) from "tibetan".

But in any case, CLDR want to wait on taking any action on these keywords until seeing a prototype ICU implementation per https://unicode-org.atlassian.net/browse/ICU-20178

PR #1281, opened 2021-05-27
CLDR-14385 XML Upload in gear menu

btangmu:
-Fix the UploadXML link in List Users

-Still not modernized, but works as non-SPA with .jsp

-Also hide some List Users menu items from users that lack permissions

CLDR-14385

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrAccount.js
+++ b/tools/cldr-apps/js/src/esm/cldrAccount.js
@@ -432,10 +432,12 @@ function getUserActionMenu(u, json) {
   let html = "<select class='userActionMenuSelect' name='" + theirTag + "'>\n";
   const theirLevel = u.data.userlevel;
   html += "<option value=''>" + LIST_ACTION_NONE + "</option>\n";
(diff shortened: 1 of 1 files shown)


btangmu:
I added comments in the ticket about fixing the link, which is accomplished by this PR.

This PR also hides some menu items in List Users from users without permission. Note: the back end already prohibited the items, so this change isn't urgent, but will help avoid confusion.

btangmu:
js unit test is failing. I'll fix that...

btangmu:
> js unit test is failing. I'll fix that...

Fixed. Needed to allow for session id being null during js unit test. Also, removed the "a=/cldr-apps/survey" part of the url when in List Users; it wasn't in the old version and it isn't needed (though for unknown reasons it's in the main menu version and seems to do no harm there).

srl295 (approved):
lgtm, i would mention the other ticket (about hiding items) in this PR if you can

btangmu:
> lgtm, i would mention the other ticket (about hiding items) in this PR if you can

Do you mean https://unicode-org.atlassian.net/browse/CLDR-14791 ?

PR #1282, opened 2021-05-27
CLDR-14736 add bal and bal_Latn

DavidLRowe:
CLDR-14736

I'm not sure if this PR completes the ticket, since there may be one other item that should change. Currently common/supplemental/likelySubtags.xml contains an entry for 'bal', but not one for 'bal_Latn'. But the data in likelySubtags.xml is generated from another source, so potentially another change is needed.


srl295 (approved):
Looks great to me!

PR #1283, opened 2021-05-27
CLDR-14790 improve error handling for vue problems

srl295:
CLDR-14790

- [ ] This PR completes the ticket. (nope)

### DIFF
--- a/tools/cldr-apps/js/src/cldrVueRouter.js
+++ b/tools/cldr-apps/js/src/cldrVueRouter.js
@@ -4,9 +4,10 @@ import { createApp } from "vue";
 
 import { getCldrOpts } from "./getCldrOpts";
 
-// components. See index.js for css imports.
+// See index.js for css imports.
 // example: import {SomeComponent} from 'whatever'
 import {
+  // components
   Alert,
   Button,
   Checkbox,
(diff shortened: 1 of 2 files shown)


srl295:
image

tools/cldr-apps/js/src/cldrVueRouter.js
```diff
@@ -92,10 +95,28 @@ function createCldrApp(component, specialPage, extraProps) {
...
+  }
+
+  let app;
+  try {
+    app = createCldrApp(component, specialPage, extraProps);
+    lastRoot = app.mount(el);
+  } catch (e) {
+    // did we successfully create the app?
+    errBox(e, (!!app)?"showing":"creating");
+    return null;
```
btangmu, line 118: needs formatting

I would move errBox to below its caller, and not make it nested

in the long run such a function might be more at home in cldrRetry.js for re-use

otherwise LGTM

btangmu (changes requested):

btangmu (approved):

macchiati:
This sounds like something we want to push to production this weekend if we
can so that people working on a Asian time on their Monday can get it.

srl295:
> This sounds like something we want to push to production this weekend if we
> can so that people working on a Asian time on their Monday can get it.

This is an older pr, not yet a fix for the retry loop. This improves messaging in exceptional situations.

PR #1287, opened 2021-05-28
CLDR-14498 Incorporate information from survey: phase 1

macchiati:
CLDR-14498

Note: this incorporates the preliminary data from the survey. There are still questions out for about a dozen locales that may result in further changes.

Most of the changes are adding new lines with a restriction to scope="units", to reflect that a subset of the grammatical forms are needed for normal messages with placeholders of formatted units (eg "3.5 kilometers")

There was also a change to sw and zu, which should be changed to 'no information included yet' (that is, removed from the data until we get more information). Also left a note that the data is not completely clear regarding when gender is really only scoped to 'semantic gender' (aka gender of people).

### DIFF
--- a/common/supplemental/grammaticalFeatures.xml
+++ b/common/supplemental/grammaticalFeatures.xml
@@ -193,7 +193,7 @@ plurals.xml and ordinals.xml. See those files and the LDML spec for more informa
     </grammaticalFeatures>
     <grammaticalFeatures targets="nominal" locales="lt">
       <grammaticalCase values="nominative genitive dative accusative instrumental vocative locative"/>
-      <grammaticalCase values="nominative genitive dative accusative instrumental locative"/>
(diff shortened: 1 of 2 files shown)


macchiati:
Also, I had to split gu and kn, because apparently they behave differently. (I'd kinda suspect that since one is Indo-European and the other Dravidian

I left cs hr sk sr together, and changed to what looked common among them, but need to revisit once the translators answer queries.

macchiati:
This will need some further refinement when I get back information from Micah's people but probably contains the bulk of the changes. I'll provide a follow-up PR with those.

macchiati:
Please review this: I'd like to get at least the first phase in for tomorrow (the rest will come at the EOW)

btangmu (approved):

srl295 (approved):

PR #1289, opened 2021-05-31
CLDR-14797 Make most aliased items show as Missing in ST

macchiati:
CLDR-14797

See the discussion in the ticket.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
@@ -51,6 +51,8 @@ import com.ibm.icu.util.ULocale;
  */
 public class VettingViewer<T> {
 
+    private static final boolean DEBUG = false;
+
     private static boolean SHOW_SUBTYPES = true; // CldrUtility.getProperty("SHOW_SUBTYPES", "false").equals("true");
 
     private static final String CONNECT_PREFIX = "₍_";
@@ -1242,12 +1244,30 @@ public class VettingViewer<T> {
         return result;
     }
 
+    /**
+     * Used to determine what the status of a particular path's value is in a given locale.
+     */
     public enum MissingStatus {
-        PRESENT, ALIASED, MISSING_OK, ROOT_OK, ABSENT
+        /** There is an explicit value for the path, including ↑↑↑,
+         * or there is an inherited value (but not including the ABSENT conditions, eg not from root). */
+        PRESENT,
+
+        /** The value is inherited from a different path. It might be in the same locale or from a parent (but not root or CODE_FALLBACK). */
+        ALIASED,
+
+        /** See ABSENT */
+        MISSING_OK,
+
+        /** See ABSENT  */
+        ROOT_OK,
+
+        /** The supplied CLDRFile is null, or the value is null, or the value is inherited from root or CODE_FALLBACK.
+         * A special ValuePathStatus.isMissingOk method allows for some exceptions, changing the result to  MISSING_OK or ROOT_OK. */
+        ABSENT
     }
 
     /**
-     * Get the MissingStatus
+     * Get the MissingStatus: for details see the javadoc for MissingStatus.
      *
      * @param sourceFile the CLDRFile
      * @param path the path
@@ -1258,15 +1278,23 @@ public class VettingViewer<T> {
         if (sourceFile == null) {
             return MissingStatus.ABSENT;
         }
-        if ("root".equals(sourceFile.getLocaleID()) || path.startsWith("//ldml/layout/orientation/")) {
+        final String sourceLocaleID = sourceFile.getLocaleID();
+        if ("root".equals(sourceLocaleID)
+            || path.startsWith("//ldml/layout/orientation/")) { // TODO: this path condition should probably be in isMissingOk
             return MissingStatus.MISSING_OK;
         }
         MissingStatus result;
 
         String value = sourceFile.getStringValue(path);
         Status status = new Status();
-        String sourceLocale = sourceFile.getSourceLocaleID(path, status);
+        String sourceLocale = sourceFile.getSourceLocaleID(path, status); // does not skip inheritance marker
+
         boolean isAliased = !path.equals(status.pathWhereFound); // this was path.equals, which would be incorrect!
+        if (DEBUG
+            && !isAliased
+            && !sourceLocale.equals(sourceLocaleID)) {
+            int debug = 0;
+        }
 
         if (value == null) {
             result = ValuePathStatus.isMissingOk(sourceFile, path, latin, isAliased) ? MissingStatus.MISSING_OK : MissingStatus.ABSENT;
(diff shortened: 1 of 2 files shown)


macchiati:
Note: I'm not absolutely sure this is the right approach, so we shouldn't merge until after discussion.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1265,7 +1265,7 @@ public static MissingStatus getMissingStatus(CLDRFile sourceFile, String path, b
 
         String value = sourceFile.getStringValue(path);
         Status status = new Status();
-        sourceFile.getSourceLocaleID(path, status);
+        String sourceLocale = sourceFile.getSourceLocaleID(path, status);
         boolean isAliased = !path.equals(status.pathWhereFound); // this was path.equals, which would be incorrect!
```
btangmu, line 1269: sourceLocale looks unused

macchiati: It is unused . I just assigned it because that way it was easier to debug I think the compiler will just ignore it

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1705,14 +1711,15 @@ public static void getStatus(Iterable<String> allPaths, CLDRFile file,
             // String value = file.getSourceLocaleID(path, status);
             MissingStatus missingStatus = VettingViewer.getMissingStatus(file, path, latin);
 
-            switch (missingStatus) {
+            MissingStatus adjustedMissingStatus = missingStatus  == MissingStatus.ALIASED ? aliasRemap : missingStatus;
+
```
btangmu, line 1715: are there any callers of VettingViewer.getMissingStatus, other than this one, for which the adjustment would be wrong? if so, a brief explanatory comment might help, and in any case, maybe the adjustment should happen in getMissingStatus itself; it could depend on a boolean parameter if needed

there's a general problem that each component tends to have its own slight variation on an idea like "missing status", and the variations lead to inconsistencies that are sometimes unintended ... might there be an opportunity for more unification and encapsulation here?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: The compiler will issue a warning, depending on config -- one option would be if (DEBUG) ... println ... sourceLocale

Very minor quibble! I thought I'd better check anyway that the assignment wasn't for the sake of something intended but missing.

srl295: quibble to the quibble: rather than if(DEBUG), use `logger.finest(() -> "sourceLocale:"+sourceLocale);`

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: there are multiple callers of VettingViewer.getMissingStatus, I think all within VettingViewer, but they all should give a consistent view of the missing status.

Perhaps getMissingStatus() shouldn't be public if it's only called within this object?

btangmu: it's called by TestUtilities -- otherwise, only 3 callers, all in VettingViewer -- unless I'm leaving out a .jsp

srl295:
Tests could still call it. Could be package level

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: DataSection.java includes:
```
        CLDRFile.Status sourceLocaleStatus = new CLDRFile.Status();
        String sourceLocale = ourSrc.getSourceLocaleID(xpath, sourceLocaleStatus);
```
Should that change as well for consistency?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
macchiati: I looked at https://www.geeksforgeeks.org/logger-finest-method-in-java-with-examples/, and that is a lot of function calls for something that is only needed when debugging. So I suggest we not entangle this PR, and that you put together a short doc about how and where we could deploy logging, and what the performance implications are, and we can discuss in the cldr infra meeting.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
macchiati: 1. getMissingStatus is only used internally or in tests. I thought about pushing the special logic to getMissingStatus, but decided it was better to have the lower-level method be simpler.
2. I don't know the relation to DataSection.java; the call you have is for getSourceLocaleID. Do you mean "should it change to use the Extended version"? If that's what you mean, I don't think so.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: Steven's the expert on the new logging features but I think it would just entail adding one line in VettingViewer:
```
    private static final Logger logger = SurveyLog.forClass(VettingViewer.class);
```
Anyway that's low priority.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: >    ... the relation to DataSection.java ...

It's not clear to me. My concern is whether the status as indicated in the dashboard matches the status as shown in the data section and Info Panel. We can study that further after this is merged.

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: yes, that will be documented in CLDR-5313

srl295 (approved):

btangmu:
I'll have to remember this trick: ALLOW_MANY_COMMITS=true

srl295:
requested CLA recheck wiht https://cla-assistant.io/check/unicode-org/cldr?pullRequest=1289

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
macchiati: It is used in testing; we could incorporate ICU's technique of marking methods with @internal or even come up with another annotation for methods that should only be public for testing purposes (@testing). But that would take more discussion in the cldr infra meeting.

I will add some better comments to the two methods.

PR #1290, opened 2021-06-01
CLDR-14597 Restrict organizations to specified locales

btangmu:
-Provide warnings to managers who try to assign users locales not covered by organization

-Enable managers to vote in all locales covered by organization

-Move LocaleNormalizer from cldr-apps...web to cldr-code...util

-Use `Set<CLDRLocale>` for sets of locales (avoid arrays, sets of strings, etc.)

-New class LocaleSet for encapsulation

-On startup, call LocaleNormalizer.setKnownLocales from SurveyMain

-Allow knownLocales to be null and skipped for testing without SurveyMain

-Cache the per-user sets of authorized locales and interest locales in UserRegistry.User

-New method u.exampleLocale for USER wildcard, encapsulate in UserRegistry.User

-Make some methods private if they are only called locally

-Revise unit test to call SurveyMain.getLocalesSet

-Remove dead code

-Comments

CLDR-14597


macchiati (approved):
One question: We wanted fr_* to mean any child of fr. Is that covered? (not a showstopper for this PR!!)

btangmu:
> One question: We wanted fr_* to mean any child of fr. Is that covered? (not a showstopper for this PR!!)

If a Vetter has "fr_CA" and their organization has "fr", then they're covered -- in Locales.txt, "fr" effectively includes "fr_*" (with this PR).

However, there's no actual notation like "fr_*" (with an actual asterisk) supported that I'm aware of -- this PR doesn't add support for such a notation.

macchiati:
Ok, that's great.

Mark

srl295 (commented):
post merge lgtm…

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
```diff
@@ -347,55 +350,19 @@ public int hashCode() {
...
-                return new String[0];
-            }
-        }
-
-        /**
-         * Set of interest locales for this user.
-         * @return null for 'all', otherwise a set of CLDRLocales
-         */
-        @Schema( hidden = true ) // TODO: does not seem to effectively hide
-        public Set<CLDRLocale> getInterestLocales(boolean ignoredParameter) {
```
srl295, line 384: !!!! @btangmu I missed this.   

Looks like @Schema was dropped (why?)  and ignoredParameter . They were added to prevent this function from being used in serialization.

btangmu: > Looks like `@Schema` was dropped (why?) and ignoredParameter . They were added to prevent this function from being used in serialization.

@srl Sorry for the misunderstanding. The comment seemed to mean that `@Schema` here was ineffective. `ignoredParameter` seemed to be cruft. Do either of them actually accomplish something useful? If so, let's restore them along with a clarifying comment.

srl295: Schema in other places is effective,  `ignoredParameter` was to change the serializer.  (This was part of the issue with serializing UserRegistry.User). Actually, you had dropped the serialization of User from Auth, so it is moot for now.  In any event, future modernization.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
```diff
@@ -1122,16 +1059,10 @@ String updateIntLocs(int id, boolean doCommit, Connection conn, PreparedStatemen
         removeIntLoc.setInt(1, id);
         removeIntLoc.executeUpdate();
 
-        String[] il = user.getInterestList();
-        if (il != null) {
-            updateIntLoc.setInt(1, id);
```
srl295, line 1127: this line was inadvertently dropped

btangmu: You're right! I see you've fixed it in https://github.com/unicode-org/cldr/pull/1348

PR #1267, opened 2021-05-21
CLDR-13513 seed data for hnj-Hmnp

srl295:
Green Hmong in Nyiakeng Puachue Hmong script
also fix for Hmnp numbers

CLDR-13513

- [ ] This PR completes the ticket.

PR #1291, opened 2021-06-02
CLDR-14801 Unicode 14 new scripts & script metadata

markusicu:
CLDR-14801

Note: I selectively copied some scripts from the latest language-subtag-registry (File-Date: 2021-05-11) but did not update the whole file, in order to avoid having to make changes for new language subtags etc.

I updated http://cldr.unicode.org/development/updating-codes/updating-script-metadata to match what I did, adjusting paths, simplifying instructions for VM arguments, and some other details.


macchiati (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
```diff
@@ -382,17 +382,17 @@ public void TestCoverageCompleteness() {
 
         final Pattern script100 = PatternCache.get("("
```
srl295, line 383: why is this a static list? should be pulled from data (maybe lstreg?)

srl295 (approved):
LGTM but i have questions about the script metadata process ( #1267 )

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
markusicu: Good question. I noted this in the instructions a few years ago but never investigated. See point 11 there:
- Add the script codes to TestCoverageLevel.java variable script100.
  - Why is this hardcoded here? Why does it not come from common/supplemental/coverageLevels.xml?

Thinking for a moment, I guess it can't pull it from coverageLevels.xml because that's what it wants to test, and while we can get the full list of script codes from ScriptMetadata, there is no other source for the coverage levels. But maybe we could get "all others" and test against those here? @macchiati 

I plan to merge this PR now, but we could revisit this.

srl295: > Good question. I noted this in the instructions a few years ago but never investigated. See point 11 there:
> 
> * Add the script codes to TestCoverageLevel.java variable script100.
>   
>   * Why is this hardcoded here? Why does it not come from common/supplemental/coverageLevels.xml?
> 
> Thinking for a moment, I guess it can't pull it from coverageLevels.xml because that's what it wants to test, and while we can get the full list of script codes from ScriptMetadata, there is no other source for the coverage levels. But maybe we could get "all others" and test against those here? @macchiati

It could check  locale validity against coverage levels perhaps?

markusicu: FTR  This variable should normally not need to be updated any more after https://github.com/unicode-org/cldr/pull/1296

PR #1294, opened 2021-06-02
CLDR-14797 push the change to a lower level

macchiati:
The previous PR made a fix, but it didn't show up in the Survey Tool; it needed to be at a lower level (Tom was right!)

NOTE: I filed a ticket for some cleanup in the vetting viewer: https://unicode-org.atlassian.net/browse/CLDR-14806

CLDR-14797

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ValuePathStatus.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ValuePathStatus.java
@@ -46,6 +46,9 @@ public class ValuePathStatus {
     }
 
     public static boolean isMissingOk(CLDRFile sourceFile, String path, boolean latin, boolean aliased) {
+        if (sourceFile.getLocaleID().equals("en")) {
+            return true;
+        }
         Output<String[]> arguments = new Output<>();
         List<String> failures = null;
 //        if (path.startsWith("//ldml/characters/parseLenients")) {
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
@@ -1249,20 +1249,26 @@ public class VettingViewer<T> {
      */
     public enum MissingStatus {
         /** There is an explicit value for the path, including ↑↑↑,
-         * or there is an inherited value (but not including the ABSENT conditions, eg not from root). */
+         * or there is an inherited value (but not including the ABSENT conditions, eg not from root).
+         */
         PRESENT,
 
-        /** The value is inherited from a different path. It might be in the same locale or from a parent (but not root or CODE_FALLBACK). */
+        /** The value is inherited from a different path. Only applies if the parent is not root.
+         * That path might be in the same locale or from a parent (but not root or CODE_FALLBACK).
+         */
         ALIASED,
 
-        /** See ABSENT */
+        /** See ABSENT
+         */
         MISSING_OK,
 
-        /** See ABSENT  */
+        /** See ABSENT
+         */
         ROOT_OK,
 
         /** The supplied CLDRFile is null, or the value is null, or the value is inherited from root or CODE_FALLBACK.
-         * A special ValuePathStatus.isMissingOk method allows for some exceptions, changing the result to  MISSING_OK or ROOT_OK. */
+         * A special ValuePathStatus.isMissingOk method allows for some exceptions, changing the result to  MISSING_OK or ROOT_OK.
+         */
         ABSENT
     }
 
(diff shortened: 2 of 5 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/ValuePathStatus.java
```diff
@@ -46,6 +46,9 @@ static int countZeros(String otherValue) {
     }
 
     public static boolean isMissingOk(CLDRFile sourceFile, String path, boolean latin, boolean aliased) {
+        if (sourceFile.getLocaleID().equals("en")) {
```
macchiati, line 49: This is a bit of cleanup; English is always ok.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/paths/missingOk.txt
```diff
@@ -28,3 +28,4 @@
 # More items ok to be missing
 //ldml/characters/parseLenients.* ; ok
 
+//ldml/layout/orientation/.* ; ok
```
macchiati, line 31: This was a bit of cleanup (should have been here from the start, since missingOk.txt is for all the path-specific exceptions.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1290,37 +1295,46 @@ public static MissingStatus getMissingStatus(CLDRFile sourceFile, String path, b
...
-                result = MissingStatus.ALIASED;
-            } else {
+
+            if (localeFoundIsRootOrCodeFallback) {
+                result = ValuePathStatus.isMissingOk(sourceFile, path, latin, isAliased) ? MissingStatus.ROOT_OK
+                    : isParentRoot ? MissingStatus.ABSENT
+                        : MissingStatus.ALIASED;
+            } else if (!isAliased) {
                 result = MissingStatus.PRESENT;
+            } else if (isParentRoot) { // We handle ALIASED specially, depending on whether the parent is root or not.
```
macchiati, line 1333: The main difference in the logic from v39 is that aliases in non-top locales (like en_NZ) are not flagged as ABSENT.

macchiati:
The main difference in the logic from v39 is that:

1. aliases show up as ABSENT ("Missing" in the survey tool)
2. non-top locales (like en_NZ), however, don't show that behavior.

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1290,37 +1295,46 @@ public static MissingStatus getMissingStatus(CLDRFile sourceFile, String path, b
...
+                result = ValuePathStatus.isMissingOk(sourceFile, path, latin, isAliased) ? MissingStatus.MISSING_OK
+                    : MissingStatus.ABSENT;
+            } else {
+                result = MissingStatus.ALIASED;
             }
```
btangmu, line 1338: Maybe related to https://unicode-org.atlassian.net/browse/CLDR-14827

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1724,12 +1738,6 @@ public static void getStatus(Iterable<String> allPaths, CLDRFile file,
         boolean latin = VettingViewer.isLatinScriptLocale(file);
         CoverageLevel2 coverageLevel2 = CoverageLevel2.getInstance(SupplementalDataInfo.getInstance(), file.getLocaleID());
 
-        // We handle ALIASED specially, depending on whether the parent is root or not.
-        boolean parentIsRoot = CLDRLocale.ROOT.equals(CLDRLocale.getInstance(file.getLocaleID()).getParent());
-        MissingStatus aliasRemap = parentIsRoot
-            ? MissingStatus.ABSENT
-                : MissingStatus.PRESENT;
-
```
btangmu, line 1732: Maybe related to https://unicode-org.atlassian.net/browse/CLDR-14827

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/paths/missingOk.txt
btangmu: Maybe related to https://unicode-org.atlassian.net/browse/CLDR-14827

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: This does appear to be the problematic area. In the test case I'm using, isAliased and isParentRoot are both true.

The old code would have first tested isAliased, and finding it to be true, it would have returned MissingStatus.ALIASED.

The new code does not handle the case of isAliased being true until after it returns ABSENT on the basis of isParentRoot and isMissingOk.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/paths/missingOk.txt
macchiati: This particular one wasn't, since it had been previously hardcoded.

PR #1296, opened 2021-06-03
CLDR-14801 TestCoverageLevel auto-skip scripts

markusicu:
CLDR-14801

Automatically skip not-recommended scripts.
Saves us from adding every new script to the `script100` pattern in the test code.

*Let me know if you want scripts other than Identifier_Type=Recommended to continue to be listed explicitly.*

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
@@ -16,6 +16,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


macchiati (approved):
Looks good. By the way, @pedberg-icu, we should really be dropping all of the Coverage level == 100 (comprehensive) lines from the coverage data files. It serves no purpose, because anything not at level 80 (modern) is by definition comprehensive

markusicu:
I updated item 11 of the instructions: http://cldr.unicode.org/development/updating-codes/updating-script-metadata

markusicu:
> By the way, @pedberg-icu, we should really be dropping all of the Coverage level == 100 (comprehensive) lines from the coverage data files. It serves no purpose, because anything not at level 80 (modern) is by definition comprehensive

Will this require a code change? Or could we even now just stop adding to these level100 lists?

PR #1297, opened 2021-06-03
CLDR-10560 change time cycle pref/use: CN: HH, TW: Bh, HK/MO: ah

pedberg-icu:
CLDR-10560

Change time cycle prefs for Chinese regions as described in the bug.
1. CN (zh, yue_Hans; also bo, ii, ug) changes from ah to HH (and make that first preference for pattern char C)
2. TW (zh_Hant) changes from ah to Bh. Affects standard formats, availableFormats, intervalFormats; the latter now have entries for greatestDifference="B".
3. HK,MO continue to use ah (and make that first preference for pattern char C). Per further feedback (as noted in the ticket) there was no need to split HK,MO time cycle behavior by script.
4. SG time cyle prefs are unchanged.
5. zh_Hans_HK, zh_Hans_MO, zh_Hans_SG must now override the standard time formats inherited from zh (which have HH)
6. zh_Hant_HK must now override standard time formats and 'h' formats inherited from zh_Hant (which have Bh)
7. Note that supplemental timeData preferred value must be one of H, h, K, k

### DIFF
--- a/common/main/bo.xml
+++ b/common/main/bo.xml
@@ -493,25 +493,25 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<timeFormats>
 					<timeFormatLength type="full">
 						<timeFormat>
-							<pattern>h:mm:ss a zzzz</pattern>
+							<pattern>HH:mm:ss zzzz</pattern>
 							<datetimeSkeleton>ahmmsszzzz</datetimeSkeleton>
 						</timeFormat>
 					</timeFormatLength>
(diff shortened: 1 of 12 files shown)


macchiati (approved):

pedberg-icu:
Since the default time cycle was changing for CN, I also had to update bo, ii, ug; then I had to update bo_IN to add timeFormats with the original 12-hour cycle, to match default for IN.

macchiati (approved):

PR #1305, opened 2021-06-04
CLDR-14814 fix email footer

srl295:
- [x] ought to fix it


btangmu (approved):

srl295:
Thanks. I'm going to test before merging!

srl295:
confirmed that it respects the actual server base, as well as `CLDR_SURVEY_BASE` (which is set on st.unicode.org)

PR #1309, opened 2021-06-08
CLDR-14562 update index.jsp, add top level root page

srl295:
- CLDR-14398 make /cldr-apps just redirect to /cldr-apps/v
- for the server root ( / in development ) add a redirection page as well

CLDR-14562

- [ ] This PR completes the ticket. (no, but it may complete CLDR-14398 )

### DIFF
--- a/tools/cldr-apps/src/cldr-root/index.html
+++ b/tools/cldr-apps/src/cldr-root/index.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset='utf-8'>
+    <title>CLDR Survey Tool</title>
+</head>
+
+<body>
+    <h1>
+        CLDR SurveyTool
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/cldr-root/index.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+
```
btangmu, line 3: for more consistent html formatting we could add this file to pretty.sh

btangmu (approved):

tools/cldr-apps/src/cldr-root/index.html
srl295: yes, though it shouldn't change very often!  I ran it through the built in formatter in vscode, so it at least has consistent indenting

PR #1310, opened 2021-06-09
CLDR-14498 Semi-final pass based on survey results.

macchiati:
CLDR-14498

This is further progress on reducing the number of grammatical forms we need. There will probably need to be one more PR. However, this takes the number of missing forms substantially in some locales so is worth integrating earlier.


srl295 (approved):
structure LGTM…

PR #1311, opened 2021-06-09
CLDR-14821 ST: need to trigger previous vetter vote copy

btangmu:
-After copying votes from user A to user B, enable repeat of auto-import for B

-New clear action for SurveyAjax.alreadyAutoImportedVotes

CLDR-14821


srl295 (approved):
great if that is sufficient!

btangmu:
> great if that is sufficient!

I added a comment to the ticket, explaining that Copy Votes needs to be repeated for those users to whom votes were already copied, if they had logged in during v40 before the copy -- or there are alternatives as I mentioned there...

PR #1312, opened 2021-06-09
CLDR-14790 handle missing 'special' pages

srl295:
CLDR-14790

handle case "iv" in the ticket.

Test with a URL such as http://cldr-smoke.unicode.org/cldr-apps/v#SOMETHING_NON_EXISTENT

image

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLoad.js
+++ b/tools/cldr-apps/js/src/esm/cldrLoad.js
@@ -36,6 +36,7 @@ import * as cldrText from "./cldrText.js";
 import * as cldrVettingParticipation from "./cldrVettingParticipation.js";
 import { isVueSpecial } from "../specialToComponentMap";
 import { notification } from "ant-design-vue";
+import { h } from "vue";
 
 const CLDR_LOAD_DEBUG = false;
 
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -620,6 +620,20 @@ function unspecialLoad(itemLoadInfo, theDiv) {
     }
   } else if (curSpecial) {
     console.log("No special js found for " + curSpecial);
```
btangmu, line 622: this whole block could go in a subroutine, to avoid making the function so long

tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -620,6 +620,20 @@ function unspecialLoad(itemLoadInfo, theDiv) {
...
+    cldrSurvey.hideLoader();
+    isLoading = false;
+    flipper.flipTo(pages.other, document.createTextNode("Page Not Found")); // stop loader
+    notification.warning({
+      message: "Page not found",
+      description: `The page “${curSpecial}” does not exist.
+        The link could be out-of-date, or you may have found a bug.
+        Click here to return to the SurveyTool.`,
+      duration: 0,
```
btangmu, line 631: I'd rather see an actual button to click on rather than making the whole notification act like a button; this would enable selecting and copying the text

btangmu (dismissed):

srl295:
image

updated

btangmu (approved):

manuelco-cldr:
It works

PR #1315, opened 2021-06-11
CLDR-14824 Generate new charts, and make any necessary fixes.

macchiati:
CLDR-14824

This is the first of many chart generations for v40. In this PR:

1. Fixes the omission of 'item' from root.
2. Adjusts the ChartGrammatical forms because some of the units may not have gender values yet.
3. Adds a tool ListGrammarInfo.java for generating the table at the top of http://cldr.unicode.org/translation/grammatical-inflection
4. Fixes GrammarInfo.java to have a superset of the case values, and to validate the enum values when supplemental data is read.
5. AdjustsValuePathStatus.java because count values may be integers. (and are)

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -6236,6 +6236,11 @@ annotations.
 				<unitPattern count="one">{0} mole</unitPattern>
 				<unitPattern count="other">{0} moles</unitPattern>
 			</unit>
+			<unit type="concentr-item">
+				<displayName>items</displayName>
+				<unitPattern count="one">{0} item</unitPattern>
+				<unitPattern count="other">{0} items</unitPattern>
+			</unit>
 			<unit type="consumption-liter-per-kilometer">
 				<displayName>liters per kilometer</displayName>
 				<unitPattern count="one">{0} liter per kilometer</unitPattern>
@@ -7178,6 +7183,11 @@ annotations.
 				<unitPattern count="one">{0} mol</unitPattern>
 				<unitPattern count="other">{0} mol</unitPattern>
 			</unit>
+			<unit type="concentr-item">
+				<displayName>item</displayName>
+				<unitPattern count="one">{0} item</unitPattern>
+				<unitPattern count="other">{0} item</unitPattern>
+			</unit>
 			<unit type="consumption-liter-per-kilometer">
 				<displayName>liters/km</displayName>
 				<unitPattern count="one">{0} L/km</unitPattern>
@@ -8046,6 +8056,11 @@ annotations.
 				<displayName>mol</displayName>
 				<unitPattern count="one">{0}mol</unitPattern>
 				<unitPattern count="other">{0}mol</unitPattern>
+			</unit>
+			<unit type="concentr-item">
+				<displayName>item</displayName>
+				<unitPattern count="one">{0}item</unitPattern>
+				<unitPattern count="other">{0}item</unitPattern>
 			</unit>
 			<unit type="consumption-liter-per-kilometer">
 				<displayName>L/km</displayName>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
@@ -620,7 +620,7 @@ public class ChartGrammaticalForms extends Chart {
                     final UnitId unitId = uc.createUnitId(bestUnit);
                     unitPattern = unitId.toString(ENGLISH, "long", pluralCategory, null, null, false);
                     if (unitPattern == null) {
-                        int debug = 0;
+                        return null;
                     }
                 }
                 String unitMeasure = MessageFormat.format(unitPattern, string.contains("/") ? "~" + bestDoubleFactor : string);
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
@@ -318,9 +318,9 @@ public class ChartGrammaticalForms extends Chart {
         }
         unitToBestUnit = ImmutableMap.copyOf(unitToBestUnit);
         // quick check
-        final BestUnitForGender u1 = unitToBestUnit.get("meter");
-        final BestUnitForGender u2 = unitToBestUnit.get("square-centimeter");
-        int comp = u1.compareTo(u2); // should be less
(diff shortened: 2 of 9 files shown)


common/main/en.xml
```diff
@@ -7178,6 +7183,11 @@ annotations.
 				<unitPattern count="one">{0} mol</unitPattern>
 				<unitPattern count="other">{0} mol</unitPattern>
 			</unit>
+			<unit type="concentr-item">
+				<displayName>item</displayName>
+				<unitPattern count="one">{0} item</unitPattern>
+				<unitPattern count="other">{0} item</unitPattern>
```
srl295, line 7189: doesn't fall back ? I guess we'd need lots of aliases in root for that.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
```diff
@@ -400,8 +400,13 @@ public void writeSubcharts(Anchors anchors) throws IOException {
                     String quantity = shortUnit.contentEquals("generic") ? "temperature" : uc.getQuantityFromUnit(shortUnit, false);
 
                     String gender = UnitPathType.gender.getTrans(cldrFile, "long", shortUnit, null, null, null, null);
+                    gender = gender == null ? "n/a" : gender;
                     Set<String> systems = uc.getSystems(shortUnit);
 
+                    if (unitCell == null || quantity == null || gender == null || sizeInBaseUnits.value == null) {
+                        throw new IllegalArgumentException("No best base unit for: " + shortUnit);
+                    }
+
```
srl295, line 409: `gender == null` can't happen due to line 403, will be at least n/a

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
```diff
@@ -455,7 +460,7 @@ public void writeSubcharts(Anchors anchors) throws IOException {
                     final String shortUnit = uc.getShortId(longUnit);
                     String gender = UnitPathType.gender.getTrans(cldrFile, "long", shortUnit, null, null, null, null);
                     final BestUnitForGender bestUnit = unitToBestUnit.get(shortUnit);
-                    if (bestUnit != null) {
+                    if (gender != null && bestUnit != null) {
                         bestUnitForGender.put(gender, bestUnit);
                     }
```
srl295, line 465: same, should just be `(bestUnit != null)` ?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListGrammarInfo.java
```diff
@@ -0,0 +1,73 @@
...
+import org.unicode.cldr.util.GrammarInfo;
+import org.unicode.cldr.util.GrammarInfo.GrammaticalFeature;
+import org.unicode.cldr.util.GrammarInfo.GrammaticalScope;
+import org.unicode.cldr.util.GrammarInfo.GrammaticalTarget;
+import org.unicode.cldr.util.LanguageTagParser;
+import org.unicode.cldr.util.SupplementalDataInfo;
+
+import com.google.common.base.Joiner;
+
+public class ListGrammarInfo {
```
srl295, line 18:
```suggestion

@CLDRTool(alias="list-grammar-info", description = "Lists Grammar Info…")
public class ListGrammarInfo {
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListGrammarInfo.java
```diff
@@ -0,0 +1,73 @@
...
+            ltp.set(locale);
+            String region = ltp.getRegion();
+            if (!region.isEmpty()) {
+                continue;
+            }
+            GrammarInfo grammarInfo = SDI.getGrammarInfo(locale, true);
+            if (grammarInfo == null || !grammarInfo.hasInfo(GrammaticalTarget.nominal)) {
+                continue;
+            }
+            //CLDRFile cldrFile = factory.make(locale, true);
```
srl295, line 42: dead code

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
```diff
@@ -36,16 +36,17 @@
...
 
     /**
-     * The ordering of these values is intended to put the default values first, and to group values together that tend to have similar forms.
+     * The ordering of these values is intended to put the default values first, and to group values together that tend to have similar forms for the most common cases,
+     * then have the rest in alphabetical order.
      */
-    public enum CaseValues {nominative, vocative, accusative, oblique, genitive, dative, locative, instrumental, prepositional,
-        ablative, adessive, allative, causal, delative, elative, essive, illative, inessive, sublative, superessive, terminative, translative;
+    public enum CaseValues {nominative, vocative, accusative, oblique, genitive, dative, locative, instrumental, prepositional, ablative,
+        abessive, adessive, allative, causal, comitative, delative, elative, ergative, essive, illative, inessive, locativecopulative, partitive, sociative, sublative, superessive, terminative, translative;
```
srl295, line 43: is this verified against DTD / supplemental metadata?

srl295 (approved):
fine w/ comments (could be future)

common/main/en.xml
macchiati: There is a general fallback between widths of the items, but en.xml needs to be "fully stocked"

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
macchiati: Good point. I'll pick that up next time.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
macchiati: ditto (these were from before there was a null check in the code.

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
macchiati: Yes, there is a general verification of the valid values in the DTD. This is to align the enum with the DTD values.

I've been thinking for some time that we could simplify our lives if we had a simple program that generated enums from the valid DTD values (both @match values and the regular DTD options). Don't know whether it would be worth the extra step of running it whenever we change the DTD.

PR #1317, opened 2021-06-11
CLDR-14790 handle another case where ST could not coldstart

srl295:
Bugs Everywhere! Well, hopefully not. This PR catches the rest of the err cases (the known knowns!) that
ought to make "ST seems stuck/dead/hung" move even more towards 0%.

- the 'waiting' page is not able to run cldrGui.run(), we just need to reload first.
- make cldrGui.run() throw an exception instead of 'doing nothing'.
- also, annotate which function generated the /v page
- show err if there's a problem mounting the main header
- fix a css err in dashboardwidget
- add an error handler when mounting vue components

CLDR-14790

Note: The following are all _simulated_ errors! (hence 'foo')

## Typo in MainHeader.vue

TypeError while loading

## Typo in Dashboard (hence 'vue component') as the special name is not available

hi  This site uses cookies

## Typo in some other page (`transfervotes`)

This site uses cookies

### DIFF
--- a/tools/cldr-apps/js/src/views/WaitingPanel.vue
+++ b/tools/cldr-apps/js/src/views/WaitingPanel.vue
@@ -113,7 +113,6 @@ export default {
             if (this.$specialPage == "retry") {
               // immediately head back to the main page.
               window.location.replace("v#");
-              run(); // reboot everything
               setTimeout(
                 () =>
                   notification.success({
@@ -122,6 +121,15 @@ export default {
                   }),
                 4000
               );
+              run().catch((e) => {
+                // We can get here if run() was not able to boot the page
+                // That's OK, it may have been a waiting page. Reload should
+                // clear it up.
+                console.log(
+                  `run() threw an error, so we will do a page reload. Err was: ${e}`
+                );
+                window.location.reload();
+              });
             } else {
               window.setTimeout(() => window.location.reload(), NORMAL_RETRY);
             }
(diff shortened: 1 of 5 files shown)


tools/cldr-apps/js/src/views/WaitingPanel.vue
```diff
@@ -122,6 +121,14 @@ export default {
                   }),
                 4000
               );
+              run()
+              .catch(e => {
+                // We can get here if run() was not able to boot the page
+                // That's OK, it may have been a waiting page. Reload should
+                // clear it up.
+                console.log(`run() threw an error, so we will do a page reload. Err was: ${e}`);
```
macchiati, line 129: Question; if there is a persistent problem loading, will this go into an infinite loop?

macchiati (approved):
Looks ok with one caveat; approved in case it is better than current anyway.

tools/cldr-apps/js/src/views/WaitingPanel.vue
srl295: @macchiati before getting to this line, the front end checks with the server to make sure that it's ready to start up.  The only error here is that it wasn't able to start the surveytool front end _without reloading the page_.  If a problem happened after reload, it should be caught by the other err handlers that this PR puts in.

srl295:
forgot formatting

macchiati (approved):

btangmu (approved):

manuelco-cldr:
The https://st.unicode.org/cldr-apps/v#abcdefghijk/// still keep saying "Loading..." perpetually.

PR #1319, opened 2021-06-14
CLDR-14827 CLDR survey tool dashboard showing incorrect number of issues

btangmu:
-getMissingStatus return ALIASED if isAliased && !localeFoundIsRootOrCodeFallback

-The logic was previously changed by PR 1294

CLDR-14827

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
@@ -1328,13 +1328,10 @@ public class VettingViewer<T> {
                 result = ValuePathStatus.isMissingOk(sourceFile, path, latin, isAliased) ? MissingStatus.ROOT_OK
                     : isParentRoot ? MissingStatus.ABSENT
                         : MissingStatus.ALIASED;
-            } else if (!isAliased) {
-                result = MissingStatus.PRESENT;
-            } else if (isParentRoot) { // We handle ALIASED specially, depending on whether the parent is root or not.
-                result = ValuePathStatus.isMissingOk(sourceFile, path, latin, isAliased) ? MissingStatus.MISSING_OK
-                    : MissingStatus.ABSENT;
-            } else {
+            } else if (isAliased) {
                 result = MissingStatus.ALIASED;
+            } else {
+                result = MissingStatus.PRESENT;
             }
         }
         return result;


btangmu:
This draft PR fails this test, for "hr":
```
    public void TestMissingGrammar() {
        // https://cldr-smoke.unicode.org/cldr-apps/v#/hu/Length/a4915bf505ffb49
        final String path = "//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"length-meter\"]/unitPattern[@count=\"one\"][@case=\"accusative\"]";
        checkGrammarCoverage("hr", path,    MissingStatus.ABSENT, DEBUG, 0, 0, 1, 1, 0);
        checkGrammarCoverage("kw", path, MissingStatus.ABSENT, false, 0, 0, 1, 1, 0);
        checkGrammarCoverage("en_NZ", path, MissingStatus.ALIASED, DEBUG, 1, 0, 0, 0, 0);
    }
```

@macchiati I don't understand why MissingStatus.ABSENT would be expected.

getMissingStatus includes this comment:
```
            /*
             * Only count it as missing IF the (localeFound is root or codeFallback)
             * AND the aliasing didn't change the path.
             * Note that localeFound will be where an item with ↑↑↑ was found even though
             * the resolved value is actually inherited from somewhere else.
             */
```

btangmu:
BTW TestMissingGrammar was also changed in https://github.com/unicode-org/cldr/pull/1294 -- in the old test, ALIASED (not ABSENT) was expected for "hr".

btangmu:
@macchiati , when I step through the code for the "hr" test, stepping into getMissingStatus, I find that localeFound and sourceLocaleID are both "hr"; path is `//ldml/units/unitLength[@type="long"]/unit[@type="length-meter"]/unitPattern[@count="one"][@case="accusative"]` which is different from status.pathWhereFound = `//ldml/units/unitLength[@type="long"]/unit[@type="length-meter"]/unitPattern[@count="one"]`. 

So, localeFound is neither root nor codeFallback, and aliasing DID change the path. For both those reasons, based on this comment,

```
            /*
             * Only count it as missing IF the (localeFound is root or codeFallback)
             * AND the aliasing didn't change the path.
             * Note that localeFound will be where an item with ↑↑↑ was found even though
             * the resolved value is actually inherited from somewhere else.
             */
```

it's correct to ALIASED, not ABSENT.

If the comment is right, then getMissingStatus is right, and TestMissingGrammar is wrong.

If the comment is wrong, what should it say?

btangmu:
See https://github.com/unicode-org/cldr/pull/1323

PR #1321, opened 2021-06-15
CLDR-14804 update Locales.txt per discussion

srl295:
- fix casing of some orgs
- (re-)add * to SIL, wikimedia
- fix corrupted chars in comments
- re-instate obvious participation locales
- add '*' to Apple, IBM, Microsoft, Google

CLDR-14804

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
@@ -373,6 +373,11 @@ public class StandardCodes {
     };
 
     /**
+     * Used for Locales.txt to mean "all"
+     */
+    public static final String ALL_LOCALES = "*";
+
+    /**
      * Returns locales according to status. It returns a Map of Maps, key 1 is
      * either IBM or Java (perhaps more later), key 2 is the Level.
      *
@@ -437,7 +442,7 @@ public class StandardCodes {
             }
             desiredLocale = LocaleIDParser.getParent(desiredLocale);
         }
-        Level status = locale_status.get("*");
+        Level status = locale_status.get(ALL_LOCALES);
         if (status != null && status != Level.UNDETERMINED) {
             coverageType.value = LocaleCoverageType.star;
             return status;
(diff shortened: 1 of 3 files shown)


srl295:
Does not add * to (Google, Apple. Microsoft, IBM) per CLDR-14832 but that could be piggybacked here.

DavidLRowe:
@srl295 In line
```
SIL	;	kk	;	moderate	;
```
I believe "kk" should be "kkj". In the past SIL has contributed to kkj (Kako, Latn, CM). This was part of the same effort that collected data for jgo, mgo and nnh (three other Cameroonian languages). To my knowledge, SIL has never contributed to kk (Kazakh, Cyrl, KZ). 

Can this be corrected as part of this ticket? or does it need a separate ticket?

nemobis:
+1, thank you. I've also commented at https://unicode-org.atlassian.net/browse/CLDR-14597?focusedCommentId=161551

srl295:
> @srl295 In line
> 
> ```
> SIL	;	kk	;	moderate	;
> ```
> 
> I believe "kk" should be "kkj". In the past SIL has contributed to kkj (Kako, Latn, CM). This was part of the same effort that collected data for jgo, mgo and nnh (three other Cameroonian languages). To my knowledge, SIL has never contributed to kk (Kazakh, Cyrl, KZ).
> 
> Can this be corrected as part of this ticket? or does it need a separate ticket?

@DavidLRowe it certainly can be corrected here, I think the focus of _this_ ticket should be unblocking organizations.  As SIL gets an `*` SIL is not restricted after this PR.   If the _only_ change that needs to be made is kk to kkj, then that's ok, otherwise I would suggest a separate PR to update any coverage levels and locales you would like to note.

srl295:
fixed tests. apparently some code did not expect `*`

macchiati:
Yes, the * needs to be added to those orgs. If not on their PR then in a
follow-up.

Mark (📱)

srl295:
> Yes, the * needs to be added to those orgs. If not on their PR then in a follow-up. Mark (📱)
> […](#)

It's done in this PR

macchiati (approved):

PR #1324, opened 2021-06-16
CLDR-13928 Value for 20-vote item moved to Winning column with 8 votes

btangmu:
-Add tests for two rules:

--- (1) For a 20 vote item, 8 votes are needed to change from non-approved to approved

--- (2) Once Approved, it requires 20 votes to change

-Revise VoteResolver to pass both tests; adjust from HIGH_BAR (20) to LOWER_BAR (8) if not approved

-Revise SurveyJSONWrapper.wrap(VoteResolver) to send client the adjusted required votes

-Remove unneeded 3rd arg of computeStatus (always trunkStatus)

CLDR-13928

### DIFF
--- a/tools/cldr-apps/src/test/resources/org/unicode/cldr/unittest/web/data/TestSTFactory.xml
+++ b/tools/cldr-apps/src/test/resources/org/unicode/cldr/unittest/web/data/TestSTFactory.xml
@@ -202,4 +202,29 @@
 	<vote name="OrgStatM1" locale="fr_CA" xpath="//ldml/annotations/annotation[@cp='🔫'][@type='tts']">pistolet à eau</vote>
 	<vote name="OrgStatG1" locale="fr_CA" xpath="//ldml/annotations/annotation[@cp='🔫'][@type='tts']">↑↑↑</vote>
 	<verify status="approved" orgStatus="ok" statusOrg="google" locale="fr_CA" xpath="//ldml/annotations/annotation[@cp='🔫'][@type='tts']">pistolet à eau</verify>
+
+	<echo>Tests involving 20-vote items</echo>
+
+	<!-- Rules:
+	       1. For a 20 vote item, 8 votes are needed to change from non-approved to approved.
+	       2. Once Approved, it requires 20 votes to change.
+	     Note: this test assumes and requires that the .../minimumGroupingDigits path
+	     is "contributed" in fr.xml, and that the .../decimal path is "approved" in fr.xml.
+	     Ideally this test might mock the data to avoid the assumption.
+	     Compare TestUtilities.TestVoteResolver where it is possible to set oldStatus, etc. -->
+	<user name="TwenA" org="apple" level="vetter" locales="*"/>
+	<user name="TwenB" org="ibm" level="vetter" locales="*"/>
+
+	<verify status="contributed" locale="fr" xpath="//ldml/numbers/minimumGroupingDigits">1</verify>
+	<vote name="TwenA" locale="fr" xpath="//ldml/numbers/minimumGroupingDigits">2</vote>
+	<vote name="TwenB" locale="fr" xpath="//ldml/numbers/minimumGroupingDigits">2</vote>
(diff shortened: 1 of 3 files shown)


macchiati (approved):

btangmu:
I'll wait until after today's push to production to merge this.

macchiati:
sounds good. This would revert any winning values anyway after we deploy,
so there is no big urgency.
Mark

tools/cldr-apps/src/test/resources/org/unicode/cldr/unittest/web/data/TestSTFactory.xml
```diff
@@ -202,4 +202,29 @@
...
 	<verify status="approved" orgStatus="ok" statusOrg="google" locale="fr_CA" xpath="//ldml/annotations/annotation[@cp='🔫'][@type='tts']">pistolet à eau</verify>
+
+	<echo>Tests involving 20-vote items</echo>
+
+	<!-- Rules:
+	       1. For a 20 vote item, 8 votes are needed to change from non-approved to approved.
+	       2. Once Approved, it requires 20 votes to change.
+	     Note: this test assumes and requires that the .../minimumGroupingDigits path
+	     is "contributed" in fr.xml, and that the .../decimal path is "approved" in fr.xml.
+	     Ideally this test might mock the data to avoid the assumption.
```
srl295, line 213: it might be good to have such a mock.

We already have the sandbox locales ( see CLDR-12002 ) so there's precedent for having multiple input dirs.  Actually, a `mul` sublocale such as `mul-AA` might make sense. it could have static test data, and get copied into the sandbox directory on startup for test use.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -738,6 +740,19 @@ public int getRequiredVotes() {
         return requiredVotes;
     }
 
+    /**
+     * What are the adjusted required votes for this item?
```
srl295, line 744:
```suggestion
     * What are the adjusted required votes for this item, taking into account the trunkStatus?
```

srl295: I would also update getRequiredVotes to say that it _only_ takes the xpath and locale into account. 

`canFlagOnLosing()` wasn't updated, should it be? 

Perhaps this function should be `getRequiredVotes()` as this is the _actual_ required votes, and the other one renamed to `getUnadjustedRequiredVotes()`

I'm splitting hairs here, but this is critical core machinery, and can be challenging to trace through. ('Adjusted' doesn't tell me much about what is happening, and I know we have several 'adjusted' functions in various parts now)

srl295 (approved):
LGTM with usual rants about naming, this is software after all :)

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: canFlagOnLosing works correctly, as it stands, since requiredVotes == HIGH_BAR even when getAdjustedRequiredVotes may return LOWER_BAR

I like your suggestions -- the "adjusted" business has a disagreeable smell! there's definitely a lot of room for improvement in the implementation and naming and comments. I was trying to avoid touching/changing more of the code for this PR than was really necessary to accomplish the goal of this ticket; maybe too cautious...

btangmu:
I forgot to merge this a week ago! The plan was to wait until after a push to production that was about to happen. I've checked there aren't more recent changes to the 3 files, so I'll merge now

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: actually maybe I was wrong about canFlagOnLosing -- I'll email cldr-core2 for policy clarification

PR #1325, opened 2021-06-17
CLDR-14801 Unicode 14 root & zh collation & translit

markusicu:
CLDR-14801


macchiati (commented):
common/uca/UCA_Rules_SHORT.txt

I was surprised to see some removals, such as removing the following line.

<<<	 ‎ٶ‎ / ‎ء

I would have just expected insertions. Do you know anything about those? (There are only a few.)

markusicu:
> common/uca/UCA_Rules_SHORT.txt
> I was surprised to see some removals, such as removing the following line.
> <<< ‎ٶ‎ / ‎ء
> 
> I would have just expected insertions. Do you know anything about those? (There are only a few.)

I don't know. I can try to investigate if you think it's important, but with this large amount of data, if the output looks largely right, I tend to trust the tool...

srl295 (approved):
RSLGTM

markusicu:
> > common/uca/UCA_Rules_SHORT.txt
> > I was surprised to see some removals, such as removing the following line.
> > <<< ‎ٶ‎ / ‎ء
> > I would have just expected insertions. Do you know anything about those? (There are only a few.)
> 
> I don't know. I can try to investigate if you think it's important, but with this large amount of data, if the output looks largely right, I tend to trust the tool...

Easier than I feared. This one is U+0676, and this is a bug fix Ken made for a UTC action item: https://github.com/unicode-org/unicodetools/commit/a427146d6d918220926ae9b5c486d3d207f39cc5

Other “removals” also look like a *change* in ordering (characters sort slightly earlier or later than they used to), or even just an insertion of another character that changes something like `<< x` to `<< y` + `= x`.

macchiati:
Sounds great, thanks.

PR #1328, opened 2021-06-22
CLDR-14841 create production tag on push to production

srl295:
- Create a tag of the form `production/2021-06-22-2206z` during
the manual push to production process.
- The date and time are UTC (June 22nd, 22:06 UTC)

CLDR-14841

- [ ] This PR completes the ticket. (no, there are other items

Example tag: https://github.com/srl295/cldr/releases/tag/production%2F2021-06-22-2206z

I have tested the exact code stanza involved, but have not tested an actual push to production (I had my test case stop before continuing that far).

Probably once this is merged, we will want to test it with an actual push. It could wait to merge until the next time we're doing such a push.

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -24,6 +24,10 @@ jobs:
         with:
           ref: ${{ github.event.inputs.git-ref }}
           lfs: false
+      - name: Calculate Tag Name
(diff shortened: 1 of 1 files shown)


btangmu (approved):
interesting!

PR #1329, opened 2021-06-24
CLDR-14850 speed up mail by reusing connection

srl295:
also, improve debugging

CLDR-14850

This is on a branch as  e547a7fb5e9 e518310dab81c1268762f18540f7c to be pushed to production as a hotfix.

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
@@ -367,24 +367,33 @@ public class MailSender implements Runnable {
             return;
         }
 
-        processAllMail();
+        logger.finest("processing all mail…");
+        Transport transport = processAllMail();
+        try {
+            if (transport != null) {
+                transport.close();
+            }
+        } catch (MessagingException me) {
+            logger.log(Level.WARNING, "Could not close transport", me);
+        }
         logger.finer("MailSender.run() done");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
```diff
@@ -72,6 +72,9 @@
                 SurveyMain.isUnofficial() ? 5 : 60);
             // Check for outbound mail every 60 seconds
             CLDR_MAIL_DELAY_EACH = CLDRConfig.getInstance().getProperty("CLDR_MAIL_DELAY_EACH", 60);
+
+
+            System.err.println("Logger settings " + logger.getLevel().toString() + " for " + logger.getName());
         }
```
btangmu, line 78: I'm assuming this println won't happen often enough to clog the logs, and/or it's temporary -- also ideally shouldn't go to ".err" if it's not an error

btangmu (dismissed):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/MailSender.java
srl295: Temporary. I'll delete it

srl295: @btangmu actually that is printed exactly one time at startup. Will remove in the PR.

btangmu (approved):

PR #1331, opened 2021-06-25
CLDR-14808 Revamp unit width tests to be more granular

macchiati:
CLDR-14808

Restructures the unit width checks for excessive length (short and narrow) to be much more granular. That is, it takes into account the subunits (eg in kilowatt-hour) and allows non-metric units to be longer (since there are often no good abbreviations for them in languages with metric units).

Information is cached for the extra unit data, so we shouldn't see significant performance cost. Would appreciate Steven's taking a look at the caching.

Note: makes some small changes in en.xml to fix excessive widths (don't want warnings from en).

### DIFF
--- a/common/annotations/en.xml
+++ b/common/annotations/en.xml
@@ -250,8 +250,8 @@ annotations.
 		<annotation cp="↷" type="tts">clockwise top semicircle arrow</annotation>
 		<annotation cp="↸">north west arrow long bar</annotation>
 		<annotation cp="↸" type="tts">north west arrow long bar</annotation>
-		<annotation cp="↹">leftwards arrow bar over rightwards arrow bar</annotation>
-		<annotation cp="↹" type="tts">leftwards arrow bar over rightwards arrow bar</annotation>
+		<annotation cp="↹">leftwards | arrow | bar | rightwards</annotation>
+		<annotation cp="↹" type="tts">leftwards over rightwards arrow bars</annotation>
 		<annotation cp="↺">anticlockwise open circle arrow</annotation>
 		<annotation cp="↺" type="tts">anticlockwise open circle arrow</annotation>
 		<annotation cp="↻">clockwise open circle arrow</annotation>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckWidths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckWidths.java
@@ -119,7 +119,7 @@ public class CheckWidths extends CheckCLDR {
             this(d, e, displayWidth, maximum, placeholders, false);
         }
 
-        boolean hasProblem(String path, String value, List<CheckStatus> result, CheckCLDR cause, Boolean aliasedAndComprenehsive) {
+        boolean hasProblem(String path, String value, List<CheckStatus> result, CheckCLDR cause, Boolean aliasedAndComprehensive) {
             double factor = 1d;
             switch (special) {
             case NUMBERFORMAT:
@@ -153,9 +153,9 @@ public class CheckWidths extends CheckCLDR {
                 if (valueMeasure >= warningReference) {
                     return false;
                 }
-                if (valueMeasure < errorReference 
-                	&& cause.getPhase() != Phase.BUILD 
-                	&& !aliasedAndComprenehsive) {
+                if (valueMeasure < errorReference
+                	&& cause.getPhase() != Phase.BUILD
+                	&& !aliasedAndComprehensive) {
                     errorType = CheckStatus.errorType;
                 }
                 break;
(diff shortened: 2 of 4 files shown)


common/annotations/en.xml
```diff
@@ -250,8 +250,8 @@ annotations.
 		<annotation cp="↷" type="tts">clockwise top semicircle arrow</annotation>
 		<annotation cp="↸">north west arrow long bar</annotation>
 		<annotation cp="↸" type="tts">north west arrow long bar</annotation>
-		<annotation cp="↹">leftwards arrow bar over rightwards arrow bar</annotation>
-		<annotation cp="↹" type="tts">leftwards arrow bar over rightwards arrow bar</annotation>
+		<annotation cp="↹">leftwards | arrow | bar | rightwards</annotation>
+		<annotation cp="↹" type="tts">leftwards over rightwards arrow bars</annotation>
```
btangmu, line 254: The addition of vertical line characters here differs from the usual pattern in this file. Does it matter?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckWidths.java
```diff
@@ -103,7 +119,8 @@ public Limit(double d, double e, Measure displayWidth, LimitType maximum, Specia
             this(d, e, displayWidth, maximum, placeholders, false);
         }
 
-        boolean hasProblem(String value, List<CheckStatus> result, CheckCLDR cause, Boolean aliasedAndComprenehsive) {
+        boolean hasProblem(String path, String value, List<CheckStatus> result, CheckCLDR cause, Boolean aliasedAndComprenehsive) {
+            double factor = 1d;
```
btangmu, line 123: "aliasedAndComprenehsive" is misspelled, but that's not new and it must be used consistently or the compiler would have complained

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckWidths.java
```diff
@@ -40,4 +42,27 @@ public void TestBasic() {
             assertEquals("errorSubType", Subtype.valueTooWide, err.getSubtype());
         }
     }
+
+    public void TestEnglishWidths() {
+        UnitWidthUtil unitWidthUtils = CheckWidths.UnitWidthUtil.getInstance();
+        CLDRConfig config = CLDRConfig.getInstance();
+        CLDRFile english = config.getEnglish();
+        CheckWidths x = new CheckWidths();
+        CheckWidths.UnitWidthUtil.getInstance();
```
btangmu, line 51: Is line 51 superfluous? The returned instance isn't assigned to a variable, and we already have an instance from line 47

btangmu (commented):
LGTM except a couple uncertain points where I added comments

common/annotations/en.xml
macchiati: That is a problem in the new symbols. I made a ticket for Andrew to take to fix them, but went ahead with this one.

https://unicode-org.atlassian.net/browse/CLDR-14859

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckWidths.java
macchiati: fixed

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckWidths.java
macchiati: fixed

macchiati:
Just fixed those issues, so ready for re-review.

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckWidths.java
```diff
@@ -369,4 +353,87 @@ public CheckCLDR setCldrFileToCheck(CLDRFile cldrFileToCheck, Options options,
...
+
+        LoadingCache<String, Double> unitToComponents = CacheBuilder.newBuilder().build(new CacheLoader<String, Double>() {
+            @Override
+            public Double load(String longUnitId) {
+                double components = 0;
+                String shortId = CONVERTER.getShortId(longUnitId);
+
+                Set<String> systems = CONVERTER.getSystems(shortId);
+                int widthFactor = systems.contains("metric") &&  !shortId.endsWith("-metric") ? 1 : 3;
+                // NOTE: allow cup-metric and pint-metric to be longer, since they aren't standard metric
```
srl295, line 388: so metric, but not ending in `-metric` ? could this collide with anything else?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckWidths.java
```diff
@@ -40,4 +42,26 @@ public void TestBasic() {
...
+        List<CheckStatus> possibleErrors = new ArrayList<>();
+        x.setCldrFileToCheck(english, options, possibleErrors);
+        assertEquals("setCLDRFile", Collections.emptyList(), possibleErrors);
+        for (String path : config.getEnglish().fullIterable()) {
+            String fullPath = english.getFullXPath(path);
+            String value = english.getStringValue(path);
+            x.check(path, fullPath, value, options, possibleErrors);
+            if (!assertEquals(path + " = " + value, Collections.emptyList(), possibleErrors) || isVerbose()) {
+                // for debug
+                double roughMax = unitWidthUtils.getRoughComponentMax(path);
```
srl295, line 61: ideally we'd have some specific API level (instead of check-level) tests here, especially checking `metric-cup`, `100-…` etc.. could be future.

srl295 (approved):
LGTM with comments for the future!

macchiati:
Left the ticket open for the last comments

PR #1335, opened 2021-06-27
CLDR-14807 fix day-person collisions

macchiati:
CLDR-14807 

The ticket will have 3 parts:

1. Make sure the special durations for persons don't collide.

2. Make the alias mechanism more robust where it could need recursion.
3. Make the iteration over aliases also more robust.

This PR should address the first item, but more work needs to be done on #2 and #3, which were encountered while doing #1. It is worth starting with #1, since it will unblock the vetters on fixing current errors.

The PR also adds a tool used in testing.


srl295 (approved):
needs to be data driven… but anyway LGTM!

PR #1336, opened 2021-06-28
CLDR-14384 Modernize Lock My Account in main menu

btangmu:
-New LockAccount.vue

-New api/auth/lock implemented in Auth.java and new LockRequest.java

-Log out the user once they have been locked

-Revise UNSUBSCRIBE message in MailSender

-New constant UserRegistry.ADMIN_EMAIL to encapsulate admin@

-Remove lock.jsp

-Eclipse settings auto-updated 39 to 40 for cldr-code-40.0-SNAPSHOT.jar

-Format LoginButton.vue (not otherwise changed in this PR)

-Remove already-dead code from MainHeader.vue (not otherwise changed in this PR)

-Comments

CLDR-14384


tools/cldr-apps/js/src/specialToComponentMap.js
```diff
@@ -17,6 +18,7 @@ const specialToComponentMap = {
   about: AboutPanel,
   add_user: AddUser,
   auto_import: AutoImport,
+  lock_account: LockAccount,
```
srl295, line 21: hooray, another modern component! 🎆 (Glad I made that the default for specials…)

tools/cldr-apps/js/src/views/LockAccount.vue
```diff
@@ -0,0 +1,178 @@
...
+      cldrStatus.setSurveyUser(null);
+      cldrStatus.setPermissions(null);
+      cldrStatus.setSessionId(null);
+      // The success message will be displayed in the main window, replacing the form,
+      // as soon as Vue renders it. However, it may quickly be replaced with something else
+      // as a result of the disconnection, since the user has been logged out on the back end.
+      // To ensure the user has a chance to read the message, also open an alert box with the
+      // same message, then redirect to the default page. Ideally (though not necessarily) the
+      // alert should open after Vue has done the rendering, and before other messages start
+      // showing up about being disconnected, for which ten milliseconds seems to work OK.
```
srl295, line 144: an alternative would be to fetch `POST /api/auth/login` with a payload of `{}` which will give a new anonymous session ID, that could be setup as the continuing session ID. It would be the effect of a refresh.

srl295: it will return:
```json
{   "sessionId": "MTczMjg4NzIwMzEzNzgyMTc0NQo",  "user": false }
```

srl295 (approved):
LGTM and thanks for modernizing. Commented w/ Idea for improving the logout experience.

tools/cldr-apps/js/src/views/LockAccount.vue
btangmu: That does sound better. I'll merge this PR as-is and then experiment with your idea...

PR #1340, opened 2021-06-29
CLDR-13853 Reduce the number of English Changed

macchiati:
CLDR-13853

Make two significant changes.

1. Values are only counted as different if their DAIP normalization is also different
2. A new parameter is introduced in GenerateBirths that lets us limit how far back we check for cases where English changed (and the locale didn't). The default is 36.0.


btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrVersion.java
```diff
@@ -28,7 +28,7 @@
     v1_1_1, v1_2, v1_3, v1_4_1, v1_5_1, v1_6_1, v1_7_2, v1_8_1, v1_9_1, v2_0_1,
     v21_0, v22_1, v23_1, v24_0, v25_0, v26_0, v27_0, v28_0, v29_0, v30_0, v31_0, v32_0, v33_0, v33_1, v34_0,
     v35_0, v35_1, v36_0, v36_1, v37_0, v38_0, v38_1, v39_0,
-    trunk;
+    baseline;
```
srl295, line 31: :+1:

srl295 (approved):

PR #1341, opened 2021-06-29
CLDR-13928 Revise canFlagOnLosing, false if need 8 votes; refactoring, renaming

btangmu:
-In VoteResolver.canFlagOnLosing the threshold is 20 (HIGH_BAR), never 8

-Revise tests to take baselineStatus into account

-Rename ...trunk... to ...baseline...

-Avoid awkward names/implementations like getAdjustedRequiredVotes

-Refactor VoteResolver to postpone all calculation of required votes until locale, path, and baseline status are set

-Reduce hard-coding of numbers 4, 8, ..., in tests

-Comments

CLDR-13928

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
@@ -106,7 +106,7 @@ public class VoteResolver<T> {
      */
     public static final int HIGH_BAR = Level.tc.votes;
 
-    private static final int LOWER_BAR = (2 * Level.vetter.votes);
+    public static final int LOWER_BAR = (2 * Level.vetter.votes);
 
     /**
      * This is the level at which a vote counts. Each level also contains the
@@ -463,7 +463,6 @@ public class VoteResolver<T> {
                 System.out.println("VoteInfo: " + info.getName() + info.getOrganization());
             }
             Organization organization = info.getOrganization();
-            //orgToVotes.get(organization).clear();
             orgToVotes.get(organization).add(value, votes, time.getTime());
             if (DEBUG) {
                 System.out.println("Adding now Info: " + organization.displayName + info.getName() + " is adding: " + votes + value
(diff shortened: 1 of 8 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1970,6 +1963,6 @@ public boolean isValueLocked() {
      * @return true or false
      */
     public boolean canFlagOnLosing() {
-        return valueIsLocked || (requiredVotes == HIGH_BAR);
+        return valueIsLocked || (getRequiredVotes() == HIGH_BAR);
```
btangmu, line 1966: This is the only change that's not refactoring or testing. If getRequiredVotes() returns 8 that's not high enough to flag.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUtilities.java
```diff
@@ -599,19 +602,20 @@ public void TestResolvedVoteCounts() {
...
+        resolver.setBaseline("foo", Status.approved);
         resolver.add("zebra", toVoterId("googleS"));
         counts = resolver.getResolvedVoteCounts();
         logln(counts.toString());
         assertEquals("", "foo", new ArrayList<>(counts.keySet()).get(0));
     }
 
     private void verifyRequiredVotes(VoteResolver<String> resolver, String locale,
-        String xpath, int required) {
+        String xpath, Status baselineStatus, int required) {
```
btangmu, line 613: This is the essential change for testing; previously the baseline status often defaulted to missing. Now we test with various status values.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -694,73 +695,65 @@ public int compare(T o1, T o2) {
...
-            return LOWER_BAR;
-        } else {
-            return requiredVotes;
+        if (requiredVotes == 0) {
+            int preliminaryRequiredVotes = supplementalDataInfo.getRequiredVotes(locale, pathHeader);
+            if (preliminaryRequiredVotes == HIGH_BAR && baselineStatus != Status.approved) {
+                requiredVotes = LOWER_BAR;
+            } else {
+                requiredVotes = preliminaryRequiredVotes;
+            }
```
btangmu, line 741: All the "adjustment" is now encapsulated here; this is refactoring, doesn't change behavior -- this is cleaner, avoids awkwardly beginning the calculation of required votes in setLocale before the VoteResolver is fully initialized

srl295 (approved):

PR #1343, opened 2021-07-01
CLDR-13853 reduce the number of english changed

macchiati:
CLDR-13853

After a few wild goose chases, tracked down the problem. The code used in VettingViewer for english changed was no longer in sync with CheckNew, so fixed that. Will have to check it out on Smoketest.

- [ ] This PR completes the ticket.

PR #1345, opened 2021-07-01
CLDR-13853 unit test for VettingViewer English Changed

srl295:
- use JUnit5 for a modern test experience
- add a VSCode settings file for test execution

CLDR-13853

depends on #1343

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "java.test.config": {
+        "env": {
+            "CLDR_DIR": "${workspaceRoot}",
+            "CLDR_ENVIRONMENT": "UNITTEST"
+        }
+    }
+}
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestVettingViewer.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestVettingViewer.java
@@ -1,0 +1,85 @@
+package org.unicode.cldr.util;
+
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import java.util.EnumSet;
+import java.util.Map.Entry;
+
+import com.ibm.icu.impl.Relation;
+import com.ibm.icu.impl.Row.R2;
+
+import org.junit.jupiter.api.Test;
+import org.unicode.cldr.test.OutdatedPaths;
+import org.unicode.cldr.util.PathHeader.PageId;
+import org.unicode.cldr.util.PathHeader.SectionId;
+import org.unicode.cldr.util.VettingViewer.Choice;
+import org.unicode.cldr.util.VettingViewer.VoteStatus;
+
+/**
+ * Also see {@link org.unicode.cldr.unittest.TestUtilities}
+ */
+class TestVettingViewer {
+    @Test
+    void testDashboardEnglishChanged() {
+        final String loc = "de";
+        final CLDRLocale locale = CLDRLocale.getInstance(loc);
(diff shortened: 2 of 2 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/util/TestVettingViewer.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Also see {@link org.unicode.cldr.unittest.TestUtilities}
+ */
+class TestVettingViewer {
+    static final String CENTURY_NARROW_DISPLAY_NAME = "//ldml/units/unitLength[@type=\"narrow\"]/unit[@type=\"duration-century\"]/displayName";
+    @Test
+    void testOutdatedPaths() {
+        String previous = VettingViewer.getOutdatedPaths().getPreviousEnglish(CENTURY_NARROW_DISPLAY_NAME);
+        assertNotEquals(OutdatedPaths.NO_VALUE, previous, "expected prev english to be present for " + CENTURY_NARROW_DISPLAY_NAME);
```
srl295, line 27: @macchiati perhaps this is to be expected, though?

macchiati: It turns out that the English value was not explicitly set until v37 (aliased value), so it is predated in some locales.

macchiati (approved):
Looks great, and we can build on the scaffolding in the future.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestVettingViewer.java
srl295: this particular test fails (for the reason you gave)  I can omit it, but keep the other test function.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "java.test.config": {
+        "env": {
+            "CLDR_DIR": "${workspaceRoot}",
+            "CLDR_ENVIRONMENT": "UNITTEST"
+        }
+    }
+}
```
btangmu, line 8: is this the first file added to the repo that's specific to VS Code? how are you using it?

I've been using VS Code for editing js files, but not for java or actually running any code

btangmu (commented):
`Error:    TestVettingViewer.testOutdatedPaths:27 expected prev english to be present for //ldml/units/unitLength[@type="narrow"]/unit[@type="duration-century"]/displayName ==> expected: not equal but was: <�>`
-- is that what you were referring to ("this particular test fails")?

srl295:
> `Error:    TestVettingViewer.testOutdatedPaths:27 expected prev english to be present for //ldml/units/unitLength[@type="narrow"]/unit[@type="duration-century"]/displayName ==> expected: not equal but was: <�>`
> 
> -- is that what you were referring to ("this particular test fails")?

Yes. It's expected , so I'll remove that test case.

.vscode/settings.json
srl295: it enables "run this test" or "debug this test" as a button right next to any test function.

macchiati:
Where does this 'button' show up?

Mark

srl295:
> Where does this 'button' show up? Mark
> […](#)

in visual studio code, next to the test class definition (for running the entire class)  and also the test function.

image

Eclipse should have similar,  "Run As -> JUnit Test Case"

srl295:
i removed the failing test case for the U+FFFD, so the test should pass now. 

The test will show as "skipped" if it didn't find any english changed to test, this won't fail but will show up in the overall logs.  Otherwise it will make sure that everything "English Changed" is present with a non-U+FFFD prior value.

macchiati:
Sounds good, thanks.
Mark

btangmu (approved):

PR #1347, opened 2021-07-01
CLDR-14866 Fix Polish missing lines

macchiati:
CLDR-14866

There was some special-case code for computing some pl and ru grammar lines, but it appears to be flaky, so I removed it.

I can't test it locally, so will have to commit to smoketest before verification.


srl295 (approved):
seems ok and simpler… will try to test locally

macchiati:
Thanks!

PR #1350, opened 2021-07-03
CLDR-14597 Prevent duplicates in cldr_interest table

btangmu:
-Fix UserRegistry.updateIntLocs

CLDR-14597

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
@@ -1066,11 +1066,20 @@ public class UserRegistry {
 
         LocaleSet intLocSet = user.getInterestLocales();
         logger.finer("uil: intlocs " + id + " = " + intLocSet.toString());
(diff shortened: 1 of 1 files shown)


srl295 (approved):
Lgtm

- did the logging not work or get in the way?
- I expected the simplification was already happening.

btangmu:
> Lgtm
> 
>     * did the logging not work or get in the way?

I didn't notice any problem with the logging

>     * I expected the simplification was already happening.

I was surprised by the duplicate rows! Not sure if the bug is old or recent; I guess re-creating the table is a pretty rare event...

btangmu:
Here's how I just "manually" removed the duplicate rows in the production db:
```
CREATE TABLE cldr_interest_temp LIKE cldr_interest;
ALTER TABLE cldr_interest_temp ADD UNIQUE INDEX temp_id (uid, forum);
INSERT IGNORE INTO cldr_interest_temp SELECT * FROM cldr_interest;
ALTER TABLE cldr_interest_temp DROP INDEX temp_id;
DROP TABLE cldr_interest;
RENAME TABLE cldr_interest_temp TO cldr_interest;
```

PR #1353, opened 2021-07-07
CLDR-14101 add examples for minimal pairs, units with grammar

macchiati:
CLDR-14101

ExampleGenerator.java
I also did a bit of cleanup, moving the fields up to the top with statics before non-statics. supplementalDataInfo could be initialized statically also.
Main changes are
- private String handleMinimalPairs(XPathParts parts, String minimalPattern) - all new
- private String getBestUnitWithGender(String gender) - all new. Not quite sure about threading with the caching pattern used
- private String handleFormatUnit(XPathParts parts, String value) - adds use of minimal pairs for context

GrammarInfo.java - just some minor reformatting

TestExampleGenerator.java
- Added tests for the two additions (minimal pairs, units with minimal pairs)

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
@@ -75,8 +75,15 @@ public class GrammarInfo implements Freezable<GrammarInfo>{
         public CharSequence getSymbol() {
             return symbol;
         }
-        public String getDefault(Collection<String> values) {
-            return this == grammaticalGender && values != null && !values.contains("neuter") ? "masculine" : defaultValue;
+        /**
+         * Gets the default value. The parameter only needs to be set for grammaticalGender
+         */
+        public String getDefault(Collection<String> featureValuesFromGrammaticalInfo) {
+            return this == grammaticalGender
+                && featureValuesFromGrammaticalInfo != null
+                    && !featureValuesFromGrammaticalInfo.contains("neuter")
+                    ? "masculine"
+                        : defaultValue;
         }
         public static Matcher pathHasFeature(String path) {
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
```diff
@@ -75,8 +75,15 @@ public String getShortName() {
...
-            return this == grammaticalGender && values != null && !values.contains("neuter") ? "masculine" : defaultValue;
+        /**
+         * Gets the default value. The parameter only needs to be set for grammaticalGender
+         */
+        public String getDefault(Collection<String> featureValuesFromGrammaticalInfo) {
+            return this == grammaticalGender
+                && featureValuesFromGrammaticalInfo != null
+                    && !featureValuesFromGrammaticalInfo.contains("neuter")
+                    ? "masculine"
+                        : defaultValue;
```
btangmu, line 86: Possibility for future improvement: define constants for "neuter", etc. If one of them were misspelled, for example, the compiler would catch the problem. Constants might facilitate updating the code when additional genders are added, etc.

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
srl295: such could be derived from the DTD +supplemental data

srl295 (approved):
lgtm on a quick check

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
macchiati: Those exist already; good point about using them.

PR #1354, opened 2021-07-07
CLDR-14870 fix items

macchiati:
Also wrote a test, which exposed other similar cases where the display name didn't match the plural, so fixed those too.

CLDR-14870


macchiati:
I did the "items" ticket, and wrote a test which revealed a bunch of related cases (where the display name didn't match the plural).

btangmu (approved):

PR #1355, opened 2021-07-07
CLDR-14858 Yukon should not show DST strings

srl295:
- add a check that CLDRFile.getRawExtraPaths() does
not show DST strings for non-DST metazones
- update getRawExtraPaths() to not include such paths
- include junit params for parameterized tests
- simplify junit pom dependencymanagement, remove an unneeded entry (bare "junit" not used anywhere)

CLDR-14858

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckMetazones.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckMetazones.java
@@ -39,20 +39,32 @@ public class CheckMetazones extends CheckCLDR {
             // MessageFormat with arguments
         }
 
-        if (path.indexOf("/long") >= 0) {
-            XPathParts parts = XPathParts.getFrozenInstance(path);
-            String metazoneName = parts.getAttributeValue(3, "type");
-            if (!metazoneUsesDST(metazoneName) && path.indexOf("/standard") < 0) {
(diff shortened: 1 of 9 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckMetazones.java
```diff
@@ -39,20 +39,32 @@ public CheckCLDR handleCheck(String path, String fullPath, String value, Options
             // MessageFormat with arguments
         }
 
-        if (path.indexOf("/long") >= 0) {
```
srl295, line 42: @macchiati is there any reason only `/long` had this restriction? Why would a short DST name for a non-DST metazone make any more sense?

srl295:
OK, so after this fix Yukon only has `standard-long` and `standard-short` in all locales, even in Comprehensive. 

Those other paths would error if you attempt to add data.  ALso, the French vs German display difference was a matter of coverage— `Yukon` is in scope for Canada!  

Also updated some junit stuff.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckMetazones.java
macchiati: Can't think of any reason.

macchiati (approved):

PR #1360, opened 2021-07-13

btangmu:
-New api/VotingCompletion.java, works with openapi/ui, not yet with ST front end

-Front-end changes in this commit are only refactoring/comments

-Remove superfluous parameter what from cldrVote.handleWiredClick

-Remove unused parameter box from cldrVote.wireUpButton

-Remove dead code

-Comments

CLDR-14877

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingCompletion.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingCompletion.java
@@ -1,0 +1,197 @@
+package org.unicode.cldr.web.api;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
+import org.eclipse.microprofile.openapi.annotations.tags.Tag;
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingCompletion.java
```diff
@@ -0,0 +1,197 @@
...
+     * again shortly thereafter. However, that 5735 milliseconds evidently involves
+     * server/locale initialization that only happens once per Survey Tool startup,
+     * and would happen anyway if this function weren't the first thing called.
+     * If the user has already loaded the Dashboard, then this function takes 358 ms
+     * the first time called, then 25 ms if called again shortly thereafter.
+     */
+    private synchronized int getTotalPathCount(String localeId, Level coverageLevel) {
+        final CLDRFile cldrFile = sm.getSTFactory().make(localeId, true, true);
+        int total = 0;
+        for (String xpath : cldrFile.fullIterable()) {
```
btangmu, line 188: create a path header factory ... pathHeader = factory.fromPath?
pathHeader.section == SUPPRESS?

btangmu: there could be votes for a suppressed item if a change has happened (though not from old votes?)

btangmu: enum is called SPECIAL
there is getShowRowAction

btangmu: PathHeader.txt
//ldml/dates/calendars/calendar[@type="%A"]/eras/eraAbbr/era[@type="%A"]   ; Special  ; Suppress ; &calendar($1) ; &calField(Eras:abbreviated:none)-$2 ; HIDE

btangmu: skip if "Suppress" or "Special" or "HIDE"
-- but there's an api for that

btangmu: Per email:

Here is the test for missing:

MissingStatus missingStatus = VettingViewer.getMissingStatus(file, path, latin);

To get at the PathHeader "filter" for the paths, use

PathHeader:

    public SurveyToolStatus getSurveyToolStatus() {

The meanings are in 

PathHeader.SurveyToolStatus

The phase can also have an effect, although I think we can probably ignore the tweaks that it makes. But we can compare to "results on the screen" and then see if we need to take it into consideration. That would be in:

CheckCLDR.Phase
CheckCLDR.StatusAction

PR #1367, opened 2021-07-15
CLDR-14408 build: timeout for JS tests

srl295:
- this test has been hanging, as a stopgap fail fast after 10m
- test should take a few seconds to run

CLDR-14408


btangmu (approved):

srl295:
working... https://github.com/unicode-org/cldr/runs/3090863953?check_suite_focus=true#step:8:48

PR #1363, opened 2021-07-14
CLDR-14803 Rename "Dizzy Face" to "Face With X Eyes"

AEApple:
Rename 😵  U+1F635 DIZZY FACE to "Face With X Eyes"

CLDR-14803

- [x] This PR completes the ticket.

PR #1369, opened 2021-07-16
CLDR-14803 improve error message for TestNamesVsEmojiData

srl295:
CLDR-14803

The source for the test says:

```java
    /** The English name should line up with the emoji-test.txt file */
```

…this change makes the test print out something to that effect.


srl295:
would have applied to #1363  also

macchiati (approved):

PR #1373, opened 2021-07-21
CLDR-14781 en: add menu variants for Sami language names

pedberg-icu:
CLDR-14781

In English, add alt=menu variants for Sami language names.
Too late in CLDR 40 to collect these for other display languages.


AEApple (approved):

srl295:
These don't seem to show up in ST… should they?

PR #1375, opened 2021-07-21
CLDR-14917 Non-ascending order of RBNF rules in "su" depends on undefined ICU behavior

grhoten:
CLDR-14917

I verified that the rules work in [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp)


richgillam (commented):
LGTM...

pedberg-icu (approved):

PR #1382, opened 2021-07-30
CLDR-14933 Fix ExampleGenerator.handleFormatUnit null pointer exception

btangmu:
-Bug fix: do not call method grammarInfo.get when grammarInfo is null

-Call getLocaleID once instead of three times

-Skip setting local variables gCase and unitCaseInfo when unused, when value is null

-Remove dead code (unitGenders, unitId, width)

-Comments

CLDR-14933

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -47,7 +47,6 @@ import org.unicode.cldr.util.SupplementalDataInfo.PluralType;
 import org.unicode.cldr.util.TimezoneFormatter;
 import org.unicode.cldr.util.TransliteratorUtilities;
 import org.unicode.cldr.util.UnitConverter;
-import org.unicode.cldr.util.UnitConverter.UnitId;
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -753,19 +752,20 @@ private String handleFormatUnit(XPathParts parts, String value) {
...
-                gCase = GrammaticalFeature.grammaticalCase.getDefault(null);
-            }
-            Collection<String> unitCaseInfo = grammarInfo.get(GrammaticalTarget.nominal, GrammaticalFeature.grammaticalCase, GrammaticalScope.units);
             if (value != null) {
+                String gCase = parts.getAttributeValue(-1, "case");
+                if (gCase == null) {
+                    gCase = GrammaticalFeature.grammaticalCase.getDefault(null);
+                }
+                Collection<String> unitCaseInfo = null;
+                if (grammarInfo != null) {
```
btangmu, line 761: This "if" is the essential bug fix, along with the unitCaseInfo != null below

srl295 (approved):

PR #1386, opened 2021-08-03
CLDR-14931 Update ansible documentation and scripts

btangmu:
-Comments in scripts and README.md

-Update oldversion, newversion, and add certbot settings in local.yml

CLDR-14931

### DIFF
--- a/tools/scripts/ansible/README.md
+++ b/tools/scripts/ansible/README.md
@@ -85,21 +85,38 @@ Add three files inside local-vars/cldrbackup-vars: id_rsa, id_rsa.pub, and known
 
 ### Setup: Config file
 
-- Create a file `local-vars/local.yml` matching the example values in [test-local-vars/local.yml](test-local-vars/local.yml) (but with a secure password instead of `hunter42`!.)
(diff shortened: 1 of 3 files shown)


tools/scripts/ansible/README.md
```diff
@@ -85,21 +85,38 @@ Add three files inside local-vars/cldrbackup-vars: id_rsa, id_rsa.pub, and known
...
+  # TODO: surveytooldeploy.password appears to be unused?
   password: hunter43
+  # vap will match CLDR_VAP in /srv/st/config/cldr.properties
   vap: hunter44
+  # testpw will match CLDR_TESTPW in /srv/st/config/cldr.properties
   testpw: hunter45
-  oldversion: 36
-  newversion: 37
+  oldversion: 39
+  newversion: 40
```
srl295, line 114: this is only used the first time the cldr.properties is created. Fine to update it, but probably doesn't need to be updated periodically.  If we do more extensive testing, we'll want to have it generated from current and (current-1)

srl295 (approved):
great

PR #1387, opened 2021-08-04
CLDR-14941 Allow script Toto to have translated name "Toto"

pedberg-icu:
CLDR-14941

Allow script Toto to have translated name "Toto" (i.e. add to exceptions list for same-as-code error).


macchiati (approved):

petross8901 (commented):
5€ bonus for winlive247@abv.net

PR #1391, opened 2021-08-04
CLDR-14952 v40 checkin common/main from SurveyTool

srl295:
CLDR-14952

ALLOW_MANY_COMMITS=true


srl295:
Splitting into multiple commits for review.

srl295:
New:

 create mode 100644 common/annotations/dsb.xml
 create mode 100644 common/annotations/hsb.xml

pedberg-icu (approved):
Everything I have looked at seems reasonable. The test failures are probably easier to fix once the data is in xml. If we can temporarily ignore the test failures to enable merging the PR, I support doing that.

srl295:
You can make a pr against my pr if you want.

srl295:
https://github.com/unicode-org/cldr/actions/runs/1099406782 I'll take a pass at fixing this sometime this morning

srl295:
- I did a re-export on top of latest master instead fo what’s in production (i.e. updated data / code ) there was no (zero) change in the output XML.  Whew.

- on a quick glance the errors seem straightforward to fix… they are all collision errors that should have been caught before.

macchiati:
I agree with Peter. If there are a small number of failures, it is simplest to override the tests, and move everything into XML. That way people can more easily work on it. Can you do that?

PR #1396, opened 2021-08-06
CLDR-14952 remove triple up arrow in annotations

srl295:
annotations: votes for triple-up-arrow are causing "must be a real value" errors. Removed from:
- kab.xml
- zu.xml

CLDR-14952


macchiati:
Yes we deliberately cause inheritance to fail because there's no sensible thing to inherit.

srl295:
> Yes we deliberately cause inheritance to fail because there's no sensible thing to inherit.

OK. Probably doesn't make sense to even emit

```xml
		<annotation cp="🤹" draft="unconfirmed">↑↑↑</annotation>
``` 

… from the ST if that's true. maybe we need an uninheritable-path filter?

srl295:
Looks like i got 'em all.

pedberg-icu (approved):

PR #1400, opened 2021-08-06
CLDR-14755 Clean up some test errors

macchiati:
CLDR-14755


pedberg-icu (approved):
Looks good, I was about to make a similar fix.

srl295:
Note for posterity: Merged during a time when tests were known to fail, however, commit message was accidentally malformed.

PR #1404, opened 2021-08-09
CLDR-14722 Fix Khmer duration patterns

AEApple:
Time durations were incorrectly replaced with localized duration units in the patterns.

CLDR-14722


AEApple:
Ideally we should have a test for this. Is there a ticket already, or should I file one?

btangmu (approved):

srl295:
> Ideally we should have a test for this. Is there a ticket already, or should I file one?

probably best to file and ref this PR

PR #1410, opened 2021-08-10
CLDR-14286 Fix Interval format GyMMM for locale fa

kipcole9:
The pattern for `<greatestDifference id="M">` has no repeating pattern
when TR35 indicates:

  The pattern is then designed to be broken up into two pieces by
  determining the first repeating field. For example, "MMM d-d, y"
  would be broken up into "MMM d-" and "d, y".

This PR replaces:

  <greatestDifference id="M">LLL تا MMM y G</greatestDifference>

with

  <greatestDifference id="M">MMM تا MMM y G</greatestDifference>

replacing LLL (standalone month) with MMM (month) thereby ensuring
there is a repeating pattern per the specification.

CLDR-14286

### DIFF
--- a/common/main/fa.xml
+++ b/common/main/fa.xml
@@ -1601,7 +1601,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						</intervalFormatItem>
 						<intervalFormatItem id="GyMMM">
 							<greatestDifference id="G">MMM y G تا MMM y G</greatestDifference>
-							<greatestDifference id="M">LLL تا MMM y G</greatestDifference>
+							<greatestDifference id="M">MMM تا MMM y G</greatestDifference>
 							<greatestDifference id="y">MMM y تا MMM y G</greatestDifference>
 						</intervalFormatItem>
 						<intervalFormatItem id="GyMMMd">


srl295:
@kipcole9 errs probably unrelated

pedberg-icu (approved):

roozbehp:
Please don't merge. This is wrong.

roozbehp (changes requested):
Since LLL and MMM values are different for some Persian month names, this would result in grammatically incorrect ranges.

There bug should be fixed in the spec or the code that parses this data. The data itself is correct.

roozbehp:
To clarify, the rule for Persian MMM/MMMM vs LLL/LLLL is that if a year immediately follows the month name, the format version is used, otherwise the standalone version is used.

kipcole9:
One approach might be to say that for the purposes of determining “repeating field”, that “standalone” and “normal” formats are equivalent. In this example, treat `L === M` when finding the break point of the interval pattern.

srl295:
@pedberg-icu i added an 'agenda' label to the ticket if it needs discussion

pedberg-icu:
> One approach might be to say that for the purposes of determining “repeating field”, that “standalone” and “format” formats are equivalent. In this example, treat `L === M` when finding the break point of the interval pattern.

@kipcole9 I like that approach. Please start a new PR for that and close this PR.

PR #1413, opened 2021-08-10
CLDR-14966 fixed various final_testing errs in be_tarask,doi,ee,ff_Adlm,gu,sv,yo_BJ

pedberg-icu:
CLDR-14966

* be_tarask 
    * delete name for 21 northern america, conflicts with 003 north america (more important)
    * delete name for 57 micronesian region, conflicts with FM micronesia (more important)
    * name for width/100 normal conflicts with weight/100 regular (which is same as in "be"); both important, so convert the "be" name for width/100 using tarask transliterator
* doi 
    * delete relative weekdays for Sunday, incomplete and do not have them for other weekdays
* ee 
    * duration/minute/short too wide, copy value from narrow
* ff_Adlm 
    * delete the indian months, only 2 supplied
* gu 
    * angle/radian/short too wide, copy value from narrow
* sv 
    * remove provisional status for narrow kilowatt-hour-per-100-kilometer/other, matches confirmed short & long
* yo_BJ 
    * light/lumen/short too wide, copy value from long (which is shorter!)

### DIFF
--- a/common/main/yo_BJ.xml
+++ b/common/main/yo_BJ.xml
@@ -3536,8 +3536,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<unitPattern count="other">{0} sídiì</unitPattern>
 			</unit>
 			<unit type="light-lumen">
-				<displayName>àmì lumɛ́ɛ̀nì</displayName>
-				<unitPattern count="other">{0} àmi lumɛ́ɛ̀nì</unitPattern>
+				<displayName>lumɛ́ɛ̀nì</displayName>
+				<unitPattern count="other">{0} lumɛ́ɛ̀nì</unitPattern>
 			</unit>
 			<unit type="mass-grain">
 				<displayName>gíréènì</displayName>
--- a/common/main/be_tarask.xml
+++ b/common/main/be_tarask.xml
@@ -511,7 +511,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<territory type="017" draft="provisional">Сярэдняя Афрыка</territory>
 			<territory type="018" draft="provisional">Паўднёвая Афрыка</territory>
 			<territory type="019" draft="provisional">Паўночная і Паўднёвая Амэрыкі</territory>
-			<territory type="021" draft="provisional">Паўночная Амэрыка</territory>
 			<territory type="029" draft="provisional">Карыбы</territory>
 			<territory type="030" draft="provisional">Усходняя Азія</territory>
 			<territory type="034" draft="provisional">Паўднёвая Азія</territory>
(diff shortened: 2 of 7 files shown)


AEApple:
Does it make sense for ee to only copy the minute value from narrow to short? It would then be inconsistent with all the other short versions?

macchiati (approved):

macchiati:
It might. Just have a phone so hard to tell. On the other hand, we don't
want to spend much time on the non target locales where people introduce
errors they don't fix.

Mark (📱)

AEApple:
narrow in ee are all single letter abbreviations, none of the short values for ee are currently abbreviated, but instead are exactly the same as the long values which is why I'm concerned about the inconsistency in ee.

pedberg-icu:
i see my "ee" fix here as a temporary measure just to work around the error issue. I am hoping that in the next round "ee" vetters will come up with a real abbreviation for minute that is between the long and single-letter narrow in length. So I do not think we should change the other short forms for consistency.

srl295:
Conflict with https://github.com/unicode-org/cldr/pull/1412 (merged)

srl295:
Your tarask fix sounds better￼

pedberg-icu:
Hmm, I did not see anyone else signed upon the spreadsheet  to fix the ff_Adlm or be_tarask errors which was why I fixed them. We probably want to avoid duplicate work.

pedberg-icu:
So how are we going to resolve the merge issue here?

srl295:
> Hmm, I did not see anyone else signed up in the spreadsheet to fix the ff_Adlm or be_tarask errors which was why I fixed them. We probably want to avoid duplicate work.

Sorry, i did not see the spreadsheet… perhaps that was Monday's discussion?

We discussed console errors and assigned them on the agenda. 

> So how are we going to resolve the merge issue here?

Can you try rebasing? Otherwise, if you want I can try updating the PR to rebase?

srl295:
@pedberg-icu I have solved the rebase locally (just be_tarask), and can update this PR if you want

srl295 (approved):
needs rebase otherwise lgtm

common/main/yo_BJ.xml
```diff
@@ -3536,8 +3536,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<unitPattern count="other">{0} sídiì</unitPattern>
 			</unit>
 			<unit type="light-lumen">
-				<displayName>àmì lumɛ́ɛ̀nì</displayName>
-				<unitPattern count="other">{0} àmi lumɛ́ɛ̀nì</unitPattern>
```
srl295, line 3540: Algorithmic, should not be in this PR, done in #1414

PR #1419, opened 2021-08-11
CLDR-14635 en-PH and en-CA to inherit en

AEApple:
Update supplemental data so that en-CA & en-PH inherit from en instead of en-001

CLDR-14635 / CLDR-14544


macchiati (approved):

srl295 (approved):

AEApple:
What is exit code 255?

Is there a ticket filed for better explanations and/or linking to the exit code descriptions, or should I file one?

AEApple:
Also why is it failing the Jira title Ticket code check? Is it only allowed to have one ticket id in the title?

srl295:
> What is exit code 255?
> 
> Is there a ticket filed for better explanations and/or linking to the exit code descriptions, or should I file one?

Just  above it it says: 

```
<< FAILURE - Error count is 5 . >>
```

and even further above it should give the `sr_Latn` errors, or any others.

I can add GitHub annotations to call these out the way I did for unit tests. I'll work on that.

AEApple:
Does that mean that exit code 255 only means that there are still 1+ errors or does it have additional meaning?

> Just above it it says:
> 
> ```
> << FAILURE - Error count is 5 . >>
> ```
> 
> and even further above it should give the `sr_Latn` errors, or any others.
> 

Thank you for adding GitHub annotations!
> I can add GitHub annotations to call these out the way I did for unit tests. I'll work on that.

srl295:
> What is exit code 255?
> 
> Is there a ticket filed for better explanations and/or linking to the exit code descriptions, or should I file one?

https://github.com/unicode-org/cldr/pull/1421

srl295:
> Does that mean that exit code 255 only means that there are still 1+ errors or does it have additional meaning?

The exit code may have a meaning for certain tests or processes, but it's probably not important compared to the information "There was at least one error". 

> 
> > Just above it it says:
> > ```
> > << FAILURE - Error count is 5 . >>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and even further above it should give the `sr_Latn` errors, or any others.
> 
> Thank you for adding GitHub annotations!
> 
> > I can add GitHub annotations to call these out the way I did for unit tests. I'll work on that.

srl295:
errs are just sr_Latn

naotoj:
Hi,
Sorry if it is already discussed, but should this fix include the currency symbol for `USD` as "US$" into `en_CA.xml`? Otherwise, a mere "$" represents the USD, which was not the case for `en-CA` locale.

stenshamn:
The currency symbol should indeed be "US$" for en_CA (as viewed here in ST before the change https://st.unicode.org/cldr-apps/v#/en_CA/C_NAmerica/196bd456f629e908) and this may very well have gotten borked by the inheritance change... Probably for en_PH as well.

macchiati:
Thanks. We discussed that in Monday's meeting, and plan a fix.
Mark

naotoj:
Thanks. I appreciate it.

reedesler:
Hi, I'm wondering if the CA and PH regions should be added to the [enUS group](https://github.com/unicode-org/cldr/blob/c686f24abae6654a2147288d51c0a34c5d4c57b3/common/supplemental/languageInfo.xml#L13) for language matching. I would expect a user with the desired language `en-CA` to be matched with `en-US` over `en-GB`.

srl295:
> Hi, I'm wondering if the CA and PH regions should be added to the [enUS group](https://github.com/unicode-org/cldr/blob/c686f24abae6654a2147288d51c0a34c5d4c57b3/common/supplemental/languageInfo.xml#L13) for language matching. I would expect a user with the desired language `en-CA` to be matched with `en-US` over `en-GB`.

As this is long merged, probably best to comment on the open bug CLDR-14635

PR #1420, opened 2021-08-11
CLDR-14966 copy some ↑↑↑ entries for units from sr to sr_Latn

pedberg-icu:
CLDR-14966

The process for generating locales like sr_Latn from sr evidently does not copy over the source locale entries with ↑↑↑, but some of them are needed in  order for the target locale to pass logical group tests. So in this case I just copied them manually.


srl295 (approved):
LGTM! Do we need a followup ticket to fix the generator? Maybe it should copy only where there's a logical group?

srl295:
with this in sr_Latn:

```
<< SUCCESS - No errors found. >>
```

PR #1378, opened 2021-07-22
CLDR-14748 Use GitHub Action Commands to improve test output clarity

srl295:
CLDR-14748

- [ ] This PR completes the ticket.

PR #1421, opened 2021-08-11
CLDR-14748 GitHub annotation for ConsoleCheck

srl295:
CLDR-14748

When in a GitHub build, annotate consolecheck output so that it shows errors prominently. c.f. #1378

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -108,7 +108,7 @@ jobs:
           name: cldr-code
           path: tools/cldr-code/target/
       - name: run CLDR console check
-        run: java -DCLDR_DIR=$(pwd) -Xmx6g -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -e -z FINAL_TESTING
+        run: java -DCLDR_GITHUB_ANNOTATIONS=true -DCLDR_DIR=$(pwd) -Xmx6g -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -e -z FINAL_TESTING
   deploy:
     # don't run deploy on manual builds!
     if: github.repository == 'unicode-org/cldr' && github.event_name == 'push' && github.ref == 'refs/heads/master' && github.event.inputs.git-ref == ''
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
@@ -94,6 +94,7 @@ public class ConsoleCheckCLDR {
     static boolean SHOW_LOCALE = true;
     static boolean SHOW_EXAMPLES = false;
     // static PrettyPath prettyPathMaker = new PrettyPath();
+    private static boolean CLDR_GITHUB_ANNOTATIONS = (Boolean.parseBoolean(System.getProperty("CLDR_GITHUB_ANNOTATIONS", "false")));
 
     private static final int HELP1 = 0,
         HELP2 = 1,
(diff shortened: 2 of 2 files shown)


.github/workflows/maven.yml
```diff
@@ -66,23 +66,24 @@ jobs:
...
-          sleep 3
-          mysql --user=root --password=root < tools/cldr-apps/test-setup.sql
-      - name: Test with maven
-        run: >
-          mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
-          '-Dorg.unicode.cldr.unittest.web.jdbcurl=jdbc:mysql://cldrtest:VbrB3LFCr6A!@localhost/cldrtest?useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC'
-          '-Dorg.unicode.cldr.unittest.web.KeepDb=true' -DCLDR_GITHUB_ANNOTATIONS=true
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        echo "TODO FIXME!!!"
```
srl295, line 69: TEMPORARY!!

.github/workflows/maven.yml
```diff
@@ -108,7 +109,7 @@ jobs:
           name: cldr-code
           path: tools/cldr-code/target/
       - name: run CLDR console check
-        run: java -DCLDR_DIR=$(pwd) -Xmx6g -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -e -z FINAL_TESTING
+        run: java -DCLDR_GITHUB_ANNOTATIONS=true -DCLDR_DIR=$(pwd) -Xmx6g -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -e -z FINAL_TESTING -f sr_Latn
```
srl295, line 112: THis is a test run…

srl295:
@AEApple Have a look at this run: https://github.com/unicode-org/cldr/actions/runs/1121190444

I'm going to add the locale ID and clean it up, so it will be:

> sr_Latn: invalid place holder @ Units|Volume|fluid-ounce-imperial|narrow-one-nominative

But you can click on an error and it will pop into the logs with the full context.

srl295:
image

image

AEApple:
That does look nice, but I'm still going to see the exit code 255 at the end and then ask what I need to do to fix it.

There is still not an explanation of what exit code 255 means in terms or resolving errors so a PR can merge.

srl295:
> That does look nice, but I'm still going to see the exit code 255 at the end and then ask what I need to do to fix it.
> There is still not an explanation of what exit code 255 means in terms or resolving errors so a PR can merge.

Exit code 255 has always been a result of errors printed on prior lines. Previously, we had to search for the word "error" and scroll up to see what the problem was.  This fix, and the one in #1378, attempt to make those prior errors stand out.

There's nothing technically that prevents some internal error in a build tool from failing, resulting in the same "process completed…" message but potentially with no further information about what happened.

srl295:
@AEApple I've taken out the testing harness for this (since our build actually passes now).  It might be possible to add more text in the build, or perhaps in documentation- I guess my point is that 'process exitted' is GitHub's message that just means something bad happened, I don't think we can customize it easily.

AEApple:
> @AEApple I've taken out the testing harness for this (since our build actually passes now). It might be possible to add more text in the build, or perhaps in documentation- I guess my point is that 'process exitted' is GitHub's message that just means something bad happened, I don't think we can customize it easily.

Documentation would work for me. I'll think about this after the site migration.

macchiati (approved):

srl295:
Build https://github.com/unicode-org/cldr/actions/runs/1121386570 should demonstrate this also, on real errs

PR #1424, opened 2021-08-11
CLDR-14783 update ICU4J libs to version of 2021-08-11 with Unicode 14 beta

pedberg-icu:
CLDR-14783

Update to use libs from latest ICU4J, tagged "cldr/2021-08-11", which adds Unicode 14 support. That requires likelySubtags & test/tool support for an additional ancient script Cpmn "Cypro-Minoan" (already have en name for it).


markusicu (approved):
lgtm

PR #1432, opened 2021-08-13
CLDR-14423 Use year marker " 'р'." consistently in uk

AEApple:
CLDR-14423 - Use the year marker 'р'." consistently in Ukrainian

Please check spreadsheet attached to issue for additional feedback from the linguist

### DIFF
--- a/common/main/uk.xml
+++ b/common/main/uk.xml
@@ -1521,7 +1521,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					</dateFormatLength>
 					<dateFormatLength type="medium">
 						<dateFormat>
-							<pattern>d MMM y G</pattern>
+							<pattern>d MMM y 'р'. G</pattern>
 							<datetimeSkeleton>GyMMMd</datetimeSkeleton>
 						</dateFormat>
 					</dateFormatLength>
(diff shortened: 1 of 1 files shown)


AEApple:
@pedberg-icu - Can you please take a look and let me know if I missed any items I should update

- I didn't add the proposed year marker for Gy, which the language specialist suggested please let me know if that does align with your expectation it wasn't clear in the issue.
- The language specialist flagged a couple of LLL forms that they believe should be MMM instead, can you let me know what you think? Happy to push that part of the feedback to 41 in another ticket if needed.
- In the interval formats I went with "'рр'." at the end if nothing followed the y but "'р'." if G followed the year for example, but let me know if your feedback agrees with our language specialist that "'р'." should always follow the y instead of using "'рр'." when the year is part of the interval on both sides.
- Our language specialist also suggested that "'р'." should be in the interval formats with MM as well as MMM. I didn't make that change since it does not align with your reported feedback. I will probably include in the clone ticket for 41 unless you think it should be implemented now.

srl295:
@AEApple I don't see a spreadsheet?

srl295 (commented):
data seems OK but i don't have any background to compare it to.

AEApple:
> @AEApple I don't see a spreadsheet?

It's included in the Jira issue, I actually exported as a PDF, let me know if it doesn't work for you and I can reattach. (Actual spreadsheet is linked from the TC notes.)

srl295:
> > @AEApple I don't see a spreadsheet?
> 
> It's included in the Jira issue, I actually exported as a PDF, let me know if it doesn't work for you and I can reattach. (Actual spreadsheet is linked from the TC notes.)

Attached to https://unicode-org.atlassian.net/browse/CLDR-14423 ? I don't see anything attached there.

AEApple:
Weird my update didn't publish, so I just published the saved draft.

srl295 (approved):
spot check lgtm

common/main/uk.xml
```diff
@@ -1592,9 +1592,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="yyyyMd">dd.MM.y G</dateFormatItem>
 						<dateFormatItem id="yyyyMEd">E, dd.MM.y G</dateFormatItem>
 						<dateFormatItem id="yyyyMMM">LLL y G</dateFormatItem>
```
stenshamn, line 1594: The ticket says "Input from our sources suggests that 'р' should be used for all formats with MMM/LLL or longer month styles." so then we should be changing this line as well, no?

stenshamn (approved):

PR #1433, opened 2021-08-13

btangmu:
-New api/VotingCompletion.java, works with openapi/ui, not yet with ST front end

-Revise DataSection and VettingViewer to be more consistent with each other

-Skip paths for which PathHeader returns null

-Refactor PathHeader and related code for encapsulation and consistency

-In some places assume both HIDE and DEPRECATED paths should be hidden

-In getShowRowAction and getAcceptNewItemAction, change ph.status to ph

-Rename cleanPath/pretty to ph for consistency

-Front-end changes in this commit are only refactoring/comments

-Remove superfluous parameter what from cldrVote.handleWiredClick

-Remove unused parameter box from cldrVote.wireUpButton

-Reduce warnings for deprecated/unused/dead

-Change public to private where possible

-Remove dead/commented-out code

-Comments, including TODO

CLDR-14877

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrGui.js
+++ b/tools/cldr-apps/js/src/esm/cldrGui.js
@@ -612,7 +612,9 @@ function refreshCounterVetting() {
     $("#nav-page .counter-infos").hide();
     return;
   }
-
+  // TODO: avoid using the DOM for calculations -- calculate total, voted, and abstain using
+  // the original data, not the elements that we've put in the DOM
+  // Reference: https://unicode-org.atlassian.net/browse/CLDR-14877
   const inputs = $(".vetting-page input:visible:checked");
   let total = inputs.length;
   const abstain = inputs.filter(function () {
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataSection.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataSection.java
@@ -90,7 +90,8 @@ public class DataSection implements JSONString {
      */
 
     /**
-     * The baseline CLDRFile for this DataSection
+     * The translation-hints CLDRFile for this DataSection
+     * This is English if TRANS_HINT_ID = "en_ZZ"; see SurveyMain.getTranslationHintsFile
      */
     private CLDRFile translationHintsFile;
 
@@ -1222,7 +1223,7 @@ public class DataSection implements JSONString {
         public StatusAction getStatusAction() {
             // null because this is for display.
             return SurveyMain.phase().getCPhase()
-                .getShowRowAction(this, InputMethod.DIRECT, getPathHeader().getSurveyToolStatus(), userForVotelist);
+                .getShowRowAction(this, InputMethod.DIRECT, getPathHeader(), userForVotelist);
         }
 
         /**
@@ -1610,7 +1611,7 @@ public class DataSection implements JSONString {
 
         DataSection section = new DataSection(pageId, sm, locale, prefix, matcher);
 
-        CLDRFile ourSrc = sm.getSTFactory().make(locale.getBaseName(), true, true);
+        CLDRFile ourSrc = sm.getSTFactory().make(locale.getBaseName());
 
         ourSrc.setSupplementalDirectory(sm.getSupplementalDirectory());
 
@@ -1850,12 +1851,8 @@ public class DataSection implements JSONString {
                     // synthesize a new row..
                     String base_xpath_string = podBase + ourSuffix + suff;
 
-                    SurveyToolStatus ststats = SurveyToolStatus.READ_WRITE;
                     PathHeader ph = stf.getPathHeader(base_xpath_string);
-                    if (ph != null) {
-                        ststats = stf.getPathHeader(base_xpath_string).getSurveyToolStatus();
-                    }
-                    if ((ststats == SurveyToolStatus.HIDE || ststats == SurveyToolStatus.DEPRECATED)) {
+                    if (ph == null || ph.shouldHide()) {
                         continue;
                     }
 
@@ -2046,12 +2043,8 @@ public class DataSection implements JSONString {
                 }
             }
 
-            SurveyToolStatus ststats = SurveyToolStatus.READ_WRITE;
             PathHeader ph = stf.getPathHeader(xpath);
-            if (ph != null) {
-                ststats = stf.getPathHeader(xpath).getSurveyToolStatus();
-            }
-            if (ststats == SurveyToolStatus.HIDE || ststats == SurveyToolStatus.DEPRECATED) {
+            if (ph == null || ph.shouldHide()) {
                 continue;
             }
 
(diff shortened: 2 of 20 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1631,8 +1631,7 @@ public int compare(Map<String, Object> o1, Map<String, Object> o2) {
                     continue; // ignore unvotes.
                 }
                 PathHeader pathHeader = (PathHeader) m.get("pathHeader");
-                if (pathHeader.getSurveyToolStatus() != PathHeader.SurveyToolStatus.READ_WRITE &&
-                    pathHeader.getSurveyToolStatus() != PathHeader.SurveyToolStatus.LTR_ALWAYS) {
+                if (pathHeader.shouldSkipForVoting()) {
```
btangmu, line 1634: This maybe should be `if (!pathHeader.canReadOrWrite())` instead -- currently the same result?  Study this further...

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataSection.java
```diff
@@ -1850,12 +1851,8 @@ private void ensureComplete(CLDRFile ourSrc, TestResultBundle checkCldr) {
                     // synthesize a new row..
                     String base_xpath_string = podBase + ourSuffix + suff;
 
-                    SurveyToolStatus ststats = SurveyToolStatus.READ_WRITE;
                     PathHeader ph = stf.getPathHeader(base_xpath_string);
-                    if (ph != null) {
-                        ststats = stf.getPathHeader(base_xpath_string).getSurveyToolStatus();
-                    }
-                    if ((ststats == SurveyToolStatus.HIDE || ststats == SurveyToolStatus.DEPRECATED)) {
+                    if (ph == null || ph.shouldHide()) {
```
btangmu, line 1855: This change assumes that getPathHeader never returns null for a valid path that shouldn't be hidden.

Is my assumption correct?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -1754,6 +1754,10 @@ protected CLDRFile handleMake(String localeID, boolean resolved, DraftStatus mad
         return get(localeID).getFile(resolved);
     }
 
+    public CLDRFile make(String loc) {
```
btangmu, line 1757: This new single-parameter make() is for both DataSection and Dashboard. It's equivalent to the current code, but better since we want DataSection and Dashboard to be consistent, and the current code arbitrarily uses equivalent 2-parameter and 3-parameter versions of make().

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
btangmu: I'm changing shouldSkipForVoting to !canReadOrWrite; no need for two functions.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
```diff
@@ -196,12 +204,14 @@ public StatusAction getShowRowAction(
 
             // if limited submission, and winner doesn't have an error, limit the values
 
-            if (LIMITED_SUBMISSION && !SubmissionLocales.allowEvenIfLimited(pathValueInfo.getLocale().toString(), pathValueInfo.getXpath(), valueStatus == ValueStatus.ERROR, pathValueInfo.getBaselineStatus() == Status.missing)) {
-                return StatusAction.FORBID_READONLY;
+            if (LIMITED_SUBMISSION) {
+                if (!SubmissionLocales.allowEvenIfLimited(pathValueInfo.getLocale().toString(), pathValueInfo.getXpath(), valueStatus == ValueStatus.ERROR, pathValueInfo.getBaselineStatus() == Status.missing)) {
+                    return StatusAction.FORBID_READONLY;
+                }
```
btangmu, line 210: This change simply prevents Eclipse from warning about dead code when LIMITED_SUBMISSION is false

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -543,8 +546,10 @@ public int compareTo(PathHeader other) {
             if (0 != (result = alphabeticCompare(code, other.code))) {
                 return result;
             }
-            if (!SKIP_ORIGINAL_PATH && 0 != (result = alphabeticCompare(originalPath, other.originalPath))) {
-                return result;
+            if (!SKIP_ORIGINAL_PATH) {
+                if (0 != (result = alphabeticCompare(originalPath, other.originalPath))) {
+                    return result;
+                }
```
btangmu, line 552: This change simply prevents Eclipse warning about dead code

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -53,6 +53,13 @@
 
     private static final boolean DEBUG = false;
 
+    /**
+     * DASHBOARD_USES_BASELINE: the "baseline" info is not used for the Dashboard;
+     * it is possibly still used for Priority Items Summary.
+     * TODO: remove code for baseline where it isn't needed. Reference: https://unicode-org.atlassian.net/browse/CLDR-14877
+     */
+    public static final boolean DASHBOARD_USES_BASELINE = false;
```
btangmu, line 61: I've added this constant so that we don't waste time getting baseline value for Dashboard when we don't need it, but we can easily restore the code for getting baseline value if we decide to make Dashboard use baseline again. However, unless we actually have plans for Dashboard to use baseline, I suggest that we simplify the code by removing this constant and simplify the code that depends on it.

btangmu: Never mind -- even though the Dashboard front end doesn't use baseline (a.k.a. "old"), the back end does need it while generating Dashboard data

tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -612,7 +612,9 @@ function refreshCounterVetting() {
     $("#nav-page .counter-infos").hide();
     return;
   }
-
+  // TODO: avoid using the DOM for calculations -- calculate total, voted, and abstain using
+  // the original data, not the elements that we've put in the DOM
+  // Reference: https://unicode-org.atlassian.net/browse/CLDR-14877
```
btangmu, line 617: This and all changes to front end are just comments and refactoring

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingCompletion.java
```diff
@@ -0,0 +1,223 @@
...
+import org.unicode.cldr.util.PathHeader.Factory;
+import org.unicode.cldr.util.SupplementalDataInfo;
+import org.unicode.cldr.web.CookieSession;
+import org.unicode.cldr.web.DBUtils;
+import org.unicode.cldr.web.SurveyMain;
+import org.unicode.cldr.web.api.VotingParticipation.VotingParticipationResults;
+
+@Path("/completion")
+@Tag(name = "completion", description = "APIs for voting completion statistics")
+public class VotingCompletion {
```
btangmu, line 36: This is the new back end API. It can be tested at openapi/ui/ though it isn't used by the Survey Tool front end yet.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -407,7 +407,6 @@ public Status getErrorStatus(String path, String value, StringBuilder statusMess
     private final SupplementalDataInfo supplementalDataInfo;
     private final String baselineTitle = "Baseline";
     private final String currentWinningTitle;
-    private Factory baselineFactory;
```
btangmu, line 410: This was unused; the getter and setter served no purpose

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -549,33 +524,23 @@ public void addAll(FileInfo other) {
...
-         * @param xpath
-         * @return
+         * @param specificSinglePath if not null, skip all paths except this one
+         * @return the FileInfo
          */
         private FileInfo getFileInfo(CLDRFile sourceFile, CLDRFile baselineFile,
             Relation<R2<SectionId, PageId>, WritingInfo> sorted,
             EnumSet<Choice> choices, String localeID,
-            T user, Level usersLevel, boolean quick, String xpath) {
+            T user, Level usersLevel, String specificSinglePath) {
```
btangmu, line 543: "quick" was unused (always false); "xpath" and "path" were too easily confused with each other, so "xpath" is descriptively renamed "specificSinglePath"

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestVettingViewer.java
```diff
@@ -52,19 +51,19 @@ protected final PathHeader getPathHeader(String xpath) {
...
+
+        final Factory sourceFactory = CLDRConfig.getInstance().getCldrFactory();
+
+        Factory baselineFactory = sourceFactory;
 
         EnumSet<VettingViewer.Choice> choiceSet = EnumSet.of(VettingViewer.Choice.englishChanged);
         CLDRFile sourceFile = sourceFactory.make(loc, true);
         CLDRFile baselineFile = baselineFactory.make(loc, true);
-        Relation<R2<SectionId, PageId>, VettingViewer<Organization>.WritingInfo> file;
+        Relation<R2<SectionId, PageId>, VettingViewer<Organization>.WritingInfo> sorted;
```
btangmu, line 62: This was named "sorted" in some places and "file" in other places. Changed to use the name "sorted" consistently. Neither name actually means anything to me. The type `Relation<R2<SectionId, PageId>, VettingViewer<Organization>.WritingInfo>` occurs repeatedly. I would need to spend a lot more time to have any hope of understanding it. Such complexity cries out for encapsulation!

macchiati:
FYI
Relation is an old class that predates Multimap; ideally we'd replace it.
Row is a class that has a list of typed elements; R2 is the instantiation
that has 2 members. There isn't a good modern replacement.

macchiati:
Steven, could you review this? Most of it is material I'm not familiar with.

srl295 (commented):
> Steven, could you review this? Most of it is material I'm not familiar with.

yes, will review

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestVettingViewer.java
srl295: I can never remember how it works until i step through it :joy:

srl295 (approved):

srl295:
Lots of good work here thanks

srl295:
@btangmu no blocker comments

PR #1436, opened 2021-08-16
CLDR-14976 CLDR 40 a0, fix en_CA symbol for USD, make tarask uppercase

pedberg-icu:
CLDR-14976

Two more fixes for CLDR alpha0 integration to ICU:
1. en_CA needs to specify USD symbol as "US$" which it used to inherit from en_001
2. For be_tarask, the variant needs to be uppercase

### DIFF
--- a/common/main/be_TARASK.xml
+++ b/common/main/be_TARASK.xml
@@ -11,7 +11,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 	<identity>
 		<version number="$Revision$"/>
 		<language type="be"/>
-		<variant type="tarask"/>
+		<variant type="TARASK"/>
 	</identity>
 	<localeDisplayNames>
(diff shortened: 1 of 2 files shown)


common/main/be_TARASK.xml
```diff
@@ -11,7 +11,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 	<identity>
 		<version number="$Revision$"/>
 		<language type="be"/>
-		<variant type="tarask"/>
+		<variant type="TARASK"/>
```
srl295, line 14: should be a test for this

srl295 (approved):

common/main/be_TARASK.xml
pedberg-icu: Filed as https://unicode-org.atlassian.net/browse/CLDR-14978

PR #1349, opened 2021-07-03
CLDR-14573 Add full data of Kaingang and Nheengatu locales

robertgmelo:
CLDR-14573 - Add full data of Kaingang and Nheengatu locales
- Kaingang (kgp) data (common and keyboard)
- Nheengatu (yrl) data (common and keyboard)
- Update supplemental data

- [x] This PR completes the ticket.

PR #1441, opened 2021-08-17
CLDR-14573 CLDRModify no-arg for kgp/yrl

srl295:
Part of CLDR-14980

CLDR-14573

- [ ] This PR completes the ticket

#1349 just barely missed #1438 so this catches the last two locales


srl295:
@robertgmelo these are just mechanical changes we always do

macchiati (approved):

PR #1444, opened 2021-08-17
CLDR-14811 fix LDML2Json for reading bcp47 dtd, add test

srl295:
CLDR-14811

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
@@ -376,12 +376,7 @@ public class Ldml2JsonConverter {
                 activeNumberingSystems.add(ns);
             }
         }
-        DtdType fileDtdType;
-        if (CLDRFile.isSupplementalName(locID)) {
-            fileDtdType = DtdType.supplementalData;
-        } else {
-            fileDtdType = DtdType.ldml;
-        }
+        final DtdType fileDtdType = file.getDtdType();
         CoverageInfo covInfo = CLDRConfig.getInstance().getCoverageInfo();
         for (Iterator<String> it = file.iterator("", DtdData.getInstance(fileDtdType).getDtdComparator(null)); it.hasNext();) {
             int cv = Level.UNDETERMINED.getLevel();
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRFile.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRFile.java
@@ -8,6 +8,9 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.nio.file.Path;
+import java.util.Comparator;
+import java.util.Iterator;
 import java.util.Set;
 
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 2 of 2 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRFile.java
```diff
@@ -64,4 +66,26 @@ public void TestDraftStatus(DraftStatus status) {
...
+    public void TestReadBCP47() {
+        Path bcp47Path = CLDRConfig.getInstance().getCldrBaseDirectory().toPath().resolve("common/bcp47");
+        Factory factory = Factory.make(bcp47Path.toString(), ".*");
+        assertNotNull(factory);
+        for (final String id : factory.getAvailable()) {
+            System.out.println(id);
+            CLDRFile file = factory.make(id, false);
+            assertNotNull(file, id);
+            for (final String xpath : file.fullIterable()) {
+                assertNotNull(xpath, id + ":" + xpath);
```
macchiati, line 80: This will assert that id + ":" + xpath is not null. That will always be the case. Same below. I think the arguments are reversed.

macchiati (dismissed):

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRFile.java
srl295: This test uses the standard `org.junit.jupiter.api.Assertions.assertNotNull` which has the args in the other order.

srl295: anyway, the iterator itself was failing previously. Working with @nordzilla to repro.

macchiati:
BTW, the bcp47 data has been read by supplementalInfo for some time.

Mark (📱)

srl295:
> BTW, the bcp47 data has been read by supplementalInfo for some time. Mark (📱)

yes, but not by CLDRFile which is used for the JSON machinery.

tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
```diff
@@ -376,12 +376,7 @@ private String transformPath(final String pathStr, final String pathPrefix) {
                 activeNumberingSystems.add(ns);
             }
         }
-        DtdType fileDtdType;
-        if (CLDRFile.isSupplementalName(locID)) {
-            fileDtdType = DtdType.supplementalData;
-        } else {
-            fileDtdType = DtdType.ldml;
-        }
```
srl295, line 384: This was the root issue, an incomplete if-tree. Better to use CLDRFile.getDtdType().

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRFile.java
```diff
@@ -64,4 +67,75 @@ public void TestDraftStatus(DraftStatus status) {
...
+        assertNotNull(file, id);
+        for (final String xpath : file.fullIterable()) {
+            assertNotNull(xpath, subdir + ":" + id + " xpath");
+            /*final String value = */ file.getStringValue(xpath);
+        }
+
+        for (Iterator<String> i = file.iterator(); i.hasNext();) {
+            final String xpath = i.next();
+            assertNotNull(xpath, subdir + ":" + id + " xpath");
+            /*final String value = */ file.getStringValue(xpath);
```
nordzilla, line 132: Would it be relevant to assert that the results of `file.getStringValue()` is both non-null and non-empty?

nordzilla (approved):
This looks good to me.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRFile.java
srl295: no, it could be null for missing values.

srl295:
@sffc  @macchiati can I get a stamp on this one? @nordzilla likes it…

sffc (approved):

PR #1445, opened 2021-08-17
CLDR-14984 update en_001 from en_GB

macchiati:
CLDR-14984

- Added tool to make it easier to compare locales.
- Copied values from en_GB to en_001 that were different than en_001
- Deleted some values from en_001 that were the same in en and en_GB

Annotations showed a lot of changes, but spot checks look correct.
Main were mostly small changes: units, adding 1 availableFormat, 24 hour time, and Sept.
(I think some of these changes were left over from the last release.)


AEApple (approved):
Looks good from my spot check

PR #1446, opened 2021-08-17
CLDR-14607 Adding metaZoneSupplemental

sffc:
CLDR-14607

This XML structure would serve our needs.  See the ticket and the email to cldr-core2 for more information.


macchiati:
To fit in with the other features, the data should be in https://github.com/unicode-org/cldr/blob/master/common/bcp47/timezone.xml, with that structure. 

eg name = bcp47 code, alias=metazone

We need comments in the file that the 4-letter forms are used for metazones, with one legacy exception:
            <type name="gaza" description="Gaza Strip, Palestinian Territories" alias="Asia/Gaza"/>

We will also need a spec change to describe this. Needn't be long.

macchiati:
Note: there will likely be some work necessary on the tests as well, since the current assumption is that all items are timezones not metazones.

srl295:
> To fit in with the other features, the data should be in https://github.com/unicode-org/cldr/blob/master/common/bcp47/timezone.xml, with that structure.
> 
> eg name = bcp47 code, alias=metazone
> 
> We need comments in the file that the 4-letter forms are used for metazones, with one legacy exception:
> 
> We will also need a spec change to describe this. Needn't be long.

I propose:

```xml
            <type name="auce" description="Metazone: Australia_Central" alias="Australia_Central"/>
```

We could have a cldrmodify type task that automatically sets the descriptions to the generic name from English, such as `description="Metazone: Central Australia Time"` or just `description="Central Australia Time"`.  A simplistic `description` should be fine for now.

macchiati:
Yes simple description done for now.

Mark (📱)

srl295:
@sffc ping?

macchiati:
Put revised version into https://github.com/unicode-org/cldr/pull/1457

sffc:
Archived in my fork as `archive/CLDR-14607-metazone`

PR #1447, opened 2021-08-17
CLDR-14635 change the en_CA data for the new inheritance

macchiati:
That is, with the new inheritance, the resolved file is the same as the resolved file with the old inheritance.

That is, the revised file has items that were in the old en_001 + en_CA but not in en.

That means people won't see a difference in the results that they saw in the ST. Note that in the process, the up arrows are removed. That won't make any effective difference: for the release they are removed anyway, and during the next cycle people's votes for inheritance should be restored (since those will be the winning values).

CLDR-14635

### DIFF
--- a/common/annotations/en_CA.xml
+++ b/common/annotations/en_CA.xml
@@ -12,3825 +12,401 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<territory type="CA"/>
 	</identity>
 	<annotations>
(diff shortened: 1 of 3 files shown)


AEApple:
So can you explain why all the inheritances are removed? Is that just because they'll automatically inherit from root either way and we want the vetters to vote again in the next cycle?

AEApple (approved):
Looks good to me from my spot check.

AEApple (approved):

PR #1448, opened 2021-08-17
CLDR-14986 Deal with lack of case data for certain locales

macchiati:
CLDR-14986

Includes

1. deletion of day-duration data from most locales. The vast majority of this across all locales was bogus.
2. fixing minimal pair data in a few cases
3. removing suspect case data from certain locales. (In those locales, the vast majority of the data was ↑↑↑)
4. making spot fixes to Hindi based on Karan's feedback (and verified by Google vetter)
5. adding corrected case data to Romanian and Serbian
6. narrowing the grammarFeatures.xml data to the units in evidence
7. modifying the CompareLocales tool to aid in comparison

### DIFF
--- a/common/main/bn.xml
+++ b/common/main/bn.xml
@@ -7638,10 +7638,10 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<ordinalMinimalPairs ordinal="one">ডান দিকে {0}ম বাঁকটি নিন।</ordinalMinimalPairs>
 			<ordinalMinimalPairs ordinal="other">ডান দিকে {0}তম বাঁকটি নিন।</ordinalMinimalPairs>
 			<ordinalMinimalPairs ordinal="two">ডান দিকে {0}য় বাঁকটি নিন।</ordinalMinimalPairs>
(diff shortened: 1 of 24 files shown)


AEApple (commented):
I am not familiar enough with the java classes that are being modified to be able to review that portion quickly, but I wanted to comment that the data looks correct to me from what I check against the data provide by the vetters in case someone else wants to review the Java portions.

Is there a way to mark specific files as reviewed and approved in GitHub?

macchiati:
Thanks, that's great. 
You can mark files as viewed, but I'm not sure how that shows up to other reviewers. This comment should be enough to let Peter know what you've done and only the tool remains.

The code file changes are only to a tool used in analyzing the data so those aren't going to disturb anything. So I'd feel comfortable with your signing off on it, or waiting for Peter. Either way.

AEApple:
One other question: Does this have to merge to generate the sr_Latn for the updated sr values?

pedberg-icu (approved):

macchiati:
Good thought. However in this case I made parallel deletions to both so it
doesn't matter whether it's emerged before or after the generation.

Mark (📱)

PR #1449, opened 2021-08-18
CLDR-14976 de: compact decimals, revert formats for 1000..100000 to use 0, not Tsd.

pedberg-icu:
CLDR-14976

Once again, for German compact decimals, revert formats for 1000..100000 to use just 0 (as in https://github.com/unicode-org/cldr/pull/151 and 2 or 3 others), not formats with Tsd.

### DIFF
--- a/common/main/de.xml
+++ b/common/main/de.xml
@@ -5875,12 +5875,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			</decimalFormatLength>
 			<decimalFormatLength type="short">
 				<decimalFormat>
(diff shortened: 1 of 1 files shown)


macchiati (approved):

AEApple (approved):

markusicu:
I recommend adding a test for this -- in the Survey Tool code, or as a unit test that would catch this at least when going back to XML.

AEApple:
> I recommend adding a test for this -- in the Survey Tool code, or as a unit test that would catch this at least when going back to XML.

Filed CLDR-14988 per TC discussion so that this won't happen.

PR #1452, opened 2021-08-19
CLDR-14976 en_001, revert add of HH time formats

pedberg-icu:
CLDR-14976

In https://github.com/unicode-org/cldr/pull/1445, as part of CLDR-14984 update en_001 from en_GB, the time formats with HH were added to en_001, overriding the time formats with h inherited from en. This breaks nearly all child locales and should *not* be done as part of the en_001 updates, as discussed in previous CLDR releases. This PR reverts that part of the en_001 updates.

### DIFF
--- a/common/main/en_001.xml
+++ b/common/main/en_001.xml
@@ -340,28 +340,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						</dateFormat>
 					</dateFormatLength>
 				</dateFormats>
-				<timeFormats>
-					<timeFormatLength type="full">
-						<timeFormat>
-							<pattern>HH:mm:ss zzzz</pattern>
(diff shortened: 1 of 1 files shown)


markusicu:
Once more, the best way to prevent this sort of thing would be a unit test.

AEApple (approved):
Can we update the BRS to state that this causes breakage? I missed the note about this issue.

AEApple:
> Once more, the best way to prevent this sort of thing would be a unit test.

Is there an open ticket for that already? If not, I will file one.

macchiati:
That'd be great, thanks!
Mark

AEApple:
> That'd be great, thanks! Mark

Filed: [CLDR-14996](https://unicode-org.atlassian.net/browse/CLDR-14996)

PR #1457, opened 2021-08-23
CLDR-14607 Revised version

macchiati:
CLDR-14607

Puts the change into timezone.xml; deprecates 'gaza'


common/bcp47/timezone.xml
```diff
@@ -176,7 +176,8 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="fotho" description="Faroe Islands" alias="Atlantic/Faeroe Atlantic/Faroe"/>
             <type name="frpar" description="Paris, France" alias="Europe/Paris"/>
             <type name="galbv" description="Libreville, Gabon" alias="Africa/Libreville"/>
-            <type name="gaza" description="Gaza Strip, Palestinian Territories" alias="Asia/Gaza"/>
+            <type name="gaza" description="Gaza Strip, Palestinian Territories" deprecated="true" alias="Asia/Gaza gazastrp"/>
+            <type name="gazastrp" description="Gaza Strip, Palestinian Territories" deprecated="true" alias="Asia/Gaza"/>
```
sffc, line 180:
```suggestion
            <type name="gazastrp" description="Gaza Strip, Palestinian Territories" alias="Asia/Gaza"/>
```

common/bcp47/timezone.xml
```diff
@@ -476,6 +477,164 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="zajnb" description="Johannesburg, South Africa" alias="Africa/Johannesburg"/>
             <type name="zmlun" description="Lusaka, Zambia" alias="Africa/Lusaka"/>
             <type name="zwhre" description="Harare, Zimbabwe" alias="Africa/Harare"/>
+            <type name="acre" description="Metazone" alias="Acre" since="40"/>
```
sffc, line 480: Optional: Should we add a new XML attribute, like `type="metazone"` ?

common/bcp47/timezone.xml
macchiati: Whoops! Yes, I meant only the 'gaza' one.

common/bcp47/timezone.xml
macchiati: I thought of that, and can bring it up in Wednesday's meeting. On the other hand, since the metazones will be all and only the 4 letter codes (other than the deprecated gaza) we wouldn't need to.

btangmu (approved):

common/bcp47/timezone.xml
srl295: would it make sense to promote gaza to a metazone to make the invariant work?

common/bcp47/timezone.xml
srl295: I think the length is sufficient.

macchiati:
No, that would break compatibility.

Mark

srl295:
- Merged into ICU in https://github.com/unicode-org/icu/pull/1843

PR #1460, opened 2021-08-24
CLDR-14984 A23.52 Update the segments

macchiati:
using unicodetools/.../Segmente……r.java

CLDR-14984


btangmu (approved):

srl295 (approved):

markusicu:
Sorry for not doing this one myself. I will make a note for next time.

Sigh: We have so many versions of these rules: The spec itself, two files in Unicode Tools, this here in CLDR, six line break versions in ICU for runtime use, and I think fourteen (2 code, 6 C++ testdata, 6 Java test data) for “monkey tests”. And each group with somewhat different syntax. :-(

PR #1461, opened 2021-08-24
CLDR-14801 v40 BRS §A40: DTD for Unicode 14.0.0

srl295:
CLDR-14801 

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -20,7 +20,7 @@ Except as contained in this notice, the name of a copyright holder shall not be 
 <!ATTLIST version cldrVersion CDATA #FIXED "40" >
     <!--@MATCH:version-->
     <!--@VALUE-->
-<!ATTLIST version unicodeVersion CDATA #FIXED "13.0.0" >
+<!ATTLIST version unicodeVersion CDATA #FIXED "14.0.0" >
     <!--@MATCH:version-->
     <!--@VALUE-->
 


markusicu:
Interesting, I wasn't aware of this unicodeVersion. Why does CLDR have it?

The Unicode update comes in stages: script metadata, likely subtags, collation root data, Chinese transliteration, ICU with Unicode.next properties, further updates towards final data. Of course, all should feed into the same CLDR release. When should this unicodeVersion be updated? Does it matter?

Should there be a unit test that checks that this Unicode version number is the same as some other, e.g., the one from ICU?

srl295:
> Interesting, I wasn't aware of this unicodeVersion. Why does CLDR have it?

The spec says:

>> _The `unicodeVersion` attribute defines the version of the Unicode standard that is used to interpret data. Specifically, some data elements such as exemplar characters are expressed in terms of UnicodeSets. Since UnicodeSets can be expressed in terms of Unicode properties, their meaning depend on the Unicode version from which property values are derived._

So I think the point is for consumers to check their own Unicode Version to make sure that `[:Arab:]` or whatever has the correct meaning.

> The Unicode update comes in stages: script metadata, likely subtags, collation root data, Chinese transliteration, ICU with Unicode.next properties, further updates towards final data. Of course, all should feed into the same CLDR release.

Right… 

> When should this unicodeVersion be updated? Does it matter?

Per process, it happens now :) 

> Should there be a unit test that checks that this Unicode version number is the same as some other, e.g., the one from ICU?

Imagine that. Yes, there probably should be a test, I'll file an issue for this. => https://unicode-org.atlassian.net/browse/CLDR-15004

srl295:
@markusicu please approve the PR if i got it right :)

markusicu (approved):
lgtm

macchiati (approved):

PR #1466, opened 2021-08-24
CLDR-14984 A19 Emoji tickets P2

macchiati:
CLDR-14984

Add 'bathroom' and 'toilet' as suggested in review.


AEApple (approved):
Thank you!

PR #1469, opened 2021-08-25

btangmu:
-Remove superfluous path exclusion checks for altProposed, /references

-Refactor and format DataSection.ensureComplete to clarify it only applies to timeZoneNames paths

-Use ph.shouldHide instead of SurveyToolStatus.DEPRECATED/HIDE where the distinction is unneeded

-Simplify code that had impossible conditions in TestPathHeader

-Fix deprecation warnings for SurveyLog, setCldrFileToCheck

-Comments, clean-up

CLDR-14877

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
@@ -169,7 +169,13 @@ abstract public class CheckCLDR {
             ) {
 
             PathHeader.SurveyToolStatus status = ph.getSurveyToolStatus();
-            // always forbid deprecated items - don't show.
+            /*
+             * Always forbid DEPRECATED items - don't show.
+             *
+             * Currently, bulk submission and TC voting are allowed even for SurveyToolStatus.HIDE,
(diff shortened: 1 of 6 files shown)


macchiati (approved):
@srl295  should look at the DataSection part.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
```diff
@@ -169,7 +169,13 @@ public StatusAction getShowRowAction(
             ) {
 
             PathHeader.SurveyToolStatus status = ph.getSurveyToolStatus();
-            // always forbid deprecated items - don't show.
+            /*
+             * Always forbid DEPRECATED items - don't show.
+             *
+             * Currently, bulk submission and TC voting are allowed even for SurveyToolStatus.HIDE,
+             * but not for SurveyToolStatus.DEPRECATED. If we ever want to treat HIDE and DEPRECATED
+             * the same here, then it would be simpler to call ph.shouldHide which is true for both.
```
srl295, line 177: I think it's an oversight that bulk submission allows `SurveyToolStatus.HIDE` and that `FORBID_READONLY` should be returned for such items. What do you think @macchiati ? Could be a separate item / bug because it's not part of progress bars.

srl295 (approved):

PR #1481, opened 2021-09-01
CLDR-14607 deprecated gaza should NOT alias to Asia/Gaza but SHOULD have a preferred mapping to gazastrp

pedberg-icu:
CLDR-14607

We had
```
    <type name="gaza" description="Gaza Strip, Palestinian Territories" deprecated="true" alias="Asia/Gaza gazastrp"/>
    <type name="gazastrp" description="Gaza Strip, Palestinian Territories" alias="Asia/Gaza"/>
```
but the first entry had two problems: the deprecated "gaza" short ID should not have the alias mapping to Asia/Gaza (that conflicts with the correct mapping from Asia/Gaza to gazastrp), and its reference to "gazastrp" should be via a preferred attribute, not an alias.


macchiati (approved):

srl295 (commented):
postmerge lgtm

sffc (commented):
LGTM

srl295:
- Correction to #1457 
- Merged into ICU in https://github.com/unicode-org/icu/pull/1843

PR #1484, opened 2021-09-02
CLDR-15024 get correct coverage information

macchiati:
CLDR-15024

Note that the old comment says specifically "does not skip inheritance marker", but made a function call that **did** skip the inheritance marker.

Note that we would still have to change our process for generating charts, as in the ticket.


srl295 (approved):

srl295:
@macchiati LGTM but test failures, summarized here: https://github.com/unicode-org/cldr/actions/runs/1192443901

srl295:
@macchiati actually, click "Files Changed" as if to review your own PR… scroll down… you'll see the test failures in context.

macchiati:
Thanks! Made a fix, and we'll see how the tests do.

Mark

macchiati:
Can you review so I can merge?

srl295 (approved):

PR #1485, opened 2021-09-03
CLDR-14607 move metazoneAliases under supplemental

srl295:
CLDR-14607


srl295:
WIP: needs dtd fix

macchiati:
Just sent out a message about this to core

Mark

srl295:
yeah, crossed in the mail…

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -417,6 +417,7 @@
 //supplementalData/metaZones/metazoneInfo/timezone[@type="%A"]/usesMetazone/_mzone			; Supplemental ; Metazone ; $1 ; -∞..∞ ; HIDE
 
 //supplementalData/metaZones/mapTimezones[@type="%A"]/mapZone[@other="%A"][@territory="%A"]/_type			; Supplemental ; Metazone ; $1 ; $2-$3 ; HIDE
+//supplementalData/metaZones/shortIds/shortId[@name="%A"]_alias    ; Supplemental ; Metazone ; $1 ; $2-$3 ; HIDE
```
srl295, line 420: @macchiati What is the test expecting here?

>     TestCompletenessNonLdmlDtd {
>      Error: (TestPathHeader.java:1318) //supplementalData/metaZones/shortIds/shortId[@name="%A"]/_deprecated: Unknown path header[], Supplemental      Unknown metaZones       shortIds/shortId[@name="acre"]/_deprecated
>        No match in PathHeader.txt for //supplementalData/metaZones/shortIds/shortId[@name="acre"]/_deprecated
>        You get only one message for all paths matching //supplementalData/metaZones/shortIds/shortId[@name="%A"]/_deprecated
>        For example, check to see if the field in PathHeader.txt is in PathHeader.PageId.
>        If not, either correct PathHeader.txt or add it to PageId
>        If you have a value attribute, you will need extra _ characters. The value attribute will show at the end with prefixed _, eg [...]/_skeleton.If there can be a value for the path then that element will add _.

srl295:
I think it's ready to go , modulo some test fixes

srl295:
@yumaoka comments:
> - "description" attribute does not make sense here. All entries in the proposed data use "Metazone". But this is under <metaZones><metazoneIds>
> - We may define since="40" as default value.

common/supplemental/metaZones.xml
```diff
@@ -1892,6 +1892,167 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<mapZone other="Yerevan" territory="001" type="Asia/Yerevan"/>
 			<mapZone other="Yukon" territory="001" type="America/Whitehorse"/>
 		</mapTimezones>
+
+		<shortIds>
+            <shortId name="acre" description="Metazone" alias="Acre" since="40"/>
```
yumaoka, line 1897: description="Metazone" is redundant. This data is now under <metaZones><shortId>, so it's metazone ID obviously. We could add meaningful description later, for example "Time used by Acre state in Brazil".

yumaoka: Because inception of this data is CLDR 40, we could default to since="40" and use since="xx" only for future additions.

srl295: The display name of the metazone should come from CLDR.

Do we want this supplemental item to be "meritable facts about this metazone"? or is it strictly "short aliases"? If it's the former, we should have a different structure and the element should be `<metazoneInfo>` or such. 

We don't currently have any chunk of data that exists one time for each metazone.

yumaoka: I actually prefer to have a complete list of metazones - with long/short IDs.
At this moment, if you want to get all valid metazone IDs, you have to walk through the mapping data and collect all of unique IDs.

srl295: @yumaoka we can do that, but that's a different type of structure. right now `metazoneInfo` AND `mapTimezones` each have mappings to metazones. I'd assume we don't want to actually change those existing items?

If we want an element that is the canonical list of metazones, I'd name it so:


```xml
<metazoneIds>
  <metazoneId type="Europe_Western" alias="euwe" since="41"/> <!-- assuming a non-default 'since' -->
</metazoneIds>
```

This structure and naming would allow for other metadata on a per-metazone basis.

yumaoka: Maybe something like below?

```
<metaZones>
  <metazoneList>
    <metazone type="Acre" short="acre" description="Time for Acre, Brazil"/>
    ....
  </metazoneList>
  <metazoneInfo>
    <!-- tz/mz mapping data -->
    ...
  </metazoneInfo>
</metaZoens>
```

If we did not have <metaZones> element, we would want to use `<metazones>` instead of `<metazoneList>` above. The element <metazone> describes the definition of each metazone - comes with long id and short id.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
macchiati: Every kind of path needs a match in Pathheader.txt. Easiest way to do this is to look for a path like the one you have.

EG //supplementalData/metaZones/mapTimezones[@type="%A"]/mapZone[@other="%A"][@territory="%A"]/_type			; Supplemental ; Metazone ; $1 ; $2-$3 ; HIDE

WARNING: the longest matches have to come first in the file, and you need to have the various combinations of #IMPLIED attributes that appear. That's why you see:

//supplementalData/metaZones/metazoneInfo/timezone[@type="%A"]/usesMetazone[@from="%A"][@to="%A"]/_mzone			; Supplemental ; Metazone ; $1 ; $2-$3 ; HIDE
//supplementalData/metaZones/metazoneInfo/timezone[@type="%A"]/usesMetazone[@from="%A"]/_mzone			; Supplemental ; Metazone ; $1 ; $2..∞ ; HIDE
//supplementalData/metaZones/metazoneInfo/timezone[@type="%A"]/usesMetazone[@to="%A"]/_mzone			; Supplemental ; Metazone ; $1 ; -∞..$2 ; HIDE
//supplementalData/metaZones/metazoneInfo/timezone[@type="%A"]/usesMetazone/_mzone			; Supplemental ; Metazone ; $1 ; -∞..∞ ; HIDE

For the attributes for shortId, only preferred may or might not occur. So two patterns should suffice.

macchiati:
That is a reasonable structure (wish this discussion had happened some time
ago though). It would correspond to the following.

(minor changes: we use the plural instead of list usually, and then the
singular, so metazoneIds/metazoneId. And it is best to avoid the generic
'type' for clarity)

<!ELEMENT metaZones ( metazoneIds?, mapTimezones?, shortIds? ) >

<!ELEMENT *metazoneIds* ( *metazoneId** ) >

<!ELEMENT metazoneId EMPTY >
<!ATTLIST metazoneId *shortId* NMTOKEN #REQUIRED >
    ***@***.***:regex/[a-z][a-z][a-z][a-z]-->
<!ATTLIST metazoneId *longId* CDATA #IMPLIED >
    ***@***.***:any-->
    ***@***.***>
<!ATTLIST metazoneId description CDATA #REQUIRED >
    ***@***.***:any-->
    ***@***.***>
<!ATTLIST metazoneId deprecated (true | false) "false" >
    ***@***.***>
<!ATTLIST metazoneId preferred NMTOKEN #IMPLIED >
    ***@***.***:bcp47/anyvalue-->
    ***@***.***>
<!ATTLIST metazoneId since CDATA "40" >
    ***@***.***:version-->
    ***@***.***>


Mark

srl295:
@macchiati Your reply got a little scrambled.

macchiati:
BTW, about PathHeader.txt. 

https://cldr.unicode.org/development/updating-dtds has an overview, and we can flesh out as needed.

srl295:
Here's what i'm proposing for the dtd.

- dropped 'description', is there a compelling reason to have a description? We have the localized text in locale data, and we have plenty of metadata about what makes up a metazone. Let me know if there's a use case here.
- changed longId and preferred to `@MATCH:metazone`

```xml
<!ELEMENT metaZones ( metazoneInfo?, mapTimezones?, metazoneIds? ) >

<!ELEMENT metazoneIds ( metazoneId* ) >

<!ELEMENT metazoneId EMPTY >
<!ATTLIST metazoneId shortId NMTOKEN #REQUIRED >
    <!--@MATCH:regex/[a-z][a-z][a-z][a-z]-->
<!ATTLIST metazoneId longId CDATA #IMPLIED >
    <!--@MATCH:metazone-->
    <!--@VALUE-->
<!ATTLIST metazoneId deprecated (true | false) "false" >
    <!--@VALUE-->
<!ATTLIST metazoneId preferred NMTOKEN #IMPLIED >
    <!--@MATCH:metazone-->
    <!--@VALUE-->
<!ATTLIST metazoneId since CDATA "40" >
    <!--@MATCH:version-->
    <!--@METADATA-->
```

srl295:
@macchiati @sffc @yumaoka DTD/XML updated, PTAL

common/supplemental/metaZones.xml
```diff
@@ -1892,6 +1892,167 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<mapZone other="Yerevan" territory="001" type="Asia/Yerevan"/>
 			<mapZone other="Yukon" territory="001" type="America/Whitehorse"/>
 		</mapTimezones>
+
+		<metazoneIds>
+            <metazoneId shortId="acre" longId="Acre"/>
```
sffc, line 1897: Nit: Inconsistent space/tab indentation

common/supplemental/metaZones.xml
```diff
@@ -1892,6 +1892,167 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<mapZone other="Yerevan" territory="001" type="Asia/Yerevan"/>
 			<mapZone other="Yukon" territory="001" type="America/Whitehorse"/>
 		</mapTimezones>
+
+		<metazoneIds>
```
sffc, line 1896: Optional: I was thinking originally that this section could contain additional metadata about metazones, not just the ID mapping.  In that case, `<metazoneInfo>` or `<metazoneDetails>` might be an appropriate tag name.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
macchiati: You have $2-$3, bu there is only one variable in the regex (%A), so only a $1. 

Best to look at the bcp47 equivalents for the pattern.

macchiati: The test results show what the paths turn into. The value parameters turn into separate items
//supplementalData/metaZones/metazoneIds/metazoneId[@shortId="%A"]/_deprecated
//supplementalData/metaZones/metazoneIds/metazoneId[@shortId="%A"]/_longId

macchiati:
We have used the convention on XXXs/XXX in a number of places.

If we had to do it all over again, I would have liked to have changed the
convention to drop the intermediate levels. But given what we have, let's
leave as Steven has it.




Mark

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
macchiati: So you need to provide values for these. The values for all three can be done with the following, I think.

//supplementalData/metaZones/shortIds/shortId[@name="%A"]_%E    ; Supplemental ; Metazone ; $1 ; $2 ; HIDE

macchiati (changes requested):
Left comment for fixing the PathHeader failures

macchiati:
Whoops, grabbed the wrong one:
//supplementalData/metaZones/metazoneIds/metazoneId[@shortId="%A"]/_%E ; Supplemental ; Metazone ; $1 ; $2 ; HIDE

common/supplemental/metaZones.xml
srl295: thanks, fixed

srl295:
@macchiati @sffc PTAL, should pass tests and be ready to go

macchiati (approved):

macchiati:
Just reviewed, looks good!

Mark

common/supplemental/metaZones.xml
sffc: Thoughts on my optional comment?

```xml
<metazoneDetails>
    <metazoneDetail ... />
    <metazoneDetail ... />
</metazoneDetails>
```

sffc (approved):

macchiati:
I'd responded 8 hours ago about that.

Mark

sffc:
> I'd responded 8 hours ago about that. Mark

I assume you're taking about this reply?

> We have used the convention on XXXs/XXX in a number of places.
>
> If we had to do it all over again, I would have liked to have changed the
> convention to drop the intermediate levels. But given what we have, let's
> leave as Steven has it.

My optional suggestion is to replace "metazoneIds" with "metazoneDetails", in case we want to add more details than just the IDs in the future.  It is consistent with the "XXXs/XXX" naming convention.

srl295:
@sffc  @macchiati  I'd lke to merge and we can iterate.

PR #1486, opened 2021-09-06
CLDR-14989 GitHub Action to produce CLDR Production Data

srl295:
CLDR-14989

This is a first step, just producing production data on request.  Manually launched. Your choice of CLDR ref to build. outputs a .zip file with production data.

See an example run with .zip attached at:

https://github.com/srl295/cldr/actions/runs/1206942094

image
image

I will be adding the ability to build with ICU, but that may take some more debugging. This workflow will be useful as is for producing production data.  Additionally, the cldr-code .jar file is attached to the run.

### DIFF
--- a/.github/workflows/build-icu.yml
+++ b/.github/workflows/build-icu.yml
@@ -1,0 +1,72 @@
+# Build ICU
+
+name: Build ICU from CLDR
+
+# Triggered on push to master,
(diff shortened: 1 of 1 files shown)


srl295:
I have building ICu data in progress too

macchiati (approved):

PR #1488, opened 2021-09-07
CLDR-14989 GitHub Action to build ICU with CLDR

srl295:
CLDR-14989

### DIFF
--- a/.github/workflows/build-icu.yml
+++ b/.github/workflows/build-icu.yml
@@ -28,7 +28,7 @@ jobs:
     name: Build Production Data
     runs-on: ubuntu-latest
     steps:
-      - name: Clone Repository (Custom Ref)
+      - name: Clone CLDR
         uses: actions/checkout@v2
         with:
           ref: ${{ github.event.inputs.cldr-ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-icu.yml
```diff
@@ -63,10 +63,82 @@ jobs:
           mvn  -s .github/workflows/mvn-settings.xml -B  -DCLDR_DIR=$(pwd) -DCLDR_GITHUB_ANNOTATIONS=true --file=tools/pom.xml -pl cldr-code  \
             exec:java -Dexec.mainClass=org.unicode.cldr.tool.GenerateProductionData \
             -Dexec.args="-d target/cldr-prod"
+          (cd target/cldr-prod/ && zip -r ../cldr-prod.zip *)
```
srl295, line 66: seems to be more performant to zip up first

srl295:
https://github.com/srl295/cldr/actions/runs/1211247730

srl295:
Generates these artifacts:
- cldr-code:  the main CLDR jar
- cldr-prod: Production CLDR xml data (aka cldr-staging/production)
- icu-data: a .zip file roughly like the icu-data.zip, with built ICU data
- icu4c:  distribution of ICU4C built w/ clang on ubuntu
- icu4j: distribution of ICU4J built w/ OpenJDK16

The latter two targets were copied from icu's

macchiati:
Thanks!

It isn't clear from https://github.com/srl295/cldr/actions/runs/1211247730
where the artifacts are stored for download. That is, although the CLDR
data and ICU data re built, I don't see where to fetch them from.

Mark

srl295:
> make: *** No rule to make target 'out/build/icudt70l/uprops.icu', needed by 'out/icu4j/icudata.jar'.  Stop.

Any ideas @yumaoka ?

srl295:
> Thanks!
> 
> It isn't clear from https://github.com/srl295/cldr/actions/runs/1211247730
> where the artifacts are stored for download. That is, although the CLDR
> data and ICU data re built, I don't see where to fetch them from.
> 

They are attached to the action run .. may not show up since this is a fork. 

Try the action in CLDR.

srl295:
Ready to merge.

srl295:
It's at the bottom of https://github.com/srl295/cldr/actions/runs/1211247730 as shown below. If it doesn't show it may be because it's in my own fork and won't give access.

image

.github/workflows/build-icu.yml
```diff
@@ -59,14 +59,194 @@ jobs:
...
+        run: >
+          env
+          TOOLS_ROOT=$(pwd)/icu/tools/
+          CLDR_DIR=$(pwd)/cldr
+          CLDR_DATA_DIR=$(pwd)/cldr-prod/
+          ant -f icu/tools/cldr/cldr-to-icu/build-icu-data.xml clean all && ln icu/icu4c/LICENSE icu/icu4c/source/data/
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Zip up icu-data.zip
+        run: 'cd icu/icu4c/source && zip -rlq ../../../icu-data.zip data'
```
jefgen, line 155: I was wondering why this created a double zip file (icu-data.zip inside icu-data.zip)... but it looks like this is a limitation of the GitHub Action:

https://github.com/actions/upload-artifact
>  One of the consequences of this limitation is that if a zip is uploaded during a workflow run and then downloaded from the UI, there will be a **double zip created**.

Oh well...  ¯\\_ (ツ)_/¯

.github/workflows/build-icu.yml
```diff
@@ -59,14 +59,194 @@ jobs:
...
+        # Do this after restoring cache.
+        run: >
+          mvn -B install:install-file   -Dproject.parent.relativePath=""   -DgroupId=org.unicode.cldr
+          -DartifactId=cldr-api   -Dversion=0.1-SNAPSHOT   -Dpackaging=jar   -DgeneratePom=true
+          -Dfile=cldr-code/cldr-code.jar
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Generate ICU Data
+        run: >
+          env
```
jefgen, line 147: In the [cldr-icu-readme.txt](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/cldr-icu-readme.txt#L125) file I noticed that it sets other environment vars, like `ANT_OPTS="-Xmx4096m"` and `ICU4C_DIR`/`ICU4J_ROOT`.

I guess they might not actually be needed?

(FWIW, I think the `ICU4C_DIR` will be auto-set [here](https://github.com/unicode-org/icu/blob/7851f708a3dd43fc1e0e9dd784e7dfe9a70915b2/icu4c/source/data/build.xml#L25), but I'm not sure about `ICU4J_ROOT`)

.github/workflows/build-icu.yml
```diff
@@ -59,14 +59,194 @@ jobs:
...
+          name: icu-data
+          path: icu-data.zip
+  # clang release build with some options to enforce useful constraints.
+  # Includes dependency checker on an in-source, optimized build.
+  # Includes checking @draft etc. API tags vs. ifndef guards like
+  # U_HIDE_DRAFT_API and U_FORCE_HIDE_DRAFT_API.
+  # (FORCE guards make this tool pass but won't compile to working code.
+  # See the testtagsguards.sh script for details.)
+  icu4c-clang:
+    runs-on: ubuntu-latest
```
jefgen, line 168: Wonder if maybe giving it a name, perhaps something like (?):
`    name: Build and test ICU4C, with new data`

jefgen (commented):
This is really neat @srl295! 😊

.github/workflows/build-icu.yml
srl295: @jefgen it seemed to be quicker to upload a zip rather than to upload a 'directory'.

jefgen: Ah, good point. 👍

.github/workflows/build-icu.yml
srl295: Didn't seem to be needed.

srl295: `ICU4J_ROOT` is for the icu4j data install step (C Makefile) and is set on line 230, but is not needed here on line 147 - this is cldr-to-icu (David's tool)

jefgen: Ah, sorry for missing that. Thanks!

srl295:
New run: https://github.com/srl295/cldr/actions/runs/1222657915

jefgen (approved):

PR #1494, opened 2021-09-10
CLDR-14642 json: bcp47 filenames and values

srl295:
- add a "-t all" that converts all types in one operation
- use a modern Timer for progress
- update progress messages
- modify some paths to use valid bcp47 values
- produce bcp47 locale ids for files, such as und, ca-ES-valencia, en-US-u-va-posix
- add a new tool option --bcp47=false which produces the old behavior (for compatibility)
- correct JSON_config_bcp47.txt (update to CLDR-14571)

CLDR-14642

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
@@ -5,12 +5,14 @@ import java.io.IOException;
 import java.io.PrintWriter;
 import java.text.ParseException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
+import java.util.logging.Logger;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -28,6 +30,7 @@ import org.unicode.cldr.tool.Option.Options;
 import org.unicode.cldr.util.Annotations;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.CLDRFile.DraftStatus;
 import org.unicode.cldr.util.CLDRPaths;
 import org.unicode.cldr.util.CLDRTool;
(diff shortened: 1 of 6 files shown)


srl295:
Snapshot of the data at https://github.com/unicode-org/cldr-json/pull/28

srl295:
Summary / migration:
- `root` is now `und`.  So, `und` subdirectory instead of root.
- parentLocales: `        "sr-Latn": "und",`
- `ca-ES-valencia` and `be-tarask` locales use their proper bcp47 names. Note that the identity section retains the original CLDR values for language-territory-variant
```json
{
  "main": {
    "ca-ES-valencia": {
      "identity": {
        "version": {
          "_cldrVersion": "40"
        },
        "language": "ca",
        "territory": "ES",
        "variant": "VALENCIA"
      },
```
- similarly, `en_US_POSIX` is now `en-US-u-va-posix`, for example:
```json
{
  "main": {
    "en-US-u-va-posix": {
      "identity": {
        "version": {
          "_cldrVersion": "40"
        },
        "language": "en",
        "territory": "US",
        "variant": "POSIX"
      },
      "delimiters": {
```
- In this change I attempted to convert all relevant-looking values into bcp47. I may not have exhausted all values, especially in supplemental.
- Note that the keys in `variants.json` are still in uppercase, so

```json
      "localeDisplayNames": {
        "variants": {
          "TARASK": "orthographe taraskievica",
```

macchiati:
Note that you should be able to check the Dtd @MATCH annotations for other
cases.

Mark

srl295:
> Note that you should be able to check the Dtd @MATCH annotations for other
> cases.
> 

Good point. I could derive the changes automatically from it. Think there's actually a bug for exactly that !

btangmu (approved):

sffc (approved):
I thought the plan was to remove en-US-u-va-posix from JSON, according to the discussions in https://unicode-org.atlassian.net/browse/CLDR-14487.  But if you leave it there, I guess it's fine.

srl295:
> I thought the plan was to remove en-US-u-va-posix from JSON, according to the discussions in https://unicode-org.atlassian.net/browse/CLDR-14487. But if you leave it there, I guess it's fine.

Hm, I do see that.  I thought the plan was to use all bcp47 IDs.

Could downstream (icu4x, etc) just omit undesired locales?  That seems less disruptive than withdrawing a locale that's otherwise in common.

macchiati:
My recollection was the same as Shane's, to not include the locale in JSON.

Mark

srl295:
@macchiati @sffc  I'll update cldr-json to just exclude this locale.

PR #1501, opened 2021-09-16
CLDR-14989 GitHub Action to build ICU with CLDR (#1488)

srl295:
CLDR-14989

cherrypick from #1488 to avoid a split ticket


srl295:
same story as #1502

pedberg-icu (approved):

PR #1506, opened 2021-09-24
CLDR-14642 WIP

srl295:
CLDR-14642


srl295:
verified that bcp47 ids would work for timeData.json (testing data, just to show structure)

```json
"timeData": {
      "UY": {
        "_allowed": "H h hB hb",
        "_preferred": "H"
      },
      "be-tarask": {
        "_allowed": "H h K",
        "_preferred": "H"
      },
      "ca-ES-valencia": {
        "_allowed": "H h K",
        "_preferred": "H"
      },
      "en-US-u-va-posix": {
        "_allowed": "H h K",
        "_preferred": "H"
      }
  }
```

tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
```diff
@@ -389,18 +389,39 @@ private String transformPath(final String pathStr, final String pathPrefix) {
...
-                result = fixXpathBcp47(result, "likelySubtag", "from", "to");
+                if (!result.contains("\"iw\"") &&
+                    !result.contains("\"in\"") &&
+                    !result.contains("\"ji\"")) {
+                    // Special case: preserve 'iw' and 'in' likely subtags
+                    result = fixXpathBcp47(result, "likelySubtag", "from", "to");
+                } else {
+                    result = underscoreToHypen(result);
+                    logger.warning("Including aliased likelySubtags: " + result);
+                }
```
srl295, line 404: @yumaoka @macchiati  some special code to preserve `iw`, `in`, `ji` in likelySubtags - otherwise, they end up duplicating `he`, `id`, `yi` due to aliasing.  As an alternative, these likelysubtags could be dropped.

macchiati:
They can't be dropped!

By design.

Mark (📱)

srl295:
OK, then probably just special casing these three may make the most sense. ￼￼

srl295:
I would expect it to be less likely that a BCP 47 user would be starting with these deprecated codes in the first place￼

srl295:
> Looks ok .
> 
> Note: we do have to preserve all likely mappongs as they are.

yes, the likely mappings are preserved.

PR #1509, opened 2021-09-29
CLDR-14684 missing DTD line in spec

macchiati:
CLDR-14684


srl295 (approved):
bug and PR title are a little… cryptic!

PR #1512, opened 2021-10-01
CLDR-14945 Abstain should not cause disappearance of inherited item

btangmu:
-In voteForValue, call setValueFromResolver with resolveMorePaths true so vote resolution is not bypassed

-In setValueFromResolver, fix handling of Status.missing: no premature return, must update delegate

-Change setValueFromResolver from public to private since only called locally

-Comments

CLDR-14945


btangmu:
I've changed this PR to "draft" since there's a failing test.

btangmu:
See https://github.com/unicode-org/cldr/pull/1513

PR #1513, opened 2021-10-01
CLDR-14945 Abstain should not cause disappearance of inherited item

btangmu:
-In voteForValue, call setValueFromResolver with resolveMorePaths true so vote resolution is not bypassed

-In setValueFromResolver, fix handling of Status.missing: no premature return, must update delegate

-Change setValueFromResolver from public to private since only called locally

-Change TestSTFactory.TestSparseVote to expect inheritance, not null

-Rename some variables in TestSTFactory.TestSparseVote

-Comments

CLDR-14945

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -204,7 +204,7 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
          *
          * TODO: reevaluate whether resolveMorePaths is needed anymore.
          */
-        public VoteResolver<String> setValueFromResolver(String path, VoteResolver<String> resolver, boolean resolveMorePaths) {
+        private VoteResolver<String> setValueFromResolver(String path, VoteResolver<String> resolver, boolean resolveMorePaths) {
             PerLocaleData.PerXPathData xpd = ballotBox.peekXpathData(path);
             String res;
             String fullPath = null;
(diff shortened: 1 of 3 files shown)


srl295 (approved):
makes sense..

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -1172,7 +1174,11 @@ public synchronized void voteForValue(User user, String distinguishingXpath, Str
                 doPermanentVote(distinguishingXpath, xpathId, value);
             }
 
-            xmlsource.setValueFromResolver(distinguishingXpath, null, false /* resolveMorePaths */);
```
macchiati, line 1175: This is an opportunity for changing the boolean into something meaningful, 

add in XmlSource
enum PathResolution {MORE_PATHS, FEWER_PATHS}

true => PathResolution.MORE_PATHS

But probable give an more descriptive enum value name than MORE; that is what is included that isn't in LESS.

srl295: yes, and, do all callers now pass true ?If so, then just drop the parameter

btangmu: loadVoteValues still calls setValueFromResolver with resolveMorePaths true for vxml generation, and false otherwise. Before we could change that, we would need to study and measure the effects, including possible performance costs. I think that's beyond the scope of this ticket, which is for fixing a well-defined bug.

For loadVoteValues, this comment describes resolveMorePaths:

         * @param resolveMorePaths false to do vote resolution only on paths with votes in current votes table, or
         *                         true to do vote resolution also on all paths in trunk (for making vxml).

Should we make another ticket with the objective of cleaning up these parts of the code? I hesitate to create another techdebt ticket when we already have so many.

macchiati:
> Should we make another ticket with the objective of cleaning up these
parts of the code?

I think there is already a broader ticket to make the code more readable
(eg replacing naked constants (including booleans) with meaningful values
(like enums)). So we could do the fixup under that.

The refactoring is pretty simple:

Add an enum.
Add method that creates from boolean

Change the API to use it
Add a copy of the old API vectoring to the new, by using the
create-from-boolean method
Refactor: Inline that copy.
Use "References" to find the call sites for create-from-boolean method, and
fix.

There might be a faster way, if either of you have suggestions. Using
refactoring helps to avoid typos creeping in.

Mark

btangmu:
Per Mark and Steven's comments, I'm going to do a little refactoring and commenting and make another PR on this ticket. I'll postpone deeper changes that would require testing vxml production, performance, etc.

PR #1515, opened 2021-10-03
CLDR-8386 fix Malayalam transliterator

mjansche:
Fix transliteration of Malayalam VOWEL SIGN O, OO, AU.
Enable transliteration of Malayalam chillu letters.
As a result, the transliteration of ജോർദ്ദാൻ into Latin
changes from "jēāർddāൻ" to "jōrddān".

Associated changes:
- Map Dravidian RRA to Bengali RA instead of RA NUKTA.

CLDR-8386


common/transforms/Malayalam-InterIndic.xml
```diff
@@ -87,6 +106,15 @@ For terms of use, see http://www.unicode.org/copyright.html
 ൭→\uE06D;      # DIGIT SEVEN
 ൮→\uE06E;      # DIGIT EIGHT
 ൯→\uE06F;      # DIGIT NINE
+൰→\uE067\uE066;            # UNMAPPED Malayalam-InterIndic: NUMBER TEN
+൱→\uE067\uE066\uE066;      # UNMAPPED Malayalam-InterIndic: NUMBER ONE HUNDRED
+൲→\uE067\uE066\uE066\uE066;# UNMAPPED Malayalam-InterIndic: NUMBER ONE THOUSAND
+ൺ→\uE023\uE04D;      # LETTER CHILLU NN, remapped to NNA VIRAMA
+ൻ→\uE028\uE04D;      # LETTER CHILLU N, remapped to NA VIRAMA
```
srl295, line 113: Interesting. Now <https://util.unicode.org/UnicodeJsps/character.jsp?a=ൻ> lists the CHILLU N as confusable with NA+VIRAMA.  IS there any Unicode Metadata that we should be leveraging here?Maybe not.

srl295:
FYI @iwsfutcmd

srl295:
Note that this will go into v41

common/transforms/Malayalam-InterIndic.xml
mjansche: Sorry, missing background: how is confusability influenced by these changes? I'm not familiar with the chain of causation. (Yes, in some sense all chillu letters are arguably confusable with the underlying simple letter plus VIRAMA.)

srl295: @mjansche https://www.unicode.org/versions/Unicode5.1.0/#Malayalam_Chillu_Characters is the table I was looking for, but doesn't look like it's in any other metadata.

I was looking to see if there was a way to avoid needing to do the remap but I don't think so.

There's no causation with confusability.

mjansche: Would it make sense to separately add a set of transform rules that map the <5.1 chillu characters to their >=5.1 equivalents?

macchiati:
That makes sense to me.

Mark

mjansche:
I have no idea why this would fail: https://github.com/unicode-org/cldr/actions/runs/1518165589

I've replicated the test setup as closely as I can locally but cannot reproduce the failure. The error messages are consistent with the changed transform not being loaded, but I see no indication that this is actually the case.

cesargerardoromo:
> Tenga en cuenta que esto entrará en v41

PR #1516, opened 2021-10-03
CLDR-14751 add generative unit number prefixes

macchiati:
CLDR-14751

Note that the modification history doesn't remove the old material; that is done in a separate ticket.


macchiati:
Created against wrong branch

PR #1517, opened 2021-10-03
CLDR-14751 unit number prefixes

macchiati:
CLDR-14751

* Note that this doesn't remove the old Modifications; that is in a separate ticket
* Also, the syntax is different than originally proposed: see ticket

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -890,9 +890,8 @@ The formal syntax for identifiers is provided below.
             </ul></td></tr>
 
 <tr><td>single_unit</td><td>:=</td>
-    <td>(dimensionality_prefix)? prefixed_unit<br/>
-		| currency_unit
-        <ul><li><em>Example: </em>square-meter</li></ul></td></tr>
+    <td>number_prefix? dimensionality_prefix? prefixed_unit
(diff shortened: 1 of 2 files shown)


macchiati:
Note: couldn't add younies as reviewer.

docs/ldml/tr35-general.md
```diff
@@ -900,6 +900,16 @@ The formal syntax for identifiers is provided below.
         <li>See Section 6.6 <a href="#Private_Use_Units">Private-Use Units</a></li>
     </ul></td></tr>
 
+<tr><td>number_prefix</td><td>:=</td>
+    <td>("1"[0-9]+ | [2-9][0-9]*) "-"
```
younies, line 904: I think we can accept `"1"` as a number prefix and assume it is exist by default if there is no number prefix.


Example: "1-meter-per-1-second"  equals "meter-per-second"
this will make no need to check if the number prefix is only one or not and it will be more easier in the coding.

younies (approved):

macchiati:
We could do that, but I think that numbers themselves are an edge case, and
it would be better not to allow extra '1-meter' when it adds no value.
However, this is something we could expand in the future.

Mark

srl295 (approved):

PR #1520, opened 2021-10-04
CLDR-14976 update ICU4J libs to release-70-rc (use different but equivalent tag)

pedberg-icu:
CLDR-14976

Update IC4J libs in CLDR to ICU release-70-rc. Since for some reason that release-70-rc tag was not available in https://github.com/unicode-org/icu/packages/411079/versions (github action failure?), created a separate but equivalent ICU tag  "cldr/2021-09-29" and used that instead.


pedberg-icu:
Oops, this was against wrong branch. Closed this and replaced with https://github.com/unicode-org/cldr/pull/1520 against correct branch.

PR #1523, opened 2021-10-05
CLDR-15035 clean up unit identifier BNF and text

macchiati:
CLDR-15035

While cleaning up the EBNF and other items listed in the ticket, I also reordered the text in that section to make it flow better. That included recasting some of the tables.

The EBNF is also reordered to have the core unit identifier first, then the mixed unit identifier, and finally the long unit identifier.

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -818,29 +818,48 @@ This data is supplied in **Part 6: [Supplemental](tr35-info.md#Contents)**: [Sec
 
 ### 6.2 <a name="Unit_Identifiers" href="#Unit_Identifiers">Unit Identifiers</a>
 
-Units are identified internally as described in this section. As with other identifiers in CLDR, the American English spelling is used for unit identifiers. These names should not be presented to end users, however. As in other cases, the translated names for different languages (or variants of English) are available in the CLDR localized data.
-
-| Name                 | Example |
-| -------------------- | ------- |
-| long unit identifier | length-meter, mass-pound, duration-day |
-| core unit identifier | meter, pound, day |
-  
-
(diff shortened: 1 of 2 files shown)


docs/ldml/tr35-general.md
```diff
@@ -818,29 +818,48 @@ This data is supplied in **Part 6: [Supplemental](tr35-info.md#Contents)**: [Sec
...
+| ------------ | ------------ |
+| meter        | length-meter |
+| pound        | mass-pound   |
+| day          | duration-day |
  
-A core unit that is not a simple unit is called a _complex unit_. It is valid to construct a complex unit identifier from multiple simple unit identifiers using multiplication (kilogram-meter) and division (kilogram-per-meter). As usual, with division the part before the (first) -per- is called the _numerator_, and the part after it is called the _denominator_.
 
-The conversion information uses the short unit identifiers, discarding the unitType. Thus “meter” is used instead of “length-meter”. The translation data currently uses the long unit identifiers, for backwards compatibility. However, that is likely to change in a future version.
+The list of valid CLDR simple unit identifiers is found in _Section [3.11 Validity Data](tr35.md#Validity_Data)_. 
+These names should not be presented to end users, however: the translated names for different languages (or variants of English) are available in the CLDR localized data.
```
srl295, line 846: > These names should not be presented to end users, however: the translated names for different languages (or variants of English) are available in the CLDR localized data. 

OK, I see this was moved rather than dropped. That's good, although I think this information might want to be right up at the top where these identifiers are introduced.

srl295 (approved):
I'm OK with it, and definite improvement.  I'd keep the 'not for end users' note more prominent ideally.

PR #1525, opened 2021-10-05
CLDR-14629 fix ordering of locale aliases

macchiati:
CLDR-14629

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -3459,7 +3459,7 @@ The `languageAlias`, `scriptAlias`, `territoryAlias`, and `variantAlias` element
 
 #### 1. Multimap interpretation
 
-Interpret each languageId as a multimap from a _fieldId_ (language, script, region, variants) to a **set** of field values.
+Interpret each languageId as a multimap from a _fieldId_ (language, script, region, variants) to a ** sorted set** of field values.
 
 _Examples:_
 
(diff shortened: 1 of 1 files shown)


pedberg-icu (approved):

docs/ldml/tr35.md
```diff
@@ -3575,23 +3575,26 @@ The data from supplementalMetadata is (logically) preprocessed as follows.
 4. Change the _type_ and _replacement_ values in the remaining rules into multimap rules, as per _Definition 1. Multimap Interpretation_.
    1. Note that the “und” value disappears.
 5. Order the set of rules by
```
FrankYFTang, line 3577: How about
"5. Order the set of rules by descending order, as defined below"

docs/ldml/tr35.md
```diff
@@ -3575,23 +3575,26 @@ The data from supplementalMetadata is (logically) preprocessed as follows.
...
-   1. the size of the union of all field value sets, with largest size first
-   2. and then alphabetically by field.
+   1. The size of the union of all field value sets, with largest size first
+   2. And then by field, where L < S < R < V. 
+     * So {L=X, S=Z} < {L=X, R=Y}.
+     * After this point we are guaranteed to have the same set of fields
+   3. And then by field values sets in the order L, S, R, V.
```
FrankYFTang, line 3582: How about just
```
3. If A.L < B.L then A < B
4. If A.L > B.L then A > B
5. If A.S < B.S then A < B
6. If A.S > B.S then A > B
7. If A.R < B.R then A < B
8. If A.R > B.R then A > B
9. If A.V < B.V then A < B
10. If A.V > B.V then A > B
```

docs/ldml/tr35.md
```diff
@@ -3575,23 +3575,26 @@ The data from supplementalMetadata is (logically) preprocessed as follows.
 4. Change the _type_ and _replacement_ values in the remaining rules into multimap rules, as per _Definition 1. Multimap Interpretation_.
    1. Note that the “und” value disappears.
 5. Order the set of rules by
-   1. the size of the union of all field value sets, with largest size first
-   2. and then alphabetically by field.
+   1. The size of the union of all field value sets, with largest size first
+   2. And then by field, where L < S < R < V.
```
FrankYFTang, line 3579: How about
"   2. And then by field, in the order of L, S,  R, and V."

docs/ldml/tr35.md
```diff
@@ -3575,23 +3575,26 @@ The data from supplementalMetadata is (logically) preprocessed as follows.
 4. Change the _type_ and _replacement_ values in the remaining rules into multimap rules, as per _Definition 1. Multimap Interpretation_.
    1. Note that the “und” value disappears.
 5. Order the set of rules by
-   1. the size of the union of all field value sets, with largest size first
-   2. and then alphabetically by field.
+   1. The size of the union of all field value sets, with largest size first
+   2. And then by field, where L < S < R < V. 
+     * So {L=X, S=Z} < {L=X, R=Y}.
```
FrankYFTang, line 3580: How about "* So {L=X, S=Z} before {L=X, R=Y}."

docs/ldml/tr35.md
```diff
@@ -3575,23 +3575,26 @@ The data from supplementalMetadata is (logically) preprocessed as follows.
...
+     * After this point we are guaranteed to have the same set of fields
+   3. And then by field values sets in the order L, S, R, V. 
+     * To determine the ordering between a field set A and B, traverse each in parallel
+     * If the field values are all equal, the result is equal.
+     * Otherwise at the first differing field values, A is less if its field value is less, otherwise B is less.
```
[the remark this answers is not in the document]
macchiati: The first differing value is fonipa vs hepburn, so the fonipa rule comes first. That's what "Otherwise at the first differing field values, A is less if its field value is less, otherwise B is less." covers.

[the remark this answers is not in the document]
FrankYFTang: The data in the file is in the form of a _ separated string , such as 
<languageAlias type="und_hepburn_heploc" replacement="und_alalc97" reason="deprecated"/>
and we change und_hepburn_heploc to the V={hepburn, heploc} notation here. If data is somehow
<languageAlias type="und_heploc_hepburn" replacement="und_alalc97" reason="deprecated"/>
should it still be V={hepburn, heploc}? which mean we need sort the fields in V ? 
or we should just in the data check in process ensure that will never happen?
My understand the {A, B} is a set notation not an ordered list notation (please correct me if I misunderstand) and if that is case then "the first" in a set does not make sense unless we specifcy the order in the set.

PR #1526, opened 2021-10-05
CLDR-14855 build: speedup by caching ~/.npm

srl295:
CLDR-14855


.github/workflows/ansible-lint.yml
```diff
@@ -14,6 +14,14 @@ jobs:
       uses: ansible/ansible-lint-action@c37fb7b4bda2c8cb18f4942716bae9f11b0dc9bc
       with:
         targets: tools/scripts/ansible/*-playbook.yml
+    - name: Cache local npm repository
```
btangmu, line 17: Interesting! Seems like this might save a lot of network traffic as well as time.

I wonder if we could avoid duplicating the same 8 lines of code in 3 yml files -- is there a good way to share code between workflows?

btangmu (dismissed):

.github/workflows/ansible-lint.yml
srl295: Excellent idea. Filed https://unicode-org.atlassian.net/browse/CLDR-15085 

I think these 8 lines are more minor compared to the Java/Maven setup and build, those would be even better to reuse.

srl295:
linter needed to have a separate cache, it actually doesn't share resources.

PR #1535, opened 2021-10-07
CLDR-15048 ignore more settings files

srl295:
- ignore and delete org.eclipse.jdt.apt.core.prefs
- cleanup .gitignores

CLDR-15048


srl295:
It looks like there needs to be some local steps run to handle other types of settings, I've updated the ticket.

PR #1541, opened 2021-10-15
CLDR-13731 Report in ConsoleCheck if alt without non-alt

btangmu:
-New CheckAltOnly.java to check if only the alt path is present but the non-alt path is absent

-New TestCheckAltOnly.java (unit test for CheckAltOnly.java)

-Fix some warnings

-Change some public to private

-Comments

CLDR-13731


btangmu:
See https://github.com/unicode-org/cldr/pull/1542 or later

PR #1566, opened 2021-10-28
CLDR-13731 Report in ConsoleCheck if alt without non-alt

btangmu:
-New CheckAltOnly.java to check if only the alt path is present but the non-alt path is absent

-New CheckCLDR.CheckStatus.Subtype.missingNonAltPath

-New TestCheckAltOnly.java (unit test for CheckAltOnly.java)

-CheckAltOnly: no error for the root locale

-CheckAltOnly: no error if the non-alt path has a constructed value

-CheckAltOnly: no error if getSourceLocaleIdExtended does not return XMLSource.CODE_FALLBACK_ID

-CheckAltOnly: no error if currency symbol with alt is same as root alt value

-XML data changes: remove alt from error-triggering paths (move to non-alt) for locale display names

-XML data changes: delete currency types RUR in pa.xml and GEL in to.xml

-Fix TestCheckDisplayCollisions: frCaResolved was clearly intended to be resolved

-Fix some warnings

-Change some public to private

-Comments

CLDR-13731

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
@@ -83,7 +83,7 @@ public class TestCheckDisplayCollisions extends TestFmwkPlus {
         CLDRFile frResolved = factory.make("fr", true);
         checkFile(cdc, fr, frResolved);
 
-        CLDRFile frCaResolved = factory.make("fr_CA", false);
+        CLDRFile frCaResolved = factory.make("fr_CA", true);
         checkFile(cdc, frCA, frCaResolved,
             scorpioEmoji, ukRegion);
     }
--- a/common/main/bo.xml
+++ b/common/main/bo.xml
@@ -155,7 +155,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<territory type="GT" draft="unconfirmed">གྷོ་ཊེ་མ་ལ།</territory>
 			<territory type="GW" draft="unconfirmed">གྷི་ནི་ཡ་བིས྄་སོ།</territory>
 			<territory type="GY" draft="unconfirmed">གྷུ་ཡཱ་ན།</territory>
-			<territory type="HK" alt="short" draft="unconfirmed">ཧོང་ཀོང༌།</territory>
+			<territory type="HK" draft="unconfirmed">ཧོང་ཀོང༌།</territory>
 			<territory type="HN" draft="unconfirmed">ཧོན་དུ་རས྄།</territory>
 			<territory type="HR" draft="unconfirmed">ཀུརོ་ཤི་ཡ།</territory>
 			<territory type="HT" draft="unconfirmed">ཧེ་ཏི།</territory>
@@ -198,7 +198,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<territory type="ML" draft="unconfirmed">མ་ལི།</territory>
 			<territory type="MM" draft="unconfirmed">འབར་མ།</territory>
 			<territory type="MN" draft="unconfirmed">སོག་ཡུལ།</territory>
-			<territory type="MO" alt="short" draft="unconfirmed">མེ་ཀའོ།</territory>
+			<territory type="MO" draft="unconfirmed">མེ་ཀའོ།</territory>
 			<territory type="MR" draft="unconfirmed">མཽ་རི་ཏ་ནི་ཡ།</territory>
 			<territory type="MT" draft="unconfirmed">མལ་ཊ།</territory>
 			<territory type="MU" draft="unconfirmed">མཽ་རིཤས྄།</territory>
--- a/common/main/ig.xml
+++ b/common/main/ig.xml
@@ -40,7 +40,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="ce">Chechen</language>
 			<language type="ceb">Cebụanọ</language>
 			<language type="chr">Cheroke</language>
-			<language type="ckb" alt="menu">Kurdish ọsote</language>
+			<language type="ckb">Kurdish ọsote</language>
 			<language type="ckb" alt="variant">Kurdish ọzọ</language>
 			<language type="co">Kọsịan</language>
 			<language type="cs">Cheekị</language>
--- a/common/main/ku.xml
+++ b/common/main/ku.xml
@@ -406,7 +406,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<territory type="GU">Guam</territory>
 			<territory type="GW">Gîne-Bissau</territory>
 			<territory type="GY">Guyana</territory>
-			<territory type="HK" alt="short">Hong Kong</territory>
+			<territory type="HK">Hong Kong</territory>
 			<territory type="HN">Hondûras</territory>
 			<territory type="HR">Kroatya</territory>
 			<territory type="HT">Haîtî</territory>
@@ -456,7 +456,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<territory type="ML">Malî</territory>
 			<territory type="MM">Myanmar (Birmanya)</territory>
 			<territory type="MN">Mongolya</territory>
-			<territory type="MO" alt="short">Makao</territory>
+			<territory type="MO">Makao</territory>
 			<territory type="MP">Giravên Bakurê Marianan</territory>
 			<territory type="MQ">Martinique</territory>
 			<territory type="MR">Morîtanya</territory>
--- a/common/main/pa.xml
+++ b/common/main/pa.xml
@@ -6572,9 +6572,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<symbol>RUB</symbol>
 				<symbol alt="narrow">₽</symbol>
 			</currency>
-			<currency type="RUR">
-				<symbol alt="narrow" draft="contributed">р.</symbol>
-			</currency>
 			<currency type="RWF">
 				<displayName>ਰਵਾਂਡਨ ਫ੍ਰੈਂਕ</displayName>
 				<displayName count="one">ਰਵਾਂਡਨ ਫ੍ਰੈਂਕ</displayName>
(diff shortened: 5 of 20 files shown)


macchiati (approved):

btangmu:
@macchiati @srl295 now this is approved, it still has 51 failing data tests as expected, so two options are:

1. fixing the data before merging to main
2. merging to main in spite of the failures (overriding github required status -- how?), then fixing the failures in a subsequent PR

I imagine pros and cons both ways. Would you like me to try to fix the data (with some guidance)? If so, then maybe I should simply combine the code and data changes in a subsequent PR. On the other hand, if others than me are going to fix the data, it could be convenient to have the tests committed now -- though also maybe a nuisance to have ongoing errors on main

btangmu:
BTW about overriding github required status, I've seen these:

PR Flags

The following "variables" can be set in the pull request body to change the behavior of the checker.

// Allow commits with mixed issue tags
DISABLE_JIRA_ISSUE_MATCH=true

// Allow multiple commits in the same PR
ALLOW_MANY_COMMITS=true

https://github.com/unicode-org/jira-github-pr-check

... but I don't see a way to override cldr-mvn/check errors...

macchiati:
There is still something a odd with the test.

*2021-10-26T16:51:12.4860622Z ff_Adlm [Fulah (Adlam)]*

▸Currencies|Western_Asia_(C)|Western_Asia:_Lebanon|LBP-symbol-narrow◂

〈L£〉

〈L£〉

⁅missing non alt path⁆

❮Error: A path with alt is present but the corresponding path without alt
is missing; a solution might be to confirm the non-alt path❯

*https://st.unicode.org/cldr-apps/v#/ff_Adlm//714ef3a5e09c1bb3*
<https://st.unicode.org/cldr-apps/v#/ff_Adlm//714ef3a5e09c1bb3>

root

The table indicates (and clicking on the link) that there is a value in
root. https://st.unicode.org/cldr-apps/v#/root/C_WAsia/7a65298807a5c2e0
Now, what may be happening is that in the ST we are reporting as "root"
items that are also just in code-fallback. Can you confirm that this is the
case?

As to the question, we should be able to fix these as part of the checkin.
Please create a spreadsheet to copy in the table, and we can review and
indicate the right change. Key questions.

   1. If the alt path is not in root, we want to know about it. In case it
   was mistakenly entered with a bulk upload, we may want to delete the alt
   path as the solution.
   2. For the others, we will generally want to add the explicit non-alt
   value


Note: when we finish the release, we also need to reset the Phase for
checking, so that introducing errors is allowed. Maybe we don't make that
clear enough though.

(When we add a test that exposes an error, we generally want to allow for
the vetters to fix it because we don't know how to. In this case, it is
much easier.)

Mark

srl295:
I missed the discussion around over writing checks, but I really don’t
think we should do that. If we can avoid it. There’s actually a way to
create a Q of requests to be merged and then another pull request could be
added to fix the errors if that’s what’s needed

macchiati:
We have often fixed data in the same PR that we fixed (or added to) a test.
Is it the volume of changes that concerns you?


Mark

srl295:
No, I think it should be fixed in the same pull request, I thought there
was a discussion about merging this to main with errors.

btangmu:
@macchiati @srl295 my understanding is that both of you prefer my option "1. fixing the data before merging to main"

in that case, the question of how to override is moot (for now at least) since that would have been my option "2."

next I should address Mark's question about something "...odd with the test..."

btangmu:
@macchiati I'm attaching the table (as an xlsx file) here:
[51.xlsx](https://github.com/unicode-org/cldr/files/7455293/51.xlsx)

btangmu:
@macchiati @srl295 a simple and obvious fix would be to remove from the xml the values for the alt paths in question. I imagine that vetters could subsequently import their old votes for those alt values, and vote for them again along with voting for corresponding non-alt values -- is that right? Is there a better way?

srl295:
> Now, what may be happening is that in the ST we are reporting as "root"
> items that are also just in code-fallback. Can you confirm that this is the
> case?

That's not what's happening here. This data is in `root.xml` and that is the cause of at least some of these.

```xml
                        <currency type="PKR">
                                <symbol alt="narrow">Rs</symbol>
                        </currency>
                        <currency type="PLN">
                                <symbol alt="narrow">zł</symbol>
                        </currency>
                        <currency type="PYG">
                                <symbol alt="narrow">₲</symbol>
                        </currency>
                        <currency type="RON">
                                <symbol alt="narrow">lei</symbol>
                        </currency>
                        <currency type="RUB">
                                <symbol alt="narrow">₽</symbol>
                        </currency>
                        <currency type="RWF">
                                <symbol alt="narrow">RF</symbol>
                        </currency>
                        <currency type="SBD">
                                <symbol alt="narrow">$</symbol>
                        </currency>
                        <currency type="SEK">
                                <symbol alt="narrow">kr</symbol>
                        </currency>
                        <currency type="SGD">
                                <symbol alt="narrow">$</symbol>
                        </currency>
                        <currency type="SHP">
                                <symbol alt="narrow">£</symbol>
                        </currency>
                        <currency type="SRD">
                                <symbol alt="narrow">$</symbol>
                        </currency>
                        <currency type="SSP">
                                <symbol alt="narrow">£</symbol>
                        </currency>
                        <currency type="STN">
                                <symbol alt="narrow">Db</symbol>
                        </currency>
                        <currency type="SYP">
                                <symbol alt="narrow">£</symbol>
                        </currency>
                        <currency type="THB">
                                <symbol alt="narrow">฿</symbol>
                        </currency>
                        <currency type="TOP">
                                <symbol alt="narrow">T$</symbol>
                        </currency>
```

Should we exclude currency symbols here? Because I think this is a case where the normal symbol would be, say, `THB` even if the narrow is `฿`.

Alternatively (if you pardon the pun!) perhaps _for completeness_ we should explicitly update all of these to be:

```xml
                        <currency type="RUB">
                                <symbol>RUB</symbol>  <!-- ««« Add these (from code-fallback) explicitly -->
                                <symbol alt="narrow">₽</symbol>
                        </currency>
```

srl295:
- `yi` name for `TL` - this should _move_ from TL-variant to TL. For some reason they translated the variant but not the regular one.  This seems like exactly what the test is designed to catch.

srl295:
These other three are also good positive fails.

- `ig` Kurdish - just move these from alt to regular. Another good positive, the translator should translate the non-alt first. This one can be solved with a move, I think.

- All of the ones around MO, HK, CD, CG:  The translators need to complete the work here: in most cases they've *declined to translate the parenthetical* SAR, SAR, Kinshasa, Brazzaville,  respectively. Here I agree to just delete these entries so they can be sorted out.
- `CZ` and `PS` corsican translator declined to translate "Czechia" and "Palestinian Territories". Delete the alt form, redo both. Another good hit.

srl295:
I think I've reviewed all of the data errors in the spreadsheet, with two possible solutions for the currency part.

btangmu:
@srl wrote "This data is in root.xml and that is the cause of at least some of these ..."
The examples you gave in root are all "alt" paths. However, CheckAltOnly is interested in the source locales of the "non-alt" paths (corresponding to "alt" paths in non-root locales). So, it's not clear to me whether these examples are relevant for this ticket -- if they are, what's a particular example out of the 51 errors?

macchiati:
Mark


On Mon, Nov 1, 2021 at 1:31 PM Tom Bishop ***@***.***> wrote:

> @macchiati <https://github.com/macchiati> Yes I can confirm that in the
> ST we are reporting as "root" items that are also in code-fallback. The
> front end, by design, changes "code-fallback" to "root" in such links per
> an old completed ticket (
> https://unicode-org.atlassian.net/browse/CLDR-11622 ).
>
That's what I thought might be happening. What is a mystery to me is why
other cases like STN don't also show up as 'root'. They should also be
coming from code-fallback.

For the non-alt path in question in root (
> http://localhost:9080/cldr-apps/v#/root/C_WAsia/7a65298807a5c2e0 ), the
> Info Panel says for "LBP", "This item is an untranslated code" and "This
> item is constructed from other values". The status in root is "Missing".
> The test (CheckAltOnly) would not trigger an error unless the source locale
> were "code-fallback" (on the back end).
>
OK, glad you confirmed.

> BTW the message "This item is constructed from other values" is
> potentially confusing: it's actually used for "code-fallback" in general,
> not to be confused with CLDRFile.getConstructedValue which returns null
> except for a very limited set of paths that always contain
> "//ldml/localeDisplayNames/languages/language".
>
Hmmm. We probably should change that message.

macchiati:
Yes, that is probably an error.

btangmu:
@macchiati @srl295 here's a google sheet with the 51 errors:
https://docs.google.com/spreadsheets/d/1rXa1XXyP76hrspLtHSis4uH6FJqzAUzcwVNoOYmMUow/edit#gid=1818827278
I've also linked to it from today's infrastructure agenda

macchiati:
Thanks!

Mark

btangmu:
@macchiati @srl295 there are still two console-check errors, both currency symbols with alt, but NOT same as root alt value:

https://st.unicode.org/cldr-apps/v#/pa//386035af20c8d53a

https://st.unicode.org/cldr-apps/v#/to//268bf25be49df55b

srl295:
I think they are both leftover from old root values and should be deleted.

One is the OLD Russian ruble as `p.` and the other is the georgian Lek symbol when it was a variant and not in the main narrow.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
```diff
@@ -83,7 +83,7 @@ public void testInheritance() {
         CLDRFile frResolved = factory.make("fr", true);
         checkFile(cdc, fr, frResolved);
 
-        CLDRFile frCaResolved = factory.make("fr_CA", false);
+        CLDRFile frCaResolved = factory.make("fr_CA", true);
```
srl295, line 86: 👍

srl295 (approved):

srl295:
i approve pending test fixes of course…

btangmu:
> I think they are both leftover from old root values and should be deleted.
> 
> One is the OLD Russian ruble as `p.` and the other is the georgian Lek symbol when it was a variant and not in the main narrow.

I'll delete both and update this PR

srl295 (approved):

PR #1567, opened 2021-10-28
CLDR-15126 Bogus xpath warning when press Previous or Next button

btangmu:
-In cldrSurvey.chgPage set id to null to prevent warning when that id is not found on new page

CLDR-15126


srl295 (approved):

macchiati:
A question: is the ID not found because of the coverage level? Or is some
underlying problem really there (eg mapping from ID to page is failing)?
Mark

btangmu:
@macchiati you asked "is the ID not found because of the coverage level? Or is some
underlying problem really there (eg mapping from ID to page is failing)?"

In the context of Previous/Next buttons, the ID -- identifying a row (xpath) -- is not found because we've switched sections and that row isn't in the new section, it was in the old section

macchiati:
got it, thanks!
Mark

srl295:
I added a ’steven recommend’ column

PR #1571, opened 2021-11-02
CLDR-15080 Fix broken links for Survey Tool instructions at sites.google.com

btangmu:
-Change some links from sites.google.com to cldr.unicode.org

-TODO comment where I could not find new page

-Remove defunct bugFeedbackUrl for legacy st_footer.jsp

CLDR-15080

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckQuotes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckQuotes.java
@@ -10,6 +10,8 @@ import org.unicode.cldr.util.PatternCache;
 import com.ibm.icu.text.UnicodeSet;
 
 public class CheckQuotes extends CheckCLDR {
+    private static final String VALID_DELIMITER_URL = "https://cldr.unicode.org/translation/characters";
+
     private static final Pattern ASCII_QUOTES = PatternCache.get("[\'\"]");
     private static final Pattern UNITS = PatternCache.get("//ldml/units/.*");
     private static final Pattern DELIMITERS = PatternCache.get("//ldml/delimiters/.*");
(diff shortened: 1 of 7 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckQuotes.java
```diff
@@ -10,6 +10,8 @@
 import com.ibm.icu.text.UnicodeSet;
 
 public class CheckQuotes extends CheckCLDR {
+    private static final String VALID_DELIMITER_URL = "https://cldr.unicode.org/translation/characters";
```
AEApple, line 13: So this page was originally at: https://cldr.unicode.org/translation/characters-emoji-symbols 

And had 3 subpages:

https://cldr.unicode.org/translation/characters/character-labels
https://cldr.unicode.org/translation/characters/emoji-names-and-keywords -> parameter would need to be updated to align with original.
https://cldr.unicode.org/translation/characters/typographic-names

I am happy to revert to match the old site, but I'd like to confirm that you agree on the approach before making the changes.

AEApple: Will update the site to link the subpages on the main header page which didn't migrate properly.

AEApple (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckQuotes.java
btangmu: Per today's meeting I'll change all remaining "characters-emoji-symbols" to "characters"

AEApple: Actually do you think this was the original characters page? https://cldr.unicode.org/translation/core-data/alphabetic-information

I am wondering after looking up some of the other links you weren't able to find originally.

PR #1576, opened 2021-11-04
CLDR-15101 Support for Ethiopic-Ethiopic (Gurage) Transliteration

dyacob:
CLDR-15101


btangmu (approved):
LGTM! Plus one for having an extensive test (8,532 lines) -- I'm not sure if we have any limits or guidelines on test size.

Some characters such as 1E7E8 𞟨 ETHIOPIC SYLLABLE GURAGE HHWA don't have fonts on my computer; their usage looks consistent though.

btangmu:
@macchiati the test is 227, 999 bytes (8,532 lines), making it the largest in its folder -- for this and future similar PRs, is there any limit or guideline for test size? I've added you as a reviewer, though I've already given my approval

dyacob:
FWIW, the long list of test words comes from the "sgw" file in the [Unilex](https://github.com/unicode-org/unilex/) repo (with single letter entries removed): https://github.com/unicode-org/unilex/blob/master/data/frequency/sgw.txt

btangmu:
@macchiati and @dyacob I went ahead and merged this -- we can revisit later with a new PR if we decide to reduce the test file size. Thanks!

PR #1578, opened 2021-11-04
CLDR-15080 Fix broken links for Survey Tool instructions

btangmu:
-Change some links from cldr.org to cldr.unicode.org

-Change some http to https

-Change translation/currency-names to translation/currency-names-and-symbols

-Change translation/numbers-currency/currency-names to translation/currency-names-and-symbols

-Change translation/characters-emoji-symbols to translation/characters

-Change some other broken links to match existing pages on cldr.unicode.org

-Encapsulate by moving some links into CLDRURLS.java

-Remove PathDescription.txt, move into .java, replace URLs with their CLDRURLS names

-Use loadFromString instead of loadFromFile; new RegexLookup.SEPARATOR encapsulates semi-colon usage

-Remove obsolete XMB code

-Fix deprecation warnings with STANDARD_CODES.zoneParser

-Format PathDescription.java with Eclipse

CLDR-15080


tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRURLS.java
```diff
@@ -88,6 +89,50 @@
      */
     public static final String UNKNOWN_REVISION = "(unknown)";
 
+    public static final String GENERAL_HELP_URL = "https://cldr.unicode.org/translation/";
+    public static final String ADMIN_HELP_URL = "https://cldr.unicode.org/index/survey-tool/survey-tool-administration";
+    public static final String CAPITALIZATION_URL = "https://cldr.unicode.org/translation/translation-guide-general/capitalization";
+    public static final String CHARACTERS_HELP = "https://cldr.unicode.org/translation/characters";
```
srl295, line 95: Maybe some factoring? 
```suggestion
    public static final String CHARACTERS_HELP = GENERAL_HELP_URL + "characters";
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRURLS.java
```diff
@@ -88,6 +89,50 @@
      */
     public static final String UNKNOWN_REVISION = "(unknown)";
 
+    public static final String GENERAL_HELP_URL = "https://cldr.unicode.org/translation/";
```
srl295, line 92:
```suggestion
    public static final String GENERAL_HELP_URL = CLDR_HOMEPAGE + "/translation/";
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
```diff
@@ -19,14 +19,651 @@
 import com.ibm.icu.util.Output;
 
 public class PathDescription {
+    /**
+     * Remember to quote any [ character!
+     */
+    final private static String pathDescriptionString = ""
```
srl295, line 25: I'm surprised by this,  why is this better than `PathDescription.txt` ?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRURLS.java
btangmu: It's worth considering building them up from pieces, as you suggest. However I think it might be better to keep the complete URL strings in the source code so they're _searchable_. Whenever someone discovers a broken link, it will be easy to find the URL in the source and fix it, without needing to remember where to look (CLDRURLS.java). The string "characters" might occur in a zillion source files. Finding CHARACTERS_HELP by way of GENERAL_HELP_URL could be time-consuming and error-prone.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
btangmu: For encapsulation and maintainability. This way we can use the CLDRURL constants instead of the string literals, and we can use them for path descriptions the same way we use them in other java files.

PathDescription.txt contains the same literal URLs repeatedly, embedded inside long message strings, which in turn are concatenated to regex followed by semi-colons. Those long lines are a pain in the horizontal scrollbar.

I considered implementing a substitution mechanism so the CLDRURL constants could be used in the .txt file. Then I realized what a waste of time that would be, and also how inferior. For example, the java compiler will tell us if we use a constant that's missing from CLDRURL. We don't need to invent a new file format with its own macro language. Then I noticed that we already had loadFromString :-)

srl295 (approved):
It can work. I suppose we can use search and replace to refactor in the future.

PR #1581, opened 2021-11-05
CLDR-15101 Support for Ethiopic-Latin (Aethiopica) Transliteration

dyacob:
CLDR-15101


dyacob:
An oddity with this one, I could not use the variant attribute value "Aethiopica" and had to use "Aeth" (I've used longer string since in other transliterations).  If I used the full name "Aethiopic" as the variant identifier, the test suite would run the new test file with the name "latn-t-Ethi-m0-aethiopica.txt" against the older Ethiopic-Latin.xml transliteration and fail.  Using the shorter name, "Aeth", this issue is avoided.

btangmu (approved):

PR #1582, opened 2021-11-06
CLDR-15101 Support for Ethiopic-Latin (Beta Maṣāḥǝft) Transliteration

dyacob:
CLDR-15101


btangmu (changes requested):
Per our slack discussion, let's get clarification about the status of Beta Maṣāḥǝft -- such as whether it's intended to define a distinct standard, and whether it's defined by a document distinct from its implementation.

btangmu (approved):

PR #1589, opened 2021-11-10
CLDR-15101 Support for Ethiopic-Latin (SERA) Transliteration

dyacob:
CLDR-15101


btangmu (approved):
que sera sera

PR #1594, opened 2021-11-10
CLDR-15137 SBRS41 Update timzone db files with 2021e

yumaoka:
CLDR-SBRS41


pedberg-icu (approved):
These look fine to me (taken from tzdb)

PR #1597, opened 2021-11-11
CLDR-15166 Fix date and time localization

sayuri-gi:
[CLDR-15166](https://unicode-org.atlassian.net/browse/CLDR-15166)

**Description**
Current date and time localization for sinhala (si) language is incorrect. See below examples for understand it and PR contain eg: traditional months names that exist since decades.

**Now**
e.g. 1: January--> ජනවාරි (pron: janavaari)
e.g. 2: February --> පෙබරවාරි (pron: pebaravaari)
e.g. 3: Minute --> මිනිත්තුව (pron: miniththuwa)

**Pull Request**
e.g. 1: January --> දුරුතු (pron: duruthu)
e.g. 2: February --> නවම් (pron: navam)
e.g. 3: Minute --> විනාඩිය (pron: vinaadiya)

Reference: https://masteranylanguage.com/c/b/o9ly878fJ/98-Sinhala-MonthsOfYear-Learn-Step-By-Step
(You can see the both 'now' and 'PR' with pronunciation)

- Fix date and time localization
- Other minor fixes

### DIFF
--- a/common/main/si.xml
+++ b/common/main/si.xml
@@ -809,7 +809,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		</keys>
 		<types>
 			<type key="calendar" type="buddhist">බොදු දින දසුන</type>
-			<type key="calendar" type="chinese">චීන දින දසුන </type>
+			<type key="calendar" type="chinese">චීන දින දසුන</type>
 			<type key="calendar" type="dangi">ඩැන්ගී දින දසුන</type>
 			<type key="calendar" type="ethiopic">ඉතියෝපියානු දින දසුන</type>
 			<type key="calendar" type="gregorian">ග්‍රෙගරියානු දින දසුන</type>
@@ -2329,7 +2329,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</relativeTime>
 			</field>
 			<field type="minute-short">
-				<displayName>මිනි.</displayName>
+				<displayName>විනා.</displayName>
 				<relativeTime type="future">
 					<relativeTimePattern count="one">විනාඩි {0}කින්</relativeTimePattern>
 					<relativeTimePattern count="other">විනාඩි {0}කින්</relativeTimePattern>
@@ -2340,7 +2340,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</relativeTime>
 			</field>
 			<field type="minute-narrow">
-				<displayName>මිනි.</displayName>
+				<displayName>විනා.</displayName>
 				<relativeTime type="future">
 					<relativeTimePattern count="one">විනාඩි {0}කින්</relativeTimePattern>
 					<relativeTimePattern count="other">විනාඩි {0}කින්</relativeTimePattern>
@@ -6195,7 +6195,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<unitPattern count="other">අංශක {0}</unitPattern>
 			</unit>
 			<unit type="angle-arc-minute">
-				<displayName>චාපමිනිත්තු</displayName>
+				<displayName>චාපවිනාඩි</displayName>
 				<unitPattern count="one">චාපවිනාඩි {0}</unitPattern>
 				<unitPattern count="other">චාපවිනාඩි {0}</unitPattern>
 			</unit>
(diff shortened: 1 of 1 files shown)


srl295:
@sayuri-gi 
Thanks for the PR! This is usually handled with the ST.

That said…
does ජනවාරි vs දුරුතු have to do with a different calendar _system_? i.e. is ජනවාරි used in some other context?

sayuri-gi:
> That said… does ජනවාරි vs දුරුතු have to do with a different calendar _system_? i.e. is ජනවාරි used in some other context?
> https://en.wikipedia.org/wiki/Buddhist_calendar#Number_of_days_per_month implies that these are buddhist calendar months instead of gregorian

දුරුතු etc are traditional months names and it can use with any calendar system (not limited to a specific). We also have other calendar systems that undocumented on internet-media as well. ජනවාරි etc english words have appeared as effort of establishing western interests.
Furthermore, several times we noticed this problem and fixed on other IT projects; and maintainers said that they extracted localization data from cldr then decided to fix this here.

image

itsnotsaved:
@srl295 How long will takes to merge this? I noticed several pre-releases on https://github.com/unicode-org/cldr/releases or does anyone have any doubt about this?

srl295:
@AEApple you own the ticket ^

DavidLRowe:
@sayuri-gi  In particular, from the page mentioned above, use the "Unicode contact form" link to go to: http://www.unicode.org/reporting.html and, once there, use the "Request CLDR Submitter ID" link so that you end up at:
https://corp.unicode.org/reporting/cldr-id.html
where you'll enter email, name, language name and language code.

AEApple:
Great! Closing PR per discussion since this should be discussed in the Survey Tool.

sayuri-gi:
@srl295 @DavidLRowe @AEApple We (our team members) do not understand how to use ST properly. Account created person did not respond to our mails.

@AEApple I saw your recent activities e.g. https://github.com/unicode-org/cldr/pull/2108 https://github.com/unicode-org/cldr/pull/2109 https://github.com/unicode-org/cldr/pull/2114 https://github.com/unicode-org/cldr/pull/2116 and more. Can you check this PR contents and apply them to unicode? we do not expect any credits.

srl295:
> @srl295 @DavidLRowe @AEApple We (our team members) do not understand how to use ST properly. Account created person did not respond to our mails.
> 

@rickmcgowan fyi

sayuri-gi:
>If your vote does NOT show up in the winning column, it is because you need to discuss it further. To do this:
Click Discuss and enter why you want to make this change, to explain it to other translators. Check your email and the forum link https://st.unicode.org/cldr-apps/v#forum/si// for any replies.
>The data is sourced from the survey tool, it is not managed in git during this phase. There are others contributing Sinhala, so it is important to coordinate with them.

It needs many votes and i know some existing translators are not active and also, we've lost their connections / collaboration in the past then this is not going to work or may be will take many years. 
And i saw unicode staff keep adding language variants, versions directly to unicode without delays or oppositions and lack contribution for core languages. Just a variants / versions mean lot of loan words / grammar etc from one or more languages that leads to kill indigenous languages, affect core language (original language of that versions/variants) and dominance of unexpected language with time 

>otherwise, we shall ignore this and work to fix affected projects that depend on unicode data
>>No, don't ignore it, that is counter productive. Please use the process I have outlined.

That's the effective solution for now. However we already added them to ST.

AEApple:
I would also recommend including an explanation of why the new value is preferred (e.g. is more commonly used) so the other vetters can understand why you are proposing it.

srl295:
Hi, @sayuri-gi 

> It needs many votes…

That's OK, that's why you give your reasoning, to encourage others to vote the same

> and i know some existing translators are not active

Actually, they are indeed active.

> and also, we've lost their connections / collaboration in the past then this is not going to work or may be will take many years.

I'm not sure what you mean by 'in the past' if you are only now being added to the Survey Tool. Did you collaborate in some other way?

>  However we already added them to ST.

Great.

> And i saw unicode staff keep adding language variants, versions directly to unicode without delays or oppositions and lack contribution for core languages. Just variants / versions mean lot of loan words / grammar etc from one or more languages that lead to kill indigenous languages, affect core language (original language of that versions/variants) and dominance of unexpected language with time

Hm, I'm not sure what you mean or what staff you are referring to.  The Unicode CLDR project—which is made up of participants from many different organizations such as yourself— is actually doing quite a bit to support all languages. Is there a specific problem?

> > otherwise, we shall ignore this and work to fix affected projects that depend on unicode data
> > > No, don't ignore it, that is counter productive. Please use the process I have outlined.
> 
> That's the effective solution for now.

Thanks.  When you get the posts in place, that will let all of the vetters in a language discuss and come to the right conclusion.

PR #1599, opened 2021-11-11
CLDR-15056 Coverage Progress Bars front end

btangmu:
-Bug fix: filter for coverage when count rows in section

-Bug fix: avoid division by zero

-Make user-friendly percentage, never round down to zero or up to 100

-Implement a front end for locale completion

-Update voter completion after vote/abstain, updateStatsOneVote just as for section completion

-Call cldrLoad.handleCoverageChanged from cldrLoad.coverageUpdate, do not wait for menu choice

-Improve layout and style consistency of nav-page

-Refactor with new class MeterData; the vue component only has display logic

-Improve cldrStatus.setCurrentId and its unit test: normalize to string

-Remove unneeded timeout for cldrProgress.insertWidget

-Remove some unused imports and exports

-Comments

CLDR-15056

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrProgress.js
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.js
@@ -5,6 +5,7 @@
  */
 import * as cldrAjax from "./cldrAjax.js";
 import * as cldrCoverage from "./cldrCoverage.js";
+import * as cldrLoad from "./cldrLoad.js";
 import * as cldrStatus from "./cldrStatus.js";
 
 import ProgressMeters from "../views/ProgressMeters.vue";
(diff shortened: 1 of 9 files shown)


tools/cldr-apps/js/src/esm/cldrProgress.js
```diff
@@ -18,30 +19,62 @@ const USE_NEW_PROGRESS_WIDGET = false;
...
+    this.level = level;
+    if (!description) {
+      this.percent = 0;
+      this.title = "No data";
+      return;
+    }
+    this.percent = friendlyPercent(votes, total);
+    const locale = cldrStatus.getCurrentLocale();
+    const localeName = cldrLoad.getLocaleName(locale);
+    this.title =
```
srl295, line 41: I'd probably do this text calculation closer to the glass (such as a `ref`  in a Vue template ) so that it can be [reactive](https://v3.vuejs.org/guide/reactivity-fundamentals.html), but perhaps this is done as a class so that it can plug in to the `<a-progress`.

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrProgress.js
btangmu: Whenever cldrProgress calls updateSectionMeter, updateVoterMeter, or updateLocaleMeter, the data in ProgressMeters.vue gets updated and displayed -- so it is _reactive_, as I understand that word.

I can't say that I fully understand https://v3.vuejs.org/guide/reactivity-fundamentals.html or when/why it would be appropriate to use the "reactive" keyword to "create a reactive state from a JavaScript object". I don't think anything like that is necessary in this case.

PR #1605, opened 2021-11-15
CLDR-15424 ansible: update upgrade playbook

srl295:
CLDR-15424

### DIFF
--- a/tools/scripts/ansible/upgrade-playbook.yml
+++ b/tools/scripts/ansible/upgrade-playbook.yml
@@ -1,7 +1,21 @@
 # A playbook for upgrading packages
 - hosts: all
+  vars_files:
+    - vars/main.yml
+    - vars/openliberty.yml
+    # - local-vars/local.yml
   become: yes
   tasks:
+  # TODO: should skip these
+  - name: Stop OpenLiberty
+    service:
+      name: "{{ cldr_openliberty_service }}"
+      state: stopped
   - name: Update all packages to the latest version
     apt:
       upgrade: dist
+  - name: Restart OpenLiberty
+    service:
+      name: "{{ cldr_openliberty_service }}"
+      state: restarted
+


tools/scripts/ansible/upgrade-playbook.yml
```diff
@@ -1,7 +1,21 @@
 # A playbook for upgrading packages
 - hosts: all
+  vars_files:
+    - vars/main.yml
+    - vars/openliberty.yml
+    # - local-vars/local.yml
   become: yes
   tasks:
+  # TODO: should skip these
```
btangmu, line 9: I don't understand this "TODO" comment -- if this PR finishes the ticket, why a TODO? What is it that should be skipped?

btangmu (changes requested):

tools/scripts/ansible/upgrade-playbook.yml
srl295: it should only restart them if they are running.

btangmu:
@srl any update on this? BRS "Upgrade server" points to CLDR-15176 which is related to this

srl295:
> @SRL any update on this? BRS "Upgrade server" points to [CLDR-15176](https://unicode-org.atlassian.net/browse/CLDR-15176) which is related to this

This playbook is OK to use, it's just less than optimal because I think it will always restart the liberty server. Since we don't run this all the time, maybe that's OK. 

Do you want to try this change and if it is successful we can merge it?

tools/scripts/ansible/upgrade-playbook.yml
btangmu: looks like it's time for me to try this revised upgrade-playbook on smoketest

btangmu: I already tested it last time (April 25, 2022)! -- old ticket https://unicode-org.atlassian.net/browse/CLDR-15581

srl295:
We probably want to reboot after upgrade. So this might be ok

PR #1608, opened 2021-11-16
CLDR-15101 Support for Ethiopic-Latin (Morse Code) Transliteration

dyacob:
CLDR-15101


srl295:
Is this the right way to represent morse code in Unicode? We might want to consider that a little further.

dyacob:
The most current specification that I can locate for Morse Code is ITU-R M.1677-1 which has an "in force" status.  The recommendation does not address Unicode representation directly, but does appear to use U+002E consistently for "dot", then U+2212 primarily for "dash", but U+2013 also appears.  The authors may have just left it to the word processor to format: 

https://www.itu.int/dms_pubrec/itu-r/rec/m/R-REC-M.1677-1-200910-I!!PDF-E.pdf

srl295:
probably blocked until CLDR-15191 is discussed

dyacob:
@srl295, understood about the blocking, I've pushed in an update that uses `am-Ethi-t-d0-morse` as a forward alias, so the transliteration is ready when CLDR-15191 is ready and can be easily tweaked if needed.

common/transforms/Ethiopic-MorseCode-Amharic.xml
```diff
@@ -0,0 +1,683 @@
...
+7 ↔ '--... ' ;
+8 ↔ '---.. ' ;
+9 ↔ '----. ' ; 
+0 ↔ '----- ' ; 
+
+#
+# TODO: Seek a better way to handle this. The conversion adds a space to the end of every sequence to terminate it.
+#       The CLDR "TestTransforms" unit test will strip off the space at the end of a line in the am-Latn-t-am-Ethi-m0-morse-code.txt
+#       test file thus breaking the comparison with generated words.  The following will strip off the space at the end of
+#       a converted string to fascilitate compatibility with the unit tests.
```
srl295, line 668: it makes some sense to not generate an extra space. However this will make  `73` -> `--... ...--` whereas `7`+`3` = `--......--`

srl295 (commented):
LG so far

common/transforms/Ethiopic-MorseCode-Amharic.xml
dyacob: Converted letters would run together too.  For example "at" is read clearly as ".- -" . With the " " added after ".-"  the conversion would be ".--" which would be misinterpreted as "w" .  The issue is the space between characters, but not between words.  At the end of a word though, the last character still gets the " " which is now superfluous since a word boundary token is naturally expected.

Thinking about it again now though... "/" is the normal word boundary in ASCII form (the conversion of " "), which this transliterator is not supporting.  I should add it if everything else looks ok.

dyacob: @srl295 , the mapping of  `73` -> `--... ...--` , might be OK.  The https://morsedecoder.com/ produces this output (taking a leap of faith here to assume the service is valid).

srl295 (approved):

PR #1611, opened 2021-11-16
CLDR-15151 Tag does not work for Push to Production

btangmu:
-Change the prompt from ...(main, hash, tag) to ...(hash: 40 hex digits)

CLDR-15151


btangmu:
Test failure -- I didn't realize the colon was reserved in that context

srl295 (approved):

PR #1613, opened 2021-11-17
CLDR-15183 move two language names to comprehensive

macchiati:
CLDR-15183

This removes kgp and yrl from the limited-submission release v41, since they are not in scope.

Note: since the regex is a bit complicated, you need to scan back to see where the start is for each of them. However, the test failed before I fixed it, so I'm confident that kgp and yrl were removed.

ALLOW_MANY_COMMITS=true


srl295:
added `ALLOW_MANY_COMMITS=true`

srl295 (approved):

srl295:
@macchiati ready to merge?

PR #1614, opened 2021-11-17
CLDR-15180 Update locales.txt with new locales

AEApple:
CLDR-15180

Update locales.txt with 3 new locales for Google

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
@@ -100,7 +100,8 @@ Google	;	be	;	modern	;	T5	Belarusian
 #Google	;	bem	;	basic	;	Bemba
 Google	;	bg	;	modern	;	T2	Bulgarian
 Google	;	bn	;	modern	;	T3	Bengali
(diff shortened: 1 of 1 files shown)


AEApple:
Why is the Jira ticket number not being recognized?

macchiati (dismissed):

srl295 (dismissed):
LGTM, warnings unrelated

srl295:
> Why is the Jira ticket number not being recognized?

because of the ` - ` hypen after the number.

AEApple:
> > Why is the Jira ticket number not being recognized?
> 
> because of the `-` hypen after the number.

Can you update the comment to put in a warning about that or update the check so it doesn't consider anything following the space as part of the check?

srl295:
> > > Why is the Jira ticket number not being recognized?
> > 
> > 
> > because of the `-` hypen after the number.
> 
> Can you update the comment to put in a warning about that or update the check so it doesn't consider anything following the space as part of the check?

- The ` - ` actually does make the commit message harder to read so I would propose to remove that hyphen.
- I did file https://github.com/unicode-org/jira-github-pr-check/issues/36 to improve the warning message.

srl295 (approved):
LGTM

PR #1616, opened 2021-11-17
CLDR-15161 locale completion API

srl295:
CLDR-15161

- API scaffolding for back end for locale (third) meeting

Front End fixes:
- reload locale progress if locale changes or if there is a vote

Also:
- implementation and tests for StandardCodes.getTargetCoverageLevel()
- implementation and tests for Level.min() and Level.max()

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrGui.js
+++ b/tools/cldr-apps/js/src/esm/cldrGui.js
@@ -416,6 +416,7 @@ function handleResize() {
 }
 
 function updateWithStatus() {
+  cldrProgress.fetchLocaleData(true);  // skip if already loaded
   mainHeaderWrapper?.updateData();
 }
 
--- a/tools/cldr-apps/js/src/esm/cldrProgress.js
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.js
@@ -19,7 +19,7 @@ import { notification } from "ant-design-vue";
 
 const USE_NEW_PROGRESS_WIDGET = true;
 const CAN_GET_VOTER_PROGRESS = true;
-const CAN_GET_LOCALE_PROGRESS = false;
+const CAN_GET_LOCALE_PROGRESS = false;  // For now
 
 let progressWrapper = null;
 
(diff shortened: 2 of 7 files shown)


tools/cldr-apps/js/src/esm/cldrProgress.js
```diff
@@ -330,14 +330,24 @@ function updateVoterStats(votes, total) {
...
+ */
+function fetchLocaleData(unlessLoaded) {
   if (
     !CAN_GET_LOCALE_PROGRESS ||
     !progressWrapper ||
     !cldrStatus.getSurveyUser()
   ) {
     return;
   }
+  if (unlessLoaded && progressWrapper.localeMeter.getPresent()) {
```
btangmu, line 346: maybe equivalent: `if (unlessLoaded && localeProgressStats)`

tools/cldr-apps/js/src/esm/cldrGui.js
```diff
@@ -416,6 +416,7 @@ function handleResize() {
 }
 
 function updateWithStatus() {
+  cldrProgress.fetchLocaleData(true);  // skip if already loaded
   mainHeaderWrapper?.updateData();
 }
```
btangmu, line 421: i think updateWithStatus is only called once per session

another good place for it might be updateCompletionOneVote

but how expensive (time-consuming) is the request? fetching the whole data set too often might be problematic

srl295: updateWithStatus() is called periodically, after what=status fetch.

> another good place for it might be updateCompletionOneVote

Added that in 57c58d4

srl295: > but how expensive (time-consuming) is the request

It should be cached. But in any event, it also won't change very often if the user isn't themselves voting.

tools/cldr-apps/js/src/esm/cldrProgress.js
srl295: done

srl295:
I'd especially like review of  the StandardCodes.getTargetCoverageLevel unit test at  https://github.com/unicode-org/cldr/pull/1616/files#diff-7d5c29b2d7b87ed785f1c91fadce05f86869de4fa99841cc190628a6c55d6d5c unit test

btangmu (approved):

PR #1618, opened 2021-11-17
CLDR-15178 Auto_import internal server error

btangmu:
-Do not try to import old votes in locale und or root

-Encapsulate with new function SurveyAjax.skipLocForImportingVotes

-Also skip loc if getInstance(loc) returns null

CLDR-15178


srl295 (approved):
Ok for today given the schedule, but I think that the haspermission function I mentioned before ought to be used, (modified to skip sandbox)on a per user basis and not just globally. 

I can make a follow up pr

btangmu:
@srl follow-up pr sounds good

under the circumstances i kept this simple

note that already voteForValue throws VoteNotAcceptedException (which we catch) if the user lacks permission

PR #1620, opened 2021-11-17
CLDR-15056 Coverage Progress Bars, bug fix for Section rows

btangmu:
-Skip row if json.canModify is false or statusAction.vote is false

-Comments

CLDR-15056


btangmu:
This should fix Section Completion (first meter), skipping rows the user can't vote for. It only changes the front end.

To fix Voting Completion (second meter) will require changes to the back end...

srl295 (approved):

PR #1623, opened 2021-11-19
CLDR-15056 Coverage Progress Bars, hide 2nd meter, fix dashboard reload

btangmu:
-Hide the 2nd meter (Voter Completion) until the back end works

-Do not try to fetch data for 2nd or 3rd meter until the back end works

-New booleans in cldrProgress.js: CAN_GET_VOTER_PROGRESS, CAN_GET_LOCALE_PROGRESS

-Fix bug: Dashboard would reload after all-row getrow request though coverage unchanged

-In cldrMenu.js, do not call cldrLoad.coverageUpdate unless effectiveName changes

-In DashboardWidget.vue, do not call this.reloadDashboard unless this.level changes

CLDR-15056

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrProgress.js
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.js
@@ -17,6 +17,8 @@ import { createCldrApp } from "../cldrVueRouter";
 import { notification } from "ant-design-vue";
 
 const USE_NEW_PROGRESS_WIDGET = true;
+const CAN_GET_VOTER_PROGRESS = false;
+const CAN_GET_LOCALE_PROGRESS = false;
 
 let progressWrapper = null;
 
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/js/src/esm/cldrProgress.js
```diff
@@ -17,6 +17,8 @@ import { createCldrApp } from "../cldrVueRouter";
 import { notification } from "ant-design-vue";
 
 const USE_NEW_PROGRESS_WIDGET = true;
+const CAN_GET_VOTER_PROGRESS = false;
+const CAN_GET_LOCALE_PROGRESS = false;
```
srl295, line 21: why are these constants? I don't see the utility of `CAN_GET_LOCALE_PROGRESS = false;` when it's already hidden if it fails…

srl295 (approved):
LGTM though not clear why th e constants are there

tools/cldr-apps/js/src/esm/cldrProgress.js
btangmu: that's to avoid needless http requests (for performance) and browser console error messages

also a switch is needed at least for VOTER progress since otherwise the back end does return data but it's inaccurate

srl295: hm… OK. if that's a problem then i'm ok with this

btangmu: thanks!

PR #1628, opened 2021-11-20
CLDR-15056 Coverage Progress Bars, refactor VettingViewer again

btangmu:
-Avoid using an entire FileInfo to combine counters from multiple locales for Priority Items Summary

-New class VettingCounters with only the parts of FileInfo needed for combining counters

CLDR-15056


btangmu:
I have further changes locally now that go beyond refactoring. Dashboard json data has new votedPathCount and votablePathCount (aka votes and total) for the 2nd meter (Voter Progress = % completion for this voter in this locale) so we don't need the VotingCompletion api. I just need to revise the front end accordingly, which should be relatively simple...

srl295 (approved):

PR #1629, opened 2021-11-21
CLDR-15165 add contrasting case examples for units

macchiati:
CLDR-15165

Previous PRs added contrasting examples for minimal pairs; this adds them for the units in different case forms.

There is a bit of refactoring also, partly for clarity, partly to be able to have 2 locales among the tests.
Also, some of the 'informational warnings' are now under an environment flag to avoid clutter in the tests.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -85,6 +85,9 @@ import com.ibm.icu.util.ULocale;
  * @author markdavis
  */
 public class ExampleGenerator {
+    private static final String EXAMPLE_OF_INCORRECT = "❌  ";
+    private static final String EXAMPLE_OF_CAUTION = "⚠️  ";
+
     private static final boolean DEBUG_EXAMPLE_GENERATOR = false;
 
     final static boolean DEBUG_SHOW_HELP = false;
@@ -679,7 +682,7 @@ public class ExampleGenerator {
             sample = bestMinimalPairSamples.getBestUnitWithGender(gender, output);
             String otherGender = getOtherGender(locale, gender);
             sampleBad = bestMinimalPairSamples.getBestUnitWithGender(otherGender, output);
-           break;
+            break;
         default:
             return null;
         }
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -747,18 +750,22 @@ private String handleFormatUnit(XPathParts parts, String value) {
...
-        if (amount != null) {
-            numberFormat = icuServiceBuilder.getNumberFormat(1);
-            examples.add(format(value, backgroundStartSymbol + numberFormat.format(amount) + backgroundEndSymbol));
+        if (amount == null) {
+            return null;
         }
+        DecimalFormat numberFormat = null;
+        String formattedAmount = null;
+        numberFormat = icuServiceBuilder.getNumberFormat(1);
+        formattedAmount = numberFormat.format(amount);
```
btangmu, line 759: does the setting these to null right before setting them to something else serve a purpose?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -769,16 +776,39 @@ private String handleFormatUnit(XPathParts parts, String value) {
...
         }
         return formatExampleList(examples);
     }
 
+    private String getConstrastingCase(String unitPattern, String gCase, Collection<String> unitCaseInfo, XPathParts parts) {
+        for (String otherCase : unitCaseInfo) {
+            if (otherCase.equals(gCase)) {
+                continue;
+            }
+            parts.putAttributeValue(-1, "case", "nominative".equals(otherCase) ? null : otherCase);
```
btangmu, line 803: could constants be used instead of literal "case" and "nominative"? maybe enums rather than just strings?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Both good comments. Will handle in followup pr

PR #1630, opened 2021-11-22
CLDR-15165 2nd batch from minimal review

macchiati:
CLDR-15165


macchiati:
Fixed the bug in Slovak, so ready for review (sorry, was in meetings all afternoon).

AEApple (approved):
LGTM

Note: It would be easier to review if we were consistent about the ordering of the cases especially between the change and the review sheet (e.g. nominative is last in sk, but first in sv). It would be nice to clean up in the future to have consistent punctation and '...' when the minimal pair is part of a sentence fragment.

macchiati:
Good points. Could you file a ticket?

Mark

AEApple:
Done. [CLDR-15196](https://unicode-org.atlassian.net/browse/CLDR-15196)

PR #1631, opened 2021-11-23
CLDR-15056 Coverage Progress Bars, enable Voter Progress meter

btangmu:
-CAN_GET_VOTER_PROGRESS = true in cldrProgress.js

-Fetch Voter Progress data and Dashboard data with the same request

-Implement Voter Progress back end as an add-on to the Dashboard back end FileInfo, etc

-New java classes DashboardData and VoterProgress

-Add new method userDidVote to UsersChoice interface

-Delete api/VotingCompletion.java since request is now handled through api/Summary.java

-Some renaming (java): T organization, not T user, for T in VettingViewer

-Revise GUI to show Voter meter (2nd meter) only when Dashboard is visible

-Omit numeric percentage in meters (showInfo=false) to work around ant-design-vue title (hover) bug

-Remove redundant call to cldrProgress.refresh from cldrGui.refreshCounterVetting

-Some renaming (js): 1st meter = page, not section; 2nd meter = voter, 3rd = locale (all-vetters)

-Add a-tooltip component from ant-design-vue to improve visibility of meter hover text

-Work around the limitation that a-tooltip converts newlines into spaces: add bullets

-Comments

CLDR-15056


btangmu:
see https://github.com/unicode-org/cldr/pull/1632

PR #1640, opened 2021-12-01
CLDR-14926 st: update server.xml jdkSourceLevel

srl295:
set to at least 1.8

CLDR-14926


srl295:
Already applied this to `/var/lib/openliberty/usr/servers/cldr/server.xml` on smoketest, can apply to production also once verified.

btangmu (approved):

PR #1641, opened 2021-12-01
CLDR-15196 common order for grammatical features

macchiati:
CLDR-15196

### DIFF
--- a/common/annotations/dsb.xml
+++ b/common/annotations/dsb.xml
@@ -832,7 +832,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="🧑‍⚕">chórobny sotša | gójc | gójc/gójcowka | strowotnistwo | terapeut</annotation>
 		<annotation cp="🧑‍⚕" type="tts">gójc/gójcowka</annotation>
 		<annotation cp="👨‍⚕">doktor | gójc | gójcojski kitel</annotation>
-		<annotation cp="👨‍⚕" type="tts"/>
+		<annotation cp="👨‍⚕" type="tts">doktor</annotation>
 		<annotation cp="👩‍⚕">gójcojski kitel | gójcowka</annotation>
 		<annotation cp="👩‍⚕" type="tts">gójcowka</annotation>
 		<annotation cp="🧑‍🎓">absolwent | student | student(ka)</annotation>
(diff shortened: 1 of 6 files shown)


common/annotations/dsb.xml
```diff
@@ -832,7 +832,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="🧑‍⚕">chórobny sotša | gójc | gójc/gójcowka | strowotnistwo | terapeut</annotation>
 		<annotation cp="🧑‍⚕" type="tts">gójc/gójcowka</annotation>
 		<annotation cp="👨‍⚕">doktor | gójc | gójcojski kitel</annotation>
-		<annotation cp="👨‍⚕" type="tts"/>
+		<annotation cp="👨‍⚕" type="tts">doktor</annotation>
```
srl295, line 835: Is this the TTS err issue we have been seeing only sometimes?

macchiati:
This is ready for review, Annemarie. I cleaned up the ordering a bit, which you can see from tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java, and also added the reasoning behind the ordering.

(Just noticed a typo in the comments and will fix that.)

srl295 (approved):
DUgCET - the DUCET for grammar

AEApple:
Can we add a note in the Survey Tool about the ordering of the cases to inform the vetters about the ordering?

macchiati:
Yes, whenever we push to production we need to make sure that we update the
documentation (and send out a note pointing to the change). We'll want to
try it out on smoketest first, of course!

Mark

macchiati:
Yes, common/annotations/dsb.xml
<https://github.com/unicode-org/cldr/pull/1641#discussion_r761416911>: was
for the test failure (other failures are hidden in non-exhaustive mode).

Mark

macchiati:
If someone reviews, then I can start checking it out on Smoketest.

srl295 (approved):

macchiati:
Gakk, had to squash because of a bad commit message... so asking for another review. (Really wish squashing didn't require rereview.)

btangmu (approved):

srl295 (approved):

srl295:
> Yes, common/annotations/dsb.xml <[#1641 (comment)](https://github.com/unicode-org/cldr/pull/1641#discussion_r761416911)>: was for the test failure (other failures are hidden in non-exhaustive mode). Mark
> […](#)

*ideally* that would be in the commit message also.

macchiati:
I merged this in and checked it out on Smoketest. Noted some problems:
https://unicode-org.atlassian.net/browse/CLDR-15196?focusedCommentId=163433
Mark

PR #1642, opened 2021-12-02
CLDR-15191 add a test for -t-d0 etc. as a warning

srl295:
CLDR-15191

It turns out there was no test to verify that `-t-m0`, etc. values were valid. As an example there is `latn-t-ethi-m0-ala` that should be `…-m0-alalc`.

As there are existing errors, this only introduces a *warning* but in the future these should be cleaned up


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTransforms.java
```diff
@@ -197,11 +205,35 @@ private String makeLegacyTransformID(String source, String target, String varian
...
+                    final String tExt = splits[i];
+
+                    if (tKeys.contains(tExt)) {
+                        if (i + 1 >= splits.length) {
+                            errln(file.getName() + ": BCP47-T identifier " + tExt + " at end of line: " + id);
+                        } else {
+                            final String tVal = splits[++i];
+                            Set<String> possibleVals = supplementalDataInfo.getBcp47Keys().get(tExt);
+                            if (!possibleVals.contains(tVal)) {
+                                warnln(file.getName() + ": BCP47-T undeclared key -t-" + tExt + "-" + tVal + " in: " + id
```
srl295, line 228: just a warning for now

macchiati (changes requested):
There is a thorough validity checker somewhere else. that should be used.

srl295:
> There is a thorough validity checker somewhere else. that should be used.

Hmm. Do you know where ? It didn't seem to fail on invalid bcp47 tags.

srl295:
> > There is a thorough validity checker somewhere else. that should be used.
> Hmm. Do you know where ? It didn't seem to fail on invalid bcp47 tags.

@macchiati  did not find another validity checker, do you know where it is?

PR #1649, opened 2021-12-03
CLDR-15212 Open nn for v41

btangmu:
-Update grammaticalFeatures.xml to include nn

CLDR-15212


stenshamn (approved):

macchiati (approved):
This looks pretty safe.

PR #1650, opened 2021-12-06
CLDR-14925 Preliminary refactor for snapshot of Priority Items Summary

btangmu:
-Use HeaderParam(Auth.SESSION_HEADER) instead of ss parameter for api/summary

-Revise cldrAjax.doFetch to set session header even if other headers are already present

-Simplify getVettingViewerOutput, rename it getPriorityItemsSummaryOutput

-Remove dead code, unused parameters

-Make private where possible

-Comments

CLDR-14925

### DIFF
--- a/tools/cldr-apps/js/src/views/VettingSummary.vue
+++ b/tools/cldr-apps/js/src/views/VettingSummary.vue
@@ -13,68 +13,76 @@
 </template>
 
 <script>
+import * as cldrAjax from "../esm/cldrAjax.js";
+
+const SECONDS_IN_MS = 1000;
+
+// timeouts
+const RETRY_ON_FETCH_ERR = 15 * SECONDS_IN_MS; // When we couldn't talk to server
+const NORMAL_RETRY = 5 * SECONDS_IN_MS; // "Normal" retry: starting or about to start
+const BUSTED_RETRY = 60 * SECONDS_IN_MS; // ST is down
+
+const SUMMARY_URL = "api/summary";
+
 export default {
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/js/src/views/VettingSummary.vue
```diff
@@ -13,68 +13,76 @@
...
 
 <script>
+import * as cldrAjax from "../esm/cldrAjax.js";
+
+const SECONDS_IN_MS = 1000;
+
+// timeouts
+const RETRY_ON_FETCH_ERR = 15 * SECONDS_IN_MS; // When we couldn't talk to server
+const NORMAL_RETRY = 5 * SECONDS_IN_MS; // "Normal" retry: starting or about to start
+const BUSTED_RETRY = 60 * SECONDS_IN_MS; // ST is down
```
srl295, line 23: maybe these should go into `cldrAjax.js`?

srl295: think these constants are in a couple of files now

srl295 (approved):

tools/cldr-apps/js/src/views/VettingSummary.vue
btangmu: that might be good -- note btw that 5 seconds is pretty frequent; compare the all-row get row which is only every 15 seconds; also i made a note in today's agenda about a bug where the complete data is getting re-downloaded every 5 seconds, redundantly

PR #1651, opened 2021-12-07
CLDR-8865 run exhaustive weekly, add a button for manual

srl295:
CLDR-8865

- This adds a manual button
- … and also runs weekly Tuesday mornings (so that the results will be ready to review Wednesday)

### DIFF
--- a/.github/workflows/exhaustive.yml
+++ b/.github/workflows/exhaustive.yml
@@ -1,0 +1,61 @@
+# Exhaustive Mode
+
+name: exhaustive
+
+on:
+  schedule:
+    - cron: '1 1 * * 2'  # Tuesday morning
+  workflow_dispatch:
+    inputs:
+      git-ref:
+        description: 'Git Ref to Build'
+        required: true
+        default: main
+jobs:
+  build:
+    name: CLDR Build with Exhaustive Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Repository
+        uses: actions/checkout@v2
+        with:
+          lfs: false
+        if: github.event.inputs.git-ref == ''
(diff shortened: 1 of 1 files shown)


srl295:
Example run: https://github.com/srl295/cldr/runs/4436693977?check_suite_focus=true (name is wrong 'build for production).

Does this look like the errors expected?

- [ ] fix title of run
- [ ] bring in error / warning highlighting
- [ ] cron - run weekly

.github/workflows/exhaustive.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Build with Maven
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B compile install package --file tools/pom.xml -pl cldr-code
+          -DskipTests=true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Test with maven
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B test --file tools/pom.xml
+          -pl cldr-code -Dorg.unicode.cldr.unittest.testArgs='-e10 -n'
```
btangmu, line 53: is "-e10" the main thing that's different here from the "build" part of production.yml?

it would be nice if the common parts could be encapsulated rather than duplicated -- but I'm not sure how to re-use code in workflows...

srl295: This does fewer steps than the others. Yes we can try to refactor, but there are more divergent changes that will happen with this workflow.

srl295: Workflows are extremely complicated to debug. I'd rather not prematurely refactor.

btangmu (approved):

PR #1657, opened 2021-12-08
CLDR-15222 hide sandbox in production, identify

srl295:
- on production, hide the sandbox locales
- on smoketest/dev, show sandbox locales in a separate section

CLDR-15222

-----

On smoketest it will look like the following, on production will be hidden:

image

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrMenu.js
+++ b/tools/cldr-apps/js/src/esm/cldrMenu.js
@@ -87,12 +87,7 @@ function loadInitialMenusFromJson(json) {
   theDiv.className = "localeList";
 
   // TODO: avoid duplication of some of this code here and in cldrLocales.js
-  addTopLocale("root", theDiv);
-  for (let n in locmap.locmap.topLocales) {
-    const topLoc = locmap.locmap.topLocales[n];
-    addTopLocale(topLoc, theDiv);
-  }
-  $("#locale-list").html(theDiv.innerHTML);
+  addTopLocales(theDiv, locmap);
 
   if (cldrStatus.isVisitor()) {
     $("#show-read").prop("checked", true);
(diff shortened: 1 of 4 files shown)


srl295:
@macchiati  comments:

> Yes, in that ticket the *name* of it needs to be "Test Locales", not
> "Multiple Languages"

- [ ] TODO: update the name elsewhere

tools/cldr-apps/js/src/esm/cldrMenu.js
```diff
@@ -90,7 +90,22 @@ function loadInitialMenusFromJson(json) {
...
+    $(theDiv).append($("<hr/><h4>Test Locales</h4>"));
+    for (let n in locmap.locmap.topLocales) {
+      const topLoc = locmap.locmap.topLocales[n];
+      const topLocInfo = locmap.getLocaleInfo(topLoc);
+      if (topLocInfo.special_type === "scratch") {
+        // Now only Sandbox locales here
+        addTopLocale(topLoc, theDiv);
+      }
+    }
   }
```
btangmu, line 109: This function needs subroutines -- adding 15 lines makes it longer than it should be

btangmu (dismissed):
lgtm except adding a subroutine would be better

srl295:
## Fixed in sidebar

image

## The top page still says:

image

… and there will be a popup message with this text.

tools/cldr-apps/js/src/esm/cldrMenu.js
srl295: @btangmu that could be good, but I think even better would be modernizing the locale list sidebar.

btangmu (approved):

srl295:
There's a bug where `Test Locales` stays in the sidebar even if a different locale is chosen.

PR #1661, opened 2021-12-13
CLDR-15056 Coverage Progress Bars, enable Locale Progress meter

btangmu:
-CAN_GET_LOCALE_PROGRESS = true in cldrProgress.js

-Ensure that Page, Voter, Locale meters are displayed as 1st, 2nd, 3rd respectively

-If Page meter is visible, make Voter meter visible even if no Dashboard data yet

-If Voter meter is visible but no data yet, display with special message and appearance

-TestCldrChecksum.js: warn without failure if time is excessive

-Format unrelated cldrLoad.js by running sh tools/cldr-apps/js/test/pretty.sh

CLDR-15056

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLoad.js
+++ b/tools/cldr-apps/js/src/esm/cldrLoad.js
@@ -416,7 +416,7 @@ function insertLocaleSpecialNote(theDiv) {
  * @param {Boolean} brief if true, keep it short
  * @param {Boolean} plain if true, strip to plain text
  */
- function localeSpecialNote(bund, brief) {
+function localeSpecialNote(bund, brief) {
   if (!bund) return null;
   let msg = null;
   if (bund.dcParent) {
--- a/tools/cldr-apps/js/src/esm/cldrProgress.js
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.js
@@ -19,7 +19,7 @@ import { notification } from "ant-design-vue";
 
 const USE_NEW_PROGRESS_WIDGET = true;
 const CAN_GET_VOTER_PROGRESS = true;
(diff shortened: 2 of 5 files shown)


btangmu:
image

tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -416,7 +416,7 @@ function insertLocaleSpecialNote(theDiv) {
  * @param {Boolean} brief if true, keep it short
  * @param {Boolean} plain if true, strip to plain text
  */
- function localeSpecialNote(bund, brief) {
+function localeSpecialNote(bund, brief) {
```
srl295, line 419: ✨

tools/cldr-apps/js/src/esm/cldrProgress.js
```diff
@@ -44,6 +44,7 @@ class MeterData {
     this.votes = votes || 0;
     this.total = total || 0;
     this.level = level || cldrText.get("coverage_unknown");
+    this.exceptional = false;
```
srl295, line 47: would 'inactive' be a better term here?

srl295 (approved):
all makes sense. 'exceptional' may be the best, because the meter is active but has some error.

tools/cldr-apps/js/src/esm/cldrProgress.js
btangmu: maybe.... I was starting to implement it with "status=exception" (per https://ant.design/components/progress/ ) and that's what inspired me to call it "exceptional", but then I ended up using "showInfo=false" (also  per https://ant.design/components/progress/  )

PR #1666, opened 2021-12-15
CLDR-15247 Adding invariants to list patterns

robertbastian:
CLDR-15247

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -2165,13 +2165,40 @@ List patterns can be used to format variable-length lists of things in a locale-
 </listPatterns>
 ```
 
-The data is used as follows: If there is a type type matches exactly the number of elements in the desired list (such as "2" in the above list), then use that pattern. Otherwise,
+Each pattern satisifies the following conditions:
+<ul>
+    <li>it contains the placeholders <code>{0}</code>, <code>{1}</code>, and <code>{2}</code> ("3"-pattern only) in order</li>
+    <li>"start" and "middle" patterns end with the <code>{1}</code> placeholder</li>
+    <li>"middle" and "end" patterns begin with the <code>{0}</code> placeholder</li> 
+</ul>
 
-1.  Format the last two elements with the "end" format.
-2.  Then use middle format to add on subsequent elements working towards the front, all but the very first element. That is, {1} is what you've already done, and {0} is the previous element.
-3.  Then use "start" to add the front element, again with {1} as what you've done so far, and {0} is the first element.
+That is,
+<ul>
+    <li>all patterns can have text between the placeholders</li>
+    <li>only the "start", "2", and "3" patterns can have text before the first placeholder, and</li>
+    <li>only the "end", "2", and "3" patterns can have text after the last placeholder.</li>
+</ul>
 
-Thus a list (a,b,c,...m, n) is formatted as: `start(a,middle(b,middle(c,middle(...end(m, n))...)))`
+The data is used as follows: If there is a type that matches exactly the number of elements in the desired list (such as "2" in the above list), then use that pattern. Otherwise,
+
+1.  Format the last two elements with the "end" pattern.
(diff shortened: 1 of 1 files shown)


sffc (dismissed):
Nicely done!

sffc (approved):

docs/ldml/tr35-general.md
```diff
@@ -2165,13 +2165,34 @@ List patterns can be used to format variable-length lists of things in a locale-
...
 ```
 
-The data is used as follows: If there is a type type matches exactly the number of elements in the desired list (such as "2" in the above list), then use that pattern. Otherwise,
+Each pattern satisifies the following conditions:
+<ul>
+    <li>it contains both a <code>{0}</code> and a <code>{1}</code> placeholder</li>
+    <li>the <code>{0}</code> placeholder appears before the <code>{1}</code> placeholder</li>
+    <li>"start" and "middle" patterns end with the <code>{1}</code> placeholder</li>
+    <li>"middle" and "end" patterns begin with the <code>{0}</code> placeholder</li> 
+</ul>
```
macchiati, line 2174: Add clarifying sentence, something like: 

That is, 
* all patterns can have text between the placeholders
* only the start and '2' patterns can have literal text after the placeholders, and 
* only the end and '2' patterns can have literal text before the placeholders.

docs/ldml/tr35-general.md
```diff
@@ -2165,13 +2165,34 @@ List patterns can be used to format variable-length lists of things in a locale-
...
+
+1. Format the first two elements with the "start" pattern.
+2. Then use the "middle" pattern to add on subsequent elements working towards the back, all but the very last element. That is, `{0}` is what you've already done, and `{1}` is the next element.
+3. Then use "end" to add the last element, again with `{0}` as what you've done so far, and `{1}` is the last element.
+
+Here, the list (a,b,c,...m, n) is formatted as:  `end(middle(..., middle(start(a, b), c) ...) m) n) `. While this prefix-expression looks less suitable, it actually only requires appends,
+so this algorithm can be used to write into append-only sinks. Both the back-to-front and the front-to back algorithm produce this expression:
+
+```
+start_before + a + start_between + b + middle_between + c + ... + middle_between + m + end_between + n + end_after
```
macchiati, line 2192: might be nice to highlight the substitutions (eg with bold): a, b, c, m, n. for clarity

docs/ldml/tr35-general.md
```diff
@@ -2165,13 +2165,34 @@ List patterns can be used to format variable-length lists of things in a locale-
...
+```
+
+where the patters are "start": `start_before{0}start_between{1}`, "middle": `{0}middle_between{1}`, and "end": `{0}end_between{1}_end_after`.
```
macchiati, line 2195: _end_after => end_after 
(extra underbar)

macchiati (changes requested):
looks pretty good; a few nits

docs/ldml/tr35-general.md
robertbastian: I don't think I can use formatting in a code block

macchiati:
Good point. But '3' would behave like '2', so we would cover it in the same
way. (No composition is necessary, it is just a plain
placeholder substitution.)
Mark

docs/ldml/tr35-general.md
robertbastian: ^ that's actually the wrong way around 😛

robertbastian:
In the clarifying segment I'd have to call out the "3" pattern as allowing text before and after the placeholders, but I'm not sure we want to draw attention to its existence. The invariants hold without having to mention it.

sffc:
Presumably the 3 pattern would have placeholders 0, 1, and 2.  Seems easy enough to make it a special case.

macchiati:
Yes, just like the 2 case.

Mark (📱)

sffc (approved):
LGTM

docs/ldml/tr35-general.md
macchiati: ok

macchiati (approved):

PR #1671, opened 2021-12-22
CLDR-15161 3rd meter: correct error count

srl295:
- send in dpath (not fullpath) to check(), avoid spurious collision errors
- also add some more debugging values to the response payload

CLDR-15161

- [X] This PR completes the ticket. (hopefully!)

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
@@ -142,25 +142,34 @@ public class LocaleCompletion {
 
         final List<CheckStatus> results = new ArrayList<>();
         for (final String xpath : file) {
+            lcr.allXpaths ++;
             final Level pathLevel = covLeveller.getLevel(xpath);
             final String fullPath = file.getFullXPath(xpath);
 
             final PathHeader ph = LocaleCompletionHelper.INSTANCE.phf.fromPath(xpath);
             SurveyToolStatus surveyToolStatus = ph.getSurveyToolStatus();
             if (surveyToolStatus == SurveyToolStatus.DEPRECATED || surveyToolStatus == SurveyToolStatus.HIDE) {
+                lcr.ignoredHidden ++;
                 continue; // not visible
             }
 
-            checkCldr.check(fullPath, results, file.getStringValue(xpath));
+            checkCldr.check(xpath, results, file.getStringValue(xpath));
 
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
```diff
@@ -142,25 +142,34 @@ static LocaleCompletionResponse getLocaleCompletion(final CLDRLocale cldrLocale,
 
         final List<CheckStatus> results = new ArrayList<>();
         for (final String xpath : file) {
+            lcr.allXpaths ++;
```
btangmu, line 145: unusual space before "++" 

i wish we were using similar formatting tools for java as we're using for js

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
```diff
@@ -193,6 +205,13 @@ static LocaleCompletionResponse getLocaleCompletion(final CLDRLocale cldrLocale,
         public int provisional = 0;
         final public String level;
 
+        // The following are more or less debug items
+        public int ignoredLimited = 0;
+        public int ignoredHidden = 0;
+        public int allXpaths = 0;
+        public int statusMissing = 0;
+        public int ignoredOutOfCov = 0;
```
btangmu, line 213: it might be worthwhile to make a stronger distinction between debug and non-debug items

is this data included in json, and if so, is it actually used/needed on the front end?

could these be private?

on the one hand, it can be convenient and flexible like this to be public (and maybe included in json), however in the long run there's a risk of cruft, dead code, and unnecessary dependencies

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
```diff
@@ -178,7 +187,10 @@ static LocaleCompletionResponse getLocaleCompletion(final CLDRLocale cldrLocale,
                     } else {
                         lcr.addOk();
                     }
-                } // else: out of coverage level, do not count the path.
+                } else {
+                    // out of coverage level, do not count the path.
+                    lcr.ignoredOutOfCov ++;
```
btangmu, line 192: there's that space plus plus again

btangmu (approved):

srl295:
Merging so we can get eyes on the functionality.. will leave ticket open to revisit formatting (thanks)

PR #1673, opened 2021-12-23
CLDR-15101 ALA → ALALOC and New Transform Identifiers

dyacob:
CLDR-15101

This PR changes the mistaken "ALA" identifier to "ALALOC" in the respective transliteration.  The d0/s0 identifier "morse" is added to the BCP 47 codes as well as the new variants introduced under CLDR-15101.

- [ ] This PR completes the ticket.

PR #1674, opened 2022-01-01
CLDR-15101 Preemptive Gemination

dyacob:
CLDR-15101

This *should* be the final PR oF CLDR-15101 and should also be merged following #1673.


dyacob:
I'm a bit stumped by the `"Caused by: java.lang.IllegalArgumentException: Illegal ID Ethiopic-Gemination"` error that's appearing.  I've cloned the repo into an entirely new git root to have a fresh start and I am not encountering the problem.

btangmu:
@dyacob i just saw this pr -- it's not assigned to anyone yet -- are you able to request a review from me? If not, I can assign it to myself anyway -- for the future if you can request a review from me (or someone) then it's more likely to come to our attention sooner

dyacob:
@btangmu I can't request reviews, self assigning would be great. thanks!

common/bcp47/transform-destination.xml
```diff
@@ -30,6 +30,7 @@
 			<type name="null" description="make no change in the string" since="29"/>
 			<type name="remove" description="remove every character in the string" since="29"/>
 			<type name="zawgyi" description="map Unicode to Zawgyi Myanmar encoding" since="34"/>
+			<type name="morse" description="map Unicode to Morse Code encoding" since="41"/>
```
btangmu, line 33: Is the "Morse" part right for this PR or should it be with the other PR?

dyacob: Adding the "morse" identifier addition to this request was mostly a matter of convenience since it was overlooked in the earlier PR.

On Ethiopic-Gemination, it would follow the **S-T** ID pattern and would be formed from the respective attributes from the Ethiopic-Gemination.xml file

`<transform source="Ethiopic" target="Gemination" direction="forward" visibility="internal">`

My guess is that the "Ethiopic-Gemination" identifier was not registered ahead of its first reference by another transliteration file (like Ethiopic-Latin-Aethiopica_Auto_Geminated.xml) which depends on it.  In this case, why the identifier doesn't get registered in time is a mystery.

The approach with the "Preemptive Gemination" files follows the approach used by the "InterIndic" conversions (and others) where "InterIndic" and "Gemination" are just used as an intermediate conversion step (thus the "internal" visibility flag).

Yes, I am open to splitting the PR into smaller PRs.

PR #1680, opened 2022-01-07
CLDR-15273 en-AU: Make abbreviated months consistent

Quppa:
Fixes CLDR-15273.


mattsmacblog:
Could this change be applied to en-NZ as well?  to get around old software problems? explained here: https://unicode-org.atlassian.net/browse/CLDR-19585 & example problem software here:  https://help.idealpos.co/windows-regional-setting-change-september-abbreviation-issue-in-nz-locale

PR #1685, opened 2022-01-12
CLDR-15277 VettingParticipation2: err on out of bound locales

srl295:
Note this is on the generated Locales.txt viewable through the admin panel

- add 'coverageLocales' to each org with a list of orgs with cov
- add an `ERROR` note to the generated Locales.txt when vetters are assigned out of bounds
- add a note for orgs with a `*` coverage with out of bound vetters

CLDR-15277

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingParticipation.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingParticipation.java
@@ -194,11 +194,14 @@ public class VotingParticipation {
 
     public static class OrgParticipation {
         public String[] localesWithVetters;
+        final public String[] coverageLocales;
         public String defaultCoverage;
 
         public OrgParticipation(Organization o) {
             this.org = o;
-            this.defaultCoverage = StandardCodes.make().getDefaultLocaleCoverageLevel(o).toString();
+            StandardCodes codes = StandardCodes.make();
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VotingParticipation.java
```diff
@@ -194,11 +194,14 @@ public void put(int count, String locale, String org) {
...
         public String[] localesWithVetters;
+        final public String[] coverageLocales;
         public String defaultCoverage;
 
         public OrgParticipation(Organization o) {
             this.org = o;
-            this.defaultCoverage = StandardCodes.make().getDefaultLocaleCoverageLevel(o).toString();
+            StandardCodes codes = StandardCodes.make();
+            this.defaultCoverage = codes.getDefaultLocaleCoverageLevel(o).toString();
+            coverageLocales = codes.getLocaleCoverageLocales(o).toArray(new String[0]);
```
macchiati, line 204: Minor, but if this doesn't change over time, best to create an immutable set.

srl295: @macchiati  ordinarily yes. in this case it's actually a string array for the purposes of shipping this class over as JSON.  I think `StandardCodes…getLocaleCoverageLocales()` actually does create an immutable set here.

macchiati:
ok
Mark

macchiati (approved):

PR #1691, opened 2022-01-19
Cldr 15291 remove invalid and duplicate files

pedberg-icu:
CLDR-15291

- [ ] This PR completes the ticket.

This fixes some problems in the brs/v41data branch for CLDR 41 data import:
* Removes the mul*.xml files, which are for a sandbox and should not be checked into CLDR
* Removes the seed/main versions of kgp.xml/kgp_BR.xml/sc.xml/sc_IT.xml in favor of the common/main versions; the latter either have identical contents of (in the case of kgp.xml) are much more complete.

PR #1692, opened 2022-01-19
CLDR-15291 Remove invalid mul* files and duplicate kgp/sc files

pedberg-icu:
CLDR-15291

This fixes some problems in the brs/v41data branch for CLDR 41 data import:
* Removes the mul*.xml files, which are for a sandbox and should not be checked into CLDR
* Removes the seed/main versions of kgp.xml/kgp_BR.xml/sc.xml/sc_IT.xml in favor of the common/main versions; the latter either have identical contents of (in the case of kgp.xml) are much more complete.


srl295:
@pedberg-icu note comment in #1691  - reopen PR wasn't needed. wil review though

srl295 (approved):

macchiati (approved):

PR #1694, opened 2022-01-20
CLDR-15248 Data fix

robertbastian:
CLDR-15248

These are the required data fixes only. New checks also need to be added.


robertbastian:
@macchiati FYI

robertbastian:
Superseded by #1747

PR #1709, opened 2022-01-24
CLDR-14665 add 'two' to Maltese plurals

srl295:
- plus clone 'few' -> 'two' to pass tests (no change
in text output)

CLDR-14665

I know that there are plurals for some of these (hence the ticket) for example:

> count=two `{0} snin ilu` ('2 years ago')

should be

> `sentejn ilu` ('two years ago')

similarly `jumejn` two days

But, since vetting is closed, I am just cloning `few` to `two` for now.


macchiati (changes requested):
Changes to the plural rules can be disruptive. 

Making this change without actually changing the data has the downside of the disruption, without the upside of better quality. So I think we need to delay to v42.

srl295:
> Changes to the plural rules can be disruptive.
> 
> Making this change without actually changing the data has the downside of the disruption, without the upside of better quality. So I think we need to delay to v42.

Are there specific changes you have in mind or just to delay this merging?

srl295:
to fix:

```
Error:  (TestSupplementalInfo.java:648)  Error: : mt, 2, two: expected true, got false
Error:  (TestSupplementalInfo.java:648)  Error: : mt, 3, two: expected true, got false
Error:  (TestSupplementalInfo.java:648)  Error: : mt, 4, two: expected true, got false
```

macchiati:
I just want to not have it in v41. After the v41 branch is created, then it could go in. 

But we *don't* want to put in fake values, because then users won't see that values are missing. We also need to have the minimal pairs before vetting starts.

srl295:
It’s not fake, but copies of few. So it will look just as it did before,
and many of them will be correct that way.

 I know some of the correct values such as xahrejn, jumejn etc but better
to get experts help.

I’ll get at least minimal pairs though.

macchiati:
Hmmm. We might want to mark these with a draft status so that the translators see them.

srl295:
@macchiati  @pedberg-icu  why does this require a placeholder:

> Error: Need at least 1 placeholder(s), but only have 0. Placeholders are: {{0}={NUMBER_OF_YEARS}, e.g. “3”}; see <a href='http://cldr.unicode.org/translation/error-codes#missingPlaceholders'  target='cldr_error_codes'>missing placeholders</a>.

```xml
					<relativeTimePattern count="two">fi żmien sentejn oħra</relativeTimePattern> <!-- n=2 -->
```

it means literally "in two years further".  `sentejn` contains the idea of 2-ness. 

Ha… Arabic doesn't have a placeholder.    Is it a special exception?  ( the arabic here sounds just like "sentejn" )

```xml
					<relativeTimePattern count="two">خلال سنتين</relativeTimePattern>
```

srl295:
@macchiati  @pedberg-icu  been banging my head on test failures…  if i'm reading this correctly, 20141a410dbb172997a6e9895bc19f248f3bbff0 CLDR actually DOES NOT use plurals.xml as its source of data, but only pulls from ICU.  So you have to regenerate ICU4J in order to update plural rules. Is this correct?

macchiati:
No, that should not be the case. We have made changes in the rules and have
had them reflected in the ST well before they have gone into ICU. One thing
you might be missing is that you have to rerun the plural formatting — that
regenerates the samples, which are actually crucial to having CLDR (and
ICU) work properly.

The case of no placeholders does happen in a few locales, and I'll have to
figure out where that is. I had thought that it was automatic, and allowed
where there was only one possible non-negative number for that category

I can see about carving out some time to look at that tomorrow.

srl295:
> No, that should not be the case. We have made changes in the rules and have had them reflected in the ST well before they have gone into ICU. 

hm.  I added a unit test that should have worked, but it didn't. I traced through the code and it calls `PluralRules.forInstance` in ICU - the overriding code was commented out and removed some time ago. The unit test is checked in here, in `TestSupplementalDataInfo.java` - passes for Arabic but fails for Maltese.  (It's asserting that `two` is in the keywords, and that `two` == `2.0`

but maybe the rules are wrong somehow.

> One thing you might be missing is that you have to rerun the plural formatting — that regenerates the samples, which are actually crucial to having CLDR (and ICU) work properly. The case of no placeholders does happen in a few locales, and I'll have to figure out where that is. I had thought that it was automatic, and allowed where there was only one possible non-negative number for that category I can see about carving out some time to look at that tomorrow.

OK.  thanks.

srl295:
i'll try to figure out how to 'rerun the plural formatting' …

srl295:
@pedberg-icu  did i get the new alt= forms correct?

macchiati:
See
https://cldr.unicode.org/development/updating-codes/addingfixing-plural-and-ordinals
,

Third section.

srl295:
> See https://cldr.unicode.org/development/updating-codes/addingfixing-plural-and-ordinals , Third section.

( By the way WritePluralRulesSpreadsheets crashes even without my changes, so filed CLDR-15647 )

I ran the tools in the third section, but it did not affect my failing test. (The arabic part is OK but maltese fails). Tracing the code, it does seem like it's simply pulling from ICU4J.

```java
    void TestMaltesePlurals() {
        PluralRules.Factory prf = PluralRulesFactory.getInstance(SupplementalDataInfo.getInstance());
        PluralRules arrules = prf.forLocale(ULocale.forLanguageTag("ar"), PluralType.CARDINAL);
        assertNotNull(arrules, "ar rules");
        assertTrue(arrules.getKeywords().contains("two"), "ar did not have two");
        assertEquals(2.0, arrules.getUniqueKeywordValue("two"), "ar unique value for 'two'");

        PluralRules rules = prf.forLocale(ULocale.forLanguageTag("mt"), PluralType.CARDINAL);
        assertNotNull(rules, "mt rules");
        assertTrue(rules.getKeywords().contains("two"), "mt did not have two");
        assertEquals(2.0, rules.getUniqueKeywordValue("two"), "mt unique value for 'two'");
    }
```

macchiati:
That is the right place to look, in checkAndReplacePlaceholders.

And you are right. It should not be calling  PluralRules.*forLocale*.

Instead, it should be calling
supplementalDataInfo.getPlurals(...).getPluralRules().

We should check to see that there are no instances of PluralRules.
*forLocale* .

A thought; we might want to investigate whether we can have a test that
scans our cldr code for ICU methods that shouldn't ever be called!

srl295:
> Instead, it should be calling `supplementalDataInfo.getPlurals(...).getPluralRules()`

yes, but `supplementalDataInfo.getPlurals(...).getPluralRules()` just calls `PluralRules.*forLocale*`. It used to have an override mechanism, but it was deleted.

macchiati:
It should never have been an "override". Instead, when PluralInfo is
created, it is passed in rules that come from supplemental data, and any
CLDR code that uses plurals should call supplementalDataInfo.
getPlurals(...).getPluralRules(), which as far as I can tell, does not call
PluralRules.forLocale.

PluralInfo:

        *public* PluralRules getPluralRules() {

            *return* pluralRules;

        }

        *private* PluralInfo(Map<Count, String> countToRule, PluralType
pluralType) {

            EnumMap<Count, String> tempCountToRule = *new* EnumMap<>(Count.
*class*);

            tempCountToRule.putAll(countToRule);

            *this*.countToRule = Collections.*unmodifiableMap*(
tempCountToRule);


            // now build rules

            NumberFormat nf = NumberFormat.*getNumberInstance*(ULocale.
*ENGLISH*);

            nf.setMaximumFractionDigits(2);

            StringBuilder pluralRuleBuilder = *new* StringBuilder();

            *for* (Count count : countToRule.keySet()) {

                *if* (pluralRuleBuilder.length() != 0) {

                    pluralRuleBuilder.append(';');

                }

                pluralRuleBuilder.append(count).append(':').append(
countToRule.get(count));

            }

            pluralRulesString = pluralRuleBuilder.toString();

            *try* {

                pluralRules = PluralRules.*parseDescription*(
pluralRulesString);

            } *catch* (ParseException e) {

                *throw* *new* IllegalArgumentException("Can't create
plurals from <" + pluralRulesString + ">", e);

            }



that is called from


SupplementalDataInfo


    *private* *void* addPluralInfo(PluralType pluralType) {

        *final* String[] locales = lastPluralLocales.split("\\s+");

        PluralInfo info = *new* PluralInfo(lastPluralMap, pluralType);

srl295:
It's kind of a mess. 

1. There's a `org.unicode.cldr.tool.PluralRulesFactory` called by a lot of CLDR code, which takes a SupplementalDataInfo, but turns around  and calls PluralRules.forLocale(). returns a PluralRules.

2. Then there is code as quoted above which calls PluralRules.forLocale() directly. returns a PluralRules

3. Then there is SupplementalDataInfo.getPlurals() which has arguments in the opposite order from forLocale, and lowercase enums instead of all caps (different PluralType) class, but returns a SupplementalDataInfo.PluralInfo which is unrelated to PluralRules… different API.

Why was PluralInfo added?

srl295:
Probably pluralrules factory can be updated to wrap a pluralinfo in a pluralrules .. then we can fix all the code sites.

macchiati (approved):

srl295:
@macchiati  grazzi ħafna

srl295:
post script - vetters are already correcting a couple of spelling errors here. https://st.unicode.org/cldr-apps/v#/mt/Fields/11d2ca76c2d8faa5

PR #1711, opened 2022-01-25
CLDR-13467 In comments, some imported data and for en display name, change "Kiev" to "Kyiv"

pedberg-icu:
CLDR-13467

Supporting evidence in ticket comment: https://unicode-org.atlassian.net/browse/CLDR-13467?focusedCommentId=163267

Change spelling from "Kiev" to "Kyiv" as follows:
1. In the _description_ field only, for the BCP47 short zone name "uaiev"; we cannot change the alias "Europe/Kiev" since that is determined by the IANA timezone project, but users should not see that anyway.
2. In the standard English name (user displayable) of the exemplar city for timezone "Europe/Kiev"; previously the "Kyiv" spelling was available as alt="formal", now we have no alt value.
3. In the external wiikidata name data for entity Q1899, the English spelling is updated to correspond to what wikidata currently has.
4. And correspondingly, in the English subdivisions data, the user-displayable name for subdivision ua30 is updated.

As noted, the IANA tzdata name "Europe/Kiev" is preserved as an internal identifier, but users on systems that make use of CLDR data should not normally see that identifier.


srl295 (approved):

pedberg-icu:
@srl295 can you re-approve this? It is exactly as before, it is just that in the meantime I had experimented with adding an alt= with the old name, but then decided that was too much of a problem and retracted.

srl295 (commented):
LGTM awaiting root fix

srl295 (dismissed):

srl295 (approved):

PR #1712, opened 2022-01-26
CLDR-15307 en: add names for seed languages and others whose names will have modern coverage

pedberg-icu:
…
CLDR-15307

Add English display names for languages:
* which have locales in seed but currently no display names in English
* whose names will be added for moden coverage in CLDR 42 (but useful to have the English names now)

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -60,6 +60,7 @@ annotations.
 			<language type="asa">Asu</language>
 			<language type="ase">American Sign Language</language>
 			<language type="ast">Asturian</language>
+			<language type="atj">Atikamekw</language>
 			<language type="av">Avaric</language>
 			<language type="avk">Kotava</language>
 			<language type="awa">Awadhi</language>
@@ -90,6 +91,7 @@ annotations.
 			<language type="bjn">Banjar</language>
 			<language type="bkm">Kom</language>
 			<language type="bla">Siksika</language>
+			<language type="blt">Tai Dam</language>
 			<language type="bm">Bambara</language>
 			<language type="bn">Bangla</language>
 			<language type="bo">Tibetan</language>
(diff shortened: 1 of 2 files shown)


pedberg-icu:
Will add translations hints for CLDR 42 with https://unicode-org.atlassian.net/browse/CLDR-13224

common/main/en.xml
```diff
@@ -265,6 +280,8 @@ annotations.
 			<language type="ig">Igbo</language>
 			<language type="ii">Sichuan Yi</language>
 			<language type="ik">Inupiaq</language>
+			<language type="ike">Eastern Canadian Inuktitut</language>
+			<language type="ikt">Western Canadian Inuktitut</language>
```
AEApple, line 284: When there are multiple language names in the ISO table how does CLDR decide which one? Should we add the alternates?

common/main/en.xml
```diff
@@ -331,6 +348,7 @@ annotations.
 			<language type="kut">Kutenai</language>
 			<language type="kv">Komi</language>
 			<language type="kw">Cornish</language>
+			<language type="kwk">Kwakʼwala</language>
```
AEApple, line 351: Several languages have different names in ISO, what is the best resource to check against?

common/main/en.xml
macchiati: We don't typically need alternates. It would only be if both names are used with the same frequency, and there is a large user community.

common/main/en.xml
macchiati: Wikipedia and SIL are reasonable sources. (SIL has a paywall though, so you can't hit it many times a month). 
Note: we can always change names later.

macchiati (approved):
Looks ok by me, but see if Annemarie has any more questions.

AEApple (approved):

PR #1713, opened 2022-01-26
CLDR-14626 update territoryInfo for CA based on 2016 census data

pedberg-icu:
CLDR-14626

Update territoryInfo (country_language_population) data for Canada based on 2016 census data.
For non-indigenous languages, included those with population percentage down to 0.1%. For indigenous languages (including many "official regional" or "recognized" languages), included those with population percentage down to 0.001% or official status. Note that compared to the previous set:
* I folded **ikt** into **iu** macrolanguage, and folded **scs** into **den** macrolanguage
* I dropped a couple of other languages that had percentages below 0.001%.

This also entailed adding some likelySubtags data, and updating language names in a few places.


rickmcgowan (dismissed):
Diffs look reasonable to me. However note that I've never done any updates in XML for such things, so you'll probably want more reviewers. :-)

common/supplemental/supplementalData.xml
```diff
@@ -2604,30 +2604,84 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="es" populationPercent="28"/>	<!--Spanish-->
 		</territory>
 		<territory type="CA" gdp="1774000000000" literacyPercent="99" population="37694100">	<!--Canada-->
-			<languagePopulation type="en" populationPercent="86" officialStatus="official" references="R1194"/>	<!--English-->
```
macchiati, line 2607: We look for fluent usage, not 'mother tongue'. As near as I can tell from X, the bilingual french-english count is near 18%, making that would be:
english	93%
french	40%

details:

lang | partition | summed
-- | -- | --
french | 22.80% | 40.70%
english | 75.40% | 93.30%
both | 17.90% |

common/supplemental/supplementalData.xml
```diff
@@ -2604,30 +2604,84 @@ XXX Code for transations where no currency is involved
...
+			<languagePopulation type="war" populationPercent="0.0019" references="R1329"/>	<!--Waray-->
+			<languagePopulation type="crg" populationPercent="0.0019" references="R1329"/>	<!--Michif-->
+			<languagePopulation type="lil" populationPercent="0.0015" references="R1329"/>	<!--Lillooet-->
+			<languagePopulation type="pqm" populationPercent="0.0014" references="R1329"/>	<!--Malecite-->
+			<languagePopulation type="oka" populationPercent="0.0014" references="R1329"/>	<!--Okanagan-->
+			<languagePopulation type="kwk" populationPercent="0.0011" references="R1329"/>	<!--Kwakʼwala-->
+			<languagePopulation type="crl" populationPercent="0.0011" references="R1329"/>	<!--Northern East Cree-->
+			<languagePopulation type="gwi" populationPercent="0.0008" officialStatus="official_regional" references="R1329"/>	<!--Gwichʼin-->
+			<languagePopulation type="crm" populationPercent="0.0003" references="R1329"/>	<!--Moose Cree-->
+			<languagePopulation type="crj" populationPercent="0.0001" references="R1329"/>	<!--Southern East Cree-->
```
macchiati, line 2684: I'm ok with this for now.

Longer term, I think we need to consider a somewhat higher bar for what we include. My concern is that people might expect the same level of detail for all countries.  That is if we list a language with (say) 35 speakers in CA, does that mean that if someone looks at (say) Ukraine, would they expect that every language with over 35 speakers would be listed?

Anyway, I think this needs more discussion (and not on this PR). We might put some strong qualifications into the spec and data file so that people are not mislead, or instead of listing communities of (say) < 10,000 we might include a reference to the source document.


100.0000% | 35,121,728 (for Canada, 2016)
-- | --
10.0000% | 3,512,173
1.0000% | 351,217
0.1000% | 35,122
0.0100% | 3,512
0.0010% | 351
0.0001% | 35

macchiati (changes requested):
I have 2 comments I'd like to see a change for. The other is marked "I'm ok with this for now" and is not a blocker.

common/supplemental/supplementalData.xml
pedberg-icu: The census data that I used was not just about "mother tongue" but about all languages that are used at home. And while I think that is good data for everything other than French and English, I agree that for those two it would undercount. I will fix the data for those.

common/supplemental/supplementalData.xml
pedberg-icu: The problem is that (1) clients want to programmatically get data on language usage (so pointing off somewhere else is not useful), and (2) some clients (at least ones we have) have a strong interest in support for indigenous languages. And some Canadian provinces (e.g. Northwest Territories) designate some languages with just a few hundred speakers as official in those provinces. So I would like to preserve the indigenous language data in this case. But what I can do (as only as I am fixing the data for French and English) is use a higher threshold for the non-indigenous languages and reduce those somewhat.

common/supplemental/supplementalData.xml
```diff
@@ -5652,5 +5706,6 @@ XXX Code for transations where no currency is involved
 		<reference type="R1326" uri="https://www.ethnologue.com/language/osa">[missing]</reference>
 		<reference type="R1327" uri="https://pt.wikipedia.org/wiki/Nheengatu">[missing]</reference>
 		<reference type="R1328" uri="https://statisticsmaldives.gov.mv/statistical-release-iii-education/">2014 Maldives: 98% literacy in Divehi, 75% in English</reference>
+		<reference type="R1329" uri="https://www12.statcan.gc.ca/census-recensement/2016/dp-pd/dt-td/Rp-eng.cfm?TABID=2&amp;Lang=E&amp;APATH=3&amp;DETAIL=0&amp;DIM=0&amp;FL=A&amp;FREE=0&amp;GC=0&amp;GID=1235625&amp;GK=0&amp;GRP=1&amp;PID=110212&amp;PRID=10&amp;PTYPE=109445&amp;S=0&amp;SHOWALL=0&amp;SUB=0&amp;Temporal=2016&amp;THEME=118&amp;VID=0&amp;VNAMEE=&amp;VNAMEF=&amp;D1=0&amp;D2=0&amp;D3=0&amp;D4=0&amp;D5=0&amp;D6=0">Canada 2016 census language data; official status from Wikipedia Languages_of_Canada</reference>
```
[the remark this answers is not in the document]
pedberg-icu: Probably because the link in xml form has ampersand entity references instead of actual ampersands. You can find the actual link (with real ampersands) in country_language_population_raw.txt.

macchiati (approved):

PR #1730, opened 2022-02-02
CLDR-14289 switch to new coverage rules

macchiati:
CLDR-14289

shift to new rules, already reviewed

fixed core checking errors: 
 * no didn't show up as having country data
 * The time cycle wasn't explicit for some locales. Added them explicitly.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLocaleCoverage.java
The bulk of the changes.  Add more information to the chart, helping to debug the data: notably the default locale, and an 'imputed' coverage level. Also remove some fields that are of lesser use.

This is where the non //ldml/ checks go. Expanded what we had, and assigned the different features to different levels. I left some of the names the same  (like the name of the class) to make diffing easier. Also fixed a bug in how the orientation was being computed.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
fixed some test failures caused by the changes.

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -3696,6 +3696,7 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="tr" populationPercent="1.2"/>	<!--Turkish-->
 		</territory>
 		<territory type="NO" gdp="381200000000" literacyPercent="100" population="5467440">	<!--Norway-->
+			<languagePopulation type="no" populationPercent="100" officialStatus="official"/>	<!--Norwegian -->
 			<languagePopulation type="nb" populationPercent="100" officialStatus="official"/>	<!--Norwegian Bokmål-->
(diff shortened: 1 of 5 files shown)


common/supplemental/supplementalData.xml
```diff
@@ -3696,6 +3696,7 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="tr" populationPercent="1.2"/>	<!--Turkish-->
 		</territory>
 		<territory type="NO" gdp="381200000000" literacyPercent="100" population="5467440">	<!--Norway-->
+			<languagePopulation type="no" populationPercent="100" officialStatus="official"/>	<!--Norwegian -->
```
pedberg-icu, line 3699: We will want to add a line for this in country_language_population_raw.txt as well (otherwise this will get overwritten), but I can do that as part of other changes I plan to make today.

pedberg-icu (approved):

PR #1733, opened 2022-02-02
CLDR-15291 check in additional fixes

macchiati:
CLDR-15291


AEApple (changes requested):
Do we not need the Yoruba fixes?

AEApple:
> 

Confirmed that we should not make the Yoruba changes since the proposed changes are not clear enough.

AEApple (dismissed):

macchiati:
Not sure how that happened, but the actual changes are fine. I'll just squash, but will need a rereview.

AEApple (approved):

AEApple:
> Not sure how that happened, but the actual changes are fine. I'll just squash, but will need a rereview.

Yes, there was only the 1 file with the expected change.

PR #1734, opened 2022-02-02
CLDR-14992 so far just add code to list locale script/regions not included in coverage targets

pedberg-icu:
CLDR-14992


macchiati (approved):

pedberg-icu:
This is closed in favor of https://github.com/unicode-org/cldr/pull/1982

PR #1737, opened 2022-02-03
CLDR-15276 XML fixes to main/hi_Latn and annotations/dsb

srl295:
CLDR-15276

These errs don't show up locally but do show up on the build system.
Perhaps a JDK or XML parser version difference.

```
Error:  (TestPaths.java:437)  Error: Inconsistency:	file=annotations/dsb.xml	elementType=PCDATA	value=«»	path=//ldml/annotations/annotation[@cp="👨‍⚕"][@type="tts"]

Error:  (TestPaths.java:468)  Error: Duplicate: hi_Latn.xml, //ldml/localeDisplayNames/scripts/script[@type="Moon"
```

### DIFF
--- a/seed/main/hi_Latn.xml
+++ b/seed/main/hi_Latn.xml
@@ -631,7 +631,6 @@ annotations.
 			<script type="Mero">Meroitic</script>
 			<script type="Mlym">Malayalam</script>
 			<script type="Mong">Mongolian</script>
-			<script type="Moon">Moon</script>
 			<script type="Mtei">Meitei Mayek</script>
 			<script type="Moon">Moon</script>
 			<script type="Nkoo">N’Ko</script>
--- a/common/annotations/dsb.xml
+++ b/common/annotations/dsb.xml
@@ -832,7 +832,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="🧑‍⚕">chórobny sotša | gójc | gójc/gójcowka | strowotnistwo | terapeut</annotation>
 		<annotation cp="🧑‍⚕" type="tts">gójc/gójcowka</annotation>
 		<annotation cp="👨‍⚕">doktor | gójc | gójcojski kitel</annotation>
-		<annotation cp="👨‍⚕" type="tts"/>
 		<annotation cp="👩‍⚕">gójcojski kitel | gójcowka</annotation>
 		<annotation cp="👩‍⚕" type="tts">gójcowka</annotation>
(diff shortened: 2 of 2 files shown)


AEApple:
Can we file a ticket to create a test for these?

srl295:
> Can we file a ticket to create a test for these or at least the missing TTS one?

we have tests for them. THe question is why are they not usually failing.

AEApple (approved):

AEApple:
> > Can we file a ticket to create a test for these or at least the missing TTS one?
> 
> we have tests for them. THe question is why are they not usually failing.

Okay a ticket for why they're not failing then?

srl295:
> > > Can we file a ticket to create a test for these or at least the missing TTS one?
> > 
> > 
> > we have tests for them. THe question is why are they not usually failing.
> 
> Okay a ticket for why they're not failing then?

ok. after i analyze

srl295:
`ubuntu-20.04` version jumped from 20220123.1 to 20220131.1 in the 8 hours between commit [666bcce](https://github.com/unicode-org/cldr/commit/666bcce731281e5bbc45e342e427d7200ee6d363) and when i tagged [snapshot/2022-02-03](https://github.com/unicode-org/cldr/releases/tag/snapshot%2F2022-02-03) which are _the same commit_

srl295:
- https://github.com/actions/virtual-environments/blob/ubuntu20/20220123.1/images/linux/Ubuntu2004-Readme.md 
- https://github.com/actions/virtual-environments/blob/ubuntu20/20220131.1/images/linux/Ubuntu2004-Readme.md

seed/main/hi_Latn.xml
```diff
@@ -631,7 +631,6 @@ annotations.
 			<script type="Mero">Meroitic</script>
 			<script type="Mlym">Malayalam</script>
 			<script type="Mong">Mongolian</script>
-			<script type="Moon">Moon</script>
```
pedberg-icu, line 634: Thus should not be changed in the data (and besides I have moved this file to common in a different PR in progress). There's a  list somewhere of codes whose display names can legitimately match the code, that should be updated.

pedberg-icu (changes requested):

seed/main/hi_Latn.xml
srl295: @pedberg-icu  the issue here is that this exact element is duplicated, with the same content.

srl295: look at like 636/635 of this same file.

pedberg-icu (approved):

PR #1738, opened 2022-02-04
CLDR-15328 move hi_Latn to common (fix identity err, delete noon); fix Ethiopic transform IDs

pedberg-icu:
CLDR-15328

Fix some issues encountered in integrating CLDR 41 data to ICU for CLDR 41-release-alpha0:
1. Move hi_Latn* from seed/ to common/ (required in order to pick up for ICU). This resulted in more tests being run, so then:
    * Fix the identity err in annotations/hi_Latn.xml (had <language type="hi_Latn"/> instead of separate entries for language and script)
    * Remove the dayPeriod entries for noon
    * Added an exception to a test that checked that locales with language & script & parent != toot must be default content locales.
    * Added likelySubtags.
2. Fixed the BCP-47-style IDs for most of the newly added Ethiopic transforms
    * Needed short names for some of the new m0 variants
    * Fixed incorrect use of Cyri instead of Cyrl for Cyrillic
    * Fixed incorrect ordering of the BCP-47 ID in most; the target is supposed to come first before the -t-. Also needed und- before script codes.
    * Updated the test data accordingly


macchiati (dismissed):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestLocale.java
```diff
@@ -277,6 +277,9 @@ public void checkLanguage(String file, String language) {
         }
     }
 
+    // Locale exceptions to rule that parentIsRoot != isDefaultContent
+    static final Set<String> ROOT_PARENT_DEFAULT_CONTENT_EXCEPTIONS = ImmutableSet.of("hi_Latn");
```
srl295, line 281: I added similar to TestInheritance.

srl295 (approved):

pedberg-icu:
Hmm. The original version of this PR was failing tests because of the problem in dsb that Steven fixed in https://github.com/unicode-org/cldr/pull/1737. So I tried updating my PR by merging in main, squashing, and then force-pushing. That seems to have messed up my PR and I do not know how to fix it. I will re-create the PR on a fresh branch and close this one.

pedberg-icu:
OK, this is replaced with https://github.com/unicode-org/cldr/pull/1739 which is the same set of changes but starting with a fresh branch off of an updated main.

PR #1741, opened 2022-02-06
CLDR-15297 generate coverage charts for v41

macchiati:
CLDR-15297

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/Level.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/Level.java
@@ -3,7 +3,7 @@ package org.unicode.cldr.util;
 import java.util.Locale;
 import java.util.Set;
 
-import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSortedSet;
 
 /**
  * A simple class representing an enumeration of possible CLDR coverage levels. Levels may change in the future.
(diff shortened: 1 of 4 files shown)


pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/Level.java
```diff
@@ -12,14 +12,14 @@
  *
  */
 public enum Level {
-    UNDETERMINED(0, "none", 0),
-    CORE(10, "G4", 100),
-    BASIC(40, "G3", 80),
-    MODERATE(60, "G2", 70),
-    MODERN(80, "G1", 50),
-    COMPREHENSIVE(100, "G0", 2);
+    UNDETERMINED(0, "none", 0, "�"),
```
srl295, line 15: intentionally U+FFFD? Would something else be better?

srl295 (approved):
lgtm with comment

tools/cldr-code/src/main/java/org/unicode/cldr/util/Level.java
macchiati: It should never show up, so that's just a visual reminder!

macchiati:
Had to squash manually; can one of you rereview?

PR #1742, opened 2022-02-07
CLDR-15328 ks, fix number symbols for arabext (decimal & group were the same)

pedberg-icu:
CLDR-15328

ks (Kashmiri) uses 'arabext' digits by default, but for arabext had the same value for &lt;decimal&gt; and &lt;group&gt;: "،" U+060C ARABIC COMMA. This caused parse failures in ICU4J number tests.

The other relevant locales such as pa_Arab, ur, fa (and root) have the following for 'arabext':
&lt;decimal&gt;: "٫" U+066B ARABIC DECIMAL SEPARATOR
&lt;group&gt;: "٬" U+066C ARABIC THOUSANDS SEPARATOR (which looks almost identical to U+060C ARABIC COMMA)
So switched ks/arabext to have the same &lt;decimal&gt; and &lt;group&gt; as those, and also added U+066B and U+066C to the ks number exemplars.

Note that ks has the following for 'latn' digits
&lt;decimal&gt;: "." U+002E FULL STOP
&lt;group&gt;:  "،" U+060C ARABIC COMMA
The &lt;group&gt; here may also be an error but is not causing parse failures so I did not change it.


srl295 (approved):

roozbehp:
LGTM

PR #1747, opened 2022-02-09
CLDR-12028 modify tests for compliance and fix data as nec.

macchiati:
CLDR-12028

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
@@ -6,6 +6,7 @@ import java.util.regex.Pattern;
 
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
 import org.unicode.cldr.util.PatternCache;
+import org.unicode.cldr.util.XPathParts;
 
 public class CheckPlaceHolders extends CheckCLDR {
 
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
```diff
@@ -46,6 +47,54 @@ public CheckCLDR handleCheck(String path, String fullPath, String value, Options
...
+            case "2":
+                int pos1 = value.indexOf("{0}");
+                int pos2 = value.indexOf("{1}");
+                if (pos1 > pos2) {
+                    result.add(new CheckStatus().setCause(this)
+                        .setMainType(CheckStatus.errorType)
+                        .setSubtype(Subtype.invalidPlaceHolder)
+                        .setMessage("Invalid list pattern «" + value + "»: the placeholder {0} must be before {1}."));
+                }
+            }
```
robertbastian, line 77: The spec also allows a special "3" case, which currently doesn't exist but should be checked:

```suggestion
            case "3":
                int pos1 = value.indexOf("{0}");
                int pos2 = value.indexOf("{1}");
                int pos3 = value.indexOf("{2}");
                if (pos1 > pos2 || pos2 > pos3) {
                    result.add(new CheckStatus().setCause(this)
                        .setMainType(CheckStatus.errorType)
                        .setSubtype(Subtype.invalidPlaceHolder)
                        .setMessage("Invalid list pattern «" + value + "»: the placeholders {0}, {1}, {2} must appear in that order."));
                }
            }

pedberg-icu (approved):

srl295 (approved):

sffc (commented):
Thank you!  This is a huge improvement that fixes a whole bunch of problems and reduces edge cases throughout the whole i18n stack.

PR #1753, opened 2022-02-10
CLDR-15341 ks, add \u0626 to aux exemplars (used in name of month "May" and elsewhere)

pedberg-icu:
CLDR-15341

Add to ks aux exemplars U+0626, which is used in the name for gregorian "May" (correctly spelled according to vetter) and in display the names of a language and region, though it is apparently not part of the normal ks orthography (may be used for words imported from other languages).

### DIFF
--- a/common/main/ks.xml
+++ b/common/main/ks.xml
@@ -927,7 +927,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	</layout>
 	<characters>
 		<exemplarCharacters>[ء آ أ ٲ ؤ ا ب پ ت ث ٹ ج چ ح خ د ذ ڈ ر ز ڑ ژ س ش ص ض ط ظ ع غ ف ق ک گ ل م ن ں ھ ہ و ۄ ۆ ی ۍ ؠ ے]</exemplarCharacters>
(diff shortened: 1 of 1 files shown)


AEApple (commented):
- Should there be a space between \u200E\u200F?
- Is there a reason that so many of the characters are the escaped version when they have literal versions, or is that they are too easily confused with other similar characters?

roozbehp:
Wikipedia software's localization uses yet another spelling:
Meem, Hamza Above, Farsi Yeh, Farsi Yeh: مٔیی

See:
https://github.com/wikimedia/mediawiki/blob/master/languages/i18n/ks-arab.json

pedberg-icu:
> For what it's worth, in the Kashmiri Wikipedia, May appears to be spelled two different ways: Meem, Hamza Above, Farsi Yeh: مٔی Meem, Yeh with Hamza Above, Farsi Yeh: مئی
> 
> The second matches what CLDR has, but the first is slightly more common in the Kashmiri Wikipedia.

@roozbehp: CLDR generally tries to follow what its vetters provide (unless there are obvious misunderstandings or other problems, or geopolitical reasons for changing etc) and in this instance Meike checked back with the vetter(s) who double-checked the spelling and said it was what they intended. We would also want to avoid introducing inconsistencies with spelling used elsewhere in the data,

PR #1758, opened 2022-02-14
CLDR-15361 move seed to common

macchiati:
CLDR-15361

Test moving all seed to common, to see what breaks.


macchiati:
Obsolete

PR #1765, opened 2022-02-16
CLDR-15366 - update iso-639-3 files, lst, tld

srl295:
- Sunu script added. (future Unicode)

CLDR-15366

Note: tools needed https://github.com/unicode-org/cldr/pull/1764 fix to proceed.


srl295:
test fails still

srl295:
I guess `Sunu` is skipped for now because it's "non-Unicode"?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/iso-639-3-macrolanguages.tab
```diff
@@ -1,454 +1,455 @@
...
-zho	gan	A
-zho	hak	A
-zho	hsn	A
-zho	lzh	A
-zho	mnp	A
-zho	nan	A
-zho	wuu	A
-zho	yue	A
-zza	diq	A
+M_Id	I_Id	I_Status
```
markusicu, line 1: Why does it look in GitHub like the whole file changed?
Is there an inadvertent whitespace change on every line?

Same question for all of the .tab files.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language-subtag-registry
```diff
@@ -44795,7 +44795,7 @@ Added: 2005-10-16
 Type: script
 Subtag: Kawi
 Description: Kawi
-Added: 2021-12-22
+Added: 2021-12-24
```
markusicu, line 44798: FYI  The old dates were my guesses from when I cherry-picked just the new records, after they were proposed but before they went live in the LSR :-)

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/iso-639-3-macrolanguages.tab
srl295: @markusicu because it was checked in with CRLF.  I'm updating it to be consistent with all other source files.

markusicu (approved):
lgtm

PR #1767, opened 2022-02-17
CLDR-15366 v41 BRS update world bank data and tool

srl295:
- update world bank data
- make AddPopulationData world bank reader not year-dependent

CLDR-15366

- [ ] Update https://cldr.unicode.org/development/updating-codes/updating-population-gdp-literacy to not recommend changing the enum every year


macchiati (changes requested):
My only hesitation is that I think that Peter may have added some exceptional cases  that are being removed by these changes.

<language type="ikt" territories="CA" alt="secondary"/>

If so, then we need to add these to (some) internal exception list so that they are not stomped by the tool.

common/supplemental/supplementalData.xml
```diff
@@ -1454,12 +1454,16 @@ XXX Code for transations where no currency is involved
 		<language type="cjm" scripts="Cham"/>
 		<language type="cjm" scripts="Arab" alt="secondary"/>
 		<language type="cjs" scripts="Cyrl"/>
-		<language type="ckb" scripts="Arab" territories="IQ IR"/>
+		<language type="ckb" scripts="Arab"/>
+		<language type="ckb" territories="IQ IR" alt="secondary"/>
```
pedberg-icu, line 1458: Until we fix [CLDR-14992](https://unicode-org.atlassian.net/browse/CLDR-14992), we need to manually keep the existing line 1457, which was changed as a temporary way to address [CLDR-13595](https://unicode-org.atlassian.net/browse/CLDR-13595.).

pedberg-icu (changes requested):

pedberg-icu:
> My only hesitation is that I think that Peter may have added some exceptional cases that are being removed by these changes.

There are just two places where I saw problems: One is the languageInfo data for cab which we need to adjust manually for the time being, and the other is the territoryInfo for MV in which dv is repeated (may be an error in the _raw file). See comments above for both issues.

srl295:
@pedberg-icu thanks for The analysis… I've been trying to figure these out

common/supplemental/supplementalData.xml
```diff
@@ -3588,9 +3598,9 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="fr" populationPercent="3" officialStatus="official"/>	<!--French-->
 			<languagePopulation type="ta" populationPercent="2.5"/>	<!--Tamil-->
 		</territory>
-		<territory type="MV" gdp="6901000000" literacyPercent="98.4" population="540542">	<!--Maldives-->
-			<languagePopulation type="dv" populationPercent="98" officialStatus="official" references="R1328"/>	<!--Divehi-->
-			<languagePopulation type="en" populationPercent="75" references="R1328"/>	<!--English-->
+		<territory type="MV" gdp="6901000000" literacyPercent="98.4" population="391904">	<!--Maldives-->
+			<languagePopulation type="dv" populationPercent="98" officialStatus="official" references="R1332"/>	<!--Divehi-->
+			<languagePopulation type="dv" populationPercent="75" references="R1332"/>	<!--Divehi-->
```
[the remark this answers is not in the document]
pedberg-icu: Confirmed that this is due to my error in tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt, for Maldives it has "English	dv	75%", should be "English	en	75%"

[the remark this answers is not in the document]
srl295: @pedberg-icu  should the tools assert that `English` ≈ `en` ?

common/supplemental/supplementalData.xml
srl295: I'll do that and add a comment.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt
```diff
@@ -2,175 +2,175 @@
 #######
 #WARNING: this file is in TSV format (for pasting from a spreadsheet); make sure you maintain tabs!
 #######
-Afghanistan	AF	"34,940,837"	28%	"69,450,000,000"	official_regional	Baluchi	bal	"233,000"			
-Afghanistan	AF	"34,940,837"	28%	"69,450,000,000"		Hazaragi	haz	5.9%			
+Afghanistan	AF	"34,940,837"	28%	"69,450,000,000"	official_regional	Baluchi	bal	"233,000"
```
pedberg-icu: Hmm, I do not think the trailing tabs should have been deleted, this was set up for import to a spreadsheet.

common/supplemental/supplementalData.xml
pedberg-icu: Still seems to be wrong here...

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt
srl295: would trailing tabs affect the spreadsheet though?

srl295:
@macchiati  @pedberg-icu 

Also, ConvertLanguageData *generates* updated versions of these two:

- tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt
- tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script_raw.txt

They have quite different content, should they be checked in also?

common/supplemental/supplementalData.xml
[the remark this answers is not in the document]
pedberg-icu: That sounds good, I can file a ticket for that.

pedberg-icu:
> @macchiati @pedberg-icu
> 
> Also, ConvertLanguageData _generates_ updated versions of these two:
> 
> * tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt
> * tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script_raw.txt
> 
> They have quite different content, should they be checked in also?

I do not know, I always thought the _raw files were just inputs.

common/supplemental/supplementalData.xml
srl295: i think you might have reviewed just before the push https://github.com/unicode-org/cldr/pull/1767/files#diff-5ea2936f709d9dc83508c70c9a589a5bdaadf9392a0ca96d9b56507621f4090bR1457

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt
pedberg-icu: Perhaps not, though I guess it could depend on the spreadsheet.

pedberg-icu (approved):

common/supplemental/supplementalData.xml
pedberg-icu: Ah yes, looks OK now.

macchiati (approved):

srl295:
> > @macchiati @pedberg-icu
> 
> > 
> 
> > Also, ConvertLanguageData _generates_ updated versions of these two:
> 
> > 
> 
> > * tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population_raw.txt
> 
> > * tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script_raw.txt
> 
> > 
> 
> > They have quite different content, should they be checked in also?
> 
> 
> 
> I do not know, I always thought the _raw files were just inputs.

@macchiati do you know about the Generated .txt files?

PR #1772, opened 2022-02-22
CLDR-15187 CLDRLocale.ROOT_NAME = root, CLDRLocale.UNKNOWN_LOCALE_NAME = und

btangmu:
-New string constants for root and und, for encapsulation

CLDR-15187

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
@@ -296,7 +296,7 @@ public class StandardCodes {
                 default:
                     for (Iterator<String> it = result.iterator(); it.hasNext();) {
                         String code = it.next();
-                        if (code.equals("root") || code.equals("QO"))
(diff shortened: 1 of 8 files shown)


srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -1076,7 +1076,7 @@ private void add(CodeType type, String code, List<String> otherData) {
     static final String registryName = CldrUtility.getProperty("registry", "language-subtag-registry");
 
     public enum LstrType {
-        language("und", "zxx", "mul", "mis", "root"),
+        language(CLDRLocale.UNKNOWN_LOCALE_NAME, "zxx", "mul", "mis", CLDRLocale.ROOT_NAME),
```
srl295, line 1079: I'm a little, or more than a little, concerned about this. CLDRLocale has static init implications.  it might be better to move these string constants under `CLDRUtility` or even `CLDRFile`, for that reason.   Otherwise it will bring StandardCodes into the static init issue.

srl295 (changes requested):

PR #1775, opened 2022-02-23
CLDR-15385 BRS CLDRModify

btangmu:
-One -fp (DAIP) change for annotations/my.xml

CLDR-15385

### DIFF
--- a/common/annotations/my.xml
+++ b/common/annotations/my.xml
@@ -2007,7 +2007,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🚏" type="tts">ဘတ်စ်ကား မှတ်တိုင်</annotation>
 		<annotation cp="🛣">ကားလမ်း | ဟိုင်းဝေး | အမြန်လမ်း</annotation>
 		<annotation cp="🛣" type="tts">ကားလမ်း</annotation>
-		<annotation cp="🛤">ရထားသံလမြး သံလမြး ရထာျး သံလမျး | သံလမ်း</annotation>
+		<annotation cp="🛤">ရထားသံလမွး သံလမွး ရထြား သံလမြး | သံလမျး</annotation>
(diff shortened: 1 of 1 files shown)


sffc:
@jslin can you take a look?  Which version of the translation is correct?

justinseinlin:
This is the text label for emoji?

common/annotations/my.xml
```diff
@@ -2007,7 +2007,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🚏" type="tts">ဘတ်စ်ကား မှတ်တိုင်</annotation>
 		<annotation cp="🛣">ကားလမ်း | ဟိုင်းဝေး | အမြန်လမ်း</annotation>
 		<annotation cp="🛣" type="tts">ကားလမ်း</annotation>
-		<annotation cp="🛤">ရထားသံလမြး သံလမြး ရထာျး သံလမျး | သံလမ်း</annotation>
+		<annotation cp="🛤">ရထားသံလမွး သံလမွး ရထြား သံလမြး | သံလမျး</annotation>
```
justinseinlin, line 2010: I believe the correct version is:

ရထားသံလမ်း | ရထားလမ်း | သံလမ်း

justinseinlin: I was just guessing `|` is the delimiter for alternate spelling.  Please let me know if that assumption is not correct.

macchiati:
That is correct. The | separates different phrases.

justinseinlin:
> That is correct. The | separates different phrases.
> […](#)

I guess in this case, can we make the change to the commit to have it corrected?  I remember there was another time also trying to correct this emoji but it didn't go anywhere.

btangmu:
It looks like the "correct" version is different from both the "-" and "+" versions for this PR, is that right?

btangmu:
See https://github.com/unicode-org/cldr/pull/1794

PR #1780, opened 2022-02-24
CLDR-15366 make GenerateMaximalLocales write in-place [v41]

srl295:
CLDR-15366

- Still write to Generated for supplementaldata and likely subtags,
but copy into the repository when done.

Related to CLDR-14120

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/draft/FileUtilities.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/draft/FileUtilities.java
@@ -13,6 +13,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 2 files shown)


pedberg-icu:
Doesn't this make it harder to compare old and new for some workflows? That is, git diff will work, but using a text editor to compare old and new is harder...

srl295:
> Doesn't this make it harder to compare old and new for some workflows? That is, git diff will work, but using a text editor to compare old and new is harder...

It makes it easier to iterate on issues and fewer steps to run the tool in the first place.

I guess one could revert the changes in-repo and then do a diff.  This code _copies_ out of Generated, it doesn't remove it from Generated.

Does it make sense to default to an in-place replace?

pedberg-icu:
> It makes it easier to iterate on issues and fewer steps to run the tool in the first place.
> ...
> Does it make sense to default to an in-place replace?

As I think about it more this approach seems to make the most sense, thanks

pedberg-icu (approved):

PR #1781, opened 2022-02-24
CLDR-15366 BRS v41 LikelySubtags and tools

srl295:
- add more exceptions to GenerateMaximalLocales to match current likelies
- minor updates to likely subtags
- GenerateMaximalLocales less noisy by default, easier to see issues
- ConvertLanguageData less noisy by default, updates files in-place
(Related to CLDR-14120)

CLDR-15366

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -524,14 +524,10 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Michif; ?; ? } => { Michif; Latin; Canada }-->
 		<likelySubtag from="crh" to="crh_Cyrl_UA"/>
 		<!--{ Crimean Tatar; ?; ? } => { Crimean Tatar; Cyrillic; Ukraine }-->
-		<likelySubtag from="crj" to="crj_Cans_CA"/>
-		<!--{ Southern East Cree; ?; ? } => { Southern East Cree; Unified Canadian Aboriginal Syllabics; Canada }-->
 		<likelySubtag from="crk" to="crk_Cans_CA"/>
 		<!--{ Plains Cree; ?; ? } => { Plains Cree; Unified Canadian Aboriginal Syllabics; Canada }-->
 		<likelySubtag from="crl" to="crl_Cans_CA"/>
 		<!--{ Northern East Cree; ?; ? } => { Northern East Cree; Unified Canadian Aboriginal Syllabics; Canada }-->
-		<likelySubtag from="crm" to="crm_Cans_CA"/>
-		<!--{ Moose Cree; ?; ? } => { Moose Cree; Unified Canadian Aboriginal Syllabics; Canada }-->
 		<likelySubtag from="crs" to="crs_Latn_SC"/>
 		<!--{ Seselwa Creole French; ?; ? } => { Seselwa Creole French; Latin; Seychelles }-->
 		<likelySubtag from="cs" to="cs_Latn_CZ"/>
@@ -1044,8 +1040,6 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Inupiaq; ?; ? } => { Inupiaq; Latin; United States }-->
 		<likelySubtag from="ikk" to="ikk_Latn_ZZ"/>
 		<!--{ Ika; ?; ? } => { Ika; Latin; Unknown Region }-->
-		<likelySubtag from="ikt" to="ikt_Latn_CA"/>
-		<!--{ Inuinnaqtun; ?; ? } => { Inuinnaqtun; Latin; Canada }-->
 		<likelySubtag from="ikw" to="ikw_Latn_ZZ"/>
 		<!--{ Ikwere; ?; ? } => { Ikwere; Latin; Unknown Region }-->
 		<likelySubtag from="ikx" to="ikx_Latn_ZZ"/>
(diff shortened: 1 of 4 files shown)


common/supplemental/supplementalMetadata.xml
```diff
@@ -1797,8 +1797,9 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<defaultContent locales="
 			aa_ET ab_GE af_ZA agq_CM ak_GH am_ET an_ES ar_001 arn_CL as_IN asa_TZ ast_ES
 			az_Arab_IR az_Cyrl_AZ az_Latn az_Latn_AZ
-			ba_RU bal_Arab bal_Arab_PK bal_Latn_PK bas_CM be_BY bem_ZM bez_TZ bg_BG bgn_PK blt_VN bm_ML bm_Nkoo_ML bn_BD
-			bo_CN br_FR brx_IN bs_Cyrl_BA bs_Latn bs_Latn_BA bss_CM byn_ER
+			ba_RU bal_Arab bal_Arab_PK bal_Latn_PK bas_CM be_BY bem_ZM bez_TZ bg_BG bgn_PK
+			blt_VN bm_ML bm_Nkoo_ML bn_BD bo_CN br_FR brx_IN bs_Cyrl_BA bs_Latn bs_Latn_BA
+			bss_CM byn_ER
```
srl295, line 1802: whitespace only change, but this returns this file to being automatically generated

common/supplemental/likelySubtags.xml
```diff
@@ -2166,14 +2160,10 @@ not be patched by hand, as any changes made in that fashion may be lost.
...
 		<!--{ Scots; ?; ? } => { Scots; Latin; United Kingdom }-->
-		<likelySubtag from="scs" to="scs_Latn_CA"/>
-		<!--{ North Slavey; ?; ? } => { North Slavey; Latin; Canada }-->
 		<likelySubtag from="sd" to="sd_Arab_PK"/>
 		<!--{ Sindhi; ?; ? } => { Sindhi; Arabic; Pakistan }-->
 		<likelySubtag from="sd_Deva" to="sd_Deva_IN"/>
 		<!--{ Sindhi; Devanagari; ? } => { Sindhi; Devanagari; India }-->
-		<likelySubtag from="sd_IN" to="sd_Deva_IN"/>
-		<!--{ Sindhi; ?; India } => { Sindhi; Devanagari; India }-->
 		<likelySubtag from="sd_Khoj" to="sd_Khoj_IN"/>
```
pedberg-icu, line 2167: This was specifically added for a ticket, https://unicode-org.atlassian.net/browse/CLDR-15347. I thought I added the necessary data to get it auto-generated but perhaps not.

pedberg-icu (changes requested):

common/supplemental/likelySubtags.xml
srl295: @pedberg-icu  I've reverted this line. I couldn't trace from the debug lines why this wasn't getting shown. If anything, there was a potential candidate `sd_IN` -> `sd_Arab_IN` which is not what you requested.

srl295: reversion here https://github.com/unicode-org/cldr/compare/unicode-org:2194483..unicode-org:f18db5b

pedberg-icu (approved):

common/supplemental/likelySubtags.xml
pedberg-icu: Thanks! Approved.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateMaximalLocales.java
```diff
@@ -124,7 +127,11 @@
         System.out.println("Keep containers " + cldrContainerToLanguages);
     }
 
-    private static final List<String> KEEP_TARGETS = Arrays.asList("und_Arab_PK", "und_Latn_ET");
+    private static final List<String> KEEP_TARGETS = Arrays.asList(
+        "und_Arab_PK",
+        "und_Latn_ET",
+        "hi_Latn"
```
macchiati, line 133: I think this might want to be hi_Latn_IN?

srl295: hi_Latn is what the tool was checking for

common/supplemental/likelySubtags.xml
```diff
@@ -524,14 +524,10 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Michif; ?; ? } => { Michif; Latin; Canada }-->
 		<likelySubtag from="crh" to="crh_Cyrl_UA"/>
 		<!--{ Crimean Tatar; ?; ? } => { Crimean Tatar; Cyrillic; Ukraine }-->
-		<likelySubtag from="crj" to="crj_Cans_CA"/>
```
macchiati, line 527: I think the likelySubtags might be normally filtered by the locales currently in CLDR, which would explain these being dropped. There is reason for this, since we don't want thousands of languages in this file.

But @pedberg-icu was the one who added the indigenous language information and might have added these manually, so he should be consulted about this.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateMaximalLocales.java
macchiati: ok. I'd thought that the TARGETS meant the maximizations not the sources. But I haven't looked at this in ages!

common/supplemental/likelySubtags.xml
pedberg-icu: No, the 4 that got dropped here were all added by John Emmons in 2014 or so. They are dropped because they did not have data in the current version of the territoryInfo for CA; that is fine.

macchiati: Great!

macchiati (commented):
Looks great, thanks for the cleanup to the code!

PR #1782, opened 2022-02-25
CLDR-15261 Arabic spellout rules render a stray space after the conjunction

grhoten:
This ticket provided additional feedback that the changes were incomplete. I have confirmed that 138000 comes out as the following as per the feedback:
مائة وثمانية وثلاثون ألف وأربعة مائة
Also 1111111111111 comes out as the following:
ترليون ومائة وإحدى عشر مليار ومائة وإحدى عشر مليون ومائة وإحدى عشر ألف ومائة وإحدى عشر

The fix was performed by searching and replacing all instances of the and (و) surrounded by spaces.

CLDR-15261

### DIFF
--- a/common/rbnf/ar.xml
+++ b/common/rbnf/ar.xml
@@ -41,21 +41,21 @@ For terms of use, see http://www.unicode.org/copyright.html
                 <rbnfrule value="70">[→%spellout-numbering→ و]سبعون;</rbnfrule>
                 <rbnfrule value="80">[→%spellout-numbering→ و]ثمانون;</rbnfrule>
                 <rbnfrule value="90">[→%spellout-numbering→ و]تسعون;</rbnfrule>
-                <rbnfrule value="100">مائة[ و →%spellout-numbering→];</rbnfrule>
-                <rbnfrule value="200">مائتان[ و →%spellout-numbering→];</rbnfrule>
-                <rbnfrule value="300">←%spellout-numbering← مائة[ و →%spellout-numbering→];</rbnfrule>
-                <rbnfrule value="1000">ألف[ و →%spellout-numbering→];</rbnfrule>
-                <rbnfrule value="2000">ألفين[ و →%spellout-numbering→];</rbnfrule>
-                <rbnfrule value="3000">←%spellout-numbering← آلاف[ و →%spellout-numbering→];</rbnfrule>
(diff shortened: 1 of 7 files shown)


roozbehp:
Would it be too difficult to add a test or two? Reviewing the rules is a little difficult considering that bidi messes them up.

grhoten:
@roozbehp There currently is no easy way to convert the XML to ICU syntax that I'm aware of. In fact the XML syntax gets in the way. I use https://st.unicode.org/cldr-apps/numbers.jsp to do my testing, and I use RBNFWriter in CLDR to convert the ICU syntax to the XML form. It would be easier to put the test into ICU than to put it into CLDR, but that would require ICU to ingest this data first.  All of the current RBNF tests are located in ICU and not CLDR.

kipcole9:
Thanks very much @grhoten !

srl295 (approved):
Looks like it is just removal of a space.

srl295:
> `all instances of the and (و) surrounded by spaces.` 

I think something may have gotten dropped between `the` and `and`.

srl295:
> @roozbehp There currently is no easy way to convert the XML to ICU syntax that I'm aware of. In fact the XML syntax gets in the way. I use https://st.unicode.org/cldr-apps/numbers.jsp to do my testing, and I use RBNFWriter in CLDR to convert the ICU syntax to the XML form. It would be easier to put the test into ICU than to put it into CLDR, but that would require ICU to ingest this data first. All of the current RBNF tests are located in ICU and not CLDR.

I have CLDR-10884 for making CLDR use the ICU syntax instead of XML, pretty much like what Collation did. (Collation used to have an XML tree structure, for those unfamiliar with that ancient history). Did not get a chance to work on it yet.

srl295:
But that said, we certainly could have a unit test.

srl295:
This is a followon to #1732

srl295:
@grhoten  can you rebase this on maint-41?

macchiati:
Yes, this would be a huge improvement. Shouldn't be that difficult, since
we already convert to that form every time we generate ICU data, so the
engine to do that is already there.

srl295:
@grhoten would you want to comment on [CLDR-10884](https://unicode-org.atlassian.net/browse/CLDR-10884) as to the proposed XML format?

PR #1783, opened 2022-02-25
CLDR-15261 Arabic spellout rules render a stray space after the conjunction

grhoten:
CLDR-15261

This ticket provided additional feedback that the changes were incomplete. I have confirmed that 138000 comes out as the following as per the feedback:
مائة وثمانية وثلاثون ألف وأربعة مائة
Also 1111111111111 comes out as the following:
ترليون ومائة وإحدى عشر مليار ومائة وإحدى عشر مليون ومائة وإحدى عشر ألف ومائة وإحدى عشر

The fix was performed by searching and replacing all instances of the "and" (و) surrounded by spaces. FYI و means "and" in Arabic. Since testing the result requires ICU and there isn't an easy way to convert this XML to ICU syntax, there isn't an easy way to add a test for this value. This is a resubmission of https://github.com/unicode-org/cldr/pull/1782 to the maint-41 branch.


srl295 (approved):
Per https://github.com/unicode-org/cldr/pull/1782#issuecomment-1051073107 and inspection, change is same as #1782 so LGTM

PR #552, opened 2020-07-08
CLDR-13792 fix error messages for bad emoji

macchiati:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-13792
- [ ] Updated PR title and link in previous line to include Issue number

PR #1784, opened 2022-02-28
CLDR-15394 remove invalid test in TestCheckNewRootFailure

pedberg-icu:
CLDR-15394

This is a follow-on to [PR-552](https://github.com/unicode-org/cldr/pull/552), which updated the test CheckNew.handleCheck to make sure that root-style emoji annotations like "E10-836" were not finding their way into non-root data (either explicitly or by inheritance), and added a new test TestCheckCLDR.TestCheckNewRootFailure to verify that the updated CheckNew.handleCheck was working correctly. However, the latter test tried to verify that an error was generated for both "E10-836" and "↑↑↑"; the latter test is not valid, and caused problems in production code, so it is removed here.

For entries like "↑↑↑", the value is obtained by following the parent chain until an explicit value if found (which may be in root, in which case for regular CLDR data it will be an entry like "E10-836". In production data the root entries are removed, so if root is reached the value is null, and CheckNew.handleCheck doe not generate an error (so TestCheckCLDR.TestCheckNewRootFailure did).

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
@@ -550,8 +550,14 @@ public class TestCheckCLDR extends TestFmwk {
         Map<String, String> options = new HashMap<>();
         for (Phase phase : Phase.values()) {
             options.put(Options.Option.phase.getKey(), phase.toString());
-            for (String value : Arrays.asList("E10-836", CldrUtility.INHERITANCE_MARKER)) {
-
+            String value = "E10-836";
+            // The following code used to check values of both "E10-836" and CldrUtility.INHERITANCE_MARKER="↑↑↑",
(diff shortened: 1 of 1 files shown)


pedberg-icu:
But we are already checking the resolved value; in CheckNew.handleCheck we never see "↑↑↑" (from what I aw debugging); we only see a real value, or "E10-836", or (only in production data where root has no value) null. So there is no value for TestCheckCLDR.TestCheckNewRootFailure to separately test "↑↑↑". In regular CLDR data that test is harmless because (the way the test is set up) "↑↑↑" will resolve to "E10-836" and CheckNew.handleCheck produces an error. But in production data, using "↑↑↑" in"yo" will resolve  to null, so CheckNew.handleCheck does not produce an error, so TestCheckCLDR.TestCheckNewRootFailure *does* produce a bogus error.

macchiati:
Ah, I didn't look at the rest of the code. If it is always testing the resolved value, then we should be good.

macchiati (approved):

PR #1785, opened 2022-03-01
CLDR-15398 Fix mistakes

ivan-pan:
CLDR-15398

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -29,7 +29,7 @@ _This is a draft document which may be updated, replaced, or superseded by other
 
 _Please submit corrigenda and other comments with the CLDR bug reporting form [[Bugs](https://cldr.unicode.org/index/bug-reports)]. Related information that is useful in understanding this document is found in the [References](#References). For the latest version of the Unicode Standard see [[Unicode](https://www.unicode.org/versions/latest/)]. For a list of current Unicode Technical Reports see [[Reports](https://www.unicode.org/reports/)]. For more information about versions of the Unicode Standard, see [[Versions](https://www.unicode.org/versions/)]._
 
->**_NOTE: The source for the LDML specification has been converted to Github Markdown (GFM) instead of HTML. The formatting is now simpler, but some features — such as formatting for table captions — may not be complete by the release date. Improvements in the formatting for the v39 specification are planned for after the release, but no substantive changes would be made to the content._**
+>**_NOTE: The source for the LDML specification has been converted to GitHub Markdown (GFM) instead of HTML. The formatting is now simpler, but some features — such as formatting for table captions — may not be complete by the release date. Improvements in the formatting for the v39 specification are planned for after the release, but no substantive changes would be made to the content._**
 
 ## <a name="Parts" href="#Parts">Parts</a>
 
@@ -198,7 +198,7 @@ _**UAX35-C1.**_ An implementation that claims conformance to this specification 
    * For example, an implementation might claim conformance to all LDML features except for _transforms_ and _segments_.
 2. Interpret the relevant elements and attributes of LDML documents in accordance with the descriptions in those sections.
    * For example, an implementation that claims conformance to the date format patterns must interpret the characters in such patterns according to [Date Field Symbol Table](tr35-dates.md#Date_Field_Symbol_Table).
-3. Declare which types of CLDR data that it uses.
+3. Declare which types of CLDR data it uses.
    * For example, an implementation might declare that it only uses language names, and those with a _draft_ status of _contributed_ or _approved_.
 
 _**UAX35-C2.**_ An implementation that claims conformance to Unicode locale or language identifiers shall:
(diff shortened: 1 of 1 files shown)


macchiati (approved):
This all looks great. Thanks for doing this!

macchiati:
@AEApple , @srl295 @pedberg-icu I looked these over and they looked good, and matched the ticket. I added your names in case you also want to review, but as far as I'm concerned we can merge these on Monday.

docs/ldml/tr35.md
```diff
@@ -396,7 +396,7 @@ The different identifiers can be converted to one another as described in this s
 
 ##### <a name="Language_Tag_to_Locale_Identifier" href="#Language_Tag_to_Locale_Identifier">BCP 47 Language Tag to Unicode BCP 47 Locale Identifier</a>
 
-A valid [[BCP47](#BCP47)] language tag can be converted to a valid Unicode BCP 47 locale identifier according to [Annex C. LocaleId Canonicalization](#LocaleId_Canonicalization)
+A valid [[BCP47](#BCP47)] language tag can be converted to a valid Unicode BCP 47 locale identifier according to [Annex C. LocaleId Canonicalization](#LocaleId_Canonicalization).
```
AEApple, line 399: Is there a reason `BCP47` was not updated to `BCP 47` here to be consistent with spacing?

docs/ldml/tr35.md
```diff
@@ -884,13 +884,13 @@ LDML inherits time zone IDs from the tz database [[Olson](#Olson)]. Because thes
...
 
-Although the first two letters of a short identifier may match an ISO 3166 two-letter country code, a user should not assume that the time zone belongs to the country. The first two letters in an identifier of length not equal to 5 has no meaning. Also, the identifiers are stabilized, meaning that they will not change no matter what changes happen in the base standard. So if Hawaii leaves the US and joins Canada as a new province, the short time zone identifier "ushnl" would not change in CLDR even if the UN/LOCODE changes to "cahnl" or something else.
+Although the first two letters of a short identifier may match an ISO 3166 two-letter country code, a user should not assume that the time zone belongs to the country. The first two letters in an identifier of length not equal to 5 have no meaning. Also, the identifiers are stabilized, meaning that they will not change no matter what changes happen in the base standard. So if Hawaii leaves the US and joins Canada as a new province, the short time zone identifier "ushnl" would not change in CLDR even if the UN/LOCODE changes to "cahnl" or something else.
```
AEApple, line 887: `The first two letters in an identifier of length not equal to 5 have no meaning.` seems confusing to me.

Does this mean, `The first two letters in an identifier have no meaning when they are in a short identifier with a length shorter than 5 letters.`? I wasn't sure what the significance of the short identifier having a 5 character length indicates. Or does this mean something else?

macchiati:
The bracketed items are short acronyms, and we usually don't have spaces in
them. (but I don't think we are completely consistent either).

docs/ldml/tr35.md
```diff
@@ -1125,7 +1125,7 @@ In the transformed content 't' data file, the `name` attribute in a `<key>` elem
 ```xml
 <key extension="t" name="m0" description="Transform extension mechanism">
     <type name="ungegn" description="United Nations Group of Experts on Geographical Names" since="21"/>
-<key>
+</key>
```
AEApple, line 1128: @srl295 - Do the tests in GitHub check for missing closing tags in the docs?

docs/ldml/tr35.md
```diff
@@ -1294,7 +1294,7 @@ Hybrid locales have intermixed content from 2 (or more) languages, often with on
...
+While text in a document can be tagged as partly in one language and partly in another, that is not the same having a hybrid locale. There is a difference between having a Spanglish document, and a Spanish document that has some passages quoted in English. Fine-grained tagging doesn't handle grammatical combinations like Tanglish “Enna matteru?” (_What’s the matter?_), which is neither standard Tamil nor standard English. More importantly, it doesn’t work for the very common use case for a [unicode_locale_id](#unicode_locale_id): _locale selection_.
```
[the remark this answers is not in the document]
AEApple: NM, misread this.

AEApple (approved):

docs/ldml/tr35.md
srl295: This actually is body text and not markup. It's in an xml preformatted block. I suppose those could have some kind of linting but that might be challenging.

But generally, no, there isn't any markdown lint, nor html lint against the generated output right now.  There probably would be _errors_ If something was completely broken.

Catching this particular block might be challenging, but can you file a ticket for checking the docs generally if you'd like to do so?

docs/ldml/tr35.md
srl295: - A short identifier of exactly 5 characters is a UN/LOCODE.  So `uslax` is US + LAX (airport). That's the namespace, and the LOCODE always starts with the 3166 country name.
- `utcw01` doesn't follow this, `UT` is not a country code. 
- `usnavajo` has a country code arbitrarily added. It could have been `navajo` (6) as `zznavajo` is too long.

So no it doesn't mean less than five, but less than five OR greater than 5.

srl295 (approved):
LGTM, great catches! 🎉

srl295:
@macchiati do you want to take the PR?

pedberg-icu (approved):

pedberg-icu:
Hmm. This was merged into main, not maint/maint-41. That makes the cherry-picks a bit more complicated.

PR #1789, opened 2022-03-08
CLDR-15404 Adjust code to use en.xml

macchiati:
CLDR-15404

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -27,8 +27,8 @@ import com.ibm.icu.util.ULocale;
 
 /**
  * Rough sketch for now
- * TODO Make classes/methods private that don't need to be public
- * TODO Check for invalid parameters
+ * TODO Mark Make classes/methods private that don't need to be public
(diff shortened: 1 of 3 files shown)


pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -400,6 +446,25 @@ public ParameterMatcher(Set<Length> lengths, Set<Style> styles, Set<Usage> usage
...
+
+        public ParameterMatcher(String lengths, String styles, String usages, String orders) {
+            this.lengths = setFrom(lengths, Length::from);
+            this.styles = setFrom(styles, Style::valueOf);
+            this.usages = setFrom(usages, Usage::valueOf);
+            this.orders = setFrom(orders, Order::valueOf);
+        }
+
+        private <T> Set<T> setFrom(String parameter, Function<String, T> func) {
+            if (parameter == null || parameter.isBlank()) {
```
srl295, line 458: @macchiati Maybe not the first, but first I've seen of JDK11 APIs in CLDR…  [isBlank()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank())

PR #1790, opened 2022-03-09
CLDR-15021 v41 Add COMMIT_METADATA.md with commits to skip

srl295:
CLDR-15021

- add a .github/COMMIT_METADATA.md
- Update commit checker workflow
- ~Also temporarily use the srl295 branch of the commit checker~ no, you will have to manually put this in.
- Fixes the hardcoded and increasingly-inaccurate `v38` in the report URL!

### DIFF
--- a/.github/COMMIT_METADATA.md
+++ b/.github/COMMIT_METADATA.md
@@ -1,0 +1,20 @@
+# Commit Metadata
+
+### The following list of commits is parsed by the ICU/CLDR commit checker.
+### It has the following structure:
+### `- <commit> <bug id or hyphen> <message>`
+
+- 00decafbad CLDR-0000 Bad commit message
+- 50257a7a3eba7bbed634961a2aa74e77b12810bf - Early commit, no bug id
+- 283dc84d81cfc47fb15cd664fce4b0151d070ea6 - Early commit, no bug id
+
+### The following are items to skip for a certain CLDR version
+### Format: `# SKIP v00` followed by a list of commits to skip for that version
+
+# SKIP v41
+
+- 56ca5d5 CLDR-14877 split ticket
+- 56ca5d563cf57990a7598f570cb9be51956cb9de - v40 commit
+- 382f36702d763713f010a3bf87d34605c0226ebd - v40 commit
+- bc8a79f9b1b71b00c716b8b4c8e0185515a6d0b8 - v40 commit
+- 4b87f2d718dd9cab45517ebe7f996ac10b6b1fff - v40 commit
--- a/.github/workflows/commit-checker.yml
+++ b/.github/workflows/commit-checker.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           echo "::group::Setup Checker"
           END_HASH=$(cd ${GITHUB_WORKSPACE}/cldr ; git rev-parse --short HEAD )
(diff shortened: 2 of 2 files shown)


srl295:
note SPECIAL PARAMS:

image

srl295:
run in progress… https://github.com/unicode-org/cldr/actions/runs/1959452528

AEApple:
- Can we add the title of the excluded commits?
- Can we add a link to where to add commits to be excluded in GitHub under the 'Excluded commits' section in the Commit Checker report to make it easy for people to find in the future?

srl295:
> * Can we add the title of the excluded commits?
> * Can we add a link to where to add commits to be excluded in GitHub under the 'Excluded commits' section in the Commit Checker report to make it easy for people to find in the future?

I'll work on adding the title. 

As to "where to add commits to be excluded" - yes, I think I can add a link for now. But probably we should consider this just the v1.0 of this feature, so it may change in the future.

AEApple:
> As to "where to add commits to be excluded" - yes, I think I can add a link for now. But probably we should consider this just the v1.0 of this feature, so it may change in the future, since this tool is shared with ICU.

Makes sense!

AEApple:
Looks like it works well to me: https://cldr-smoke.unicode.org/cldrcc/CLDR-Report-2022-03-15-v41-3813592f96.html

What else still remains for this?

AEApple (approved):
Looks good to me once it is rebased. Feel free to skip moving my test commits over. I can always add them again later once the change merges into main.

srl295:
> Looks like it works well to me: https://cldr-smoke.unicode.org/cldrcc/CLDR-Report-2022-03-15-v41-3813592f96.html
> 
> 
> 
> What else still remains for this?

That's all!

.github/workflows/commit-checker.yml
```diff
@@ -17,11 +17,11 @@ on:
       tool-repository:
         description: Commit Checker ICU repo
         required: true
-        default: 'unicode-org/icu'
+        default: 'srl295/icu'
```
srl295, line 20: @AEApple maybe we don't want to commit this to maint. Would it be too much hassle to type in srl295 each time?

AEApple: Sure, as long as we're sure it's running against maint-41 for the actual commit check and only using the srl925 for the list of commits to skip.

srl295:
Updated parameters: 
image

.github/COMMIT_METADATA.md
```diff
@@ -0,0 +1,20 @@
...
+### The following are items to skip for a certain CLDR version
+### Format: `# SKIP v00` followed by a list of commits to skip for that version
+
+# SKIP v41
+
+- 56ca5d5 CLDR-14877 split ticket
+- 56ca5d563cf57990a7598f570cb9be51956cb9de - v40 commit
+- 382f36702d763713f010a3bf87d34605c0226ebd - v40 commit
+- bc8a79f9b1b71b00c716b8b4c8e0185515a6d0b8 - v40 commit
+- 4b87f2d718dd9cab45517ebe7f996ac10b6b1fff - v40 commit
```
srl295, line 20: @AEApple  things to skip need to go under the SKIP v41 section

PR #1792, opened 2022-03-11
CLDR-15405 PersonNames coverage, PathHeader, PathDescription

pedberg-icu:
CLDR-15405

Fill out coverage, PathHeader, and PathDescription for Person Name Formats data (but not yet AttributeValueValidity). I did not find any place in PathHeader where I needed to do something special for the nameFormat placeholders.

Note that as part of this I created a stub translation guidelines page for Person Name Formats, since PathDescription needed to refer to it: https://cldr.unicode.org/translation/miscellaneous-person-name-formats


macchiati (approved):
Looks good. Later we can refine to use a function to replace the spaces in the attributes. And add test that the values are in order.

PR #1795, opened 2022-03-14
CLDR-15420 document num&alpha3 codes for CLDR region subtags

markusicu:
CLDR-15420


markusicu:
Please ignore... replacement: #1796

PR #1796, opened 2022-03-14
CLDR-15420 document num&alpha3 codes for CLDR region subtags

markusicu:
CLDR-15420

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -550,17 +550,18 @@ The private use subtags listed as **excluded** in _Section 3.5.3 [Private Use Co
 
 Subtags in the region.xml file (see _Section 3.11 [Validity Data](#Validity_Data)_). These are based on [[BCP47](#BCP47)] subtag values marked as **Type: region**
 
-Unicode identifiers give specific semantics to the following subtags:
+Unicode identifiers give specific semantics to the following subtags.
+(The alpha2 codes are used as Unicode region subtags. The alpha3 and numeric codes are derived according to _Section 3.5.2 [Numeric Codes](#Numeric_Codes)_ and listed here for additional documentation.)
 
-|      | Name                         | Comment | ISO 3166-1 status |
-| ---- | ---------------------------- | ------- | ----------------- |
-| `QO` | Outlying Oceania             | countries in Oceania [009] that do not have a [subcontinent](https://unicode-org.github.io/cldr-staging/charts/38/supplemental/territory_containment_un_m_49.html). | private use |
-| `QU` | European Union               | **deprecated**: the _canonicalized_ form is EU | private use |
-| `UK` | United Kingdom               | **deprecated**: the _canonicalized_ form is GB | exceptionally reserved |
-| `XA` | Pseudo-Accents               | special code indicating derived testing locale with English + added accents and lengthened | private use |
-| `XB` | Pseudo-Bidi                  | special code indicating derived testing locale with forced RTL English | private use |
-| `XK` | Kosovo                       | industry practice | private use |
-| `ZZ` | Unknown or Invalid Territory | used in APIs or as replacement for invalid code | private use |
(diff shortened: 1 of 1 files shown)


srl295 (approved):
I'm approving this for v41.

As I noted in the ticket, I really do not like this kind of data living in the spec at all.  It should be a vector to appropriate charts, which could have appropriate highlighting on data items which are private use.

srl295:
this easily gets out of date if the metadata changes but not the spec, or if someone navigates to the wrong spec version

markusicu:
> this easily gets out of date if the metadata changes but not the spec

I don't think it *can* get out of date: AFAIU the relationship between the alpha2/alpha3/numeric codes for private use region subtags is fixed and algorithmic. Besides, codes are stable. At most one of these will be deprecated (like QU has been).

> I really do not like this kind of data living in the spec at all. It should be a vector to appropriate charts, which could have appropriate highlighting on data items which are private use.

I agree in principle, but when someone is looking for a question like "what is the Unicode/CLDR alpha3 code for Kosovo?" then the table that I am pointing to seems like the place they could find, and that place should give them the answer.

docs/ldml/tr35.md
```diff
@@ -550,17 +550,18 @@ The private use subtags listed as **excluded** in _Section 3.5.3 [Private Use Co
...
+
+| alpha2 | alpha3 | num | Name                         | Comment | ISO 3166-1 status |
+| ------ | ------ | --- | ---------------------------- | ------- | ----------------- |
+| `QO`   | `QOO`  | 961 | Outlying Oceania             | countries in Oceania [009] that do not have a [subcontinent](https://unicode-org.github.io/cldr-staging/charts/38/supplemental/territory_containment_un_m_49.html). | private use |
+| `QU`   | `QUU`  | 967 | European Union               | **deprecated**: the _canonicalized_ form is EU | private use |
+| `UK`   | -      | -   | United Kingdom               | **deprecated**: the _canonicalized_ form is GB | exceptionally reserved |
+| `XA`   | `XAA`  | 973 | Pseudo-Accents               | special code indicating derived testing locale with English + added accents and lengthened | private use |
+| `XB`   | `XBB`  | 974 | Pseudo-Bidi                  | special code indicating derived testing locale with forced RTL English | private use |
+| `XK`   | `XKK`  | 983 | Kosovo                       | industry practice | private use |
```
pedberg-icu, line 563: Note that ICU has been using XXK for Kosovo (per different industry practice), as I mentioned in the ticket: https://unicode-org.atlassian.net/browse/ICU-21341?focusedCommentId=160540

markusicu: > Note that ICU has been using XXK for Kosovo (per different industry practice)

Well, I don't know what practice that is. [CLDR specifies](https://www.unicode.org/reports/tr35/#Numeric_Codes): “Unicode identifiers supply a standard mapping to these: (...) for the 3-letter codes it doubles the final letter.”

And https://github.com/unicode-org/cldr/blob/main/common%2Fsupplemental%2FsupplementalMetadata.xml#L859 has `<territoryAlias type="XKK" replacement="XK" reason="overlong"/> <!-- Kosovo -->` but no mapping for XXK.

In your comments on ICU-21341 you referred to CLDR for the Kosovo codes. Therefore, as CLDR specifies XKK, I have to assume that XXK was a typo that we should fix in ICU. --> new bug ICU-21942

Clearer documentation here in the LDML spec might have helped...

pedberg-icu: OK, thanks, guess I will file a ticket to update ICU

pedberg-icu (approved):

docs/ldml/tr35.md
markusicu: > OK, thanks, guess I will file a ticket to update ICU

I did: new bug ICU-21942

pedberg-icu: I took that ticket and have a PR: https://github.com/unicode-org/icu/pull/2034

PR #1798, opened 2022-03-15
CLDR-15404 add compact and expand

macchiati:
CLDR-15404

This moves some preliminary code for expand/compact into the code file.

The compaction also adds an ANY case, for the inputs that have the largest number of name pattern duplicates.

```
COMPACTED:
1)	{length='medium short' style='formal' usage='addressing'}
⇒		︎"{given} {given2-initial}. {surname}"
		︎"{given} {surname}"
2)	ANY
⇒		︎"{prefix} {given} {given2} {surname} {surname2} {suffix}"

``` 

That doesn't do any good in some cases, as in the current English rules.
```
COMPACTED:
1)	{length='monogram-narrow' style='informal'}
⇒		︎"{given-initial-informal}"
2)	{length='long medium short' style='informal' usage='addressing'}
⇒		︎"{given-informal}"
3)	{length='short' style='formal' usage='referring'}
⇒		︎"{given-initial}. {given2-initial}. {surname}"
4)	{length='monogram' style='formal'}
⇒		︎"{given-initial}{given2-initial}{surname-initial}"
5)	{length='short' style='formal' usage='sorting'}
⇒		︎"{surname}, {given-initial}. {given2-initial}."
6)	{length='long' style='formal' usage='referring'}
⇒		︎"{given} {given2} {surname} {suffix}"
7)	{length='medium' style='formal' usage='referring'}
⇒		︎"{given} {given2-initial}. {surname} {suffix}"
8)	{length='long' style='formal' usage='sorting'}
⇒		︎"{surname}, {given} {given2} {suffix}"
9)	{length='medium' style='formal' usage='sorting'}
⇒		︎"{surname}, {given} {given2-initial}. {suffix}"
10)	{length='monogram' style='informal'}
⇒		︎"{given-initial-informal}{surname-initial}"
11)	{length='long medium' style='informal' usage='referring'}
⇒		︎"{given-informal} {surname}"
12)	{length='long medium short' style='informal' usage='sorting'}
⇒		︎"{surname}, {given-informal}"
13)	{length='short' style='informal' usage='referring'}
⇒		︎"{given-informal} {surname-initial}. {suffix}"
14)	{length='monogram-narrow' style='formal'}
⇒		︎"{surname-initial}"
15)	ANY
⇒		︎"{prefix} {surname}"
```

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -16,6 +16,7 @@ import java.util.TreeSet;
 import java.util.function.Function;
 
 import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.Counter;
 import org.unicode.cldr.util.XPathParts;
 
 import com.google.common.base.Joiner;
@@ -26,10 +27,14 @@ import com.google.common.collect.ImmutableBiMap;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
+import com.google.common.collect.TreeMultimap;
 import com.ibm.icu.impl.Pair;
+import com.ibm.icu.impl.Row.R2;
 import com.ibm.icu.util.ULocale;
 
 /**
@@ -49,6 +54,8 @@ public class PersonNameFormatter {
         surname,
         surname2,
         suffix;
+        public static final Comparator<Iterable<Field>> ITERABLE_COMPARE = Comparators.lexicographical(Comparator.<Field>naturalOrder());
+        public static final Set<Field> ALL = ImmutableSet.copyOf(Field.values());
     }
 
     public enum Length {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -553,22 +618,40 @@ public static ParameterMatcher from(String string) {
...
+            showAttributes("style", styles, items);
+            showAttributes("usage", usages, items);
+            showAttributes("order", orders, items);
+            return items.isEmpty() ? "ANY" : "{" + JOIN_SPACE.join(items) + "}";
+        }
+
+        public String toLabel() {
+            StringBuilder sb  = new StringBuilder();
+            addToLabel(lengths, sb);
+            addToLabel(lengths, sb);
```
pedberg-icu, line 631: shouldn't this be styles, not lengths? And the next ones usages and orders?

macchiati: good catch fixed this. Also fixed the identifiers

macchiati: You can just look at the commits after this one for changes:
https://github.com/unicode-org/cldr/pull/1798/commits/f64d51655f40a7324819a72d145d345fe5410849
https://github.com/unicode-org/cldr/pull/1798/commits/11fae69c6d9e8d6c3157dd4e8388f9a0f4e5e6d1

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -581,6 +681,86 @@ public boolean equals(Object obj) {
...
+            return ComparisonChain.start()
+                .compare(lengths, other.lengths, Length.ITERABLE_COMPARE)
+                .compare(styles, other.styles, Style.ITERABLE_COMPARE)
+                .compare(usages, other.usages, Usage.ITERABLE_COMPARE)
+                .compare(orders, other.orders, Order.ITERABLE_COMPARE)
+                .result();
+        }
+
+        public ParameterMatcher merge(ParameterMatcher that) {
+            // if 3 of the 4 rows are equal, merge the other two
```
wasker, line 696: Nit: "rows" -> "attributes"; "the other two" -> "the two".

wasker, line 696: We do not anticipate that all 4 attributes match?

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -581,6 +681,86 @@ public boolean equals(Object obj) {
...
+                    styles.equals(Style.ALL) ? ImmutableSet.of() : styles,
+                        usages.equals(Usage.ALL) ? ImmutableSet.of() : usages,
+                            orders.equals(Order.ALL) ? ImmutableSet.of() : orders
+                );
+        }
+
+        /**
+         * Compact a set of FormatParameters into a (smaller) collection of ParameterMatchers
+         */
+        public static Set<ParameterMatcher> compact(Collection<FormatParameters> expanded) {
```
wasker, line 729: I think it's worth noting that the order of matchers (whether it's important or not) is not preserved.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -581,6 +681,86 @@ public boolean equals(Object obj) {
...
+                                result.remove(item2);
+                                result.add(item12);
+                                continue main; // retry everything
+                            }
+                        }
+                    }
+                    break;
+                }
+
+            // now replace any "complete" items by empty.
```
wasker, line 756: Does it really need to happen? IMO, if you to keep attribute sets non-empty, this will simplify logic of the matcher: no need to treat "empty set" in a special way, just check if the item is in the set.

The slimming would probably make sense if you used this logic to generate XML or something, but if it's used only as an internal structure, I'd rather not do that.

wasker (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
macchiati: I went ahead and merged this, so that you can go ahead and rebase on it. I'll look at your comments in a subsequent PR.

PR #1800, opened 2022-03-16
CLDR-15446 Replace FixedDecimal with DecimalQuantity for plural rule samples

echeran:
CLDR-15446

Instructions to reproduce:

1. Switch to your working copy of ICU, and then...
2. Checkout the [PR #2007](https://github.com/unicode-org/icu/pull/2007) that does the ICU side of work. You can use the [Github CLI tool](https://cli.github.com/) to do this with the command `gh pr checkout 2007`
3. Build the ICU4J jar that also includes CLDR-specific dependencies: `cd icu4j`;  `ant clean`; `ant -noinput releaseCLDR -f build.xml`
4. Give that new jar file a unique testing-specific version number and register ("install") it to your local file system's Maven cache of artifacts (jar files).  For a version string `71.0.1-SNAPSHOT-icu-pr-2007`, we would install it with the proper Maven groupId, artifactId coordinates as:  `mvn install:install-file -Durl=file:repo -DgroupId=com.ibm.icu -DartifactId=icu4j-for-cldr -Dversion=71.0.1-SNAPSHOT-icu-pr-2007 -Dpackaging=jar -Dfile=icu4j.jar -Dsources=icu4j-src.jar`
5. Switch to your working copy of CLDR, and then...
6. Checkout this PR.  Ex: `gh pr checkout 1800`
7. Edit the file `tools/pom.xml`, so that in the `<dependency>` entry for `<artifactId>icu4j-for-cldr</artifactId>`, update the version so that it reads as `<version>71.0.1-SNAPSHOT-icu-pr-2007</version>`
8. Run tests via `mvn -s .github/workflows/mvn-settings.xml -B compile install package --file tools/pom.xml`

### DIFF
--- a/common/supplemental/ordinals.xml
+++ b/common/supplemental/ordinals.xml
@@ -1,9 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE supplementalData SYSTEM "../../common/dtd/ldmlSupplemental.dtd">
 <!--
-Copyright © 1991-2015 Unicode, Inc.
-CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+Copyright © 1991-2022 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
 -->
 <supplementalData>
     <version number="$Revision$"/>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/BestMinimalPairSamples.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/BestMinimalPairSamples.java
@@ -30,11 +30,12 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.TreeMultimap;
 import com.ibm.icu.impl.locale.XCldrStub.ImmutableMap;
+import com.ibm.icu.impl.number.DecimalQuantity;
 import com.ibm.icu.text.DecimalFormat;
 import com.ibm.icu.text.PluralRules;
-import com.ibm.icu.text.PluralRules.FixedDecimal;
-import com.ibm.icu.text.PluralRules.FixedDecimalRange;
-import com.ibm.icu.text.PluralRules.FixedDecimalSamples;
+import com.ibm.icu.text.PluralRules.DecimalQuantitySamples;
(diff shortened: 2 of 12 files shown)


macchiati (changes requested):
Before this goes in, the plural data file should be regenerated to verify that none of the samples change because of this.

echeran:
note to self: these are the CLDR instructions for regenerating the plural data: https://cldr.unicode.org/development/updating-codes/addingfixing-plural-and-ordinals

echeran:
@srl295 I finished off the checklist you presented, after first making more code changes to address the remaining test failures. 

Tests pass now, but of course, they need to be run against the pending change to ICU, so you can only do that locally.  I am following the steps that I listed above in the PR description to reproduce that proper CLDR-ICU dependency during tests with in-flight development in both places. I believe those testing steps to be correct, but please lmk if you disagree.

I continue to re-run this PR's tests after any changes to the "upstream" ICU PR https://github.com/unicode-org/icu/pull/2007.

@macchiati I ran the instructions you provided to update the plural rules and included the diffs in this PR.

Also, the diffs from the auto-generated output included updates to the file header, which updates the things like the year in the copyright notice. I included that as well in this PR -- is that what you all prefer?

I see that the Jira ticket is not accepted. What is the process for that, could that be handled in the next weekly TC meeting?

Ultimately, though, I recognize this PR can't be merged as-is right now.  Instead, only after the ICU PR is merged and a new snapshot is created for icu4j-for-cldr, then the next time that @pedberg-icu or someone updates that snapshot version in CLDR, that version update will want to be combined with these corresponding code changes.  But since the ICU PR won't make be included for the 71 release, we will have to wait until we're ready for new v72 changes to begin the process.

common/supplemental/pluralRanges.xml
```diff
@@ -99,6 +98,23 @@
 			<pluralRange start="other" end="other" result="other"/>
 		</pluralRanges>
 
+<!-- one, many, other -->
+		<pluralRanges locales="fr pt">
+			<pluralRange start="one"   end="one"   result="one"/>
```
macchiati, line 103: We'll need to check this see if there are difference for the 'many' case.

macchiati (commented):
Changes look good. I didn't see any changes in the samples in the rule files, which is great.

macchiati:
I put the PR on the agenda for Monday. I have no doubt that it will be accepted.

Agreed that we can't merge this until the version of ICU it depends on is merged in.

macchiati (approved):

macchiati:
Looks good except for the failing tests.

markusicu (approved):
lgtm

PR #1803, opened 2022-03-17
CLDR-15115 explain inheritance with multiple variants

macchiati:
CLDR-15115

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -1357,7 +1357,7 @@ In version 28.0, the subdivisions in the validity files used the ISO format, upp
 
 The XML format relies on an inheritance model, whereby the resources are collected into _bundles_, and the bundles organized into a tree. Data for the many Spanish locales does not need to be duplicated across all of the countries having Spanish as a national language. Instead, common data is collected in the Spanish language locale, and territory locales only need to supply differences. The parent of all of the language locales is a generic locale known as _root_. Wherever possible, the resources in the root are language & territory neutral. For example, the collation (sorting) order in the root is based on the [[DUCET](#DUCET)] (see _[Root Collation](tr35-collation.md#Root_Collation)_). Since English language collation has the same ordering as the root locale, the 'en' locale data does not need to supply any collation data, nor do the 'en_US', 'en_GB' or the any of the various other locales that use English.
 
-Given a particular locale id "en_US_someVariant", the search chain for a particular resource is the following.
+Given a particular locale id "en_US_someVariant", the default search chain for a particular resource is the following.
 
 ```
 en_US_someVariant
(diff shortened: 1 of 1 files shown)


richgillam (approved):
Looks good and addresses the ticket.  I'm assuming I was added to the review list because of the previous discussions we'd had about resource fallback, and specifically--
1) Adding the default script if the script isn't specified.
2) Using the parentLocales resource (and not just %%Parent directives in the resource files) to determine the parent.
Neither of these issues is addressed here, although the second one is alluded to.  Are those separate tickets?  What is their status?

macchiati:
#1. There isn't a ticket for adding the default script, so I wasn't sure
that everyone is in agreement on that. (I'm all in favor, BTW) Can you file
a CLDR ticket for that and we can try to fast track it? (We have until
Monday)

#2. That's an ICU issue, not CLDR (which already uses parentLocale).

richgillam:
@macchiati 

>There isn't a ticket for adding the default script, so I wasn't sure that everyone is in agreement on that. (I'm all in favor, BTW) Can you file a CLDR ticket for that and we can try to fast track it? (We have until Monday)

There already is one, authored by you: https://unicode-org.atlassian.net/browse/CLDR-15265  For some reason, it's assigned to me, which might have been the problem.

>That's an ICU issue, not CLDR (which already uses parentLocale).

Okay.  Thanks!

docs/ldml/tr35.md
```diff
@@ -1368,6 +1368,23 @@ root
 
 _The inheritance is often not simple truncation, as will be seen later in this section._
 
+The default search chain is slighly different for multiple variants. 
+In that case, the inheritance chain covers all combinations of variants, with longest number of variants first. 
+For example, where the requested locale ID is en_fonipa_scouse, the inheritance chain is as follows:
+
+```
+en_GB_fonipa_scouse
+en_GB_scouse_fonipa // extra step, only needed if not canonical
```
sffc, line 1377: Question: which is canonical: "en_GB_fonipa_scouse" or "en_GB_scouse_fonipa"? I thought "en_GB_fonipa_scouse" was canonical because the variants are sorted.

docs/ldml/tr35.md
```diff
@@ -1368,6 +1368,23 @@ root
 
 _The inheritance is often not simple truncation, as will be seen later in this section._
 
+The default search chain is slighly different for multiple variants. 
+In that case, the inheritance chain covers all combinations of variants, with longest number of variants first.
```
sffc, line 1372: Question: How do you sort the combinations?

In your example below, how do you know "en-GB-fonipa" comes before "en-GB-scouse"?

sffc (changes requested):

docs/ldml/tr35.md
sffc: The ticket says "use alphabetical order". The spec should say this too.

macchiati:
> There already is one, authored by you: https://unicode-org.atlassian.net/browse/CLDR-15265 For some reason, it's assigned to me, which might have been the problem.

It was blocked on the ICU ticket [ICU-21125](https://unicode-org.atlassian.net/browse/ICU-21125), and thus set for the next release.

docs/ldml/tr35.md
macchiati: Yes, en_GB_fonipa_scouse is the canonical one, and checked first. en_GB_scouse_fonipa only needs to be checked if the implementation allows non-canonical locale names in its data (as explained right below). Does that need to be clarified? 

If the data for the implementation performing the inheritance doesn't require canonical locale identifiers, then extra locale IDs need to be inserted in the chain. That is indicated in the example above, marked with "only needed if not canonical".

docs/ldml/tr35.md
macchiati: The spec does say so. See https://github.com/unicode-org/cldr/blob/8d829bbb16598868fc536555c4cca1c626a0d550/docs/ldml/tr35.md#5-canonicalizing-syntax

"Put any variants into alphabetical order (eg, en-fonipa-scouse, not en-scouse-fonipa)"
Do you want a pointer to that section here?

macchiati:
Added a couple of lines to address Shane's comments.

srl295 (approved):

sffc (approved):

sffc (approved):

PR #1804, opened 2022-03-17
CLDR-14927 Explain TZDB links and aliases

macchiati:
CLDR-14927


macchiati:
Can one of you review this?

srl295 (approved):

srl295:
@macchiati  looks like 'check' was not required for merge.  Can you squash it so that it attempts a rebuild?

srl295:
looks like tests didn't run, probably due to github issues yesterday

macchiati:
Steven, the "check Expected — Waiting for status to be reported" appears to be hung. How can we unblock it?

srl295:
> Steven, the "check Expected — Waiting for status to be reported" appears to be hung. How can we unblock it?

I suggested above squashing it. Basically you need to make some commit change to it so that it rebuilds.

PR #1808, opened 2022-03-18
CLDR-15378 revise language on coverage in the spec

macchiati:
CLDR-15378

* The more extensive bits of Basic were merged into Moderate.
* Some information was removed, now points to the "Core Data" Sites page instead. That way we can keep it updated more easily.

### DIFF
--- a/docs/ldml/tr35-info.md
+++ b/docs/ldml/tr35-info.md
@@ -486,18 +486,22 @@ That is, this data provides recommended fallbacks for use when a charset or supp
 
 ## 8 <a name="Coverage_Levels" href="#Coverage_Levels">Coverage Levels</a>
 
-The following describes the coverage levels used for the current version of CLDR. This list will change between releases of CLDR. Each level adds to what is in the lower level.
+The following describes the structure used to set coverage levels used for CLDR. 
+That structure is primarily intended for internal use in CLDR tooling — it is not anticipated that users of CLDR data would need it.
+
+Each level adds to what is in the lower level. This list will change between releases of CLDR, and more detailed information for each level is on [Coverage Levels](https://cldr.unicode.org/index/cldr-spec/coverage-levels).
+
 
 | Level | Description   |     |
 | ----: | ------------- | --- |
 | 0     | undetermined  | Does not meet any of the following levels. |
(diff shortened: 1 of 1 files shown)


macchiati:
Rerunning the build, because it appears that something got stuck.

srl295:
> Rerunning the build, because it appears that something got stuck.

same here. can you squash? looks like the test never fired off

docs/ldml/tr35-info.md
```diff
@@ -486,18 +486,22 @@ That is, this data provides recommended fallbacks for use when a charset or supp
 
 ## 8 <a name="Coverage_Levels" href="#Coverage_Levels">Coverage Levels</a>
 
-The following describes the coverage levels used for the current version of CLDR. This list will change between releases of CLDR. Each level adds to what is in the lower level.
+The following describes the structure used to set coverage levels used for CLDR. 
+That structure is primarily intended for internal use in CLDR tooling — it is not anticipated that users of CLDR data would need it.
+
+Each level adds to what is in the lower level. This list will change between releases of CLDR, and more detailed information for each level is on [Core Data](https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales).
```
AEApple, line 492: Do we need to rename the page and update the link? I think calling the page 'Core Data for new locales' may cause people to not realize that it is about all the coverage levels when they're in sites.

I would recommend a 'Coverage Levels' page if we don't have one already and then a subheading for core data for new locales.

AEApple: Propose using https://cldr.unicode.org/index/survey-tool/coverage as the main page about coverage levels instead and/or putting one about coverage levels under the /cldr-spec/ directory instead of under survey tool. Happy to create the page today if needed so there is something at that link.

AEApple: Per discussion created https://cldr.unicode.org/index/cldr-spec/coverage-levels please use this instead

AEApple (changes requested):

AEApple (approved):

AEApple:
I updated the link to the new page as discussed you will have to squash into a new commit since I was not able to squash the two commits, and I missed updating the commit message. Sorry!

AEApple (approved):

PR #1809, opened 2022-03-18
CLDR-15369 Priority Items Summary, save in db; manager access; org-neutral

btangmu:
-New SurveySnapshotDb.java; json.status.userCanCreateSummarySnapshot

-New table SUMMARY_SNAPSHOTS, with LONGBLOB for json data

-Managers can view snapshots; only Admin can create them

-Org-neutral level for each locale per StandardCodes.getLocaleCoverageLevel

-Move help message to front end; fix it for org-neutral levels; remove bogus instructions link

-Move progress bar from back end to front end Vue; json has message and percent instead of ret

-Revise GUI: only show Stop button when there is something to stop

-Change button label from Refresh to Create New Summary, shown when Stop is not shown

-Minor refactoring; move some string literals out of getPriorityItemsSummaryOutput

-Remove unused method openConsoleProgress

CLDR-15369

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrPriorityItems.js
+++ b/tools/cldr-apps/js/src/esm/cldrPriorityItems.js
@@ -45,20 +45,34 @@ class SummaryArgs {
 
 let latestArgs = new SummaryArgs();
 
+let canSum = false;
 let canSnap = false;
+let canCreateSnap = false;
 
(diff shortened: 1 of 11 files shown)


srl295 (approved):
LGTM.  Seems to be an update of #1799 and #1807

PR #1811, opened 2022-03-19
CLDR-15405 Fully enumerated paths in root (with aliases) and en; namePattern no longer @ORDERED

pedberg-icu:
CLDR-15405

Changes from what we discussed in meeting:
- We had said the monogram lengths should ignore attributes except for order. But we need style too, as shown in the English examples in our own spec.
- We had said the data paths would include all attribute combinations, and where we were ignoring one (like usage for monogram lengths, or order for sorting), we would alias one to the other and use coverage to exclude the non-used one. But that does not make much sense; better to just have the fleshed-out paths in xml data not include the ignored attribute value, then no aliasing necessary.
- I had to add the English data for order=surnameFirst, that was not in the spec.
- I have not yet added the patterns for initial and initial sequence.

dtd changes (affects other items like data, coverage, PathHeader)
- length attibute value `monogram-narrow` -> `monogramNarrow`
- namePattern element is not longer `@ORDERED`, instead can have alt="2" or alt="3" (oops, I guess this was supposed to be "1" or "2", will fix in next round)

data changes
- root: Add all paths with non-ignored attributes, and alias to fallback path.
- en: Re-order and flesh out all paths, including adding new data for oder=surnameFirst.

other
- Update coverage, PathDescription, tests, etc.

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9129,50 +9129,122 @@ annotations.
 		<featureName type="zero">slashed zero</featureName>
 	</typographicNames>
 	<personNames>
-		<personName length="monogram-narrow" style="informal">
-			<namePattern>{given-informal-initial}</namePattern>
-		</personName>
-		<personName length="monogram-narrow">
-			<namePattern>{surname-initial}</namePattern>
-		</personName>
-		<personName length="monogram" style="informal">
-			<namePattern>{given-informal-initial}{surname-initial}</namePattern>
-		</personName>
-		<personName length="monogram">
-			<namePattern>{given-initial}{given2-initial}{surname-initial}</namePattern>
-		</personName>
-		<personName usage="addressing" style="informal">
-			<namePattern>{given-informal}</namePattern>
-		</personName>
-		<personName usage="addressing">
-			<namePattern>{prefix} {surname}</namePattern>
-		</personName>
-		<personName usage="sorting" style="informal">
-			<namePattern>{surname}, {given-informal}</namePattern>
-		</personName>
-		<personName length="long" usage="sorting">
-			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
-		</personName>
-		<personName length="medium" usage="sorting">
-			<namePattern>{surname}, {given} {given2-initial}. {suffix}</namePattern>
-		</personName>
-		<personName length="short" usage="sorting">
-			<namePattern>{surname}, {given-initial}. {given2-initial}.</namePattern>
-		</personName>
-		<personName length="long" usage="referring" style="formal">
-			<namePattern>{given} {given2} {surname} {suffix}</namePattern>
-		</personName>
-		<personName length="medium" usage="referring" style="formal">
-			<namePattern>{given} {given2-initial}. {surname} {suffix}</namePattern>
-		</personName>
-		<personName length="short" usage="referring" style="formal">
-			<namePattern>{given-initial}. {given2-initial}. {surname}</namePattern>
-		</personName>
-		<personName length="short" usage="referring" style="informal">
-			<namePattern>{given-informal} {surname-initial}. {suffix}</namePattern>
-		</personName>
 		<personName>
 			<namePattern>{given-informal} {surname}</namePattern>
+		</personName>
+		<personName length="long" usage="addressing" style="formal" order="surnameFirst">
+			<namePattern>{prefix} {surname}</namePattern>
+		</personName>
+		<personName length="long" usage="addressing" style="formal" order="givenFirst">
+			<namePattern>{prefix} {surname}</namePattern>
+		</personName>
+		<personName length="long" usage="addressing" style="informal" order="surnameFirst">
+			<namePattern>{given-informal}</namePattern>
+		</personName>
+		<personName length="long" usage="addressing" style="informal" order="givenFirst">
+			<namePattern>{given-informal}</namePattern>
+		</personName>
+		<personName length="long" usage="referring" style="formal" order="surnameFirst">
+			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
+		</personName>
+		<personName length="long" usage="referring" style="formal" order="givenFirst">
+			<namePattern>{given} {given2} {surname} {suffix}</namePattern>
+		</personName>
+		<personName length="long" usage="referring" style="informal" order="surnameFirst">
+			<namePattern>{surname}, {given-informal}</namePattern>
+		</personName>
+		<personName length="long" usage="referring" style="informal" order="givenFirst">
+			<namePattern>{given-informal} {surname}</namePattern>
+		</personName>
+		<personName length="long" usage="sorting" style="formal">
+			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
+		</personName>
+		<personName length="long" usage="sorting" style="informal">
+			<namePattern>{surname}, {given-informal}</namePattern>
+		</personName>
+		<personName length="medium" usage="addressing" style="formal" order="surnameFirst">
+			<namePattern>{prefix} {surname}</namePattern>
+		</personName>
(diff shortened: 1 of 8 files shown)


common/supplemental/coverageLevels.xml
```diff
@@ -904,23 +906,17 @@ For terms of use, see http://www.unicode.org/copyright.html
...
-		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/personName[@usage='%anyAttribute']/namePattern[@_q='%anyAttribute']"/>
-		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/personName[@style='%anyAttribute'][@order='%anyAttribute']/namePattern[@_q='%anyAttribute']"/>
-		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/personName[@style='%anyAttribute']/namePattern[@_q='%anyAttribute']"/>
-		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/personName[@order='%anyAttribute']/namePattern[@_q='%anyAttribute']"/>
-		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/personName/namePattern[@_q='%anyAttribute']"/>
+		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/nameOrder[@nameLocales='%anyAttribute']"/>
+		<coverageLevel inLanguage="%personNameLanguages" value="modern" match="personNames/personName[@length='%personNameNonMonograms'][@usage='addressing'][@style='%anyAttribute'][@order='%anyAttribute']/namePattern[@alt='%anyAttribute']"/>
```
macchiati, line 910: We have two situations:
In our source, we will have fully qualified. So just lines 911 and 912.

In the production data, we would have to have all the combinations

- length, usage, style, order
- length, usage, style
- length, usage, order
- length, style, order
- usage, style, order
- length, usage
- length, style
- length, order
- usage, style
- usage, order
- style, order
- length
- usage
- style
- order

EXCEPT: we should never have to apply coverage to the production data.

macchiati: So I think just lines 911 and 912 are good enough.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -285,26 +285,15 @@
...
-//ldml/personNames/personName[@style=\"%A\"][@order=\"%A\"]/namePattern[@_q=\"%A\"]                 ; Misc ; Person Name Formats ; Name Patterns for Length any ; usage-any-style-$1-order-$2
-//ldml/personNames/personName[@style=\"%A\"]/namePattern[@_q=\"%A\"]                        ; Misc ; Person Name Formats ; Name Patterns for Length any ; usage-any-style-$1-order-any
-//ldml/personNames/personName[@order=\"%A\"]/namePattern[@_q=\"%A\"]                        ; Misc ; Person Name Formats ; Name Patterns for Length any ; usage-any-style-any-order-$1
-
+//ldml/personNames/nameOrder[@nameLocales=\"%A\"]   ; Misc ; Person Name Formats ; Name Order For Languages ; $1
+//ldml/personNames/personName/namePattern[@alt=\"%A\"]      ; Misc ; Person Name Formats ; Fallback Name Pattern ; any-$1
+//ldml/personNames/personName/namePattern           ; Misc ; Person Name Formats ; Fallback Name Pattern ; any
+//ldml/personNames/personName[@length=\"%A\"][@usage=\"%A\"][@style=\"%A\"][@order=\"%A\"]/namePattern[@alt=\"%A\"]     ; Misc ; Person Name Formats ; Name Patterns for Length $1 ; $1-$2-$3-$4-$5
```
macchiati, line 291: For the ST, the same as above; the full attributes are enough.
However, for the charts, I think we need all of them.
Double however, because the 'code' will always be just all of the attributes concatenated, I think we can use something like:

... [@%att=\"%A\"][@%att=\"%A\"][@%att=\"%A\"][@%att=\"%A\"]/namePattern[@alt=\"%A\"]
... [@%att=\"%A\"][@%att=\"%A\"][@%att=\"%A\"][@%att=\"%A\"]/namePattern
... [@%att=\"%A\"][@%att=\"%A\"][@%att=\"%A\"]/namePattern[@alt=\"%A\"]
...

common/supplemental/coverageLevels.xml
pedberg-icu: See my comments in the PR overview and the e-mail I sent. I think there is no need to have fully qualified in the source for attributes that are irrelevant, like usage for monograms, or order for usage=sorting. What would be the point? The regular cldr data should b fully fleshed out *except* for those cases (i.e. for monograms we don't need any entries with usage attribute). That is the way I entered the data, so that is the way I have coverage setup (and PathHeader, etc.).

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
pedberg-icu: See my comment about coverage. I think that the regular data should *not* have entries for all values of ignored attributes, it should should just have entries without that attribute (i.e. for monograms no entries with usage attribute). So I have set up PathHeader, coverage etc. with that assumption. That is different than what we discussed in the meeting; I mentioned that in the PR overview and also in an e-mail.

common/supplemental/coverageLevels.xml
macchiati: Ok, makes sense.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
macchiati: Ok. I still think we will have to do something for the charts, because they will need to work with production data. For them I think we will need to have a (simple) function &hyphen(...) for the output, that replaces spaces with hyphens. That way for the compacted data
length="long short" => long-short

However, we don't need to do that here.

macchiati (approved):

richgillam (commented):
A lot of this I don't fully understand, but I think it all looks okay.  I notice you moved to listing all the combinations of attribute values (or most of them) exhaustively.  I'm still a bit foggy on the rationale for that change.  Would you mind cluing me in (again)?

common/main/en.xml
```diff
@@ -9129,50 +9129,122 @@ annotations.
...
+			<namePattern>{given-informal}</namePattern>
 		</personName>
-		<personName usage="addressing" style="informal">
+		<personName length="long" usage="addressing" style="informal" order="givenFirst">
 			<namePattern>{given-informal}</namePattern>
 		</personName>
-		<personName usage="addressing">
-			<namePattern>{prefix} {surname}</namePattern>
+		<personName length="long" usage="referring" style="formal" order="surnameFirst">
+			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
```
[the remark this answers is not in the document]
pedberg-icu: We can discuss in the meeting this week. I will be doing another pass over the data and dtd for some other changes (like the initial patterns).

pedberg-icu:
> I notice you moved to listing all the combinations of attribute values (or most of them) exhaustively. I'm still a bit foggy on the rationale for that change. Would you mind cluing me in (again)?

This is mainly for the Survey Tool. There is not really a good way in Survey Tool for vetters in different locales to say that they just want to add entries for certain combinations of attribute values, with the particualr combinations being different from locale to locale. Instead, for the regular CLDR data, we will have separate entries for all of the supported combinations, and ask vetters to enter values for each (possibly repeating the same values for several different entries). Then when we generate the "production" version of the CLDR data, we will use tooling to minimize the actual combinations we have in the data for each locale.

PR #1812, opened 2022-03-20
CLDR-15305 fix discussion of plural rules

macchiati:
CLDR-15305

This is intended to match Robin's recommendations in 
* https://unicode-org.atlassian.net/browse/CLDR-15305?focusedCommentId=164007
* https://unicode-org.atlassian.net/browse/CLDR-15305?focusedCommentId=164028
with a few wording differences.


macchiati:
Wrong target

PR #1813, opened 2022-03-20
CLDR-15305 fix discussion of plural rules

macchiati:
CLDR-15305

This is intended to match Robin's recommendations in

https://unicode-org.atlassian.net/browse/CLDR-15305?focusedCommentId=164007
https://unicode-org.atlassian.net/browse/CLDR-15305?focusedCommentId=164028
with a few wording differences.

It also includes one clarifying example from the ticket discussion:
  * For example, e = 0 and i != 0 and i % 1000000 = 0 and *+v = 0+* or e != 0..5 is parsed as if it were (e = 0 and i != 0 and i % 1000000 = 0 and v = 0) or (e != 0..5)


eggrobin (approved):
I like your use of 𝑁 so that the source number has a name (leaving lowercase letters exclusively for operands).

sffc (approved):

sffc:
CC @echeran

PR #1818, opened 2022-03-21
CLDR-14111 First draft of number range formatting spec

sffc:
CLDR-14111

### DIFF
--- a/docs/ldml/tr35-numbers.md
+++ b/docs/ldml/tr35-numbers.md
@@ -79,6 +79,10 @@ The LDML specification is divided into the following parts:
     *   5.2 [Plural Ranges](#Plural_Ranges)
 *   6 [Rule-Based Number Formatting](#Rule-Based_Number_Formatting)
 *   7 [Parsing Numbers](#Parsing_Numbers)
+*   8 [Number Range Formatting](#Number_Range_Formatting)
+    *   8.1 [Approximate Number Formatting](#Approximate_Number_Formatting)
+    *   8.2 [Collapsing Number Ranges](#Collapsing_Number_Ranges)
+    *   8.3 [Range Pattern Processing](#Range_Pattern_Processing)
 
 ## 1 <a name="Numbering_Systems" href="#Numbering_Systems">Numbering Systems</a>
 
@@ -441,7 +445,7 @@ The miscPatterns supply additional patterns for special purposes. The currently 
 
 **approximately**
 
-> indicates an approximate number, such as: “~99”
+> indicates an approximate number, such as: “\~99”. This pattern is not currently in use; see ICU-20163.
 
 **atMost**
 
@@ -1241,6 +1245,8 @@ The `pluralRanges` element provides information allowing an implementation to de
 
 The data has been gathered presuming that in any usage, the start value is strictly less than the end value, and that no values are negative. Results for any cases that do not meet these criteria are undefined.
 
+For the formatting of number ranges, see <a name="Number_Range_Formatting" href="#Number_Range_Formatting">Number Range Formatting</a>.
+
 ## 6 <a name="Rule-Based_Number_Formatting" href="#Rule-Based_Number_Formatting">Rule-Based Number Formatting</a>
 
 ```xml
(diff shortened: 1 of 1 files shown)


sffc:
Let me know if this is what you had in mind. There is still one TODO item I need to resolve.

docs/ldml/tr35-numbers.md
```diff
@@ -1315,6 +1317,54 @@ Here is a set of heuristic rules that may be helpful:
...
+
+1. Format the lower bound and the upper bound independently following the steps in <a name="Number_Format_Patterns" href="#Number_Format_Patterns">Number Format Patterns</a>.
+1. If the resulting values are identical, stop evaluating these steps and, instead, perform the steps in <a name="Approximate Number Formatting" href="#Approximate Number Formatting">Approximate Number Formatting</a>.
+    1. Note: This behavior may be customized in order to, for example, print the range despite the endpoints being identical. However, a spec-compliant implementation must support approximate number formatting.
+1. Perform the steps in <a name="Collapsing_Number_Ranges" href="#Collapsing_Number_Ranges">Collapsing Number Ranges</a>, obtaining modified *lower* and *upper* values.
+1. Obtain a number range pattern by following the steps in <a name="Range_Pattern_Processing" href="#Range_Pattern_Processing">Range Pattern Processing</a>.
+1. Substitute *lower* as `{0}` and *upper* as `{1}` into the range pattern from the previous step.
+
+For plural rule selection of number ranges, see <a name="Plural_Ranges" href="#Plural_Ranges">Plural Ranges</a>.
```
macchiati, line 1333: This construct
<a name="Plural_Ranges" href="#Plural_Ranges">Plural Ranges</a>
is only appropriate if you want an href to Plural_Ranges to go here.
Should instead be a regular Markdown link. 

Same problem in other cases, so search for <a name=...

macchiati (changes requested):

docs/ldml/tr35-numbers.md
```diff
@@ -1315,6 +1317,54 @@ Here is a set of heuristic rules that may be helpful:
...
+
+### 8.1 <a name="Approximate_Number_Formatting" href="#Approximate_Number_Formatting">Approximate Number Formatting</a>
+
+TODO
+
+### 8.2 <a name="Collapsing_Number_Ranges" href="#Collapsing_Number_Ranges">Collapsing Number Ranges</a>
+
+*Collapsing* a number range refers to the process of removing duplicated information in the *lower* and *upper* values. For example, if the lower string is "3.2 centimeters" and the upper string is "4.5 centimeters", it is desirable to remove the extra "centimeters" token.
```
macchiati, line 1341: This is not going to be easily implementable, since programs would have to check for the ill-defined concept of semantic equivalents. Hard to do across languages.

To make this work in practice, I think it needs to go a bit lower in the process. One approach (you can probably think of a better): when the number is being formatted, we need to get the "measure" in the broad sense, including currencies, and any compact decimal prefix/suffix, and their locations in the resulting string (relative to themselves and the number. EG, 

<3.5, number> <M, compact-designation> <centimeters, id:centimeter>
vs
<M, compact-designation> <3.5, number> <centimeters, id:centimeter>
vs
...
Then we can describe what to do.

docs/ldml/tr35-numbers.md
sffc: Fixed.

docs/ldml/tr35-numbers.md
sffc: Fixed.

docs/ldml/tr35-numbers.md
```diff
@@ -1315,6 +1321,73 @@ Here is a set of heuristic rules that may be helpful:
...
+
+1. "centimeter" is semantically equivalent to "centimeters".
+1. "K" (the thousands symbol in compact decimals) is NOT semantically equivalent to "K" (the measurement unit Kelvin).
+
+Two semantically equivalent tokens can be *collapsed* if they appear at the start of both values or the end of both values. However, the implementation may choose different levels of aggressiveness with regard to collapsing tokens.
+
+**Heuristic:** Only collapse semantically equivalent *unit of measurement* tokens, and only if the tokens are more than one code point in length. This is to avoid ambiguous strings such as "3–5K" (could represent 3–5000 or 3000–5000).
+
+**To collapse tokens:** Remove the token from both values, and then re-compute the token based on the number range. If the token depends on the plural form, follow [Plural Ranges](#Plural_Ranges) to calculate the correct form. If the tokens originated at the beginning of the string, prepend the new token to the beginning of the *lower* string; otherwise, append the new token to the end of the *upper* string.
+
```
macchiati, line 1377: Add:
The above description describes the expected output. Internally, the implementation may determine the equivalent units of measurement by passing the codes back from the number formatters, allowing for a precise determination of "semantically equivalent".

macchiati (changes requested):
Github was not letting me add comments inline:

Change to 

* Only collapse semantically equivalent *unit of measurement* tokens. This is to avoid ambiguous strings such as "3–5K" (could represent 3–5000 or 3000–5000).
* Only collapse if the tokens are more than one code point in length.
* These heuristics may be refined in the future.

Add:
The above description describes the expected output. Internally, the implementation may determine the equivalent units of measurement by passing the codes back from the number formatters, allowing for a precise determination of "semantically equivalent".


Add: 

The current recommended heuristic is: 
 * If the *lower* string ends with a character other than a digit, or if the *upper* string begins with a character other than a digit, and if the range pattern does not contain a character having the `White_Space` binary Unicode property after the `{0}` or before the `{1}` placeholders, insert a non-breaking space (U+00A0) at those positions.

docs/ldml/tr35-numbers.md
```diff
@@ -1315,6 +1321,73 @@ Here is a set of heuristic rules that may be helpful:
...
+- `US dollars` → unit of measurement for the currency USD
+
+Two tokens are *semantically equivalent* if they have the same *semantic annotations*, even if they are not the exact same string. For example:
+
+1. "centimeter" is semantically equivalent to "centimeters".
+1. "K" (the thousands symbol in compact decimals) is NOT semantically equivalent to "K" (the measurement unit Kelvin).
+
+Two semantically equivalent tokens can be *collapsed* if they appear at the start of both values or the end of both values. However, the implementation may choose different levels of aggressiveness with regard to collapsing tokens.
+
+**Heuristic:** Only collapse semantically equivalent *unit of measurement* tokens, and only if the tokens are more than one code point in length. This is to avoid ambiguous strings such as "3–5K" (could represent 3–5000 or 3000–5000).
```
macchiati, line 1374: Change to

The current recommended heuristic is:

* Only collapse semantically equivalent unit of measurement tokens. This is to avoid ambiguous strings such as "3–5K" (could represent 3–5000 or 3000–5000).
* Only collapse if the tokens are more than one code point in length. [[Shane, what is the reason?]]

These heuristics may be refined in the future.

docs/ldml/tr35-numbers.md
```diff
@@ -1315,6 +1321,73 @@ Here is a set of heuristic rules that may be helpful:
...
+**To collapse tokens:** Remove the token from both values, and then re-compute the token based on the number range. If the token depends on the plural form, follow [Plural Ranges](#Plural_Ranges) to calculate the correct form. If the tokens originated at the beginning of the string, prepend the new token to the beginning of the *lower* string; otherwise, append the new token to the end of the *upper* string.
+
+### 8.3 <a name="Range_Pattern_Processing" href="#Range_Pattern_Processing">Range Pattern Processing</a>
+
+To obtain a number range pattern, the following steps are taken:
+
+1. Load the range pattern found in [Miscellaneous Patterns](#Miscellaneous_Patterns).
+1. Optionally add spacing to the range pattern.
+
+**Heuristic:** If the *lower* string ends with a character other than a digit, or if the *upper* string begins with a character other than a digit, add spacing to the range pattern.
```
macchiati, line 1385: Change to:

The current recommended heuristic is:

 * If the lower string ends with a character other than a digit or if the upper string begins with a character other than a digit
 * **and** if the range pattern does not contain a character having the White_Space binary Unicode property after the {0} or before the {1} placeholders
 * then insert a non-breaking space (U+00A0) at those positions.

These heuristics may be refined in the future.

[[Note change of 0020 to 00A0]]

macchiati (changes requested):

macchiati:
Was able to add inline comments now.

docs/ldml/tr35-numbers.md
sffc: Added, but a few lines above where you suggested

docs/ldml/tr35-numbers.md
sffc: Done

docs/ldml/tr35-numbers.md
sffc: done

pedberg-icu:
@sffc This needs to be retargeted to the maint/maint-41 branch (and then probably rebased)

macchiati (approved):

macchiati:
Shane, we'd like to get this merged ASAP. You'll have to squash to fix a commit message (I would have done that but https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/cldr/1818 won't work for me on your branch :-)

sffc:
My git operations were:

1. Add one more commit https://github.com/unicode-org/cldr/commit/385d6838181d9d13536e015896395a46bb62fc4d fixing a small typo
2. Force-push to rebase the branch, carrying the 9 commits from the old branch into the new branch
3. Squash using the online tool

sffc:
As soon as CI is green, this PR is mergeable. Anyone can feel free to push the button if I don't get to it first.

macchiati:
Thanks!

PR #1821, opened 2022-03-22
CLDR-15408 Remove most of Annex A Deprecated Structure, refer to DTD …

pedberg-icu:
…Delta charts

CLDR-15408

Remove most of spec Part 1 Annex A Deprecated Structure, instead referring to the [DTD Deltas](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/dtd_deltas.html) chart. But keep the following:
- Anchors in the Deprecated Structure section that could be the target of links from other places.
- Information about replacements for deprecated items (and a few other useful tidbits that are not about explaining the deprecated structure).

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -1180,7 +1180,7 @@ LDML 1.7 or older specification used different syntax for representing unicode l
 
 |                               | EBNF |
 | ----------------------------- | ---- |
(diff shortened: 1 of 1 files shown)


macchiati (approved):

pedberg-icu:
@macchiati, a third stuck PR that needs admin power to merge...thanks

PR #1825, opened 2022-03-22
CLDR-15456 Currency: SLL»SLE

srl295:
CLDR-15456

- [X] This PR completes the ticket. (but will need a followon to track in v42)

- There’s an undetermined transition period. SLE is an internal code until that transition period is announced.
- For now SLE is 'new leone' but it will eventually just be 'leone' when SLE is in circulation.
- Updated ISO4217 and currency validity. SLE is not available in the 4217 lists yet.
- Also update tests that would otherwise fail on NPEs found during development

https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/amendments/dl_currency_iso_amendment_171.pdf

### DIFF
--- a/common/bcp47/currency.xml
+++ b/common/bcp47/currency.xml
@@ -238,6 +238,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="shp" description="Saint Helena Pound"/>
             <type name="sit" description="Slovenian Tolar" since="1.9"/>
             <type name="skk" description="Slovak Koruna"/>
+            <type name="sle" description="Sierra Leonean New Leone"/>
             <type name="sll" description="Sierra Leonean Leone"/>
             <type name="sos" description="Somali Shilling"/>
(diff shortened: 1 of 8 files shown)


pedberg-icu (approved):

macchiati (approved):

roozbehp:
I have a major problem with this. The old currency (SLL) is changed in the PR to have two decimal digits instead of the previous 0. Sure, this is what ISO formally says, but it doesn't match the facts on the ground, which is that the currency is so small (117 SLL is 0.01 USD) that subunits are not used. A new PR should be done that restores the 0 for SLL.

srl295:
That’s fair. Are you ok with leaving SLE at 2?

roozbehp:
> That’s fair. Are you ok with leaving SLE at 2?

Yes, SLE should remain at two. There's no practice on the ground for it yet, so we should go with ISO there.

srl295:
If anyone lands here, the followon is CLDR-15900

PR #1828, opened 2022-03-22
CLDR-15469 Improve name pattern parser

wasker:
CLDR-15469

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -5,6 +5,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
@@ -16,6 +17,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.function.Function;
 
+import org.checkerframework.checker.units.qual.C;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.Counter;
 import org.unicode.cldr.util.XPathParts;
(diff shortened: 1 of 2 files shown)


wasker:
@macchiati FYI - started on parser. Incomplete, I'll need to add successful tests.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -327,28 +327,63 @@ public static NamePattern from(String patternString) {
...
+            int i = 0;
+            while (i < patternLength) {
+                final Character currentCharacter = patternString.charAt(i);
+
+                switch (currentCharacter) {
+                case '\\':
+                    if (i + 1 < patternLength) {
+                        rawValue += patternString.charAt(i + 1);
+                        i += 2;
+                        continue;
```
richgillam, line 343: I wonder if we want to restrict this to just escape sequences that we actually define.  I guess right now that'd just be `\{`, `\\`, and (I guess) `\}`.  Then again, maybe there's no good reason to be so restrictive and this approach future-proofs us if we add more syntax characters,,,

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPersonNameFormatter.java
```diff
@@ -158,5 +167,30 @@ public void TestLabels() {
...
+            code.run();
+            assertTrue("Exception was not thrown as expected.", true);
+        }
+        catch (Exception e) {
+            fail(String.format("%s was not supposed to throw an exception.", subject));
+        }
+    }
+
+    @FunctionalInterface
+    private interface Runable {
```
richgillam, line 191: The correct spelling is `Runnable`, but I'm pretty sure Java already defines a "Runnable" interface.  Although it looks a little bit like the built-in Java `Runnable` interface might be what you want here...

richgillam (approved):
Looks good.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
wasker: I actually do want to restrict escape sequences here. This will have an effect on toString, which is used for testing positive cases.

macchiati (dismissed):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPersonNameFormatter.java
wasker: Oh, thanks, that's exactly what I wanted!

macchiati:
Note: you need to have "CLDR-15469 " at the start of each commit and in the PR pattern. Otherwise we have to squash and fix before merging.

wasker:
> Note: you need to have "[CLDR-15469](https://unicode-org.atlassian.net/browse/CLDR-15469) " at the start of each commit and in the PR pattern. Otherwise we have to squash and fix before merging.

Yep, was going through the motions. :)

Switch this issue's state to Accepted, please: https://unicode-org.atlassian.net/browse/CLDR-15469

macchiati:
done

PR #1835, opened 2022-03-24
CLDR-15105 spec: update ToC

srl295:
- use tool https://gfmtoc.relevant.space

CLDR-15105

Sample run: https://corp.unicode.org/~srloomis/tmp/dist13781/dist/


srl295:
#1836 will cause this to need a redo… PTAL but will re-run and resubmit

docs/ldml/tr35-info.md
```diff
@@ -38,35 +38,35 @@ The LDML specification is divided into the following parts:
...
-*   12 [Parent Locales](#Parent_Locales)
-*   13 [Unit Conversion](#Unit_Conversion)
-*   14 [Unit Preferences](#Unit_Preferences)
+* 1 Introduction [Supplemental Data](#Supplemental_Data)
+* 2 [Territory Data](#Territory_Data)
+  * 2.1 [Supplemental Territory Containment](#Supplemental_Territory_Containment)
+  * 2.2 [Subdivision Containment](#Subdivision_Containment)
+  * 2.3 [Supplemental Territory Information](#Supplemental_Territory_Information)
+  * 2.4 [Territory-Based Preferences](#Territory_Based_Preferences)
+    * 2.4.1 [Preferred Units for Specific Usages](#Preferred_Units_For_Usage)
```
pedberg-icu, line 47: The entry for the Table after 2.4.1 is missing here

docs/ldml/tr35-keyboards.md
```diff
@@ -44,46 +44,47 @@ The LDML specification is divided into the following parts:
...
-*   10 [Platform Behaviors in Edge Cases](#Platform_Behaviors_in_Edge_Cases)
+* 1 [Keyboards](#Introduction)
+* 2 [Goals and Non-goals](#Goals_and_Nongoals)
+* 3 [Definitions](#Definitions)
+  * 3.1 [Escaping](#Escaping)
+* 4 [File and Directory Structure](#File_and_Dir_Structure)
+* 5 [Element Hierarchy - Layout File](#Element_Heirarchy_Layout_File)
+  * 5.1 [Element: keyboard](#Element_Keyboard)
+  * 5.2 [Element: version](#Element_version)
+  * 5.3 ~~[Element: generation](#Element_generation)~~
```
pedberg-icu, line 55: What is the ~~ in this entry?

docs/ldml/tr35-info.md
srl295: It's a table caption. I'll see if I can use an option to bring it back.

docs/ldml/tr35-keyboards.md
srl295: It's strikeout, and is actually in the original element because it is deprecated.  It works in the ToC and so might be a positive change.

miloush: it's deprecated

pedberg-icu:
IN general the main difference I see is that this removes the "Table:" prefix for TOC entries for tables. However in at least one case (in tr35-info.md after the entry for 2.4.1) the table entry has gone missing from he TOC entirely; and there is an odd ~~ preceding one TOC entry in tr35-keyboards.md.

docs/ldml/tr35-dates.md
```diff
@@ -38,52 +38,53 @@ The LDML specification is divided into the following parts:
...
-*   3 [Calendar Fields](#Calendar_Fields)
-*   4 [Supplemental Calendar Data](#Supplemental_Calendar_Data)
-    *   4.1 [Calendar Data](#Calendar_Data)
-    *   4.2 [Calendar Preference Data](#Calendar_Preference_Data)
-    *   4.3 [Week Data](#Week_Data)
-        *   Table: [Week Designation Types](#Week_Designation_Types)
-    *   4.4 [Time Data](#Time_Data)
-    *   4.5 [Day Period Rule Sets](#Day_Period_Rule_Sets)
-        *   4.5.1 [Day Period Rules](#Day_Period_Rules)
-            *   4.5.1.1 [Fixed periods](#Fixed_periods)
```
macchiati, line 64: The formatting is inconsistent. In many cases it includes the number in the following anchor text, and in many cases it does not.

**[Turn on Hide Whitespace changes to do the review and this becomes far more obvious.]**

macchiati (dismissed):

docs/ldml/tr35-dates.md
macchiati: Note: other than that the formatting looks pretty good. It would be nice to have the Table: back if that can be done.

srl295: this probably points to inconsistencies in the source text, since this is an automated generation.

Table: will come back soon, but probably not with Table

srl295:
PTAL .I think i have resolved everthing.  And, added docs for running the automated tool.  This can definitely be a part of the BRS to regenerate ToC if needed.

docs/ldml/tr35-dates.md
macchiati: Ok, that makes sense. We should fix the inconsistency in another ticket. (Having them all adjacent in the TOC makes them obvious!)

docs/ldml/tr35-info.md
macchiati: Still missing?

docs/ldml/tr35-dates.md
macchiati: I suppose that we could do the same with "Table: ", that is, fix on the source line.

I presume in the production version the table captions will be in a larger font size.

srl295:
OK i made another update to document the ToC fixer code a bit, and to have it run automatically when the archive is created. This means that for the site, the ToC will be kept up to date automatically in HTML.

I'll stop working on it now to get some more review comments.

docs/ldml/tr35-info.md
srl295: There's actually no table there! Maybe it used to be there.  It reads like this:

```md
#### 2.4.1 <a name="Preferred_Units_For_Usage" href="#Preferred_Units_For_Usage">Preferred Units for Specific Usages</a>

_For information about preferred units and unit conversion, see Section 13 [Unit Conversion](#Unit_Conversion) and Section 14 [Unit Preferences](#Unit_Preferences)._

### 2.5 <a name="rgScope" href="#rgScope">`<rgScope>`: Scope of the “rg” Locale Key</a>
```

So no table between 2.4.1 and 2.5 … maybe it was moved… 

if we are otherwise happy with this, and this PR is uncovering errors/inconsistencies, i propose merging this and then fixing the actual text afterwards.

docs/ldml/tr35-dates.md
srl295: if we want to make that part of the title, we could. per the other ticket, the 'captions' are now just H6 (`######`).

srl295:
BTW https://unicode-org.atlassian.net/browse/CLDR-15206 covers any stylistic improvements… so can make changes there.

docs/ldml/tr35-info.md
pedberg-icu: Agree...

pedberg-icu (approved):

srl295:
@macchiati ok to merge ?

macchiati:
Yup

PR #1838, opened 2022-03-24
CLDR-15404 Add example generation 2

macchiati:
CLDR-15404

The main part of this PR is to add the Example Generator. However, in doing so I found that the tests were not being run, so I fixed those.

Notes

* There is a to-do to use the sample person names once those are in en.xml.

* Had to change from using _q to using alt. Simplest was to move the ordering rank into the NamePattern.
* Changed the precedence order to match the DTD. usage before style
* Added some checks for bad data

* Removed incorrect .java

* Added test for the example generator.


macchiati:
I worked out the kinks in this. Can one of you do the review?

pedberg-icu (approved):

PR #1839, opened 2022-03-25
CLDR-15404 add fallbacks

macchiati:
CLDR-15404

- It adds fallback behavior for modifiers, including the first grapheme for -initial, and uppercasing for -allCaps. That also involved newly passing around the locale and a BreakIterator
- There is a default order if none specified by the name object, to prevent inadvertent matches.
- There is a working bestValue for field values implementation in the SimpleNameObject to figure out the best match for field+modifier combinations.
  - It uses longest match first
  - Among the longest matches (if more than one), favors modifiers according to the following order (see the Modifier enum)
    - informal, allCaps, initial, prefix, core
    - We need to decide on the right order and add to the spec.
  - This meant changing the internal data structure so that we could go right from the field to the modifier combinations (better performance anyway). I also impose an invariants:
    - If you have a modified field, you have to have the unmodified field. That is, the name object wouldn't have 'given-informal' if it didn't have 'given'.
    - We need to talk about other possible invariants. Eg, given is required; surname2 only occurs if there is a surname.
  - A few modifications to tests. Eg no more "Smith, null. B." values

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -33,9 +33,11 @@ import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
-import com.google.common.collect.Sets.SetView;
 import com.google.common.collect.TreeMultimap;
 import com.ibm.icu.impl.Row.R2;
+import com.ibm.icu.lang.UCharacter;
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.text.MessageFormat;
 import com.ibm.icu.util.ULocale;
 
 /**
@@ -112,11 +114,24 @@ public class PersonNameFormatter {
 
 
     public enum Modifier {
-        initial,
-        allCaps,
         informal,
-        core;
+        allCaps,
+        initial,
+        prefix,
+        core
+        ;
         public static final Comparator<Iterable<Modifier>> ITERABLE_COMPARE = Comparators.lexicographical(Comparator.<Modifier>naturalOrder());
+        public static final Comparator<Collection<Modifier>> LONGEST_FIRST = new Comparator<Collection<Modifier>>() {
+
+            @Override
+            public int compare(Collection<Modifier> o1, Collection<Modifier> o2) {
+                return ComparisonChain.start()
+                    .compare(o2.size(), o1.size()) // reversed order for longest first
+                    .compare(o1,  o2, ITERABLE_COMPARE)
+                    .result();
+            }
+
+        };
         public static final Set<Modifier> ALL = ImmutableSet.copyOf(Modifier.values());
     }
 
@@ -267,6 +282,36 @@ public class PersonNameFormatter {
     }
 
     /**
+     * Format fallback results, for when modifiers are not found
+     */
+
+    public static class FallbackFormatter {
+        private ULocale formatterLocale;
+        private BreakIterator characterBreakIterator;
+        private MessageFormat initialFormatter;
+
+        public FallbackFormatter(ULocale uLocale, String formatterPattern) {
+            formatterLocale = uLocale;
+            characterBreakIterator = BreakIterator.getCharacterInstance(uLocale);
+            initialFormatter = new MessageFormat(formatterPattern);
+        }
+
+        public String formatInitial(String bestValue) {
+            // TODO Alex if multiword, apply the algorithm to each part
+            BreakIterator breakIterator = characterBreakIterator;
+            breakIterator.setText(bestValue);
+            bestValue = bestValue.substring(0, breakIterator.next());
+            bestValue = initialFormatter.format(new String[] {bestValue});
+            // TODO Alex use initial pattern after defined
+            return bestValue;
+        }
+
+        public String formatAllCaps(String bestValue) {
+            return UCharacter.toUpperCase(formatterLocale, bestValue);
+        }
+    }
+
+    /**
      * A name pattern, corresponding to a string such as "{given-initial} {surname}"
      * Immutable
      */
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -292,10 +315,38 @@ public String format(NameObject nameObject) {
...
                     result.append(literal);
                 } else {
-                    final String bestValue = nameObject.getBestValue(element.getModifiedField(), remainingModifers);
-                    if (!remainingModifers.isEmpty()) {
+                    String bestValue = nameObject.getBestValue(element.getModifiedField(), remainingModifers);
+                    if (bestValue == null) {
+                        // TODO Rich to handle collapsing whitespace / literals
+                        continue;
+                    } if (!remainingModifers.isEmpty()) {
                         // TODO Alex Apply unhandled modifiers algorithmically where possible
```
richgillam, line 323: I'm confused-- isn't this what you're doing below?

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -292,10 +315,38 @@ public String format(NameObject nameObject) {
...
+                            case allCaps:
+                                bestValue = UCharacter.toUpperCase(formatterLocale, bestValue);
+                                break;
+                            case prefix:
+                                // TODO unhandled prefix is special; treat as if null
+                                // TODO if there is no plain, but there is a prefix and core, use that; otherwise use core
+                                break;
+                            case core:
+                            case informal:
+                                // no option, just fall back
```
richgillam, line 345: "Fall back" here means just return the unmodified version of the field (or at least skip this particular modifier), right?

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -283,7 +298,15 @@ public int getFieldsSize() {
...
-        public String format(NameObject nameObject) {
+        /**
+         * Return the rank order (0, 1, ...) in a list
+         * @return
+         */
+        public int getRank() {
+            return rank;
+        }
+
+        public String format(NameObject nameObject, ULocale formatterLocale, BreakIterator characterBreakIterator) {
```
richgillam, line 309: It seems unfortunate to require the break iterator to be passed in.  Couldn't we just instantiate it here if we need it?  For that matter, it also seems unfortunate that we need the locale.  It seems like we should be able to get that from the rules somehow, or at least store it in a property on the formatter instead of passing it to the format() method.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -292,10 +315,38 @@ public String format(NameObject nameObject) {
                 if (literal != null) {
                     result.append(literal);
                 } else {
-                    final String bestValue = nameObject.getBestValue(element.getModifiedField(), remainingModifers);
-                    if (!remainingModifers.isEmpty()) {
+                    String bestValue = nameObject.getBestValue(element.getModifiedField(), remainingModifers);
+                    if (bestValue == null) {
+                        // TODO Rich to handle collapsing whitespace / literals
```
richgillam, line 320: I'll try to do this on Monday; I ran out of time today.

richgillam (commented):
I ran out of time before I got through the whole PR; I stopped before I got to SimpleNameObject.  I'm sorry; I'll pick it up there Monday morning.

The part I _did_ review looks good!

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
macchiati: For the breakIterator, I didn't want to pay the creation cost on every format call, and it seemed overkill to store it in every NamePattern.

Note: this does not have to be heavily optimized code, since the use will be in the CLDR tooling. However, I wanted to take a reasonable stab at what could make sense in a production environment.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
macchiati: Thanks!

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
macchiati: Ok. I should delete that, since it is no longer applicable.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
macchiati: It means for most fields, skip the modifier. For the prefiix modifier, however, we actually want return an empty value, so that {surname-prefix} gets ignored if there is no explicit field.

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/SimpleNameObject.java
```diff
@@ -1,85 +1,140 @@
...
+    public ULocale getNameLocale() {
+        return nameLocale;
+    }
 
-        public SimpleNameObject(ULocale nameLocale, ImmutableMap<ModifiedField, String> patternData) {
-            this.nameLocale = nameLocale == null ? ULocale.ROOT : nameLocale;
-            this.patternData = patternData;
+    public SimpleNameObject(ULocale nameLocale, Map<ModifiedField, String> patternData) {
+        this.nameLocale = nameLocale == null ? ULocale.ROOT : nameLocale;
+        Map<Field, Map<Set<Modifier>, String>> _patternData = new EnumMap<>(Field.class);
```
macchiati, line 70: BTW, I liked the way this code worked out; by changing the data structure I was able to minimize the number of lookups. One to get the field data, then walk through the cases where the NameObject also had explicit modifier strings. (For our current data, the bulk have just 1 item.)

macchiati:
Peter, can you review this one?

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -844,7 +923,8 @@ public ParameterMatcher slim() {
 
         public NamePattern getBestMatch(NameObject nameObject, FormatParameters nameFormatParameters) {
             if (nameFormatParameters.order == null) {
-                nameFormatParameters = nameFormatParameters.setOrder(localeToOrder.get(nameObject.getNameLocale()));
+                final Order mappedOrder = localeToOrder.get(nameObject.getNameLocale());
+                nameFormatParameters = nameFormatParameters.setOrder(mappedOrder == null ? Order.givenFirst : mappedOrder);
```
pedberg-icu, line 927: Hmm, I wonder whether default to surnameFirst is better. We can discuss in meeting.

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
macchiati: Will add to agenda

macchiati: added.

richgillam (commented):
Belated ship-it.  The SimpleNameObject class looks good (actually I was having trouble grokking some of it, but I blame myself and a general lack of sleep for that).  The new helper class addresses my other concern and looks good.

PR #1841, opened 2022-03-27
CLDR-15484 Fix mistakes

ivan-pan:
CLDR-15484

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -123,11 +123,11 @@ The LDML specification is divided into the following parts:
 <!ELEMENT localeDisplayNames ( alias | ( localeDisplayPattern?, languages?, scripts?, territories?, subdivisions?, variants?, keys?, types?, transformNames?, measurementSystemNames?, codePatterns?, special* ) )>
 ```
 
-Display names for scripts, languages, countries, currencies, and variants in this locale are supplied by this element. They supply localized names for these items for use in user-interfaces for various purposes such as displaying menu lists, displaying a language name in a dialog, and so on. Capitalization should follow the conventions used in the middle of running text; the `<contextTransforms>` element may be used to specify the appropriate capitalization for other contexts (see _Section 12 [ContextTransform Elements](#Context_Transform_Elements)_ ). Examples are given below.
+Display names for scripts, languages, countries, currencies, and variants in this locale are supplied by this element. They supply localized names for these items for use in user-interfaces for various purposes such as displaying menu lists, displaying a language name in a dialog, and so on. Capitalization should follow the conventions used in the middle of running text; the `<contextTransforms>` element may be used to specify the appropriate capitalization for other contexts (see _Section 12 [ContextTransform Elements](#Context_Transform_Elements)_). Examples are given below.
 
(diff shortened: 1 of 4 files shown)


docs/ldml/tr35-general.md
```diff
@@ -1303,7 +1303,7 @@ The durationUnit is a special type of unit used for composed time unit durations
 </durationUnit>
 ```
 
-The type contains a skeleton, where 'h' stands for hours, 'm' for minutes, and 's' for sections. These are the same symbols used in availableFormats, except that there is no need to distinguish different forms of the hour.
+The type contains a skeleton, where 'h' stands for hours, 'm' for minutes, and 's' for seconds. These are the same symbols used in availableFormats, except that there is no need to distinguish different forms of the hour.
```
srl295, line 1306: thanks for catching these…

srl295 (dismissed):
Great finds.

srl295:
@ivan-pan can you rebase?

ivan-pan:
I noticed Гриинуич in Part 3. This should be changed to [Гринуич](https://bg.wikipedia.org/wiki/%D0%93%D1%80%D0%B8%D0%BD%D1%83%D0%B8%D1%87) in the locale data and, when done, in the LDML specification.

ivan-pan:
On another note, the HTML version of [Part 5](https://unicode-org.github.io/cldr/ldml/tr35-collation.html) contains `:point_right:` instead of `👉`.

ivan-pan:
In Part 6, something is wrong here:

> \* The type attribute contained only a `unicode_region_subtag` `unicode_subdivision_suffix` values were used in the `contains` attribute; these are not unique across multiple territories, so for lower levels a now-deprecated

> Therer mayreflecting that _force per area_ can be referenced as either _pressure_ or _stress_, for example).

srl295 (approved):

srl295:
> In Part 6, something is wrong here:
> 
> > * The type attribute contained only a `unicode_region_subtag` `unicode_subdivision_suffix` values were used in the `contains` attribute; these are not unique across multiple territories, so for lower levels a now-deprecated
> 
> > Therer mayreflecting that _force per area_ can be referenced as either _pressure_ or _stress_, for example).

First, thanks once again! Second, I'll take a look at this.

ivan-pan:
@srl295 Please also take a look at these two comments of mine:

> I noticed Гриинуич in Part 4. This should be changed to [Гринуич](https://bg.wikipedia.org/wiki/%D0%93%D1%80%D0%B8%D0%BD%D1%83%D0%B8%D1%87) in the locale data and, when done, in the LDML specification.

> On another note, the HTML version of [Part 5](https://unicode-org.github.io/cldr/ldml/tr35-collation.html) contains `:point_right:` instead of `👉`.

srl295:
> I noticed Гриинуич in Part 4. This should be changed to [Гринуич](https://bg.wikipedia.org/wiki/%D0%93%D1%80%D0%B8%D0%BD%D1%83%D0%B8%D1%87) in the locale data and, when done, in the LDML specification.

Can you file a separate ticket?

ivan-pan:
@srl295 Oh, it actually is already corrected in the current CLDR version, now it needs to be corrected in the specification.

PR #893, opened 2020-12-20
CLDR-14251 Modernize: remove v.jsp; add SurveyTool.java, CldrGui.js

btangmu:
-New SurveyTool.java with most of the Java from v.jsp

-New CldrGui.js with most of the HTML from v.jsp

-Remove SurveyAjax.includeEvilJavaScript; handle more such items with CldrStatus.js

-Move SurveyAjax.includeJavaScript and related code to SurveyTool.java

-Extend statusJSON, new class StatusForFrontEnd to include more info needed by front end

-More js unit tests; encapsulate parseAsMimeType in new TestCldrTest.js

-Change dojo config parseOnLoad to false, instead use dojo/parser in cldrGui.run

-In redesign/review.js, use dojo/ready instead of jquery equivalent: wait for cldrGui.run

-Skip dojo/ready where superfluous

-Use XMLHttpRequest instead of dojo/request in cldrStAjax, per vanilla-js.com

-Rename dojo dependencies in showV for clarity, e.g. dojoxBusyButton for BusyButton

-Normalize error handlers for cldrStAjax.sendXhr: err parameter is always a string

-New CldrText.js to replace stui.str and stui.sub; no more dojo/string or loadStui

-New WHAT_ADMIN_PANEL so front end does not need vap for URL if logged in as admin

-Restore/revise comment about fonts in surveytool.css (removed in PR 877)

-Fix CLDR-14365 by calling hideRightPanel from loadHandler for Date, Zones, Numbers reports

-Run npx prettier on all .js files except those that are minified

-Remove dead code; clean up

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14251
- [x] Updated PR title and link in previous line to include Issue number

PR #1843, opened 2022-03-28
CLDR-15369 Priority Items Summary, fix memory shortage; refactoring

btangmu:
-Add liberty/config/jvm.options with 7 GB to avoid summary failure due to low memory on cldr-smoke

-Report Survey Tool memory stats in the About page

-Make CookieSession.stuff private, use its synchronized accessor method in SurveyAjax

-Move some of SurveyMain.init code into new subroutines for clarity

-Comments

CLDR-15369


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -410,44 +410,64 @@ public final void init(final ServletConfig config) throws ServletException {
...
+        if (!startUpDatabase()) {
+            return;
+        }
+        // TODO: call Summary.scheduleAutomaticSnapshots here or in doStartup
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-15369
 
+        // TODO: check whether doStartup and startUpDatabase have a synchronization problem,
+        // sometimes this.dbUtils null when doStartup calls doStartupDb
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-15472
+    }
```
srl295, line 434: does sound like some static init may have changed.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/About.java
```diff
@@ -99,7 +99,14 @@ public Response getAbout() {
...
         return Response.ok(r).build();
     }
+
+    private static String getMemoryStats() {
+        Runtime rt = Runtime.getRuntime();
+        return "total: " + rt.totalMemory() +
+            "; max: " + rt.maxMemory() +
+            "; free: " + rt.freeMemory();
+    }
 }
```
srl295, line 112: instead, please send individual numbers instead of a string, so that we can programmatically track it!  Then, assemble the string on the client side.

srl295: also #893 (2 years ago) removed these fields from the `SurveyAjax?what=status` endpoint -> https://github.com/unicode-org/cldr/pull/893/files#diff-d994bfe2804309e4400822a8cd2bed66b3a1e05ab5ef0f518a81e9738af6ee70L1317

Can you reinstate those?

srl295: I'll open a separate PR

srl295: Opened https://github.com/unicode-org/cldr/pull/1845

srl295 (approved):
Approving, but would recommend sending memory as separate fields

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/About.java
btangmu: I'll make a subsequent PR to split up the memory stats

tools/cldr-apps/src/main/java/org/unicode/cldr/web/CookieSession.java
```diff
@@ -60,7 +60,7 @@ public class CookieSession {
     static final boolean DEBUG_INOUT = false;
     public String id;
     public String ip;
-    public Hashtable<String, Object> stuff = new Hashtable<>(); // user data
+    private Hashtable<String, Object> stuff = new Hashtable<>(); // user data
```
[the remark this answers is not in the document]
btangmu: ooops -- searching for callers doesn't work with jsp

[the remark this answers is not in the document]
btangmu: sorry about that -- and sorry we still have jsp!

PR #1844, opened 2022-03-28
CLDR-15405 Add initialPatterns, sampleNames & data; PathHeader functions for ordering

pedberg-icu:
CLDR-15405

Finish up the remaining changes to dtd, coverage, PathHeader etc. modulo discussion about changing some of the added items, and add new data items for root/English:
- Add initialPattern element with type=initial|initialSequence, and data for root/English. Since this is PCDATA, it is also required to have an alt attribute, so it allows alt=variant.
- Add sampleName element with item=minimal|simple|full|multiword, and child element nameField with type=prefix|given|given2|surname|surname2|suffix, and initial data for English. Since nameField is PCDATA, it also allows alt=variant. **Note: we need to discuss (in meeting): (1) whether "item" is the best name for the sampleName attribute; (2) what additional values might be needed besides minimal|simple|full|multiword; and (3) whether the initial English data is what we want.**
- Fix the namePattern alt attribute to be 1|2, not 2|3.
- Update coverage, PathHeader, PathDescription, TextExampleGenerator, etc. for the above changes.
- Add PathHeader ordering functions personNameOrder and sampleNameOrder, use them in  PathHeader.txt for correct ordering of paths.

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9129,6 +9129,8 @@ annotations.
 		<featureName type="zero">slashed zero</featureName>
 	</typographicNames>
 	<personNames>
+		<initialPattern type="initial">{0}.</initialPattern>
+		<initialPattern type="initialSequence">{0} {1}</initialPattern>
 		<personName>
 			<namePattern>{given-informal} {surname}</namePattern>
 		</personName>
@@ -9246,5 +9248,26 @@ annotations.
 		<personName length="monogramNarrow" style="informal" order="givenFirst">
 			<namePattern>{given-informal-initial}</namePattern>
 		</personName>
+		<sampleName item="minimal">
+			<nameField type="given">Katherine</nameField>
+			<nameField type="surname">Johnson</nameField>
+		</sampleName>
+		<sampleName item="simple">
+			<nameField type="given">Alberto</nameField>
+			<nameField type="given2">Pedro</nameField>
+			<nameField type="surname">Calderón</nameField>
+		</sampleName>
+		<sampleName item="full">
+			<nameField type="prefix">Dr.</nameField>
+			<nameField type="given">Dorothy</nameField>
+			<nameField type="given2">Lavinia</nameField>
+			<nameField type="surname">Brown</nameField>
+			<nameField type="suffix">M.D.</nameField>
+		</sampleName>
+		<sampleName item="multiword">
+			<nameField type="given">Erich Oswald</nameField>
+			<nameField type="given2">Hans Carl Maria</nameField>
+			<nameField type="surname">von Stroheim</nameField>
+		</sampleName>
 	</personNames>
 </ldml>
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -5369,6 +5369,8 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 	</typographicNames>
 	<personNames>
 		<nameOrder nameLocales="und" order="surnameFirst"/>
+		<initialPattern type="initial">{0}.</initialPattern>
+		<initialPattern type="initialSequence">{0} {1}</initialPattern>
 		<!-- fallback/any pattern -->
 		<personName>
 			<namePattern>{surname} {surname2}, {prefix} {given} {given2} {suffix}</namePattern>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
@@ -382,9 +382,17 @@ public class PathDescription {
         + RegexLookup.SEPARATOR
         + "Person name order for languages. For more information, please see "
         + CLDRURLS.PERSON_NAME_FORMATS + ".\n"
+        + "^//ldml/personNames/initialPattern\\[@type=\"([^\"]*)\"]"
+        + RegexLookup.SEPARATOR
+        + "Initials patterns for person name formats. For more information, please see "
+        + CLDRURLS.PERSON_NAME_FORMATS + ".\n"
         + "^//ldml/personNames/personName"
         + RegexLookup.SEPARATOR
         + "Person name formats by length, usage, style, order. For more information, please see "
+        + CLDRURLS.PERSON_NAME_FORMATS + ".\n"
+        + "^//ldml/personNames/sampleName"
+        + RegexLookup.SEPARATOR
+        + "Sample names for person name forma examples. For more information, please see "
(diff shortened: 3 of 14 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
```diff
@@ -382,10 +382,18 @@
         + RegexLookup.SEPARATOR
```
macchiati, line 382: I think we also need to do the PlaceHolderDescriptions (but not in this PR). I think our tests might ignore placeholders that aren't {0}, {1}, etc.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -1851,6 +1851,60 @@ public String transform(String source) {
...
+                    final List<String> lengthValues = Arrays.asList("long", "medium", "short", "monogram", "monogramNarrow");
+                    final List<String> usageValues = Arrays.asList("addressing", "referring", "sorting");
+                    final List<String> styleValues = Arrays.asList("formal", "informal");
+                    final List<String> orderValues = Arrays.asList("surnameFirst", "givenFirst");
+
+                    List<String> parts = HYPHEN_SPLITTER.splitToList(source);
+                    order = 0;
+                    for (String part: parts) {
+                        if (lengthValues.contains(part)) {
+                            order += lengthValues.indexOf(part) * 64;
```
macchiati, line 1868: I think this needs to be refined.
We might get collisions. 
long = 0 would be zero, and collide with addressing = 0.
Also medium + short (1 + 2) = monogram (3).

macchiati: What I suggest instead is having one List with all the values. When you get an index, add (1 << index) to the order.

macchiati (changes requested):
Changes could go into a later PR, if you want to get this progress in now.

common/main/en.xml
```diff
@@ -9246,5 +9248,26 @@ annotations.
...
 			<namePattern>{given-informal-initial}</namePattern>
 		</personName>
+		<sampleName item="minimal">
+			<nameField type="given">Katherine</nameField>
+			<nameField type="surname">Johnson</nameField>
+		</sampleName>
+		<sampleName item="simple">
+			<nameField type="given">Alberto</nameField>
+			<nameField type="given2">Pedro</nameField>
+			<nameField type="surname">Calderón</nameField>
```
richgillam, line 9258: For a Spanish name like this, would you want a `surname2` field as well?  Or would an example like that go somewhere else?

richgillam (commented):
I don't feel qualified to review a lot of this; I don't know enough about what it's doing or how it's supposed to work.  I'll defer to Mark.

common/main/en.xml
pedberg-icu: We need to discuss in the meeting what kinds of example names we want for English.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
pedberg-icu: I will add that to the TODOs in the agenda

common/main/en.xml
```diff
@@ -9246,5 +9248,26 @@ annotations.
...
+		</sampleName>
+		<sampleName item="full">
+			<nameField type="prefix">Dr.</nameField>
+			<nameField type="given">Dorothy</nameField>
+			<nameField type="given2">Lavinia</nameField>
+			<nameField type="surname">Brown</nameField>
+			<nameField type="suffix">M.D.</nameField>
+		</sampleName>
+		<sampleName item="multiword">
+			<nameField type="given">Erich</nameField>
```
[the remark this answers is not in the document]
pedberg-icu: Since thee were so many names in given2 I just moved the first to add to given.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
pedberg-icu: Oh, good idea, I will do that; I forgot we could have multiple instances of the same type.

pedberg-icu:
@macchiati I made some changes to address your feedback, thanks. I will leave un-squashed for now so you can just review the deltas, then will squash (and will need a re-approval).

common/main/root.xml
```diff
@@ -5369,6 +5369,8 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 	</typographicNames>
 	<personNames>
 		<nameOrder nameLocales="und" order="surnameFirst"/>
+		<initialPattern type="initial">{0}.</initialPattern>
+		<initialPattern type="initialSequence">{0}{1}</initialPattern>
```
macchiati, line 5373: I think more languages use spaces between initials than not, so maybe this should also be space separated.

common/main/en.xml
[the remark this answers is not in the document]
macchiati: ok

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
macchiati: SGTM

common/main/en.xml
macchiati: We don't have to have names for specific people. I think it is actually useful to have someone with surname2 for English, to show that we *don't* included in normal formatting there.

macchiati:
> @macchiati I made some changes to address your feedback, thanks. I will leave un-squashed for now so you can just review the deltas, then will squash (and will need a re-approval).

Ok, will do now.

macchiati (approved):
Changes you made look good. If you want to merge, you can address other comments in a later PR.

common/main/root.xml
pedberg-icu: @macchiati OK, I added spaces (I only had it without spaces because that is the way it was in the agenda to-dos), and have squashed. Those are the only changes, needs re-approval to merge, thanks.

common/main/en.xml
pedberg-icu: Sounds good. I need to update the names here based on discussion in meeting; also need to add more same name item types. Will do that all in another PR.

pedberg-icu:
@macchiati Needs re-approval after squashing (and adding spaces in the root initialSequence pattern), thanks. - Peter

macchiati (approved):

macchiati:
Now mergeable

PR #1845, opened 2022-03-28
CLDR-15369 re-instate memfree/memtotal

srl295:
- reinstate 'memfree' and 'memtotal'
These were removed in #893 but are read by Prometheus.

CLDR-15369

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
@@ -3678,6 +3678,10 @@ public class SurveyMain extends HttpServlet implements CLDRProgressIndicator, Ex
         return new StatusForFrontEnd(this, request).toJSONObject();
     }
 
+    /**
+     * NOTE: the data in this status object is actually used by Prometheus monitoring.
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -3701,6 +3701,10 @@ public JSONObject statusJSON(HttpServletRequest request) throws JSONException {
         private int users = CookieSession.getUserCount();
         private String sessionMessage = null;
 
+        private Runtime r = Runtime.getRuntime();
+        double memtotal = r.totalMemory() / 1024000.0;
+        double memfree = r.freeMemory() / 1024000.0;
+
         private JSONObject toJSONObject() throws JSONException {
```
btangmu, line 3708: how about adding a comment about prometheus here? Otherwise it's natural to assume Survey Tool is the only client for status requests, and someone will just delete these again in the belief that they're dead code

srl295:
Example return from `SurveyAjax?what=status` (other fields removed for clarity)

```json
{
  "status": {
    "memtotal": 3325.952,
    "memfree": 1393.606625
  }
}
```

btangmu (approved):

PR #1854, opened 2022-03-29
CLDR-15384 Add code to elide literal text around empty fields.

richgillam:
CLDR-15384

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -262,8 +262,7 @@ public class PersonNameFormatter {
         }
         @Override
         public String toString() {
-            // TODO Rich escape \ and { in literals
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -262,8 +262,7 @@ public static NamePatternElement from(Object element) {
         }
         @Override
         public String toString() {
-            // TODO Rich escape \ and { in literals
-            return literal != null ? literal : modifiedField.toString();
+            return literal != null ? literal.replaceAll("\\", "\\\\").replaceAll("{", "\\{") : modifiedField.toString();
```
macchiati, line 265: .replace doesn't use regex, and would be a bit simpler.

macchiati (approved):
The change to replaceAll can wait on a later PR

macchiati:
Rich, do you need me to merge?

PR #1858, opened 2022-03-30
CLDR-14332 update docs to fix links that were to 'latest'

srl295:
CLDR-14332

Did not add a notice yet, but fixed links


srl295:
@pedberg-icu I noticed that you had some links that went into maint/maint-41 instead of 'latest', is this purposeful?

pedberg-icu:
> @pedberg-icu I noticed that you had some links that went into maint/maint-41 instead of 'latest', is this purposeful? (I left them for now)

That was intentional since the old versions went to maint-40, and I think the intent was that they link to versioned data that matches the spec version

PR #1865, opened 2022-03-31
CLDR-14543 disable tracking XMLSource.sourceLocation unless UNITTEST 

srl295:
only enable source location tracking in UNITTEST

reduces memory use and object creation

CLDR-14543


tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
```diff
@@ -1878,6 +1879,18 @@ public boolean equalsOrInheritsCurrentValue(String value, String curValue, Strin
         return false;
     }
 
+    /**
+     * Enable SourceLocation in this Source. Call this once before populating the XMLSource.
```
macchiati, line 1883: I'm guessing that these must be overridden to make sense. Is that true?

srl295: Yes, and they are overridden in SimpleXMLSource

btangmu (approved):

PR #1824, opened 2022-03-22
CLDR-15467 update maven.yml to improve builds

srl295:
CLDR-15467

- [ ] This PR completes the ticket.

PR #1868, opened 2022-03-31
CLDR-15460 build: restructure workflows

srl295:
- run ansible/vagrant only if ansible changes
- run js test only if js changes
- move yaml lint of the workflows into maven

CLDR-15460

Includes fix from: CLDR-15467 update maven.yml to improve builds (#1824) 2ea074bfc0716359c24e47b757b6027bc78eb69f

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -1,6 +1,14 @@
 name: Ansible Lint
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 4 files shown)


srl295:
@pedberg-icu  #1824  might not have merged cleanly, so I've included it here as it is related.

pedberg-icu (approved):
All seems to make sense...

PR #1869, opened 2022-04-01
CLDR-15405 nameOrder→nameOrderLocales, mv sorting from usage→order, sampleNames in root / ∅∅∅ in en

pedberg-icu:
CLDR-15405

Various items to unblock other work:
1. Change nameOrder to nameOrderLocales, make order attribute distinguishing, move locale list from attribute to value. Use root/en data suggested by Mark.
2. Move sorting from usage to order. Rearrange root/en data to match.
3. For sampleName, expanded & renamed items types, included givenInformal. Added root data with all nameFields for each item; adjusted the en data with ∅∅∅ for unused fields (add PathDescriotion note for that). The en sampleName data needs better values, current ones are just to have something for testing 
4. Misc en updates. remove . in mononym patterns since it is in initialPattern; use -initial-allCaps for all mononym patterns. 
5. Use PathHeader ordering functions for personName/sampleName section heads too.

Still to do (another PR): 
6. PlaceHolderDescriptions for non-{0},{1}-type placeholders
7. Better en sampleName data

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9129,145 +9129,212 @@ annotations.
 		<featureName type="zero">slashed zero</featureName>
 	</typographicNames>
 	<personNames>
+		<nameOrderLocales order="givenFirst">und</nameOrderLocales>
+		<nameOrderLocales order="surnameFirst">ja zh ko</nameOrderLocales>
 		<initialPattern type="initial">{0}.</initialPattern>
 		<initialPattern type="initialSequence">{0} {1}</initialPattern>
 		<personName>
 			<namePattern>{given-informal} {surname}</namePattern>
 		</personName>
-		<personName length="long" usage="addressing" style="formal" order="surnameFirst">
-			<namePattern>{prefix} {surname}</namePattern>
+		<personName length="long" usage="addressing" style="formal" order="sorting">
+			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
 		</personName>
 		<personName length="long" usage="addressing" style="formal" order="givenFirst">
 			<namePattern>{prefix} {surname}</namePattern>
 		</personName>
-		<personName length="long" usage="addressing" style="informal" order="surnameFirst">
-			<namePattern>{given-informal}</namePattern>
+		<personName length="long" usage="addressing" style="formal" order="surnameFirst">
+			<namePattern>{prefix} {surname}</namePattern>
+		</personName>
+		<personName length="long" usage="addressing" style="informal" order="sorting">
+			<namePattern>{surname}, {given-informal}</namePattern>
 		</personName>
 		<personName length="long" usage="addressing" style="informal" order="givenFirst">
 			<namePattern>{given-informal}</namePattern>
 		</personName>
-		<personName length="long" usage="referring" style="formal" order="surnameFirst">
+		<personName length="long" usage="addressing" style="informal" order="surnameFirst">
+			<namePattern>{given-informal}</namePattern>
+		</personName>
+		<personName length="long" usage="referring" style="formal" order="sorting">
 			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
 		</personName>
 		<personName length="long" usage="referring" style="formal" order="givenFirst">
 			<namePattern>{given} {given2} {surname} {suffix}</namePattern>
 		</personName>
-		<personName length="long" usage="referring" style="informal" order="surnameFirst">
+		<personName length="long" usage="referring" style="formal" order="surnameFirst">
+			<namePattern>{surname}, {given} {given2} {suffix}</namePattern>
+		</personName>
+		<personName length="long" usage="referring" style="informal" order="sorting">
 			<namePattern>{surname}, {given-informal}</namePattern>
 		</personName>
 		<personName length="long" usage="referring" style="informal" order="givenFirst">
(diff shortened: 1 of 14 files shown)


common/main/en.xml
```diff
@@ -9129,145 +9129,212 @@ annotations.
...
+		<personName length="monogramNarrow" style="informal" order="surnameFirst">
+			<namePattern>{given-informal-initial-allCaps}</namePattern>
 		</personName>
-		<sampleName item="minimal">
+		<!-- The following samples are temporary, to be filled out with something better -->
+		<sampleName item="givenSurname">
+			<nameField type="prefix">∅∅∅</nameField>
 			<nameField type="given">Katherine</nameField>
+			<nameField type="given2">∅∅∅</nameField>
+			<nameField type="givenInformal">∅∅∅</nameField>
```
macchiati, line 9289: this needs to be given-informal

macchiati: and elsewhere.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -1878,8 +1878,8 @@ public String transform(String source) {
             functionMap.put("sampleNameOrder", new Transform<String, String>() {
                 @Override
                 public String transform(String source) {
-                    final List<String> sampleItemValues = Arrays.asList("minimal", "simple", "full", "multiword");
-                    final List<String> fieldTypeValues = Arrays.asList("prefix", "given", "given2", "surname", "surname2", "suffix");
+                    final List<String> sampleItemValues = Arrays.asList("givenSurname", "given2Surname", "givenSurname2", "informal", "full", "multiword");
+                    final List<String> fieldTypeValues = Arrays.asList("prefix", "given", "given2", "givenInformal", "surname", "surname2", "suffix");
```
macchiati, line 1882: given-informal

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -1308,8 +1313,10 @@ public PersonNameFormatter(CLDRFile cldrFile) {
             if (path.startsWith("//ldml/personNames/sampleName")) {
                 //ldml/personNames/sampleName[@item="full"]/nameField[@type="prefix"]
                 String value = cldrFile.getStringValue(path);
-                XPathParts parts = XPathParts.getFrozenInstance(path);
-                names.put(SampleType.valueOf(parts.getAttributeValue(-2, "item")), ModifiedField.from(parts.getAttributeValue(-1, "type")), value);
+                if (!value.equals("∅∅∅")) { // TODO is this how we want to handle ∅∅∅?
```
macchiati, line 1316: yes, that should work.

macchiati (changes requested):
Nice! I had only one substantive comment; use of -informal

richgillam (dismissed):

common/main/en.xml
pedberg-icu: I thought we said that we wanted attribute values to be camelCase (as opposed to the placeholder values, which can have appended hyper-separated modifiers)? Of course this means a mapping in the enums between the attribute value and the placeholder value).

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -208,6 +210,9 @@ public ModifiedField(Field field, Modifier... modifiers) {
 
         /** convenience method for testing */
         public static ModifiedField from(String string) {
+            if (string.equals("givenInformal")) {
+                string = "given-informal"; // TODO: hack for handling nameField types vs namePattern values
+            }
```
pedberg-icu, line 215: This is where I currently map from the attribute value to the placeholder value

macchiati:
No, it uses the same syntax. It is only *within* a single field or modifier
(eg allCaps) that we use 🐪 instead of -

common/main/en.xml
macchiati, line 9289:
```suggestion
			<nameField type="given-informal">∅∅∅</nameField>
```

macchiati: Hmmm, I didn't even know there was an "add suggestion"

pedberg-icu:
> No, it uses the same syntax. It is only *within* a single field or modifier (eg allCaps) that we use 🐪 instead of -

OK, will fix. Then for ordering of the nameField type attribute values I will treat modifiers like "-informal" as less significant than the base names they modify.

macchiati:
Sounds good, thanks.

Note: if it would make the code cleaner to put ordering utilities into
PersonNameFormatter we could do that.

pedberg-icu:
> Sounds good, thanks. Note: if it would make the code cleaner to put ordering utilities into PersonNameFormatter we could do that.

I think better to have them in PathHeader.java, that is the main place they are relevant (I think)

pedberg-icu:
OK, changed nameField type attribute value givenInformal → given-informal (like the namePattern placeholder), adjusted OathHeader sampleNameOrder function accordingly, and squashed. @macchiati, can you re-review? Thanks.

macchiati (approved):

PR #1875, opened 2022-04-05
CLDR-15405 replace "any" entry with 3 root fallbacks by order; add/use &personNameSection

pedberg-icu:
CLDR-15405

- Remove the fallback "any" personName entry in root & en. Instead, in root provide explicit patterns for the 3 different orders (with length="long" usage="referring" style="formal" ) and have other personName entries fall back to the appropriate one of those.
- As a result, remove ParameterMatcher.MATCH_ALL and adjust code accordingly. Update tests that passed an empty string of attribute values to now pass one or more fallback strings with order attribute set.
- Add PathHeader function personNameSection and use it in PathHeader.txt
- Remove the logKnownIssue skip around a test in TestCldrFactory.testWrite, the original test now works fine.
-  Other minor fixes

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -5372,159 +5372,155 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<nameOrderLocales order="surnameFirst">ja zh ko hu</nameOrderLocales>
 		<initialPattern type="initial">{0}.</initialPattern>
 		<initialPattern type="initialSequence">{0} {1}</initialPattern>
-		<!-- fallback/any pattern -->
-		<personName>
-			<namePattern>{surname} {surname2}, {prefix} {given} {given2} {suffix}</namePattern>
-		</personName>
 		<personName length="long" usage="addressing" style="formal" order="sorting">
-			<alias source="locale" path="../personName"/>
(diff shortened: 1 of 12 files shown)


richgillam (approved):
I'm sorry-- I'm giving a ship-it in case you need to get this in quickly, but I don't feel like I understand this code well enough for my ship-it to have any real value.  You might want to wait for somebody else (Mark?) to take a look.

common/main/root.xml
```diff
@@ -5372,159 +5372,155 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
 		</personName>
 		<personName length="monogramNarrow" style="informal" order="surnameFirst">
-			<alias source="locale" path="../personName"/>
+			<alias source="locale" path="../personName[@length='long'][@usage='referring'][@style='formal'][@order='surnameFirst']"/>
 		</personName>
 		<sampleName item="givenSurname">
 			<nameField type="prefix">Prefix</nameField>
 			<nameField type="given">Given</nameField>
-			<nameField type="given-informal">given-informal</nameField>
+			<nameField type="given-informal">givenInformal</nameField>
```
pedberg-icu, line 5523: Note the "givenInformal" here is just a dummy name, not an attribute or placeholder value.

macchiati (approved):

PR #1877, opened 2022-04-06
CLDR-15404 Enhance error checking

macchiati:
CLDR-15404

- Add checks for nameOrderLocales and sampleName
- Make LocaleMatchValue public, for ease of checking validity of locales
- Add information to exceptions thrown by NamePattern.parse
- Fix the TestPersonNameFormatter to account for the more detailed messages

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
@@ -3,6 +3,7 @@ package org.unicode.cldr.test;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
```diff
@@ -38,146 +53,198 @@ public CheckCLDR handleCheck(String path, String fullPath, String value, Options
...
+                    result.add(new CheckStatus().setCause(this)
+                        .setMainType(CheckStatus.errorType)
+                        .setSubtype(Subtype.invalidPlaceHolder)
+                        .setMessage("Invalid locales: \"" + JOIN_SPACE.join(orderErrors) + "\""));
+                }
+                return this;
+
+            case "sampleName":
+                //ldml/personNames/sampleName[@item="informal"]/nameField[@type="surname"]
+                if (value.equals("∅∅∅")) {
```
pedberg-icu, line 86: why this? (not blocking, just curious)

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
macchiati: Right now it raises an error 
1. on given if it is empty or 
2. on surname if it is empty while surname2 isn't

So in both cases the one if flags is empty. I could structure it a bit different if you think I should also flag surname2 in case 2.

pedberg-icu: Oh right, I get it now, thanks. Fine as is.

PR #1881, opened 2022-04-06
CLDR-15088 specs: fix mixed content, improve build

srl295:
CLDR-15088

- fix mixed content in html header
- move html header to separate html file
- commented the archive.js generator
- made it so it doesn't download css or gif unnecessarily


srl295:
not new issues for 41, so i don't think it's worth copying to maint 41… at least not now

rickmcgowan (approved):

pedberg-icu (approved):

PR #1884, opened 2022-04-09
CLDR-14906 Retain Dashboard hidden status across releases

btangmu:
-Revise DASH_HIDE db table: unversioned; new column val (blob)

-DASH_HIDE subtype column, CheckCLDR.CheckStatus.Subtype instead of VettingViewer.Choice

-Include value and subtype in http requests and responses

-Consistently spell subtype with lowercase t in all files

-New small class PathValuePair.java

-New HiddenNotifications.java to encapsulate complex HashMap

-New subclasses of VettingViewer etc to encapsulate arguments and results

-Refactor single-path dashboard to use the same code as multi-path dashboard

-Refactor ReviewHide using HiddenNotifications and move inline SQL to constants

-Rename some items for consistency to reduce confusion

-Distinguish ReviewNotificationGroup groups from ReviewEntry entries

-Shorten methods by moving code into new subroutines

-New unit tests for cldrDash.js in TestCldrDash.js and dash_data.js

-Remove Deprecated annotation from SurveyAjax.getRow

-Postpone revising unfinished modern replacement VoteAPIHelper.handleGetRows; error if getDashboard

-Comments

CLDR-14906

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
@@ -1,26 +1,14 @@
 package org.unicode.cldr.web;
 
-import java.util.ArrayList;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
 
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
-import org.json.JSONArray;
-import org.unicode.cldr.util.CLDRFile;
-import org.unicode.cldr.util.CLDRLocale;
-import org.unicode.cldr.util.Factory;
-import org.unicode.cldr.util.Level;
-import org.unicode.cldr.util.Organization;
+import org.unicode.cldr.test.CheckCLDR;
+import org.unicode.cldr.util.*;
 import org.unicode.cldr.util.PathHeader.PageId;
 import org.unicode.cldr.util.PathHeader.SectionId;
-import org.unicode.cldr.util.VettingViewer;
 import org.unicode.cldr.util.VettingViewer.Choice;
-import org.unicode.cldr.util.VoterProgress;
 
 import com.ibm.icu.impl.Relation;
 import com.ibm.icu.impl.Row;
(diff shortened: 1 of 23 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
```diff
@@ -64,33 +52,42 @@ public ReviewNotification add(ReviewNotification notification) {
...
         /**
-         * Notifications that the user has chosen to hide
+         * Get a map from each subtype (String from CheckCLDR.CheckStatus.Subtype)
+         * to an array of path-value pairs
+         *
+         * @return the map
          */
-        public HashMap<String, List<String>> hidden;
+        public Map<String, PathValuePair[]> getHidden() {
+            return hidden.getHidden();
```
btangmu, line 65: For single-path requests, getHidden and voterProgress are superfluous -- for performance, I intend to bypass them in a subsequent PR

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
```diff
@@ -1,26 +1,14 @@
 package org.unicode.cldr.web;
 
-import java.util.ArrayList;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
+import java.util.*;
```
srl295, line 3: Why the *? seems to be the best practice to give explicit imports.

tools/cldr-apps/src/main/webapp/tc-all-errors.jsp
```diff
@@ -59,7 +59,7 @@ if(request.getParameter("flush") != null) {
 <body>
 
 
-<h1>All error subTypes</h1>
+<h1>All error subtypes</h1>
```
srl295, line 62: this and following, not a bad change, doesn't seem related to dashboard…

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
btangmu: IntelliJ or VS Code did it automatically without my asking, sorry!

tools/cldr-apps/src/main/webapp/tc-all-errors.jsp
btangmu: right -- globally replacing subType with subtype was easier and seems to me safer and better in the long run

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
srl295: ok. must have some threshhold, 'over 10 lines and we put a * in'. No objection…

PR #1890, opened 2022-04-11
CLDR-15219 Survey Tool is missing favicon

btangmu:
-Add favicon.ico (copied from home.unicode.org) in webapp folder

-Link to favicon.ico from html header generated in SurveyTool.java

-Also add a favicon.ico to go with tools/cldr-apps/js/test/Test.html

CLDR-15219


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -20,6 +20,7 @@ public class SurveyTool extends HttpServlet {
     static final Logger logger = SurveyLog.forClass(SurveyTool.class);
 
     private static final long serialVersionUID = 1L;
+    private static final String FAVICON_LINK = "<link rel='icon' href='/cldr-apps/favicon.ico' type='image/x-icon'>\n";
```
srl295, line 23:
```suggestion
    private static final String FAVICON_LINK = "<link rel='icon' href='./favicon.ico' type='image/x-icon'>\n";
```

or relative to context root.  We need to get rid of all static `cldr-apps` paths

srl295 (changes requested):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
btangmu: good idea!

srl295 (approved):

PR #1892, opened 2022-04-11
CLDR-15404 fix the internal caching in example generator, part 1

macchiati:
CLDR-15404

There were problems where a change in a sample name wasn't reflected in the examples for patterns. That doesn't work for two reasons:

1. The examples are cached per path, so wouldn't get automatically refreshed. 
2. For *generating* those examples, two objects are used that relatively expensive to create, and were being cached across invocations.  

For the first problem, I added paths to ExampleDependencies. There is a call that uses that in ExampleGenerator to clear examples that are dependent on given paths.

For the second problem, there was no built-in mechanism for that. You could clear examples, but not arbitrary other constructs. So I extended the cache mechanism a bit so that I could clear out other items when a path changed.

The synchronization is a bit heavy handed, because I wanted to get the functionality working. There might be a bit of contention when the same locale is being worked on by 2 people.

In a subsequent PR I want to get some advice on better synchronization to decrease possible contention, but want at least to get the functionality correct on smoketest.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleCache.java
Adds the ability to add other kinds of objects to be handled if a path changes.

Adds a test that the examples for a path change when a path it depends on gets a different value.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleDependencies.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleDependencies.java
@@ -261,5 +261,11 @@ public class ExampleDependencies {
     .put("//ldml/units/unitLength[@type=\"*\"]/unit[@type=\"*\"]/unitPattern[@count=\"*\"]", "//ldml/listPatterns/listPattern[@type=\"*\"]/listPatternPart[@type=\"*\"]")
     .put("//ldml/units/unitLength[@type=\"*\"]/unit[@type=\"*\"]/unitPattern[@count=\"*\"]", "//ldml/units/unitLength[@type=\"*\"]/compoundUnit[@type=\"*\"]/compoundUnitPattern")
     .put("//ldml/units/unitLength[@type=\"*\"]/unit[@type=\"*\"]/unitPattern[@count=\"*\"]", "//ldml/units/unitLength[@type=\"*\"]/compoundUnit[@type=\"*\"]/compoundUnitPattern1[@count=\"*\"]")
+    .putAll("//ldml/personNames/sampleName[@item=\"*\"]/nameField[@type=\"*\"]",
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleDependencies.java
```diff
@@ -261,5 +261,11 @@ public class ExampleDependencies {
     .put("//ldml/units/unitLength[@type=\"*\"]/unit[@type=\"*\"]/unitPattern[@count=\"*\"]", "//ldml/listPatterns/listPattern[@type=\"*\"]/listPatternPart[@type=\"*\"]")
     .put("//ldml/units/unitLength[@type=\"*\"]/unit[@type=\"*\"]/unitPattern[@count=\"*\"]", "//ldml/units/unitLength[@type=\"*\"]/compoundUnit[@type=\"*\"]/compoundUnitPattern")
     .put("//ldml/units/unitLength[@type=\"*\"]/unit[@type=\"*\"]/unitPattern[@count=\"*\"]", "//ldml/units/unitLength[@type=\"*\"]/compoundUnit[@type=\"*\"]/compoundUnitPattern1[@count=\"*\"]")
+    .putAll("//ldml/personNames/sampleName[@item=\"*\"]/nameField[@type=\"*\"]",
+        "//ldml/personNames/personName[@length=\"*\"][@usage=\"*\"][@style=\"*\"][@order=\"*\"]/namePattern",
```
pedberg-icu, line 265: Should we also put any .../namePattern[@alt=\"*\"] if we have them? Perhaps a future addition...

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleDependencies.java
btangmu: Ideally the dependency will be auto-detected by TestExampleGenerator.main. Then ExampleDependencies.java will go back to being auto-generated, not hand-edited.

macchiati:
Agreed.

PR #1910, opened 2022-04-13
CLDR-15369 Priority Items Summary snapshots, improvements

btangmu:
-For snapshots, include only locales in SubmissionLocales.CLDR_LOCALES

-Fix snapshot schedule in case server is started between midnight and 1:00 AM

-Remove the word slow from the menu item

-Automatically show the most recent summary snapshot

-Display controls below the summary rather than above it

-Unrelated formatting change for Test.html (forgot to run prettier for favicon PR)

CLDR-15369

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrText.js
+++ b/tools/cldr-apps/js/src/esm/cldrText.js
@@ -467,7 +467,7 @@ const strings = {
   special_users: "Users",
   special_vetting_participation: "Vetting Participation",
   special_vetting_participation2: "Vetting Participation v2",
-  special_vsummary: "Priority Items Summary (slow)",
+  special_vsummary: "Priority Items Summary",
   special_test_panel: "Experimental Test Panel",
 
   lock_account_admin: "The Admin account cannot be locked.",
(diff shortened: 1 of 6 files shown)


btangmu:
@macchiati @srl295 if there's no objection, after this is merged, I'll set CLDR_AUTO_SNAP=true again in cldr.properties on cldr-smoke. Then we should start getting auto snapshots at one o'clock every morning, and if those work OK we can do the same on the production server -- that might complete the ticket

macchiati:
Sgtm

tools/cldr-apps/js/src/esm/cldrText.js
```diff
@@ -467,7 +467,7 @@ const strings = {
   special_users: "Users",
   special_vetting_participation: "Vetting Participation",
   special_vetting_participation2: "Vetting Participation v2",
-  special_vsummary: "Priority Items Summary (slow)",
```
srl295, line 470: Maybe when CLDR-14543 lands we can change this to `(fast)` :)

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrText.js
btangmu: good idea

PR #1911, opened 2022-04-13
CLDR-15381 Java formatting with prettier, tabWidth 4

btangmu:
-Add package.json with custom setting for .java files: tabWidth 4

CLDR-15381


package.json
```diff
@@ -0,0 +1,12 @@
+{
```
srl295, line 1: it's… a little unusual and confusing to have a package.json at the top level.  I think it might misrepresent what structure is underneath.

srl295: https://prettier.io/docs/en/configuration.html could we use a `.prettierrc.json` file instead?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -1,5 +1,19 @@
 package org.unicode.cldr.util;
 
+import com.google.common.base.Joiner;
```
srl295, line 3: is it going to conflict with eclipse and intellij's import structure?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -679,25 +854,54 @@ private static void writeList() {
...
-                .join(keywords) + "\"},");
+            System.out.println(
+                "{\"" +
+                s +
+                "\",\"" +
+                shortName +
+                "\",\"" +
+                Joiner.on("|").join(keywords) +
+                "\"},"
+            );
```
srl295, line 904: very much expanded

package.json
btangmu: yes, I think so

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: maybe in the same sense that it will conflict with their java formatting in general -- anytime we edit a file, regardless of which IDE or editor we use, we'll need to re-run prettier -- though we might get the IDE/editor to run it automatically, in which case i don't know if the IDE/editor will get in a fight with itself over the imports

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: right -- compare this version of the same code in my other PR https://github.com/unicode-org/cldr/pull/1913:
```
            System.out.println(
                "{\"" + s + "\",\"" + shortName + "\",\"" + Joiner.on("|").join(keywords) + "\"},"
            );
```
if you like that better, that's an argument for printWidth 100

srl295: i was just looking for a width parameter! Yes, let's do printwidth 100

package.json
srl295: I think that might make sense.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
srl295: the width is the only real issue i have otherwise lgtm

btangmu: great, lets see what @macchiati and others think

package.json
btangmu: see https://github.com/unicode-org/cldr/pull/1919

btangmu:
I prefer always to include braces even where optional. This formatter (prettier) doesn't add or remove braces, but as you say, Mark, there's always the option to run a different formatter to add braces, and then run prettier again

macchiati:
What I found strange is that prettier did make non-whitespace changes in
the text, notably adding () to returns, eg return a+b+c ==> return (a+b+c)
(but not always)

btangmu:
> What I found strange is that prettier did make non-whitespace changes in the text, notably adding () to returns, eg return a+b+c ==> return (a+b+c) (but not always)

Yes it's sort of inscrutable and sometimes does things that would never occur to me -- overall, though, I've been really happy with how it works for javascript. When the result is ugly it's sometimes because the original code is too dense, and creation of a new variable or subroutine will make it format better and be more maintainable

srl295:
> > What I found strange is that prettier did make non-whitespace changes in the text, notably adding () to returns, eg return a+b+c ==> return (a+b+c) (but not always)
> 
> Yes it's sort of inscrutable and sometimes does things that would never occur to me -- overall, though, I've been really happy with how it works for javascript. When the result is ugly it's sometimes because the original code is too dense, and creation of a new variable or subroutine will make it format better and be more maintainable

@btangmu is this the same prettier? It doesn't list Java nor options for java. https://prettier.io/docs/en/index.html

srl295:
OK… it's a plugin https://github.com/jhipster/prettier-java  … looks like there _may_ be some options in https://github.com/jhipster/prettier-java/blob/main/packages/prettier-plugin-java/src/options.js

btangmu:
> @btangmu is this the same prettier? It doesn't list Java nor options for java. https://prettier.io/docs/en/index.html

it's based on that, as a plugin:

npm install -g prettier prettier-plugin-java

https://www.npmjs.com/package/prettier-plugin-java

macchiati:
Closing, because we should at this point just import the mechanism that the Unicodetools group is using
> The Unicodetools group is looking at this, and considering incorporating a formatter in the build process. We should coordinate with them.

srl295:
@macchiati maybe, would need consensus on changing all source files and invalidating many in-progress PRs.  not impossible, just a bit higher impact than with unicodetools.

btangmu:
I don't have any in-progress PRs myself currently -- it will be a morale-booster when we do have Java formatting!

macchiati:
I agree. But also think that a "pull off the bandage" reformatting of all
the source code is by far the cleanest approach. Would take some timing.

PR #1913, opened 2022-04-13
CLDR-15381 Java formatting with prettier, tabWidth 4, printWidth 100

btangmu:
-Add package.json with custom settings for .java files

-Default for printWidth is 80; setting it to 100 results in fewer line breaks

CLDR-15381


btangmu:
@macchiati @srl295 this is the same as https://github.com/unicode-org/cldr/pull/1911 except that some lines tend to be longer

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -216,7 +242,10 @@ public static final class AnnotationSet {
         static final Factory factory = CONFIG.getCldrFactory();
         static final CLDRFile ENGLISH = CONFIG.getEnglish();
         static final CLDRFile ENGLISH_ANNOTATIONS = null;
-        static final SubdivisionNames englishSubdivisionIdToName = new SubdivisionNames("en", "main");
```
macchiati, line 219: This is definitely splitting lines that it doesn't need to. Maybe set the width higher?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -278,11 +319,18 @@ private String getStringValue(String xpath) {
...
             }
             String sourceLocale = cldrFile2.getSourceLocaleID(xpath, null);
-            if (sourceLocale.equals(XMLSource.CODE_FALLBACK_ID) || sourceLocale.equals(XMLSource.ROOT_ID)) {
-                if (!xpath.equals("//ldml/characterLabels/characterLabelPattern[@type=\"category-list\"]")) {
+            if (
+                sourceLocale.equals(XMLSource.CODE_FALLBACK_ID) ||
+                sourceLocale.equals(XMLSource.ROOT_ID)
+            ) {
+                if (
+                    !xpath.equals(
```
macchiati, line 330: In general, I find it harder to read, because it expands to too many lines.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -370,21 +421,26 @@ private Annotations synthesize(String code, Transform<String, String> otherSourc
...
-                        subdivisionName = MISSING_MARKER + subdivisionCode;
-//                    }
+                    //                    subdivisionName = englishSubdivisionIdToName.get(subdivisionCode);
+                    //                    if (subdivisionName != null) {
+                    //                        subdivisionName = ENGLISH_MARKER + subdivisionCode;
+                    //                    } else {
+                    subdivisionName = MISSING_MARKER + subdivisionCode;
+                    //                    }
                 }
-                String flagName = flagLabel == null ? subdivisionName : initialPattern.format(flagLabel, subdivisionName);
```
macchiati, line 387: But sometimes it is clearer.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -679,25 +758,48 @@ private static void writeList() {
...
         Set<String> keys = new TreeSet<>(ChartAnnotations.RBC);
         map.keySet().addAllTo(keys);
-//        keys.add("👩🏻‍⚖");
+        //        keys.add("👩🏻‍⚖");
         for (String key : keys) {
-            System.out.println(Utility.hex(key, 4, "_").toLowerCase(Locale.ROOT)
-                + "\t" + key
-                + "\t" + map.get(key).getShortName()
-                + "\t" + Joiner.on(" | ").join(map.get(key).getKeywords()));
+            System.out.println(
```
macchiati, line 763: This is really uglier and takes too many vertical lines. Original line breaks were more principled; new ones erase that.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -679,25 +758,48 @@ private static void writeList() {
...
         }
         for (String s : Arrays.asList(
-            "💏", "👩‍❤️‍💋‍👩",
-            "💑", "👩‍❤️‍👩",
-            "👪", "👩‍👩‍👧",
-            "👦🏻", "👩🏿",
-            "👨‍⚖", "👨🏿‍⚖", "👩‍⚖", "👩🏼‍⚖",
-            "👮", "👮‍♂️", "👮🏼‍♂️", "👮‍♀️", "👮🏿‍♀️",
-            "🚴", "🚴🏿", "🚴‍♂️", "🚴🏿‍♂️", "🚴‍♀️", "🚴🏿‍♀️")) {
+            "💏",
```
macchiati, line 774: This one really breaks the intent, since it flattens out the logical structure.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRPaths.java
```diff
@@ -34,105 +33,211 @@ public class CLDRPaths {
...
-        common_annotations(DtdType.ldml), common_casing(DtdType.ldml), common_collation(DtdType.ldml), common_main(DtdType.ldml), common_rbnf(
-            DtdType.ldml), common_segments(DtdType.ldml), common_subdivisions(DtdType.ldml),
-
-        common_supplemental(DtdType.supplementalData), common_transforms(DtdType.supplementalData), common_validity(DtdType.supplementalData),
-
-        keyboards_android(DtdType.keyboard, DtdType.platform), keyboards_chromeos(DtdType.keyboard, DtdType.platform), keyboards_dtd(DtdType.keyboard,
-            DtdType.platform), keyboards_osx(DtdType.keyboard,
-                DtdType.platform), keyboards_und(DtdType.keyboard, DtdType.platform), keyboards_windows(DtdType.keyboard, DtdType.platform),
-                ;
+        common_annotations(DtdType.ldml),
```
macchiati, line 223: This is much better. The jumbled structure was from an earlier application of a formatter that arbitrarily joined lines.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CaseIterator.java
```diff
@@ -48,164 +49,593 @@ public class CaseIterator {
...
         // a\N{MODIFIER LETTER RIGHT HALF RING}
-        { "a\u02BE", "A\u02BE", "a\u02BE", },
+        { "a\u02BE", "A\u02BE", "a\u02BE" },
         // ff
-        { "ff", "FF", "Ff", "fF", "ff", },
+        { "ff", "FF", "Ff", "fF", "ff" },
         // ffi
-        { "ffi", "FFI", "FFi", "FfI", "Ffi", "F\uFB01", "fFI", "fFi", "ffI", "ffi", "f\uFB01", "\uFB00I", "\uFB00i", },
+        {
+            "ffi",
```
macchiati, line 57: Another bad case. No advantage at all in splitting all these up; just wastes vertical lines.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ChainedMap.java
```diff
@@ -103,9 +129,18 @@ public Iterator<Entry<K3, Map<K2, Map<K1, V>>>> iterator() {
         public Iterable<Row.R4<K3, K2, K1, V>> rows() {
             List<R4<K3, K2, K1, V>> result = new ArrayList<>();
             for (Entry<Object, Object> entry0 : super.mapBase.entrySet()) {
-                for (Entry<Object, Object> entry1 : ((Map<Object, Object>) entry0.getValue()).entrySet()) {
-                    for (Entry<Object, Object> entry2 : ((Map<Object, Object>) entry1.getValue()).entrySet()) {
-                        R4<K3, K2, K1, V> item = (R4<K3, K2, K1, V>) Row.of(entry0.getKey(), entry1.getKey(), entry2.getKey(), entry2.getValue());
+                for (Entry<Object, Object> entry1 : (
+                    (Map<Object, Object>) entry0.getValue()
+                ).entrySet()) {
```
macchiati, line 134: Why on earth is it breaking (...).method() before ")."?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: I'll make another PR with printWidth 120

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: one way to address that, in this case, might be to assign the string literal "//ldml/characterLabels/characterLabelPattern[@type=\"category-list\"]" to a constant first

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: with this kind of formatter, that's bound to happen -- one approach would be to make the structure more explicit by defining separate arrays and then combining them, and that might be beneficial anyway in terms of separating the code from the data

if we really want a formatter to leave such code unchanged, it will have to be a very different kind of formatter, one that only enforces certain rules and otherwise doesn't change line-breaking or anything else. I don't know if such formatters exist...

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRPaths.java
btangmu: this illustrates why we might need to compromise; the formatter has to be allowed to change line-breaking in order to fix this

maybe there are formatters that will break lines but won't join them, I don't know

tools/cldr-code/src/main/java/org/unicode/cldr/util/ChainedMap.java
btangmu: looks like it's balancing parentheses

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: See https://github.com/unicode-org/cldr/pull/1919 -- it leaves this line unchanged

tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
btangmu: again, see https://github.com/unicode-org/cldr/pull/1919 which leaves this line unchanged

PR #1915, opened 2022-04-14
CLDR-15468 Locales.txt - steps towards support for focussed orgs

srl295:
CLDR-15468 Locales.txt: steps to support for focussed orgs

- new pathMatch element in coverageLevels.xml
- document Locales.txt format
- Split Locales.txt to a separate reader class
- parse @-commands
- tests for Locales.txt reader

Co-authored-by: Peter Edberg <42151464+pedberg-icu@users.noreply.github.com>

Note that this does NOT add IAAP, nor make any changes to Locales.txt content.
It also does not parse pathMatch.

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -1196,7 +1196,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!-- # Coverage levels -->
 
-<!ELEMENT coverageLevels ( approvalRequirements, coverageVariable*, coverageLevel* ) >
+<!ELEMENT coverageLevels ( approvalRequirements, coverageVariable*, coverageLevel*, pathMatch* ) >
     <!--@METADATA-->
 
 <!ELEMENT approvalRequirements ( approvalRequirement* ) >
@@ -1230,6 +1230,14 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@MATCH:literal/basic, core, minimal, moderate, modern, posix-->
     <!--@VALUE-->
 <!ATTLIST coverageLevel match CDATA #REQUIRED >
+    <!--@MATCH:any-->
+
+<!ELEMENT pathMatch EMPTY >
+<!ATTLIST pathMatch id NMTOKENS #IMPLIED >
+    <!--@MATCH:any-->
+
+<!ATTLIST pathMatch match CDATA #REQUIRED >
+    <!--@VALUE-->
     <!--@MATCH:any-->
 
 <!ELEMENT idValidity ( id* ) >
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -198,25 +198,25 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageVariable key="%teluNativeLanguages" value="(te)"/>
 		<coverageVariable key="%tibtDefaultLanguages" value="(dz)"/>
 		<coverageVariable key="%tibtNativeLanguages" value="(bo)"/>
-		
+
 		<!-- Additional variables for the restructuring -->
-		
+
 		<coverageVariable key="%basicDateSkeletons" value="(yMMMMEd|yMMMMd|yMMMd|yMd|Hmsv|hmsv|Hms|hms|Hm|hm)"/>
-		
+
 		<!-- -->
 		<!-- Coverage levels begin here -->
 		<!-- -->
 		<coverageLevel value="core" match="characters/exemplarCharacters"/>
 		<coverageLevel value="core" match="characters/exemplarCharacters[@type='%coreExemplarTypes']"/>
-		
+
 		<coverageLevel value="moderate" match="characters/exemplarCharacters[@type='%anyAlphaNum']"/>
 
 		<!-- *********************
-		Modified Basic rules (v41) 
+		Modified Basic rules (v41)
 		These contain a mixture of basic and moderate.
 		The only hard requirement is that each basic rule comes after any moderate rule *that could match the same path*.
 		********************* -->
-		
+
 		<coverageLevel	value="basic"	 	match="delimiters/alternateQuotationEnd"/>
 		<coverageLevel	value="basic"	 	match="delimiters/alternateQuotationStart"/>
 		<coverageLevel	value="basic"	 	match="delimiters/quotationEnd"/>
@@ -228,7 +228,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/localeDisplayPattern/localePattern"/>
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/localeDisplayPattern/localeSeparator"/>
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/codePatterns/codePattern[@type='(language|script|territory)']"/>
-					
+
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/languages/language[@type='${Target-Language}']"/>
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/languages/language[@type='${Target-Language}'][@alt='%anyAttribute']"/>
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/languages/language[@type='en']"/>
@@ -237,24 +237,24 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/languages/language[@type='%language30'][@alt='%anyAttribute']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/languages/language[@type='%language40']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/languages/language[@type='%language40'][@alt='%anyAttribute']"/>
-					
+
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/scripts/script[@type='${Target-Scripts}']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/scripts/script[@type='%script30']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/scripts/script[@type='%script30'][@alt='%anyAttribute']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/scripts/script[@type='%script40']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/scripts/script[@type='%script40'][@alt='%anyAttribute']"/>
-					
+
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/territories/territory[@type='${Target-Territories}']"/>
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/territories/territory[@type='${Target-Territories}'][@alt='(%anyAttribute)']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/territories/territory[@type='%territory30']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/territories/territory[@type='%territory30'][@alt='%anyAttribute']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/territories/territory[@type='%territory40']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/territories/territory[@type='%territory40'][@alt='%anyAttribute']"/>
-					
+
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/measurementSystemNames/measurementSystemName[@type='(metric|UK|US)']"/>
-					
+
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/annotationPatterns/annotationPattern[@type='(flag|keycap|emoji|combined)']"/>
-					
+
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/localeDisplayPattern/localeKeyTypePattern"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/types/type[@key='calendar'][@type='${Calendar-List}']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/types/type[@key='calendar'][@type='gregorian']"/>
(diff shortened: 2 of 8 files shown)


common/supplemental/coverageLevels.xml
```diff
@@ -924,5 +924,17 @@ For terms of use, see http://www.unicode.org/copyright.html
...
+		<pathGroups>
+			<pathGroup id="characterLabel">
+				<pathMatch match="characterLabels/characterLabelPattern[@type='%anyAttribute']"/>
+				<pathMatch match="characterLabels/characterLabelPattern[@type='%anyAttribute'][@count='%allPlurals']"/>
+			</pathGroup>
+			<pathGroup id="annotations">
+				<pathMatch match="annotations/annotation[@cp='%modernEmoji'][@type='%anyAttribute']"/>
+				<pathMatch match="annotations/annotation[@cp='%modernEmoji']"/>
+			</pathGroup>
+			<pathGroup id="characters_and_annotations" includeAll="annotations characterLabel"/>
```
srl295, line 936: this is a pathGroup that inherits from two others (and has no other paths, but it could)

common/supplemental/coverageLevels.xml
```diff
@@ -924,5 +924,17 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel value="modern" match="numbers/minimalPairs/caseMinimalPairs[@case='%anyAlphaNum']"/>
         <coverageLevel value="modern" match="numbers/minimalPairs/genderMinimalPairs[@gender='%anyAlphaNum']"/>
 
+		<pathGroups>
+			<pathGroup id="characterLabel">
+				<pathMatch match="characterLabels/characterLabelPattern[@type='%anyAttribute']"/>
+				<pathMatch match="characterLabels/characterLabelPattern[@type='%anyAttribute'][@count='%allPlurals']"/>
+			</pathGroup>
```
srl295, line 931: this mirrors the `<coverageLevel>` lines for characterLabelPattern, but without the coverage level.  I'm trying to make these easy to build and maintain, and also to keep Locales.txt as simple as possible.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
```diff
@@ -968,3 +968,5 @@ wikimedia ; * ; moderate ;
 
 yahoo ; mul ; modern ;
 # ^no votes in 30…40 for this org
+
+iaap ; @weight=4 @pathGroup=characters_and_annotations * ; moderate ;
```
srl295, line 972: I put the @ in the locale section, because it is a modifier on the allowed locales.

common/dtd/ldmlSupplemental.dtd
```diff
@@ -1225,6 +1225,25 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+    <!--@MATCH:any-->
+
+<!-- a pathMatch is basically a coveragelevel minus thevalue -->
+<!ELEMENT pathMatch EMPTY >
+<!ATTLIST pathMatch inLanguage CDATA #IMPLIED >
+    <!--@MATCH:any-->
+<!ATTLIST pathMatch inScript CDATA #IMPLIED >
+    <!--@MATCH:validity/script-->
+<!ATTLIST pathMatch inTerritory CDATA #IMPLIED >
+    <!--@MATCH:any-->
```
pedberg-icu, line 1242: Uh, why not `@MATCH:validity/region`?

srl295:
```suggestion
    <!--@MATCH:validity/region-->
```

sure... maybe coverageLevel needs that also

common/dtd/ldmlSupplemental.dtd
```diff
@@ -1225,6 +1225,25 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+<!ELEMENT pathGroup ( pathMatch* ) >
+<!ATTLIST pathGroup id NMTOKEN #REQUIRED >
+    <!--@MATCH:any-->
+<!ATTLIST pathGroup includeAll NMTOKENS #IMPLIED wip -  >
+    <!--@MATCH:any-->
+
+<!-- a pathMatch is basically a coveragelevel minus thevalue -->
+<!ELEMENT pathMatch EMPTY >
+<!ATTLIST pathMatch inLanguage CDATA #IMPLIED >
+    <!--@MATCH:any-->
```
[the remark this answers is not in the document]
srl295:
```suggestion
    <!--@MATCH:validity/language-->
```

srl295:
> Format looks good to me. But I am not involved with adding orgs/vetters. Maybe have Meike and /or Fredrik take a look? (At least at the Locales.txt example)

OK.

srl295:
@meikehubert  @stenshamn   ?thoughts on the format?

stenshamn:
Don't see any red flags as such, but also not sure if I'm looking in the right places. I see we have some old comment sections so I suggest you add to those, so when someone comes to these files in a few years there is guidance on how to augment them.

srl295:
> Don't see any red flags as such, but also not sure if I'm looking in the right places. I see we have some old comment sections so I suggest you add to those, so when someone comes to these files in a few years there is guidance on how to augment them.

I'm going to add that as a todo:

- [ ] Add link at top of Locales.txt to documentation page on the file format. Make one if it doesn't exist!

common/dtd/ldmlSupplemental.dtd
```diff
@@ -1225,6 +1225,25 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST coverageLevel match CDATA #REQUIRED >
     <!--@MATCH:any-->
 
+<!ELEMENT pathGroups ( pathGroup+ ) >
```
macchiati, line 1228: I think this can be simpler. All we need is an on-off switch.

srl295: It is an on off. Trying to colocate with coverageLevels since it is a similar dataset. Let's discuss tomorrow.

macchiati: What I mean is, we just need the match attribute. We don't need the other flags(inLanguage, inScript/) etc. 

All we need to map an ID to is a set of paths.

macchiati: Simplest would be a single new element:

<pathMatch id=**NMTOKENS** match=**pathMatch, like coverageLevel**/>

srl295:
@macchiati  I went with this format, to avoid issues with `id` being non-distinguishing:

```xml
		<pathMatches id="characterLabel">
			<pathMatch match="characterLabels/characterLabelPattern[@type='%anyAttribute']"/>
			<pathMatch match="characterLabels/characterLabelPattern[@type='%anyAttribute'][@count='%allPlurals']"/>
		</pathMatches>
		<pathMatches id="annotations">
			<pathMatch match="annotations/annotation[@cp='%modernEmoji'][@type='%anyAttribute']"/>
			<pathMatch match="annotations/annotation[@cp='%modernEmoji']"/>
		</pathMatches>
```

srl295:
i also have a parser for Locales.txt (split out)

srl295:
@macchiati  it's now:

```xml
		<pathMatch id="characterLabel" match="characterLabels/characterLabelPattern[@type='%anyAttribute']"/>
		<pathMatch id="characterLabel" match="characterLabels/characterLabelPattern[@type='%anyAttribute'][@count='%allPlurals']"/>
		<pathMatch id="annotations" match="annotations/annotation[@cp='%modernEmoji'][@type='%anyAttribute']"/>
		<pathMatch id="annotations" match="annotations/annotation[@cp='%modernEmoji']"/>
```

srl295:
@macchiati actually it has to be:

```xml
		<pathMatch id="characterLabel1" match="characterLabels/characterLabelPattern[@type='%anyAttribute']"/>
		<pathMatch id="characterLabel2" match="characterLabels/characterLabelPattern[@type='%anyAttribute'][@count='%allPlurals']"/>
		<pathMatch id="annotations1" match="annotations/annotation[@cp='%modernEmoji'][@type='%anyAttribute']"/>
		<pathMatch id="annotations2" match="annotations/annotation[@cp='%modernEmoji']"/>
```

because id is distinguishing, and then

```
iaap ; @weight=4 @pathMatch=annotations1,anotations2,characterLabel1,characterLabel2 * ; moderate ;
```

for Locales.txt.

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -553,117 +521,18 @@ public Level getTargetCoverageLevel(String localeId) {
...
-        platform_locale_levelString = new TreeMap<>();
-        platform_level_locale = new EnumMap<>(Organization.class);
-        for (Organization platform : platform_locale_level.keySet()) {
-            Map<String, String> locale_levelString = new TreeMap<>();
-            platform_locale_levelString.put(platform.toString(), locale_levelString);
-            Map<String, Level> locale_level = platform_locale_level.get(platform);
-            for (String locale : locale_level.keySet()) {
-                locale_levelString.put(locale, locale_level.get(locale).toString());
+    private LocalesTxtReader loadPlatformLocaleStatus() {
+        if (localesTxt == null) {
```
macchiati, line 527: This can fail in a multi-threading situation. The more reliable and cleaner way do do a lazy load is to have LocalesTxtReader method that returns a private static SINGLETON.

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
```diff
@@ -0,0 +1,205 @@
...
+                }
+
+                if (locales.size() != 1) {
+                    // require there to be exactly one locale.
+                    // This would allow collapsing into fewer lines.
+                    throw new IllegalArgumentException("Expected one locale entry in Locales.txt but got " + locales.size() + ": " + line);
+                }
+
+                // extract the single locale, process as before
+                String locale = locales.toArray(new String[1])[0];
```
macchiati, line 113: This is ugly. use ocales.iterator().next()

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
```diff
@@ -0,0 +1,205 @@
...
+
+                    // verify that the locale is not a default content locale
+                    if (defaultContentLocales.contains(locale)) {
+                        throw new IllegalArgumentException("Cannot have default content locale in Locales.txt: " + line);
+                    }
+                }
+
+                Level status = Level.get(stuff.get(2));
+                if (status == Level.UNDETERMINED) {
+                    System.out.println("Warning: Level unknown on: " + line);
```
macchiati, line 131: Let's make this an exception and fail. We don't want stuff be be screwed in up production.

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
```diff
@@ -0,0 +1,205 @@
...
+                language = parser.getLanguageScript();
+                if (!language.equals(oldLanguage)) {
+                    Level languageLevel = locale_level.get(language);
+                    if (languageLevel == null || languageLevel.compareTo(childLevel) < 0) {
+                        locale_level.put(language, childLevel);
+                    }
+                }
+            }
+        }
+        // backwards compat hack
```
macchiati, line 186: Can you explain what is going on here, and why we need this?

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
srl295: I will see if I can figure it out, but this function was moved from StandardCodes.java

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
srl295: Sounds like a good improvement. Note that this function was moved from StandardCodes.java

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
srl295: I'm not sure this specific one would fail, it might load multiple times.  In any event, StandardCodes is also needed for parsing Locales.txt (circular dependency).  I've fixed this so that the singleton uses StandardCodes.make() to read Locales.txt

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
macchiati: Ok, if it is just copied that's fine for now.

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
```diff
@@ -0,0 +1,205 @@
...
+public class LocalesTxtReader {
+    Map<Organization, Map<String, Level>> platform_locale_level = null;
+    Map<Organization, Relation<Level, String>> platform_level_locale = null;
+    Map<String, Map<String, String>> platform_locale_levelString = null;
+    Map<Organization, Map<String, Integer>> organization_locale_weight = null;
+    Map<Organization, Map<String, Set<String>>> organization_locale_match = null;
+
+    public static final String DEFAULT_NAME = "Locales.txt";
+
+    public LocalesTxtReader() {
```
macchiati, line 26: does this need to be public?

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * Read from Locales.txt, from the default location
+     * @param lstreg stream to read from
+     */
+    public LocalesTxtReader loadPlatformLocaleStatus(StandardCodes sc) {
```
macchiati, line 33: I don't see what is going on here. It looks this will be performed multiple times.

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
srl295: filed subtask CLDR-15625 for straightening this out

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
srl295: No, only once. But I'll rename the function in this class to `read()` to make it clearer.

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalesTxtReader.java
srl295: fixed

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
macchiati: Double-check loading is notoriously tricky.

Easiest and safest paradigm to use is the https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom

```
public class Something {
    private Something() {}

    private static class LazyHolder {
        static final Something INSTANCE = new Something();
    }

    public static Something getInstance() {
        return LazyHolder.INSTANCE;
    }
}
```

(For more on why it is tricky, you can later read https://en.wikipedia.org/wiki/Double-checked_locking.)

srl295: that's what is in the PR now

macchiati (approved):

PR #1919, opened 2022-04-14
CLDR-15381 Java formatting with prettier, tabWidth 4, printWidth 120

btangmu:
-Add .prettierrc.json with custom settings for .java files

CLDR-15381


tools/cldr-code/src/main/java/org/unicode/cldr/util/Annotations.java
```diff
@@ -679,25 +699,46 @@ private static void writeList() {
...
+            "🚴‍♂️",
+            "🚴🏿‍♂️",
+            "🚴‍♀️",
+            "🚴🏿‍♀️"
+        )) {
             final String shortName = eng.getShortName(s);
             final Set<String> keywords = eng.getKeywords(s);
-            System.out.println("{\"" + s + "\",\"" + shortName + "\",\"" + Joiner.on("|")
-                .join(keywords) + "\"},");
+            System.out.println("{\"" + s + "\",\"" + shortName + "\",\"" + Joiner.on("|").join(keywords) + "\"},");
```
srl295, line 741: … but generally, it's not mangling these lines anymore

srl295 (commented):
generally LGTM…

tools/cldr-code/src/main/java/org/unicode/cldr/util/CharUtilities.java
```diff
@@ -102,17 +103,44 @@ public static int compare(CharSource text1, CharSource text2) {
...
             }
         }
     }
 
     private static final char utf16Fixup[] = {
-        0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0x2000, 0xf800, 0xf800, 0xf800, 0xf800
+        0,
```
macchiati, line 112: uggg. Any way to manually override (eg with comments) to prevent mangling? We shouldn't have to make code changes just because of a pretty printer!

srl295: > If I use // @formatter:off and // @formatter:on, it works well.

per https://github.com/jhipster/prettier-java/issues/505#issue-1007366698

srl295:
@macchiati  @btangmu  prettier is explicitly an opinionated formatter.

per https://prettier.io/docs/en/option-philosophy.html

> Please note that as option requests are out of scope for Prettier, they will be closed without discussion

tools/cldr-code/src/main/java/org/unicode/cldr/util/CharUtilities.java
btangmu: "// @Formatter:off and // @Formatter:on" -- that looks promising

maybe that wouldn't be needed very often; the utf16Fixup array is unusual due to the performance optimization

btangmu: actually "@Formatter" doesn't work for me, but "// prettier-ignore" does, for java, although the documentation is for javascript:

https://prettier.io/docs/en/ignore.html#javascript

PR #1927, opened 2022-04-18
CLDR-15558 DtdData err msg improvements

srl295:
- indicate which DTD type failed to read
- indicate whether an illegal annotation was on an ELEMENT or an ATTLIST (i.e. @MATCH on an ELEMENT!)
- indicate the name of the element / attribute

CLDR-15558

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
@@ -221,7 +221,6 @@ public class DtdData extends XMLFileReader.SimpleHandler {
 
         public void addComment(String commentIn) {
             if (commentIn.startsWith("@")) {
-                // there are exactly 2 cases: deprecated and ordered
                 switch (commentIn) {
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -726,7 +738,9 @@ private static DtdData getInstance(DtdType type, String version, File directory)
             }
         }
         if (simpleHandler.ROOT.children.size() == 0) {
-            throw new IllegalArgumentException(); // should never happen
+            throw new IllegalArgumentException("Internal Error: DtdData.getInstance(" +
+                type + ", ...): readFile() failed to return any children!");
+            // should never happen
```
srl295, line 743: with user error, many such things happen that shouldn't happen :)

pedberg-icu (approved):
Thanks, this will be very helpful in debugging dtd additions!

macchiati (approved):
These all look good.

PR #1932, opened 2022-04-20
CLDR-14956 ht should fallback to French fr_HT

AEApple:
CLDR-14956

Haitian Creole language `ht` should fallback to French `fr-HT`

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -5372,6 +5372,7 @@ XXX Code for transations where no currency is involved
 		<parentLocale parent="en_150" locales="en_AT en_BE en_CH en_DE en_DK en_FI en_NL en_SE en_SI"/>
 		<parentLocale parent="en_IN" locales="hi_Latn"/>
 		<parentLocale parent="es_419" locales="es_AR es_BO es_BR es_BZ es_CL es_CO es_CR es_CU es_DO es_EC es_GT es_HN es_MX es_NI es_PA es_PE es_PR es_PY es_SV es_US es_UY es_VE"/>
+		<parentLocale parent="fr_HT" locales="ht"/>
 		<parentLocale parent="no" locales="nb nn"/>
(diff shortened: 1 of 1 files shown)


macchiati:
LGTM

AEApple:
I don't understand why it is failing the Jira ticket ID check.

srl295 (dismissed):

btangmu:
> I don't understand why it is failing the Jira ticket ID check.

That's not clear to me either -- maybe squashing will fix it, I'll try...

btangmu:
there were two spaces between "CLDR-14956" and "ht ..." -- not sure if that was a factor -- I clicked on Details for both the red-x checks, but there are still two red x's

AEApple:
> there were two spaces between "[CLDR-14956](https://unicode-org.atlassian.net/browse/CLDR-14956)" and "ht ..." -- not sure if that was a factor -- I clicked on Details for both the red-x checks, but there are still two red x's

Ah I had thought originally it was due to having an extra `-` so I switched to an underscore last night, but that didn't fix it.

btangmu:
i also had to delete the second space from the pull request title (not just the commit message)

AEApple:
> i also had to delete the second space from the pull request title (not just the commit message)

Why is the space affecting the ability of it to read the Jira ticket id?

macchiati:
There is a special check to make sure that the commit starts with the jira
ticket, and it wants to do an exact match for the start of the string.
However, I don't know why an additional space bothers it.

srl295:
> There is a special check to make sure that the commit starts with the jira
> ticket, and it wants to do an exact match for the start of the string.
> However, I don't know why an additional space bothers it


It's purposefully strict so that when some future or present person or process has to decode it it is simpler.

There are already several different apps that read the data. Commit checker, pr checker, etc.  And hopefully we won't ever move off git but you never know. 

There's already a ticket for making the message more helpful. I have not learned how to run the pr checker locally. Yet. But maybe I can discuss priorities with Shane sometime.

pedberg-icu (approved):

srl295 (approved):

PR #1934, opened 2022-04-20
CLDR-15561 read git hash directly for data

srl295:
- if possible, read directly from .git/HEAD and related files
- due to a recent git update, 'git rev-parse HEAD' may refuse to operate,
details at https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory
- if we do use git rev-parse HEAD, be clearer when there is an error

CLDR-15561

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
@@ -43,6 +43,7 @@ import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
```diff
@@ -1573,6 +1574,37 @@ public static String getCldrBaseDirHash() {
...
      */
     public final static String getGitHashForDir(String dir) {
+        // First, try just reading .git/HEAD
+        final File parent = new File(dir);
+        final File gitDir = new File(dir, ".git");
+        final File headfile = new File(gitDir, "HEAD");
+        if (headfile.canRead()) {
+            // Try this first, fallback to git commands
+            try {
+                String s = Files.readString(headfile.toPath());
```
srl295, line 1584: new in jdk11

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
```diff
@@ -1573,6 +1574,37 @@ public static String getCldrBaseDirHash() {
...
+                    if(s.startsWith("ref: ")) {
+                        s = s.substring(5); // refs/heads/main
+                        final Path refPath = gitDir.toPath().resolve(s);
+                        if (refPath.startsWith(gitDir.toPath())) {
+                            s = Files.readString(refPath);
+                            if (s != null && !s.isBlank()) {
+                                return s.trim();
+                            }
+                        } else { // ignore something like refs: ../../../yourfiles
+                            System.err.println("Ignoring strange git refPath " + refPath);
```
srl295, line 1596: this will avoid reading a random file outside of the .git dir

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
```diff
@@ -1573,6 +1574,37 @@ public static String getCldrBaseDirHash() {
...
+                        }
+                    } // else, maybe detached head
+                    return s.trim();
```
srl295, line 1599: sometimes `.git/HEAD` is just a hash, if so then use it

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
```diff
@@ -1573,6 +1574,37 @@ public static String getCldrBaseDirHash() {
...
+                }
+            } catch (IOException e) {
+                System.err.println(e + ": readString failed for " + headfile);
+                e.printStackTrace();
+            }
+
+        }
+
```
btangmu, line 1607: this method is too long, around 70 lines; most of what's above here should be in one subroutine, and most of what's below here should be in another subroutine

btangmu (approved):
needs subroutines and formatting, otherwise great

srl295:
image
also, confirmed that this works even if a different user owns the .git directory.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrUtility.java
srl295: rewrote as separate subroutines

btangmu (approved):

PR #1936, opened 2022-04-21
CLDR-15567 fixes for old DTDs

srl295:
CLDR-15567 fixes for DTD and chart generation

- ToolConstants: add 1.1, 1.4, 1.5.0.1, 41 versions
( [1.1](https://github.com/unicode-org/cldr/releases/tag/release-1-1) is newly remanufactured)

- DtdData: map 1.1.1, 1.4.1, 1.5.1 to prior DTDs (those versions did
not have a shipped DTD.)

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartDtdDelta.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartDtdDelta.java
@@ -109,6 +109,7 @@ public class ChartDtdDelta extends Chart {
             for (DtdType type : TYPES) {
                 String firstVersion = type.firstVersion; // FIRST_VERSION.get(type);
                 if (firstVersion != null && current != null && current.compareTo(firstVersion) < 0) {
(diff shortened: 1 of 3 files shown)


srl295:
@macchiati  this gets further, but I think there's a logic bug in ChartDtdDelta.  It's checking `current` for when a DTD first was released, but I think it needs to check `last` otherwise the compare fails.

It is trying to compare ldmlBCP47.dtd v1.6.1 to v1.7.2 ,when 1.7.2 was the first version for ldmlBCP47.

```
Caused by: java.io.FileNotFoundException: /Users/srl295/src/cldr-archive/cldr-1.6.1/common/dtd/ldmlBCP47.dtd (No such file or directory)
```

srl295:
With this fix, including rerunning `CheckoutArchive`, I am able to complete `GenerateAllCharts` with no errors.

I suspect that the DTD chart may be missing some versions before?

macchiati (approved):
Thanks!

macchiati:
Thanks, I'll check that out.

PR #1937, opened 2022-04-21
CLDR-13843 Change the About panel github links

btangmu:
-For the About panel, links to commit, compare release, and compare main

-Construct links in js instead of java; configurable in cldrText.js

-Most of the changes are to fix bugs that blocked this minor GUI change

-Do not assume cldr-apps.war has a hash, since liberty:package drops it

-Two hashes: Data and Code, where Code refers to cldr-code.jar

-For cldr-code, use the name Code consistently, not (Java) Tools, UTILITIES, etc.

-Encapsulate string literals with new constants such as CldrUtility.CODE_SLUG

-Remove deprecated and dead code such as getCurrevCldrApps, setCldrAppsHash

-Fix some minor compiler warnings

-Comments

CLDR-13843

### DIFF
--- a/tools/cldr-apps/js/src/views/AboutPanel.vue
+++ b/tools/cldr-apps/js/src/views/AboutPanel.vue
@@ -23,15 +23,22 @@
               <a class="aboutValue" v-bind:href="v">{{ v }}</a>
             </span>
             <span v-else-if="valueIsHash(v, k)">
-              <!-- Looks like a Git hash, 6…40 hex chars.
-                   We need to check that CLDR_COMMIT_BASE is set, because that’s the
-                   base URL for the linkification -->
-              <a
-                class="aboutValue"
-                v-bind:href="aboutData.CLDR_COMMIT_BASE + v"
-              >
+              <!-- Looks like a Git hash, 6…40 hex chars. -->
+              <a class="aboutValue" v-bind:href="commitPrefix + v">
                 <i class="glyphicon glyphicon-cog" />{{ v }}
               </a>
+              |
+              <a
+                class="aboutValue"
+                v-bind:href="comparePrefix + compareRelease + '...' + v"
+                >compare to release</a
+              >
+              |
+              <a
+                class="aboutValue"
+                v-bind:href="comparePrefix + v + '...' + compareMain"
+                >compare to main</a
+              >
             </span>
             <span v-else class="aboutValue">
               <!-- some other type -->
@@ -45,14 +52,25 @@
 </template>
 
 <script>
+import * as cldrText from "../esm/cldrText.js";
+
 export default {
   data() {
     return {
       aboutData: null,
+      commitPrefix: null,
+      comparePrefix: null,
+      compareMain: null,
+      compareRelease: null,
     };
   },
 
   created() {
+    this.commitPrefix = cldrText.get("git_commit_url_prefix");
+    this.comparePrefix = cldrText.get("git_compare_url_prefix");
+    this.compareMain = cldrText.get("git_compare_url_main");
+    this.compareRelease = cldrText.get("git_compare_url_release");
+
     fetch("api/about")
       .then((r) => r.json())
       .then((data) => (this.aboutData = data));
@@ -60,11 +78,7 @@ export default {
 
   methods: {
     valueIsHash(value, key) {
-      if (
-        this.aboutData.CLDR_COMMIT_BASE &&
-        key.includes("HASH") &&
-        /^[a-fA-F0-9]{6,40}$/.test(value)
-      ) {
+      if (key.includes("HASH") && /^[a-fA-F0-9]{6,40}$/.test(value)) {
         return true;
       }
       return false;
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
@@ -35,7 +35,11 @@ import org.unicode.cldr.web.WebContext;
  * rather than environment variables.
  */
 public class CLDRConfigImpl extends CLDRConfig implements JSONString {
+    private static final String CODE_HASH_KEY = "CLDR_CODE_HASH";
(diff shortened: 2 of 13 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -3344,7 +3312,7 @@ private static void busted(String what, Throwable t, String stack) {
         if (t != null) {
             SurveyLog.logException(logger, t, what /* , ignore stack - fetched from exception */);
         }
-        logger.warning("SurveyTool " + SurveyMain.getCurrevCldrApps() + " busted: " + what + " ( after " + pages + "html+" + xpages
```
srl295, line 3347: Could fallback to utility or data hash here. This message might be in the logs out of context, and would provide information on which rev failed. 

Also, I would expect that cldr-apps would get a hash back at some point, so I'm not sure about ripping out support for a cldr-apps hash.

tools/cldr-apps/src/test/java/org/unicode/cldr/unittest/web/TestMisc.java
```diff
@@ -96,18 +91,13 @@ public void TestGitHash() {
                 errln("❌ getCldrBaseDirHash is shorter than 8 chars: " + hash);
             }
         }
-        final String strings[] = {"CLDR-Apps", "CLDR-Tools"};
```
srl295, line 99: again, i dont' think ther'es a reason to remove this functionality entirely.

tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
```diff
@@ -200,14 +204,14 @@ private synchronized void init() {
...
-
-        survprops.put("CLDRHOME", cldrHome);
+        // codeHash = the git version of the cldr-code.jar file embedded in cldr-apps.
+        String codeHash = getGitHashForSlug(CldrUtility.CODE_SLUG);
+        // dataHash = the git version of the CLDR_DIR currently in use.
+        String dir = survprops.getProperty(CldrUtility.DIR_KEY, null);
+        String dataHash = CldrUtility.getGitHashForDir(dir);
+        survprops.put(CODE_HASH_KEY, codeHash);
+        survprops.put(DATA_HASH_KEY, dataHash);
+        survprops.put(CldrUtility.HOME_KEY, cldrHome);
```
srl295, line 211: Should leave support for a cldr-apps hash as well.

tools/cldr-apps/js/src/views/AboutPanel.vue
```diff
@@ -23,12 +23,10 @@
...
             </span>
             <span v-else-if="valueIsHash(v, k)">
-              <!-- Looks like a Git hash, 6…40 hex chars.
-                   We need to check that CLDR_COMMIT_BASE is set, because that’s the
-                   base URL for the linkification -->
+              <!-- Looks like a Git hash, 6…40 hex chars. -->
               <a
                 class="aboutValue"
-                v-bind:href="aboutData.CLDR_COMMIT_BASE + v"
+                v-bind:href="gitUrlPrefix + v + gitUrlSuffix"
```
srl295, line 29: I actually don't like this change, because it's only correct in some situations, as I noted on the original ticket. For example, if an alternate branch or back level version is deployed, the "..main" diff may not return much.

What I would propose (and what I proposed there) is to have at least two links:
- a link directly to the commit (under the commit hash)
- a link with the comparison to "..main" (as 'vs. main' or some such)

I use this link often, and a link directly to the commit is what is most useful.

srl295 (changes requested):
First, +1 to the bulk of the changes. Great cleanup.

I would request to leave the original commit link there, adding an additional diff (..main) link. Knowing that that diff link may not always be useful.

I'd prefer leaving cldr-apps hash in, but fine if that should wait until that is sorted out. The cldr-apps war ought to be self identifying, so I consider it a configuration bug that it is currently not.

tools/cldr-apps/js/src/views/AboutPanel.vue
btangmu: good idea -- it's trivial to include both links in the about panel

srl295: in https://unicode-org.atlassian.net/browse/CLDR-13843?focusedCommentId=158448 @macchiati  also asked for the 'changes since last released version'.  Not sure how you'd calculate that, but could use the OLDVERSION field from cldr.properties.

srl295: `gitUrlPrefix + 'release-' + '41'  + '...' + v` would give all changes since v41  [example](https://github.com/unicode-org/cldr/compare/release-41...91c8dcdcce09065085c23a471ed1736f00aded35)

srl295: OLDVERSION with `.` changed to `-`

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
btangmu: I'll make it show the cldr-code hash here

tools/cldr-apps/src/test/java/org/unicode/cldr/unittest/web/TestMisc.java
btangmu: we shouldn't keep dead/broken code around just based on wishful thinking -- there's too much of that already -- so it should be decided whether we're going to implement a cldr-apps war hash during v42 -- the return on investment seems dubious

as a compromise, we could have a brief comment indicating that a cldr-apps war hash could be implemented similar to the cldr-code jar hash

srl295: It's already out so I'm fine to keep it so. It's more of a bug than wishful thinking though!

srl295 (approved):
It should use the variable instead of a hard coded 41, but that can be a follow up

PR #1943, opened 2022-04-23
CLDR-15404 fix for monogram after swapping to usage

macchiati:
CLDR-15404

1. Restore the special handling for monograms (now simpler to test for).
2. Fix the tests, but also replace the numbers as suggested in a review of the translators guide.


macchiati:
Note: after doing this, I had the thought that it might be clearer to use super/subscripts in the tests (I also copy those examples to the translator's guide). eg

⓵{prefix}𝟙 ⓶{given}𝟚 ... =>
¹{prefix}₁ ²{given}₂ ...

What do you think?

richgillam:
I like the superscript/subscript idea; I think it looks a lot better.

Also, did you see my comment on your last PR about how we already had a test for skipping literals?  If we can't consolidate my old test with your new ones, can we at least move them to be closer together (and maybe named more similarly?

richgillam (approved):

PR #1945, opened 2022-04-25
CLDR-15369 Priority Items Summary snapshots, more improvements

btangmu:
-Enforce a max duration for auto-snapshot to run: AUTO_SNAP_MAX_DURATION_MINUTES

-Provide constants for time of day including minute: AUTO_SNAP_MINUTE_OF_HOUR

-Set percent to 100 (not 0) when complete successfully

-Refactor with subroutines

-Comments

CLDR-15369

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
@@ -175,6 +175,7 @@ public class VettingViewerQueue {
 
             @Override
             public void done() {
+                // note: this method is possibly never called
                 progress.update("Done!");
             }
 
@@ -353,6 +354,7 @@ public class VettingViewerQueue {
         if (args.loadingPolicy != LoadingPolicy.FORCESTOP) {
             VVOutput res = entry.output.get(args.usersOrg);
             if (res != null) {
+                setPercent(100);
                 results.status = Status.READY;
                 results.output.append(res.output);
                 stop(entry);
@@ -434,7 +436,6 @@ public class VettingViewerQueue {
     }
 
     private void stop(QueueEntry entry) {
-        setPercent(0);
         Task t = entry.currentTask;
         if (t != null) {
             if (t.myThread.isAlive() && !t.stop) {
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
```diff
@@ -175,6 +175,7 @@ public void nudge() {
 
             @Override
             public void done() {
+                // note: this method is possibly never called
```
srl295, line 178: why isn't it?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Summary.java
```diff
@@ -347,6 +356,25 @@ private void makeAutoPriorityItemsSnapshot() throws IOException, JSONException {
         log("Automatic Summary Snapshot, finished; status = " + sr.status);
     }
 
+    private boolean autoSnapTooLong(long startMillis) {
+        final long elapsedMillis = System.currentTimeMillis() - startMillis;
+        final long elapsedMinutes = TimeUnit.MINUTES.convert(elapsedMillis, TimeUnit.MILLISECONDS);
+        if (elapsedMinutes > AUTO_SNAP_MAX_DURATION_MINUTES) {
+            log("Automatic Summary Snapshot timed out, stopping");
```
srl295, line 363: should have some way to log this… maybe SurveyLog.logException since those are tracked?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
btangmu: good question!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Summary.java
btangmu: it will show up in console.log -- what kind of "tracked" do you mean?

srl295: For later, the new `SurveyMetrics` class tracks number of exceptions thrown.  We could track metrics around how long snapshots take to run, and a timestamp of when the last snapshot was taken, with alerts if something is out of range

btangmu: it will show up in console.log -- what kind of "tracked" do you mean?

PR #1947, opened 2022-04-26
CLDR-14906 Retain Dashboard hidden status across releases

btangmu:
-For efficiency, skip unused getHidden and voterProgress for single-path dash

-New DashboardArgs.isOnlyForSinglePath

-Minor refactoring of parameters, subroutines, accessor methods

-Fix bug in setUserAndOrganization, parameter names vs field names

CLDR-14906

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
@@ -9,6 +9,7 @@ import org.unicode.cldr.util.*;
 import org.unicode.cldr.util.PathHeader.PageId;
 import org.unicode.cldr.util.PathHeader.SectionId;
 import org.unicode.cldr.util.VettingViewer.Choice;
+import org.unicode.cldr.util.VettingViewer.DashboardArgs;
 
 import com.ibm.icu.impl.Relation;
 import com.ibm.icu.impl.Row;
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
```diff
@@ -170,53 +174,54 @@ public ReviewEntry(String code, String xpath) {
...
          * are generally the last release values plus any changes that have been made by the
          * technical committee by committing directly to version control rather than voting.
          */
-        CLDRFile sourceFile = sourceFactory.make(loc);
-        CLDRFile baselineFile = baselineFactory.make(loc, true);
-        VettingViewer.DashboardArgs args = new VettingViewer.DashboardArgs(choiceSet, locale, coverageLevel);
-        args.setUserAndOrganization(user.id, usersOrg);
+        final String localeId = locale.getBaseName();
+        final CLDRFile sourceFile = sourceFactory.make(localeId);
+        final Factory baselineFactory = CookieSession.sm.getDiskFactory();
```
srl295, line 201: for testability, there should be another overload of ReviewEntry() where the baselineFactory and stFactory are passed as parameters, instead of going through the static CookieSession.sm

srl295 (approved):
minor comments but definitely an improvement.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
btangmu: is there a ticket for that testability?

PR #1948, opened 2022-04-26
CLDR-15538 Expedite page refresh after user votes

btangmu:
-After user votes, when no more votes are pending responses, schedule update in 3 seconds

-New function cldrSurvey.expediteStatusUpdate

-New fastTimerSpeed = 3000 ms = 3 seconds

-Fix bug in updateDashboardRow, updateRow was renamed to updatePath

CLDR-15538

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.js
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.js
@@ -76,10 +76,11 @@ const statusActionTable = {
 /**
  * How often to fetch updates. Default 15s.
  * Used only for delay in calling updateStatus.
- * May (theoretically, if it were visible) be changed by js written by SurveyMain.showOfflinePage, etc.
  * @property timerSpeed
  */
 let timerSpeed = 15000; // 15 seconds
+let fastTimerSpeed = 3000; // 3 seconds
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/js/src/esm/cldrSurvey.js
```diff
@@ -562,7 +564,7 @@ function updateStatusLoadHandler(json) {
   if (wasBusted == false && json.status.isSetup && loadOnOk != null) {
     window.location.replace(loadOnOk);
   } else {
-    setTimeout(updateStatus, timerSpeed);
+    statusTimeout = setTimeout(updateStatus, timerSpeed);
```
srl295, line 567: should this also clear the old timeout? or is this only called the first time.

srl295 (approved):
lgtm w/ a question

tools/cldr-apps/js/src/esm/cldrSurvey.js
btangmu: this is in updateStatus, which only gets called in response to the timer (except the first time), and it sets

  statusTimeout = null;

right at the beginning -- I think clearTimeout would be redundant there since it's called in response to the timer itself

PR #1953, opened 2022-04-27
CLDR-14922 When user presses Agree in forum, vote automatically

btangmu:
-New cldrVoteAgree.js, accessed by cldrForum and cldrTable

-Call cldrVoteAgree.vote from cldrForum when user presses Agree

-Load the table to make sure it is visible and to trigger vote while loading

-Call cldrVoteAgree.voteIfAgreed from cldrTable when loading each row

-The back end already automatically makes Agree forum posts when votes occur

-Refactor cldrForum a bit to shorten methods, reduce nesting

CLDR-14922


tools/cldr-apps/js/src/esm/cldrForum.js
```diff
@@ -373,68 +374,73 @@ function submitPost(event) {
...
+      "<i>Your post was added, #" + data.postId + " but could not be shown.</i>"
+    );
+  }
+}
+
+function errorHandlerForSubmit(err) {
+  const post = $(".post").first();
+  post.before("<p class='warn'>error! " + err + "</p>");
+}
+
```
btangmu, line 443: above here, in this file it's just refactoring which I thought would be needed -- though it turned out not to be needed, it results in shorter methods and less nesting

tools/cldr-apps/js/src/esm/cldrForum.js
```diff
@@ -954,15 +961,19 @@ function makeOneReplyButton(post, postType, label) {
...
-      /*
-       * Don't specify locale/xpath/subject/value/open for reply. Instead they will be set to
-       * match the original post in the thread.
-       */
-      replyTo: post.id,
-      replyData: post,
-      postType: postType,
-    });
+    if (postType === "Agree") {
+      cldrVoteAgree.vote(post.locale, post.xpath, post.value);
```
btangmu, line 965: calling cldrVoteAgree.vote here is the essential change in this file

macchiati (commented):
Looks good to me, but Steven should review the semantics.

tools/cldr-apps/js/src/esm/cldrForum.js
srl295: That makes it look easy !

tools/cldr-apps/js/src/esm/cldrVoteAgree.js
```diff
@@ -0,0 +1,107 @@
...
+ */
+function vote(locale, xpstrid, value) {
+  rememberToVote[agreeKey(locale, xpstrid)] = value;
+  cldrStatus.setCurrentLocale(locale);
+  cldrStatus.setCurrentId(xpstrid);
+  cldrStatus.setCurrentSpecial("");
+  cldrStatus.setCurrentPage("");
+  // cldrLoad.reloadV will cause the page table to be loaded,
+  // and then cldrTable will call voteIfAgreed for each row
+  cldrLoad.reloadV();
```
srl295, line 54: seems like odd layering for a vote…

tools/cldr-apps/js/src/esm/cldrVoteAgree.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+  notifyNotFound(value);
+}
+
+function reallyVote(tr, theRow, value, valueHash) {
+  const pseudoButton = {
+    value: value,
+  };
+  cldrVote.handleWiredClick(tr, theRow, valueHash, undefined, pseudoButton);
```
srl295, line 86: this seems fragile…  why not just use the voting API?

POST `/api/voting/{locale}/row/{hexXpathHash}`

payload is: 

```json
{
  "value": "The Vote Value",
  "voteLevelChanged": 4
}
```

`voteLevelChanged` vote downgrade, can omit or null to use default.

srl295: and then mark the row for reload somehow

srl295 (changes requested):

tools/cldr-apps/js/src/esm/cldrVoteAgree.js
btangmu: the comments at top of cldrVoteAgree.js try to explain some of the rationale

a fundamental question is, for the new agree-vote feature, do we re-use the existing system or implement a separate parallel system?

the existing voting system keeps track of which rows have votes pending; how many votes are pending; distinctive display styles for rows depending on their vote-pending status; when an all-row response is received, avoidance of replacing newer single-row data with older all-row data for a particular row; how a row is displayed in response to a vote-submission failure; ...

a parallel system would risk screwing up the existing system -- is there a vote pending on this row? yes, but the existing system doesn't know about it, so the display doesn't match the reality

the existing system doesn't use api/voting -- that's not actually used at all, yet, is it? we keep postponing modernizing the page view since it's such a large task, involving lot of highly complex tightly coupled systems

when the page view is modernized we won't have dependencies on data in the DOM; in the meantime we do, so in this pr i've tried to adapt to the current situation with minimal complexity

btangmu: @srl295 thanks! i'll go ahead and merge this and follow up soon with another PR or two:
* add a dialog along the lines of the suggestion by @stenshamn "You indicated agreement with Vetter N, and your vote for value Z will be changed from X to Y. Please note that only the vote for Z will be changed. Other vote changes needed for consistency with this value will have to be made by you individually for those items."
* possibly refactor to reduce entanglement with legacy code

PR #1946, opened 2022-04-25
CLDR-14766 regression tests for UserRegistry.Level

srl295:
- first steps towards moving logic into VoteResolver
- should be a data driven approach, but this is an improvent

CLDR-14766

- [ ] This PR completes the ticket.

PR #1956, opened 2022-04-28
CLDR-14766 move permissions policy from UserRegistry to VoteResolver.Level

srl295:
CLDR-14766

👉  **Note** because of the way phases work, canSubmit returns false not only in the SurveyTool `READONLY` phase, but also `VETTING_CLOSED`.  This is because CheckCLDR.Phase only has `FINAL_TESTING` and does not have a readonly option. But voting during `VETTING_CLOSED` should not have been allowed before, either, and probably stopped by other code.

- [X] add some VoteResolver.Level specific tests
- [x] refactor UserRegistry to call into VoteResolver.Level (moving code there)
- [x] add a test that makes the same test directly into VoteResolver.Level

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
@@ -29,7 +29,6 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.json.JSONString;
 import org.unicode.cldr.test.CheckCLDR;
-import org.unicode.cldr.test.CheckCLDR.Phase;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRConfig.Environment;
 import org.unicode.cldr.util.CLDRInfo.UserInfo;
@@ -459,10 +458,7 @@ public class UserRegistry {
          * @param org
          */
         public boolean isAdminForOrg(String org) {
-            boolean adminOrRelevantTc = UserRegistry.userIsAdmin(this) ||
-
-                ((UserRegistry.userIsTC(this) || this.userlevel == MANAGER) && (org != null) && this.org.equals(org));
-            return adminOrRelevantTc;
+            return getLevel().isAdminForOrg(getOrganization(), Organization.fromString(org));
         }
 
         @Override
@@ -504,7 +500,7 @@ public class UserRegistry {
         }
 
         public boolean canImportOldVotes(CheckCLDR.Phase inPhase) {
-            return UserRegistry.userIsVetter(this) && (inPhase == Phase.SUBMISSION);
+            return getLevel().canImportOldVotes(inPhase);
         }
 
         @Schema( description="how much this user’s vote counts for")
@@ -1158,6 +1154,7 @@ public class UserRegistry {
         return msg;
     }
 
+    // TODO: should be refactored.
     public boolean canSetUserLevel(User me, User them, int newLevel) {
         final VoteResolver.Level myLevel = VoteResolver.Level.fromSTLevel(me.userlevel);
         if (!myLevel.canCreateOrSetLevelTo(VoteResolver.Level.fromSTLevel(newLevel))) {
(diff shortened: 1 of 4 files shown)


srl295:
I think this is ready to go, just pending PR #1946 being merged. 
Because of what this is, i'd like these to be separate commits.

srl295:
@macchiati  this one is the refactoring but depends on #1946 being merged first

srl295:
Maybe it's actually `tCORGiVETTER` 

[corgi](https://commons.wikimedia.org/wiki/File:14_year_old_Corgi.jpg)

macchiati:
🐶

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -262,7 +264,108 @@ public ImmutableSet<Integer> getVoteCountMenu() {
             admin.voteCountMenu = ImmutableSet.of(vetter.votes, admin.votes); /* Not LOCKING_VOTES; see canVoteWithCount */
             tc.voteCountMenu = ImmutableSet.of(vetter.votes, tc.votes, PERMANENT_VOTES);
         }
-    }
+
+        // The following methods were moved here from UserRegistry
+        // TODO: remove this todo notice
```
btangmu, line 269: "TODO: remove this todo notice"?!

macchiati:
🤔

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -262,7 +264,108 @@ public ImmutableSet<Integer> getVoteCountMenu() {
...
+        }
+
+        public boolean canSubmit(CheckCLDR.Phase inPhase) {
+            if (inPhase == Phase.FINAL_TESTING) {
+                return false;
+                // TODO: Note, this will mean not just READONLY, but VETTING_CLOSED will return false here.
+                // This is probably desired!
+            }
+            return isStreet();
+        }
```
btangmu, line 351: where the old code uses SurveyMain.Phase.READONLY, the new code uses CheckCLDR.Phase.FINAL_TESTING, so this is potentially a change in functionality

@srl295 as you wrote in the commit message, "voting during VETTING_CLOSED should not have been allowed before, either, and probably stopped by other code" -- I imagine you're right; to be cautious I'll ask at today's infrastructure meeting, could there have been an intentional loophole for Admin or TC to vote during VETTING_CLOSED

srl295: @btangmu thanks.

srl295: i would lean towards making it more restrictive unless we have a reason this should be open.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: can i remove this in a future PR?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: @srl295 at yesterday's infrastructure meeting it was expressed that ST may have VETTING_CLOSED during resolution, and that TC/Admin _need_ to be able to vote at that time

i'll study the question...

btangmu: locally, i've confirmed that TC can submit a new item during VETTING_CLOSED

btangmu: SurveyMain.Phase:
```
    public enum Phase {
        SUBMIT("Data Submission", CheckCLDR.Phase.SUBMISSION),
        VETTING("Vetting", CheckCLDR.Phase.VETTING),
        /*
         * VETTING_CLOSED means, "closed after vetting, while still open for admin tasks"
         */
        VETTING_CLOSED("Vetting Closed", CheckCLDR.Phase.FINAL_TESTING),
        CLOSED("Closed", CheckCLDR.Phase.FINAL_TESTING),
        DISPUTED("Dispute Resolution", CheckCLDR.Phase.VETTING),
        FINAL_TESTING("Final Testing", CheckCLDR.Phase.FINAL_TESTING),
        READONLY("Read-Only", CheckCLDR.Phase.FINAL_TESTING),
```

btangmu: CheckCldr.Phase:
```
    public enum Phase {
        BUILD, SUBMISSION, VETTING, FINAL_TESTING("RESOLUTION");

```

btangmu: it would be less confusing if there were only a single enum -- it could make all the fine distinctions while also providing some methods for treating some values as equivalent for some purposes

srl295: Yup there's an issue for that. 

Maybe this one function shouldn't be refactored yet ?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
```diff
@@ -1686,22 +1683,19 @@ static final boolean userCanDeleteUser(User managerUser, int targetId, int targe
...
-        return (userIsVetter(managerUser));
+        return managerUser.getLevel().canDoList();
     }
 
     public static final boolean userCanCreateUsers(User u) {
-        return (userIsTC(u) || userIsExactlyManager(u));
+        return u.getLevel().canCreateUsers();
     }
 
     static final boolean userCanSubmit(User u) {
```
btangmu, line 1693: This method -- with only one arg -- appears to be unused, in which case it should be removed

srl295: Could be. For near future ?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
```diff
@@ -1686,22 +1683,19 @@ static final boolean userCanDeleteUser(User managerUser, int targetId, int targe
...
         return userCanSubmit(u, SurveyMain.phase());
     }
 
     static final boolean userCanSubmit(User u, SurveyMain.Phase phase) {
-        if (phase == SurveyMain.Phase.READONLY) {
-            return false;
-        }
-        return ((u != null) && userIsStreet(u));
+        return u.getLevel().canSubmit(phase.getCPhase());
     }
```
btangmu, line 1699: This method -- with two args -- appears to be unused except by TestCompatUserLevelDataDriven

so this is all much ado about not very much? if so, we could simply remove this method as well, and remove the call to it in the test

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: see my note above -- userCanSubmit (both versions), and canSubmit, are all essentially dead code if i'm not mistaken

srl295:
A proposed way forward here(probably next week):

- [ ] remove todo todo
- [ ] somehow add override for tc to fix the vetting closed phase issue, add special test. 
- [ ] remove canCreateUsers and tests for it

btangmu (approved):
since userCanSubmit is essentially dead code, this looks harmless

a follow-up PR should remove the dead code

btangmu:
> * somehow add override for tc to fix the vetting closed phase issue

not necessary? my guess is this PR will not prevent TC from voting when VETTING_CLOSED -- i approved the pr, if you want to merge then we can confirm that on smoketest

macchiati:
> locally, i've confirmed that TC can submit a new item during
VETTING_CLOSED
Good, thanks.

PR #1957, opened 2022-04-28
CLDR-15475 levels: vetters=6 for TC, bump tc=50

srl295:
CLDR-15475 levels: vetter=6 for tcorg, bump tc=50

- 'vetter' has 6 votes for tc, 4 otherwise
- new Organization.isTCOrg() function
- 'tc' now has 50 votes
- remove UserRegistry.ALL_LEVELS, obsolete and hard to maintain- Some code changes to call sites for this.
- TestUserLevel test and data updates
- changes to required votes, from 20 to 50
- update supplemental to support `<approvalRequirement votes="=tc" .../>` with a symbolic vote number
- documentation for above
- unit test updates supporting vote requirement changes


srl295:
would like some clarity on the `tcorgvetter` capabIlities - you can review the `tcorgvetter` lines in the .csv file in this PR.

- does the `tcorgvetter` act like a manager? or like tc? or like a vetter? I am assuming manager. 
- are 6 votes correct for `tcorgvetter`?
- [ ] I didn't change this yet, but should TC then have 50 votes as the ticket mentions ?Not clear what the consensus is. Thanks

macchiati:
tcOrgVetter is precisely like vetter except with higher vote.

Not like manager.

Tc (us) bumped to 50 is right

srl295:
> tcOrgVetter is precisely like vetter except with higher vote.
> 
> Not like manager.

OK. It's a vetter in a ( tc org ).  It's not a TC (Org Vetter).   I parsed it wrongly.

> Tc (us) bumped to 50 is right.

will do. thanks

srl295:
@macchiati actually, it's basically "expert" but 6 instead of 8 votes.

srl295:
@macchiati  if you review the 2nd commit (ignoring the 1st) the simple change to VoteResolver is clear, as is the csv changes.

srl295:
as expected there are some other test failures. I'll work on those.

srl295:
Maybe it's actually `tCORGiVETTER` 

[corgi](https://commons.wikimedia.org/wiki/File:14_year_old_Corgi.jpg)

tests should pass now.

macchiati:
One other thing. 
No non-tc org should be able to have tcOrgVetters

Question. Is it easy for org admins to move all vetters to tcOrgVetter status?

srl295:
> One other thing. No non-tc org should be able to have tcOrgVetters

OK. That will take some more logic.

Should a tc org have non-tcOrgVetters? 

> Question. Is it easy for org admins to move all vetters to tcOrgVetter status?

all vetters? no there's not a bulk move. 

But that brings up a related question.  We had talked about a situation where "in scope" locales get one vote, but "out of scope" locales get a lower vote such as 1 vote.  Going back to CLDR-15475, Is a separate level the best way to handle this, or should it really be something rule based? where "vetter" means 4 votes normally, but "6" for certainl locales in certain orgs,  and "1" for other locales and other orgs?

AEApple:
Yes, for most languages TC orgs should have regular TC vetters, TC org vetters are for times when there is only one TC org that is able to support a language that cycle.

srl295:
I think for now I could put some logic in on 'create user' and 'update user' to prevent non-tc-org from creating a tcorguser. I'll look into that.

(also a reminder please review #1956 so this can be unblocked)

srl295:
ok. @macchiati  would it work to just say that only `tc` level users can create `tcorgvetters` ? that would be a lot simpler to enforce.

srl295:
OK PTAL.  Now only TC (or admin)  can create tcorgvetter accts.

srl295:
@macchiati  @btangmu  rebased…

srl295:
@btangmu @macchiati  rebased, but this does not include "Instead of a separate Level.TcOrgVetter, should we have all the vetters in a TC org get a higher vote?"'

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -2787,7 +2787,12 @@ public String toString() {
         ApprovalRequirementMatcher(String xpath) {
             XPathParts parts = XPathParts.getFrozenInstance(xpath);
             if (parts.containsElement("approvalRequirement")) {
-                requiredVotes = Integer.parseInt(parts.getAttributeValue(-1, "votes"));
+                final String votesStr = parts.getAttributeValue(-1, "votes");
+                if (votesStr.charAt(0) == '=') {
+                    requiredVotes = VoteResolver.Level.valueOf(votesStr.substring(1)).getVotes();
+                } else {
+                    requiredVotes = Integer.parseInt(votesStr);
+                }
```
btangmu, line 2795: this would make a good subroutine: requiredVotes = getRequiredVotes(parts)

btangmu (dismissed):
great!

srl295:
image
^ working locally

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
```diff
@@ -230,7 +230,7 @@ private static VoteEntry[] calculateVotes(Set<User> votes, Map<User, Integer> ov
             if (overrides != null) {
                 override = overrides.get(u);
             }
-            entries.add(new VoteEntry(u.id, u.getLevel().getVotes(), override));
+            entries.add(new VoteEntry(u.id, u.getLevel().getVotes(u.getOrganization()), override));
```
btangmu, line 233: u.getLevel().getVotes(u.getOrganization())
could be shortened to
u.getVoteCount()

docs/ldml/tr35-info.md
```diff
@@ -573,16 +573,18 @@ In this example, the coverge variables %dayTypes and %wideAbbr are used to subst
...
+
 Here is an example of the approvalRequirements section.
 
 ```xml
 <approvalRequirements>
     <!--  "high bar" items -->
-    <approvalRequirement votes="20" locales="*" paths="//ldml/numbers/symbols[^/]++/(decimal|group)"/>
+    <approvalRequirement votes="=tc" locales="*" paths="//ldml/numbers/symbols[^/]++/(decimal|group)"/>
     <!--  established locales - https://cldr.unicode.org/index/process#h.rm00w9v03ia8 -->
     <approvalRequirement votes="8" locales="ar ca cs da de el es fi fr he hi hr hu it ja ko nb nl pl pt pt_PT ro ru sk sl sr sv th tr uk vi zh zh_Hant" paths=""/>
```
macchiati, line 585: For later, we probably want this to be a variable also.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Organization.java
```diff
@@ -56,10 +57,24 @@ public enum Organization {
...
-
-    // To be removed.
     ;
 
+    /**
+     * Get a list of the TC Organizations
+     * @return
+     */
+    public static Set<Organization> getTCOrgs() {
+        return EnumSet.of(google, apple, microsoft);
```
macchiati, line 67: Make a private final static TC_ORGS = ImmutableSet.copyOf(EnumSet.of(google, apple, microsoft));

No reason to recreate a mutable set each time this is called.

macchiati (dismissed):

btangmu:
eventually maybe Level could move out of VoteResolver, and rather than just an enum, a class could encapsulate the increasingly complex relationships between level, vote count, org, and user -- then ideally the other classes wouldn't need to know, for example, whether/when the vote count depends on the org...

docs/ldml/tr35-info.md
srl295: yes. Need to choose a good name, maybe =HIGH_BAR or similar (caps, so not a level)

srl295:
> eventually maybe Level could move out of VoteResolver, and rather than just an enum, a class could encapsulate the increasingly complex relationships between level, vote count, org, and user -- then ideally the other classes wouldn't need to know, for example, whether/when the vote count depends on the org...

I'm not sure how that specific move would help. Per this ticket, vote count always depends on level+org and not on user.

I think what might help some is to remove STLevel and basically inline all calls that go through UserRegistry.User.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Organization.java
srl295: I think I did this originally because i forgot to make TC_ORGS static and this wasn't allowed. Fixed now.

common/supplemental/coverageLevels.xml
```diff
@@ -14,31 +14,31 @@ For terms of use, see http://www.unicode.org/copyright.html
...
-			<approvalRequirement votes="20" locales="*" paths="//ldml/numbers/(defaultNumberingSystem|otherNumberingSystems.*)"/>
-			<approvalRequirement votes="20" locales="*" paths="//ldml/characters/exemplarCharacters.*"/>
-			<approvalRequirement votes="20" locales="*" paths="//ldml/characters/parseLenients.*"/>
-			<approvalRequirement votes="20" locales="*" paths="//ldml/numbers/minimalPairs/pluralMinimalPairs.*"/>
-			<approvalRequirement votes="20" locales="*" paths="//ldml/numbers/minimumGroupingDigits"/>
-			<approvalRequirement votes="20" locales="*" paths="//ldml/numbers/symbols[^/]++/timeSeparator"/>
-			<approvalRequirement votes="20" locales="*" paths="//ldml/numbers/currencies/currency\[@type=.([A-Z]{3}).\]/symbol(\[@alt=.(narrow|variant).\])?"/>
-            <approvalRequirement votes="20" locales="*" paths="//ldml/dates/timeZoneNames/metazone[^/]++/short/[^/]++"/>
-			<approvalRequirement votes="20" locales="ar ca cs da de el es fi fr he hi hr hu it ja ko nl no pl pt pt_PT ro ru sk sl sr sv th tr uk vi zh zh_Hant"
+			<approvalRequirement votes="=tc" locales="Cldr:modern" paths="//ldml/numbers/symbols[^/]++/(decimal|group|(plus|minus)Sign)"/>
```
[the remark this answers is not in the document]
macchiati: ditto

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -216,9 +219,11 @@ public boolean canManageSomeUsers() {
          * on more than this user's level and the other user's desired new level
          */
         public boolean canCreateOrSetLevelTo(Level otherLevel) {
-            return
-                canManageSomeUsers() && // must be some sort of manager
-                otherLevel.getSTLevel() >= getSTLevel(); // can't gain higher privs
+            // Must be a manager at all
+            if (!canManageSomeUsers()) return false;
+            // Cannot elevate privilege
+            if (getSTLevel() > otherLevel.getSTLevel()) return false;
```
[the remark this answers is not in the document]
macchiati: We could add a method morePowerfulThan(...)

tools/cldr-code/src/main/java/org/unicode/cldr/util/Organization.java
```diff
@@ -56,10 +57,26 @@ public enum Organization {
     wikimedia("Wikimedia Foundation"),
     wod_nko("WOD N’ko", "World Organization for the Development of N’ko", "WODN"),
     yahoo("Yahoo"),
-
-    // To be removed.
     ;
 
+    private final static Set<Organization> TC_ORGS = EnumSet.of(google, apple, microsoft);
```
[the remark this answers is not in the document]
srl295: done

docs/ldml/tr35-info.md
srl295: done

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -227,7 +232,7 @@ public boolean canCreateOrSetLevelTo(Level otherLevel) {
          * @param withVotes the given vote count
```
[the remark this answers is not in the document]
srl295: done

PR #1958, opened 2022-04-29
CLDR-15411 Add high coverage locales to Locales.txt

AEApple:
CLDR-15411

Add high coverage locales to Locales.txt so that they show up in the priority view.

I will file a separate ticket about updating the docs.


macchiati:
Looks like there is a test of that in TestStandardCodes.java, so a line needs to be fixed there also.

pedberg-icu (approved):

macchiati (approved):

PR #1967, opened 2022-05-02
CLDR-15599 Update plural many category for Catalan

AEApple:
CLDR-15599

Update plural many category for Catalan. Catalan should have the same behavior as the other romance languages.


AEApple:
So I went back and looked at PR #1426 and it looks like `ca` should be with `it` & `pt_PT` based on the previous changes and the current `ca` `one` and `other` forms. I'm going to wait to see how the tests run and then make any changes in response to the tests.

macchiati (dismissed):

AEApple:
Is it okay if I put in the updated minimal pairs in a 2nd PR? I will ask the ticket submitter to provide the minimal pairs and update separately as long as there are no concerns.

macchiati (approved):

srl295:
> Is it okay if I put in the updated minimal pairs in a 2nd PR? I will ask the ticket submitter to provide the minimal pairs and update separately as long as there are no concerns.

Was there ever a second PR? because `many` is still missing in minimal pairs, and it's breaking the spreadsheet generation CLDR-15647

@macchiati should missing minimal pairs be an error?

PR #1976, opened 2022-05-04
CLDR-15598 Update Google coverage in Locales.txt

meikehubert:
Commented out nqo in Google
Added in Google and Cldr at basic: bho, bgc, raj

CLDR-_____

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
@@ -140,6 +140,8 @@ Google	;	az	;	modern	;	T3.1	Azerbaijani
 Google	;	be	;	modern	;	T5	Belarusian
 #Google	;	bem	;	basic	;	Bemba
 Google	;	bg	;	modern	;	T2	Bulgarian
+Google	;	bgc	;	basic	;	Haryanvi
+Google	;	bho	;	basic	;	Bhojpuri
 Google	;	bn	;	modern	;	T3	Bengali
 Google  ; br  ; moderate ;
 Google  ; brx ; basic ; Bodo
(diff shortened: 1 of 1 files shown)


macchiati:
Meike, this fails because it can't locale bcg.xml (probably doesn't exist yet). 

Note: when you go to https://github.com/unicode-org/cldr/pull/1976/files and search for "Check failure", you'll find an abbreviated error message for the problems.

You should modify this to comment out bcg for now, then make a new PR that includes the bcg locale and this change later.

srl295:
> Meike, this fails because it can't locale bcg.xml (probably doesn't exist yet).
> 
> Note: when you go to https://github.com/unicode-org/cldr/pull/1976/files and search for "Check failure", you'll find an abbreviated error message for the problems.
> 
> You should modify this to comment out bcg for now, then make a new PR that includes the bcg locale and this change later.

@macchiati 
it's in TestPathHeader.TestCompleteness

>             toTest = StandardCodes.make().getLocaleCoverageLocales(Organization.cldr);

So…… basically, `bcg` becomes a "CLDR Locale" because of the entry. I wonder if `getLocaleCoverageLocales` should actually _exclude_ nonexistent locales to avoid this. What do you think?

srl295 (dismissed):
LGTM but needs to pass

macchiati:
We want the test to remain. No point in adding a locale until it exists,
and the test prevents errors.

stenshamn (dismissed):

pedberg-icu:
> I think we actually need to add the core data before we can add the new locales in the locale.txt file.

Yes

srl295:
@meikehubert  I rebased this PR

AEApple (approved):

srl295 (approved):

PR #1980, opened 2022-05-05
CLDR-15545 Update locales.txt for Meta

AEApple:
CLDR-15545

- Update locales.txt with new languages as requested by Meta
- Rename FB to Meta in Locales.txt


tools/cldr-code/src/main/java/org/unicode/cldr/util/Organization.java
```diff
@@ -37,6 +36,7 @@ public enum Organization {
     lakota_lc("Lakota LC"),
     lao_dpt("Lao Posts/Telecom??"),
     longnow("The Long Now Foundation", "Long Now", "PanLex", "Utilka Foundation"),
+    meta("Meta"),
```
srl295, line 39: Needs to retain an alias to Facebook.

srl295 (changes requested):
Needs an alias. I’ll add a more specific comment.

srl295:
@AEApple probably should have a (short) procedure somewherer for this process

tools/cldr-code/src/main/java/org/unicode/cldr/util/Organization.java
srl295:
```suggestion
    meta("Meta", "Facebook"),
```

srl295 (approved):

pedberg-icu (approved):

rickmcgowan (approved):
The changed files look reasonable to me. Note I've updated the cldr-reply address list on corp.unicode.org to add Meta, and also retained FB for now.

PR #1982, opened 2022-05-06
CLDR-14992 SDI.getTargetTerritories/Scripts should include the ones for actual CLDR locales

pedberg-icu:
CLDR-14992

Per the ticket and TC discussion, the locales returned by SupplementalDataInfo getTargetScripts and getTargetTerritories for a given target language should include not just the primary script and region from languageData, but also the explicit scripts and regions from the locale IDs for all (common/main) CLDR locales for that language (except for regions that are 3-digit UN M.49 codes, which we decided in TC to exclude). See the scripts/regions that will be added in [Additions to Target-Scripts, Target-Regions](https://docs.google.com/document/d/1TqsXBswiOSCgHckeAJ--day9WR08EXSNBuJUeJz48Dk/)

This will help do things like ensure that there is Persian calendar coverage in ckb_IR, for example.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
@@ -2639,6 +2639,47 @@ public class SupplementalDataInfo {
         return Level.OPTIONAL.getLevel(); // If no match then return highest possible value
     }
 
+    // The following is for mapping language to the explicit script and region codes in
+    // the locale IDs for CLDR locales. We don't need to worry about the unmarked default
+    // script or region since they are already supplied by the <languageData>.
+    private Map<String, BasicLanguageData> doMapLanguagesToScriptsRegion() {
+        Map<String, BasicLanguageData> langToScriptsRegions = new TreeMap<>();
+        org.unicode.cldr.util.Factory factory = CLDRConfig.getInstance().getCldrFactory();
+        for (CLDRLocale locale : factory.getAvailableCLDRLocales()) {
+            String language = locale.getLanguage();
+            if (language.length() == 0 || language.equals("root")) {
+                continue;
+            }
+            BasicLanguageData scriptsAndRegions = langToScriptsRegions.get(language);
+            if (scriptsAndRegions == null) {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -2639,6 +2639,47 @@ public int getCoverageValueOld(String xpath, ULocale loc) {
...
+            if (scriptsAndRegions == null) {
+                scriptsAndRegions = new BasicLanguageData();
+                langToScriptsRegions.put(language, scriptsAndRegions);
+            }
+            String script = locale.getScript();
+            if (script.length() > 0) {
+                scriptsAndRegions.addScript(script);
+            }
+            String region = locale.getCountry();
+            if (region.length() > 0 && region.length() < 3) { // per CLDR TC, do not want 001, 419 etc.
```
srl295, line 2663: this works, though perhaps a character text (alpha vs digits) would be more robust?

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
pedberg-icu: > It makes some sense, but shouldn't this also be a test to make sure we update supplemental data? Maybe that's already the case and it's an exhaustive test?

I believe that updating supplemental languageData would not actually help here; the languageData has somewhat different intended goals and usages. Many of the script and regions that would be added to coverage per the code here are already in languageData as secondary scripts or regions, but such secondary items are excluded for getTargetScripts and getTargetTerritories. But we don't want to just change getTargetScripts and getTargetTerritories to include the secondary items because that would also add many script and regions for which we do *not* have locales.

pedberg-icu: > this works, though perhaps a character text (alpha vs digits) would be more robust?

Not sure it makes it more robust unless I added both a length and character type check. But not sure it is necessary and is (just slightly) less performant.

pedberg-icu:
Thanks Steven!

PR #1986, opened 2022-05-09
CLDR-14831 add dateTimeFormat type="atTime", move "at" style to that, base form is minimal

pedberg-icu:
CLDR-14831

The "at" style for date-time combining patterns was introduced in CLDR-22, with English changing per [CLDR-4781](https://unicode-org.atlassian.net/browse/CLDR-4781) and many other locales following sure soon after. However, those patterns with "at" are not suitable for many necessary uses of date-time combining patterns, such as combining a date with a time range (for interval formats) or a relative time + absolute time (both as in CLDR-14831), or even just a wall clock time (which is not designating an event at a particular time).

So this PR moves the "at" style patterns to a new type="atTime" variant of the dateTimeFormat, and keeps the base "standard" form  as one without literal text (other than punctuation), for minimal grammatical inflection requirements. In the future we may add more types of date-time combining patterns for different grammatical requirements.

Notes:
- I added the type="atTime" in the dateTimeFormat element, one level above the lev-level pattern element, so that I could add aliases in root from the "atTime" version to the base "standard" version, for each length. Aliases cannot be handled in leaf elements. And then I added these aliases to root.
- In most cases, locales that had the "at" style used it for just the "full" and "long" lengths; for the "medium" and "short" lengths they typically had a punctuation only variant, generally using comma-space or just space. Some locales (e.g. Finnish) used the "at" style for "medium" as well.
- So how I handled the data was to move the "at" style to the "atTime" variant, and copy the longest punctuation-only form (typically medium) to the standard form for long and full as well. However, if the current date-time combining patterns in a locale were the same as before CLDR 22 - as with the Persian locale for example - I dod not make any changes, even though these patterns have literal text.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -1738,8 +1738,8 @@ public class ExampleGenerator {
         String calendar = parts.findAttributeValue("calendar", "type");
 
         if (parts.contains("dateTimeFormat")) {
-            String dateFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "dateFormat"));
-            String timeFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "timeFormat"));
+            String dateFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "dateFormat").replaceAll("atTime", "standard"));
+            String timeFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "timeFormat").replaceAll("atTime", "standard"));
             String dateFormatValue = cldrFile.getWinningValue(dateFormatXPath);
             String timeFormatValue = cldrFile.getWinningValue(timeFormatXPath);
             parts = XPathParts.getFrozenInstance(cldrFile.getFullXPath(dateFormatXPath));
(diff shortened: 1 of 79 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1738,8 +1738,8 @@ private String handleDateFormatItem(String xpath, String value) {
         String calendar = parts.findAttributeValue("calendar", "type");
 
         if (parts.contains("dateTimeFormat")) {
-            String dateFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "dateFormat"));
-            String timeFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "timeFormat"));
+            String dateFormatXPath = cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "dateFormat").replaceAll("atTime", "standard"));
```
macchiati, line 1741: We might want to general multiple examples, one with atTime and one with standard. But that can be a followup.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
pedberg-icu: This will already generate examples for each of the date-time combining pattern styles (the standard and the atTime); the code above is about getting the date and time formats to use in those examples, since there is no type="atTime" in the date & time patterns. But what we probably *do* want to do is generate multiple examples for the "standard" style, including  one with a single time (as generated by the above), and one with a time range(for intervalFormats). I as thinking about that and filed https://unicode-org.atlassian.net/browse/CLDR-15622 for this.

PR #1987, opened 2022-05-09
CLDR-15017 Check for back-and-forth flip-flop values

btangmu:
-New FindFlipFlops.java, partly based on GenerateBirth.java

-List all paths in all locales where value changed from A to B then back to A

CLDR-15017

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/FindFlipFlops.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/FindFlipFlops.java
@@ -1,0 +1,204 @@
+package org.unicode.cldr.tool;
+
+import com.google.common.base.Joiner;
+import com.ibm.icu.util.ICUException;
+import com.ibm.icu.util.VersionInfo;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.unicode.cldr.test.DisplayAndInputProcessor;
+import org.unicode.cldr.tool.Option.Options;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.Factory;
+import org.unicode.cldr.util.LanguageTagParser;
+import org.unicode.cldr.util.SimpleFactory;
+
+public class FindFlipFlops {
+
+    private static int flipFlopCount = 0;
+
+    private static CldrVersion[] VERSIONS;
+
+    private static Factory[] factories;
+
+    private static final boolean USE_RESOLVED = false;
+
+    static final Options myOptions = new Options()
+        .add(
+            "file",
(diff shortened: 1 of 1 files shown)


btangmu:
@macchiati when I run this, 12,590 flip-flops are found; the first and last items are:
🏓🏓 1 //ldml/numbers/currencies/currency[@type="PHP"]/displayName[@count="one"] en
Philippine peso
Philippine peso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine piso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
Philippine peso
...
...
🏓🏓 12590 //ldml/dates/fields/field[@type="day"]/displayName zu
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
Usuku
usuku
usuku
Usuku
Usuku
Usuku

tools/cldr-code/src/main/java/org/unicode/cldr/tool/FindFlipFlops.java
```diff
@@ -0,0 +1,204 @@
...
+                continue; // skip region locales
+            }
+            System.out.println(fileName);
+            findFlops(fileName);
+        }
+        System.out.println("Total " + flipFlopCount + " flip-flops");
+    }
+
+    private static void confirmVersionArchiveIsPresent() {
+        try {
```
btangmu, line 64: some of these methods are based on code in GenerateBirth.java -- if this pr is accepted, some code should go in a shared utility class, in a subsquent pr

macchiati (changes requested):
I think this is almost there, but the information needs to be in a form that makes it easy for use to see patterns.

That is, we need to see not just a raw listing of the same string over and over again (which from the description is what is produced), but rather when each change was made. Eg,
``
#Version[TAB]From[TAB]⇒[TAB]To
21.0[TAB]«abc»[TAB]⇒[TAB]«adc»
23.0[TAB]«adc»[TAB]⇒[TAB]«abc»
``

And (it is hard to tell whether you are doing this from the listing) a line should only be listed if 
  * its From value equals some other lines To value, or 
  * its To value equals some other line's From value.

btangmu:
See https://github.com/unicode-org/cldr/pull/1991

PR #1989, opened 2022-05-10
CLDR-15404 Check the nameOrderLocales

macchiati:
CLDR-15404

Checks that between the two nameOrderLocales, both 'und', and the user's language are mentioned. 

It only checks if neither values is missing, so we don't cause an error on a missing value.

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9239,7 +9239,7 @@ annotations.
 		<featureName type="zero">slashed zero</featureName>
 	</typographicNames>
 	<personNames>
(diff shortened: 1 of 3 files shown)


pedberg-icu:
Getting a unittest error:
```
  TestPersonNameFormatter {
    ...
    TestExampleGenerator
##[error] (TestPersonNameFormatter.java:415)  Error: : Example for und en: expected "〖und = «any other»〗", got "〖und = «any other»〗〖en = English〗"
```

macchiati:
Thanks, made a fix.

pedberg-icu (approved):

PR #1991, opened 2022-05-10
CLDR-15017 Check for back-and-forth flip-flop values

btangmu:
-New FindFlipFlops.java, partly based on GenerateBirth.java

-Find all paths in all locales where value changed from A to B then back to A

-Output tab-separated data including a row for each significant change

CLDR-15017

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/FindFlipFlops.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/FindFlipFlops.java
@@ -1,0 +1,261 @@
+package org.unicode.cldr.tool;
+
+import com.google.common.base.Joiner;
+import com.ibm.icu.util.ICUException;
+import com.ibm.icu.util.VersionInfo;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.unicode.cldr.test.DisplayAndInputProcessor;
+import org.unicode.cldr.tool.Option.Options;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.Factory;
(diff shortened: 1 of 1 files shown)


btangmu:
This finds total 12,646 flip-flops

For partial sample output (a few locales) see https://docs.google.com/spreadsheets/d/1wE2812-SxthrGLSSZYdyKtVRXqDW0Hhv6wOy0dz3G4I/edit#gid=0

macchiati:
We need to be able to see the locale (I'd thought you were going to have
the locally as a subheader)

Better anyway to add a first column with the locale code.

btangmu:
Hi Mark,

Column F is the locale.

Cheers,

Tom

汤姆*文林
mobile

macchiati:
Great thanks. Was viewing on phone and missed that!

macchiati:
I looked more carefully at the data.

1. ↑↑↑ needs to be replaced. When it occurs it needs to be replaced by the resolved value using a resolved CLDRFile instance before any comparisons are done.
2. Please include the path in all (non-empty) rows. We need to be able to view and manipulate the data in a spreadsheet
3. Let's skip releases before 21.0. Early versions were less stable, and we don't care as much about that era
4. When you do comparisons, let's skip differences in whitespace. Again, the goal is to reduce noise so we can see the problem cases more easily. I think simplest is to define a method equalsWithoutWhitespace and use 

return x.equals(y) ? true : x.replaceAll("\h*","").equals(y.replaceAll("\h*","")) 

(I'm writing this without checking. I'm sure that can be optimized, but unless the tool takes a long time to run, probably not worth it).

btangmu:
See https://github.com/unicode-org/cldr/pull/2003

PR #1993, opened 2022-05-10
CLDR-14336 for currency patterns add alts alphaNextToNumber,noCurrency; add element currencyPatternAppendISO

pedberg-icu:
CLDR-14336

This does the following:
-  For currency patterns (standard, accounting, and compact decimal) adds two alt= forms: alphaNextToNumber (for when the currency symbol character nearest to the number is an alpha character) and noCurrency (to get a currency-style pattern without the actual currency symbol). These are mutually exclusive.
    - The alphaNextToNumber form is generally only needed to add space between the symbol and number if the pattern does not already have a space there.
    - For compact decimals, the noCurrency form is only needed if the result needs to be different from the short compact decimal formats, which provide the fallback for these. Currently there are no such cases (formerly the `pt` compact currency formats would have needed this). However, the noCurrency form is still useful for the standard & accounting currency formats (e.g. to get the format with 2 trailing decimals but without a currency sign).
- Under currencyFormats adds a new child element currencyPatternAppendISO which provides a message pattern intended for combining a currency format that uses a symbol with the ISO 4217 code for the currency (e.g. to get $1,432.00 USD). In this PR that pattern exists only in root.
- Adds the necessary data for alphaNextToNumber and noCurrency to locales in common/main/. The alphaNextToNumber patterns are based on the current currencySpacing behavior. All of this data is draft="provisionsal", except for `en`.
- Updates tests, coverage, PathHeader, PathDescription, etc.

Also in SandboxLocalesTest I had to change the minimalDraftStatus for Factory.make from null to unconfirmed, otherwise there was an NPE in XMLNormalizingLoader for locale "root" in the call to cache.getUnchecked(key).

### DIFF
--- a/tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
+++ b/tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
@@ -14,6 +14,7 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.unicode.cldr.test.CheckCLDR;
+import org.unicode.cldr.util.CLDRFile.DraftStatus;
 
 class SandboxLocalesTest {
 
@@ -45,9 +46,14 @@ class SandboxLocalesTest {
         // our factory includes common/main, so limit here.
         for(final CLDRLocale l : SpecialLocales.getByType(SpecialLocales.Type.scratch)) {
             System.out.println("Testing " + l);
-            // TODO With the current changes for CLDR-14336 there is an NPE in the following,
-            // see further notes in SimpleFactory.handleMake
-            CLDRFile f = factory.make(l.getBaseName(), true, null);
+            // As part of CLDR-14336 I had to change the minimalDraftStatus parameter below from null to unconfirmed.
+            // Without that an NPE exception was thrown below the following. The Factory.make call ends up in
+            // SimpleFactory.handleMake; the first call there for "mul" with resolved=true calls makeResolvingSource
+            // which recursively calls SimpleFactory.handleMake, eventually ending up with a call for "root" with
+            // resolved=false that calls new CLDRFile. That ends up in XMLNormalizingLoader.getFrozenInstance. With
+            // the null DraftStatus, the NPE occurred inside the call there to cache.getUnchecked(key) which is
+            // Google library code.
+            CLDRFile f = factory.make(l.getBaseName(), true, DraftStatus.unconfirmed);
             List<CheckCLDR.CheckStatus> errs = new LinkedList<>();
             check.setCldrFileToCheck(f, options, errs);
             for(final CheckCLDR.CheckStatus err : errs) {
--- a/tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
+++ b/tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
@@ -45,6 +45,8 @@ class SandboxLocalesTest {
         // our factory includes common/main, so limit here.
         for(final CLDRLocale l : SpecialLocales.getByType(SpecialLocales.Type.scratch)) {
             System.out.println("Testing " + l);
+            // TODO With the current changes for CLDR-14336 there is an NPE in the following,
+            // see further notes in SimpleFactory.handleMake
             CLDRFile f = factory.make(l.getBaseName(), true, null);
             List<CheckCLDR.CheckStatus> errs = new LinkedList<>();
             check.setCldrFileToCheck(f, options, errs);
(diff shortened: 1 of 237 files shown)


tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
```diff
@@ -45,6 +45,8 @@ void test() throws IOException {
         // our factory includes common/main, so limit here.
         for(final CLDRLocale l : SpecialLocales.getByType(SpecialLocales.Type.scratch)) {
             System.out.println("Testing " + l);
+            // TODO With the current changes for CLDR-14336 there is an NPE in the following,
+            // see further notes in SimpleFactory.handleMake
```
pedberg-icu, line 49: To investigate

tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleFactory.java
```diff
@@ -524,6 +524,11 @@ public CLDRFile handleMake(String localeName, boolean resolved, DraftStatus mini
                 try {
                     makeResolvingSource = makeResolvingSource(localeName, minimalDraftStatus);
                 } catch (Exception e) {
+                    // TODO: The current changes for CLDR-14336 result in a NPE being caught below when this
+                    // is called from SandboxLocalesTest.test(). In that case this handleMake method is called
+                    // recursively via Factory.makeResolvingSource; the third time it is called is with
+                    // localeName=root and resolved=false, and for that case the call to new CLDRFile below
+                    // generates the NPE.
```
pedberg-icu, line 531: The focus of the failure investigation

tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
```diff
@@ -45,7 +46,14 @@ void test() throws IOException {
         // our factory includes common/main, so limit here.
         for(final CLDRLocale l : SpecialLocales.getByType(SpecialLocales.Type.scratch)) {
             System.out.println("Testing " + l);
-            CLDRFile f = factory.make(l.getBaseName(), true, null);
```
srl295, line 48: Could you just omit the third argument? Then it uses the factory's default.  This was definitely a mistake, I didn't have a reason to pass null here.

btangmu (approved):

tools/cldr-apps/src/test/java/org/unicode/cldr/util/SandboxLocalesTest.java
pedberg-icu: I will try that in a future cleanup, I would like to get this merged so we can start shakedown.

pedberg-icu: I did that in https://github.com/unicode-org/cldr/pull/2009

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Placeholders.txt
```diff
@@ -203,6 +203,8 @@
 
 ^//ldml/numbers/currencyFormats/unitPattern ; {0}=NUMBER 3; {1}=CURRENCY_NAME Dollars
 ^//ldml/numbers/currencyFormats\[@numberSystem="%A"]/unitPattern ; {0}=NUMBER 3; {1}=CURRENCY_NAME Dollars
+^//ldml/numbers/currencyFormats/currencyPatternAppendISO ; {0}=CURRENCY_FORMAT ¤#,##0.00
```
sffc, line 206: I don't see any seed data for `currencyPatternAppendISO`; is that intentional?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -138,12 +138,14 @@ public void testCurrency() {
         "//ldml/numbers/currencyFormats/currencySpacing/afterCurrency/currencyMatch",
         "//ldml/numbers/currencyFormats/currencySpacing/afterCurrency/surroundingMatch",
         "//ldml/numbers/currencyFormats/currencySpacing/afterCurrency/insertBetween",
+        "//ldml/numbers/currencyFormats/currencyPatternAppendISO", // TODO see CLDR-14831
```
sffc, line 141: CLDR-14831 doesn't mention anything about currency formatting.

sffc (commented):
Thank you for doing this! I left a couple questions.

(This is my long-delayed Jira ticket review)

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
pedberg-icu: Yes, that is in the linked ticket https://unicode-org.atlassian.net/browse/CLDR-15286

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Placeholders.txt
pedberg-icu: There is a pattern in root. Locales can override it if they choose to.

PR #1970, opened 2022-05-03
CLDR-15588 add nl and nl_BE PersonNames

mimckenna:
CLDR-15588

- [x] This PR completes the ticket.

Add nl and nl_BE personName patterns and sample names.

Note that nl uses surname-prefix and surname-core to handle tussenvoegsels, but -prefix and -core are not used in Belgium

Double-Barreled names are not handled.  See: https://onzetaal.nl/taaladvies/geachte-mevrouw-jansen-de-jong-jansen
In double-barreled surnames, e.g. maiden plus married, defacto usage it to only use the first surname in `usage="addressing"`

  "mevrouw Jansen-de Jong" -> "Dear mevrouw Jansen" (mevrouw = "Ms.")

Test data are in the "nl" and "nl-BE" tabs of the [PersonName Test Data Google Sheet ](https://docs.google.com/spreadsheets/d/1YwWVSgaMNx13O0yf3f86MQjpMqy8VS1gfqGrT38fQ8g/edit?usp=sharing)

PR #1994, opened 2022-05-11
CLDR-15632 prefix & core modifiers for sampleName

mimckenna:
CLDR-15632

Add surname-prefix and surname-core options for personNames/sampleName

* Updates allowed nameField values
* Adds surname-prefix and surname-core test values to root.xml and en.xml

This PR blocks PR #1970 , add nl and nl_BE personNames

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9443,6 +9443,8 @@ annotations.
 			<nameField type="given-informal">Neele</nameField>
 			<nameField type="given2">Eva Sophia</nameField>
 			<nameField type="surname">Meyer Wolf</nameField>
+			<nameField type="surname-prefix">van den</nameField>
+			<nameField type="surname-core">Wolf</nameField>
 			<nameField type="surname2">Becker Schmidt</nameField>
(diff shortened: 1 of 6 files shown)


pedberg-icu (approved):
This all looks good, assuming it passes tests.

common/main/en.xml
```diff
@@ -9443,6 +9443,8 @@ annotations.
 			<nameField type="given-informal">Neele</nameField>
 			<nameField type="given2">Eva Sophia</nameField>
 			<nameField type="surname">Meyer Wolf</nameField>
+			<nameField type="surname-prefix">van den</nameField>
```
macchiati, line 9446: if both 'surname' and the pair 'surname-prefix' and 'surname-core' exist, they should be aligned. So this needs to change "Meyer Wolf" to "van den Wolf"

macchiati: We probably need a test for that also.

macchiati (commented):
Needs one small change.

PR #1995, opened 2022-05-11
CLDR-15615 Add new locale Bhojpuri (bho)

AEApple:
CLDR-15615

Add core date for new locale Bhojpuri (bho)

### DIFF
--- a/seed/main/bho.xml
+++ b/seed/main/bho.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2022 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="bho"/>
+	</identity>
+	<localeDisplayNames>
+		<languages>
+			<language type="bho" draft="contributed">भोजपुरी</language>
+		</languages>
+		<scripts>
+			<script type="Deva" draft="contributed">देवानागारी</script>
+		</scripts>
+	</localeDisplayNames>
+	<characters>
+		<exemplarCharacters draft="contributed">[क ख ग घ डं च छ ज झ ञ ट ठ ड ढ ण ड़ ढ़ त थ द ध न प फ ब भ म य र ल व श ष स ह क्ष त्र ज्ञ श्र ा  ि  ी  ु  ू  ृ  े  ै  ो  ौ  ँ  ं  ः  ॅ  ्]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[अ आ इ ई उ ऊ ऋ ए ऐ ओ औ]</exemplarCharacters>
+		<exemplarCharacters type="numbers" draft="contributed">[\- ‑ , . % ‰ + 0० 1१ 2२ 3३ 4४ 5५ 6६ 7७ 8८ 9९]</exemplarCharacters>
+		<exemplarCharacters type="punctuation" draft="provisional">[\- _ – — , ; ~ \: \\ ! | ? . … ` &quot; ‘ ’ “ ” ( ) \{ \} \[ \] § @ * / + \&amp; # ′ ″]</exemplarCharacters>
+	</characters>
+	<numbers>
+		<defaultNumberingSystem>deva</defaultNumberingSystem>
+	</numbers>
+</ldml>
--- a/common/supplemental/supplementalMetadata.xml
+++ b/common/supplemental/supplementalMetadata.xml
@@ -1801,7 +1801,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<defaultContent locales="
 			aa_ET ab_GE af_ZA agq_CM ak_GH am_ET an_ES ann_NG ar_001 arn_CL as_IN asa_TZ ast_ES
 			az_Arab_IR az_Cyrl_AZ az_Latn az_Latn_AZ
-			ba_RU bal_Arab bal_Arab_PK bal_Latn_PK bas_CM be_BY bem_ZM bez_TZ bg_BG bgn_PK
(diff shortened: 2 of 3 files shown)


seed/main/bho.xml
```diff
@@ -0,0 +1,30 @@
...
+<!-- Copyright © 1991-2022 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+-->
+<ldml>
+		<identity>
+			<version number="$Revision$"/>
+			<language type="bho"/>
+			<script type="Deva"/>
```
pedberg-icu, line 12: The identity element should only reflect items in the locale ID of the file, so delete the `<script type="Deva"/>`

common/supplemental/attributeValueValidity.xml
```diff
@@ -12,7 +12,7 @@
 			<!-- The following are in $language but not in comon/main or in modern coverage: cu lij prg vo -->
 			<variable id='$language' type='choice'>
 				af agq ak am ann ar as asa ast az
-				bas be bem bez bg bm bn bo br brx bs
+				bas be bem bez bg bho bm bn bo br brx bs
```
pedberg-icu, line 15: Actually we do not want to add this. We already have `bho` in the $languageExceptions section of this file, so vetters are already being asked for its display name in different languages. The codes in the $language section are generally only for locales we have in common/main/.

pedberg-icu:
We also need to add an empty bho_IN.xml file (you can look at other empty region stubs in seed/main for an example). And then in the supplementalMetadata.xml file, we need to add bho_IN in the defaultContent section.

AEApple:
Made all requested changes, please let me know if there is anything further, otherwise I will submit PRs for the two other new locales following the same format followed by updating locales.txt file.

DavidLRowe:
Didn't have time this evening for more than a cursory look, but it seems that the auxiliary exemplarCharacter element is missing its closing tag.

AEApple:
> Didn't have time this evening for more than a cursory look, but it seems that the auxiliary exemplarCharacter element is missing its closing tag.

Thanks! Fixed, I'll check again after the tests complete running again.

seed/main/bho.xml
```diff
@@ -19,10 +19,10 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		</scripts>
 	</localeDisplayNames>
 	<characters>
-		<exemplarCharacters draft="contributed">[क ख ग घ डं च छ ज झ ञ ट ठ ड ढ ण ड़ ढ़ त थ द ध न प फ ब भ म य र ल व श ष स ह क्ष त्र ज्ञ श्र ा  ि  ी  ु  ू  ृ  े  ै  ो  ौ  ँ  ं  ः  ॅ  ॔  ्]</exemplarCharacters>
-		<exemplarCharacters type="auxiliary">[अ आ इ ई उ ऊ ऋ ए ऐ ओ औ अं अ:]</exemplarCharacters>
+		<exemplarCharacters draft="contributed">[क ख ग घ डं च छ ज झ ञ ट ठ ड ढ ण ड़ ढ़ त थ द ध न प फ ब भ म य र ल व श ष स ह क्ष त्र ज्ञ श्र ा  ि  ी  ु  ू  ृ  े  ै  ो  ौ  ँ  ं  ः  ॅ  ्]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[अ आ इ ई उ ऊ ऋ ए ऐ ओ औ अ:]</exemplarCharacters>
 		<exemplarCharacters type="numbers" draft="contributed">[\- ‑ , . % ‰ + 0० 1१ 2२ 3३ 4४ 5५ 6६ 7७ 8८ 9९]</exemplarCharacters>
-		<exemplarCharacters type="punctuation" draft="provisional">[- _ – — , ; ~ : \ ! | ? . … ` ' ‘ ’ " “ ” ( ) { } [ ] § @ * / + - # ′ ″]</exemplarCharacters>
+		<exemplarCharacters type="punctuation" draft="provisional">[\- _ – — , ; ~ : \ ! | ? . … ` ' ‘ ’ " “ ” ( ) { } [ ] § @ * / + # ′ ″]</exemplarCharacters>
```
pedberg-icu, line 25: Some of the characters in th punctuation set need to be escaped (with backslash in front of them) because they have special meaning in a UnicodeSet. In particular the square brackets, otherwise e.g. the closing one is interpreted as the end of the set. Also the colon. So these need be represented as \:   \[   \]

pedberg-icu:
Oh good catch on the additional escapes, I missed that those characters were in there.

pedberg-icu (approved):
Looks good now, assuming tests pass

DavidLRowe:
The punctuation exemplar has `&quot;` to represent U+0022, but it also has `"` (three items later in the list). The `"` should be removed.

pedberg-icu (approved):

seed/main/bho.xml
```diff
@@ -20,7 +20,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	</localeDisplayNames>
 	<characters>
 		<exemplarCharacters draft="contributed">[क ख ग घ डं च छ ज झ ञ ट ठ ड ढ ण ड़ ढ़ त थ द ध न प फ ब भ म य र ल व श ष स ह क्ष त्र ज्ञ श्र ा  ि  ी  ु  ू  ृ  े  ै  ो  ौ  ँ  ं  ः  ॅ  ्]</exemplarCharacters>
-		<exemplarCharacters type="auxiliary">[अ आ इ ई उ ऊ ऋ ए ऐ ओ औ अ:]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[अ आ इ ई उ ऊ ऋ ए ऐ ओ औ]</exemplarCharacters>
```
pedberg-icu, line 23: I wonder whether that character with colon in aux was actually supposed to have \u0903 DEVANAGARI VISARGA instead, which is a combining char that looks like colon

pedberg-icu: (But \u0903 is already in the main exemplars so no need to have the combination in aux)

DavidLRowe:
I think (based on what I see in other locale files) that `{` and `}` need escaping as well, that is: `\{` and `\}` .

DavidLRowe (dismissed):

seed/main/bho.xml
AEApple: Makes sense!

macchiati (approved):

PR #2000, opened 2022-05-11
CLDR-15314 add tok Toki Pona

DavidLRowe:
CLDR-15314


DavidLRowe:
@srl295 I modeled this on eo, but did not include any population entries in territoryInfo in supplementalData.xml (whereas eo has an entry under San Marino).

srl295 (approved):

PR #2001, opened 2022-05-12
CLDR-14032 More space adjustments in DAIP; add test. Run it on en.xml, adjust tests

pedberg-icu:
CLDR-14032

Enhanced DAIP to make the space adjustments per the ticket and associated [proposal](https://docs.google.com/document/d/1MuHnd5hvrwI1TqgxW_js4lgzh7nQRySjNHOsFJCF2qY/edit#heading=h.xcfpr8iehq4e) as discussed in TC:
- In Latin-script locales intervalFormats, spaces around \u2013 become thin space \u2009
- In time formats, space between time and AM/PM marker becomes NNBSP \u202F
- In Cyrillic-script locales date formats with year, space between y and year marker becomes NNBSP \u202F
- In narrow units, space(s) on either side of {0} become NNBSP \u202F
- In short units,  space(s) on either side of {0} become NBSP \u00A0

Added a unit test for these changes.

Then ran CLDRModify -fp on en.xml to show the result of these changes (that also reordered some data items not related to the DAIP changes here). The resulting spacing changes required adjusting the expected results of some unit tests.

I  plan to run CLDRModify on the other locales as a separate PR before start of regular submission, either under the ticket for this or another ticket.

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
@@ -364,6 +364,10 @@ public class TestDisplayAndInputProcessor extends TestFmwk {
         if (path.contains("/foreignSpaceReplacement")) {
             return null; // CLDR-15384 typically inherited; no DAIP processing desired
         }
+        if (logKnownIssue("CLDR-15635", "Skip TestAll() for /intervalFormatItem until we update xml data") &&
+            (path.contains("/timeFormat") || path.contains("/dateFormatItem") || path.contains("/intervalFormatItem"))) {
+            return null; // CLDR-14032 changing normalization for intervalFormats but xml data not yet updated
+        }
         if (path.contains("/exemplarCharacters") || path.contains("/parseLenient")) {
             try {
                 UnicodeSet s1 = new UnicodeSet(value);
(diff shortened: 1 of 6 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
```diff
@@ -477,6 +481,64 @@ public static PathSpaceData[] getArray() {
...
+                    "{0} g", "{0} g"), //  \u202F
+            new PathSpaceAdjustData("en",
+                    "//ldml/units/unitLength[@type=\"narrow\"]unit[@type=\"mass-gram\"]/unitPattern[@count=\"other\"]",
+                    "g {0}", "g {0}"), //  \u202F
+            new PathSpaceAdjustData("en",
+                    "//ldml/units/unitLength[@type=\"short\"]unit[@type=\"mass-gram\"]/unitPattern[@count=\"other\"]",
+                    "{0} g", "{0} g"), // \u00A0
+            new PathSpaceAdjustData("en",
+                    "//ldml/units/unitLength[@type=\"short\"]unit[@type=\"mass-gram\"]/unitPattern[@count=\"other\"]",
+                    "g {0}", "g {0}"), // \u00A0
```
btangmu, line 532: the comments on lines 529 and 532 refer to \u00A0, which, however, doesn't seem to occur in the data on those lines

btangmu (changes requested):
the comments on lines 529 and 532 of TestDisplayAndInputProcessor.java refer to \u00A0, which, however, doesn't seem to occur in the data on those lines

it might be more robust if the \u escapes were in the strings themselves rather than in comments -- anyway, here the comments don't seem to match the data

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
btangmu: sorry, never mind, they do have \u00A0 but somehow it was changed to \u0020 when copy-pasting from browser

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -834,13 +834,13 @@ public void TestDayPeriods() {
         checkDayPeriod("pl", "format", "morning1", "〖06:00 – 10:00⁻〗〖❬8:00 ❭rano〗");
         checkDayPeriod("pl", "stand-alone", "morning1", "〖06:00 – 10:00⁻〗");
 
-        checkDayPeriod("en", "format", "night1", "〖00:00 – 06:00⁻; 21:00 – 24:00⁻〗〖❬3:00 ❭at night〗");
+        checkDayPeriod("en", "format", "night1", "〖00:00 – 06:00⁻; 21:00 – 24:00⁻〗〖❬3:00 ❭at night〗");
```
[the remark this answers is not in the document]
pedberg-icu: Yes, filed a separate ticket for that: CLDR-15636

pedberg-icu:
I will file a separate ticket to update data files per CLDRMofiy before general submission. Should probably add to BRS too.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -834,13 +834,13 @@ public void TestDayPeriods() {
...
         checkDayPeriod("en", "stand-alone", "night1", "〖00:00 – 06:00⁻; 21:00 – 24:00⁻〗");
 
-        checkDayPeriod("en", "format", "noon", "〖12:00〗〖❬12:00 ❭noon〗");
-        checkDayPeriod("en", "format", "midnight", "〖00:00〗〖❬12:00 ❭midnight〗");
-        checkDayPeriod("en", "format", "am", "〖00:00 – 12:00⁻〗〖❬6:00 ❭AM〗");
-        checkDayPeriod("en", "format", "pm", "〖12:00 – 24:00⁻〗〖❬6:00 ❭PM〗");
+        checkDayPeriod("en", "format", "noon", "〖12:00〗〖❬12:00 ❭noon〗");
+        checkDayPeriod("en", "format", "midnight", "〖00:00〗〖❬12:00 ❭midnight〗");
```
macchiati, line 841: I think it is probably a mistake to use a NBSP (or NNBSP) between the time (13:00) and an unabbreviated day period. The am/pm are fine, because they are short. But forcing a string like "12:00 midnight" to break as a unit could leave lines unnecessarily ragged.

However, I think that can be discussed and corrected afterwards if necessary.

pedberg-icu: @macchiati I did *not* do anything to the spacing in the patterns that have B (for day periods). I think the issue here is just that the ExampleGenerator code is using time patterns with 'a' to generate examples for dayPeriods. That seems like an ExampleGenerator problem.

pedberg-icu: Oh, actually I think it might be in the DateTimePatternGenerator code which, if given a skeleton with 'B', will adjust a time pattern that has 'a' by substituting in the 'B'. Hmm, need to think about how to address that.

macchiati:
Or replace

pedberg-icu:
> Or replace
> […](#)

Yeah, that is actually what I meant, sorry.

macchiati:
NP. that's what I figured 🙂

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
pedberg-icu: Actually I do not think this is a necessarily problem in DTPG. We have separate availableFormats and intervalFormats for Bh, Bhm and Bhms - and this PR did not change those - so if Example Generator is correctly using DTPG and DIF it should not have gotten the NNBSP above. It may be an ExampleGenerator problem after all. I will investigate.

PR #2004, opened 2022-05-13
CLDR-15617 Rajasthani (raj) Core Data

AEApple:
CLDR-15617

Add Core data for Rajasthani (raj)

### DIFF
--- a/common/supplemental/supplementalMetadata.xml
+++ b/common/supplemental/supplementalMetadata.xml
@@ -1801,7 +1801,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<defaultContent locales="
 			aa_ET ab_GE af_ZA agq_CM ak_GH am_ET an_ES ann_NG ar_001 arn_CL as_IN asa_TZ ast_ES
 			az_Arab_IR az_Cyrl_AZ az_Latn az_Latn_AZ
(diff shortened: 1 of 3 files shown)


AEApple:
A couple of questions:

1. Where should I add the time cycle?
2. Where do I add the Romanization data?
3. Where do I add the country & population data? (especially since they have specified a specific Indian state)
4. They listed `अ` under aux, but if a word is `अ` initial then it would be commonly used, should I move it to main?

macchiati:
> A couple of questions:
> 
> 1. Where should I add the time cycle?
It is normally inherited from the region, in this case India. So unless you have some reason to believe it's different you don't need to do anything.
> 2. Where do I add the Romanization data?
For the romanization data, we don't need that until a higher level.
> 3. Where do I add the country & population data? (especially since they have specified a specific Indian state)
That is in the supplemental data, under language / territory. It would go just under India, we don't keep data about the individual subdivisions.
> 4. They listed `अ` under aux, but if a word is `अ` initial then it would be commonly used, should I move it to main?
If it is a letter in common words, then it would go into Main.

AEApple:
> 

I clearly did this completely backwards. I will fix, I just haven't had time yet this morning :)

macchiati (approved):

PR #2005, opened 2022-05-13
CLDR-15637 update sampleName surname-core

mimckenna:
CLDR-15637

- [ ] This PR does NOT complete the ticket.

Update en.xml personNames/sampleName with matching surname and surname-core.

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9471,7 +9471,7 @@ annotations.
 			<nameField type="given">Ada Cornelia</nameField>
 			<nameField type="given-informal">Neele</nameField>
 			<nameField type="given2">Eva Sophia</nameField>
-			<nameField type="surname">Meyer Wolf</nameField>
+			<nameField type="surname">van den Wolf</nameField>
 			<nameField type="surname-prefix">van den</nameField>
 			<nameField type="surname-core">Wolf</nameField>
 			<nameField type="surname2">Becker Schmidt</nameField>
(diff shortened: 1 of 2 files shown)


macchiati (changes requested):
The problem is that the surname has to equal surname-prefix + " " + surname-core.

Your change has:

Meyer Wolf == van den + Meyer Wolf

which is false. 

Change to:
surname: van den Wolf
surname-core: Wolf
Then we will have:

van den Wolf == van den + Wolf

mimckenna:
I'm updating the surname to equal prefix + core, but the remainder check logic allows surname to equal core without the prefix, such as 
```
{ surname: "Wolf", surname-prefix: "van den", surname-core: "Wolf" }
```

from test at: https://jsfiddle.net/mimckenna/oyq1sdap/42/
image
 
Also, Malaysian will have name where the surname may be independent of a surname-prefix (see: https://www.upu.int/UPU/media/upu/PostalEntitiesFiles/s42/mysEn.pdf)

Example: ms_MY: `{ prefix:"Puasn",  given:"Aisyah", surname-prefix:"binti Haji", surname:"Musa" }`
image

Malaysian _could_ be represented with only surname-prefix and surname-core, but it would be incorrect to assume that surname on its own would always be composed of surname-prefix and surname-core that way it is in Dutch.

macchiati (approved):

macchiati:
Let's discuss this via email

PR #2007, opened 2022-05-13
CLDR-10952 add mdf Moksha

DavidLRowe:
CLDR-10952


DavidLRowe:
Note that the original CLDR ticket ([CLDR-10952](https://unicode-org.atlassian.net/browse/CLDR-10952)) included `|` in the exemplar lists. What I included in this pull request is my understanding of what was intended.

srl295 (approved):

srl295:
@rueter  hi! that webhook "approved" because the ticket is in the correct state. It's merged. If there's more to change, the best is now to update it via the Survey Tool, at https://st.unicode.org/cldr-apps/v#/mdf/Alphabetic_Information/7ed88347aa1b55ed for example.

PR #2011, opened 2022-05-16
CLDR-7813 Add islamic to calendarPreferenceData & coverage for more regions

pedberg-icu:
CLDR-7813

Add Islamic calendar to supplemental calendarPreferenceData and to coverage (if not already there) for more regions, but after Gregorian in preference.

In calendarPreferenceData the following prefer
- astronomical islamic: BD ID MY NE PK
- civil islamic: AL AZ MV TJ TM TR UZ XK


roozbehp:
I have no easy way of confirming the preference for astronomical vs civil Islamic calendar in the various countries listed, but I can confirm that the Islamic calendar is very important in the newly added countries, all of which have a significant Muslim population.

roozbehp (approved):

PR #2013, opened 2022-05-16
CLDR-13263 Merge getConstructedBaileyValue and getBaileyValue

btangmu:
-In progress...

CLDR-13263

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -36,6 +36,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.unicode.cldr.test.CheckMetazones;
+import org.unicode.cldr.tool.resolver.CldrResolver;
 import org.unicode.cldr.util.DayPeriodInfo.DayPeriod;
 import org.unicode.cldr.util.GrammarInfo.GrammaticalFeature;
 import org.unicode.cldr.util.GrammarInfo.GrammaticalScope;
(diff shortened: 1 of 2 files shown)


btangmu:
@macchiati @srl295 this draft PR is just work in progress -- it does fix one bug, if I'm not mistaken, that is, it results in the sublocale de_AT inheriting "Moldauisch" not "Rumänisch (Republik Moldau)"...

- Policy question, see email to cldr-core (subject "Policy question for constructed versus inherited values")
- https://st.unicode.org/cldr-apps/v#/de/Languages_O_S/43862f6b62df23e5 
- https://st.unicode.org/cldr-apps/v#/de_AT/Languages_O_S/43862f6b62df23e5 
-Have multiple candidates for inheritance; if there is a valid parent value that would be the bailey value instead of the constructed value

More generally this PR makes getConstructedBaileyValue identical to getBaileyValue, and it causes getBaileyValue to get the constructed value if, and only if, the value would otherwise be null or inherited from root

Probably some code that now calls getConstructedBaileyValue or getConstructedValue needs simplification -- the very existence of constructed values ought to be encapsulated, and most callers shouldn't need to know about constructed values

macchiati:
Sounds good.

> the very existence of constructed values ought to be encapsulated, and
most callers shouldn't need to know about constructed values

There are test programs and console checks that need to look at the 'raw'
values, so we need to allow access to the lower levels. But at the
higher levels most callers should probably get the constructed values. So I
think it is a good idea, but we need to take a look at what is going on
before changing a resolved "getStringValue()"

btangmu:
See https://github.com/unicode-org/cldr/pull/2020

PR #2024, opened 2022-05-18
CLDR-14064 Prevent submission of root annotation values/codes

btangmu:
-Revise CheckForCopy to recognize annotation codes like E10-836

-Move some code into new methods

-New class AnnotationUtil.java encapsulates info about annotation paths/values

-Revise import statements, done automatically by IntelliJ

CLDR-14064

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/AnnotationUtil.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/AnnotationUtil.java
@@ -1,0 +1,23 @@
+package org.unicode.cldr.util;
+
+import java.util.regex.Pattern;
+
+public class AnnotationUtil {
+    private static final String PATH_PREFIX = "//ldml/annotations/annotation";
+
+    private static final Pattern BAD_EMOJI = Pattern.compile("E\\d+(\\.\\d+)?-\\d+");
+
+    /**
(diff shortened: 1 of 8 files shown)


pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/AnnotationUtil.java
```diff
@@ -0,0 +1,23 @@
+package org.unicode.cldr.util;
+
+import java.util.regex.Pattern;
+
+public class AnnotationUtil {
```
srl295, line 5: OK, but I would just fold this into org.unicode.cldr.util.Annotations itself.

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/AnnotationUtil.java
srl295: could be future

btangmu: i considered it and then worried about initialization problems, seeing the complexity of Annotations.java and not being very familiar with it -- there's been enough excitement recently :-)

PR #2025, opened 2022-05-18
CLDR-15652 integration cleanups: preserve space in foreignSpaceReplacement

pedberg-icu:
CLDR-15652

Need to preserve a space as the value of the foreignSpaceReplacement element (it was getting stripped by e.g. CLDRModify -fp). There are a couple of issues:
- XML behavior for handling spaces (mainly an issue for third party clients of CLDR data). There is a standard xml:space attribute for controlling whether spaces are preserved in values.
- DAIP behavior. DAIP stripped except for hardcoded paths. Added foreignSpaceReplacement to that list but also made it pay attention to xml:space.

After fixing the above, restore the root value of foreignSpaceReplacement to be space.

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3170,6 +3170,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@METADATA-->
 
 <!ELEMENT foreignSpaceReplacement ( #PCDATA ) >
+<!ATTLIST foreignSpaceReplacement xml:space (default | preserve) "preserve" >
+    <!--@VALUE-->
 <!ATTLIST foreignSpaceReplacement alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
 <!ATTLIST foreignSpaceReplacement draft (approved | contributed | provisional | unconfirmed) #IMPLIED >
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/api/AttributeKey.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/api/AttributeKey.java
@@ -92,9 +92,9 @@ public final class AttributeKey {
         // 1) we don't expect the attribute name to have a namespace either,
         // 2) the attribute key should be in our cache of known instances.
         if (elementName.indexOf(':') == -1) {
-            checkArgument(attributeName.indexOf(':') == -1,
-                "attributes in an external namespace cannot be present in elements in the default"
-                    + " namespace: %s:%s",
+            checkArgument((attributeName.startsWith("xml:") || attributeName.indexOf(':') == -1),
+                "attributes in an external namespace other than xml: cannot be present in"
+                    + " elements in the default namespace: %s:%s",
                 elementName, attributeName);
             return checkNotNull(KNOWN_KEYS.get(elementName, attributeName),
                 "unknown attribute (was it deprecated?): %s:%s",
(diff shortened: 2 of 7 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/api/AttributeKey.java
```diff
@@ -92,9 +92,9 @@ public static AttributeKey keyOf(String elementName, String attributeName) {
         // 1) we don't expect the attribute name to have a namespace either,
         // 2) the attribute key should be in our cache of known instances.
         if (elementName.indexOf(':') == -1) {
-            checkArgument(attributeName.indexOf(':') == -1,
-                "attributes in an external namespace cannot be present in elements in the default"
-                    + " namespace: %s:%s",
+            checkArgument((attributeName.startsWith("xml:") || attributeName.indexOf(':') == -1),
```
macchiati, line 95: Does this want to be !attributeName.startsWith("xml:") &&

macchiati (changes requested):
Looks good except for a question I had.

tools/cldr-code/src/main/java/org/unicode/cldr/api/AttributeKey.java
pedberg-icu: @macchiati No, the first (boolean) argument to checkArgument should be true to indicate argument is OK, i.e. it either starts with "xml:" or does not have ':'

pedberg-icu: @macchiati No, the first (boolean) argument to checkArgument should be true to indicate argument is OK, i.e. it either starts with "xml:" or does not have ':'

The message is for if the check fails, i.e. the boolean value is false

btangmu (approved):

common/dtd/ldml.dtd
```diff
@@ -3170,6 +3170,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@METADATA-->
 
 <!ELEMENT foreignSpaceReplacement ( #PCDATA ) >
+<!ATTLIST foreignSpaceReplacement xml:space (default | preserve) "preserve" >
+    <!--@VALUE-->
```
srl295, line 3174: `@VALUE` is probably not right here. THis should be handled globally because it is part of XML. It's actually not a 'value' of this xpath, its' structural.  CLDR-15671

tools/cldr-code/src/main/java/org/unicode/cldr/api/AttributeKey.java
srl295: Ideally, this should be a function and check against a list.

common/dtd/ldml.dtd
pedberg-icu: Agreed, but as I recall I was getting a unittest error without it. Also I think SurveyTool is not paying attention to this attribute and is not actually displaying the space.

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -429,8 +429,10 @@ public synchronized String processInput(String path, String value, Exception[] i
                 value = normalizeWhitespace(path, value);
             }
 
-            // all of our values should not have leading or trailing spaces, except insertBetween
-            if (!path.contains("/insertBetween") && !isUnicodeSet) {
+            // all of our values should not have leading or trailing spaces, except insertBetween,
+            // foreignSpaceReplacement, and anything with
```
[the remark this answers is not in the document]
pedberg-icu: Fixed that in https://github.com/unicode-org/cldr/pull/2056

PR #2032, opened 2022-05-18
CLDR-15656 refactor DB stuff out of UserRegistry

srl295:
- DBUserRegistry for 'normal' db use
- incomplete MemoryUserRegistry subclass for testing

CLDR-15656


srl295:
massive conflicts now - will have to pick this up again another time

srl295:
@btangmu  As you can imagine, this PR from *may 2022* is more than a little bit bit-rotten…

but I think it is a worthwhile effort, so I'm going to start on redoing it.  The basic process is to make `UserRegistry` abstract, with the SQL stuff in a subclass, and an in-memory version for testing use.  

Anyway. WIP on https://github.com/srl295/cldr/tree/wip/cldr-15656/refactor-userreg and eventually I'll update this PR.

PR #2033, opened 2022-05-19
CLDR-15223 feedback for adding users

srl295:
- add API endpoint for locale normalizer
- add LocaleNormalizer.getMessageArray()
- live feedback on add user panel

CLDR-15223

image below from adding `und doi ff es FF` as locales  for an Adlam user

image

the field gets updated to just 'ff'

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrText.js
+++ b/tools/cldr-apps/js/src/esm/cldrText.js
@@ -529,6 +529,10 @@ const strings = {
     "The coverage level for each locale is organization-neutral.",
   summary_access_denied:
     "To see the summary, you must be logged in as a manager, TC, or admin.",
+
+  // LocaleNormalizer.LocaleRejection
+  locale_rejection_unknown: "Unknown or not in CLDR",
+  locale_rejection_outside_org_coverage: "Outside of org’s coverage",
 };
 
 /**
--- a/tools/cldr-apps/js/src/views/AddUser.vue
+++ b/tools/cldr-apps/js/src/views/AddUser.vue
@@ -67,14 +67,31 @@
               id="new_locales"
               name="new_locales"
               v-model="newUser.locales"
+              @change="validateLocales"
+              placeholder="en de de_CH fr zh_Hant"
             />
             <button v-on:click="setAllLocales()">All Locales</button><br />
-            (Space separated. Examples: "en de de_CH fr zh_Hant". Use the All
-            Locales button to grant access to all locales. )
+            (Space separated. Use the All Locales button to grant access to all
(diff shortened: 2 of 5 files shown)


macchiati:
"Locale list was updated due to:"

That is the *opposite* of what the message wants to be.

What we want to say is:

The following locales were *not* added due to problems:
- **und** — the locale "und" does not match an individual language
- **doi (Dogri)** — is outside of your organization's coverage
- FF — does not have the right casing for a language code
- **es (Spanish)** — is outside of your organization's coverage
- $$ — is not a valid language code

(I added the $$ as a weird case)

macchiati (changes requested):
(see comment)

macchiati:
PS, I can see the rationale for not accepting FF. While it is a valid language code, it is not in canonical form, and might be from the user's typing in a region code by mistake.

srl295:
> "Locale list was updated due to:"
> 
> 
> 
> That is the *opposite* of what the message wants to be.

Good feedback. Quick notes:

- "updated" means what you typed was normalized. It does this when you exit the field. 
- the normalization and messages are pre existing. But can definitely work on the wording. 
- I'll add a test for the normalization logic if there isn't one.

DavidLRowe:
Does this check (for example) a user entering "kmr" (ISO 639-3 code for northern Kurdish), when "ku" is what is needed?

macchiati:
I like that. Have a low-level common method to check a locale string, with
options. Returns locale string (or und), and fills enum set

Enum set indicates whether canonicalized and how, whether disallowed and
how. Each enum value has a message.

srl295:
> I like that. Have a low-level common method to check a locale string, with options. Returns locale string (or und), and fills enum set Enum set indicates whether canonicalized and how, whether disallowed and how. Each enum value has a message.
> […](#)

it's LocaleNormalizer and it was added in #1290 a year ago.

srl295:
> Does this check (for example) a user entering "kmr" (ISO 639-3 code for northern Kurdish), when "ku" is what is needed?

`kmr` (and `eng` for that matter) return just 'unknown'.

srl295:
OK current update. 

* First, removed 'example' and made it a placeholder:
image
(shows as grayed out `en de de_CH…` if you delete all text)

* Next, changed LocaleNormalizer to return enums.    Localizable in cldrText.js.  Also, use a display name wherever we have it.

image

If the UI itself looks OK, let's iterate on the normalizer's behavior separately, because it's already called from many other spots.

DavidLRowe:
That's probably reasonable. Good.

macchiati:
LGTM

PR #2034, opened 2022-05-20
CLDR-13879 Add core data for vec

AEApple:
CLDR-13879


DavidLRowe:
Error log says:
```
Error: This field must be a set of the form [a b c-d ...]: Error: '^' not after '[' at "[' \u2019 / \\\\ \u2044 \\: \u2053 \u2010 \u2012 \u2013 \u2014 \u2015 ( ) \\[ \\] \\{ \\} \u2329 \u232A\u2026 . \u00B7 , ; ! ? \u00AB \u00BB \u2018 \u2019 \u201C \u201D @ \\& # \u20AC $ ^| + \\- =]"
```

DavidLRowe:
The `^` needs to be escaped (or removed). 

Also U+2019 appears twice. I suggest you delete the one just after the `&apos;`.

DavidLRowe:
Just out of curiosity I checked if any other locales have `\^` and found only one: kby.xml (in exemplars/main/). 

So, either side with the majority and remove `^` or else change it to `\^`.

srl295:
> Just out of curiosity I checked if any other locales have `\^` and found only one: kby.xml (in exemplars/main/). 
> 
> 
> 
> So, either side with the majority and remove `^` or else change it to `\^`.

Generally there are many more punctuation characters then usually present. ^ is not normally used in punctuation I'd think.

AEApple:
Okay, updated. Tests are running again. I'll squash and send for review if it passes all the tests.

AEApple:
I think we need to clarify the exemplars page for people who are contributing for new languages (especially around what not to include.)

DavidLRowe (approved):
AFAICT this looks fine. This locale will need a numbers exemplar to reach Core minimum. (Not sure if that is needed before Survey Tool can be used?)

srl295:
> AFAICT this looks fine. This locale will need a numbers exemplar to reach Core minimum. (Not sure if that is needed before Survey Tool can be used?)

Italian's numbers exemplars are the same as root, and probably same as `vec`.  Worst case is it requests confirmation of that data item, but it won't affect other processing.

srl295 (approved):
Looks good! Pasa na bèƚa xornada!

AEApple:
Great! I will merge this and then we should look at what else is remaining to do so people can contribute in Survey Tool

PR #2035, opened 2022-05-23
CLDR-15646 Priority Items: fix coverage level if orgIsNeutralForSummary

btangmu:
-This affects snapshots, and summaries by Admin

CLDR-15646


srl295 (approved):
good, but wonder if this should be a function. I guess the other callers aren't quite identical.

macchiati:
Agreed.

PR #2037, opened 2022-05-23
CLDR-15662 LocaleCompletion (third) Meter fix missing count

btangmu:
-Do not assume hasError is true when addMissing needs to be called

-Test (hasError || statusMissing) instead of only hasError

-Fix a few warnings shown by IntelliJ

-Comments

CLDR-15662

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
@@ -4,7 +4,6 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -46,7 +45,6 @@ import org.unicode.cldr.web.STFactory;
 import org.unicode.cldr.web.SurveyLog;
 import org.unicode.cldr.web.SurveyMain;
 
-import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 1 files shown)


srl295:
Can you update the unit test also? TestLocaleCompletion

btangmu:
@srl295 this draft pr points to a something, however I don't think it can be a full solution

further testing shows that we can get statusMissing == true, hasError == false, and results.size() == 0 at the point in question

locally, for French, dash has 159 missing; this pr changes locale completion from 2594 / 2595 to 2594 / 3123 -- increasing the difference from 1 to 529 ( = 2594 / 3123) -- that's a lot more than the 159 "Missing" per Dash, so what are the other 529 - 159 = 370?

the distinction between "missing" and "statusMissing" is mysterious...

seems like if LocaleCompletion.handleGetLocaleCompletion shared more code with VettingViewer.handleOnePath they would be more likely to be consistent -- as it stands they're both incredibly complicated and it would seemingly take a miracle for them to be consistent with each other; presumably VettingViewer.handleOnePath is currently the more accurate of the two

btangmu:
> Can you update the unit test also? TestLocaleCompletion

there's a comment: "TODO Temporarily disabled for CLDR-15585, restore when data for es is fleshed out"

does it still need to be disabled, or has es been fleshed out? how would it get fleshed out?

srl295:
> > Can you update the unit test also? TestLocaleCompletion
> 
> there's a comment: "TODO Temporarily disabled for [CLDR-15585](https://unicode-org.atlassian.net/browse/CLDR-15585), restore when data for es is fleshed out"
> 
> does it still need to be disabled, or has es been fleshed out? how would it get fleshed out?

Ouch. I thought that had to do with limited submission locales, let me check.

srl295:
> > > Can you update the unit test also? TestLocaleCompletion
> > 
> > 
> > there's a comment: "TODO Temporarily disabled for [CLDR-15585](https://unicode-org.atlassian.net/browse/CLDR-15585), restore when data for es is fleshed out"
> > does it still need to be disabled, or has es been fleshed out? how would it get fleshed out?
> 
> Ouch. I thought that had to do with limited submission locales, let me check.

https://github.com/unicode-org/cldr/pull/2038 should re-enable this test.

PR #2038, opened 2022-05-23
CLDR-15662 re-enable TestLocaleCompletion

srl295:
CLDR-15662

### DIFF
--- a/tools/cldr-apps/src/test/java/org/unicode/cldr/web/api/TestLocaleCompletion.java
+++ b/tools/cldr-apps/src/test/java/org/unicode/cldr/web/api/TestLocaleCompletion.java
@@ -21,6 +21,8 @@ import org.opentest4j.MultipleFailuresError;
 import org.unicode.cldr.test.CheckCLDR;
 import org.unicode.cldr.test.TestCache;
 import org.unicode.cldr.test.TestCache.TestResultBundle;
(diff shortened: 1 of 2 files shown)


srl295:
@btangmu  with no code or changes (other than what's needed to run) it fails:

- completion votes should be equal completion total but was  2477/2518 ==> expected: <2518> but was: <2477>
- missing ==> expected: <0> but was: <41>

This probably means that `es.xml` (used as a model in this) is missing 41 items this cycle, hence the comment.

I guess I could fix these numbers, then the test would at least pass until later in BRS.

tools/cldr-apps/src/test/java/org/unicode/cldr/web/api/TestLocaleCompletion.java
```diff
@@ -45,14 +47,22 @@ public class TestLocaleCompletion {
      * Locale to use. We will start with a blank slate.
      */
     private static final String TEST_LOCALE = "es";
+    /**
+     * Baseline number of items missing in the above locale. This is sensitive to data changes.
```
btangmu, line 51: does "data" in this case mean only xml?

srl295: yes, only xml. will update.

PR #2039, opened 2022-05-23
CLDR-8666 Report Status: low level plumbing

srl295:
- a new interface VoteReportStatus intended to be used by VettingViewer and others

CLDR-8666


srl295:
- this way we can call `User.getCompletedReports()` to get an enumset of the completed reports (and a similar function to mark them complete)
- Will add some sort of an interface into VettingViewer to fetch this information for dashboard use, and also the UI for completion.

srl295:
especially check the `VoterReportStatus` class to see if it looks OK

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Reports.java
```diff
@@ -0,0 +1,56 @@
...
+    public static String getSettingsName(ReportId r) {
+        return "report_" + r + "_v" + SurveyMain.getNewVersion();
+    }
+    public static String getSettingsNameAcceptable(ReportId r) {
+        return "report_" + r + "_v" + SurveyMain.getNewVersion() + "_acceptable";
+    }
+
+    public static EnumSet<ReportId> getCompletion(UserSettings settings) {
+        final Set<ReportId> have = new TreeSet<ReportId>();
+        for(ReportId r : ReportId.values()) {
```
btangmu, line 22: needs formatting

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Reports.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+
+import org.unicode.cldr.util.VoterReportStatus.ReportId;
+
+/**
+ * Utilities for managing report status in settings
+ */
+public class Reports {
+    public static String getSettingsName(ReportId r) {
```
btangmu, line 13: could these be private? so far, only called locally

tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestReports.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import java.util.EnumSet;
+
+import org.junit.jupiter.api.Test;
+import org.unicode.cldr.util.VoterReportStatus.ReportId;
+
+public class TestReports {
```
btangmu, line 11: +1 for unit test

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Reports.java
srl295: I'll make them private for now.

btangmu (approved):

PR #2040, opened 2022-05-24
CLDR-15662 LocaleCompletion (third) Meter refactor

btangmu:
-New LocaleCompletionCounter.java to encapsulate actual counting

-Plan: LocaleCompletionCounter.handleOnePath should resemble VettingViewer.handleOnePath

-Fix a few warnings shown by IntelliJ

-Comments

CLDR-15662

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
@@ -1,11 +1,7 @@
 package org.unicode.cldr.web.api;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Logger;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.ws.rs.GET;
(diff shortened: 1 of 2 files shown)


btangmu:
this is only refactoring, in preparation for fixing the bug -- the inner loop has its own new method in the new class: LocaleCompletionCounter.handleOnePath; it should use the same code as VettingViewer.handleOnePath to count "missing" -- unless of course there's a good reason to do it differently...

srl295:
Could you factor out the stfactory part and put it in CLDR-code?

btangmu:
@srl295 how would you feel about merging this as-is, as a first step?

srl295:
> > Could you factor out the stfactory part and put it in CLDR-code?
> 
> good question... compare FileInfo which doesn't depend on STFactory at all... but what to do about this:
> 
> `checkCldr = stFactory.getTestResult(cldrLocale, options)`
> 
> ?
> 
> there's nothing like that in FileInfo that i can find --

but maybe it should.  if DefaultErrorStatus used TestCache it could share the already-run test results instead of having to re-run them itself.

> the strategy i'd like to try next is to study whether the code in VettingViewer.FileInfo.handleOnePath could be shared with LocaleCompletionCounter.handleOnePath -- does that make sense? in what ways do they need to differ?

I think it makes sense.

srl295 (approved):

srl295:
> @srl295 how would you feel about merging this as-is, as a first step?

I was reviewing it (finally) at the same time as you wrote that. So, yes. Just didn't get to read through yet.

btangmu:
thanks!

PR #2045, opened 2022-05-26
CLDR-15665 Add VeC as an organization

AEApple:
CLDR-15665

Add VeC as an organization


AEApple:
Main question I have is whether the org name should be 'VeC - Lengua Veneta' instead of 'VeC — Venetian Language' which is what was listed in the body of the ticket.

macchiati:
Lengua Veneta looks better, good idea

AEApple:
Okay, updated :)

rickmcgowan (approved):

macchiati (approved):

PR #2048, opened 2022-05-28
CLDR-15298 add oc-ES Occitan (Spain)

DavidLRowe:
CLDR-15298

This PR address only oc-ES.


DavidLRowe:
@srl295 I'll need to squash this eventually, but I thought it might be helpful to see what I needed to do in each commit to get the tests to pass, and whether the end result is correct.
- Initially I added oc_ES.xml plus made some changes to supplementalData.xml and likelySubtags.xml
- After that build failed, I realized that there were oc.xml and oc_FR.xml files in seed that needed to move to common
- Then the build failure claimed there was no main exemplar for oc.xml, so I removed the draft="provisional" on the exemplar data in oc.xml (for Occitan in France)

srl295:
Whether it's in seed or common doesn't make as much difference at this point and we can reevaluate at the end of vetting.

macchiati (approved):

PR #2050, opened 2022-05-30
CLDR-8666 reports: update per PR comments, improve API

srl295:
- code cleanup in ReportsDB and VoterReportStatus
- break up long functions in ReportResponse.vue
New module cldrClient:
- use swagger-client to generate dynamic API (no manual URL construction)

CLDR-8666

### DIFF
--- a/tools/cldr-apps/js/src/views/ReportResponse.vue
+++ b/tools/cldr-apps/js/src/views/ReportResponse.vue
@@ -33,7 +33,7 @@
 </template>
 
 <script>
-import * as cldrAjax from "../esm/cldrAjax.js";
+import * as cldrClient from "../esm/cldrClient.js";
 import * as cldrStatus from "../esm/cldrStatus.js";
 import * as cldrText from "../esm/cldrText.js";
 export default {
@@ -50,6 +50,7 @@ export default {
     };
   },
   created: async function () {
+    this.client = await cldrClient.getClient();
     await this.reload();
   },
   computed: {
@@ -67,15 +68,11 @@ export default {
     },
   },
   methods: {
-    async changed() {
-      this.loaded = false;
-      const user = cldrStatus.getSurveyUser();
-      if (!user) {
-        this.error = "Not logged in.";
-        this.loaded = true;
-        return;
-      }
-      switch (this.state) {
+    /**
+     * Map state to 2 fields
+     */
+    setFields(state) {
+      switch (state) {
         case "acceptable":
           this.completed = true;
           this.acceptable = true;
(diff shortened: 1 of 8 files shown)


tools/cldr-apps/js/src/views/ReportResponse.vue
```diff
@@ -90,60 +87,77 @@ export default {
...
         return;
       }
-      const theUrl = `api/voting/reports/users/${user.id}/locales/${this.$cldrOpts.locale}`;
-      return cldrAjax
-        .doFetch(theUrl, {
-          headers: {
-            "X-SurveyTool-Session": this.$cldrOpts.sessionId,
+      try {
+        this.setFields(this.state);
+        await this.client.apis.voting.updateReport(
```
srl295, line 114: @btangmu  is that shorter? This function `voting.updateReport()` is synthesized by the API definition the server side

srl295:
@yumaoka  may recognize the approach

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReportsDB.java
```diff
@@ -67,9 +67,8 @@ public ReportStatus getReportStatus(Integer user, CLDRLocale locale) {
...
                 String.format("SELECT report, completed, acceptable FROM %s WHERE submitter=? AND locale=?", table),
                 user, locale.getBaseName());
-            ResultSet rs = ps.executeQuery();
-        ) {
-            while(rs.next()) {
+            ResultSet rs = ps.executeQuery();) {
+            while (rs.next()) {
                 final String report = rs.getString("report");
                 final Boolean completed = rs.getBoolean("completed");
                 final Boolean acceptable = rs.getBoolean("acceptable");
```
btangmu, line 74: future improvement: constants instead of string literals like "acceptable", for java/js/vue -- more maintanable, since these are common words that also occur in irrelevant contexts

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
```diff
@@ -10,9 +10,7 @@
      * Enumeration for the reports. In order.
      */
     public enum ReportId {
-        datetime,
-        zones,
-        compact,  // aka 'numbers'
+        datetime, zones, compact, // aka 'numbers'
```
btangmu, line 13: supposing we eventually make the naming more consistent, which name is better, "compact" or "numbers"?

tools/cldr-apps/js/src/views/ReportResponse.vue
btangmu: really amazing -- i need to study this new ability to use swagger on the front end!

btangmu (approved):

tools/cldr-apps/js/src/views/ReportResponse.vue
```diff
@@ -90,60 +87,77 @@ export default {
...
+        console.error(e);
+        this.error = e;
+        this.loaded = true;
+      }
+    },
+    async reload() {
+      this.loaded = false;
+      const user = cldrStatus.getSurveyUser();
+      if (!user) {
+        this.error = "Not logged in.";
```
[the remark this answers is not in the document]
srl295: thanks i'll use that.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
srl295: These are the URLs also. I think what we should really do is think about the whole structure of reports and come up with a new system for them.

srl295:
> swagger

this was one of my reasons for going with the openapi spec. Not just the 'ui' browser, but for this client capability. I had kind of forgotten about it.  It was also the reason for going with the session ID as a HTTP header instead of a parameter, was for this simplicity in use. (you will note that the vue script doesn't mention the session anymore).

PR #2051, opened 2022-05-31
CLDR-15628 Handle non-numeric placeholders for personName nameField, add descriptions

pedberg-icu:
CLDR-15628

Handle non-numeric placeholders for personName nameField, add descriptions:
- Change regex for matching placeholders to accommodate the personName nameField style of placeholder
- Add new PlaceholderStatus OPTIONAL("Any specific placeholder is optional (and non-numeric); there must be at least one.")
- For a first pass, when generating placeholder id ignore some of the modifiers to allow one description to handle various modifier combinations.
- Add the placeholder descriptions for {prefix} {given} {given-informal} {given2} {surname} {surname-prefix} {surname-core} {surname2} {suffix} (in any combination with ignored modifiers -allCaps -initialCap -initial -monogram)

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PatternPlaceholders.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PatternPlaceholders.java
@@ -17,7 +17,8 @@ public class PatternPlaceholders {
         DISALLOWED("No placeholders allowed."),//
         REQUIRED("Specific number of placeholders allowed."),//
         LOCALE_DEPENDENT("Some placeholders may be omitted in certain locales"),//
-        MULTIPLE("May have multiple instances of the same placeholder, eg “{0} cats and {0} dogs”.")//
+        MULTIPLE("May have multiple instances of the same placeholder, eg “{0} cats and {0} dogs”."),//
+        OPTIONAL("Any specific placeholder is optional (and non-numeric); there must be at least one.")//
         ;
 
         private final String message;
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/PatternPlaceholders.java
```diff
@@ -106,6 +110,12 @@ public PlaceholderData transform(String source) {
                         } else {
                             example = "";
                         }
+                        // TODO Some normalization of personName namePattern placeholder ids.
+                        // Hack for now, later do something maybe using PersonNameFormatter.ModifiedField
+                        id = id.replace("-allCaps", "");
+                        id = id.replace("-initialCap", "");
+                        id = id.replace("-initial", "");
+                        id = id.replace("-monogram", "");
```
macchiati, line 118: Should include all the modifiers?

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/PatternPlaceholders.java
pedberg-icu: @macchiati No because I have explicit placeholders for {given-informal} {surname-prefix} {surname-core}, per the summary above. I am only ignoring (stripping) the modifiers that can be used with multiple fields.

macchiati (approved):

PR #2054, opened 2022-05-31
CLDR-13263 Merge getConstructedBaileyValue and getBaileyValue

btangmu:
-Remove getConstructedBaileyValue, call getBaileyValue instead

-Remove getWinningValueForVettingViewer, call getWinningValue instead

-In getStringValue and getBaileyValue, get constructed value when appropriate

-New GlossonymConstructor.java to encapsulate language name construction

-New getStringValueWithBaileyNotConstructed prevents getName from using constructed values recursively

-Disallow constructed names that would get error CheckStatus.Subtype.nameContainsYear

-Change TestConstructedBailey to TestConstructedValue, test getConstructedValue

-Delete INHERITANCE_MARKER from syr.xml, since it inherits code es_419

-Delete language type=gsw from uz_Cyrl.xml, duplicates constructed type=de_CH

-Rename isExplicitBailey to topValueIsInheritanceMarker for clarity

-Delete dead code in VoteAPIHelper and elsewhere

-Change method from public to private if possible

-Comments

CLDR-13263

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
@@ -1,0 +1,124 @@
+package org.unicode.cldr.util;
+
+import com.ibm.icu.util.Output;
+import org.unicode.cldr.test.CheckNames;
+
+/**
+ * Automatically construct language names (glossonyms)
+ *
+ * Example: in German (de), for the path
+ *
+ *       //ldml/localeDisplayNames/languages/language[@type="ro_MD"]
+ *
+ * the value "Rumänisch (Republik Moldau)" is automatically constructed based on the code "ro_MD".
+ *
+ * The constructed value is a default if no preferable value is submitted or inherited.
+ * A different (non-constructed) value, such as "Moldauisch", may become the winning
+ * value instead of the constructed value.
+ */
+public class GlossonymConstructor {
+
+    /**
+     * Some paths with this prefix can get automatically constructed values
+     */
+    public static final String PATH_PREFIX = "//ldml/localeDisplayNames/languages/language[@type=\"";
+
+    /**
+     * The code such as "ro_MD" must contain an underscore, otherwise there is no constructed value.
+     * Underscore also serves as a clue for recognizing a value that is from code-fallback; for example,
+     * when the value "ro_MD" is inherited, the underscore implies it's a raw code ("bogus value") and should
+     * be replaced by a constructed value like "Rumänisch (Republik Moldau)"; but when the value "Moldauisch"
+     * (without underscore) is inherited, it should not be replaced by a constructed value
+     */
+    private static final String CODE_SEPARATOR = "_";
+
+    /**
+     * For "pathWhereFound" when the value is constructed.
+     * It is non-null to satisfy TestPathHeadersAndValues.
+     * It should not be treated as an actual path; for example, the
+     * Survey Tool Info Panel should not show a broken "Jump to original" link.
+     */
+    public static final String PSEUDO_PATH = "constructed";
+
+    /**
+     * Is the given path eligible for getting a constructed value?
+     *
+     * @param xpath the given path
+     * @return true if eligible
+     */
+    public static boolean pathIsEligible(String xpath) {
+        return xpath.startsWith(PATH_PREFIX) && xpath.contains(CODE_SEPARATOR);
+    }
+
+    /**
+     * Is the given value bogus, and therefore eligible for getting replaced by a constructed value?
+     *
+     * @param value the given value
+     * @return true if bogus
+     */
+    public static boolean valueIsBogus(String value) {
+        return value == null || value.contains(CODE_SEPARATOR) || CheckNames.matchDigitPattern(value);
+    }
+
+    private final CLDRFile cldrFile;
+
+    public GlossonymConstructor(CLDRFile cldrFile) {
+        this.cldrFile = cldrFile;
+        if (!cldrFile.isResolved()) {
+            throw new IllegalArgumentException("Unresolved CLDRFile in GlossonymConstructor constructor");
+        }
+    }
+
+    /**
+     * Get the constructed value and fill in tracking information about where it was found
+     *
+     * @param xpath the path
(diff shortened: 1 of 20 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Is the given value bogus, and therefore eligible for getting replaced by a constructed value?
+     *
+     * @param value the given value
+     * @return true if bogus
+     */
+    public static boolean valueIsBogus(String value) {
+        return value == null || value.contains(CODE_SEPARATOR) || CheckNames.matchDigitPattern(value);
```
srl295, line 60: odd layering to have a util call a 'test' - matchDigitPattern probably belongs under util ideally

srl295 (approved):

btangmu:
@macchiati I'll wait a bit in case you'd like to review this one as well

tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
```diff
@@ -0,0 +1,124 @@
...
+        return xpath.startsWith(PATH_PREFIX) && xpath.contains(CODE_SEPARATOR);
+    }
+
+    /**
+     * Is the given value bogus, and therefore eligible for getting replaced by a constructed value?
+     *
+     * @param value the given value
+     * @return true if bogus
+     */
+    public static boolean valueIsBogus(String value) {
```
macchiati, line 59: We should be telling whether a value is bogus base on where it is found, not what it looks like. That is, if it comes from root or code fallback, then it's 'bogus'.

It looks like this is being called on the result from cldrFile.getName(type, true, altPicker); instead of guessing, we should enhance that method to return the information we need.

tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
```diff
@@ -0,0 +1,124 @@
...
+            final CLDRFile.SimpleAltPicker altPicker = (alt == null) ? null : new CLDRFile.SimpleAltPicker(alt);
+            final String value = cldrFile.getName(type, true, altPicker);
+            if (!valueIsBogus(value)) {
+                return value;
+            }
+        }
+        return null;
+    }
+
+    private void track(Output<String> pathWhereFound, Output<String> localeWhereFound) {
```
macchiati, line 116: In some cases it is harder to read to have extremely small private methods, because you have to look somewhere else to figure out what is actually happening. I would inline this method, for example.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
btangmu: I completely agree that should the goal. This PR avoids changing cldrFile.getName more than necessary. I think that realistically, any improvements to getName should be in subsequent PRs. This PR is already extensive. It fixes bugs, clears up discrepancies, and encapsulates key functionalities in a new class separate from CLDRFile. Adding changes to getName to this same PR would have too much risk of breaking code that depends on getName, unrelated to constructed language names (in the sense of this ticket). getName is extremely complex, crufty, and tightly coupled. It should be moved into its own new class separate from CLDRFile. That needs to be done very carefully as a task in its own right, not mixed in with other changes.

Tracking where a value is found also cries out for a new class. The way we currently pass around Output<String> pathWhereFound, Output<String> localeWhereFound -- is too cumbersome and ad hoc. We don't do it consistently, and it's not always clear what these strings should be. We do have some classes such as XMLSource.AliasLocation and XMLSource.SourceLocation, that might be the beginnings of something more generally useful. Such a class should encapsulate a notion of "where found" that's general enough to include constructed language names. As it stands, XMLSource doesn't know about constructed language names.

About valueIsBogus specifically, as it stands the test is only applied to values for a very limited set of paths, and the assumptions it makes about values with underscores and digits aren't new with this PR. I think it's relatively safe to merge this PR and follow up with enhancements such as "where found" tracking capability for getName.

tools/cldr-code/src/main/java/org/unicode/cldr/util/GlossonymConstructor.java
btangmu: OK, I'll inline it. Part of my motivation for separating it was anticipation of an improved tracking method with its own class along the lines of XMLSource.AliasLocation but more general.

macchiati (approved):
Ok, the rest can be done in a followup. (including any inlining)

(I disagree about getName. We can have an overload that passes back additional information if the last parameter is not null; that wouldn't disturb any other callers.)

btangmu:
>  ...We can have an overload that passes back additional information if the last parameter is not null...

There are already ten cldr.getName functions:
```
public String getName(int type, String code)
public String getName(int type, String code, Transform<String, String> codeToAlt)
public String getName(String type, String code)
public synchronized String getName(String localeOrTZID)
public synchronized String getName(String localeOrTZID, boolean onlyConstructCompound,
    String localeKeyTypePattern, String localePattern, String localeSeparator)
public synchronized String getName(String localeOrTZID, boolean onlyConstructCompound,
    String localeKeyTypePattern, String localePattern, String localeSeparator,
    Transform<String, String> altPicker) {
public String getName(
        LanguageTagParser lparser,
        boolean onlyConstructCompound,
        Transform<String, String> altPicker,
        String localeKeyTypePattern,
        String localePattern,
        String localeSeparator) {
public String getName(LanguageTagParser lparser,
                      boolean onlyConstructCompound,
                      Transform<String, String> altPicker)
public synchronized String getName(String localeOrTZID, boolean onlyConstructCompound)
public synchronized String getName(String localeOrTZID,
    boolean onlyConstructCompound,
    Transform<String, String> altPicker)
```
These call each other in ways that are too complicated for my brain to disentangle. Adding an eleventh, with an additional parameter or two that can be null or not (like pathWhereFound, localeWhereFound), getting passed around through various versions of getName -- might be effective in the short term; but risky and not good for the maintainability of the code in the long run.

PR #2055, opened 2022-05-31
CLDR-15610 Regenerate Births

macchiati:
CLDR-15610

NOTE: I had held off doing this because I was worried about spacing changes, but when I checked the code,
it does do a DAIP transformation before comparing strings, so we should be ok.


srl295 (approved):
binary data but SGTM if tests pass

PR #2061, opened 2022-06-01
CLDR-15684 ST crashes when accessing PersonNameFormats

btangmu:
-Remove the offending js code which has not worked in a very long time

-Postpone any new implementation of Info Panel explanation of extra attributes

CLDR-15684

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.js
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.js
@@ -86,21 +86,6 @@ function show(str, tr, hideIfLast, fn) {
         placeholderInfo
       );
     }
-    // extra attributes
-    if (
-      theRow.extraAttributes &&
-      Object.keys(theRow.extraAttributes).length > 0
-    ) {
-      var extraHeading = cldrDom.createChunk(
-        cldrText.get("extraAttribute_heading"),
-        "h3",
-        "extraAttribute_heading"
-      );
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/js/src/esm/cldrInfo.js
```diff
@@ -86,21 +86,6 @@ function show(str, tr, hideIfLast, fn) {
...
-      theRow.extraAttributes &&
-      Object.keys(theRow.extraAttributes).length > 0
-    ) {
-      var extraHeading = cldrDom.createChunk(
-        cldrText.get("extraAttribute_heading"),
-        "h3",
-        "extraAttribute_heading"
-      );
-      var extraContainer = cldrDom.createChunk("", "div", "extraAttributes");
-      appendExtraAttributes(extraContainer, theRow);
```
srl295, line 100: shouldn't it just be `cldrSurvey.appendExtraAttributes` ?

btangmu:
@srl295 this PR removed the only reference to this in cldrText.js:
```
extraAttribute_heading:
    "Note: there are additional specifiers for this item. Read the help page for further details.",
```

Also `extraAttribute_desc` seems to have been dead code already -- should I follow up by deleting those?

Since you wrote "...duplicate of what's already shown in purple on the main row... OK to leave as is ..." is there anything else that needs to happen for this ticket?

srl295:
@btangmu  OK to delete, because yes there is no additional information provided by that code.  If there was some kind of deep analysis that would be different.

PR #2062, opened 2022-06-01
CLDR-15676 Cannot do a bulk upload with the Survey Tool 42

btangmu:
-The field named stuff must be public since it is referenced by jsp

CLDR-15676


btangmu:
I haven't confirmed yet that bulk upload works with this change; at least this is a necessary condition, as Steven pointed out

srl295:
someday that should eb an automated test case :/

PR #2064, opened 2022-06-01
CLDR-13263 Constructed-value-related clean-up

btangmu:
-Move matchDigitPattern into util/RegexUtilities

-Remove GlossonymConstructor.track, inline the code instead

-Move ten cldr.getName functions and related subroutines/data to bottom of CLDRFile.java in preparation for refactoring as separate class

-Comments

CLDR-13263

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckNames.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckNames.java
@@ -5,10 +5,9 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
-import org.unicode.cldr.util.PatternCache;
+import org.unicode.cldr.util.RegexUtilities;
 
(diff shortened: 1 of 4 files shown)


btangmu:
@macchiati @srl295 the diff for CLDRFile looks worse than it is; in fact there is no change in CLDRFile other than moving the getName-specific parts to the bottom of the file.

srl295:
> @macchiati @srl295 the diff for CLDRFile looks worse than it is; in fact there is no change in CLDRFile other than moving the getName-specific parts to the bottom of the file.

why is the diff so 'bad' then? If it's just reordering functions is that worthwhile?

btangmu:
Yes, it does seem worthwhile to me since it prepares the way to encapsulate and improve getName. Otherwise, I'd be at a dead end --  I feel it would be irresponsible for me to start making changes in code that's so entangled, without first untangling it.

macchiati (changes requested):
You've clearly done a lot of good work here, but when you combining refactoring with substantial changes, it is nearly impossible to tell what the substantial changes are. Would it be possible for you to separate those into two separate PRs?

PR #2066, opened 2022-06-01
CLDR-15691 fix sampleName field order in ST; remove surname from full if we have surname-prefix/-suffix

pedberg-icu:
CLDR-15691

- In ST display of sampleName fields for full sample, fix ordering of surname-prefix vs surname-core
- In full sample data, remove surname field if we have both surname-prefix and surname-core (note that nl_BE does not have surname-prefix and surname-core, just surname)

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -5596,7 +5596,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<nameField type="given">∅∅∅</nameField>
 			<nameField type="given-informal">∅∅∅</nameField>
 			<nameField type="given2">∅∅∅</nameField>
-			<nameField type="surname">∅∅∅</nameField>
 			<nameField type="surname-prefix">∅∅∅</nameField>
 			<nameField type="surname-core">∅∅∅</nameField>
 			<nameField type="surname2">∅∅∅</nameField>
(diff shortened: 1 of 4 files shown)


common/main/root.xml
```diff
@@ -5596,7 +5596,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<nameField type="given">∅∅∅</nameField>
 			<nameField type="given-informal">∅∅∅</nameField>
 			<nameField type="given2">∅∅∅</nameField>
-			<nameField type="surname">∅∅∅</nameField>
```
pedberg-icu, line 5599: Should we leave the surname field in root?

macchiati (approved):
We could go ahead and merge this. (But see my note for a slightly more robust formulation.)

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -1924,8 +1924,8 @@ public String transform(String source) {
                     // The various nameField attribute values: each group in desired
                     // sort order, but groups from least important to most
                     final List<String> attrValues = Arrays.asList(
-                        "informal", // modifiers for nameField type
-                        "prefix", "given", "given2", "surname", "surname-prefix", "surname-core", "surname2", "suffix"); // values for nameField type
+                        "informal", "prefix", "core", // modifiers for nameField type
+                        "prefix", "given", "given2", "surname", "surname2", "suffix"); // values for nameField type
```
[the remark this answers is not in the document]
pedberg-icu: Thanks, will do in next PR

common/main/root.xml
macchiati: No, let's remove there.

PR #2073, opened 2022-06-03
CLDR-8666 fix the swagger client: broken in smoke/prod

srl295:
- affects report view and votring
- CLDR-15702 to revisit this workaround

CLDR-8666

### DIFF
--- a/tools/cldr-apps/src/main/webapp/META-INF/microprofile-config.properties
+++ b/tools/cldr-apps/src/main/webapp/META-INF/microprofile-config.properties
@@ -1,0 +1,2 @@
+mp.openapi.servers=/cldr-apps
+


srl295:
to test:

1. https://cldr-smoke.unicode.org/cldr-apps/v#r_datetime/ab//

should not have a red box: 
image


2. in the Priority Items Summary, clicking Load should work and not show an error
(need to be logged in as TC)
image

image

srl295:
more explanation:

-Makes the default server API path just “/cldr-apps” 
- previously it defaulted to something including the URL, <http://st.unicode.org/cldr-apps> (note incorrect HTTP instead of HTTPS)

macchiati (dismissed):

srl295:
Commit message was wrong. Can i get a re-review?

AEApple (approved):

PR #2076, opened 2022-06-04
CLDR-15353 adjust PK language populations

DavidLRowe:
CLDR-15353

Outstanding request: renaming "lah" from "Lahnda" to "Western Punjabi"

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -355,7 +355,7 @@ annotations.
 			<language type="la">Latin</language>
 			<language type="lad">Ladino</language>
 			<language type="lag">Langi</language>
-			<language type="lah">Lahnda</language>
+			<language type="lah">Western Panjabi</language>
 			<language type="lam">Lamba</language>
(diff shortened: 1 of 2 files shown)


DavidLRowe:
Figures updated in the XML file directly (rather than the raw .txt file). 

The order of some entries changed in order to keep the entries in decreasing order according to percentage of population.

No evidence for "tg_Arab" in PK, so Tajik (Arabic) entry removed.

macchiati (approved):
The changes look good. However, as for the language name change, those would revert if the file is reformatted. For 'lah', best is to change the language name in en.com.

DavidLRowe:
@macchiati change in en.xml (rather than en.com)?

Should I add that to this PR?

macchiati:
Yup, typo

DavidLRowe:
The only change since the last review is that en.xml is now updated to use "Western Panjabi" for "lah".

macchiati (approved):

PR #2077, opened 2022-06-06
CLDR-8666 fix nginx, not microprofile

srl295:
- add X-Forwarded-Proto: header as it should have been
- remove microprofile-config.properties as not needed

CLDR-8666

this fixes CLDR-15702


srl295:
need to test this, but i think i missed this in the nginx config

btangmu (approved):

macchiati (approved):
A rubber stamp, since I'm not familiar with the files.

PR #2078, opened 2022-06-06
CLDR-10720 TestCLDRLinks: fix URL parsing, was too restrictive

srl295:
CLDR-10720

( Note: https://unicode-org.atlassian.net/browse/CLDR-15678 has been updated with  the results of this test.)


AEApple (approved):
Thank you for doing this!

PR #2081, opened 2022-06-07
CLDR-15646 Priority Items: ticket items B/C/D

btangmu:
-New Progress column gets percentage from LocaleCompletion (item B in ticket)

-New VettingViewerQueue.getLocaleToProgress pre-fetches all percentages

-New VettingViewer.setLocaleToProgress, getProgressPercentage

-Change SHOW_SUBTYPES from true to false (item C in ticket)

-New first column Level (item D in ticket)

-Fix an unrelated recent bug in VettingViewer.vue: change tt to code

-Comments

CLDR-15646

### DIFF
--- a/tools/cldr-apps/js/src/views/VettingSummary.vue
+++ b/tools/cldr-apps/js/src/views/VettingSummary.vue
@@ -70,7 +70,7 @@
             :key="locale"
           >
             <td>
-              <tt>{{ locale }}—{{ humanizeLocale(locale) }}</tt>
+              <code>{{ locale }}—{{ humanizeLocale(locale) }}</code>
             </td>
             <td>
               <span
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
@@ -1,21 +1,17 @@
 package org.unicode.cldr.web;
 
 import java.io.IOException;
-import java.util.EnumSet;
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
 import java.util.concurrent.Semaphore;
 import java.util.logging.Logger;
 
 import org.json.JSONException;
 import org.json.JSONObject;
-import org.unicode.cldr.util.CLDRFile;
-import org.unicode.cldr.util.CldrUtility;
-import org.unicode.cldr.util.Organization;
-import org.unicode.cldr.util.VettingViewer;
+import org.unicode.cldr.util.*;
 import org.unicode.cldr.web.CLDRProgressIndicator.CLDRProgressTask;
 
 import com.ibm.icu.dev.util.ElapsedTimer;
+import org.unicode.cldr.web.api.LocaleCompletion;
 
 /**
  * @author srl
@@ -43,7 +39,7 @@ public class VettingViewerQueue {
     /**
      * Count the # of paths in this CLDRFile
      *
-     * @param file
+     * @param f
      * @return
      */
     private static int pathCount(CLDRFile f) {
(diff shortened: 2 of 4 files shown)


tools/cldr-apps/js/src/views/VettingSummary.vue
```diff
@@ -70,7 +70,7 @@
             :key="locale"
           >
             <td>
-              <tt>{{ locale }}—{{ humanizeLocale(locale) }}</tt>
+              <code>{{ locale }}—{{ humanizeLocale(locale) }}</code>
```
btangmu, line 73: we started getting "[Vue warn]: Failed to resolve component: tt at <VettingSummary >" recently -- tt is obsolete in html5 -- see https://www.w3schools.com/tags/tag_tt.asp

tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
```diff
@@ -279,17 +275,43 @@ private void processCriticalWork(final CLDRProgressTask progress) {
...
+            final HashMap<String, String> map = new HashMap<>();
+            final ArrayList<String> localeList = vv.getLocaleList(org);
+            for (String localeId : localeList) {
+                final CLDRLocale loc = CLDRLocale.getInstance(localeId);
+                final LocaleCompletion.LocaleCompletionResponse response = LocaleCompletion.handleGetLocaleCompletion(loc);
+                final int percent = (100 * response.votes / response.total);
+                map.put(localeId, Integer.toString(percent));
+            }
+            return map;
+        }
```
btangmu, line 313: the dubious assumption that "Progress" is same as "Local Completion" is encapsulated here in getLocaleToProgress

Possibly "org" should always be "surveytool" here -- meaning "org-neutral" -- as it is anyway in the case of Snapshots?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -53,7 +53,7 @@
 
     private static final boolean DEBUG = false;
 
-    private static final boolean SHOW_SUBTYPES = true; // CldrUtility.getProperty("SHOW_SUBTYPES", "false").equals("true");
+    private static final boolean SHOW_SUBTYPES = false;
```
btangmu, line 56: per ticket: "C. Priority Items: Remove the detailed Errors/Warnings (leave under a flag, so we can enable if we want later)"

btangmu: there was already a flag, just make it false

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -856,15 +866,36 @@ public int getLocaleCount(Organization org) {
...
                 .append("<tr class='tvs-tr'>")
                 .append(TH_AND_STYLES)
+                .append("Level</th>")
+                .append(TH_AND_STYLES)
                 .append("Locale</th>")
                 .append(TH_AND_STYLES)
-                .append("Codes</th>");
+                .append("Codes</th>")
+                .append(TH_AND_STYLES)
+                .append("Progress</th>");
```
btangmu, line 898: all this html-generating code on the back end is techdebt, not improved in this pr

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1084,7 +1115,7 @@ private void computeOne(int n) {
                 System.out.println("writeAction.compute(" + n + ") - got fileinfo " + name + ": " + localeID);
             }
             try {
-                writeSummaryRow(output, choices, fileInfo.vc.problemCounter, name, localeID);
+                writeSummaryRow(output, choices, fileInfo.vc.problemCounter, name, localeID, level);
```
btangmu, line 1118: Six params for writeSummaryRow is too many, made worse by this pr -- more techdebt to be addressed when moving all the presentation to the front end

tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
btangmu: about my comment here, "this method may be inefficient" -- it could be made twice as fast -- however, there are two assumptions: (1) progress = locale completion and (2) locale completion uses VettingViewer

PR #2084, opened 2022-06-08
CLDR-15714 Change the formal sorting for English to use surname -prefix and -core

macchiati:
CLDR-15714


macchiati:
Don't review until tests pass.

pedberg-icu (approved):

PR #2086, opened 2022-06-08
CLDR-11640 add frr North Frisian

DavidLRowe:
CLDR-11640


DavidLRowe:
[q ß v x y z] are in the main exemplar for "de", but not in the main exemplar for "frr", so I added them to the auxiliary exemplar for "frr".

PR #2087, opened 2022-06-08
CLDR-15671 treat xml: attributes as metadata, should not show in ST

pedberg-icu:
CLDR-15671

In DtdData, treat all xml: attributes as metadata. Hopefully this is enough to make the xml:space attribute not show in ST,
- I don't want to just ignore/discard them entirely, since we want to be able to use the xml:space attribute as an indicator for space trimming in places like DAIP
- This should produce the desired result in XPathParts.getSpecialNondistinguishingAttributes()
- This allows us to remove `@VALUE` from the xml:space attribute in the dtd (and we don't need to replace it with `@METADATA`)

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3171,7 +3171,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT foreignSpaceReplacement ( #PCDATA ) >
 <!ATTLIST foreignSpaceReplacement xml:space (default | preserve) "preserve" >
-    <!--@VALUE-->
 <!ATTLIST foreignSpaceReplacement alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
 <!ATTLIST foreignSpaceReplacement draft (approved | contributed | provisional | unconfirmed) #IMPLIED >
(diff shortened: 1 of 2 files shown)


srl295 (approved):

macchiati:
It doesn't appear to work in smoketest. It is not a problem with PathHeader: I added (locally) the following:
```
    public void TestForeignSpaceReplacement() {
        PathHeader ph = pathHeaderFactory.fromPath("//ldml/personNames/foreignSpaceReplacement[@xxml:space=\"preserve\"]");
        assertEquals("foreignSpaceReplacement SectionId", SectionId.Misc, ph.getSectionId());
        assertEquals("foreignSpaceReplacement PageId", PageId.PersonNameFormats, ph.getPageId());
        assertEquals("foreignSpaceReplacement Header", "AuxiliaryItems", ph.getHeader());
        assertEquals("foreignSpaceReplacement Code", "foreignSpaceReplacement", ph.getCode());
        assertEquals("foreignSpaceReplacement status", SurveyToolStatus.READ_WRITE, ph.getSurveyToolStatus());
    }
```

srl295:
Any update step that had to run after changing the dtd?

pedberg-icu:
I have a follow-on PR #2097 that does update the dtd after running GenerateDtd and adds some more tests to pin down the problem (and also adds prettyPaths for personNames stuff)

PR #2097, opened 2022-06-09
CLDR-15671 xml: attribute tests, explicitly list as @METADATA, prettyPaths for personNames

pedberg-icu:
CLDR-15671

This is a follow-on to #2087 that attempts to do a few more things that should fix the problem with the xml:space=preserve attribute showing in Survey Tool for the foreignSpaceReplacement element.
1. First I re-ran GenerateDtd; this added an explicit `@METADATA` annotation in the dtd for xml:space (even though we set this status in code anyway). This also responds to an issue Steven raised in a comment on #2087.
2. Finally understand what XPathParts.getSpecialNondistinguishingAttributes() is supposed to do (return  the collection of non-distinguishing attribute values that *should* appear with blue background in the Survey Tool left column) so fixed it to exclude xml: attributes from this collection, and added a function description to alleviate future confusion. Also added a comment for CLDRFile.getDistinguishingXPath(), which *should* and *does* remove the xml:space attribute when normalizing a path.
3. Probably not relevant to this issue but needed doing anyway: Added prettyPaths for the personNames items (and for the dateTimeFormat atTime type).

### DIFF
--- a/tools/cldr-apps/src/test/java/org/unicode/cldr/unittest/web/TestXPathTable.java
+++ b/tools/cldr-apps/src/test/java/org/unicode/cldr/unittest/web/TestXPathTable.java
@@ -133,13 +133,15 @@ public class TestXPathTable extends TestFmwk {
             "numbers=hebrew",
             "//ldml/dates/calendars/calendar[@type=\"chinese\"]/dateFormats/dateFormatLength[@type=\"medium\"]/dateFormat[@type=\"standard\"]/pattern[@type=\"standard\"][@alt=\"variant-proposedx333\"][@numbers=\"hanidec\"][@draft=\"true\"]",
             "numbers=hanidec",
+            "//ldml/personNames/foreignSpaceReplacement[@xml:space=\"preserve\"]",
+            "",
 
         };
 
         for (int i = 0; i < xpaths.length; i += 2) {
             String xpath = xpaths[i + 0];
             String expect = xpaths[i + 1];
-            Map<String, String> ueMap = xpt.getUndistinguishingElementsFor(xpath);
+            Map<String, String> ueMap = xpt.getUndistinguishingElementsFor(xpath); // just calls XPathParts.getSpecialNondistinguishingAttributes()
             if (ueMap != null) {
                 logln(xpath + "\n -> " + ueMap.toString() + " expect " + expect);
             } else {
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3171,6 +3171,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT foreignSpaceReplacement ( #PCDATA ) >
 <!ATTLIST foreignSpaceReplacement xml:space (default | preserve) "preserve" >
+    <!--@METADATA-->
 <!ATTLIST foreignSpaceReplacement alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
 <!ATTLIST foreignSpaceReplacement draft (approved | contributed | provisional | unconfirmed) #IMPLIED >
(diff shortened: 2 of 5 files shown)


tools/cldr-apps/src/test/java/org/unicode/cldr/unittest/web/TestXPathTable.java
```diff
@@ -133,13 +133,15 @@ public void TestNonDistinguishing() throws SQLException {
             "numbers=hebrew",
             "//ldml/dates/calendars/calendar[@type=\"chinese\"]/dateFormats/dateFormatLength[@type=\"medium\"]/dateFormat[@type=\"standard\"]/pattern[@type=\"standard\"][@alt=\"variant-proposedx333\"][@numbers=\"hanidec\"][@draft=\"true\"]",
             "numbers=hanidec",
+            "//ldml/personNames/foreignSpaceReplacement[@xml:space=\"preserve\"]",
+            "xml:space=preserve",
```
srl295, line 137: i thought the goal was to NOT have this listed in the results from `getSpecialNondistinguishingAttributes()`

srl295:
`XPathParts.getSpecialNondistinguishingAttributes` is interpreted by SurveyTool as a list of things that _should_ be handled "specially", that is, displayed.  

maybe I'm reading the code and 09523e300501f86e85bb7601388274bcd4c5f117 wrongly. 

Also, in running a local test, I can't find any xpaths *that are marked as visible in SurveyTool by PathHeader* which have specialNonDistinguishingAttributes, including personNames.  Maybe I'm doing something wrong in my test. I was trying to find what else in our data would show up in purple like this.

srl295:
More strange behavior. This is `main` not this PR:   but getFullPath does NOT include xml:space for, say, ja.xml - please note that `xml:space` is only on the root element and not on ja.xml

tools/cldr-apps/src/test/java/org/unicode/cldr/unittest/web/TestXPathTable.java
pedberg-icu: Hmm, you may be right. I was assuming that non-distinguishing attributes should *not* be shown. But I see that the numbers=hanidec attribute shown above is also displayed in blue in the ST: https://st.unicode.org/cldr-apps/v#/zh/Chinese/4c684156dd09b01c

pedberg-icu: So maybe I *do* need to add xml:space to the explicit list in this function, which are things that should be treated as distinguishing even though that are not actually distinguishing. Seems backward but maybe I do not really understand the intent of the function.

pedberg-icu:
> More strange behavior. This is `main` not this PR: but getFullPath does NOT include xml:space for, say, ja.xml - please note that `xml:space` is only on the root element and not on ja.xml

In ja.xml the value does not have space and so is not needed (and as I recall I may have gotten an error when I added it). But anyway I will revise the PR to change the behavior of  XPathParts.getSpecialNondistinguishingAttributes. Thanks for looking at this!!!

srl295:
The getSpecial… function does not affect the distinguishedness of the path
for ST,  just display

pedberg-icu:
> The getSpecial… function does not affect the distinguishedness of the path for ST, just display

@srl295 thanks, I finally get what you have been trying to tell me all along, and now understand what `XPathParts.getSpecialNondistinguishingAttributes()` is supposed to do (and added some documentation to alleviate future confusion). I fixed it to exclude the xml: attributes from the collection of non distinguishing attributes that *should* be displayed. That should finally fix the problem.

macchiati (approved):

pedberg-icu:
And... in smoketest this *does* fix the problem, i.e. with this change the xml:space=preserve attribute is not shown in Survey Tool for the foreignSpaceAttribute

srl295:
> More strange behavior. This is `main` not this PR: but getFullPath does NOT include xml:space for, say, ja.xml - please note that `xml:space` is only on the root element and not on ja.xml

PR #2098, opened 2022-06-10
CLDR-15687 Add test for 8 votes, make changes in coverageLevels.xml

macchiati:
CLDR-15687

This adds a test for the 8 vote locales, according to the following policy.
 * If a locale is in Locales.txt under Organization=Cldr, and has > 2 TC org locales at Modern, then it needs 8 votes.
 * Otherwise it needs 4.

This involved making some small changes to coverageLevels.txt
* setting 3 locales to 4 votes that didn't make the bar: am, ga, ky
* setting some locales that did make the bar to 8 votes: bg, 'is' and the sublocales en_AU, en_GB, es_MX, fr_CA,

And some unrelated cleanup
* dropping 1 warning that isn't relevant anymore
* reformatting the output of another to be more readable.

Review the ticket with hiding whitespace to avoid some noise introduced by re-indenting.

### DIFF
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -36,7 +36,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<approvalRequirement votes="=HIGH_BAR" locales="tr" paths="//ldml/localeDisplayNames/territories/territory\[@type=.CY.\]"/>
 			<approvalRequirement votes="=HIGH_BAR" locales="kea pt_CV" paths="//ldml/numbers/currencies/currency\[@type=.CVE.\]/(symbol|decimal)"/>
 			<!--  established locales - http://cldr.unicode.org/index/process#TOC-Draft-Status-of-Optimal-Field-Value -->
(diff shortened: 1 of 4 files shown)


macchiati:
Adding some reviewers to give visibility to the changes in the 8 vote status.

AEApple:
Can we consider adding some way in organizations.java to signal that an organization is a TC org so that we can just update that in one place if orgs start attending the TC meetings regularly or stop sending representation?

macchiati:
> Can we consider adding some way in organizations.java to signal that an organization is a TC org so that we can just update that in one place if orgs start attending the TC meetings regularly or stop sending representation?

That is in Organization.java, at https://github.com/unicode-org/cldr/blob/109ccb318e8d46ec93c9bbc40c54e5454b13d37e/tools/cldr-code/src/main/java/org/unicode/cldr/util/Organization.java#L72

However, I do need to change the test at TestCldrSuperset() to adapt to the change in the TCs. I can do that, but not immediately, so I'll file a ticket.

macchiati:
Filed as: https://unicode-org.atlassian.net/browse/CLDR-15726

pedberg-icu (approved):

PR #2100, opened 2022-06-10
CLDR-15646 Priority Items Summary: improve snapshot id, subheaders

btangmu:
-Display snapshot id like 2022-05-16 08:15 GMT instead of 2022-05-16T08:15:25.083077935Z

-Fix the subheader lines, they are supposed to be between different first letters of locales

CLDR-15646

### DIFF
--- a/tools/cldr-apps/js/src/views/VettingSummary.vue
+++ b/tools/cldr-apps/js/src/views/VettingSummary.vue
@@ -41,7 +41,7 @@
               title="Show the indicated snapshot of the Priority Items Summary"
               @click="showSnapshot(snapshotId)"
             >
-              {{ snapshotId }}
+              {{ humanizeSnapId(snapshotId) }}
             </button>
           </span>
         </span>
@@ -181,7 +181,7 @@ export default {
         return null;
       }
       if (cldrPriorityItems.snapshotIdIsValid(snapshotId)) {
-        return "Snapshot " + snapshotId;
+        return "Snapshot " + this.humanizeSnapId(snapshotId);
       } else if (this.canUseSnapshots) {
         return "Latest (not a snapshot)";
       } else {
@@ -230,6 +230,19 @@ export default {
 
     humanizeLocale(locale) {
       return cldrLoad.getLocaleName(locale);
+    },
+
+    /**
+     * Display a more user-friendly snapshot id
+     *
+     * @param id like “2022-05-16T08:15:25.083077935Z”
+     * @return like “2022-05-16 08:15 GMT”
+     */
+    humanizeSnapId(id) {
+      if (!id.match(/^\d+\-\d+\-\d+T\d\d:\d\d:\d\d\..+Z$/)) {
+        return id;
+      }
+      return id.replace("T", " ").replace(/:\d\d\..+Z/, " GMT");
     },
 
     reportClass(kind) {
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
@@ -995,13 +995,12 @@ public class VettingViewer<T> {
                 if (DEBUG_THREADS) {
                     System.out.println("Appending " + name + " - " + outputs[n].length());
                 }
-                output.append(outputs[n]);
-
                 char nextChar = name.charAt(0);
                 if (lastChar != nextChar) {
                     output.append(this.header);
                     lastChar = nextChar;
                 }
+                output.append(outputs[n]);
             }
         }
 


tools/cldr-apps/js/src/views/VettingSummary.vue
```diff
@@ -232,6 +232,19 @@ export default {
...
+     *
+     * @param id like “2022-05-16T08:15:25.083077935Z”
+     * @return like “2022-05-16 08:15 GMT”
+     */
+    humanizeSnapId(id) {
+      if (!id.match(/^\d+\-\d+\-\d+T\d\d:\d\d:\d\d\..+Z$/)) {
+        return id;
+      }
+      return id.replace("T", " ").replace(/:\d\d\..+Z/, " GMT");
+    },
```
btangmu, line 246: this implementation on the front end seemed slightly easier (for me) than using ICU to format the id on the back end; also it allows for more than one id per minute (though they're displayed the same); also it allows for further customization of the displayed id (e.g., timezone/locale-specific) without changing the underlying format in the database and http requests/responses

btangmu:
After this PR, there remain two tasks for this ticket:
* Revise the Progress column and the locale progress (3rd) meter to be relative to baseline (maybe 2 Progress columns?)
* Omit locales that are generated (sr_Latn)

Question: how to recognized generated locales? Is there a function like cldrLocale.isGenerated()?

btangmu:
@srl295 something's wrong with github: I requested your review -- it says "btangmu requested review from macchiati and srl295 36 minutes ago" -- but you're not displayed under reviewers, only Mark is!

tools/cldr-apps/js/src/views/VettingSummary.vue
```diff
@@ -232,6 +232,19 @@ export default {
...
+     * Display a more user-friendly snapshot id
+     *
+     * @param id like “2022-05-16T08:15:25.083077935Z”
+     * @return like “2022-05-16 08:15 GMT”
+     */
+    humanizeSnapId(id) {
+      if (!id.match(/^\d+\-\d+\-\d+T\d\d:\d\d:\d\d\..+Z$/)) {
+        return id;
+      }
+      return id.replace("T", " ").replace(/:\d\d\..+Z/, " GMT");
```
srl295, line 245: you know, there's this CLDR thing…  😄 

```suggestion
    return new Date(id).toLocaleString();
```

```shell
> new Date('2022-05-16T08:15:25.083077935Z').toLocaleString()
'5/16/2022, 3:15:25 AM'
```

tools/cldr-apps/js/src/views/VettingSummary.vue
srl295: format on the front end. Display the raw ID as a mouseover if needed.

srl295 (changes requested):

tools/cldr-apps/js/src/views/VettingSummary.vue
```diff
@@ -41,7 +41,7 @@
               title="Show the indicated snapshot of the Priority Items Summary"
```
srl295, line 41: ^ the title here could contain the snapshotId

srl295 (changes requested):

srl295:
> After this PR, there remain two tasks for this ticket:
> 
> * Revise the Progress column and the locale progress (3rd) meter to be relative to baseline (maybe 2 Progress columns?)
> * Omit locales that are generated (sr_Latn)
> 
> Question: how to recognized generated locales? Is there a function like cldrLocale.isGenerated()?

maybe something like this

```js
const {
 readonly,  // if readonly
 readonly_in_limited, // if readonly BECAUSE OF limited submission
 special_type, // 'readonly', 'scratch'
 special_comment, // "The English locale is maintained by the CLDR-TC…" or "Serbian Latin is generated from Serbian Cyrillic"
special_comment_raw, // "@sr_Latn is generated from @sr_Cyrl" (no keyword substitution)
} =  cldrLocale.getTheLocaleMap().getLocaleInfo('sr_Latn');
```

tools/cldr-apps/js/src/views/VettingSummary.vue
btangmu: that's a good option -- the ticket called for GMT, should we change that?

it will be more robust to keep a fallback (keep id unchanged) if id isn't a date or doesn't match the expected pattern

tools/cldr-apps/js/src/views/VettingSummary.vue
btangmu: good idea

tools/cldr-apps/js/src/views/VettingSummary.vue
srl295: can use the following then (will still use preferred format
```shell
> new Date('2022-05-16T08:15:25.083077935Z').toLocaleString([], {timeZone: 'UTC', timeZoneName: 'short'})
'5/16/2022, 8:15:25 AM UTC'
```

or if you want to lock down the format

```shell
> new Date('2022-05-16T08:15:25.083077935Z').toLocaleString('en-US', {timeZone: 'UTC', timeZoneName: 'short'})
'5/16/2022, 8:15:25 AM UTC'
```

btangmu:
@srl295 "I thought we wanted to mark ones that were generated" -- the ticket says "Omit", not "mark" -- I'm happy to implement whatever people decide -- for "omit" it makes sense to skip them on the back end, resulting in less code and data; for "mark" it would be more of a front-end thing...

btangmu:
@mark @srl295 @AEApple @pedberg-icu please advise, should generated locales be "omitted" from Priority Items Summary, or "marked"? (The ticket says "omit" however it's unclear whether that was the latest consensus)

btangmu:
also, any consensus or preferences about GMT versus something localized? the ticket says "GMT" but it's easily changed -- personally dates like 01/02/2022 drive me crazy since in the US that's usually Jan 2 while in Canada it's Feb 1, though you really just never know

btangmu:
or UTC instead of GMT as suggested by @srl295 ?

how would you feel about merging this PR as-is, with a follow-up soon to address the suggested further improvements? 

we've got a push to production scheduled in just over an hour, right?

srl295:
> also, any consensus or preferences about GMT versus something localized? the ticket says "GMT" but it's easily changed -- personally dates like 01/02/2022 drive me crazy since in the US that's usually Jan 2 while in Canada it's Feb 1, though you really just never know

There are two separate questions:
- UTC time vs local time zone?
- fixed format vs user's local format?

srl295 (approved):
fine to include in currrent push

macchiati:
Ticket says omit, but we agreed in meeting to have extra status column, eg

TC
GEN // for generated
HC // for non. TC high coverage

srl295:
CLDR-9943 for GMT vs UTC.  GMT is an ambiguous id.

macchiati:
Ok to merge this and follow up.

PR #2101, opened 2022-06-10
CLDR-15357 LocaleCompletion: add baseline count to back end

srl295:
CLDR-15357 

It's an incremental step.

`GET /api/completion/locale/es` for example now returns:

```json
{
  "baselineCount": 6,
  "error": 0,
  "level": "MODERN",
  "missing": 7,
  "provisional": 0,
  "total": 8653,
  "votes": 8632
}
```

That `6` is missing+error+provisional from HEAD (current .xml).

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletionCounter.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletionCounter.java
@@ -15,8 +15,10 @@ public class LocaleCompletionCounter {
     private final LocaleCompletion.LocaleCompletionResponse lcr;
     private final VettingViewer<Organization> vv;
     private final VettingViewer.DashboardArgs args;
+    private final boolean isBaseline;
 
     public LocaleCompletionCounter(CLDRLocale cldrLocale, STFactory stFactory) {
+        this.isBaseline = false;
         localeId = cldrLocale.toString();
         level = StandardCodes.make().getTargetCoverageLevel(localeId);
         lcr = new LocaleCompletion.LocaleCompletionResponse(level);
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/VettingViewerQueue.java
```diff
@@ -305,7 +305,7 @@ private HashMap<String, String> getLocaleToProgress(VettingViewer<Organization>
             final ArrayList<String> localeList = vv.getLocaleList(org);
             for (String localeId : localeList) {
                 final CLDRLocale loc = CLDRLocale.getInstance(localeId);
-                final LocaleCompletion.LocaleCompletionResponse response = LocaleCompletion.handleGetLocaleCompletion(loc);
+                final LocaleCompletion.LocaleCompletionResponse response = LocaleCompletion.getLocaleCompletion(loc);
```
srl295, line 308: @btangmu this was calling the uncached worker function before. Perhaps because the permissions were backwards.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletionCounter.java
```diff
@@ -28,9 +30,25 @@ public LocaleCompletionCounter(CLDRLocale cldrLocale, STFactory stFactory) {
...
+        localeId = cldrLocale.toString();
+        level = StandardCodes.make().getTargetCoverageLevel(localeId);
+        lcr = new LocaleCompletion.LocaleCompletionResponse(level);
+        final SurveyMain sm = CookieSession.sm;
+        vv = new VettingViewer<>(sm.getSupplementalDataInfo(), factory, new STUsersChoice(sm));
+        final EnumSet<VettingViewer.Choice> set = VettingViewer.getLocaleCompletionCategories();
+        args = new VettingViewer.DashboardArgs(set, cldrLocale, level);
+        args.setUserAndOrganization(0, VettingViewer.getNeutralOrgForSummary());
+        Dashboard.setFilesForBaseline(args, cldrLocale, factory);
+    }
```
btangmu, line 47: a future improvement would be to share more code between the two constructors  -- it's only the first and last lines that differ, right? seems like the first constructor should just call the second one with isBaseline false

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletionCounter.java
srl295: that's true. 

There's a lot of code churn here now (in fact, merge conflicts during development) so I went for the copy-paste simpler option first

PR #2105, opened 2022-06-13
CLDR-15722 Add alt for NZ country name

AEApple:
CLDR-15722

- Add alt for NZ country name
- Make new item appear as missing in all locales.


pedberg-icu (approved):

macchiati:
The result of the conversation was to add "Aotearoa" as an alt, not "Aotearoa New Zealand". Can you create another PR?

stenshamn:
@macchiati , that's not what I recall, nor what the sources in https://unicode-org.atlassian.net/browse/CLDR-15722 indicated. Official sources have started using "Aotearoa New Zealand", not just the Māori "Aotearoa" in isolation.

PR #2108, opened 2022-06-14
CLDR-15706 Add Hinglish variant

AEApple:
CLDR-15706 Add Hinglish variant

- Add Hinglish variant
- Make visible in to all languages

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -258,6 +258,8 @@ annotations.
 			<language type="hdn">Northern Haida</language>
 			<language type="he">Hebrew</language>
 			<language type="hi">Hindi</language>
+			<language type="hi_Latn">Hindi (Latin)</language>
+			<language type="hi_Latn" alt="variant">Hinglish</language>
 			<language type="hif">Fiji Hindi</language>
 			<language type="hil">Hiligaynon</language>
 			<language type="hit">Hittite</language>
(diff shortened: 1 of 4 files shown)


btangmu:
Test description:
```
    public void TestCoverageCompleteness() {
        /**
         * Check that English paths are, except for known cases, at least modern coverage.
         * We filter out the things we know about and have determined are OK to be in comprehensive.
         * If we add a path that doesn't get its coverage set, this test should complain about it.
         */
```

btangmu:
I went to https://st.unicode.org/cldr-apps/v#lookup///
and entered //ldml/localeDisplayNames/languages/language[@type="hi_Latn"]
and it says "❌ Not Found"

btangmu:
@AEApple in TestCoverageLevel.java right below this line
```
        for (String path : english.fullIterable()) {
```

if this were inserted

```
            if (path.equals("//ldml/localeDisplayNames/languages/language[@type=\"hi_Latn\"]") ||
                    path.equals("//ldml/localeDisplayNames/languages/language[@type="\hi_Latn\"][@alt=\"variant\"]")) {
                continue;
            }
```

then I think the test would be skipped for those two paths

I'm not familiar with this part of the code though, and I'm not it's the best workaround -- at least for a draft pr there's no harm in trying it

AEApple:
Okay, thank you!

Do you know if I need to include the standard name for hi-Latn in order to add an alt? I had added it since I didn't see any languages with composed names which have alts currently.

macchiati (approved):

PR #2111, opened 2022-06-14
CLDR-15662 LocaleCompletion (third) Meter: use baselineCount

btangmu:
-Change how the front-end 3rd meter calculates percentage, using baselineCount

-Percentage cannot be less than 0 or greater than 100

CLDR-15662

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrProgress.js
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.js
@@ -188,12 +188,21 @@ function refreshVoterMeter() {
 }
 
 function refreshLocaleMeter() {
-  if (localeProgressStats) {
-    progressWrapper?.updateLocaleMeter(
+  if (localeProgressStats && progressWrapper) {
+    const baselineProblems = localeProgressStats.baselineCount;
+    const currentProblems =
+      localeProgressStats.error +
+      localeProgressStats.missing +
+      localeProgressStats.provisional;
+    const solvedProblems =
+      currentProblems >= baselineProblems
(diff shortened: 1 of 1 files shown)


btangmu:
this changes the 3rd meter

it doesn't change Priority Items Summary "Progress" column yet -- that will be next

tools/cldr-apps/js/src/esm/cldrProgress.js
```diff
@@ -188,12 +188,21 @@ function refreshVoterMeter() {
...
+      currentProblems >= baselineProblems
+        ? 0
+        : baselineProblems - currentProblems;
+    progressWrapper.updateLocaleMeter(
       new MeterData(
         cldrText.get("progress_all_vetters"),
-        localeProgressStats.votes,
-        localeProgressStats.total,
+        solvedProblems,
+        baselineProblems,
```
btangmu, line 205: I'm thinking we should switch to a better terminology for all 3 meters: "done" and "total".

Here solvedProblems would be "done", and baselineProblems would be "total".

For voter completion and page completion, votes would be "done" and total would be "total".

macchiati: sounds good.

macchiati (commented):
Looks good, but I suggest that we don't integrate until @srl295 has a chance to look at the baseline issue (see email).

btangmu:
See https://github.com/unicode-org/cldr/pull/2115 which combines these 3rd-meter changes with changes for Priority Items Summary

PR #2112, opened 2022-06-15
CLDR-15739 allow vote for none in NameOrder

macchiati:
CLDR-15739


srl295 (approved):
looks ok- i'm not sure that voting for an empty string works everywhere else, though.  I think that is usually disallowed.

macchiati:
Right, I'll need to try this out in smoketest to make sure it works. I'm going to be driving for the next hour, so will check it out when I get there.

PR #2115, opened 2022-06-15
CLDR-15646 Use baselineCount for 3rd meter and summary Progress column

btangmu:
-Fix baselineCount by using a resolved CLDRFile for sourceFile

-Throw exception in generateLocaleCompletion unless sourceFile.isResolved

-Change calculation of locale completion percentage to use baselineCount

-Percentage cannot be less than 0 or greater than 100

-New interface LocaleBaselineCount, implemented by VVQueueLocaleBaselineCount

-New VettingViewer.localeBaselineCount, with setter

-New static LocaleCompletion.getBaselineCount

-Refactor LocaleCompletionCounter constructors to avoid duplication

-New CompletionPercent.java in util

-Move LocaleCompletionData to its own file LocaleCompletionData.java in util

-Encapsulate formula errors+missing+provisional with method LocaleCompletionData.problemCount

-Minor refactoring; final where possible; reduce public members, use getters/setters

-Java formatting, including some re-ordering of imports

-Remove dead code

-Comments

CLDR-15646

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
@@ -1,8 +1,10 @@
 package org.unicode.cldr.web.api;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import java.time.Duration;
 import java.util.concurrent.ExecutionException;
-
 import javax.enterprise.context.ApplicationScoped;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
@@ -10,7 +12,6 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-
 import org.eclipse.microprofile.metrics.annotation.Counted;
 import org.eclipse.microprofile.metrics.annotation.Timed;
 import org.eclipse.microprofile.openapi.annotations.Operation;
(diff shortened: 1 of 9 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
```diff
@@ -149,47 +143,73 @@ public void valueChanged(String xpath, XMLSource source) {
...
-        // add in baseline
-        lcr.baselineCount = LocaleCompletionHelper.INSTANCE.basecache.get(cldrLocale);
-        return lcr;
+    public static int getBaselineCount(CLDRLocale cldrLocale) {
+        try {
+            return LocaleCompletionHelper.INSTANCE.basecache.get(cldrLocale);
+        } catch (ExecutionException e) {
+            e.printStackTrace();
+            return 0;
+        }
```
srl295, line 162: why?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1304,6 +1267,14 @@ private void writeSummaryRow(Appendable output, EnumSet<Choice> choices, Counter
         output.append("</tr>\n");
     }
 
+    private String getLocaleProgressPercent(String localeId, Counter<Choice> problemCounter) {
+        final LocaleCompletionData lcd = new LocaleCompletionData(problemCounter);
+        final int problemCount = lcd.problemCount();
+        final int total = localeBaselineCount.getBaselineProblemCount(CLDRLocale.getInstance(localeId));
```
srl295, line 1273: should there be a column for the baseline count also?  since this is HTML it could be in a `title=`

srl295 (dismissed):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
btangmu: it saves adding "throws ExecutionException" to every caller -- if it's problematic, though, I'll change it

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: I can't really judge the value of adding another column -- I'm just trying to implement what's called for in the ticket for now

btangmu:
@macchiati @srl295 my understanding is this PR may need to wait for changes in how baselineCount is calculated...

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleCompletion.java
srl295: Hm, ok i see you are calling getBaselineCount in another context.  I still think it should propagate, because the information isn't available if something goes wrongfetching

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
macchiati: I agree that the baseline count would be useful, but that can come later; shouldn't interfer with this PR

macchiati (approved):

btangmu:
@srl "still think it should propagate" -- I just revised this pr so that the ExecutionException does propagate

macchiati (dismissed):

btangmu:
the latest commit makes sourceFile.isResolved true, resulting in more consistency between baseline and non-baseline problem counts

btangmu:
@macchiati it looks like you approved right before my last change, so this needs a review again

macchiati (approved):

srl295 (commented):
post merge lgtm

PR #2118, opened 2022-06-17
CLDR-15646 Prevent baselineCount from depending on votes

btangmu:
-New VettingViewer.EmptyUsersChoice implements UsersChoice

-Use EmptyUsersChoice instead of STUsersChoice if isBaseline

CLDR-15646

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
@@ -263,6 +263,34 @@ public class VettingViewer<T> {
         VoteResolver<String> getVoteResolver(CLDRLocale loc, String path);
     }
 
+    public static class EmptyUsersChoice implements UsersChoice<Organization> {
+        final PathHeader.Factory phf = PathHeader.getFactory();
+
+        @Override
+        public String getWinningValueForUsersOrganization(CLDRFile cldrFile, String path, Organization org) {
+            CLDRLocale loc = CLDRLocale.getInstance(cldrFile.getLocaleID());
+            return getVoteResolver(loc, path).getOrgVote(org);
+        }
+
+        @Override
+        public VoteStatus getStatusForUsersOrganization(CLDRFile cldrFile, String path, Organization org) {
+            CLDRLocale loc = CLDRLocale.getInstance(cldrFile.getLocaleID());
+            return getVoteResolver(loc, path).getStatusForOrganization(org);
+        }
+
+        @Override
+        public VoteResolver<String> getVoteResolver(final CLDRLocale loc, final String path) {
+            VoteResolver<String> r = new VoteResolver<>();
+            r.setLocale(loc, phf.fromPath(path));
+            return r;
+        }
+
+        @Override
+        public boolean userDidVote(int userId, CLDRLocale loc, String path) {
+            return false;
+        }
+    }
+
     public interface ErrorChecker {
         enum Status {
             ok, error, warning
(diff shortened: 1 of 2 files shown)


btangmu:
@macchiati @srl295 this results in baselineCount 8611 for fr, which I imagine is wrong -- maybe this draft PR is useful anyway for discussion

btangmu:
See our slack discussion for context

This probably still needs fixing:
```
public VoteStatus getStatusForUsersOrganization(CLDRFile cldrFile, String path, Organization orgOfUser) {
    CLDRLocale loc = CLDRLocale.getInstance(cldrFile.getLocaleID());
    BallotBox<User> ballotBox = getBox(sm, loc);
    return ballotBox.getResolver(path).getStatusForOrganization(orgOfUser);
}
```

there's no connection between the cldrFile (disk-based for baseline) and ballotBox, other than being for the same locale -- so here's one likely place that the votes are consulted when they shouldn't be

srl295:
Good catch

srl295:
Maybe it should be a constant VoteStatus.ok_novotes or provisionalOrWorse
here depending on path status in git?

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -263,6 +263,34 @@ public enum VoteStatus {
...
+        @Override
+        public VoteStatus getStatusForUsersOrganization(CLDRFile cldrFile, String path, Organization org) {
+            CLDRLocale loc = CLDRLocale.getInstance(cldrFile.getLocaleID());
+            return getVoteResolver(loc, path).getStatusForOrganization(org);
+        }
+
+        @Override
+        public VoteResolver<String> getVoteResolver(final CLDRLocale loc, final String path) {
+            VoteResolver<String> r = new VoteResolver<>();
+            r.setLocale(loc, phf.fromPath(path));
```
btangmu, line 284: probably the VoteResolver needs additional initialization

btangmu: compare: STFactory.PerLocaleData#getResolverInternal, which calls setUsingKeywordAnnotationVoting, setBaseline, setBaileyValue

btangmu: setUsingKeywordAnnotationVoting can be simplified to be entirely private within VoteResolver, so it won't be necessary to duplicate that code in EmptyUsersChoice

setBaileyValue might be problematic since EmptyUsersChoice#getVoteResolver doesn't have the CLDRFile

unfortunately, setBaseline is very problematic since the corresponding code in STFactory has complex private dependencies:

```
            final ValueChecker vc = ERRORS_ALLOWED_IN_VETTING ? null : new ValueChecker(path);
            ...
            final String currentValue = diskData.getValueAtDPath(path);
            final Status currentStatus = calculateStatus(diskFile, diskFile, path);
            if (ERRORS_ALLOWED_IN_VETTING || vc.canUseValue(currentValue)) {
                r.setBaseline(currentValue, currentStatus);
                r.add(currentValue);
            }
```

btangmu: aside from having zero votes, we need EmptyUsersChoice to act the same as STUsersChoice -- otherwise it's meaningless to compare current stats with baseline

btangmu:
See https://github.com/unicode-org/cldr/pull/2119 and https://github.com/unicode-org/cldr/pull/2120

PR #2122, opened 2022-06-21
CLDR-15646 Make baselineCount more consistent with non-baseline

btangmu:
-Treat constructed values as missing in VoteResolver.calculateStatus

-In VotelessUsersChoice set baseline value using xmlSource.isHere. getValueAtDPath

CLDR-15646


btangmu:
This draft PR was just for discussion. I'm closing it now. So far, per discussion:
* the change for constructed values isn't needed or wanted
* baseline already matches non-baseline with respect to constructed values
* the change to use xmlSource, affecting the example "alt" path, may or may not be appropriate -- some change is needed though to prevent arbitrary differences between baseline and non-baseline algorithms making the percent=100*done/total arbitrary

PR #2123, opened 2022-06-22
CLDR-15749 Add extra paths for ca, es, gl names

macchiati:
CLDR-15749

As per discussions in person-names WG.
* These are added as alt='1' and provisional, so vetters will need to confirm the value and change as needed.


richgillam (approved):
Does this mean these extra entries now show up in the Survey Tool, where the translators can edit them?  Or is there some other process at work here?

macchiati:
Yes, they will show up in the survey tool, marked as provisional. So the vetters can edit from there (and confirm the result).

richgillam:
>Yes, they will show up in the survey tool, marked as provisional. So the vetters can edit from there (and confirm the result).

Great!  Thanks for confirming that.  I wasn't sure how this process was supposed to work.

PR #2127, opened 2022-06-23
CLDR-15737 Problems with compact number report

macchiati:
CLDR-15737

Add test of problem cases; will fail for now, so just draft.

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCompactNumbers.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCompactNumbers.java
@@ -18,6 +18,7 @@ import org.unicode.cldr.util.VerifyCompactNumbers;
 import com.ibm.icu.text.CompactDecimalFormat;
 import com.ibm.icu.text.CompactDecimalFormat.CompactStyle;
 import com.ibm.icu.text.NumberFormat;
+import com.ibm.icu.util.Currency;
 import com.ibm.icu.util.ULocale;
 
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCompactNumbers.java
```diff
@@ -13,15 +13,19 @@
...
 import com.ibm.icu.text.CompactDecimalFormat;
 import com.ibm.icu.text.CompactDecimalFormat.CompactStyle;
 import com.ibm.icu.text.NumberFormat;
+import com.ibm.icu.text.PluralRules.FixedDecimal;
+import com.ibm.icu.util.Currency;
 import com.ibm.icu.util.ULocale;
 
 public class TestCompactNumbers  extends TestFmwkPlus {
-    static final boolean DEBUG = false;
+    static final boolean DEBUG = true;
```
btangmu, line 28: is it intentional and safe to make DEBUG = true in production?

btangmu (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCompactNumbers.java
btangmu: I don't see any problem with it, just checking it's intentional; it's in a unit test anyway

macchiati:
That test isn't actually run yet. Once we have the ICU fix in, I want to enable the test and fix that.

PR #2128, opened 2022-06-23
CLDR-8666 Implementation for Vote Report Resolver

srl295:
CLDR-8666 Vote Report Resolver and tests

- also make VoterReportStatus abstract instead of interface
- move ReportStatusUpdater into org.unicode.cldr.util (not web specific)
- tests added, but they depend on CLDR-15765 to be safe (so disabled for now)

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReportsDB.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReportsDB.java
@@ -5,9 +5,6 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
@@ -16,9 +13,10 @@ import java.util.logging.Logger;
 
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.unicode.cldr.util.CLDRLocale;
+import org.unicode.cldr.util.ReportStatusUpdater;
 import org.unicode.cldr.util.VoterReportStatus;
 
-public class ReportsDB implements VoterReportStatus<Integer>, ReportStatusUpdater<Integer> {
+public class ReportsDB extends VoterReportStatus<Integer> implements ReportStatusUpdater<Integer>{
     public static ReportsDB getInstance() {
         return ReportsDBHelper.INSTANCE;
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
@@ -1,11 +1,13 @@
 package org.unicode.cldr.util;
 
 import java.util.EnumSet;
+import java.util.Set;
 
 /**
  * This interface is for objects which can expose information on which reports have been completed.
+ * @param T the type for each voter. Must be an Integer.
  */
-public interface VoterReportStatus<T> {
+public abstract class VoterReportStatus<T> {
     /**
      * Enumeration for the reports. In order.
      */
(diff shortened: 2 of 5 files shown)


srl295:
see if this makes some sense

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReportsDB.java
```diff
@@ -5,9 +5,6 @@
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
```
srl295, line 10: unrelated cleanup of imports

srl295 (commented):
Note that the test does pass if enabled, but would conflict with other tests.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
```diff
@@ -1,19 +1,39 @@
 package org.unicode.cldr.util;
 
 import java.util.EnumSet;
+import java.util.Set;
 
 /**
  * This interface is for objects which can expose information on which reports have been completed.
+ * @param T the type for each voter. Must be an Integer.
```
btangmu, line 8: "Must be an Integer" -- then why generic?

srl295: See the note. It was originally generic because it didn't need to be more specific. But voteresolver's user Id is fixed as an int. I was trying to minimize churn by not removing the generic just yet.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
```diff
@@ -35,5 +55,43 @@ public ReportStatus mark(ReportId r, boolean asComplete, boolean asAcceptable) {
...
+     * Update a Resolver for a particular Report. The resolver will be cleared.
+     * Note that T must be an Integer for this to succeed.
+     * @param l locale
+     * @param r which report
+     * @param userList set of users
+     * @param status the source
+     * @param res which
+     * @return resolver
+     */
+    public VoteResolver<ReportAcceptability>
```
btangmu, line 79: ordinarily we have "VoteResolver<String>" where the strings are candidate items...

is the plan to use "VoteResolver<ReportAcceptability>" to resolve votes between "acceptable" and "notAcceptable"?

srl295: yes, that's exactly what it does.  it's a voteresolver applied to an enum.

srl295: it means that all the usual voting rules apply- per-organization, strengths, etc.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
srl295: @btangmu  i added a note to https://unicode-org.atlassian.net/browse/CLDR-15765 to also remove the generic when that ticket goes in (post close).

srl295:
Translating the test case:

- fr: Nobody Votes.
    - Status: missing
- fr: Meta vetter votes ‘acceptable’
    - Status: approved
    - Value: acceptable
- fr: Guest guest votes ‘not acceptable’
    - Status:approved
    - Value: acceptable
- fr: Wikimedia vetter votes ‘not acceptable’
    - Status: approved
    - Value: not acceptable (5 vs 4)
- fr: Guest guest abstains
    - Status: provisional
    - Value: acceptable (arbitrary?)


- ssy: Guest votes notAcceptable
    - Status: unconfirmed
    - Value: not Acceptable


With this change, the dashboard, etc, will be able to calculate and display the 'final' result for each of the reports:
- Missing (nobody has voted)
- Acceptable (with status: approved, provisional, etc…)
- NotAcceptable (with status: approved, provisional, etc…)

macchiati (approved):

PR #2126, opened 2022-06-22
CLDR-15763 set Noto Sans on .cldrValue

srl295:
- remove unusable body {font-face: } from redesign.css
(overriden by antdv2)
- set an explicit font on the .cldrValue class

CLDR-15763

- [ ] This PR completes the ticket.

PR #2130, opened 2022-06-23
CLDR-10674 Survey Tool to display bidi examples in default and RL context; pass 1 currencyFmts

pedberg-icu:
CLDR-10674

This PR adds the infrastructure to display bidi examples in both default and RTL context, and implements it for just currencyFormats. Once we see how that looks, I will create a separate PR to implement it for more cases (units, short date formats?) and also to refine the behavior based on this first pass.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -64,6 +64,7 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.ibm.icu.impl.Row.R3;
 import com.ibm.icu.impl.Utility;
+import com.ibm.icu.lang.UCharacter;
 import com.ibm.icu.text.BreakIterator;
 import com.ibm.icu.text.DateFormat;
 import com.ibm.icu.text.DateFormatSymbols;
@@ -81,6 +82,7 @@ import com.ibm.icu.text.PluralRules.SampleType;
 import com.ibm.icu.text.SimpleDateFormat;
 import com.ibm.icu.text.SimpleFormatter;
 import com.ibm.icu.text.Transliterator;
+import com.ibm.icu.text.UnicodeSet;
 import com.ibm.icu.text.UTF16;
 import com.ibm.icu.util.Calendar;
 import com.ibm.icu.util.Output;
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1810,8 +1869,24 @@ private String handleCurrencyFormat(XPathParts parts, String value) {
         }
 
         double sampleAmount = 1295.00;
-        String example = formatNumber(df, sampleAmount);
-        example = addExampleResult(formatNumber(df, -sampleAmount), example);
+        example = addExampleResult(formatNumber(df, sampleAmount), example, showContexts);
+        example = addExampleResult(formatNumber(df, -sampleAmount), example, showContexts);
+        
+        if (showContexts && !altAlpha && countValue == null) {
```
btangmu, line 1875: suggestion for future improvement: a block this size warrants adding a subroutine

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2379,14 +2465,33 @@ public static String simplify(String exampleHtml) {
...
-            .replace("<span class='cldr_substituted'>", "❬")
-            .replace("</span>", "❭");
+        if (exampleHtml == null) {
+            return null;
+        }
+        if (internal) {
+            return "〖"
+                    + exampleHtml
+                        .replace("", "❬")
+                        .replace("", "❭")
```
btangmu, line 2475: string literals with PUA characters make this code obscure and fragile... not your fault since the code was already like this... still it would be better to use backgroundStartSymbol (already defined) instead of "", and so forth

btangmu (approved):

macchiati (approved):
Looks good. When we have some more time I think we could clean up some tech debt to make it simpler.

srl295:
I would have rejected this. It reverts changes in https://github.com/unicode-org/cldr/pull/2126

btangmu:
"reverts changes in https://github.com/unicode-org/cldr/pull/2126" -- that's a surprise! It doesn't show up under "Files changed"; is that a defect of git(hub) or an indication that we should be setting up PRs differently?

srl295:
> "reverts changes in #2126" -- that's a surprise! It doesn't show up under "Files changed"; is that a defect of git(hub) or an indication that we should be setting up PRs differently?

No, my client must have somehow shown me incorrect data. I had bad data coverage, maybe a client bug with partial loading?  SOrry for the false alarm, everything is OK

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2219,22 +2294,33 @@ private String addTransliteration(String input, String value) {
      * @return string with HTML for the background.
      */
     private String finalizeBackground(String input) {
```
srl295, line 2296: This function shows why the example content should be shipped as *structured data* to the front end.  This is not the best practice for generating HTML.

srl295 (commented):
Changes OK but example generator has some areas for improvement.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Let's carve that off into a separate ticket. I think the product of the Example generator could be changed from a string to a class. The data in the class can be is convenient to ship to the client, but can also be convenient to convert to a purely textual form that we can use in the low-level tests/tools.

pedberg-icu:
> * "Fill=>context: neutral, RTL" looks odd. maybe: Key: neutral, RTL

Good idea, fixed that in #2137

> * I'm guessing we should show both native and ASCII numbers

We already do that by virtue of having different Survey Tool sections for number patterns using different number systems (i.e. the number patterns are specific to one particular number system)

> * Is there a locale that would show the differences between the neutral context and RTL

I will look for a good example.

PR #864, opened 2020-11-30
CLDR-14324 Fix NullPointerException in OutputFileManager.getLocaleTime

btangmu:
-Effectively make permanent CLDR_NOUPDATE=TRUE

-Delete dead code

-Change some items from public to private

-Comments

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14324
- [x] Updated PR title and link in previous line to include Issue number

PR #2132, opened 2022-06-23
CLDR-15765 refactor VoteResolver to fix static init

srl295:
- new class VoterInfoList, replaces hidden statics in VoteResolver
- VoterInfoList must be passed as the first arg of new VoteResolver()
- VoterInfoList is mutable (synchronized)
- Fixup tests to use VoterInfoList
- For SurveyTool, the UserRegistry manages the VoterInfoList along with the underlying
map

CLDR-15765

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
@@ -59,6 +59,7 @@ import org.unicode.cldr.util.StringId;
 import org.unicode.cldr.util.SupplementalDataInfo;
 import org.unicode.cldr.util.UnicodeSetPrettyPrinter;
 import org.unicode.cldr.util.VoteResolver;
+import org.unicode.cldr.util.VoterInfoList;
 import org.unicode.cldr.util.VoteResolver.CandidateInfo;
 import org.unicode.cldr.util.VoteResolver.UnknownVoterException;
 import org.unicode.cldr.util.XMLSource;
@@ -358,9 +359,9 @@ public class ConsoleCheckCLDR {
         if (options[VOTE_RESOLVE].doesOccur) {
             resolveVotesDirectory = CldrUtility.checkValidFile(CLDRPaths.BASE_DIRECTORY + "incoming/vetted/votes/",
                 true, null);
-            VoteResolver.setVoterToInfo(CldrUtility.checkValidFile(CLDRPaths.BASE_DIRECTORY
+            voterInfoList.setVoterToInfo(CldrUtility.checkValidFile(CLDRPaths.BASE_DIRECTORY
                 + "incoming/vetted/usersa/usersa.xml", false, null));
-            voteResolver = new VoteResolver<>();
+            voteResolver = new VoteResolver<>(voterInfoList);
         }
 
         String user = options[USER].value;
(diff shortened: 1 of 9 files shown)


srl295:
If it's too much churn we could hang onto this till close. I'll use it for unit test branches meanwhile.

macchiati:
Probably better to hold off for now.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterInfoList.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.unicode.cldr.util.VoteResolver.Level;
+import org.unicode.cldr.util.VoteResolver.VoterInfo;
+
+public class VoterInfoList {
```
btangmu, line 13: this new class is a huge improvement for separation of concerns!

let's consider similar treatment for VoteResolver.Level and VoteResolver.VoterInfo

tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
```diff
@@ -358,9 +359,9 @@ public static void main(String[] args) throws IOException {
         if (options[VOTE_RESOLVE].doesOccur) {
             resolveVotesDirectory = CldrUtility.checkValidFile(CLDRPaths.BASE_DIRECTORY + "incoming/vetted/votes/",
                 true, null);
-            VoteResolver.setVoterToInfo(CldrUtility.checkValidFile(CLDRPaths.BASE_DIRECTORY
+            voterInfoList.setVoterToInfo(CldrUtility.checkValidFile(CLDRPaths.BASE_DIRECTORY
                 + "incoming/vetted/usersa/usersa.xml", false, null));
```
btangmu, line 363: just curious -- where does usersa.xml come from and what does it look like?

srl295: @btangmu  I haven't been able to find the generation side. Did it get removed? OutputFileManager used to produce it.

srl295: https://github.com/unicode-org/cldr-aux/blob/master/voting/36/usersa/usersa.xml

btangmu: Removed as dead code in https://github.com/unicode-org/cldr/pull/864

srl295: Yes, and its cause of death has been lost in the mists of time.

btangmu: I vaguely remember we discussed it in a meeting and the consensus was that the db backups were sufficient for a history of voting, so we didn't need an xml version. At that time I wasn't aware that console check could read the xml. Now I see we have VOTE_RESOLVE and VoteResolver.setVoterToInfo... setVoterToInfo(String fileName) is only called by console check, so that's also dead unless/until usersa.xml gets resurrected. There's also setVoterToInfo(Map<Integer, VoterInfo> testVoterToInfo) which still has callers...

macchiati (approved):

PR #2133, opened 2022-06-27
CLDR-8666 Reports: Updated API for using VoteResolver 

srl295:
- move MemVoterReportStatus out of tests and back into cldr-code
- add a 'date' field to ReportStatus and MemVoterReportStatus, for date dependent
vote resolution
- add a ReportsDB.clone() function which make a subset copy of the database
This makes it much more performant to update VoteResolvers- one query instead of
thousands

CLDR-8666

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
@@ -1,6 +1,9 @@
 package org.unicode.cldr.web.api;
 
 import java.sql.SQLException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
@@ -22,11 +25,14 @@ import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
 import org.unicode.cldr.util.CLDRLocale;
+import org.unicode.cldr.util.VoteResolver;
 import org.unicode.cldr.util.VoterReportStatus;
+import org.unicode.cldr.util.VoterReportStatus.ReportAcceptability;
 import org.unicode.cldr.util.VoterReportStatus.ReportId;
 import org.unicode.cldr.web.CookieSession;
 import org.unicode.cldr.web.ReportsDB;
 import org.unicode.cldr.web.ReportsDB.UserReport;
+import org.unicode.cldr.web.SurveyMain;
 import org.unicode.cldr.web.UserRegistry;
 
 @Path("/voting/reports")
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/MemVoterReportStatus.java
```diff
@@ -1,19 +1,24 @@
...
-class MemVoterReportStatus<T extends Comparable<T>> extends VoterReportStatus<T> implements ReportStatusUpdater<T> {
+public class MemVoterReportStatus<T extends Comparable<T>> extends VoterReportStatus<T> implements ReportStatusUpdater<T> {
     Map<Pair<T, CLDRLocale>, ReportStatus> data = new HashMap<>();
 
     @Override
     public ReportStatus getReportStatus(T user, CLDRLocale locale) {
         return data.computeIfAbsent(Pair.of(user, locale), k -> new ReportStatus());
     }
 
+    public void markReportComplete(T user, CLDRLocale locale, ReportId r, boolean marked, boolean acceptable, Date date) {
```
btangmu, line 18: six parameters is a lot, might improve with new class for parameter

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
```diff
@@ -38,8 +40,13 @@ static ReportAcceptability fromPair(boolean isComplete, boolean isAcceptable) {
     public static class ReportStatus {
         public EnumSet<ReportId> completed = EnumSet.noneOf(ReportId.class);
         public EnumSet<ReportId> acceptable = EnumSet.noneOf(ReportId.class);
+        public Date date = null;
```
btangmu, line 43: could be private?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
```diff
@@ -108,6 +114,87 @@ public Response getReport(
...
+        public String locale;
+        private Set<ReportVettingResult> reports = new HashSet<ReportVettingResult>();
+        public ReportVettingResult[] getReports() {
+            return reports.toArray(new ReportVettingResult[0]);
+        }
+    }
+
+    public static class ReportVettingResult {
+        // @Schema(description = "raw VoteResolver output")
+        // public String raw;
```
btangmu, line 182: commented out; not needed?

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
srl295: for other voting items we do pass VoteResolver.toString() to the client for debugging how the voting turned out, this could be enabled here also.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoterReportStatus.java
srl295: this class is used more like a struct, but it could be private here. With getters for the other fields as well.

tools/cldr-code/src/main/java/org/unicode/cldr/util/MemVoterReportStatus.java
srl295: could be.  it has only two call sites, though, so I don't expect the API to have wide use.

srl295:
will merge and improve in subsequent PRs

PR #2140, opened 2022-06-28
CLDR-15667 More DAIP adjustments for intvFmtFallback, space in short/narrow AM/PM, ...

pedberg-icu:
CLDR-15667

This is a follow-on to [CLDR-14032: Propose more sophisticated DAIP handling of spacesDONE](https://unicode-org.atlassian.net/browse/CLDR-14032) to refine the space adjustments from that. Here:
- When converting spaces around em-dash intervalFormats to THIN SPACE \u2009, need to also handle the intervalFormatFallback element
- Per the original ticket we changed space between 12-hour time and AM/PM marker 'a' to NNBSP \u202F for Latn, Cyrl, Grek. Two refinements:
    - We should _not_ do that for formats with the long AM/PM marker “aaaa”. There may not be many of those (if any) in CLDR data, but check anyway. (This check is currently a bit awkward)
    - Spaces in short and narrow AM/PM markers (e.g. “a. m.”) should be converted to \u202F NNBSP. Some short and narrow AM/PM markers are not actually very short; perhaps we should skip doing this for those?

To get tests to pass, I also had to fix the intervalFormatFallback data in root, en, and wae to be what DAIP would produce.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
@@ -69,7 +69,7 @@ public class DisplayAndInputProcessor {
         + "numbers/symbols.*|"
         + "numbers/miscPatterns.*|"
         + "numbers/(decimal|currency|percent|scientific)Formats.+/(decimal|currency|percent|scientific)Format.*)");
-    private static final Pattern INTERVAL_FORMAT_PATHS = PatternCache.get("//ldml/dates/.+/intervalFormatItem.*");
+    private static final Pattern INTERVAL_FORMAT_PATHS = PatternCache.get("//ldml/dates/.+/intervalFormat(Item.*|Fallback)");
     private static final Pattern NON_DECIMAL_PERIOD = PatternCache.get("(?<![0#'])\\.(?![0#'])");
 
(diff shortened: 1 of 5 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -1082,8 +1093,14 @@ private String normalizeWhitespace(String path, String value) {
         // Further whitespace adjustments per CLDR-14032
         if ((scriptCode.equals("Latn") || scriptCode.equals("Cyrl") || scriptCode.equals("Grek")) && 
                 HOUR_FORMAT_XPATHS.matcher(path).matches()) {
-            value = AMPM_SPACE_BEFORE.matcher(value).replaceAll("$1\u202Fa");
-            value = AMPM_SPACE_AFTER.matcher(value).replaceAll("a\u202F$1");
+            String test = AMPM_SPACE_BEFORE.matcher(value).replaceAll("$1$2"); // value without a+
+            if (value.length() - test.length() != 4) { // exclude patterns with aaaa
```
pedberg-icu, line 1097: This way of checking for "aaaa" in the original value is a bit awkward. I tried getting the Matcher result and then calling group(n).length() for the appropriate group but kept getting illegal argument exceptions, so I went with this approach.

macchiati (approved):

PR #2142, opened 2022-06-29
CLDR-15746 os_ES, copy missing MMMM ("de MMMM") date formats from es

pedberg-icu:
CLDR-15746

For `oc_ES`, the vetter requested adding some missing `MMMM` date formats (cannot be added in Survey Tool), copying the missing forms from `es` which has them.


DavidLRowe:
Commit message has "os_ES" but locale is actually "oc_ES". Actual data (in oc_ES.xml) is fine.

Is it important to change the commit message? If so, dismiss review.

PR #2144, opened 2022-06-29
CLDR-10564 Logical groups should be in same coverage level

btangmu:
-New unit test TestCheckLogicalGroupings.testSameCoverageLevel

-Revise LogicalGrouping.PathType.RELATIVE, addPaths to separate groups

-Distinct groups for -1, 0, 1 (close to zero) and -3, -2, 2, 3 (far from zero)

CLDR-10564

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/LogicalGrouping.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/LogicalGrouping.java
@@ -77,8 +77,10 @@ public class LogicalGrouping {
 
     /**
      * Return a sorted set of paths that are in the same logical set as the given path
+     *
+     * @param cldrFile the CLDRFile
      * @param path the distinguishing xpath
-     * @param pathType TODO
+     * @param pathTypeOut if not null, gets filled in with the PathType
      *
      * @return the set of paths, or null (to be treated as equivalent to empty set)
      *
@@ -96,13 +98,10 @@ public class LogicalGrouping {
      * Caches: Most of the calculations are independent of the locale, and can be cached on a static basis.
      * The paths that are locale-dependent are /dayPeriods and @count. Those can be computed on a per-locale basis;
      * and cached (they are shared across a number of locales).
-     *
-     * Reference: https://unicode.org/cldr/trac/ticket/11854
      */
     public static Set<String> getPaths(CLDRFile cldrFile, String path, Output<PathType> pathTypeOut) {
         if (path == null) {
             return null; // return null for null path
-            // return new TreeSet<String>(); // return empty set for null path
         }
         XPathParts parts = null;
         PathType pathType = null;
@@ -126,7 +125,6 @@ public class LogicalGrouping {
         if (pathType == PathType.SINGLETON) {
             /*
              * Skip cache for PathType.SINGLETON and simply return a set of one.
-             * TODO: should we ever return null instead of singleton here?
              */
             Set<String> set = new TreeSet<>();
             set.add(path);
@@ -341,19 +339,12 @@ public class LogicalGrouping {
             @Override
             @SuppressWarnings("unused")
             void addPaths(Set<String> set, CLDRFile cldrFile, String path, XPathParts parts) {
-                String fieldType = parts.findAttributeValue("field", "type");
                 String relativeType = parts.findAttributeValue("relative", "type");
                 Integer relativeValue = relativeType == null ? 999 : Integer.valueOf(relativeType);
                 if (relativeValue >= -3 && relativeValue <= 3) { // This is just a quick check to make sure the path is good.
+                    String fieldType = parts.findAttributeValue("field", "type");
                     if (!(nowUnits.contains(fieldType) && relativeValue == 0)) { // Workaround for "now", "this hour", "this minute"
-                        int limit = 1;
-                        if (fieldType != null && fieldType.startsWith("day")) {
-                            limit = 3;
-                        }
-                        for (Integer i = -1 * limit; i <= limit; i++) {
-                            parts.setAttribute("relative", "type", i.toString());
-                            set.add(parts.toString());
-                        }
+                        addRelativePaths(set, parts, relativeValue, fieldType);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


btangmu:
The new test succeeds in failing:

```
[INFO] Running org.unicode.cldr.test.TestCheckLogicalGroupings
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.02 s <<< FAILURE! - in org.unicode.cldr.test.TestCheckLogicalGroupings
[ERROR] testSameCoverageLevel  Time elapsed: 0.013 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
Locale: am
Group size: 7
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-1"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-2"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-3"]
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="0"]
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="1"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="2"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="3"]
 ==> expected: <moderate> but was: <comprehensive>
	at org.unicode.cldr.test.TestCheckLogicalGroupings.testGrouping(TestCheckLogicalGroupings.java:109)
	at org.unicode.cldr.test.TestCheckLogicalGroupings.sameLevel(TestCheckLogicalGroupings.java:97)
	at org.unicode.cldr.test.TestCheckLogicalGroupings.testSameCoverageLevel(TestCheckLogicalGroupings.java:86)
```

Questions:
1. Should we fix the problem with the logical group before merging this PR?
2. If so, how should we fix the logical group: by dividing it into two groups, or by making all the paths moderate, or by making all the paths comprehensive?

pedberg-icu:
Code looks good. Here is my take on the questions:

> 1. Should we fix the problem with the logical group before merging this PR?

Yes.

> 2. If so, how should we fix the logical group: by dividing it into two groups, or by making all the paths moderate, or by making all the paths comprehensive?

In this case I think it should be fixed by dividing the items into two different logical groups.

> 3. The title of the ticket is "Logical groups that generate errors should be in same coverage level" -- is the qualification "that generate errors" significant here? Couldn't any logical group generate an error, if some but not all items are missing?

I do not have an answer for this one.

Once the logical groups are fixed and the new error goes away I can approve PR.

srl295 (dismissed):
approving, test looks good

macchiati:
I think the problem with logical groups is when there are optional members. That is:

macchiati:
A core group might be A (yesterday, today, tomorrow), in English.
In German, you also have B (vorgestern, and übermorgen) // the "day before yesterday", and the "day after tomorrow".
We don't force (say) English speakers to produce terms in B that would be artificial and unnecessary in their language.

The distinction is:

1. If you have any member of A, then you require the other members of A. And raise an error if you don't have them.
2. If you have a member of B, then you need to have the members of A — but _not_ vice versa.

The determination of the logical group allows some members to be optional. It is ok in that case to have any optional members (eg B) at comprehensive, and core group (A) at a lower level. 

So we need to have any test (or data changes) take this asymmetry into account.

- For members of A, all other members need to be at the same level
- For members of B, that's not required.

btangmu:
> It is ok in that case to have any optional members (eg B) at comprehensive, and core group (A) at a lower level.

@macchiati are you picturing A and B as two logical groups, or as two subsets of a single logical group?

@pedberg-icu as you suggest I'll investigate dividing this group into two:

moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-1"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-2"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-3"]
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="0"]
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="1"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="2"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="3"]

btangmu:
I added changes to LogicalGroupings.java so that these are two distinct groups:

```
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-1"]
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="0"]
moderate //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="1"]

comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-2"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="-3"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="2"]
comprehensive //ldml/dates/fields/field[@type="day-narrow"]/relative[@type="3"]
```

srl295 (commented):
code looks OK- seems like the discussion is still in progress?

btangmu:
> seems like the discussion is still in progress?

@srl295 yes -- if @macchiati agrees with splitting the group in two, at least for now, then the PR can be made ready for review as-is -- maybe it's best to postpone merging, pending response to my question "are you picturing A and B as two logical groups, or as two subsets of a single logical group?"

macchiati:
There is a provision in the code already for having a 'core' group, and
then optionals (that require the core group). I forget the exact
APIs/structure.

(The full generalization would be a dependency graph: X requires {Y, Z, W},
Y requires {W}, etc. But I don't think we need that.)

macchiati:
Right. If it is optional, and there is a missing required value, then we
should present an error.

The question is: which item to attach the error to. What we try to do is
attach it to the thing you should change. In this case, we don't want the
error to appear on the optional item, but on (at least) one of the missing
required items (or maybe all of them).

btangmu:
I'm closing this PR as-is, so we can refer back to it as needed.

I've made another PR: https://github.com/unicode-org/cldr/pull/2160 . It revises the unit test to take isOptional() into account, and refactors handleCheck (without change in functionality) in preparation for changing its functionality. One such change will be to trigger an error on a missing non-optional path if there is an optional non-missing path in the same group.

PR #2149, opened 2022-07-01
CLDR-8666 Report summary: improve view

srl295:
- sticky table header
- include counts for each report type
- includes updated spreadsheet

CLDR-8666

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrReport.js
+++ b/tools/cldr-apps/js/src/esm/cldrReport.js
@@ -8,6 +8,7 @@ import * as cldrLoad from "./cldrLoad.js";
 import * as cldrSurvey from "./cldrSurvey.js";
 import * as cldrText from "./cldrText.js";
 import * as cldrVueRouter from "../cldrVueRouter.js";
+import * as cldrXlsx from "./cldrXlsx.js";
 import ReportResponse from "../views/ReportResponse.vue";
 import XLSX from "xlsx";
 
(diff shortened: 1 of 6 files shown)


srl295:
@macchiati  
image


Still want to add spreadsheet data, but this is what i have so far.

srl295:
there's a data problem I'm working on (shown in the 'zh' column) so will keep this draft for now

srl295:
Draft until #2155  merged.

Current look:
image

- [ ] Need to fix spreadsheet download

srl295:
[survey_reports.xlsx](https://github.com/unicode-org/cldr/files/9031303/survey_reports.xlsx)

Spreadsheet example. (I've set the headers to be angled, they aren't like that by default)
image

macchiati:
I can't see the original (blank in browser, and can't download). The format
looks good.

Question: in terms of generating "approved" or not, I had thought that the
votes were weighted by the voter's level (eg 1 for Guest, 4 for Mozilla,
etc). If so, that should show up in the count of acceptable, eg Amharic
Numbers-Acceptable should be (I'm guessing) 4 or 6, not 1.

srl295:
> I can't see the original (blank in browser, and can't download). The format looks good. Question: in terms of generating "approved" or not, I had thought that the votes were weighted by the voter's level (eg 1 for Guest, 4 for Mozilla, etc). If so, that should show up in the count of acceptable, eg Amharic Numbers-Acceptable should be (I'm guessing) 4 or 6, not 1.

Aha…

Actually, I am count the number of *voters* for each item. If you want the VoteResolver vote (the score) that is actually even simpler to produce.

macchiati:
Yes, the score is what determines whether it is approved or not; so that is
the main information of interest.

tools/cldr-apps/js/src/esm/cldrReport.js
```diff
@@ -181,61 +151,42 @@ async function reportTypes() {
...
-    [
-      "Locale", // 0
-      "Code", // 1
-      "Completion", // 2
-      "Acceptable", // 3
-      "Unacceptable", // 4
-      "Total Votes",
-    ],
-  ];
+  var ws_data = [["Locale", "Code", "Total Votes"]];
```
btangmu, line 162: best practice is to use "let" instead of "var", the latter being bug-prone and rarely having any advantage -- in this case it looks like "const" would be even better

also this function is 50 lines long, should be shortened with subroutines; no need for pushComment to be nested; short simple functions with descriptive names are more maintainable

srl295: @btangmu  done and improved.  This function is a spreadsheet export, I think it would hurt readability to split it up further, as it would make it harder to follow the flow of the data generation.

macchiati:
Right. Suppose we have a score of 8 for acceptable (2 voters) and score of
3 for unacceptable (3 voters), the scores are the important figures.

srl295:
updated to use scores:

image

Amharic: one voter, but score of 6

image

srl295:
> Right. Suppose we have a score of 8 for acceptable (2 voters) and score of 3 for unacceptable (3 voters), the scores are the important figures.
> […](#)

Done. PTAL

PR #2156, opened 2022-07-04
CLDR-15725 compound unit transforms

macchiati:
CLDR-15725


macchiati:
Put in a proof of concept, with rules using regex.
Added spreadsheet with current results, just a few rule sets.
https://docs.google.com/spreadsheets/d/1L58nWgxn8sWiOmfTf1VgdejRp12fJqZ91fG5ru4v-VQ/edit#gid=0

Need to analyze more cases, seeing what rules would result.

tools/cldr-code/src/main/java/org/unicode/cldr/util/BoundaryTransform.java
```diff
@@ -0,0 +1,219 @@
+package org.unicode.cldr.util;
```
pedberg-icu, line 1: Should this be in an existing or new icu class (since presumably it will go there eventually)?

pedberg-icu:
This looks very promising! {presumably we want the code in ICU eventually...)

tools/cldr-code/src/main/java/org/unicode/cldr/util/BoundaryTransform.java
macchiati: Once we get this further along, it (or an optimized version of it) could go into ICU. We should work first on the rules, and make sure that the syntax is powerful enough for what we need (and reasonably clear).

macchiati: Note that the current implementation uses regex for convenience. We would need to make sure that we use syntax that is easily supported across platforms. It doesn't have to be the same, but would have to be easily transformed into "native" regex.

macchiati:
Fleshed out a bit more, added first cut at the list rules for ICU. Also tested inserting spaces between chinese and non-chinese letters, and changing 'a' to 'an' before vowels in English.

tools/cldr-code/src/main/java/org/unicode/cldr/util/BoundaryTransform.java
```diff
@@ -0,0 +1,271 @@
...
+            + " ⦅o ❙⦆([uU8]|[hH][oU]|11($| ))→u ;" // o
+            ))
+        .put("he", BoundaryTransform.from(""
+            + " \\u05D5⦅❙⦆\\P{sc=Hebr}→-;" // y
+            ))
+        .put("en", BoundaryTransform.from(
+            " a⦅ ❙⦆[aeiou]→n ;" // doesn't handle "an hour", or "an underground"
+            ))
+        .put("zh", BoundaryTransform.from(
+            "[\\p{L}&&\\p{sc=hani}]⦅❙⦆[\\p{L}&&\\P{sc=hani}]→ ;" // add space at chinese/non-chinese letter junction
```
pedberg-icu, line 230: won't this add a space between Han chars and shared CJK punctuation like brackets etc?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBoundaryTransform.java
```diff
@@ -0,0 +1,484 @@
...
+            return pattern2;
+        }
+    }
+
+    public void TestGeneralFormat() {
+        String[][] tests = {
+            {"en", "Take a {0}", "book", "Take a book"},
+            {"en", "Take a {0}", "apple", "Take an apple"},
+            {"zh", "舘{0}", "豈", "舘豈"},
+            {"zh", "舘{0}", "a", "舘 a"},
```
pedberg-icu, line 463: Need som test cases with punctuation like CJK brackets etc.

pedberg-icu:
Looks good but I had a concern about  spacing getting added between Han characters and CJK punctuation like brackets.

macchiati:
Good question. That is only adding spaces between letters (between a Han
and a non-Han). I suspect it will need some refinement....

pedberg-icu:
If the current code does *not* add code between Han chars and punctuation, that is good, I was afraid that it did. I think we *may* need to add spaces between Han chars and Latin digits, will check on that. But it seems like the current code makes some of the improvements we want without doing any harm. so that is good.

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/BoundaryTransform.java
macchiati: The \\p{L}&& restricts it to letters

PR #1084, opened 2021-03-04
CLDR-14199 Add new locale file shn.xml to seed/main

i18ncraig:
CLDR-14199 Add new Myanmar locale file for Shan (ISO shn) to seed/main.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/CLDR-14199
- [X] Updated PR title and link in previous line to include Issue number

PR #2158, opened 2022-07-05
Add files via upload

SZTSIIT:
Adding the fixed version of Shn.xml file. 
Follow up with this PR [](https://github.com/unicode-org/cldr/pull/1084)

CLDR-_____


SZTSIIT:
https://github.com/unicode-org/cldr/pull/2159
Another Developer PR with proper files.

PR #2159, opened 2022-07-06
CLDR-14199 Add new shn locale file to seed/main

sengkyaut:
CLDR-14199 Add new shn locale file to seed/main


srl295 (dismissed):
looks complete and well formed. @macchiati  too late for 42? or at least wait for ST to close?

DavidLRowe:
I looked through the shn.xml file and have some questions.

The early part of the file looks reasonable (with the caveat that I don't know the Shan language or the Burmese script).

However, looking at exemplars:
- main has a-z, whereas I would have expected Burmese characters
- other exemplars have triple up arrow (which seems odd in this context)

The pattern elements for decimalFormats[@numberSystem="latn"]/decimalFormatLength[@type="long"]/decimalFormat
contain "thousand", "million", "billion", "trillion" rather than Shan words. (Similarly for "short", pattern elements use K, M, B, T.)

The currency type elements are not translated, for example, currencies/currency[@type="ADP"]/displayName has "Andorran Peseta" rather than a Shan translation.

Also:
minimalPairs info seems suspect.
units info seems suspect.
listPatterns info seems suspect.
posix info seems suspect.
characterLabels info seems suspect.

@macchiati Is this what you would expect?

petariordanoo72 (approved):
Start 5€ bonus

srl295:
> I looked through the shn.xml file and have some questions.
> 
> The early part of the file looks reasonable (with the caveat that I don't know the Shan language or the Burmese script).
> 
> However, looking at exemplars:
> 
> * main has a-z, whereas I would have expected Burmese characters
> * other exemplars have triple up arrow (which seems odd in this context)
> The pattern elements for decimalFormats[@numbersystem="latn"]/decimalFormatLength[@type="long"]/decimalFormat contain "thousand", "million", "billion", "trillion" rather than Shan words. (Similarly for "short", pattern elements use K, M, B, T.)
> The currency type elements are not translated, for example, currencies/currency[@type="ADP"]/displayName has "Andorran Peseta" rather than a Shan translation.
> Also: minimalPairs info seems suspect. units info seems suspect. listPatterns info seems suspect. posix info seems suspect. characterLabels info seems suspect.

@sengkyaut maybe en.xml was used as a baseline?  In any event, the exemplars need to be fixed, but any other data should just be dropped from the xml file. 

> @macchiati Is this what you would expect?

None of the errors you mention are expected.

macchiati:
All new locales should have *only* the core data (see
https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales).

All the other data should be entered in using the Survey Tool, to help
prevent errors from creeping in.

srl295:
> All new locales should have *only* the core data (see
> https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales).
> 
> All the other data should be entered in using the Survey Tool, to help
> prevent errors from creeping in.
> 


Delete the non core data and put it in the pr. But keep the original file. Next release, upload that file via the surveytool. 

TC:

1. We should update our docs to specify to only include core. 

2. And we need to have a skeleton xml file available  for folks.

sengkyaut:
According to https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales.
- [x] correct language code
- [x] four exemplar sets
- [x] Verified country data
- [x] Default content script and region
- [x] correct time cycle used with the language in the default content region
And
- [x] removed the other non core data.

srl295:
I can't comment on the file due to Github /browser being slow to render Shan!

- The exemplar set looks good.
- There are many lines such as `<ellipsis type="final">↑↑↑</ellipsis>` - does this really mean that you are confirming the root data? 

In any event, this is much more than core data. Core data would be a subset such as https://gist.github.com/srl295/e7d5a70b6512b8f9eb709a0aca992b8f

sengkyaut:
I've update the shn.xml version that @srl295 provided as in https://gist.github.com/srl295/e7d5a70b6512b8f9eb709a0aca992b8f.

sengkyaut:
@macchiati What do you think sir?

sffc (approved):
LGTM; looking forward to the full data landing!

srl295 (approved):

srl295:
@macchiati  @DavidLRowe  any comments or ok to land?

DavidLRowe:
@srl295 Has this PR passed the tests? It looks as though the tests need permission to run. Often it's only from the test failures that I see what's missing.

srl295:
> @srl295 Has this PR passed the tests? It looks as though the tests need permission to run. Often it's only from the test failures that I see what's missing.

good call!  I approved it before, but since they are still a first time contributor (not merged yet) it re-requests permission.

DavidLRowe:
It looks as though the tests are passing. Just a couple of questions: 

languageData element in supplementalData.xml currently has:
```
<language type="shn" scripts="Mymr"/>
<language type="shn" territories="MM" alt="secondary"/>
```
Do we need entries for TH as well?

Is it a problem that "shn" isn't listed in languageGroup.xml?

sengkyaut:
> It looks as though the tests are passing. Just a couple of questions:
> 
> languageData element in supplementalData.xml currently has:
> 
> ```
> <language type="shn" scripts="Mymr"/>
> <language type="shn" territories="MM" alt="secondary"/>
> ```
> 
> Do we need entries for TH as well?
> 
> Is it a problem that "shn" isn't listed in languageGroup.xml?

https://en.wikipedia.org/wiki/Shan_people
https://shn.wikipedia.org/wiki/တႆး
Shan people also known as the **Tai Long**, or **Tai Yai** primarily live in the Shan State of Myanmar and Northern Thailand.
That is why I add shn_TH entries. I am not sure if it will be a problem or not because it successfully passed the check in my repo's workflow. I just read the https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales and only edit **supplementalMetadata.xml**. I have no idea what **languageGroup.xml** is used for to be sure.

sengkyaut:
> It looks as though the tests are passing. Just a couple of questions:
> 
> languageData element in supplementalData.xml currently has:
> 
> ```
> <language type="shn" scripts="Mymr"/>
> <language type="shn" territories="MM" alt="secondary"/>
> ```
> 
> Do we need entries for TH as well?
> 
> Is it a problem that "shn" isn't listed in languageGroup.xml?

https://en.wikipedia.org/wiki/Shan_people
https://shn.wikipedia.org/wiki/တႆး
Shan people also known as the **Tai Long**, or **Tai Yai** primarily live in the Shan State of Myanmar and Northern Thailand.
That is why I add shn_TH entries. I am not sure if it will be a problem or not because it successfully passed the check in my repo's workflow. I just read the https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales and only edit **supplementalMetadata.xml**. I have no idea what **languageGroup.xml** is used for to be sure.

srl295 (approved):
approving now that tests pass

srl295:
I'll put this on for discussion in today's call

PR #2160, opened 2022-07-06
CLDR-10564 Logical groups; unit test and refactoring

btangmu:
-New unit test TestCheckLogicalGroupings.testSameCoverageLevel

-Take isOptional into account for the unit test

-Refactor CheckLogicalGroupings.java, especially extremely long method handleCheck

-New LogicalGroupingChecker.java encapsulates the refactored code with shorter methods

CLDR-10564

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckLogicalGroupings.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckLogicalGroupings.java
@@ -3,21 +3,14 @@ package org.unicode.cldr.test;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
-import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRFile.DraftStatus;
 import org.unicode.cldr.util.Factory;
-import org.unicode.cldr.util.Level;
 import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.LogicalGrouping;
-import org.unicode.cldr.util.LogicalGrouping.PathType;
 import org.unicode.cldr.util.With;
-import org.unicode.cldr.util.XMLSource;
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/test/TestCheckLogicalGroupings.java
```diff
@@ -79,4 +76,53 @@ private void assertCheckStatus(String localeID, CheckLogicalGroupings clg, Strin
...
+
+    void sameLevel(String localeId) {
+        final CLDRConfig config = CLDRConfig.getInstance();
+        final CLDRFile cldrFile = config.getCLDRFile(localeId, true);
+        final SupplementalDataInfo sdi = SupplementalDataInfo.getInstance();
+        final CoverageLevel2 coverageLevel = CoverageLevel2.getInstance(sdi, localeId);
+        for (final String path : cldrFile.fullIterable()) {
+            Set<String> grouping = LogicalGrouping.getPaths(cldrFile, path);
+            if (grouping != null && grouping.size() > 1) {
+                LogicalGrouping.removeOptionalPaths(grouping, cldrFile);
```
btangmu, line 99: This "removeOptionalPaths" is the only difference between this PR and the preceding draft PR, as far as the unit test is concerned

macchiati (commented):
Looks good. Let's hold off integrating until we are done vetting, unless you think there are good reasons to integrate.

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestCheckLogicalGroupings.java
srl295: also the test "if (grouping != null) … " was added.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckLogicalGroupings.java
```diff
@@ -342,7 +145,7 @@ public int size() {
         }
 
         public static Fingerprint make(String value) {
-            return FINGERPRINT_CACHE.computeIfAbsent(value, v -> new Fingerprint(v));
+            return FINGERPRINT_CACHE.computeIfAbsent(value, Fingerprint::new);
```
srl295, line 148: yes, this is the idiom i should be using elsewhere.

srl295 (approved):
lgtm

PR #2164, opened 2022-07-08
CLDR-13082 Add link for headers on ST pages

btangmu:
-Move code to new method cldrLoad.addSectionHeader; add onclick

-New methods cldrTable.isHeaderId, makeHeaderId, and goToHeaderId

-Encapsulate row_ (formerly r@) prefix with new cldrTable.makeRowId

-Remove unused parameters doPush/replace for cldrLoad.replaceHash, setHash, updateHashAndMenus

-Remove deprecated and pointless substr(0) in cldrLoad.js

-Use CLDR_LOAD_DEBUG instead of local variable reportProblems

-New unit test for cldrTable

-Comments

CLDR-13082

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.js
+++ b/tools/cldr-apps/js/src/esm/cldrTable.js
@@ -22,6 +22,9 @@ import * as cldrText from "./cldrText.js";
 import * as cldrVote from "./cldrVote.js";
 import * as cldrXPathUtils from "./cldrXpathUtils.js";
 
+const HEADER_ID_PREFIX = "header_";
+const ROW_ID_PREFIX = "row_"; // formerly "r@"
+
 const CLDR_TABLE_DEBUG = false;
 
 const TABLE_USES_NEW_API = false;
@@ -214,14 +217,7 @@ function insertRowsIntoTbody(theTable, reuseTable) {
 
       if (newPartition != curPartition) {
         if (newPartition.name != "") {
-          var newPar = cldrSurvey.cloneAnon(parRow);
-          var newTd = cldrSurvey.getTagChildren(newPar);
-          var newHeading = cldrSurvey.getTagChildren(newTd[0]);
-          newHeading[0].innerHTML = newPartition.name;
-          newHeading[0].id = newPartition.name;
-          tbody.appendChild(newPar);
-          newPar.origClass = newPar.className;
-          newPartition.tr = newPar; // heading
+          addSectionHeader(newPartition, tbody, parRow);
         }
         curPartition = newPartition;
       }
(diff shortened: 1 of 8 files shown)


tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -22,6 +22,16 @@ import * as cldrText from "./cldrText.js";
 import * as cldrVote from "./cldrVote.js";
 import * as cldrXPathUtils from "./cldrXpathUtils.js";
 
+const HEADER_ID_PREFIX = "header:";
+
+/*
+ * TODO: usage of '@' in tr.id appears to be problematic for jQuery:
+ * if we try to use selectors like $('#' + tr.id), we get
+ * "Syntax error, unrecognized expression: #r@f3d4397b739b287"
+ * Is there a good reason to keep '@'?
```
srl295, line 31: no, fine to change it

srl295: I don't think it shows up in user URLs, so should not have an impact.

COlon `:` may need escaping in urls.

tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -1301,12 +1321,37 @@ function getValidWinningValue(theRow) {
...
+
+function isHeaderId(id) {
+  return id.startsWith(HEADER_ID_PREFIX);
+}
+
+function makeHeaderId(name) {
+  // Use encodeURIComponent since, for example, name might
+  // be "Western Asia", and spaces are illegal in id attributes;
+  // add HEADER_ID_PREFIX to identify this kind of id uniquely
+  return HEADER_ID_PREFIX + encodeURIComponent(name);
```
srl295, line 1336: Better (more human friendly if the links are passed around) would probably be to strip out spaces, leaving `#WesternAsia`

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrTable.js
btangmu: sounds good! I'll strip the spaces if there's no objection...

maybe stripping out (rather than escaping) some other punctuation would be even more human-friendly, however I'm not very confident the headers would be unique (in a page) if all punct were removed -- any thoughts on that?

srl295: i think they probably are unique, but safest would be to have a test to confirm.  @macchiati  thoughts?

tools/cldr-apps/js/src/esm/cldrTable.js
btangmu: "Technically, in [HTML5](https://developer.mozilla.org/en-US/docs/Glossary/HTML5), the value for an id attribute may contain any character, except [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) characters. However, to avoid inadvertent errors, only [ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) letters, digits, '_', and '-' should be used and the value for an id attribute should start with a letter." https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

btangmu: I've revised this PR -- now the prefixes are "header_" and "row_", and all non-alphanumeric characters are replaced by underscore in the id

tools/cldr-apps/js/src/esm/cldrTable.js
btangmu: Although I haven't done an exhaustive test that header_ ids are unique, I'm fairly confident; eventually we should have a mode for cldr-apps-webdriver to cycle through all pages and validate the html -- that would catch any non-unique id in a more general way

PR #2165, opened 2022-07-09
CLDR-15801 workaround for data from Turkmenistan

macchiati:
CLDR-15801

1. The data is marked as draft=contributed.
2. Some of the XML data was commented out, because they caused errors. 
    1. The date formats were malformed. Using AAAA instead of MMMM for long month, and ý for year.
    2. The abbreviated islamic months require language knowledge to fix. 
    3. In both cases, the Survey tool alerts people to the problems as they enter them, allowing them to fix the syntax.

### DIFF
--- a/common/main/tk.xml
+++ b/common/main/tk.xml
@@ -30,14 +30,17 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="alt">günorta Altaý dili</language>
 			<language type="am">amhar dili</language>
 			<language type="an">aragon dili</language>
+			<language type="ann" draft="contributed">Obolo dili</language>
 			<language type="anp">angika dili</language>
 			<language type="ar">arap dili</language>
 			<language type="ar_001">häzirki zaman standart arap dili</language>
 			<language type="arn">mapuçe dili</language>
 			<language type="arp">arapaho dili</language>
+			<language type="ars" draft="contributed">Nejdi arap dili</language>
 			<language type="as">assam dili</language>
 			<language type="asa">asu dili</language>
 			<language type="ast">asturiý dili</language>
+			<language type="atj" draft="contributed">Atikamekw dili</language>
 			<language type="av">awar dili</language>
 			<language type="awa">awadhi dili</language>
 			<language type="ay">aýmara dili</language>
@@ -63,6 +66,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="bug">bugiý dili</language>
 			<language type="byn">blin dili</language>
 			<language type="ca">katalan dili</language>
+			<language type="cay" draft="contributed">Kaýuga dili</language>
 			<language type="ccp">çakma dili</language>
 			<language type="ce">çeçen dili</language>
 			<language type="ceb">sebuan dili</language>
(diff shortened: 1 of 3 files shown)


common/main/tk.xml
```diff
@@ -1690,6 +1784,169 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 					</intervalFormats>
 				</dateTimeFormats>
 			</calendar>
+			<calendar type="islamic">
+				<months>
+					<monthContext type="format">
+						<!-- UPLOAD ERRORS
```
macchiati, line 1790: Commented out due to errors (inconsistent . usage)

common/main/tk.xml
```diff
@@ -1690,6 +1784,169 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+							<month type="7" draft="contributed">Rejep</month>
+							<month type="8" draft="contributed">Meret</month>
+							<month type="9" draft="contributed">Oraza</month>
+							<month type="10" draft="contributed">Baýram</month>
+							<month type="11" draft="contributed">Boş aý</month>
+							<month type="12" draft="contributed">Gurban</month>
+						</monthWidth>
+					</monthContext>
+					<monthContext type="stand-alone">
+						<!-- UPLOAD ERRORS
```
macchiati, line 1836: Commented out due to errors (inconsistent . usage)

common/main/tk.xml
```diff
@@ -1690,6 +1784,169 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+							<month type="12" draft="contributed">Gurban</month>
+						</monthWidth>
+					</monthContext>
+				</months>
+				<eras>
+					<eraAbbr>
+						<era type="0" draft="contributed">HS</era>
+					</eraAbbr>
+				</eras>
+				<!-- UPLOAD ERRORS
```
macchiati, line 1887: Commented out due to errors (bad placeholder formats, eg AAAA instead of MMM)

tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
```diff
@@ -1,196 +1,305 @@
...
+locale=tk ; action=add ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="2"][@draft="contributed"] ; new_value=Sap.
+locale=tk ; action=add ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="3"][@draft="contributed"] ; new_value=4 tir 1
+locale=tk ; action=add ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="4"][@draft="contributed"] ; new_value=4 tir 2
+locale=tk ; action=add ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="5"][@draft="contributed"] ; new_value=4 tir 3
+locale=tk ; action=add ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="6"][@draft="contributed"] ; new_value=4 tir 4
```
srl295, line 110: Should this file really be checked in? Maybe it should be .gitignored

srl295 (approved):
Lgtm. This whole sub project for tk is some percentage of heroic on all sides!

tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
macchiati: I think it is actually useful to have a store of these as examples.

srl295: I haven't used this tool.  However, it does seem (looking at history) a little random to check in the file each time with the latest result.  If we want to have an example modify_config.txt or set of them for different purposes, that could be good.

Might be better to take a filesystem path rather than an in-source-tree path, at least as an option.  That would be what I'd expect from an occasional tool.

PR #2169, opened 2022-07-13
CLDR-15646 Priority Items Summary, fix bug in last PR; add debug logging

btangmu:
-Was missing this. in this.canCreateSnapshots

-Debug logging conditional on SUMMARY_DEBUG

CLDR-15646

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrPriorityItems.js
+++ b/tools/cldr-apps/js/src/esm/cldrPriorityItems.js
@@ -6,6 +6,8 @@
 import * as cldrAjax from "../esm/cldrAjax.js";
 import * as cldrStatus from "../esm/cldrStatus.js";
 
+const SUMMARY_DEBUG = true;
+
 const SECONDS_IN_MS = 1000;
 
 const NORMAL_RETRY = 10 * SECONDS_IN_MS; // "Normal" retry: starting or about to start
@@ -85,11 +87,20 @@ function viewCreated(setData, setSnap) {
 function fetchStatus() {
   if (!canSum || "vsummary" !== cldrStatus.getCurrentSpecial()) {
     canSum = canSnap = canCreateSnap = false;
+    if (SUMMARY_DEBUG) {
+      console.log("cldrPriorityItems.fetchStatus resetting for other special");
+    }
     return;
   }
   if (canSum) {
     if (canSnap) {
       listSnapshots();
+    }
+    if (SUMMARY_DEBUG) {
+      console.log(
+        "cldrPriorityItems.fetchStatus using latestArgs.summarizeAllLocales = " +
+          latestArgs.summarizeAllLocales
+      );
     }
     requestSummary(new SummaryArgs(latestArgs, LOAD_NOSTART));
   }
@@ -97,11 +108,20 @@ function fetchStatus() {
 
 function start(summarizeAllLocales) {
   const sa = new SummaryArgs(null, LOAD_START);
+  if (SUMMARY_DEBUG) {
+    console.log(
+      "cldrPriorityItems.start using summarizeAllLocales = " +
+        summarizeAllLocales
+    );
+  }
   sa.setSummarizeAllLocales(summarizeAllLocales);
   requestSummary(sa);
 }
 
 function stop() {
+  if (SUMMARY_DEBUG) {
+    console.log("cldrPriorityItems.stop");
+  }
   requestSummary(new SummaryArgs(latestArgs, LOAD_FORCESTOP));
 }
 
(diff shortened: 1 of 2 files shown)


btangmu:
@macchiati @srl295 there seem to be some timing issues making behavior pseudo random, possibly depending on browser (Firefox vs Chrome), debugging tools on front and back end, etc. This morning, occasionally I seem to be able to repro problems similar to what Mark has reported.

On the front end, this code seems suspicious:
```
    if (latestArgs.loadingPolicy === LOAD_NOSTART) {
        window.setTimeout(fetchStatus.bind(this), NORMAL_RETRY);
    }
```
Possibly it should be:
```
    if (latestArgs.loadingPolicy !== LOAD_FORCESTOP) {
...
```

On the back end, this code seems suspicious:
```
status = "Waiting...";
progress.update("Waiting...");
OnlyOneVetter.acquire();
try {
    if (stop) {
        status = "Stopped on request.";
        statusCode = Status.STOPPED;
        return;
    }
    processCriticalWork(progress);
} finally {
    OnlyOneVetter.release();
}
status = "Finished.";
statusCode = Status.READY;
```
I'm occasionally seeing it seeming to get stuck on "Waiting..."

btangmu:
I turned on DEBUG in VettingViewerQueue and the console shows:
[INFO] Creating task for Priority Items Summary
[INFO] Starting up vv task: Priority Items Summary
[INFO] Starting generation of Priority Items Summary
[INFO] Creating task for Priority Items Summary
[INFO] Starting up vv task: Priority Items Summary
[INFO] [ERROR   ] While VettingViewer processing Priority Items Summary
[INFO] java.lang.InterruptedException
[INFO] Creating task for Priority Items Summary
[INFO] Starting up vv task: Priority Items Summary

srl295 (approved):

srl295:
> @macchiati @srl295 there seem to be some timing issues making behavior pseudo random, possibly depending on browser (Firefox vs Chrome), debugging tools on front and back end, etc. This morning, occasionally I seem to be able to repro problems similar to what Mark has reported.
> 
> On the front end, this code seems suspicious:
> 
> ```
>     if (latestArgs.loadingPolicy === LOAD_NOSTART) {
>         window.setTimeout(fetchStatus.bind(this), NORMAL_RETRY);
>     }
> ```
> 
> Possibly it should be:
> 
> ```
>     if (latestArgs.loadingPolicy !== LOAD_FORCESTOP) {
> ...
> ```
> 
> On the back end, this code seems suspicious:
> 
> ```
> status = "Waiting...";
> progress.update("Waiting...");
> OnlyOneVetter.acquire();
> try {
>     if (stop) {
>         status = "Stopped on request.";
>         statusCode = Status.STOPPED;
>         return;
>     }
>     processCriticalWork(progress);
> } finally {
>     OnlyOneVetter.release();
> }
> status = "Finished.";
> statusCode = Status.READY;
> ```
> 
> I'm occasionally seeing it seeming to get stuck on "Waiting..."

I'd think 'finally' gets fired even if there was an exception. so it should only wait if someone still has the lock acquired.

btangmu:
> I'd think 'finally' gets fired even if there was an exception. so it should only wait if someone still has the lock acquired.

I think the thread is killed after acquire, before release... I put in additional println as shown in the new ticket https://unicode-org.atlassian.net/browse/CLDR-15819 and the release never happens

srl295:
> > I'd think 'finally' gets fired even if there was an exception. so it should only wait if someone still has the lock acquired.
> 
> I think the thread is killed after acquire, before release... I put in additional println as shown in the new ticket https://unicode-org.atlassian.net/browse/CLDR-15819 and the release never happens

Very strange.  I wonder why the thread was killed, not just interrupted. I'd think an InterruptedException first.

btangmu:
> Very strange

Yes. We have Survey-Tool-specific code tangled up with long-running-task-management code; the latter should be a separate reusable class, preferably one that we don't have to write and maintain ourselves...

srl295:
> > Very strange
> 
> Yes. We have Survey-Tool-specific code tangled up with long-running-task-management code; the latter should be a separate reusable class, preferably one that we don't have to write and maintain ourselves...

If you are referring to the threadpool, note that directly workign with threads is [forbidden within J2EE](https://www.martellosecurity.com/kb/mitre/cwe/383/) because all threads are managed by the web container.

btangmu:
does VettingViewerQueue.java currently obey that rule?

srl295:
> does VettingViewerQueue.java currently obey that rule?

Yes, although I think I see a bug:

```java
// VVQ.java 397ff

        // TODO: May be better to use SurveyThreadManager.getExecutorService().invoke() (rather than a raw thread) but would require
        // some restructuring
        t = entry.currentTask = new Task(entry, args.usersOrg);
        t.myThread = SurveyThreadManager.getThreadFactory().newThread(t);
        t.myThread.start();
        SurveyThreadManager.getThreadFactory().newThread(t);  // <<<<<<<<<<<<<<<<<<<<
```

newThread is called twice…

srl295:
> yikes!

not clear this is the cause. but it's definitely wrong

PR #2187, opened 2022-07-21
CLDR-15547 Add back group separator to pa and ta

AEApple:
CLDR-15547


AEApple:
Note: I didn't touch the accounting formats since they weren't modified as part of the ticket originally, nor did I touch Urdu since Urdu didn't have the second group separator originally (at least not immediately prior to the removal of the space)

macchiati (approved):

PR #2193, opened 2022-07-22
CLDR-15804 cs: add availableFmts/intervalFmts for vvvv, with comma

pedberg-icu:
CLDR-15804

In CLDR 42 Survey Tool submission, Czech vetters added a comma to the standard full time format (with zzzz) `H:mm:ss, zzzz` (and this won), leaving the standard long time format (with just z) without a comma: `H:mm:ss z`. To match this and prevent associated warnings, we need to update the availableFormats and intervalFormats to add patterns that distinguish skeletons with `vvvv` (which should have comma) from skeletons with just `v` (which should not). Adding these was something the vetters could not do in the Survey Tool.

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -1615,7 +1615,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ELEMENT intervalFormatItem ( alias | ( greatestDifference*, special* ) ) >
 <!ATTLIST intervalFormatItem id NMTOKEN #REQUIRED >
     <!-- TODO: check to see if this should be minimized -->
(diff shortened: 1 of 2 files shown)


macchiati (approved):

macchiati:
Looks good; I approved. Note for later; we might consider converting
identical items between gregorian and other calendars to be ^^^ where the
inheritance would be identical.

pedberg-icu:
> Looks good; I approved. Note for later; we might consider converting identical items between gregorian and other calendars to be ^^^ where the inheritance would be identical.
> […](#)

I think that will happen automatically in GenerateProductionData (any value same as Bailey will be removed)

PR #2194, opened 2022-07-23
CLDR-15857 simplify person names code

macchiati:
CLDR-15857

Simplify the code by replacing ParameterMatcher by FormatParameters.
ROUGH CUT for now.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -33,7 +33,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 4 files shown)


macchiati:
Hi, this is a very first cut at simplifying the code.

macchiati:
The tests pass, and I added a test to check that we can format all the sample names. But it still needs cleanup.
- remove commented-out code
- mark methods that are only public for testing
- add the API we discussed (eg to option to uppercase surname for surnameFirst if the locale is givenFirst
- other cleanup

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -981,263 +1041,273 @@ private <T> void addToLabel(T item, StringBuilder sb) {
...
+//            // now replace any "complete" items by empty.
+//            Set<ParameterMatcher> result2 = new TreeSet<>();
+//            for (ParameterMatcher item1 : result) {
+//                result2.add(item1.slim());
+//            }
+//            return ImmutableSet.copyOf(result2);
+//        }
+//
+//    }
+
```
richgillam, line 1302: Do you need to keep all this commented-out code in here?  If you can just get it out of a previous revision in git, that'd be cleaner.  (Or is this just still here until you're completely done with this simplification effort, and THEN you'll delete it?)

richgillam (approved):
Didn't read it EXTREMELY closely, but it all looks pretty straightforward.

macchiati:
Yes, I left it in just during development; will delete.

richgillam (approved):
Looks good!

PR #2203, opened 2022-07-27
CLDR-15860 test that all defaultCollations exist

srl295:
CLDR-15860 support collation inheritance, test all defaultCollations

- TestCollation: test that all defaultCollations exist
- TestLocaleIDParser: add a test for LocaleIDParser
- add bool param to LocaleIDParser.getParent() for collation trees
- add Factory.setIgnoreExplicitParentLocale(true) for collation trees


macchiati (approved):

srl295:
thanks @macchiati  - but zh_Hant doesn't have a stroke collation .  So this can't be merged until it passes.

pedberg-icu (dismissed):

pedberg-icu:
> thanks @macchiati - but zh_Hant doesn't have a stroke collation . So this can't be merged until it passes.

@srl295 The stroke collation data is in zh.xml; and collation data is supposed to inherit across script boundaries, unlike locale data.

srl295:
@pedberg-icu hm. somehow it is not! 🤔

pedberg-icu:
> @pedberg-icu hm. somehow it is not! 🤔

Well, I cannot find a unit test that demonstrates that it works in CLDR. But it works in ICU, for example
- TestGetLocale() in main/tests/collate/src/com/ibm/icu/dev/test/collator/CollationAPITest.java shows that if you request a collation for zh_TW, the actualLocale for the collation you get is "zh@collation=stroke"
- TestGetKeywordValues in main/tests/collate/src/com/ibm/icu/dev/test/collator/CollationServiceTest.java shows that the collation keywords available for zh_Hant are "stroke", "pinyin", "eor", "search", "standard"

srl295:
> > @pedberg-icu hm. somehow it is not! 🤔
> 
> Well, I cannot find a unit test that demonstrates that it works in CLDR. But it works in ICU, for example
> 
> * TestGetLocale() in main/tests/collate/src/com/ibm/icu/dev/test/collator/CollationAPITest.java shows that if you request a collation for zh_TW, the actualLocale for the collation you get is "zh@collation=stroke"
> * TestGetKeywordValues in main/tests/collate/src/com/ibm/icu/dev/test/collator/CollationServiceTest.java shows that the collation keywords available for zh_Hant are "stroke", "pinyin", "eor", "search", "standard"

It's in the tr35, at least under the definition of `<parentLocales>`.   However, neither LocaleIDParser nor CLDRFile+ understand this structure. Will add some options and tests.

pedberg-icu (approved):

pedberg-icu:
Nice. At some point we will need to think about things like ars → ar and hi_Latn → en_IN where we probably _do_ want to follow parent data for collation. Need to figure out how to indicate that.

macchiati (approved):

pedberg-icu:
> Nice. At some point we will need to think about things like ars → ar and hi_Latn → en_IN where we probably _do_ want to follow parent data for collation. Need to figure out how to indicate that.

Maybe what we do is follow non-root parent locales for collation data.

srl295:
> > Nice. At some point we will need to think about things like ars → ar and hi_Latn → en_IN where we probably _do_ want to follow parent data for collation. Need to figure out how to indicate that.
> 
> 
> 
> Maybe what we do is follow non-root parent locales for collation data.

We could go back(?)  to having an attribute in the collation indicating parentage.  Or have the supplemental data have an attribute that indicates that collation is affected also.

In any event it will be a change for collation consumers.

PR #2210, opened 2022-07-28
CLDR-15599 Add Catalan minimal pair value for many

AEApple:
CLDR-15599

Add Catalan minimal pair value for many


srl295 (approved):
Approving.  A similar (but linguistically uninformed) change locally caused the tool to pass

PR #2211, opened 2022-07-28
CLDR-15858 Revert main exemplar update in my

AEApple:
CLDR-15858

Revert main exemplar update in my


AEApple:
So I checked to make sure that everything in the original exemplar from v41 is listed on page 21 of https://www.unicode.org/notes/tn11/UTN11_4.pdf

Due to some of the characters requiring combinations of unicode points, I'm not sure if there is anything that should be added, but at least this would prevent the issue that is blocking integration currently since those characters are valid unicode code points in Burmese.

pedberg-icu (approved):

sven-oly (approved):

PR #2212, opened 2022-07-29
CLDR-15874 Revert change to decimal separator

AEApple:
CLDR-15874

Revert change to decimal separator in ks_Deva

### DIFF
--- a/common/main/ks_Deva.xml
+++ b/common/main/ks_Deva.xml
@@ -733,7 +733,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<minusSign>↑↑↑</minusSign>
 		</symbols>
 		<symbols numberSystem="latn">
-			<decimal>।</decimal>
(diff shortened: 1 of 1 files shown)


AEApple:
Since this was found after the survey tool closed, do we need to do anything so that the votes from this cycle don't inherit during the next submission cycle?

pedberg-icu (approved):

pedberg-icu:
> Since this was found after the survey tool closed, do we need to do anything so that the votes from this cycle don't inherit during the next submission cycle?

Hmm, not sure. @srl295 @btangmu , any thoughts on that?

srl295:
> > Since this was found after the survey tool closed, do we need to do anything so that the votes from this cycle don't inherit during the next submission cycle?
> 
> 
> 
> Hmm, not sure. @srl295 @btangmu , any thoughts on that?

There's a ticket for admin vote for removal. Need that.

PR #2216, opened 2022-07-30
CLDR-15840 hy collation: sort ech-yiwn ligature between lowercase and titlecase ech+vew

pedberg-icu:
CLDR-15840

For hy: sort ech-yiwn ligature between lowercase and titlecase ech+vew, per [L2/22-124](https://www.unicode.org/L2/L2022/22124-utc172-properties-recs.pdf) item Coll2.


markusicu (approved):
lgtm tnx!

PR #2217, opened 2022-07-30
CLDR-15877 Add English name for bgc

AEApple:
CLDR-15877

Add English name for bgc

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -84,6 +84,7 @@ annotations.
 			<language type="bfd">Bafut</language>
 			<language type="bfq">Badaga</language>
 			<language type="bg">Bulgarian</language>
+			<language type="bgc">Haryanvi</language>
 			<language type="bgn">Western Balochi</language>
 			<language type="bho">Bhojpuri</language>
 			<language type="bi">Bislama</language>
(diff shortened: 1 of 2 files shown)


AEApple:
So neither bgc nor raj are at modern coverage. Should I add an the exception, or should we move both to modern and remove the exception for raj?

Also should we add a test so that if a locale is added as a TC org locale that we automatically add the locale name to modern since by adding the locale to locales.txt the TC org is signaling that they plan to bring the locale up to at least Basic coverage that cycle?

pedberg-icu:
> So neither bgc nor raj are at modern coverage. Should I add an the exception, or should we move both to modern and remove the exception for raj?

I would say that in general we should add them both at modern coverage and remove the exceptions. However at the end of a release that will cause the coverage statistics for all CLDR languages to drop. So for 42 I think we add the exception, then file a separate ticket for 43 to put them both at modern coverage and remove the exception.

> Also should we add a test so that if a locale is added as a TC org locale that we automatically add the locale name to modern since by adding the locale to locales.txt the TC org is signaling that they plan to bring the locale up to at least Basic coverage that cycle?

Agree, also sounds like a se[arate ticket for 43.

macchiati:
Agreed

pedberg-icu (approved):

AEApple:
For future reference filed the following tickets:

- CLDR-15883 to add language names for bgc & raj in v43
- CLDR-15882 to consider adding tests against locales.txt to make sure locales listed by TC orgs have their language names covered at modern.

Maxxisthename (approved):

PR #2222, opened 2022-08-01
CLDR-14660 he currency formats, add RLM before currency sign

pedberg-icu:
CLDR-14660

In the Hebrew standard currency format, add RLM before the currency symbol in both the positive and negative formats. As demonstrated in smoketest using the new bidi examples, this makes currency lay out the same regardless of positive of negative form, ISO 4217 symbol (strong LR) or shekel symbol, and R-L or neutral context.

And then since the accounting format did not seem to be trying to do anything special, copy the standard currency format to the accounting format.

I plan to discuss this during the CLDR TC meeting 2022-08-01.


richgillam (approved):
👍

pedberg-icu:
We did discuss this during the CLDR TC meeting 2022-08-01, and approved this approach.

PR #2215, opened 2022-07-29
CLDR-14981 make CLDRModify -fQ work for seed/annotations

srl295:
- CLDRConfig.getAnnotationsFactory now includes seed
- but, Annotations.getAvailable() only includes locales in common
- Annotations.getAllAvailable() includes seed+common
- CLDRModify has a special path which includes common/annotations when resolving
seed/annotations files, otherwise root.xml is not found.

CLDR-14981

- [X] This PR completes the ticket.

PR #2233, opened 2022-08-02
CLDR-15889 CLDRModify passes before alpha1

pedberg-icu:
CLDR-15889

ALLOW_MANY_COMMITS=true

This is for the results of the three CLDRModify passes per https://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-passes:
1. CLDRModify with no arguments against 6 directories. This only affected common/annotations, reordering some of the entries in en.xml and root.xml
2. CLDRModify -fQ against common/annotations and seed/annotations. Running against common/annotations this added many tts names to the regular keywords. Running against seed/annotations crashed; there is an outstanding [PR-2215](https://github.com/unicode-org/cldr/pull/2215) to fix that, so I just skipped running CLDRModify -fQ  against seed/annotations.
3. CLDRModify -fp against 6 directories. This produced no changes (I had previously run CLDRModify -fp against common/main and seed/main).

**Note: This should be merged with a MERGE COMMIT to preserve the two separate commits for different passes. That option needs to be enabled from the github admin panel. I can take care of this.**


srl295 (approved):
RSLGTM

> I can take care of this.

Please remember to edit the commit message on the merge, thanks!

pedberg-icu:
> Please remember to edit the commit message on the merge, thanks!

OK, I think I edited it successfully...

PR #2234, opened 2022-08-02
CLDR-15863 rename zDefault to zz-default (ruleset names should not have uppercase)

pedberg-icu:
CLDR-15863


richgillam (approved):
Thank you for the fix!

grhoten:
I think "default" would have been more appropriate. I don't see value in the "z".

pedberg-icu:
> I think "default" would have been more appropriate. I don't see value in the "z".

Is there not a fallback to the last ruleset (sorted alphabetically) in the list?

grhoten:
I sorted the rules manually in the past so that I could find the rules more easily. The rules used to be just appended to the end. There is no requirement that the rules are alphabetic. In fact, all of the other RBNF rules are not sorted. The default is really just the last rule for anything beyond spellout and ordinal.

PR #2235, opened 2022-08-02
CLDR-15866 Fix translation for LSL in ig

AEApple:
Fix "this is not a translation" names and symbol for Igbo translation of Lesotho Loti (LSL)

CLDR-15866


macchiati (approved):

pedberg-icu:
Merged to get in stuff before start of alpha1 integration

PR #2236, opened 2022-08-02
CLDR-15557 keyboard: work around XPath issue for doublequote

srl295:
- use \u{22} instead of &quot; as with other files

CLDR-15557


srl295:
this shows up in exhaustive mode only

`mvn --file=tools/pom.xml -pl cldr-code '-Dorg.unicode.cldr.unittest.testArgs=-n -q -filter:TestDtds -e10' -Dtest=org.unicode.cldr.unittest.TestShim test`

macchiati (approved):

pedberg-icu:
Merged to get in stuff before start of alpha1 integration, hope that is OK

PR #2240, opened 2022-08-02
CLDR-15359 build: always build exhaustive

srl295:
CLDR-15359

This only makes sure cldr-archive is present before mvn tests

ALLOW_MULTI_COMMITS=true


srl295:
@macchiati  this doesn't explicitly change any test code or behavior.

It does add a couple of minutes to the build, because usually the fetch only gets one commit instead of all of them, plus the time to process the checkout of all of the tags.

srl295:
we could merge this, but may not be worth it unless we use it for tests. 

Is there a way we could make the tests optionally run without requiring cldr-archive to be present for all test runs?

.github/workflows/maven.yml
```diff
@@ -17,11 +17,13 @@ jobs:
         uses: actions/checkout@v3
         with:
           lfs: false
+          fetch-depth: 0  # Needed for cldr-archive
```
btangmu, line 20: I'm not sure what this means...

I found this, not sure what the implication is:
https://github.com/changesets/action/issues/97 "This is no longer necessary: the tool will automatically deepen shallow clones as necessary"

.github/workflows/maven.yml
```diff
@@ -52,6 +54,9 @@ jobs:
           -DskipTests=true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Checkout CLDR archve
```
btangmu, line 57: typo "archve" --> "archive"

btangmu (dismissed):

.github/workflows/maven.yml
srl295: I think that's for an unrelated tool. cldr-archive does _not_ automatically deepen shallow clones as necessary.

.github/workflows/maven.yml
srl295:
```suggestion
      - name: Checkout CLDR archive
```

.github/workflows/maven.yml
btangmu: good to know! so I'm guessing zero means no limit

btangmu (approved):

macchiati (approved):

PR #2241, opened 2022-08-03
CLDR-15486 fix problems in nl, tk, and the tests.

macchiati:
CLDR-15486

This will (hopefully) fix the remaining exhaustive test failures. Some of these are workarounds, so I filed a different ticket for taking care of them when we have time (https://unicode-org.atlassian.net/browse/CLDR-15894)


macchiati:
It is ready for review, if the tests pass.

srl295 (approved):
LGTM. commented on ticket.

PR #2242, opened 2022-08-03
CLDR-13349 Chinese traditional exemplar set contains a simplified character

btangmu:
-Change U+7B80 简 to U+7C21 簡 in yue.xml and zh_Hant.xml

CLDR-13349


DavidLRowe (approved):
This pull request does exactly what it says: changes U+7B80 to U+7C21 in the main exemplar for these two locales.

PR #2244, opened 2022-08-03
CLDR-15900 BRS v42: update SLE/SLL legal tender

srl295:
- SLE became legal tender 2022-07-01
- SLL will cease to be tender 2022-10-01

This PR:
- updates metadata
- moves SLL to SLE in symbol
- add (1964-2022) to SLL in en name

References:
- https://www.six-group.com/dam/download/banking-services/interbank-clearing/en/news/sierra-leone.pdf
- https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/amendments/dl_currency_iso_amendment_172.pdf

CLDR-15900 followon to #1825 

ALLOW_MULTI_COMMITS=true
ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -5795,14 +5795,14 @@ annotations.
 				<displayName count="other">Slovak korunas</displayName>
 			</currency>
 			<currency type="SLE">
-				<displayName>Sierra Leonean New Leone</displayName>
-				<displayName count="one">Sierra Leonean new leone</displayName>
-				<displayName count="other">Sierra Leonean new leones</displayName>
-			</currency>
-			<currency type="SLL">
 				<displayName>Sierra Leonean Leone</displayName>
 				<displayName count="one">Sierra Leonean leone</displayName>
 				<displayName count="other">Sierra Leonean leones</displayName>
+			</currency>
+			<currency type="SLL">
+				<displayName>Sierra Leonean Leone (1964—2022)</displayName>
+				<displayName count="one">Sierra Leonean leone (1964—2022)</displayName>
+				<displayName count="other">Sierra Leonean leones (1964—2022)</displayName>
 			</currency>
 			<currency type="SOS">
 				<displayName>Somali Shilling</displayName>
--- a/common/main/en_AU.xml
+++ b/common/main/en_AU.xml
@@ -6711,12 +6711,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<symbol>SHP</symbol>
 				<symbol alt="narrow">↑↑↑</symbol>
 			</currency>
-			<currency type="SLL">
-				<displayName>Sierra Leonean Leone</displayName>
-				<displayName count="one">Sierra Leonean leone</displayName>
-				<displayName count="other">Sierra Leonean leones</displayName>
-				<symbol>SLL</symbol>
-			</currency>
 			<currency type="SOS">
 				<displayName>Somali Shilling</displayName>
 				<displayName count="one">Somali shilling</displayName>
--- a/common/main/en_SL.xml
+++ b/common/main/en_SL.xml
@@ -39,7 +39,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	</dates>
 	<numbers>
 		<currencies>
-			<currency type="SLL">
+			<currency type="SLE">
 				<symbol>Le</symbol>
 			</currency>
 		</currencies>
--- a/common/main/ff_Adlm_SL.xml
+++ b/common/main/ff_Adlm_SL.xml
@@ -49,7 +49,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<currency type="GNF">
 				<symbol>GNF</symbol>
 			</currency>
-			<currency type="SLL">
+			<currency type="SLE">
 				<symbol>Le</symbol>
 			</currency>
 		</currencies>
--- a/common/main/ff_Latn_SL.xml
+++ b/common/main/ff_Latn_SL.xml
@@ -46,7 +46,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	</dates>
 	<numbers>
 		<currencies>
-			<currency type="SLL">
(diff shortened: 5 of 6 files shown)


macchiati (approved):

srl295:
had to update the localization, please check

macchiati (changes requested):
Our practice is to put a date range in, not an English word. Eg ...>Le (1964-2021)<...

common/main/ff_Adlm_SL.xml
```diff
@@ -50,6 +50,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<symbol>GNF</symbol>
 			</currency>
 			<currency type="SLL">
+				<symbol>Le (old)</symbol>
```
pedberg-icu, line 53: Don't we usually do a date range like "Le (1964–2022)" instead of "(old)"?

common/main/ff_Latn_SL.xml
```diff
@@ -47,6 +47,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	<numbers>
 		<currencies>
 			<currency type="SLL">
+				<symbol>Le (old)</symbol>
```
pedberg-icu, line 50: Don't we usually do a date range like "Le (1964–2022)" instead of "(old)"?

common/main/ff_Adlm_SL.xml
srl295: yes, @macchiati had the same comment, will do

srl295:
> Our practice is to put a date range in, not an English word. Eg ...>Le (1964-2021)<...

OK will do.  also there are other translations of `SLE` that may need the same treatment

srl295:
@pedberg-icu @macchiati please go ahead and merge if this passes

pedberg-icu (dismissed):

pedberg-icu:
> > Our practice is to put a date range in, not an English word. Eg ...>Le (1964-2021)<...
> 
> OK will do. also there are other translations of `SLE` that may need the same treatment

@macchiati @srl295 so actually I think both the comment from Mark and the one from me about using a date range were meant to apply to the currency _name_, not the currency _symbol_, which iI now see is what this PR is about. And the date range is causing a too-long error. How about just removing the symbol entry for the old name so the ISO 4217 code will be used? Or putting in the ISO 4217 code explicitly as the old name?

srl295:
@pedberg-icu OK, please take a look now. thanks for the review

common/main/en_SL.xml
```diff
@@ -40,6 +40,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	<numbers>
 		<currencies>
 			<currency type="SLL">
+				<symbol draft="contributed">Le (1964-2022)</symbol>
```
pedberg-icu, line 43: Should we also remove this symbol like the others?

common/main/en.xml
```diff
@@ -5795,15 +5795,15 @@ annotations.
...
-			<currency type="SLL">
 				<displayName>Sierra Leonean Leone</displayName>
 				<displayName count="one">Sierra Leonean leone</displayName>
 				<displayName count="other">Sierra Leonean leones</displayName>
 			</currency>
+			<currency type="SLL">
+				<displayName>Sierra Leonean Leone (1964-2022)</displayName>
+				<displayName count="one">Sierra Leonean leone (1964-2022)</displayName>
+				<displayName count="other">Sierra Leonean leones (1964-2022)</displayName>
+			</currency>
```
pedberg-icu, line 5806: Good change for display name...

common/main/en_SL.xml
srl295: thanks, thought I caught them all

srl295: forgot to save that file.

common/main/en.xml
pedberg-icu: But the test seems to be complaining that these are not en-dashes....

srl295: OK, yeah trying to work on this and bits in between meetings… I'll try again

pedberg-icu: And TestSupplementalInfo.TestCurrency seems to be complaining that the name has a year range even though it is still in current use as of today< that might need a test exception.

srl295: It could check to see if this year is the final year.

pedberg-icu: Oh, good idea

srl295: @pedberg-icu Hah…  changing hyphen-minus to em-dash incorrectly quells the "Error: Has 'old' in name" because it didn't check for em-dash.

pedberg-icu: I am happy to just go with that and have a separate ticket to fix the test to check for en-dash

pedberg-icu (dismissed):

macchiati (approved):

pedberg-icu:
Aarghh. Display collision with non-updated currency name in en_AU.

srl295:
@pedberg-icu  would you be able to take over this one? I'm booked until Tuesday and I think this may be blocking you.

common/main/en_AU.xml
```diff
@@ -6711,12 +6711,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<symbol>SHP</symbol>
 				<symbol alt="narrow">↑↑↑</symbol>
 			</currency>
-			<currency type="SLL">
```
AEApple, line 6714: @srl295 - Just out of curiosity, why are we removing instead of changing to inheritance?

pedberg-icu: Well, the result is the same... and we cannot say that en_AU vetters would vote to always inherit, what we do know is they voted for an explicit value that happened to be the same (at the time) as the inherited value, _instead_ of voting for inheritance.

AEApple: Thank you for the explanation!

AEApple (approved):

pedberg-icu (approved):

PR #2252, opened 2022-08-05
CLDR-13449 ar short date formats, improve consistency

pedberg-icu:
CLDR-13449

Improve the consistency of short date formats in Arabic.
- One part of this was just making sure that in the combination of RLM and '/', the RLM consistently comes first (easier to compare formats). That fix was also related to https://unicode-org.atlassian.net/browse/CLDR-14076. Also removed unnecessary RLMs on the en-dash in a couple of intervalFormats.
- For Md intervalFormats, most greatestDifference values used d/M (with RLM), but:
    - 3 used d-M (one of these was mixed with d/M for the same skeleton).
    - 1 used M/d without RLM to try and get right layout (but with wrong field order).
    - Made them all consistently use d/M (with RLM before /).


richgillam (approved):
The diff in GitHub is useless, but I trust you...

PR #2256, opened 2022-08-09
CLDR-15767 Treat @ as ALetter in WordBreak

macchiati:
CLDR-15767

### DIFF
--- a/common/segments/root.xml
+++ b/common/segments/root.xml
@@ -348,7 +348,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 				<!-- Now normal variables -->
 				<variable id="$Format">\p{Word_Break=Format}</variable>
 				<variable id="$Katakana">\p{Word_Break=Katakana}</variable>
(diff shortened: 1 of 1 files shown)


macchiati:
Needs to be reviewed by TC first, just a draft for now.

pedberg-icu:
@macchiati Are we sure that we want to make @ ALetter? Then something like @@ ·@@ would be a word. How about making @ MidLetter? That seems to better capture the desired behavior.

Note: It looks like HYPHEN-MINUS is no longer in MidLetter (ever since its linebreak class was split out into HY). That might be an oversight.

macchiati:
I don't know whether the - was deliberate or not. For double-click (for
example), my intuition is that most of the following, having the hyphen
separate words is the right thing to do.

out-of-the-box
black-bird
pre-enable

If you think this should be investigated, we can file a ticket for that.

srl295 (approved):

PR #2257, opened 2022-08-09
CLDR-15912 un-parallelize GenerateProductionData to avoid crashes

pedberg-icu:
CLDR-15912

When running the CLDR-ICU integration, I periodically encounter hard-to-diagnose crashes in GenerateProductionData (usually NPE) that sometimes go away if I re-run it or if there is a slight data change, and go away if I temporarily de-parallelize GenerateProductionData. So I am just going to commit the de-parallelized version, the 2 or so minutes of speedup from the parallel version are not worth the occasional crashes. At some point we can investigate these crashes.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
@@ -138,8 +138,9 @@ public class GenerateProductionData {
 
         // get directories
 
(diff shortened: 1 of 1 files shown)


macchiati (approved):

srl295 (approved):
OK, and follow up in another ticket

pedberg-icu:
> OK, and follow up in another ticket

https://unicode-org.atlassian.net/browse/CLDR-15913

PR #2261, opened 2022-08-10
CLDR-14634 Adjust Hebrew plural rules

macchiati:
CLDR-14634

Make changes as per ticket, dropping many and adjusting one.

### DIFF
--- a/common/supplemental/plurals.xml
+++ b/common/supplemental/plurals.xml
@@ -155,9 +155,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
         <!-- 4: one,two,many,other -->
 
         <pluralRules locales="he iw">
-            <pluralRule count="one">i = 1 and v = 0 @integer 1</pluralRule>
+            <pluralRule count="one">i = 1 and v = 0 or i = 0 and v != 0 @integer 1</pluralRule>
             <pluralRule count="two">i = 2 and v = 0 @integer 2</pluralRule>
-            <pluralRule count="many">v = 0 and n != 0..10 and n % 10 = 0 @integer 20, 30, 40, 50, 60, 70, 80, 90, 100, 1000, 10000, 100000, 1000000, …</pluralRule>
+            <!--  <pluralRule count="many">v = 0 and n != 0..10 and n % 10 = 0 @integer 20, 30, 40, 50, 60, 70, 80, 90, 100, 1000, 10000, 100000, 1000000, …</pluralRule> -->
             <pluralRule count="other"> @integer 0, 3~17, 101, 1001, … @decimal 0.0~1.5, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
         </pluralRules>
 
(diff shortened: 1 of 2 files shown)


common/supplemental/plurals.xml
```diff
@@ -155,9 +155,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
         <!-- 4: one,two,many,other -->
```
srl295, line 155:
```suggestion
        <!-- 4: one,two,other -->
```

common/supplemental/plurals.xml
```diff
@@ -155,9 +155,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
 
         <pluralRules locales="he iw">
-            <pluralRule count="one">i = 1 and v = 0 @integer 1</pluralRule>
+            <pluralRule count="one">i = 1 and v = 0 or i = 0 and v != 0 @integer 1</pluralRule>
             <pluralRule count="two">i = 2 and v = 0 @integer 2</pluralRule>
-            <pluralRule count="many">v = 0 and n != 0..10 and n % 10 = 0 @integer 20, 30, 40, 50, 60, 70, 80, 90, 100, 1000, 10000, 100000, 1000000, …</pluralRule>
+            <!--  <pluralRule count="many">v = 0 and n != 0..10 and n % 10 = 0 @integer 20, 30, 40, 50, 60, 70, 80, 90, 100, 1000, 10000, 100000, 1000000, …</pluralRule> -->
```
srl295, line 160:
```suggestion
```

srl295 (commented):
Does this intend to apply to Yiddish (iw) also?

common/supplemental/plurals.xml
macchiati: After all of the plural rules are done, I'm going to apply the program to normalize the output

macchiati:
> Does this intend to apply to Yiddish (iw) also?

No, that language has Germanic plurals.

common/supplemental/plurals.xml
macchiati: When the program normalizes the file, that will disappear; it is just for comparison.

srl295:
> > Does this intend to apply to Yiddish (iw) also?
> 
> No, that language has Germanic plurals.

OK.  Then the plural rules need to be split out to reflect that, as `iw` is affected as the PR stands.

macchiati:
Yiddish has the code 'yi'.

The rules are for Hebrew with either the old or new code for:         <
pluralRules locales=*"he iw"*>
You know that, so I suspect your eye just confused 'iw' with 'yi'.

srl295:
> Yiddish has the code 'yi'. The rules are for Hebrew with either the old or new code for: < pluralRules locales=*"he iw"*> You know that, so I suspect your eye just confused 'iw' with 'yi'.
> […](#)

I think you're right.

srl295 (approved):

common/supplemental/plurals.xml
AEApple, line 160: I would like to understand why you are commenting out the plural rule for "many" instead of removing it?

srl295: @macchiati explained that this is for comparison and will be removed by later automated processes.

PR #2262, opened 2022-08-10
CLDR-14421 Change metric-ton to tonne

macchiati:
CLDR-14421

This makes the changes from the ticket.

1. Deprecates metric-ton in favor of tonne
2. Adds structure to ensure that we don't get future ambiguities. Most of the work was here and in the tests.

The test code ensures that none of the defined Ids have the ambiguity. The structure in the units.xml file is currently commented out (so the data is hard-coded in the test file).

Most of the changes (by number of files) are changing metric-ton to tonne. The test code checks that the deprecation works, and that the new structural elements for test work.


common/supplemental/units.xml
```diff
@@ -8,6 +8,15 @@ For terms of use, see http://www.unicode.org/copyright.html
 
 <supplementalData>
     <version number="$Revision$"/>
+    <!--  
+    <unitIdComponents>
```
srl295, line 12: Why is this here if it's commented out? Is it meant for later, or backed out?

macchiati: I'm getting a strange error that I can't figure out. Will be sending out a note, but wanted to get the bulk of this in.

PR #2264, opened 2022-08-10
CLDR-13808 change default Hans→Hans conversion of 脏, add exceptions as necessary

pedberg-icu:
CLDR-13808

Change the default Hans→Hant conversion of 脏 from `臟` to 髒, add as exceptions compounds in which 脏 _should_ convert to 臟 (check that the other characters in the compounds convert correctly as well).


btangmu (approved):
there are other exceptions such as 腑脏→腑臟...

pedberg-icu:
> there are other exceptions such as 腑脏→腑臟...

Thanks, I will file another ticket to add more....

PR #2266, opened 2022-08-10
CLDR-13043 bg currency patterns should use grouping separator

pedberg-icu:
CLDR-13043

bg currency patterns should use grouping separator


sffc (approved):
I think I used this locale for testing no-grouping behavior once, but I guess we won't be able to do that any more :)

PR #2270, opened 2022-08-11
CLDR-14421 fix XML problem in unitIdComponents

macchiati:
The problem was a hidden NBSP.
Also, hide the display of information in TestUnits under the flags. This drastically reduces the number of warnings for the normal case.

CLDR-14421


srl295:
- unitIdComponent regex may need to be `[a-z][a-z0-9]*`
- pathheader missing

macchiati:
Made the changes you suggested.

srl295 (approved):

srl295:
seems to build now, too!

PR #2271, opened 2022-08-11
CLDR-15191 test issues

srl295:
CLDR-15191


srl295:
- improve transliterator test
- morse transliterator

What is the right test file name? und-t-d0-morse.txt is what I am trying but something (maybe ICU)  does not like that as a transliterator.

@macchiati any idea on this?

PR #2272, opened 2022-08-11
CLDR-14857 fix light-year, parsec

macchiati:
CLDR-14857


srl295 (approved):
I see we are not adding light-week, etc yet (wasn't clear from the ticket) but this change LGTM

PR #2283, opened 2022-08-16
CLDR-13542 revert change from apostrophe grouping for Swiss locales

macchiati:
CLDR-13542

This has the effect of reverting https://unicode-org.atlassian.net/browse/CLDR-9865, as in https://github.com/unicode-org/cldr/commit/abb5ba2f0b8b8d3fc3dc16c9ecd4abcbba0bf2dc#diff-873bba593c30f754801f660191b8b500L60


macchiati:
There is some question as to whether to do this: see the ticket.

PR #2285, opened 2022-08-16
CLDR-15932 issues from prod test: copy empty annotation parents, fix en_AU/oc_ES/sw_KE

pedberg-icu:
CLDR-15932

Fix issues from running tests on the production version of the data (BRS task):
1. `en_AU` fluid ounces: There was a display name collision because `en_AU` was inheriting from `en_001` the plain name "fluid ounce" for `volume-fluid-ounce-imperial`, but also defined the plain name "fluid ounce" for `volume-fluid-ounce` (i.e. the US fluid ounce). Although `en_AU` is mostly metric (more so than `en_GB`), per https://www.legislation.gov.au/Details/F2012C00400 the fluid ounce they do use is the US one, i.e. 0.0000295735296 cubic meters. So changed the display names for `volume-fluid-ounce-imperial` to explicitly use "Imperial".
2. `oc_ES` missing currency symbols: There were two cases in which `oc_ES` was overriding a narrow currency symbol inherited from `os`, while inheriting (with ↑↑↑) what `os` used for the standard currency symbol, which was just the ISO 4217 code inherited from root. But when the ↑↑↑ entry is deleted for production data, there is an error for alt entry without the non-alt entry. So just explicitly specified the ISO 4217 code as the standard currency symbol.
3. `sw_KE` display name collisions,  delete problem entries to inherit from `sw`:
    * For Hant, `sw_KE` was using something akin to "common" (which conflicted with the name for Zyyy) instead of "simplified".
    * For 018 Southern Africa, name was colliding with ZA South Africa
    * For 419 Latin American, name was colliding with 005 South America
4. GenerateProductionData fix: The unit test TestCLDRFile/TestFileIds tests that for annotation files, not only is the true parent present, but also the "simple" parent (i.e. from subtag stripping) is present, if different. And for that reason we added an empty annotations/ff.xml file. However GenerateProductionData did not copy empty files that had no children, resulting in the following error on production data: `Error: (TestCLDRFile.java:743) Missing simple parent (ff) for ff_Adlm  in common/annotations; likely=ff_Adlm`. So the solution is to update GenerateProductionData so that for annotation files, it ensures that empty files are copied if they are either the true or simple (truncation) parent of a file with content.

### DIFF
--- a/common/main/en_AU.xml
+++ b/common/main/en_AU.xml
@@ -7938,9 +7938,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<unitPattern count="other">{0} fluid ounces</unitPattern>
 			</unit>
 			<unit type="volume-fluid-ounce-imperial">
-				<displayName>↑↑↑</displayName>
-				<unitPattern count="one">↑↑↑</unitPattern>
-				<unitPattern count="other">↑↑↑</unitPattern>
+				<displayName>Imp. fluid ounces</displayName>
+				<unitPattern count="one">{0} Imp. fluid ounce</unitPattern>
+				<unitPattern count="other">{0} Imp. fluid ounces</unitPattern>
 			</unit>
 			<unit type="volume-tablespoon">
 				<displayName>tablespoons</displayName>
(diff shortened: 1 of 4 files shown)


srl295 (approved):
LGTM but question: should there be tests to make sure these get caught before it goes to production?

pedberg-icu:
> LGTM but question: should there be tests to make sure these get caught before it goes to production?

Yes. Some of the locale file issues I think already generate warning, some should be errors. I will file a ticket. The GenerateProductionData issue was a problem in copying to production, not a problem in the data itself.

PR #2286, opened 2022-08-16
CLDR-15924 finalize changes to person names

macchiati:
CLDR-15924

### DIFF
--- a/common/main/ff.xml
+++ b/common/main/ff.xml
@@ -821,4 +821,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<nostr>Alaa:A</nostr>
 		</messages>
 	</posix>
+	<personNames>
+		<nameOrderLocales order="givenFirst">und ff</nameOrderLocales>
+		<nameOrderLocales order="surnameFirst">ko vi yue zh</nameOrderLocales>
+	</personNames>
 </ldml>
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
@@ -1,25 +1,151 @@
 # constructed line for modify_config.txt file
-locale=tk ; action=addNew ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="1"] ; new_value=Aşy
-locale=tk ; action=addNew ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="2"] ; new_value=Sap
-locale=tk ; action=addNew ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="3"] ; new_value=Tir I
-locale=tk ; action=addNew ; new_path=//ldml/dates/calendars/calendar[@type="islamic"]/months/monthContext[@type="format"]/monthWidth[@type="abbreviated"]/month[@type="4"] ; new_value=Tir II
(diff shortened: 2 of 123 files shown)


macchiati:
If the tests pass, I think we can go ahead and merge this much, then follow up with a different PR for Mike's changes.

common/main/ff.xml
```diff
@@ -821,4 +821,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<nostr>Alaa:A</nostr>
 		</messages>
 	</posix>
+	<personNames>
+		<nameOrderLocales order="givenFirst">und ff</nameOrderLocales>
+		<nameOrderLocales order="surnameFirst">ko vi yue zh</nameOrderLocales>
+	</personNames>
```
srl295, line 827: Just curious… what value does this add to add this to several locales?  

There doesn't seem to be any documentation for these elements in the spec.  But does does `<nameOrderLocales order="givenFirst">und</nameOrderLocales>` already apply to all locales?  Why add this to, say, ff?

The "surnameFirst" data is also in root, is there a reason to repeat it here?

tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
```diff
@@ -1,25 +1,151 @@
...
+locale=te ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko te vi yue zh
+locale=ti ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko te vi yue zh
+locale=am ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=ar ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=ast ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=az ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=be ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=bg ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=bn ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
+locale=br ; action=add ; new_path=//ldml/personNames/nameOrderLocales[@order="surnameFirst"]; new_value=ko vi yue zh
```
srl295, line 53: same comment. maybe a better data model is needed if this information is repeated in each locale?

srl295 (commented):
Changes seem OK, I do have questions about the repetitiveness of the data model.  

Elements seem to be undocumented.

srl295 (approved):

srl295 (approved):

tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
macchiati: There is a lot of commonality among the locales, but each one is a separate decision, based on what is customary in the locale.

common/main/ff.xml
macchiati: These are in the design doc (haven't made it into LDML yet, but will). They say:

For formatting these languages in this locale, use the surnameFirst or givenFirst patterns, respectively.

macchiati:
Just fixed a last (hopefully) typo.

srl295 (approved):

common/main/ff.xml
srl295: Perhaps the spec could link to the design doc as an interim step? 

My specific question was if `und ff` was any different than `und` where ff is not mentioned.

PR #2287, opened 2022-08-17
CLDR-15452 Replace FixedDecimal with DecimalQuantity to gen samples

echeran:
CLDR-15452

This PR addresses the work described in the ticket in [this comment](https://unicode-org.atlassian.net/browse/CLDR-15452?focusedCommentId=166447).

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratedPluralSamples.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratedPluralSamples.java
@@ -28,8 +28,9 @@ import org.unicode.cldr.util.TempPrintWriter;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
+import com.ibm.icu.impl.number.DecimalQuantity;
+import com.ibm.icu.impl.number.DecimalQuantity_DualStorageBCD;
 import com.ibm.icu.impl.Relation;
-import com.ibm.icu.text.FixedDecimal;
 import com.ibm.icu.text.PluralRules;
 import com.ibm.icu.text.PluralRules.Operand;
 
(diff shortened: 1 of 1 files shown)


echeran:
@macchiati @sffc Is there a way to test this? Can you take a look?  Because even though I've made changes in this PR to address the ticket, I'm not entirely sure how to know that it's doing the correct thing.

macchiati:
This program reformats the plurals.xml file to add/replace the samples. So the test to make sure it is working is to run it, then verify that the results are identical to the HEAD file.

Thanks!

DavidLRowe:
@echeran This PR seems to be failing the single-commit test because there are two commits, and the jira-ticket test because the commit message doesn't begin with the CLDR ticket number. I think you can click on the "Details" link to the right of the single-commit entry, and use the Squash branch button. Be sure that the commit message then begins with "CLDR-15452". That should rerun all the tests.

echeran:
> @echeran This PR seems to be failing the single-commit test 

The real correctness test, as described above, is being able to run the tool in question in this PR (`GeneratedPluralSamples`).  I've been waiting until the PR's logical correctness bugs get sorted out before squashing the commits into one.  Mark is going to help debug what the correct logic should be to fix the failing assertions that I'm still seeing when I try to run the tool with the current state of this PR (details in email).

I created [CLDR-15947](https://unicode-org.atlassian.net/browse/CLDR-15947) in order to represent the generalized need to have implicit correctness tests become explicit and plugged into automatic test infrastructure.

macchiati:
Thanks for filing that.

I've traced it down to the following (line 337)
When the samples are computed, it calls *boolean* isKnownBounded = rule.
computeLimited(keyword, sampleType);
Where the rules are for be, and the keyword is 'few'.

one: n % 10 = 1 and n % 100 != 11 @integer 1, 21, 31, 41, 51, 61, 71, 81,
101, 1001, … @decimal 1.0, 21.0, 31.0, 41.0, 51.0, 61.0, 71.0, 81.0, 101.0,
1001.0, …;

few: n % 10 = 2..4 and n % 100 != 12..14 @integer 2~4, 22~24, 32~34, 42~44,
52~54, 62, 102, 1002, … @decimal 2.0, 3.0, 4.0, 22.0, 23.0, 24.0, 32.0,
33.0, 102.0, 1002.0, …;

many: n % 10 = 0 or n % 10 = 5..9 or n % 100 = 11..14 @integer 0, 5~19,
100, 1000, 10000, 100000, 1000000, … @decimal 0.0, 5.0, 6.0, 7.0, 8.0, 9.0,
10.0, 11.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …;

other:  @decimal 0.1~0.9, 1.1~1.7, 10.1, 100.1, 1000.1, …
That is correctly returning false.
But countNoTrailing is zero, and that shouldn't be happening.

Will have to pick this up again in 30 minutes

echeran:
FYI, you want to `git pull` my new commit that prevents a runtime error that I didn't catch. It's a more via a more mechanical fix that doesn't appear to change much about the logic invariant failures.

macchiati:
will do

pedberg-icu:
This is the last thing I am waiting on before starting the CLDR-ICU integration (because changes to the compact decimal stuff either cause new problems or solve existing ones...)

macchiati:
Peter, it looks like the work for Elango's work is too complicated to do in
limited time. So please don't wait for it; go ahead with the integration.

pedberg-icu:
> Peter, it looks like the work for Elango's work is too complicated to do in limited time. So please don't wait for it; go ahead with the integration.

OK, will do, thanks

PR #2291, opened 2022-08-17
CLDR-15938 Run CLDRModify passes before alpha2

pedberg-icu:
CLDR-15938

ALLOW_MANY_COMMITS=true

This is for the results of the three CLDRModify passes per https://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-passes:

1. CLDRModify with no arguments against 6 directories. This produced no changes.
2. CLDRModify -fQ against common/annotations and seed/annotations. This added tts names to the regular keywords (but also seemed to remove some keywords for de_CH). 
3. CLDRModify -fp against 6 directories. This re-ordered exemplars for yue and Hant, and updated nameOrderLocales for gaa.

**Note: This should be merged with a MERGE COMMIT to preserve the two separate commits for different passes. That option needs to be enabled from the github admin panel. I can take care of this (plus editing the commit message on the merge).**


srl295 (approved):
RSLGTM

PR #2296, opened 2022-08-18
CLDR-15931 Update HCL list in locales.txt

AEApple:
CLDR-15931

- Add additional locales to high-coverage locales list which meet the coverage criteria
- Add new HCLs to the special org
- Remove hi_Latn from the special org and add to Google org list


AEApple:
> You need to set the new lines also for the "special" locale at the end of the file, to make sure that we don't think they are monitored.

Updated the special org. Also removed hi_Latn since I don't think it belongs since it's a locale that is vetted by one of the TC orgs.

macchiati (approved):

macchiati (approved):

PR #2300, opened 2022-08-18
CLDR-15945 Updating tzdata to 2022b

yumaoka:
CLDR-15945


pedberg-icu (approved):

yumaoka:
OK - some test cases are sensitive for the change.
I could not run CLDR full build on my local machine - it causes a crash when I run mvn verify.
Anyway, I'll fix the test errors.

yumaoka:
```aa [Afar]	error	▸Timezones|Unknown_Region|Cities_and_Regions|Kyiv◂	〈Kyiv〉	【】	〈Kyiv〉	«=»	【】	⁅display collision⁆	❮Error: Can't have same translation as [Cities and Regions: Kiev]. Please change either this name or the other one. See <a target='doc' href='http://cldr.unicode.org/translation/short-names-and-keywords#TOC-Unique-Names'>Unique-Names</a>.❯	https://st.unicode.org/cldr-apps/v#/aa//5750f6e448df159	code-fallback```

Same error for many locales. It looks it complains timezone region name "Kyiv" is conflicting with fallback region name of zone ID Europe/Kyiv. But this new ID is not a CLDR canonical zone ID. So this should not be handled as conflict.

File: tools/cldr-code/src/main/java/org/unicode/cldr/util/ZoneParser.java [hunk 1]
```diff
@@ -672,6 +672,7 @@ private List<String> getData(String s) {
             { "Asia/Ho_Chi_Minh", "Asia/Saigon" },
             { "Asia/Yangon", "Asia/Rangoon" },
             { "Asia/Kathmandu", "Asia/Katmandu" },
+            { "Europe/Kyiv", "Europe/Kiev" },
```
yumaoka, line 675: @pedberg-icu I figured out we need new/old zone ID mapping in this code. This code is used for collecting zone IDs from tzdata files.

pedberg-icu: Ah, thanks for figuring this out. Sorry, I have been in meetings all afternoon...

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/ZoneParser.java
```diff
@@ -545,7 +545,7 @@ public Comparator<String> getTZIDComparator() {
     private static List<String> errorData = Arrays.asList(new String[] {
         new Double(Double.MIN_VALUE).toString(), new Double(Double.MIN_VALUE).toString(), "" });
 
-    private Comparator<String> TZIDComparator = new Comparator<String>() {
```
Yqwed, line 548: Was this change intentional?

On old Android branches we use Java 8 and this won't compile there. It is easy to revert this change, so I am just curious.

Thanks!

PR #2302, opened 2022-08-19
unicode-org/cldr

dextersarrosa:
base: AEpple-CLDR-15874___


dextersarrosa:
AEApple-CLDR-15874

PR #2306, opened 2022-08-22
CLDR-15957 rdf: fix logging, speedup query

srl295:
- speedup childToParent query a few orders of magnitude
- also enable JDK 1.4 logging from Jena
(this matches what we do with OpenLiberty)

The actual data update will be a separate PR.

CLDR-15957

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/Containment.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/Containment.java
@@ -214,11 +214,19 @@ public class Containment {
             targets.add(target);
         } else if (size == 1) {
             for (String parent : parents) {
-                getAllDirected(multimap, parent, target, targets);
+                if (parent.equals(lang)) {
+                    System.err.println("ERR: "+ lang + " is its own parent");
+                } else {
+                    getAllDirected(multimap, parent, target, targets);
+                }
             }
         } else {
             for (String parent : parents) {
-                getAllDirected(multimap, parent, (ArrayList<String>) target.clone(), targets);
+                if (parent.equals(lang)) {
+                    System.err.println("ERR: "+ lang + " is its own parent");
+                } else {
+                    getAllDirected(multimap, parent, (ArrayList<String>) target.clone(), targets);
+                }
             }
         }
     }
--- a/tools/cldr-rdf/src/main/resources/org/unicode/cldr/rdf/sparql/wikidata-childToParent.sparql
+++ b/tools/cldr-rdf/src/main/resources/org/unicode/cldr/rdf/sparql/wikidata-childToParent.sparql
@@ -13,14 +13,18 @@ PREFIX bd: <http://www.bigdata.com/rdf#>
 
 
 SELECT DISTINCT ?child ?parent WHERE {
-  ?child p:P279/ps:P279 ?parent.    #child is subclass of parent
   {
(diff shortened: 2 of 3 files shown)


tools/cldr-rdf/src/main/resources/org/unicode/cldr/rdf/sparql/wikidata-childToParent.sparql
```diff
@@ -13,14 +13,18 @@ PREFIX bd: <http://www.bigdata.com/rdf#>
 
 
 SELECT DISTINCT ?child ?parent WHERE {
-  ?child p:P279/ps:P279 ?parent.    #child is subclass of parent
```
srl295, line 16: @macchiati  this line referenced 3.3MM rows.  I moved it into the dependent union items

tools/cldr-rdf/src/main/resources/org/unicode/cldr/rdf/sparql/wikidata-childToParent.sparql
```diff
@@ -13,14 +13,18 @@ PREFIX bd: <http://www.bigdata.com/rdf#>
...
     ?child wdt:P31 wd:Q25295.  #child is language family, OR
   } UNION {
+    ?child p:P279/ps:P279 ?parent.    #child is subclass of parent
     ?parent wdt:P31 wd:Q34770. #parent is language, OR
   } UNION {
+    ?child p:P279/ps:P279 ?parent.    #child is subclass of parent
     ?child wdt:P31 wd:Q34770.  #child is language
   }
 }
+ORDER BY ?parent ?child
```
srl295, line 30: It's already getting distinct items, so ordering does not slow it down a lot

tools/cldr-code/src/main/java/org/unicode/cldr/util/Containment.java
```diff
@@ -214,11 +214,19 @@ private static void getAllDirected(Multimap<String, String> multimap, String lan
...
+                }
             }
         } else {
             for (String parent : parents) {
-                getAllDirected(multimap, parent, (ArrayList<String>) target.clone(), targets);
+                if (parent.equals(lang)) {
+                    System.err.println("ERR: "+ lang + " is its own parent");
+                } else {
+                    getAllDirected(multimap, parent, (ArrayList<String>) target.clone(), targets);
+                }
```
srl295, line 229: https://www.wikidata.org/wiki/Q2873230 is a subcategory of itself

srl295: I assume this is a bug, but asked https://www.wikidata.org/wiki/Property_talk:P279#P279_is_not_reflexive

macchiati: The wikidata often contains cycles (they have no test that ensure a clean tree).

macchiati: Looks like an excellent speedup.

tools/cldr-rdf/src/main/resources/org/unicode/cldr/rdf/sparql/wikidata-childToParent.sparql
macchiati: Sounds good; I didn't know about that option

macchiati (approved):

PR #2309, opened 2022-08-23
CLDR-15962 Update supplementalData's nameOrder

macchiati:
CLDR-15962

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -5412,8 +5412,8 @@ XXX Code for transations where no currency is involved
 	</parentLocales>
 
 	<personNamesDefaults>
-		<nameOrderLocalesDefault order="givenFirst">en und</nameOrderLocalesDefault>
-		<nameOrderLocalesDefault order="surnameFirst">hu ja ko vi yue zh</nameOrderLocalesDefault>
+		<nameOrderLocalesDefault order="givenFirst">und</nameOrderLocalesDefault>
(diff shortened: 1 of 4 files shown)


richgillam (dismissed):
I *think* this looks good, but I frankly don't know enough about what's going on here to be absolutely sure.

macchiati:
We have some supplemental data that is supposed to tell us which locales are themselves surnameFirst. This checks that if we look at the locale date, no locale X has X in its surnameFirst list, if X is not in the supplemental data.

Is that at all clearer?

macchiati:
As a concrete case, the supplemental data was saying that Tamil was givenFirst, but the Tamil locale (ta.xml) was formatting Tamil names as surnameFirst. That was a whoops, so I put in a test and updated the data so it doesn't happen for Tamil (or similar other ones).

richgillam:
>We have some supplemental data that is supposed to tell us which locales are themselves surnameFirst. This checks that if we look at the locale date, no locale X has X in its surnameFirst list, if X is not in the supplemental data.
>
>Is that at all clearer?

Yes, that helps a lot.  I didn't know we had that.

>As a concrete case, the supplemental data was saying that Tamil was givenFirst, but the Tamil locale (ta.xml) was formatting Tamil names as surnameFirst. That was a whoops, so I put in a test and updated the data so it doesn't happen for Tamil (or similar other ones).

Makes sense.

richgillam (approved):
Looks good.

PR #2310, opened 2022-08-24
CLDR-15903 update charts

macchiati:
CLDR-15903

### DIFF
--- a/common/properties/coverageLevels.txt
+++ b/common/properties/coverageLevels.txt
@@ -16,6 +16,8 @@ ast ;	basic
 az ;	modern
 be ;	modern
 bg ;	modern
+bgc ;	basic
+bho ;	basic
 bn ;	modern
 br ;	moderate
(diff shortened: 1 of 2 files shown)


macchiati:
The charts software also generates coverageLevels.txt

common/properties/coverageLevels.txt
```diff
@@ -95,6 +97,7 @@ pl ;	modern
 ps ;	modern
 pt ;	modern
 qu ;	moderate
+raj ;	basic
```
btangmu, line 100: I wonder if this is related to https://unicode-org.atlassian.net/browse/CLDR-15954

btangmu (approved):

srl295:
i should rerun my chart build after this i assume

macchiati:
Yes, but let's wait until some of the other items come (like containment,...)

PR #2312, opened 2022-08-24
CLDR-15921 BRS v42: regen likely subtags

srl295:
CLDR-15921

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -713,9 +713,9 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="fan" to="fan_Latn_GQ"/>
 		<!--{ Fang; ?; ? } => { Fang; Latin; Equatorial Guinea }-->
 		<likelySubtag from="ff" to="ff_Latn_SN"/>
-		<!--{ Fulah; ?; ? } => { Fulah; Latin; Senegal }-->
+		<!--{ Fula; ?; ? } => { Fula; Latin; Senegal }-->
 		<likelySubtag from="ff_Adlm" to="ff_Adlm_GN"/>
-		<!--{ Fulah; Adlam; ? } => { Fulah; Adlam; Guinea }-->
+		<!--{ Fula; Adlam; ? } => { Fula; Adlam; Guinea }-->
 		<likelySubtag from="ffi" to="ffi_Latn_ZZ"/>
 		<!--{ Foia Foia; ?; ? } => { Foia Foia; Latin; Unknown Region }-->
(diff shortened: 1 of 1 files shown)


common/supplemental/likelySubtags.xml
```diff
@@ -3550,6 +3550,24 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ ?; Lanna; ? } => { Northern Thai; Lanna; Thailand }-->
 		<likelySubtag from="und_Laoo" to="lo_Laoo_LA"/>
 		<!--{ ?; Lao; ? } => { Lao; Lao; Laos }-->
+		<likelySubtag from="und_Laoo_AU" to="hnj_Laoo_AU"/>
```
macchiati, line 3553: I think this is a bad result. Just from quick checks, it appears that there are far more Lao speakers in these countries than there are Hmong Njua speakers. It is just that we haven't had anyone submit data for those. I suggest that we remove these for now, and I'll file a ticket to work on a longer term solution.

common/supplemental/likelySubtags.xml
```diff
@@ -2139,7 +2139,7 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="saf" to="saf_Latn_GH"/>
 		<!--{ Safaliba; ?; ? } => { Safaliba; Latin; Ghana }-->
 		<likelySubtag from="sah" to="sah_Cyrl_RU"/>
-		<!--{ Sakha; ?; ? } => { Sakha; Cyrillic; Russia }-->
+		<!--{ Yakut; ?; ? } => { Yakut; Cyrillic; Russia }-->
```
btangmu, line 2142: At first I thought there was a problem with alphabetical order here; now I see the ordering is by code, and "Yakut" has the code "saq"; "Yakut, also known as Yakutian, Sakha, Saqa or Saxa" per https://en.wikipedia.org/wiki/Yakut_language so OK!

btangmu (dismissed):
alphabetize Yakut?

common/supplemental/likelySubtags.xml
macchiati: I opened a ticket for possible approach for the longer term.

macchiati (changes requested):

common/supplemental/likelySubtags.xml
srl295: makes sense.

There are still some manual reverts here already. I can file a ticket with those to make sure they are tracked.

srl295: removed, ptal

macchiati (approved):

PR #2313, opened 2022-08-26
CLDR-15966 ja-JP numbers ≤ 99999999999999999999

glebm:
Previously, the maximum supported number was:
999999999999999

Adds support for numbers up to:
99999999999999999999

CLDR-15966

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -1421,7 +1421,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT pattern ( #PCDATA ) >
 <!ATTLIST pattern type NMTOKEN "standard" >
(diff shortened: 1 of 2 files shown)


macchiati:
The tests failed, because the values exceeded what was planned for.

You can try changing ldml.dtd line 1424 from:
```
<!ELEMENT pattern ( #PCDATA ) >
<!ATTLIST pattern type NMTOKEN "standard" >
    <!--@MATCH:literal/1000, 10000, 100000, 1000000, 10000000, 100000000, 1000000000, 10000000000, 100000000000, 1000000000000, 10000000000000, 100000000000000, approximately, atLeast, atMost, range, standard-->

```
to add the new values from 1000000000000000 to 10000000000000000000

I checked, and we are still within a 64 bit long int (probably not a requirement by clients, but this late in the release cycle I don't want any changes that would trigger a problem in clients). 

Can you make that change and commit it?

BTW, I don't think you actually need the lines beyond the following
```
<pattern type="1000000000000000" count="other">0000兆</pattern>
<pattern type="10000000000000000" count="other">0京</pattern>
```
because the software uses the highest pattern for any larger value. But it shouldn't hurt to add them.

glebm:
Thanks, updated ldml.dtd

macchiati (approved):

PR #2322, opened 2022-08-31
CLDR-15963 Missing country name should show in Dashboard

btangmu:
-Fall back to use secondary territories if there are no target territories

CLDR-15963

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
@@ -2773,14 +2773,19 @@ public class SupplementalDataInfo {
 
     private Set<String> getTargetTerritories(String language) {
         Set<String> targetTerritories = new HashSet<>();
+        Set<String> secondaryTerritories = new HashSet<>();
         try {
             Set<BasicLanguageData> langData = getBasicLanguageData(language);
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -2794,7 +2799,11 @@ private Set<String> getTargetTerritories(String language) {
             // fall through
         }
         if (targetTerritories.size() == 0) {
-            targetTerritories.add("ZZ");
+            if (secondaryTerritories.size() > 0) {
+                targetTerritories.addAll(secondaryTerritories);
```
macchiati, line 2803: Looks better, but this could add multiple secondary territories when we only really want the top one. Best would be to use LikelySubtags first. Something like:

```
    if (targetTerritories.size() == 0) {
        String region = null;
        String maximized = new LikelySubtags().maximize("raj");
        if (maximized != null) {
            CLDRLocale cldrLocale = CLDRLocale.getInstance(maximized);
            region = cldrLocale.getCountry();
        }
        if (region != null) {
            targetTerritories.addAll(region);
        } else if (secondaryTerritories.size() > 0) {
            targetTerritories.addAll(secondaryTerritories);
```

macchiati (changes requested):

btangmu:
see https://github.com/unicode-org/cldr/pull/2358

PR #2328, opened 2022-09-02
CLDR-15976 Remove some occurrences of ZERO WIDTH SPACE

btangmu:
-Add feature to DAIP to remove ZWS from only two locales for now: da and fr

-Apply the resulting changes to 3 xml files

-Comments

CLDR-15976

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
@@ -311,7 +311,6 @@ public class DisplayAndInputProcessor {
      *
      * @param path
      * @param value
-     * @param fullPath
      * @return
      */
     public synchronized String processForDisplay(String path, String value) {
@@ -377,8 +376,6 @@ public class DisplayAndInputProcessor {
      * @param path
      * @param value
      * @param internalException
-     *            TODO
-     * @param fullPath
      * @return
      */
     public synchronized String processInput(String path, String value, Exception[] internalException) {
@@ -541,7 +538,7 @@ public class DisplayAndInputProcessor {
                     value = annotationsForDisplay(value);
                 }
             }
-
+            value = normalizeZeroWidthSpace(value);
             return value;
         } catch (RuntimeException e) {
             if (internalException != null) {
@@ -1217,5 +1214,32 @@ public class DisplayAndInputProcessor {
             }
             return false;
         }
+    }
+
+    private static final Pattern ZERO_WIDTH_SPACES = PatternCache.get("\\u200B+");
+    private static final Set<String> LOCALES_NOT_ALLOWING_ZWS = new HashSet<>(Arrays.asList("da", "fr"));
+
+    /**
+     * Remove occurrences of U+200B ZERO_WIDTH_SPACE under certain conditions
+     *
+     * @param value the value to be normalized
+     * @return the normalized value
+     *
+     * TODO: extend this method to address more concerns, after clarifying the conditions
+     *   - enlarge the set LOCALES_NOT_ALLOWING_ZWS?
+     *   - strip initial and final ZWS in all locales?
+     *   - reduce two or more adjacent ZWS to one ZWS?
+     *   - allow or prohibit ZWS by itself as currency symbol, as currently in locales kea, pt_CV, pt_PT
+     *   - allow or prohibit ZWS preceding URL as in "as per [U+200B]http://unicode.org/repos/cldr/trunk/specs/ldml/tr35-general.html#Annotations"
+     * Reference: https://unicode-org.atlassian.net/browse/CLDR-15976
+     */
+    private String normalizeZeroWidthSpace(String value) {
+        if (ZERO_WIDTH_SPACES.matcher(value).find()) {
+            final String localeId = locale.getBaseName();
+            if (LOCALES_NOT_ALLOWING_ZWS.contains(localeId)) {
+                value = ZERO_WIDTH_SPACES.matcher(value).replaceAll("");
+            }
+        }
+        return value;
     }
 }
--- a/common/annotations/fr.xml
+++ b/common/annotations/fr.xml
@@ -297,8 +297,8 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="⇓" type="tts" draft="contributed">flèche double vers le bas</annotation>
 		<annotation cp="⇔" draft="contributed">flèche double vers la gauche et la droite</annotation>
 		<annotation cp="⇔" type="tts" draft="contributed">flèche double vers la gauche et la droite</annotation>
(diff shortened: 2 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -1218,4 +1215,31 @@ private static boolean wantsNNBSP(String path) {
...
+     *   - strip initial and final ZWS in all locales?
+     *   - reduce two or more adjacent ZWS to one ZWS?
+     *   - allow or prohibit ZWS by itself as currency symbol, as currently in locales kea, pt_CV, pt_PT
+     *   - allow or prohibit ZWS preceding URL as in "as per [U+200B]http://unicode.org/repos/cldr/trunk/specs/ldml/tr35-general.html#Annotations"
+     * Reference: https://unicode-org.atlassian.net/browse/CLDR-15976
+     */
+    private String normalizeZeroWidthSpace(String value) {
+        if (ZERO_WIDTH_SPACES.matcher(value).find()) {
+            final String localeId = locale.getBaseName();
+            if (LOCALES_NOT_ALLOWING_ZWS.contains(localeId)) {
```
markusicu, line 1239: This excludes fixing fr-CA etc., right? Maybe check only the language subtag, or language+script?

btangmu: That's a good idea for how this code could be improved in v43. As it stands for last-minute correction in v42, there's no actual occurrence of ZWS in fr_CA or other sublocales of "fr" or "da". For v43, I suggest we investigate possible mechanisms for getting a much much larger more comprehensive set of LOCALES_NOT_ALLOWING_ZWS. Might there be cases where variants of the same language differ in whether they should be included in the set?

srl295: @markusicu  true, but da and fr are not the only locales that should exclude this.  I think the algorithm is to be improved in the future.


No other main/*.xml files had unexpected zwsp, except root.xml has it in the garbled part of the top comment block (!!)

srl295 (approved):
OK but the list of locales needs to be improved in the future

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
srl295: @btangmu 

> Might there be cases where variants of the same language differ in whether they should be included in the set?

Does not seem likely.

srl295:
@btangmu  actually there are two other cases:

- pt_PT.xml currency symbol for PTE
- pt_CV currency symbol for CVE

and they've both been there a long time.  In both cases, these are using a currency decimal, so : `​1,234$56` for 1,234.56 units.  So the ZWSP seems to stand in for a 0-length symbol.

btangmu:
@srl295 "the ZWSP seems to stand in for a 0-length symbol" -- I'm still totally mystified by the currency occurrences; do you mean they might be intentional and/or serve some purpose?

btangmu:
New ticket for v43 follow-up: https://unicode-org.atlassian.net/browse/CLDR-15976

srl295:
> @srl295 "the ZWSP seems to stand in for a 0-length symbol" -- I'm still totally mystified by the currency occurrences; do you mean they might be intentional and/or serve some purpose?

Yes, I think it's intentional.   Note that the PTE is obsolete (for the Euro).

It's a currency format where something replaces the decimal point, in this case a `$`

So instead of `$1.23` it is `1$23` (a 'dollar'(or whatever) + 23 cents).

For CVE, I note that CVE0.10 ( i.e.  0$10 ) is worth USD$0.0009 so i actually wonder how many decimal points are being used these days…

btangmu:
@srl295 "a currency format where something replaces the decimal point" -- I still don't get how that would be related to ZWS; is there existing code that replaces ZWS with something? If so, where, and is it documented?

srl295:
> @srl295 "a currency format where something replaces the decimal point" -- I still don't get how that would be related to ZWS; is there existing code that replaces ZWS with something? If so, where, and is it documented?

I don't know if there's code or not, but I did notice that CVE and PTE in the above referenced files have:  `<symbol>`*ZWSP*`</symbol>`

srl295:
> > I did notice that CVE and PTE in the above referenced files have: `<symbol>`_ZWSP_`</symbol>`
> 
> That is effectively an empty string. My question is whether it would be ok for it to be an actual empty string in this kind of case. I wouldn't push it for CLDR 42/ICU 72, but this seems like something to consider for 43.

I'm guessing it is a standin for an empty string.  CLDR generally is _not_ OK with an empty string, if i recall correctly.

> Actually, it seems like the symbol should be `$` (or whatever is appropriate for the currency), and then there should be something that says to use the currency symbol as the decimal separator, and don't output it again in a prefix or suffix.

it does seem conceivable that the symbol would be different from the decimal marker.  But who knows. 

Actually, there are a few currencies which use this field but don't have an 'empty' currency symbol.

```
currencies.northern_western_europe.html:<tr><th class='path'><a name='5121554065b9d512' href='#5121554065b9d512'>EEK-name-decimal</a></th><th class='path'></th></tr>
currencies.northern_western_europe.html:<tr><th class='path'><a name='5e651499d56495da' href='#5e651499d56495da'>LUF-name-decimal</a></th><th class='path'></th></tr>
currencies.southern_eastern_europe.html:<tr><th class='path'><a name='1e6723482b81864a' href='#1e6723482b81864a'>ESP-name-decimal</a></th><th class='path'></th></tr>
currencies.southern_eastern_europe.html:<tr><th class='path'><a name='6aad5398e5379f95' href='#6aad5398e5379f95'>GRD-name-decimal</a></th><th class='path'></th></tr>
currencies.southern_eastern_europe.html:<tr><th class='path'><a name='482824985aa379ea' href='#482824985aa379ea'>ITL-name-decimal</a></th><th class='path'></th></tr>
currencies.southern_eastern_europe.html:<tr><th class='path'><a name='7e8aeb252b5b9b4e' href='#7e8aeb252b5b9b4e'>PTE-name-decimal</a></th><th class='path'></th></tr>
currencies.western_africa.html:<tr><th class='path'><a name='190108d7b3f47090' href='#190108d7b3f47090'>CVE-name-decimal</a></th><th class='path'></th></tr>
```

btangmu:
I'm still bewildered by the currency usage of ZWS. Anyway, I've added a link to this PR in the new ticket, so we won't lose track of these valuable comments, and I encourage further comments to go in the new ticket: https://unicode-org.atlassian.net/browse/CLDR-15987

macchiati:
Empty strings can be handled in CLDR (with some doing — it is only allowed
in special cases because of (a) tooling and (b) it is almost always wrong).
I'm not sure, however, about the clients of cldr, and this is pretty
fundamental. So we want to be very careful about any changes.

PR #2334, opened 2022-09-07
CLDR-13837 Make the Survey Tool status icons be characters

btangmu:
-Revise seven styles: d-dr-approved, d-dr-missing, etc.

-Delete seven .png files: approved.png, missing.png, etc.

-Revise style statuscol and move it from redesign.css to surveytool.css

-Remove references to undefined style d-dr-status

-Change div to span and remove nbsp in two .vue files

CLDR-13837


srl295 (approved):

srl295:
> This may not complete the ticket since the icons aren't searchable -- seemingly the css "::before" technique isn't appropriate for that

if we want them searchable it will have to be regular text

macchiati (changes requested):
The text needs to be searchable, so wants to be real characters, not just ccs decorations.

btangmu:
See https://github.com/unicode-org/cldr/pull/2336

PR #2343, opened 2022-09-08
CLDR-13837 Survey Tool status icons, fix css error

btangmu:
-Delete style in redesign.css that was .data .statuscol (without comma)

-This should have been deleted when statuscol was moved to surveytool.css

CLDR-13837


pedberg-icu (approved):

btangmu:
The bug is visible in cldr-smoke, the main Page table isn't always full-width.

".data .statuscol" means statuscol as a descendent of data; statuscol doesn't occur elsewhere so the ".data" isn't necessary. I had treated it as though it were ".data,.statuscol" with a comma, and I didn't want to mess with the data part.

PR #2347, opened 2022-09-09
CLDR-16002 json: update 'modern' logic

srl295:
- update logic for 'modern' vs 'full'
- new StandardCodes.getHighestLocaleCoverageLevel() which checks parent locales (but not root)

CLDR-16002

Yes, this is on maint-42

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
@@ -392,6 +392,25 @@ public class StandardCodes {
         return getLocaleTypes().get(org);
     }
 
+    /**
+     * returns the highest level in the hierarchy, not including root.
+     */
+    public Level getHighestLocaleCoverageLevel(String organization, String locale) {
+        // first get parent
+        final String parentId = LocaleIDParser.getParent(locale);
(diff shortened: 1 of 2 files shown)


srl295:
@sffc  the symptom was that only four en sublocales were included in 'modern', this restores all 106.

macchiati (changes requested):
This will want to be recursive.

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -392,6 +392,25 @@ public Map<String, Level> getLocaleToLevel(Organization org) {
...
 
+    /**
+     * returns the highest level in the hierarchy, not including root.
+     */
+    public Level getHighestLocaleCoverageLevel(String organization, String locale) {
+        // first get parent
+        final String parentId = LocaleIDParser.getParent(locale);
+        Level parentLevel = Level.UNDETERMINED;
+        if (parentId != null && !parentId.equals("root")) {
+            parentLevel = getHighestLocaleCoverageLevel(organization, parentId); // recurse
```
srl295, line 403: @macchiati Recurses here

macchiati (approved):

PR #2359, opened 2022-09-16
CLDR-15034 kbd: keyboard test DTD

srl295:
CLDR-15034
 
- DTD, docs and sample file for keyboard regression testing
 
Also, CLDR fixes:
- fix unclear message in ElementAttributeInfo and TestDtdData

ALLOW_MULTI_COMMITS=true
ALLOW_MANY_COMMITS=true


docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
...
+> Occurrence: Optional, Multiple
+> </small>
+
+This element represents a text output event.
+
+_Attribute:_ `to` (required)
+
+This attribute specifies a a string of output text, which is intended to match a key’s `to` attribute.
+
+Tooling should give a warning if this attribute matches more than one key, or does not match any keys.
```
mcdurdin, line 2298: Be good to note that this is helpful for testing transforms without needing reference to the key definitions.

Don't think we should be generating a warning if more than one key generates the same `to`. This will be common, for example, with punctuation on touch layouts. A warning if the `to` cannot be generated by any key/flick/longPress/multiTap is appropriate.

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
...
+
+> <small>
+>
+> Parents: [check](#test-element-check)
+>
+> Children: _none_
+>
+> Occurrence: Optional, Multiple
+> </small>
+
```
mcdurdin, line 2316: 👏🏼

docs/ldml/tr35-keyboards.md
```diff
@@ -101,6 +101,18 @@ The LDML specification is divided into the following parts:
...
 * [CLDR VKey Enum](#CLDR_VKey_Enum)
+* [Keyboard Test Data](#keyboard-test-data)
+  * [Test Doctype](#test-doctype)
+  * [Test Element: keyboardTest](#test-element-keyboardtest)
+  * [Test Element: info](#test-element-info)
+  * [Test Element: repertoire](#test-element-repertoire)
+  * [Test Element: tests](#test-element-tests)
+  * [Test Element: test](#test-element-test)
+  * [Test Element: startContext](#test-element-startcontext)
+  * [Test Element: event](#test-element-event)
```
mcdurdin, line 112:
```suggestion
  * [Test Element: keystroke](#test-element-keystroke)
```

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
...
+Specifies the starting context. This text may be escaped with `\u` notation, see [Escaping](#Escaping).
+
+**Example**
+
+```xml
+<startContext to="abc\u0022"/>
+```
+
+
+### Test Element: event
```
mcdurdin, line 2236:
```suggestion
### Test Element: keystroke
```

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
 
 * * *
 
+## Keyboard Test Data
+
+Keyboard Test Data allows the keyboard author to provide regression test data to validate the repertoire and behavior of a keyboard. Tooling can run these regression tests against an implementation, and can also be used as part of the development cycle to validate that keyboard changes do not deviate from expected behavior.
+
+Test data files have a separate DTD, named `ldmlKeyboardTest.dtd`.  Sample test data files are located in the `keyboards/test` subdirectory.
```
mcdurdin, line 2050: We should add some wording about best practice of grouping and naming of tests, multiple files per keyboard possible, multiple tests per file.

Think this may help head off a future quagmire of randomly organised keyboard tests...

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
...
+> <small>
+>
+> Parents: [keyboardTest](#test-element-keyboardtest)
+>
+> Children: _none_
+>
+> Occurrence: Optional, Multiple
+> </small>
+
+This element represents a repertoire test, to validate the available characters and their reachability. This test ensures that each of the specified characters is somehow typeable on the keyboard.
```
mcdurdin, line 2122:
```suggestion
This element represents a repertoire test, to validate the available characters and their reachability. This test ensures that each of the specified characters is somehow typeable on the keyboard, after transforms have been applied. The characters in the repertoire will be matched against the complete set of possible generated outputs, post-transform, of all keys on the keyboard.
```

I know I wrote this but it strikes me that this may be computationally complex to verify -- particularly where we are limiting the output to 'simple' keys:

> The characters in the repertoire will be matched against the complete set of possible generated outputs, post-transform, of all keys on the keyboard.

It's not going to be sufficient to just look at raw transform outputs. A transform may have outputs that can never be reached because we cannot satisfy the inputs. But how would we test this efficiently? Thoughts?

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
...
+
+_Attribute:_ `name`
+
+This attribute specifies a unique name for this repertoire test. These names could be output to the user during test execution, used to summarize success and failure, or used to select or deselect test components.
+
+_Attribute:_ `type`
+
+This attribute is either `simple` (the default) or `gesture`. A value of `simple` indicates that each of the characters in `chars` are typeable by simple single keystrokes without needing any gestures such as flicks, long presses, or multiple taps. A value of `gesture` indicates that the characters are typeable by either simple keystrokes or by use of gestures such as flicks, long presses, or multiple taps.
```
mcdurdin, line 2130: Let's further split this out by gesture type -- 'any' covers any method of generating output; 'gesture' for any gesture; 'flick', 'longPress', 'multiTap', and 'simple' are the specific methods. This will be no harder for us to code and gives useful additional flexibility to the tests.

mcdurdin (changes requested):
Looking good. Like `check` for splitting action from result.

docs/ldml/tr35-keyboards.md
srl295: To be clear the warning isn't that keys match more than one to, but that such a key is used in the test. Better to use keystroke to be un ambiguous.

docs/ldml/tr35-keyboards.md
srl295: That's automatic. Maybe out of date.

docs/ldml/tr35-keyboards.md
mcdurdin: I don't think so: when we want to write tests for transforms, we probably don't really care which key is generating the character(s); this would only be a concern if no key could generate the character.

Realised just now that we need to clarify that `<emit>` will inject `to` into the transform+reorder engine for the keyboard -- that it isn't emitted directly into the text store.

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,316 @@ Footnotes:
...
+> Children: _none_
+>
+> Occurrence: Optional, Multiple
+> </small>
+
+This element represents a text output event.
+
+_Attribute:_ `to` (required)
+
+This attribute specifies a a string of output text, which is intended to match a key’s `to` attribute.
```
mcdurdin, line 2296:
```suggestion
This attribute specifies a string of output text, which is intended to match a key’s `to` attribute.
```

docs/ldml/tr35-keyboards.md
srl295: Such a transform situation should be flagged in lint.  Yes, it will be complex.

docs/ldml/tr35-keyboards.md
srl295: Good point. Since flicks and such are out of the picture, it doesn't matter.

docs/ldml/tr35-keyboards.md
srl295: Not sure what the best practices should be at the moment.  This seems somewhat like a repository policy discussion, which we need to have more discussions around.  Right now I have a `3.0/` directory with keyboards, and a `test/` directory.

docs/ldml/tr35-keyboards.md
srl295: added hardware for good measure

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+| simple    | Each of the characters must be typeable by simple single keystrokes without needing any gestures.        |
+| gesture   | The characters are typeable by use of any gestures such as flicks, long presses, or multiple taps.       |
+| flick     | The characters are typeable by use of any `flick` element.                                               |
+| longPress | The characters are typeable by use of any `longPress` value.                                             |
+| multiTap  | The characters are typeable by use of any `multiTap` value.                                              |
+| hardware  | The characters are typeable by use of any simple keystrokes on any hardware layout.                      |
+
+either `simple` (the default) or `gesture`. A value of `simple` indicates that each of the characters in `chars` are typeable by simple single keystrokes without needing any gestures such as flicks, long presses, or multiple taps. A value of `gesture` indicates that the characters are typeable by either simple keystrokes or by use of gestures such as flicks, long presses, or multiple taps.
```
mcdurdin, line 2144: This doesn't match up with the table above now -- probably delete.

srl295:
filed CLDR-16040 to track complaints about `<backspace/>` element

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+> Children: _none_
+>
+> Occurrence: Optional, Multiple
+> </small>
+
+This element represents a single keystroke or other gesture event, identified by a particular key element.
+
+Optionally, one of the gesture attributes, either `flick`, `longPress`, or `tapCount` may be specified. If none of the gesture attributes are specified, then a regular keypress is effected on the key.  It is an error to specify more than one gesture attribute.
+
+If a key is not found, or a particular gesture has no definition, the output should be behave as if the user attempted to perform such an action.  For example, an unspecified `flick` would result in no output.
```
mcdurdin, line 2265: Should the regression test tool emit a hint on missing keys? I think that would be exceptionally helpful for writing regression tests. (May not require an update to the text here though?)

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+
+_Attribute:_ `longPress`
+
+This attribute specifies that a long press gesture should be performed on the specified key instead of a keypress. For example, `longPress="2"` indicates that the second character in a longpress series should be chosen. `longPress="0"` indicates that the `longPressDefault` value, if any, should be chosen. This corresponds to `longPress` and `longPressDefault` on [`<key>`](#Element_key).
+
+_Attribute:_ `tapCount`
+
+This attribute specifies that a multi-tap gesture should be performed on the specified key instead of a keypress. For example, `tapCount="3"` indicates that the key should be tapped three times in rapid succession. This corresponds to `multiTap` on [`<key>`](#Element_key).
+
+Note that `tapCount="1"` is valid, but represents an ordinary keypress.
```
mcdurdin, line 2285: I think we should disallow `tapCount="1"` because it introduces an arbitrary distinction between hardware keypresses and touch layout keypresses -- it would be valid for a hardware keyboard but as gestures are not allowed, `tapCount` would not otherwise be permitted.

Note that while `tapCount` could be technically allowable on a hardware keyboard, we have not up until this point considered doing this!

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+> <small>
+>
+> Parents: [test](#test-element-emit)
+>
+> Children: _none_
+>
+> Occurrence: Optional, Multiple
+> </small>
+
+This element also represents a keystroke event, except that the keystroke is specified in terms of textual value rather than key or gesture identity. This element is particularly useful for testing transforms.
```
mcdurdin, line 2308:
```suggestion
This element also represents an input event, except that the input is specified in terms of textual value rather than key or gesture identity. This element is particularly useful for testing transforms.
```

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+
+Processing of the specified text continues with the transform and other elements before updating the test output buffer.
+
+_Attribute:_ `to` (required)
+
+This attribute specifies a string of output text representing a single keystroke or gesture. This string is intended to match the output of a `key`, `flick`, `longPress` or `multiTap` element or attribute.
+Tooling should give a warning if this attribute does not match at least one keystroke or gesture. Note that the specified text is not injected directly into the output buffer.
```
mcdurdin, line 2315:
```suggestion
Tooling should give a hint if this attribute does not match at least one keystroke or gesture. Note that the specified text is not injected directly into the output buffer.
```

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+> Children: _none_
+>
+> Occurrence: Optional, Multiple
+> </small>
+
+This element represents a check on the current output buffer.
+
+_Attribute:_ `result` (required)
+
+This attribute specifies the expected resultant text in a document after processing this event and all prior events, and including any `preContext` text.  This text may be escaped with `\u` notation, see [Escaping](#Escaping).
```
mcdurdin, line 2358:
```suggestion
This attribute specifies the expected resultant text in a document after processing this event and all prior events, and including any `startContext` text.  This text may be escaped with `\u` notation, see [Escaping](#Escaping).
```

docs/ldml/tr35-keyboards.md
```diff
@@ -2030,6 +2043,353 @@ Footnotes:
...
+
+This attribute specifies a unique name for this repertoire test. These names could be output to the user during test execution, used to summarize success and failure, or used to select or deselect test components.
+
+_Attribute:_ `type`
+
+This attribute is one of the following:
+
+|  type     | Meaning                                                                                                  |
+|-----------|----------------------------------------------------------------------------------------------------------|
+| default   | This is the default, indicates that _any_ flick or keystroke may be used to generate each character      |
```
mcdurdin, line 2136:
```suggestion
| default   | This is the default, indicates that _any_ gesture or keystroke may be used to generate each character    |
```

keyboards/dtd/ldmlKeyboardTest.dtd
```diff
@@ -0,0 +1,78 @@
...
+<!ATTLIST info author CDATA #IMPLIED >
+    <!--@METADATA-->
+<!ATTLIST info name NMTOKEN #IMPLIED >
+
+<!ELEMENT repertoire EMPTY >
+    <!--@ORDERED-->
+<!ATTLIST repertoire chars CDATA #REQUIRED >
+    <!--@MATCH:any-->
+    <!--@VALUE-->
+<!ATTLIST repertoire type (default | simple | gesture | flick | longPress | multiTap | hardware) #REQUIRED >
```
mcdurdin, line 33: Should this be `#IMPLIED` for `default`?

keyboards/dtd/ldmlKeyboardTest.dtd
```diff
@@ -0,0 +1,78 @@
...
+    <!--@VALUE-->
+<!ATTLIST repertoire name NMTOKEN #IMPLIED >
+
+<!ELEMENT tests ( test+, special* ) >
+<!ATTLIST tests name NMTOKEN #IMPLIED >
+
+<!ELEMENT test ( startContext?, ( emit | keystroke | check | backspace )*, special* ) >
```
mcdurdin, line 40:
```suggestion
<!ELEMENT test ( startContext?, ( keystroke | emit | backspace | check )*, special* ) >
```

I like it better in that order. Or could go with alpha order.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
```diff
@@ -349,6 +353,13 @@ public void TestNewDtdData() {
 //
 //    }
 
+    /**
+     * paths that can be empty erlements. Each item starts with '!' because of showPath.
```
mcdurdin, line 357:
```suggestion
     * paths that can be empty elements. Each item starts with '!' because of showPath.
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
```diff
@@ -432,8 +443,17 @@ public void TestNewDtdData() {
             "row" // keyboard
             )));
 
+    static final Set<String> orderedKeyboardTestElements = Collections.unmodifiableSet(new HashSet<>(Arrays
+            .asList(    "repertoire", "event", "emit", "keystroke", "check", "backspace" )));
```
mcdurdin, line 447: Why is `repertoire` in this list? And I think `event` should be deleted?

srl295: fixed

srl295 (commented):
Note: 'name' is now required everywhere. That keeps the elements distinguishable.

mcdurdin (approved):

PR #2361, opened 2022-09-19
CLDR-15933 Remove unnecessary & incorrect legacy alias added for key "mu"

pedberg-icu:
CLDR-15933

The bcp47 alias added in #2303 for new key "mu" is unnecessary (don't need aliases for new keys), and is malformed.


macchiati (approved):

markusicu:
+1 tnx!

PR #2366, opened 2022-09-21
CLDR-15626 Document currencyPattern alts alphaNextToNumber,noCurrency and currencyPatternAppendISO

pedberg-icu:
CLDR-15626

### DIFF
--- a/docs/ldml/tr35-numbers.md
+++ b/docs/ldml/tr35-numbers.md
@@ -405,19 +405,27 @@ The short format is designed for UI environments where space is at a premium, an
 
 #### 2.4.2 <a name="Currency_Formats" href="#Currency_Formats">Currency Formats</a>
 
-Pattern for use with currency formatting. This format contains a few additional structural options that allow proper placement of the currency symbol relative to the numeric quantity. Refer to _[Section 4 - Currencies](#Currencies)_ for additional information on the use of these options.
+Patterns for use with currency formatting:
 
 ```xml
-<!ELEMENT currencyFormats (alias | (default*, currencySpacing*, currencyFormatLength*, unitPattern*, special*)) >
+<!ELEMENT currencyFormats (alias | (default*, currencySpacing*, currencyFormatLength*, currencyPatternAppendISO*, unitPattern*, special*)) >
+<!ELEMENT currencyFormatLength (alias | (default*, currencyFormat*, special*)) >
+<!ATTLIST currencyFormatLength type ( full | long | medium | short ) #IMPLIED >
+<!ELEMENT currencyFormat (alias | (pattern*, special*)) >
+<!ATTLIST currencyFormat type NMTOKEN "standard" >
+    <!--@MATCH:literal/accounting, standard-->
+<!ELEMENT currencyPatternAppendISO ( #PCDATA ) >
+```
+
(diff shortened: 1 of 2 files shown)


srl295 (approved):
thanks!

docs/ldml/tr35-numbers.md
```diff
@@ -427,14 +435,39 @@ In addition to a standard currency format, in which negative currency amounts mi
...
+            <pattern type="100000000000000" count="other" alt="alphaNextToNumber">¤ 000T</pattern>
         </currencyFormat>
     </currencyFormatLength>
 </currencyFormats>
 ```
 
+The `alt="alphaNextToNumber"` pattern, if available, should be used instead of the standard pattern when the currency symbol character closest to the numeric value has Unicode General Category L (letter). The `alt="alphaNextToNumber"` pattern is typically provided when the standard currency pattern does not have a space between currency symbol and numeric value; the alphaNextToNumber variant adds a non-breaking space if appropriate for the locale.
+
+The `alt="noCurrency"` pattern can be used when a currency-style format is desired but without the currency symbol. This sort of display may be used when formatting a large column of values all in the same currency, for example. For compact currency formats (`<currencyFormatLength type="short">`), the cpmpact decimal format (`<decimalFormatLength type="short">`) should be used if no `alt="noCurrency"` pattern is present (so the `alt="noCurrency"` pattern is typically not needed for compact currency formats).
```
sffc, line 463:
```suggestion
The `alt="noCurrency"` pattern can be used when a currency-style format is desired but without the currency symbol. This sort of display may be used when formatting a large column of values all in the same currency, for example. For compact currency formats (`<currencyFormatLength type="short">`), the compact decimal format (`<decimalFormatLength type="short">`) should be used if no `alt="noCurrency"` pattern is present (so the `alt="noCurrency"` pattern is typically not needed for compact currency formats).
```

docs/ldml/tr35-numbers.md
```diff
@@ -427,14 +435,39 @@ In addition to a standard currency format, in which negative currency amounts mi
...
+
+```xml
+<currencyPatternAppendISO>{0} ¤¤</currencyPatternAppendISO>
+```
+
+The `currencyPatternAppendISO` element provides a pattern that can be used to combine currency format that uses a currency symbol (¤) with the ISO 4217 3-letter code for the same currency (¤¤), to produce a result such as “$1,432.00 USD”. Most locales will not need to override the pattern provided in root, shown above.
```
sffc, line 469: How do we specify when implementations should use this pattern? Making the pattern available is a step in the right direction, but it's not very useful if we don't specify how to use it. The main example is when the currency symbol contains the ISO code, like "CHF". Should implementations do an exact string match? But then how about cases without ambiguity, like "CA$"; seems CAD is not really necessary for that. Maybe they should be paired with the narrow symbol, but only if the symbol is not equal to the ISO code?

docs/ldml/tr35-numbers.md
pedberg-icu: Fixed

docs/ldml/tr35-numbers.md
pedberg-icu: @sffc OK, I updated with a suggested recommendation for when to use.

pedberg-icu:
Note in the recent update I also added a Mods item for https://github.com/unicode-org/cldr/pull/2367 since while this PR was outstanding I could not add it there due to collision.

sffc (commented):
LGTM

PR #2367, opened 2022-09-22
CLDR-15286 Document new dateTimeFormat "atTime" variant and when to use it

pedberg-icu:
CLDR-15286

Document new dateTimeFormat "atTime" variant and when to use it.

Note in this PR I am not adding the corresponding entry in the Mods section of tr35 because it would conflict with a prior PR. I will updated the Mods section for this addition as part of another spec PR.


srl295 (approved):
makes sense now!

sffc (approved):

PR #2372, opened 2022-09-23
CLDR-16018 fix table

macchiati:
CLDR-16018

Used https://www.tablesgenerator.com/markdown_tables to regenerate and fix the table

1. first paste from HTML spec page into spreadsheet (https://unicode.org/reports/tr35/tr35.html#1-multimap-interpretation)
2. then adjust spreadsheet to have no merged cells (and in this case, fix the misalignment)
3. copy into  https://www.tablesgenerator.com/markdown_tables
4. hit Generate, and copy result into LDML page

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -3394,15 +3394,11 @@ Interpret each languageId as a multimap from a _fieldId_ (language, script, regi
 
 _Examples:_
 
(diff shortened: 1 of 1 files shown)


AEApple (approved):
Should we cherrypick this fix back to the last region too if we're doing revisions to update the links?

@srl295

PR #2378, opened 2022-09-24
CLDR-15955 Add constraints for unit IDs

macchiati:
CLDR-15955

Note that this deviates somewhat from the ticket: simple_unit is now above prefixed_unit, since the names and structure make more sense that way.

(Also, the table in the ticket got gummed up because of the presence of | characters in the text. The sheet https://docs.google.com/spreadsheets/d/1QpVPnIgJdYWZil2sr_L4P3z6O_AsT3sKZhP_SOyTVqQ/edit#gid=1228405746 has a cleaned up version.)

I considered changing the table structure to markdown, but refrained because that would make it harder to review. 

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -749,6 +749,8 @@ The `measurement` element is deprecated in the main LDML files, because the data
 ```xml
 <!ELEMENT units (alias | (unit*, unitLength*, durationUnit*, special*) ) >
 
+<!ELEMENT unitIdComponents ( unitIdComponent* ) >
+
 <!ELEMENT unitLength (alias | (compoundUnit*, unit*, coordinateUnit*, special*) ) >
 <!ATTLIST unitLength type (long | short | narrow) #REQUIRED >
 
@@ -895,7 +897,8 @@ As with other identifiers in CLDR, the American English spelling is used for uni
 
 #### Syntax
 
-The formal syntax for identifiers is provided below.
+The formal syntax for identifiers is provided below. 
+Some of the constraints reference data from the unitIdComponents in [Unit_Conversion](tr35-info.md#Unit_Conversion).
 
 <!-- HTML: no header -->
 
@@ -913,6 +916,7 @@ The formal syntax for identifiers is provided below.
                 <li>per-second</li>
             </ul></li>
             <li><em>Note:</em> The normalized form will have only one "per"</li>
+	  <li><em>Note:</em>The token 'per' is the single value in &lt;unitIdComponent type=”per”&gt;</li>
         </ul></td></tr>
 
 <tr><td>product_unit</td><td>:=</td>
@@ -923,7 +927,7 @@ The formal syntax for identifiers is provided below.
             </ul></td></tr>
 
 <tr><td>single_unit</td><td>:=</td>
-    <td>number_prefix? dimensionality_prefix? prefixed_unit
+    <td>number_prefix? dimensionality_prefix? simple_unit
         <ul><li><em>Examples: </em>square-meter, or 100-square-meter</li></ul></td></tr>
 
 <tr><td>pu_single_unit</td><td>:=</td>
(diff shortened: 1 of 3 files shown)


docs/ldml/tr35-general.md
```diff
@@ -960,25 +976,33 @@ The formal syntax for identifiers is provided below.
...
-        </ul></td></tr>
-
-<tr><td>unit_component</td><td>:=</td>
+<tr><td>prefix_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must be value in: &lt;unitComponent type=”prefix_component”&gt;.</li></ul></td></tr>
+
+<tr><td>base_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must not be a value in: &lt;unitComponent type=”suffix_component”&gt;
```
sffc, line 985:
```suggestion
        <ul><li><em>Constraint:</em> must not be a value in: &lt;unitComponent type=”prefix_component”&gt; 
```

docs/ldml/tr35-general.md
```diff
@@ -960,25 +976,33 @@ The formal syntax for identifiers is provided below.
...
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must be value in: &lt;unitComponent type=”prefix_component”&gt;.</li></ul></td></tr>
+
+<tr><td>base_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must not be a value in: &lt;unitComponent type=”suffix_component”&gt; 
+			or &lt;unitComponent type=”suffix_component”&gt; 
+			or &lt;unitComponent type=”power”&gt;
+			or &lt;unitComponent type=”and”&gt;
+			or &lt;unitComponent type=”per”&gt;.
```
sffc, line 989: "and" and "per" are literals, not sets

docs/ldml/tr35-general.md
```diff
@@ -960,25 +976,33 @@ The formal syntax for identifiers is provided below.
...
-<tr><td>unit_component</td><td>:=</td>
+<tr><td>prefix_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must be value in: &lt;unitComponent type=”prefix_component”&gt;.</li></ul></td></tr>
+
+<tr><td>base_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must not be a value in: &lt;unitComponent type=”suffix_component”&gt; 
+			or &lt;unitComponent type=”suffix_component”&gt; 
+			or &lt;unitComponent type=”power”&gt;
```
sffc, line 987: The grammar construction is named "dimensionality_prefix"

docs/ldml/tr35-general.md
```diff
@@ -960,25 +976,33 @@ The formal syntax for identifiers is provided below.
...
-
-<tr><td>unit_component</td><td>:=</td>
+<tr><td>prefix_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must be value in: &lt;unitComponent type=”prefix_component”&gt;.</li></ul></td></tr>
+
+<tr><td>base_component</td><td>:=</td>
+    <td>[a-z]{3,∞}
+        <ul><li><em>Constraint:</em> must not be a value in: &lt;unitComponent type=”suffix_component”&gt; 
+			or &lt;unitComponent type=”suffix_component”&gt;
```
sffc, line 986: Question: what is a unitComponent? I don't see that term defined anywhere.

I would have thought to instead write this as, "Constraint: does not match any of the following: ..."

docs/ldml/tr35-general.md
macchiati: They are members of singleton sets in the unitComponents

docs/ldml/tr35-general.md
macchiati: Hmmm. It would have been better to name it that in the XML. We'd have to check and see if that's a change we can make. It might be ok, since I don't think ICU uses those.

docs/ldml/tr35-general.md
macchiati: It is a new structure in the XML file. Which is a good point; it needs to be documented. Will have subsequent PR for that.

docs/ldml/tr35-general.md
macchiati: No, it means that it can't be in any of the unitComponent values, so it is listing each of them. I could make it "... not be in X nor in Y nor in .." if nor is clearer than or is clearer.

docs/ldml/tr35-general.md
macchiati: > I would have thought to instead write this as, "Constraint: does not match any of the following: ..."

That's better wording.

docs/ldml/tr35-general.md
```diff
@@ -945,10 +949,26 @@ The formal syntax for identifiers is provided below.
...
     <td>"square-"<p>| "cubic-"<p>| "pow" ([2-9]|1[0-5]) "-"
-        <ul><li><em>Note:</em> "pow2-" and "pow3-" canonicalize to "square-" and "cubic-"</li></ul></td></tr>
+        <ul>
+			<li><em>Note:</em> "pow2-" and "pow3-" canonicalize to "square-" and "cubic-"</li>
+			<li><em>Note:</em>These are values in &lt;unitIdComponent type=”power”&gt;</li>
+		</ul></td></tr>
 
+<tr><td>simple_unit</td><td>:=</td>
+    <td>prefixed_unit<br/>
+		| (prefix_component "-")* base_component ("-" suffix_component)*<br/>
```
sffc, line 959: I think this should reference prefixed_unit instead of base_component

docs/ldml/tr35-general.md
sffc: The bug I was pointing out was that you have suffix_component listed twice

macchiati: Ah!

docs/ldml/tr35-general.md
macchiati: It is more restrictive than what we had before. 
That is taking advantage of the fact that the metric units (roughly, units that can have SI or binary prefixes attached to them) don't have prefix_components or suffix_components.
It would mean that while before you could say "3 megaquarts-imperial" and with this change you can't.
However, thinking about it now, a full solution would require more changes, so I'll fix this.

docs/ldml/tr35-general.md
macchiati: done

docs/ldml/tr35-general.md
macchiati: I filed https://unicode-org.atlassian.net/browse/CLDR-16038, and you can leave comments there.

docs/ldml/tr35-general.md
```diff
@@ -945,10 +949,26 @@ The formal syntax for identifiers is provided below.
...
+		|  currency_unit<br/>
+		| “em” | “g” | “us” | “hg” | "of"
+        <ul><li><em>Examples:</em>kilometer, meter, cup-metric, fluid-ounce, curr-chf, em</li></ul></li>
+            <li><em>Constraint:</em> At least one unit_component must not itself be a simple_unit</li>
```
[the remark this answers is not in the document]
macchiati: done

docs/ldml/tr35-general.md
macchiati: done

macchiati:
And thanks for the detailed review!

PR #2381, opened 2022-09-25
CLDR-15608 Specify ranges of numbers with qualifiers

macchiati:
CLDR-15608

### DIFF
--- a/docs/ldml/tr35-numbers.md
+++ b/docs/ldml/tr35-numbers.md
@@ -1429,16 +1429,18 @@ The currently recommended heuristic is:
 1. Never collapse scientific or compact notation. This is to avoid producing ambiguous strings such as "3–5M" (could represent 3–5,000,000 or 3,000,000–5,000,000).
 2. Only collapse if the tokens are more than one code point in length. This is to increase clarity of strings such as "$3–$5".
 3. To perform the collapse, remove the token that is closest to the range separator.
-That is, for a prefix element, remove from the end value, and for a suffix element remove it from the start value:
+That is, for a prefix element, remove from the end value of the range, and for a suffix element remove it from the start value of the range:
     * USD 2 – USD 5 ⇒ USD 2 – 5
     * 2M EUR – 5M EUR ⇒  2M – 5M EUR
     * 2 km – 5 km ⇒ 2 – 5 km
     * 2M ft – 5M ft ⇒ 2M – 5M ft
-    * In bidi contexts, the data is built so that this works **visually**. 
-        For example, if a range from 2 km to 5 km would be presented visually as "_mk 5 – mk 2_", the collapsed form would be "_mk 5 – 2_". 
-        (The _mk_ is a stand-in for the native representation.) This requires consistent ordering among the elements. 
-        Thus a prefix value will be reordered to be visually a suffix value, and the order of the range will be reversed.
-4. When the tokens can have distinct plural forms, modify the remaining token so that it has the correct plural form, . That is, use [Plural Ranges](#Plural_Ranges) to calculate the correct plural category for the range, and pick the variant of that the remaining token corresponding to that plural form.
+4. When the tokens can have distinct plural forms, modify the remaining token so that it has the correct plural form. That is, use [Plural Ranges](#Plural_Ranges) to calculate the correct plural category for the range, and pick the variant of that the remaining token corresponding to that plural form.
+
+In bidi contexts, the data is built so that rule #3 works **visually**. 
+For example, if a range from 2 km to 5 km would be presented visually as "_mk 5 – mk 2_", the collapsed form would be "_mk 5 – 2_". 
+(The _mk_ is a stand-in for the native representation.) 
+This requires consistent visualy reordering among the elements: the range, the prefixes and the suffixes.
+Thus a prefix value will be reordered to be visually a suffix value, and the order of the range will be visually reversed.
 
 ### 8.3 <a name="Range_Pattern_Processing" href="#Range_Pattern_Processing">Range Pattern Processing</a>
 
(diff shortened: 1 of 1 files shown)


macchiati:
Hi, I'd like to get this merged by tomorrow morning. I think I've addressed all of Shane's comments.

macchiati:
Shane, I think the original ticket was inspired by you; could you take a look at it?

docs/ldml/tr35-numbers.md
```diff
@@ -1423,12 +1423,22 @@ The above description describes the expected output. Internally, the implementat
...
-These heuristics may be refined in the future.
+To perform the collapse:
+
+1. Remove the token that is closest to the range separator.
+That is, for a prefix element, remove from the end value, and for a suffix element remove it from the start value:
+    * USD 2 – USD 5 ⇒ USD 2 – 5
+    * 2M EUR – 5M EUR ⇒  2 M – 5M EUR
+    * 2 km – 5 km ⇒ 2 – 5 km
+    * 2M ft – 5M ft ⇒ 2M – 5M ft
+2. In bidi contexts, rule #1 is applied **visually**. For example, if a range from 2 km to 5 km would be presented visually as "_mk 5 – mk 2_", the collapsed form would be "_mk 5 – 2_". (The _mk_ is a stand-in for the native representation.)
```
sffc, line 1437: Question about the new bidi clause. The example you gave should be fine for both visual and logical order, right? In the ICU implementation we definitely only do this in logical order (because visual order would require pulling in the bidi code, which we don't do).

docs/ldml/tr35-numbers.md
```diff
@@ -1423,12 +1423,22 @@ The above description describes the expected output. Internally, the implementat
...
+1. Never collapse scientific or compact notation. This is to avoid producing ambiguous strings such as "3–5M" (could represent 3–5,000,000 or 3,000,000–5,000,000).
 1. Only collapse if the tokens are more than one code point in length. This is to increase clarity of strings such as "$3–$5".
 
-These heuristics may be refined in the future.
+To perform the collapse:
+
+1. Remove the token that is closest to the range separator.
+That is, for a prefix element, remove from the end value, and for a suffix element remove it from the start value:
+    * USD 2 – USD 5 ⇒ USD 2 – 5
+    * 2M EUR – 5M EUR ⇒  2 M – 5M EUR
```
sffc, line 1434:
```suggestion
    * 2M EUR – 5M EUR ⇒  2M – 5M EUR
```

docs/ldml/tr35-numbers.md
```diff
@@ -1423,12 +1423,22 @@ The above description describes the expected output. Internally, the implementat
...
+    * 2 km – 5 km ⇒ 2 – 5 km
+    * 2M ft – 5M ft ⇒ 2M – 5M ft
+2. In bidi contexts, rule #1 is applied **visually**. For example, if a range from 2 km to 5 km would be presented visually as "_mk 5 – mk 2_", the collapsed form would be "_mk 5 – 2_". (The _mk_ is a stand-in for the native representation.)
+
+The removal of  a token requires modification of the remaining token for the correct plural form, if the token can have distinct plural forms. That is, use [Plural Ranges](#Plural_Ranges) to calculate the correct form for the remaining token, and pick the variant of that token corresponding to that plural form.
```
sffc, line 1439:
```suggestion
3. If necessary, modify the remaining token with the correct plural form, if the token can have distinct plural forms. That is, use [Plural Ranges](#Plural_Ranges) to calculate the correct form for the remaining token, and pick the variant of that token corresponding to that plural form.
```

macchiati:
Yes. A prefix currency will appear on the right in English and on the left
in Arabic, etc. Sometimes that means that the patterns need to have RLM or
LRM marks in appropriate places, but that is up to CLDR, not the ICU code.

docs/ldml/tr35-numbers.md
srl295, line 1437:
```suggestion
2. In bidi contexts, rule #1 is applied logically. For example, if a range from 2 km to 5 km would be presented visually as "_mk 5 – mk 2_",  `(٢كم-٥كم)` the collapsed form would be "_mk 5 – 2_"  `(٢-٥كم)`. (The _mk_ is a stand-in for the native representation.)
```
my suggestion rendered below:

image

macchiati:
It shouldn't require any application of the bidi algorithm by the
implementer (as I said); it just requires that the source CLDR data be
properly constructed. It sounds like this is more confusing than helpful;
perhaps I should remove it.

srl295:
> It shouldn't require any application of the bidi algorithm by the implementer (as I said); it just requires that the source CLDR data be properly constructed. It sounds like this is more confusing than helpful; perhaps I should remove it.
> […](#)

“applied visually” is definitely confusing, and I think incorrect. I think it's worth noting that “closest” is still applied logically, so that “closest” is not interpreted as a visual statement.

I think what was meant is that the CLDR data ought to be properly constructed so that the visual order is correct, when this rule is applied.

macchiati:
I think I have it now. Please take a look.

docs/ldml/tr35-numbers.md
macchiati: Revised, with the bidi note restructured.

docs/ldml/tr35-numbers.md
macchiati: done

docs/ldml/tr35-numbers.md
macchiati: Reworded as suggested.

srl295 (approved):
LGTM, no Arabic this time?

macchiati:
No Arabic this time; we can consider later.

PR #2386, opened 2022-09-26
CLDR-16037 DON'T MERGE

macchiati:
CLDR-16037

### DIFF
--- a/docs/ldml/tr35-collation.md
+++ b/docs/ldml/tr35-collation.md
@@ -4,7 +4,7 @@
 
 <!-- HTML: no header -->
 <table><tbody>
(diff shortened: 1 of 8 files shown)


macchiati:
@pedberg-icu @srl295 This is a trial of showing rich diff. Some files are too big or too complicated (I guess).

srl295:
I did the same in #2385 ! 
anyway, yeah, i think the rich diff is failing here. 

So it might be better to do our own markdown diff. Something like, "diff vs prior maint release" could be generated automatically.

srl295:
@macchiati  close this?

PR #2390, opened 2022-09-26
CLDR-13370 Fix hybrid locales tags

AEApple:
CLDR-13370

Fix hybrid locales tags and clarify descriptions.

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -1350,38 +1350,39 @@ Hybrid locales have intermixed content from 2 (or more) languages, often with on
 
 While text in a document can be tagged as partly in one language and partly in another, that is not the same having a hybrid locale. There is a difference between having a Spanglish document, and a Spanish document that has some passages quoted in English. Fine-grained tagging doesn't handle grammatical combinations like Tanglish “Enna matteru?” (_What’s the matter?_), which is neither standard Tamil nor standard English. More importantly, it doesn’t work for the very common use case for a [unicode_locale_id](#unicode_locale_id): _locale selection_.
 
-To communicate requests for localized content and internationalization services, locales are used. When people pick a language from a menu, internally they are picking a locale (en-GB, es-419, etc.). To allow an application to support Spanglish or Hinglish locale selection, [unicode_locale_id](#unicode_locale_id)s can represent hybrid locales using the T extension key-value 'h0-hybrid'. (For more information on the T extension, see _Section 3.7 [Unicode BCP 47 T Extension](#t_Extension)._)
+To communicate requests for localized content and internationalization services, locales are used. When people pick a language from a menu, internally they are picking a locale (en-GB, es-419, etc.). To allow an application to support Spanglish or Hinglish locale selection, [unicode_locale_id](#unicode_locale_id)s can represent hybrid locales using the T Extension key-value 'h0-hybrid'. (For more information on the T extension, see _Section 3.7 [Unicode BCP 47 T Extension](#t_Extension)._)
 
 _However, if users typically expect a their language in non-default script to contain a significant amount of text due to lexical borrowing, then the -t- and hybrid subtags may be omitted. An example of this is when Hindi is written in Latin script since Romanized Hindi typically contains a significant amount of English text, ‘hi-Latn’ can be used instead of ‘hi-Latn-t-en-h0-hybrid’._
 This tends to work better in implementations that don't yet handle the -t- extension.
 
 Examples:
 
-<!-- HTML: no header, colspan -->
-<table><tbody>
-<tr><td>hi-t-<u>en-h0-hybrid</u></td>   <td>Hinglish</td>   <td>Hindi-English hybrid locale</td></tr>
-<tr><td>ta-t-<u>en-h0-hybrid</u></td>   <td>Tanglish</td>   <td>Tamil-English hybrid locale</td></tr>
-<tr><td>ba-t-<u>en-h0-hybrid</u></td>   <td>Banglish</td>   <td>Bangla-English hybrid locale</td></tr>
-<tr><td colspan="3">…</td></tr>
-<tr><td>en-t-<u>hi-h0-hybrid</u></td>   <td>Hinglish</td>   <td>English-Hindi hybrid locale</td></tr>
-<tr><td>en-t-<u>zh-h0-hybrid</u></td>   <td>Chinglish</td>  <td>English-Chinese hybrid locale</td></tr>
-<tr><td colspan="3">…</td></tr>
-</tbody></table>
+|Locale ID			| Base script	| Hybrid name	| Description									|
+|-------------------------------|---------------|---------------|-------------------------------------------------------------------------------|
+|hi-t-***en-h0-hybrid***	| Deva		| Hinglish	| Hindi-English hybrid where the script is Devanagari\*				|
+|hi-Latn-t-***en-h0-hybrid***	| Latin		| Hinglish	| Hindi-English hybrid where the script is Latin\*				|
+|hi-Latn			| Latin		| Hinglish	| Hindi written in Latin script; in practice usually a hybrid with English	|
+|ta-t-***en-h0-hybrid***  	| Tamil		| Tanglish	| Tamil-English hybrid where the script is Tamil\*				|
+|...																		||
+|en-t-***hi-h0-hybrid***	| Latin		| Hinglish	| English-Hindi hybrid	where the script is Latin\*				|
+|en-t-***zh-h0-hybrid***	| Latin		| Chinglish	| English-Chinese hybrid where the script is Latin\*				|
+|...																		||
 
-> _Note: The [unicode_language_id](#unicode_language_id) should be the language used as the ‘scaffold’: for the fallback locale for internationalization services, typically used for more of the core vocabulary/structure in the content. Thus Hinglish should be represented as hi-t-h0-en where Hindi is the scaffold, and as en-t-h0-hi where English is._
+\* When used as a request for international services (such as date formatting), the request is for everything to be in the base script if possible. When used to tag arbitrary content on a coarse level, the expectation is that it be the predominant script — that is, there may be certain passages or phrases that are in the other script but are not tagged on a fine-grained level.
 
-The value of -t- is a full _[unicode_language_id](#unicode_language_id)_, and can contain subtags for region where it is important to include it, as in the following. The value can also include the script, although that is is not normally included: the only instance where it should be is where the content of the source text varies by script. So because zh-Hant has different vocabulary and expressions, it could make sense to have en-t-zh-hant to make that distinction.
+> _Note: The [unicode_language_id](#unicode_language_id) should be the language used as the ‘scaffold’: for the fallback locale for internationalization services, typically used for more of the core vocabulary/structure in the content. Thus where Hindi is the scaffold, Hinglish should be represented as hi-t-en-h0-hybrid (when written in Devanagari script) or hi-Latn-t-en-h0-hybrid (when written in Latin characters). Where English is the scaffold, Hinglish should be represented as en-t-hi-h0-hybrid (or possibly en-Deva-t-hi-h0-hybrid)._
+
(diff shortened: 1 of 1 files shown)


AEApple:
So the one thing I wasn't sure about and I'm happy to update is at least for Indic languages people will generally write in one script or the other and are not likely to switch back and forth which is why Hinglish is clarified as it is. I wasn't sure how true that is for other languages and I wasn't sure if there should be any clarification about whether each language would be in it's own script unless otherwise specified.

From how the spec currently reads to me it computes the locale tag of the scaffold language but I wasn't sure if that is true of the other language, or if the other language follows the scaffold script unless otherwise specified?

markusicu:
> From how the spec currently reads to me it computes the locale tag of the scaffold language but I wasn't sure if that is true of the other language, or if the other language follows the scaffold script unless otherwise specified?

The initial part of the language tag and the part right after the `-t-` are separate [unicode_language_id](https://unicode-org.github.io/cldr/ldml/tr35.html#unicode_language_id)s, so if either has implied subtags, they should be implied separately. Where the spec mentions `tlang` that's how I read it.

AEApple:
Okay, does it otherwise look good? Unless you disagree I would like to explicitly include in the description in each table what script each language's content would be in for better clarity. e.g. For the top row of Hinglish it would be Hindi in Devanagari & English in Latin script (even though I think what people would actually do is 'hi-t-***en-Deva-h0-hybrid***), although I think in Chinese they would do 'zh-t-en-h0-hybrid' with zh in Hans & en in Latin.

macchiati:
> So the one thing I wasn't sure about and I'm happy to update is at least for Indic languages people will generally write in one script or the other and are not likely to switch back and forth which is why Hinglish is clarified as it is. I wasn't sure how true that is for other languages and I wasn't sure if there should be any clarification about whether each language would be in it's own script unless otherwise specified.
> 
> From how the spec currently reads to me it computes the locale tag of the scaffold language but I wasn't sure if that is true of the other language, or if the other language follows the scaffold script unless otherwise specified?

There is a note below that "> Note: The default script for the language is computed without reference to the hybrid subtags. Thus the default script for 'ru' is “Cyrl”, no matter what the source is in the -t- tag." Maybe we should bold that, or move it up?

macchiati:
> Okay, does it otherwise look good? Unless you disagree I would like to explicitly include in the description in each table what script each language's content would be in for better clarity. e.g. For the top row of Hinglish it would be Hindi in Devanagari & English in Latin script (even though I think what people would actually do is 'hi-t-_**en-Deva-h0-hybrid**_), although I think in Chinese they would do 'zh-t-en-h0-hybrid' with zh in Hans & en in Latin.

That's a good idea. It would probably be cleaner to make that an extra column.

docs/ldml/tr35.md
```diff
@@ -1350,38 +1350,37 @@ Hybrid locales have intermixed content from 2 (or more) languages, often with on
...
-<tr><td colspan="3">…</td></tr>
-</tbody></table>
+|Hybrid locale code		| Hybrid name	| Description						|
+|-------------------------------|---------------|-------------------------------------------------------|
+|hi-t-***en-h0-hybrid***	| Hinglish	| Hindi-English hybrid locale written in Devanagari	|
+|hi-Latn-t-***en-h0-hybrid***	| Hinglish	| Hindi-English hybrid locale written in Latin script	|
+|hi-Latn			| Hinglish	| Hindi-English hybrid locale written in Latin script	|
+|ta-t-***en-h0-hybrid***  	| Tanglish	| Tamil-English hybrid locale written in Tamil script	|
+|...													||
+|en-t-***hi-h0-hybrid***	| Hinglish	| English-Hindi hybrid locale written in Latin script	|
```
sffc, line 1367: Suggestion: Clarify when you should prefer en-t-hi-h0-hybrid versus hi-t-en-h0-hybrid

docs/ldml/tr35.md
```diff
@@ -1350,38 +1350,37 @@ Hybrid locales have intermixed content from 2 (or more) languages, often with on
...
-</tbody></table>
+|Hybrid locale code		| Hybrid name	| Description |
+|-------------------------------|---------------|-------------|
+|ru-t-***en-h0-hybrid***	| Runglish	| Russian with an admixture of American English|
+|ru-t-***en-gb-h0-hybrid***	| Runglish	| Russian with an admixture of British English|
+|ru-Latn-t-***-en-gb-h0-hybrid***| Runglish	| Russian with an admixture of British English in Latin script|
+|en-t-***zh-h0-hybrid***	| Chinglish	| American English with an admixture of Chinese (Simplified Mandarin Chinese)|
+|en-t-***zh-Hant-h0-hybrid***	| Chinglish	| American English with an admixture of Chinese (Traditional Mandarin Chinese)|
 
 Should there ever be strong need for hybrids of more than two languages or for other purposes such as hybrid languages as the source of translated content, additional structure could be added.
```
markusicu, line 1385: CLDR-13370 is asking for a change of this sentence as well:
> As a native English speaker, this was ambiguous to me (and implied that adding more structure — e.g. `-h0-` tags— this could be done.  I've submitted a ticket for enabling such things, but until then, this sentence should state unequivocally “It is not currently possible to repreesent hybrids of more than two languages, nor enable hybrid languages as the source of translated content.  Please file a ticket if you need this support”.

@macchiati do you agree? Should we change it as suggested, or reword it?

docs/ldml/tr35.md
markusicu: That's already in the note under the examples table, except that we are also fixing the ill-formed language tag syntax in that note.

markusicu:
> > ... I would like to explicitly include in the description in each table what script each language's content would be in for better clarity. e.g. For the top row of Hinglish it would be Hindi in Devanagari & English in Latin script (even though I think what people would actually do is 'hi-t-_**en-Deva-h0-hybrid**_), although I think in Chinese they would do 'zh-t-en-h0-hybrid' with zh in Hans & en in Latin.
> 
> That's a good idea. It would probably be cleaner to make that an extra column.

Good catch, @AEApple! I had assumed that the hybrid mixture is in the script of the initial unicode_language_id, but yes, it makes sense that the mixed-in language remains in its own implied script unless specified otherwise.

The most popular form of Hinglish is still `hi-Latn-t-en-h0-hybrid`.

AEApple:
> > > ... I would like to explicitly include in the description in each table what script each language's content would be in for better clarity. e.g. For the top row of Hinglish it would be Hindi in Devanagari & English in Latin script (even though I think what people would actually do is 'hi-t-_**en-Deva-h0-hybrid**_), although I think in Chinese they would do 'zh-t-en-h0-hybrid' with zh in Hans & en in Latin.
> > 
> > 
> > That's a good idea. It would probably be cleaner to make that an extra column.
> 
> Good catch, @AEApple! I had assumed that the hybrid mixture is in the script of the initial unicode_language_id, but yes, it makes sense that the mixed-in language remains in its own implied script unless specified otherwise.
> 
> The most popular form of Hinglish is still `hi-Latn-t-en-h0-hybrid`.

Yes in software :) not as sure about when written elsewhere

AEApple:
Okay, I think I picked up all the comments, can you please take a look again and let me know if you think I missed anything?

re 'As a native English speaker, this was ambiguous to me (and implied that adding more structure ...' I read the spec and I thought that had already been addressed earlier in a previous ticket, but happy to update that line as well if it's still not clear.

sffc (approved):
I haven't reviewed the ticket in detail but the changes seem fine

macchiati (changes requested):
Ah, your wording exposes a flaw in the specification.

The "output" of the -t- will be in a single script, the script of the base language (or the default script of the base language).

So hi-t-en-h0-hybrid will not be "Hindi-English hybrid locale where Hindi is in Devanagari script and English in Latin"

"Hindi-English hybrid locale in Devanagari"

I will write more tomorrow.


--


<br class="Apple-interchange-newline">

srl295 (dismissed):

srl295:
I'm removing my review until the script issues noted by Mark are addressed

macchiati:
I haven't seen any changes since my last comments.

Note: hi-t-en-h0-deva-hybrid is ill-formed; it would have to be hi-t-en-deva-h0-hybrid to be syntactically correct. But the script doesn't have any effect in that position, so it is equivalent to hi-t-en-h0-hybrid

Maybe it would help to have something like:

|Locale ID|Base Script|Description|
|-|-|
|hi-Latn-t-en-h0-hybrid|Latin| Hindi-English hybrid locale where the script is Latin*|
|hi-Latn|Latin| Equivalent to hi-Latn-t-en-h0-hybrid|
|hi-Deva-t-en-h0-hybrid|Devanagari| Hindi-English hybrid locale where the script is predominantly Devanagari*|

\* The request is that all or most of the text be in this script.

docs/ldml/tr35.md
```diff
@@ -1350,38 +1350,39 @@ Hybrid locales have intermixed content from 2 (or more) languages, often with on
...
-<tr><td colspan="3">…</td></tr>
-<tr><td>en-t-<u>hi-h0-hybrid</u></td>   <td>Hinglish</td>   <td>English-Hindi hybrid locale</td></tr>
-<tr><td>en-t-<u>zh-h0-hybrid</u></td>   <td>Chinglish</td>  <td>English-Chinese hybrid locale</td></tr>
-<tr><td colspan="3">…</td></tr>
-</tbody></table>
+|Locale ID			| Base script	| Hybrid name	| Description									|
+|-------------------------------|---------------|---------------|-------------------------------------------------------------------------------|
+|hi-t-***en-h0-hybrid***	| Deva		| Hinglish	| Hindi-English hybrid where the script is Devanagari*				|
+|hi-Latn-t-***en-h0-hybrid***	| Latin		| Hinglish	| Hindi-English hybrid where the script is Latin*				|
+|hi-Latn			| Latin		| Hinglish	| Hindi written in Latin script in practice usually a hybrid with English	|
```
markusicu, line 1364:
```suggestion
|hi-Latn			| Latin		| Hinglish	| Hindi written in Latin script; in practice usually a hybrid with English	|
```

docs/ldml/tr35.md
```diff
@@ -1350,38 +1350,39 @@ Hybrid locales have intermixed content from 2 (or more) languages, often with on
...
 
-> _Note: The [unicode_language_id](#unicode_language_id) should be the language used as the ‘scaffold’: for the fallback locale for internationalization services, typically used for more of the core vocabulary/structure in the content. Thus Hinglish should be represented as hi-t-h0-en where Hindi is the scaffold, and as en-t-h0-hi where English is._
+> _Note: The [unicode_language_id](#unicode_language_id) should be the language used as the ‘scaffold’: for the fallback locale for internationalization services, typically used for more of the core vocabulary/structure in the content. Thus where Hindi is the scaffold, Hinglish should be represented as hi-t-en-h0 (when written in Devanagari) or hi-Latn-en-h0 (when written in Latin characters). Where English is the scaffold, Hinglish should be represented as en-t-hi-h0 (or possibly en-Deva-t-hi-h0)._
```
markusicu, line 1373: The syntax here went back to ill-formed language tags.
```suggestion
> _Note: The [unicode_language_id](#unicode_language_id) should be the language used as the ‘scaffold’: for the fallback locale for internationalization services, typically used for more of the core vocabulary/structure in the content. Thus where Hindi is the scaffold, Hinglish should be represented as hi-t-en-h0-hybrid (when written in Devanagari script) or hi-Latn-t-en-h0-hybrid (when written in Latin characters). Where English is the scaffold, Hinglish should be represented as en-t-hi-h0-hybrid (or possibly en-Deva-t-hi-h0-hybrid)._
```

macchiati: Good catches, thanks

markusicu (approved):
lgtm tnx

markusicu:
> Hooray! The files in the branch are the same across the force-push. 😃

Hi @AEApple, it looks like you are proactively squashing your branch. That's not actually necessary, since CLDR uses squash-and-merge, right? Or am I missing something?

(During review, it's actually useful not to squash or rebase unless necessary, because when you do those, GitHub has an even harder time tracking comments to code than without those.)

AEApple:
I keep thinking I'm done :) I won't in the future if you find it better it just then fails the single commit and the jira-ticket check....

macchiati (approved):

PR #2395, opened 2022-09-27
CLDR-15618 spec: improve docs about value+children

srl295:
- document the invariant against nondistinguishing elements and children
- update the test to link to the docs
- test: add an unreached enum value to the switch statement

CLDR-15618

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -182,7 +182,7 @@ The LDML specification is divided into the following parts:
   * Table: [Part 6 Links](#Part_6_Links): [Supplemental](tr35-info.md) (supplemental data)
   * Table: [Part 7 Links](#Part_7_Links): [Keyboards](tr35-keyboards.md) (keyboard mappings)
 * [Annex C. LocaleId Canonicalization](#LocaleId_Canonicalization)
-  * [Definitions](#definitions)
+  * [LocaleId Definitions](#LocaleId_Definitions)
     * [1. Multimap interpretation](#1.-multimap-interpretation)
     * [2. Alias elements](#2.-alias-elements)
     * [3. Matches](#3.-matches)
@@ -1737,7 +1737,9 @@ Attributes that serve to distinguish multiple elements at the same level are cal
 <language type="ab">Abkhazian</language>
 ```
 
-Distinguishing attributes affect inheritance; two elements with different distinguishing attributes are treated as different for purposes of inheritance. For more information, see [Section 5.5 Valid Attribute Values](#Valid_Attribute_Values). Other attributes are called nondistinguishing (or informational) attributes. These carry separate information, and do not affect inheritance.
+Distinguishing attributes affect inheritance; two elements with different distinguishing attributes are treated as different for purposes of inheritance. For more information, see [Section 5.5 Valid Attribute Values](#Valid_Attribute_Values). Other attributes are called value attributes. Value attributes do not affect inheritance, and elements with value attributes may not have child elements (see [XML Format](#XML_Format)).
+
+Non-distinguishing attributes are identified by [DTD Annotations](#DTD_Annotations) such as `@VALUE`.
 
 For any element in an XML file, _an element chain_ is a resolved [[XPath](#XPath)] leading from the root to an element, with attributes on each element in alphabetical order. So in, say, [https://github.com/unicode-org/cldr/blob/main/common/main/el.xml](https://github.com/unicode-org/cldr/blob/main/common/main/el.xml) we may have:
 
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
```diff
@@ -222,6 +223,9 @@ private void checkEmpty(Multimap<String, String> m, DtdType type, Element elemen
             case distinguished:
                 distAttributes.add(attribute);
                 break;
+            case metadata:
+                metadataAttributes.add(attribute);
+                break;
```
srl295, line 228: This was previously missed in the switch statement. We don't do anything with these currently, but could.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
```diff
@@ -269,7 +273,7 @@ private void checkEmpty(Multimap<String, String> m, DtdType type, Element elemen
                     logKnownIssue("cldrbug:9982", "Lower priority fixes to bad xml");
                     break;
                 default:
-                    m.put("error", "\t||" + showPath(parents) + "||path has both children AND value attributes"
+                    m.put("error", "\t||" + showPath(parents) + "||DTD has both children AND value attributes: tr35.md#XML_Format"
```
srl295, line 276: This line is already _very long_ in practice, and will (should) only be hit by poor souls trying to update the DTD, so did not include the full URL.

srl295:
cc'ing keyboard folks because we discussed this extensively in the keyboard-sc

pedberg-icu (approved):

mcdurdin (approved):

srl295:
and of course i conflicted with myself in #2396 … please re review

pedberg-icu (approved):

mcdurdin (approved):

docs/ldml/tr35.md
```diff
@@ -2323,12 +2323,57 @@ The XML structure is stable over releases. Elements and attributes may be deprec
...
+<parent id="abc"> <!-- Allowed: 'id' is distinguishing -->
+  <someValue value="Some Value"/>
+  <text>Here is some text</text>
+  <child id="def"/>
+  <child id="ghi"/>
+</parent>
+```
 
-Rule elements do not have this restriction, but also do not inherit, except as an entire block. The rule elements are listed in serialElements in the supplemental metadata. See also _[Section 4.2 Inheritance and Validity](#Inheritance_and_Validity)_. For more technical details, see [Updating-DTDs](https://cldr.unicode.org/development/updating-dtds).
+Rule elements do not have these restrictions, but also do not inherit, except as an entire block. The rule elements are listed in serialElements in the supplemental metadata. See also _[Section 4.2 Inheritance and Validity](#Inheritance_and_Validity)_. For more technical details, see [Updating-DTDs](https://cldr.unicode.org/development/updating-dtds).
```
macchiati, line 2376: This is old, and we might as well fix it while you are here.

Instead of serialElements we have the DTD annotation @ORDERED
`<!ELEMENT rbnfrule ( #PCDATA ) >`
`   <!--@ORDERED-->`

docs/ldml/tr35.md
```diff
@@ -2323,12 +2323,57 @@ The XML structure is stable over releases. Elements and attributes may be deprec
...
+```xml
+<parent id="abc"> <!-- Allowed: 'id' is distinguishing -->
+  <someValue value="Some Value"/>
+  <text>Here is some text</text>
+  <child id="def"/>
+  <child id="ghi"/>
+</parent>
+```
+
+The following fictional (non-LDML) example illustrates the first and third restrictions on structure elements:
```
macchiati, line 2355: This looks like a repeat. Also, would be better to use real elements.

macchiati (changes requested):

docs/ldml/tr35.md
macchiati: I suggest
`		<localeDisplayPattern someValueAttribute="foo">
                         some text data
			<localePattern>{0} ({1})</localePattern>
			<localeSeparator>{0}, {1}</localeSeparator>
`

docs/ldml/tr35.md
```diff
@@ -3403,7 +3432,7 @@ The `languageAlias`, `scriptAlias`, `territoryAlias`, and `variantAlias` element
 > canonicalization.
 >See §3.8.2 [Legacy Variants](#Legacy_Variants).
 
-### Definitions
+### <a name="LocaeId_Definitions">LocaleId Definitions</a>
```
macchiati, line 3435: LocaeId_Definitions
=>
LocaleId_Definitions

macchiati (changes requested):

docs/ldml/tr35.md
macchiati: Otherwise looks good.

macchiati (approved):

pedberg-icu:
I will go ahead and merge, hope that is OK

PR #2396, opened 2022-09-27
CLDR-15524 doc: fix http to https

srl295:
- also fix one bad link broken in CLDR-15655

CLDR-15524


srl295:
> I don't think we should change to _exactly_ what is shown for some of these. I recommend to use https://www.unicode.org rather than https://unicode.org _sans_ www.

i can change https://unicode.org to https://www.unicode.org , will do

srl295:
@rickmcgowan check now

rickmcgowan (approved):
Updates seem to be in order, thanks. I hope someone is running a link check when this is all done, to make sure all links resolve, especially the outside links to external (non Unicode) sites. :-)

pedberg-icu (approved):

.github/pull_request_template.md
```diff
@@ -4,7 +4,7 @@ CLDR-_____
 
 <!--
 Thank you for your pull request.
-Please see http://cldr.unicode.org/index/process for general
```
kachgold, line 7: .github/pull_request_template.md

kachgold (approved):

macchiati:
*www.*xxxxxx.yyyy was the style 20 years ago. But no longer. What you see
is xxxxxx.yyyy.

Now, they both need to resolve to the same content in any event. But there
is no need to hold on to the Model T approach.

PR #2397, opened 2022-09-27
CLDR-15613 More info on matching pattern field lengths to requested skeleton field lengths

pedberg-icu:
CLDR-15613


richgillam (approved):
These changes all look good and make sense.

As I remember it, the issues I dealt with in this area had to do mostly with the hour and day-period fields.  The original logic (maybe just in DateIntervalFormat, I don't remember) didn't recognize different lengths of the day-period field and didn't let the user use "b" (I think "B" is handled separately).  I also seem to remember that the caller could choose an hour cycle with "h" or "H", but you couldn't use "k" and "K".  So we might need language that covers these issues (I could go through and see what else I messed with-- I'm sure there were other things), but none of that directly affects the changes you're making here, which look great.

PR #2398, opened 2022-09-28
CLDR-13722 Roll in Person Names body converted from GoogleDoc (pass 1)

pedberg-icu:
CLDR-13722

Here is a first pass at converting the Person Names spec GoogleDoc version (https://docs.google.com/document/d/1uvv6gdkuFwtbNV26Pk7ddfZult4unYwR6DnnKYbujUo/edit?pli=1) to markdown (using a tool plus lots of hand editing) and rolling into the existing stub. We won't really know what this looks like (as processed by github actions) until the PR is merged and we can see the html posted to smoketest, so there will probably be some iterations on this. It will be hard to tell much from looking at the markdown additions in this PR.

Note that I did leave one pair of tables (with spans) as embedded html, at least for now.


macchiati:
Note: we can see most of the formatting in the files view with the right button:
Screen Shot 2022-09-28 at 07 09 22
I use two windows side by side when reviewing (or fixing my own .md PRs), one in the diff view and one with the rich diff.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+* 8 [PersonName Data Interface Examples](#8-personname-data-interface-examples)
 
-## 1 <a name="AaAaaa" href="#AaAaaa">Aa Aaaa</a>
+## 1 <a name="CLDRPersonNames" href="#CLDRPersonNames">CLDR Person Names</a>
 
-Section 1 body text xxxxx.
+### 1.1 <a name="Introduction" href="#Introduction">Introduction</a>
 
-### 1.1 <a name="BbBbbb" href="#BbBbbb">Bb Bbbb</a>
+CLDR provides formatting for person names, such as John Smith or 宮崎駿. These use patterns to show how a name record (for example, from a database) should be formatted for a particular locale. Name data has fields for the parts of people’s names, such as a **given **field with a value of “Maria”, and a **surname **field value of “Schmidt”.
```
macchiati, line 86: I think the format \*\*given \*\*field isn't interpreted right; it needs to be \*\*given\*\* field (the space moved over). Same for italics. Multiple instances, so easiest to search for (\*\*|\_\_) and adjust.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+* Some languages, such as Spanish, have two surnames (where each can be composed of multiple words).
+* The ordering of name fields can be different across languages, as well as the spacing (or lack thereof) and punctuation.
+* Name formatting needs to be adapted to different circumstances, such as a need to be presented shorter or longer; formal or informal context; or when talking about someone, or talking to someone, or as a monogram (JFK).
 
-Section 2 body text xxxxx.
+This document provides the [LDML](http://www.unicode.org/reports/tr35/) specification for formatting of personal names, using data, structure, and examples. 
 
-### 2.1 <a name="DdDddd" href="#DdDddd">Dd Cddd</a>
+    **This is a technology preview; thus not intended for production software (except where itself marked as a technology preview). We have gathered a first round of data, and intend to refine the way in which we gather data. We are looking for additional feedback on the tech preview so that we can make improvements. For example, there are a few areas where we intend enhancements: handling native vs foreign names (in the native script); handling prefix and suffix fields; and so on.**
```
macchiati, line 98: for block quotes the following works https://github.github.com/gfm/#block-quotes

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+The information that is to be formatted logically consists of a data object containing a number of fields. This data object is a construct for the purpose of formatting, and doesn’t represent the source of the name data. That is, the original source may contain more information. The PersonName object is merely a logical ‘transport’ of information to formatting; it may physically consist of, for example, an API that fetches fields from a database.
+
+Note that an application might have more than one set of name data for a given person, such as data for both a legal name and a nickname or preferred name. Or the source data may contain two whole sets of name data for a person from an Eastern Slavic region, one in Cyrillic characters and one in Latin characters. Or it might contain phonetic data for a name (commonly used in Japan). The additional application-specific information in person’s names is out of scope for the CLDR Person Name formatting data. Thus a calling application may produce more than one name object to format depending on the purpose. 
+
+For illustration, the following is a sample PersonName object.
+
+```
+person_name ⇒
```
macchiati, line 319: We probably want to make this a table. (just a nice-to-have)

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+| Parameter    | Description |
+| ------------ | ----------- |
+| `addressing` | Used when speaking “to” a person, or “vocative” case. This may also have an effect on the formality.<br>example: “Robert” [medium, informal] |
+| `referring`  | Used when speaking “about” a person, or “nominative” case.<br>example: “Robert Smith”  [medium, informal] |
+| `monogram`   | The `monogram` usage is for a specific abbreviated form.<br>Example: monogram for Robert James Smith may be “RS” or “RJS”.<br><br>`usage=”referring”, formality=”formal”`<br>_“RJS”_ |
+
+Slavic languages provide a good  example of `addressing` vs `referring`. An example _uk-Cyrl_ PersonName object:
+
+```
+person_name ⇒
```
macchiati, line 385: also a table here

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+| `informal` | A less formal name for the individual. The composition depends upon the language. For example, a language might exclude the prefix, suffix and given2 (middle) name. Depending on the length, it may also exclude the surname. The formatting algorithm should choose any passed in name data that has an _informal_ attribute, if available.<br><br>`length="medium", formality="informal"`<br>“Bob Smith” |
+
+## 5 <a name="5-namepattern-syntax" href="#5-namepattern-syntax">namePattern Syntax</a>
+
+A _namePattern_  is composed of a sequence of field IDs, each enclosed in curly braces, and separated by zero or more literal characters (eg, space or comma + space). An Extended Backus Normal Form (EBNF) is used to describe the namePattern format for a specific set of attributes. It has the following structure. This is the `( #PCDATA )` reference in the element specification above.
+
+|              | EBNF           | Comments    |
+| ------------ | -------------- | ----------- |
+| namePattern  | = literal?<br>( modField  literal? )+; | Two literals cannot be adjacent |
+| modField     | = '{' field modifierList? '}';         | Parts of a name, described below |
```
macchiati, line 422: Change "Parts of a name, described below" to a name field, optionally modified.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+| field        | = 'prefix'<br>&#124; 'given'<br>&#124; 'given2'<br>&#124; 'surname'<br>&#124; 'surname2'<br>&#124; 'suffix' ; | Parts of a name, described below |
```
macchiati, line 423: Change "Parts of a name, described below" to be pointer to the descriptions of them.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+| modifierList | = '-informal'?<br>( '-allCaps' &#124; ‘-initialCap' )?<br>( '-initial' &#124; '-monogram' )?<br>( '-prefix' &#124; '-core' )? | Optional modifiers that can be applied to name parts. Note that some modifiers are exclusive: only `prefix` or `core`, only `initial` or `monogram`, only `allCaps` or `initialCap`. |
```
macchiati, line 424: Add pointer to descriptions

docs/ldml/tr35-personNames.md
```diff
@@ -5,7 +5,7 @@
 <!-- HTML: no header -->
```
macchiati, line 5: Looks great, thanks for all the hard work!

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+In some cultures, two surnames are used to indicate the paternal and maternal family names or generational names indicating father, grandfather. The `surname2` field is used to indicate this. The CLDR PersonName formatting data assumes that if a PersonName object to be formatted does not have two surnames, then the `surname2` field is not populated. (That is, no pattern should have a `surname2` field without a surname field.) Order of fields in a pattern can vary arbitrarily by locale.
+
+In most cultures, there is a concept of nickname or preferred name, which is used in informal settings or sometimes to represent a “public” or “stage name”. The nickname or preferred name may be submitted as a separate PersonName object to be formatted, or included with a modifier such as `given-informal`. 
+
+| Field      | Description |
+| ---------- | ----------- |
+| `prefix`   | Typically a title, honorific, or generational qualifier.<br>Example: ‘Ms.’, ‘Mr.’, ’Dr’, ‘President’<br><br>Note that CLDR PersonaName formats data does not define regional or locale-specific lists of prefixes, honorifics, or titles such as “Mr”, “Ms”, “Mx”, “Prof”, “Jr”, etc. |
```
macchiati, line 441: Add: The values are as supplied by the name object, via the PersonName data interface.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+Some other examples:
+
+* British name: _John Ronald Reuel Tolkien_: `given` name is "John", `given2` name would be  "Ronald Reuel", and the `surame` is "Tolkien".
+* Dutch name: _Anneliese Louise van der Pol_: `given` name: "Anneliese", `given2` name: "Louise", `surname`: "van der Pol"
+    * Also surname-prefix: “van der”, surname-core: “Pol” — see below.	
+* French name: “Jean-Louis Trintignant” would _not _be Jean (`given`) Louis (`given2`) Trintignant (`surname`), since “Louis” wouldn’t be discarded when formatting. Instead it would be Jean-Louis (`given`) Trintignant (`surname`)
+
+Note: If the legal name, stage name, etc. are substantially different, then that information can be logically in a separate Name record. That is, it is up to the implementation to maintain any distinctions that are important to it: CLDR PersonName formats is focusing on formatting a Name record that is given to it.
```
macchiati, line 455: Check for "name record" and change to "name object"

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+| allCaps    | Requests the element in all caps, which is desired In some contexts. For example, a new guideline in Japan is that for the Latin representation of Japanese names, the family name comes first and is presented in all capitals. This would be represented as<br>“{surname-allCaps} {given}”<br><br>Hayao Miyazaki (宮崎 駿) would be represented in Latin characters in Japan (ja-Latn-JP) as “_MIYAZAKI Hayao_”<br><br>_The default implementation uses the default Unicode uppercase algorithm; if the PersonName object being formatted has a locale, and CLDR supports a locale-specific algorithm for that locale, then that algorithm is used. The PersonName object can override this, as detailed below._<br><br>Only the “-_allCaps_” or the “-_initalCap_” modifier may be used, but not both. They are mutually exclusive. |
```
macchiati, line 475: check for instances of ", as detailed below." In this case, it isn't needed. In others, may want to point to a part of the document.

macchiati: Also, we probably want to replace "name object" by "PersonName object"

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+The patterns are in personName elements, which are themselves in a personNames container element. The following describes how these patterns are chosen. If the name locale is different than the formatting locale, then additional processing needs to take place: see [Handling foreign names](#6-8-handling-foreign-names).
+
+The details of the XML structure behind the data referenced here are in [XML Structure](#2-xml-structure).
+
+### 6.1 <a name="6-1-derive-the-name-locale" href="#6-1-derive-the-name-locale">Derive the name locale</a>
+
+If there is no name locale available via the PersonName data interface, then an effective name locale is created based on the characters in the name and the formatting locale. 
+
+* Get the predominant script of the given and surname, and find the likely language as specified by the likely subtags. Use the locale form from the likely language and the script as the name locale.
```
macchiati, line 497: Use the locale form from => Use the locale identifier constructed from the

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+### 6.2 <a name="6-2-derive-the-formatting-locale" href="#6-2-derive-the-formatting-locale">Derive the formatting locale</a>
+
+If the name locale is different from the formatting locale, and the script of the name is different from the script of the formatting locale, then change the formatting locale to the name locale.
+
+### 6.3 <a name="6-3-derive-the-name-order" href="#6-3-derive-the-name-order">Derive the name order</a>
+
+The name order data provides a mapping from the name locale to an order parameter. In current data, the important part of the name locale is the region code, but allowing a locale code provides for easy extension in the future.`  `
```
macchiati, line 505: Probably drop "In current data, the important part of the name locale is the region code, but allowing a locale code provides for easy extension in the future."

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+<nameOrderLocales order="surnameFirst">zh ja und-CN und-TW und-SG und-HK und-MO und-HU und-JP</nameOrderLocales>
+```
+
+The nameOrderLocales will match any locale with a zh or ja [unicode_language_subtag](https://unicode.org/reports/tr35/#unicode_language_subtag) and any locale with a CN, TW, SG, HK MO, HU, or JP [unicode_region_subtag](https://unicode.org/reports/tr35/#unicode_region_subtag).
+
+Here are some more examples. Note that if there is no order field or locale field in the PersonName object to be formatted, and the script of the PersonName data is different from that of the formatting locale, then the default result is givenFirst.
+
+| Name Object preferredOrder | Name Object locale | Resulting Order |
+| -------------------------- | ------------------ | --------------- |
+| surnameFirst               |                    | surnameFirst    |
```
macchiati, line 529: for the name object locale in this line, maybe put <any>?

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+* The usage attribute values contain U or there is no usage attribute, and
+* The formality attribute values contain F or there is no formality attribute
+
+Example for input parameters 
+
+    order=**givenFirst** length=**long** usage=**referring** formality=**formal**
+
+To match a personName, all four attributes in the personName must match:
+
+| Sample personName attributes                                 | Matches? | Comment |
```
macchiati, line 574: We need to fix this, since we changed our model. It can be simpler; there is either single value or there is no attribute at all. And "no attribute at all" matches any attribute value.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+| ------- | -------- | --------- |
+| Foo     |          | Baz       |
+
+The output is:
+
+    F. Baz
+
+### 6.7 <a name="6-7-deriving-initials" href="#6-7-deriving-initials">Deriving initials</a>
+
+The following process is used to produce initials when they are not supplied by the name object.
```
macchiati, line 733: Add: The input example is "Mary Beth"

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+See the “initial” modifier in the [Modifiers](#5-2-modifiers) section for more details.
+
+### 6.8 <a name="6-8-handling-foreign-names" href="#6-8-handling-foreign-names">Handling foreign names</a>
+
+There are two main challenges in dealing with foreign name formatting that needs to be considered. One is the ordering, which is dealt with under the section [[2.3 nameOrderLocales Element](#2-3-nameorderlocales-element)]. The other is spacing.
+
+Some writing systems require spaces (or some other non-letters) to separate words. For example, [Hayao Miyazaki](https://en.wikipedia.org/wiki/Hayao_Miyazaki) is written in English with given name first and with a space between the two name fields, while in Japanese there is no space with surname first: 
+
+    [宮崎駿](https://ja.wikipedia.org/wiki/%E5%AE%AE%E5%B4%8E%E9%A7%BF)
```
macchiati, line 750: Not sure why this isn't interpreted right in markdown

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+
+### 7.1 <a name="7-1-syntax" href="#7-1-syntax">Syntax</a>
+
+```xml
+<!ELEMENT sampleName ( nameField+ )  >
+<!ATTLIST sampleName item NMTOKENS #REQUIRED >
+```
+
+* `NMTOKENS` must be one of `( givenOnly | givenSurnameOnly | given12Surname | full )`
+
```
macchiati, line 932: Add: However, these may change arbitrarily in the future.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,929 @@ The LDML specification is divided into the following parts:
...
+* `givenOnly` is for an example name with only the given is presented: “Richard” or “Rich” (informal)
+* `givenSurnameOnly` is for an example name with only the given name and surname: “Richard Smith” or “Rich Smith” (informal)
+* `given12Surname` is for an example using all given names and a surname: “Richard Edward Smith” and “Rich E. Smith” (informal)
+* `full` is used to present a name using all fields: “Mr. Richard Edward Smith Iglesias, Ph.D.”
+
+The `nameField` values and their modifiers are described in the [Person Name Object](#3-person-name-object) and [namePattern Syntax](#5-namepattern-syntax) sections.
+
+## 8 <a name="8-personname-data-interface-examples" href="#8-personname-data-interface-examples">PersonName Data Interface Examples</a>
+
+### Example 1. {#example-1}
```
macchiati, line 946: the {#example 1} needs to be changed to a link reference, same below.

macchiati (changes requested):

docs/ldml/tr35-personNames.md
pedberg-icu: Thanks, fixed these globally (was due to style misalignment in the GoogleDoc)

docs/ldml/tr35-personNames.md
pedberg-icu: Changed to use ">"

docs/ldml/tr35-personNames.md
pedberg-icu: Yes, converted to table, done.

docs/ldml/tr35-personNames.md
pedberg-icu: Yes, converted to table, done.

docs/ldml/tr35-personNames.md
pedberg-icu: Done

docs/ldml/tr35-personNames.md
pedberg-icu: Done

docs/ldml/tr35-personNames.md
pedberg-icu: Done

docs/ldml/tr35-personNames.md
pedberg-icu: Done (in header)

docs/ldml/tr35-personNames.md
pedberg-icu: Actually I changed to PersonName object everywhere (per your later comment), except in the first occurrence in the Intro where we are speaking generally, which I just changed to name object.

docs/ldml/tr35-personNames.md
pedberg-icu: For ", as detailed below." it was just 3 places in this table which referred to steps immediately following the table, so I did not add a pointer . I still could in a future version I guess, if it seems warranted.

docs/ldml/tr35-personNames.md
pedberg-icu: Done

docs/ldml/tr35-personNames.md
pedberg-icu: Done

docs/ldml/tr35-personNames.md
pedberg-icu: Added a note and deleted a couple of table rows.

docs/ldml/tr35-personNames.md
pedberg-icu: Done

docs/ldml/tr35-personNames.md
pedberg-icu: Changed to use block quote instead of indent (which implies code block)

docs/ldml/tr35-personNames.md
pedberg-icu: Thanks, fixed. These were still in the form created by the conversion from GoogleDocs. Fixed them to the standard form.

pedberg-icu:
@macchiati Thanks for your detailed review, addressed all of your comments, please re-check and approve if OK.

docs/ldml/tr35-personNames.md
```diff
@@ -494,25 +491,25 @@ The details of the XML structure behind the data referenced here are in [XML Str
...
+* Get the predominant script of the given and surname, and find the likely language as specified by the likely subtags. Use the locale identifier constructed from the likely language and the script as the name locale.
 
 ### 6.2 <a name="6-2-derive-the-formatting-locale" href="#6-2-derive-the-formatting-locale">Derive the formatting locale</a>
 
 If the name locale is different from the formatting locale, and the script of the name is different from the script of the formatting locale, then change the formatting locale to the name locale.
 
 ### 6.3 <a name="6-3-derive-the-name-order" href="#6-3-derive-the-name-order">Derive the name order</a>
 
-The name order data provides a mapping from the name locale to an order parameter. In current data, the important part of the name locale is the region code, but allowing a locale code provides for easy extension in the future.`  `
+The name order data provides a mapping from the name locale to an order parameter.`  `
```
macchiati, line 502: minor, trailing `  `

macchiati (approved):

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+
+The patterns are in personName elements, which are themselves in a personNames container element. The following describes how these patterns are chosen. If the name locale is different than the formatting locale, then additional processing needs to take place: see [Handling foreign names](#6-8-handling-foreign-names).
+
+The details of the XML structure behind the data referenced here are in [XML Structure](#2-xml-structure).
+
+### 6.1 <a name="6-1-derive-the-name-locale" href="#6-1-derive-the-name-locale">Derive the name locale</a>
+
+If there is no name locale available via the PersonName data interface, then an effective name locale is created based on the characters in the name and the formatting locale. 
+
+* Get the predominant script of the given and surname, and find the likely language as specified by the likely subtags. Use the locale identifier constructed from the likely language and the script as the name locale.
```
richgillam, line 494: I think we left out a step here.  If the name doesn't specify a locale, we check the name's script, and if the name's script is supported by the formatting locale, the name locale becomes the formatting locale.  If the name's script is _different_ from the formatting locale, _then_ we use the likely-subtags data to derive a name locale.

I'm also wondering if we want to nail down what we mean by the "predominant script" or just leave that unspecified for now.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+### 6.3 <a name="6-3-derive-the-name-order" href="#6-3-derive-the-name-order">Derive the name order</a>
+
+The name order data provides a mapping from the name locale to an order parameter.`  `
+
+A PersonName object’s fields are used to derive an order, as follows:
+
+1. If the PersonName object to be formatted has a `preferredOrder` field, then return that field’s value
+2. Otherwise, if the PersonName object has a `nameLocale` field, and there is a `nameOrderLocales` element:
+    1. Map the `nameLocale` field using the likelySubtags data, eg `zh` ⇒ `zh_Hans_CN`
+    2. Traverse the `nameOrderLocales` elements, and within each order attribute value to see if there is a match.
```
richgillam, line 509: I think we're missing a couple steps here too.  We expand the name locale to include all the optional parts and search both `nameOrderLocales` list for the expanded name locale.  If we don't find it, we lop a field off the end and try again, and if we lop off all the fields, we search the elements for `und`.  The lopping-off part is missing.

But there are two wrinkles to that, I think:
1. Do we really need the script code here?  I'm a little worried it might interact uncomfortably with the script-based selection of name and formatting locales.
2. The truncation-fallback logic I outline above (and that I think both Mark and I implemented in our implementations) doesn't allow for `und` to be used as a wildcard language (so you can use `und_CN` to mean "any language, but in China".  I don't know if we're still mentioning this usage in this doc (I don't remember seeing it this far), but if we want to support that, we need a different algorithm.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+
+The name order data provides a mapping from the name locale to an order parameter.`  `
+
+A PersonName object’s fields are used to derive an order, as follows:
+
+1. If the PersonName object to be formatted has a `preferredOrder` field, then return that field’s value
+2. Otherwise, if the PersonName object has a `nameLocale` field, and there is a `nameOrderLocales` element:
+    1. Map the `nameLocale` field using the likelySubtags data, eg `zh` ⇒ `zh_Hans_CN`
+    2. Traverse the `nameOrderLocales` elements, and within each order attribute value to see if there is a match. 
+    3. As usual, an **und** (undefined) language subtag or empty script or region in the order attribute locale matches any language subtag, script subtag, or region subtag (respectively) in the PersonName object’s locale field.
```
richgillam, line 510: I guess this is intended to address my comments above, but I think we need to be clearer about how the matching is supposed to happen.  Also, what happens if multiple items match?  If the lists mention `zh`, `und_CN`, and `und` and the name locale is `zh_CN`, they all match.  We all agree that `und` is the match of last resort, but between `zh` and `und_CN`, which one wins, and how do we decide that?  I don't think the answer should be based on which order they locale IDs are listed in the resource.

(My current logic just does ordinary truncation fallback, by the way-- can somebody file a ticket to request something that'll actually work with `und_CN`?)

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+A set of input parameters { order=O length=L usage=U formality=F } matches a personName element when:
+
+* The order attribute values contain O or there is no order attribute, and
+* The length attribute values contain L or there is no length attribute, and
+* The usage attribute values contain U or there is no usage attribute, and
+* The formality attribute values contain F or there is no formality attribute
+
+Example for input parameters 
+
+    order=**givenFirst** length=**long** usage=**referring** formality=**formal**
```
richgillam, line 567: At least in GitHub's Markdown view, the asterisks around `givenFirst`, `long`, `referring`, and `formal` are visible; they're not causing those words to show up in boldface.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+| :----------------------------------------------------------- | :------: | :------ |
+| `order=_"givenFirst"_ length=_"long"_ usage=_"referring"_ formality=_"formal"_` | Y | exact match |
+| `length=_"long"_ usage=_"referring"_ formality=_"informal"_` | N | mismatch for formality |
+| `length=_"long"_ formality=_"formal"_`                       | Y | missing usage = all! |
+
+To find the matching personName element, traverse all the personNames in order until the first one is found. This will always terminate since the data is well-formed in CLDR.
+
+### 6.5 <a name="6-5-choose-a-namepattern" href="#6-5-choose-a-namepattern">Choose a namePattern</a>
+
+To format a name, the fields in a namePattern are replaced with the actual PersonName object to be formatted values from the input record. The personName element can contain multiple namePattern elements. We choose one based on the fields in the input PersonName object that are populated: The namePattern that has the most populated fields wins; if more than one have the same number of populated fields, the one with the least unpopulated fields wins. For example:
```
richgillam, line 581: "...are replaced with the actual PersonName object to be formatted values from the input record" reads really funny.  Maybe "...are replaced with the values of the named fields in the actual PersonNameObject"?

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+
+1. Pattern A has 12 fields total, pattern B has 10 fields total, and pattern C has 8 fields total.
+2. Both patterns A and B can be populated with 7 fields from the input PersonName object, pattern C can be populated with only 3 fields from the input PersonName object.
+3. Pattern C is discarded, because it has the least number of populated name fields.
+4. Out of the remaining patterns A and B, pattern B wins, because it has only 3 unpopulated fields compared to pattern A.
+
+If the “winning” namePattern still has fields that are unpopulated in the PersonName object, we alter the pattern algorithmically as follows:
+
+1. For each empty field from the start of the pattern, that field and all whitespace and literal text between it and the previous field and it and the next field is deleted. (This process stops with the first non-empty field.)
+2. For each empty field from the end of the pattern, that field and all whitespace and literal text between it and the next field and it and the previous field is deleted. (This process stops with the first non-empty field.)
```
richgillam, line 591: These two steps need to also mention that if the first field is empty, we also remove any whitespace and literal text at the beginning of the string, and that if the last field is empty, we also remove any whitespace and literal text at the end of the string.  It might be better to reword step 1 as follows:

>If one or more fields at the start of the pattern are empty, all fields, whitespace, and literal text before the first populated field are deleted.

...and to reword step 2 in the same way.

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+    F. Baz
+
+### 6.7 <a name="6-7-deriving-initials" href="#6-7-deriving-initials">Deriving initials</a>
+
+The following process is used to produce initials when they are not supplied by the PersonName object. Assuming the input example is “Mary Beth”:
+
+| Action              | Result |
+| ------------------- | ------ |
+| 1. Split into words | “Mary” and “Beth” |
+| 2. Each converted to initials | “M” and “B” |
```
richgillam, line 733: Do we need to spell out what "converted to initials" means?  It seems like it might be better to say "Take the first grapheme of each word."

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+        * [アルベルト・アインシュタイン](https://ja.wikipedia.org/wiki/%E3%82%A2%E3%83%AB%E3%83%99%E3%83%AB%E3%83%88%E3%83%BB%E3%82%A2%E3%82%A4%E3%83%B3%E3%82%B7%E3%83%A5%E3%82%BF%E3%82%A4%E3%83%B3) 
+        * [阿尔伯特·爱因斯坦](https://zh.wikipedia.org/wiki/%E9%98%BF%E5%B0%94%E4%BC%AF%E7%89%B9%C2%B7%E7%88%B1%E5%9B%A0%E6%96%AF%E5%9D%A6) 
+    2. **The foreign name is written in a different script.** In that case, the patterns from the **locale of the name** are used as is.
+        * [Albert Einstein](https://de.wikipedia.org/wiki/Albert_Einstein) 
+
+In both cases, the ordering may be changed according to the **Name Order for Locales** settings that each locale provides. If the PersonName object does not supply a locale for a name, then a default locale will be derived based on other information (such as the script of the characters in the name fields).
+
+**Known Issue:** The first version of the data doesn’t handle cases with `foreignSpaceReplacement `where the formatting locale doesn’t need spaces between words, but the name locale has the same ordering as the formatting locale. For example, consider where the formatting locale is Thai, and the name is in English, but transliterated into Thai.
```
richgillam, line 762: Do we want this in here?

docs/ldml/tr35-personNames.md
```diff
@@ -41,26 +41,924 @@ The LDML specification is divided into the following parts:
...
+`	<personName order="**givenFirst**" length="medium" usage="referring" formality="formal">`<br>
+`		<namePattern>{given} {given2} {surname}{suffix}</namePattern>`<br>
+`	</personName>`<br>
+`	. . .`<br>
+`	<personName order="**surnameFirst**" length="medium" usage="referring" formality="formal">`<br>
+`		<namePattern>{surname}{given2}{given}{suffix}</namePattern>`<br>
+`	</personName>`<br>
+`	. . .`<br>
+`</personNames>` | 
+
```
richgillam, line 791: Do we really want this as a table?  It looks kind of weird (at least in GitHub's Markdown viewer).  You might just want a header in boldface followed by the XML text inside a ```.  (Same thing with the other similar tables.)

richgillam (commented):
Looks great (and I know it's too late to give an actual approval)!  That said, I did have a number of nits I wanted to mention, even if we don't want to do them before publishing.

macchiati:
We can make changes after the beta (even immediately after!) I'll respond
separately on these issues.

docs/ldml/tr35-personNames.md
macchiati: Good catches.

I suggest: 

If there is no name locale available via the PersonName data interface, then an effective name locale is created based on the characters in the name and the formatting locale. 

* Find the predominant script for the name in the following way.
  * For each character in the given and surname, find the script(s) of the character using the Script_Extensions property.
    * For each of those scripts, increment a counter for that script, and record the position of the first character encountered in that script.
  * The predominant script is the script with the highest counter value. If there are multiple counters, take the one with the lowest position. If there is still more than one, take the script whose script code is alphabetically lowest (this step is only needed in edge cases).

* If the predominant script is the same as the script of the formatting locale, then the name locale becomes the formatting locale.

* Otherwise, find the likely locale for the predominant script as specified by the likely subtags. (This will add a language and region.) Use that locale as the name locale.

richgillam: The "alphabetically lowest" thing feels a little weird to me, but hopefully it'll never happen in practice.  Other than that, this looks great.

docs/ldml/tr35-personNames.md
macchiati: I suggest:

Replace
> 2. Otherwise, if the PersonName object has a `nameLocale` field, and there is a `nameOrderLocales` element:
>     1. Map the `nameLocale` field using the likelySubtags data, eg `zh` ⇒ `zh_Hans_CN`
>     2. Traverse the `nameOrderLocales` elements, and within each order attribute value to see if there is a match. 
By 

2. Otherwise use the nameOrderLocales elements to find the most best match for the full name locale, as follows.
  * For each locale L in the parent locale lookup chain* for the full name locale, do the following
    b. If there is a precise match among the givenFirst nameOrderLocales for L, then set the nameOrder to givenFirst, and stop.
    d. Otherwise if there is a precise match among the surnameFirst nameOrderLocales for the L, then set the nameOrder to givenFirst, and stop.
    e. Otherwise repeat (c) and (d) with "und" substituted for the language subtag in L.

* For example, here is a parent locale lookup chain: de_Latn_de ⇒ de_Latn ⇒ de_de ⇒ de ⇒ und

This process will always terminate, because there is always a `und` value in one of the two nameOrderLocales elements.

docs/ldml/tr35-personNames.md
macchiati: I tried to address this above. See what you tink.

docs/ldml/tr35-personNames.md
macchiati: Yes, we found that the GFM format is sometimes fussy. Might have to have a space around the = sign.

docs/ldml/tr35-personNames.md
macchiati: How about: are replaced with fields fetched from the PersonName Data Interface.

(We should search for 'record' and fix those instances to either be name object or PN data interface)

docs/ldml/tr35-personNames.md
macchiati: That is much cleaner!

docs/ldml/tr35-personNames.md
macchiati: Yes, better to say 
Fetch the first grapheme cluster of each word.

docs/ldml/tr35-personNames.md
macchiati: I'm conflicted. I sorta want to tell the reader that we need to do some more work on FSR.

docs/ldml/tr35-personNames.md
richgillam: >b. Otherwise if there is a precise match among the surnameFirst nameOrderLocales for the L, then set the nameOrder to givenFirst, and stop.

I think you mean "set the nameOrder to surnameFirst" here...

>c. Otherwise repeat (a) and (b) with "und" substituted for the language subtag in L.

That'll never match "und_XX" because the parent chain the first time through step a includes "und".  You might be able to get around this just by taking "und" out of your example parent chain (although you probably want to include some verbiage highlighting this case a little better).

Other than those things, this looks good to me.

docs/ldml/tr35-personNames.md
macchiati: It could be pulled out of a table. I think the biggest problem is that inside of ' ticks, the ** don't work.
`order="**givenFirst**"`
==>
`order="`**`givenFirst`**`"`

macchiati: But that's also kind ugly.

docs/ldml/tr35-personNames.md
macchiati: Ya, that is just to give a determinant result. We could have a footnote to that effect.

docs/ldml/tr35-personNames.md
macchiati: for #1, agreed.

You won't hit 'und' until the very end of the chain. The c step is on each iteration. That is:
```
for (L in chain) {
  for (L2 in L, L') {
    if there is a precise match for L2 in givenFirst locales, return givenFirst
    else if there is a precise match for L2 in surnameFirst locales, return surnameFirst
  }
}
```
(Where L' is obtained from L by replacing the language subtag by 'und'.)

macchiati: I agree that my first formulation was confusing. Suggestions for wording?

richgillam: Okay, yeah, I totally misread it.  I think you can clarify this by including an example at the end, maybe instead of the current parent-chain example, maybe after it:

>In other words, you'll check the `givenFirst` and `surnameFirst` resources for the following locales, in this order:
>
>de_Latin_DE, und_Latn_DE, de_Latn, und_Latn, de_DE, und_DE, de, und

Doing it this way biases the algorithm in favor of matching the country code (that is, `und_CN` will beat `zh` if you're searching for `zh_CN`), which I'm a little leery of, but that probably doesn't matter in practice because it wouldn't really make sense for somebody to have `zh` and `und_CN` (or whatever pair) have different name orders.

PR #2399, opened 2022-09-28
CLDR-16045 Remove Australia/India/Vietnam as keywords for non-globe emoji in some langs

pedberg-icu:
CLDR-16045

Remove (the equivalents of) Australia, India, Vietnam from annotations for non-globe emoji in selected languages (based on the region names we have in the language). This is just the preliminary manual change for CLDR 42 beta2, a more comprehensive removal using a tool will follow under a different ticket.

Note: besides the emoji we discussed (lotus blossom, kangaroo, boomerang) this was also relevant for some other emoji in some languages:
- sari had a keyword of India in some
- koala bear had a keyword of Australia in some
- om sign had a keyword of India in zh

### DIFF
--- a/common/annotations/bn.xml
+++ b/common/annotations/bn.xml
@@ -1819,7 +1819,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🦦" type="tts">উদ্বিড়াল</annotation>
 		<annotation cp="🦨">দুর্গন্ধ | স্কাংক</annotation>
 		<annotation cp="🦨" type="tts">স্কাংক</annotation>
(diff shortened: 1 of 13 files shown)


stenshamn:
I forgot to ask, do you want me to fix up the Scandinavian languages while we're at it? I could do it within this PR I guess?

pedberg-icu:
> I forgot to ask, do you want me to fix up the Scandinavian languages while we're at it? I could do it within this PR I guess?

Thanks. This PR was just to handle a few of the most visible or relevant cases for beta2. I think other languages (and country names) can wait for the tool Tom is putting together. The fixes will still be in CLDR 42, just not the beta2 tag.

stenshamn (approved):

macchiati:
I'm having second thoughts about Australia (and for the future, for
others). I'm not saying we should back this out (in view of the schedule),
but I don't think we should go further without a lot of thought. Will send
email.

PR #2401, opened 2022-09-29
*DON'T MERGE* Test program for country names in annotations

macchiati:
CLDR-_____


btangmu:
That looks amazing!

Possibly the values would still need to be reviewed by experts in the affected locales. There might be false positives due to homographs, for example. Also false negatives (instances not found) for inflected/adjective forms (like "French" versus "France")?

Maaajjssj:
> CLDR-_____
> 
> * [ ]  This PR completes the ticket.

[Uploading gc.today.txt…]()

PR #2404, opened 2022-09-29
CLDR-15034 kbd: add mt

srl295:
CLDR-15034

- MSA 100:2002 is available from <https://mccaa.org.mt/Section/Content?contentId=1242>

Some references mentioning this layout:
- <http://web.archive.org/web/20160414071246/https://www.mita.gov.mt/MediaCenter/Images/1_Fonts_Pic1.jpg>
- <https://www.farah.cl/Keyboardery/A-Visual-Comparison-of-Different-National-Layouts/#mt>

ALLOW_MULTI_COMMITS=yes
ALLOW_MANY_COMMITS=yes

PR #2405, opened 2022-09-29
CLDR-15034 kbd: name needs to be ORDERED

srl295:
CLDR-15034

hit this in #2404  - otherwise more than one `<name>` element is not allowed.

### DIFF
--- a/keyboards/3.0/fr-t-k0-azerty.xml
+++ b/keyboards/3.0/fr-t-k0-azerty.xml
@@ -10,7 +10,7 @@
 	<info author="Team Keyboard" normalization="NFC" layout="AZERTY" indicator="FR" />
 	<names>
 		<name value="French" />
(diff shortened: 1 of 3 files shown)


keyboards/3.0/fr-t-k0-azerty.xml
```diff
@@ -10,6 +10,7 @@
 	<info author="Team Keyboard" normalization="NFC" layout="AZERTY" indicator="FR" />
 	<names>
 		<name value="French" />
+		<name value="French AZERTY" /> <!-- a second name just to expose an issue -->
```
mcdurdin, line 13: Added the second name just to test and expose this issue.

```suggestion
		<name value="French AZERTY" />
```

The comment can stay in this PR -- it won't be useful for future readers of the code.

mcdurdin (approved):
LGTM

PR #2411, opened 2022-10-01
CLDR-15368 Enable API to replace WHAT_GETROW

btangmu:
-Enable Survey Tool to work correctly with TABLE_USES_NEW_API = true

-Keep TABLE_USES_NEW_API = false for now pending further testing

-Make VoteAPI produce json more like DataSection.getRow

-Supply numerous data items that were missing from VoteAPI json

-Remove duplicate VoteResolver in VoteAPI

-Distinguish front-end votingResults from back-end voteResolver

-Lazy initialization of PathHeader alphabetic collator to avoid CLDRConfig conflict

-A few front end changes, no need for displaySets.default; fix getSingleRowUrl; simplify testItem.cause

-Comments

CLDR-15368

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
@@ -376,15 +376,6 @@ public class PathHeader implements Comparable<PathHeader> {
         "\\[@alt=\"([^\"]*+)\"]")
         .matcher("");
 
-    static final Collator alphabetic = CLDRConfig.getInstance().getCollatorRoot();
-
-//    static final RuleBasedCollator alphabetic = (RuleBasedCollator) Collator
-//            .getInstance(ULocale.ENGLISH);
-//    static {
-//        alphabetic.setNumericCollation(true);
-//        alphabetic.freeze();
-//    }
-
     static final SupplementalDataInfo supplementalDataInfo = SupplementalDataInfo.getInstance();
     static final Map<String, String> metazoneToContinent = supplementalDataInfo
         .getMetazoneToContinentMap();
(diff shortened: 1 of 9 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -2116,11 +2107,24 @@ public static String getMetazonePageTerritory(String source) {
...
 
+    private static synchronized void initializeAlphabetic() {
+        // Lazy initialization: don't call CLDRConfig.getInstance() too early or we'll get
+        // "CLDRConfig.getInstance() was called prior to SurveyTool setup" when called from
+        // com.ibm.ws.microprofile.openapi.impl.core.jackson.ModelResolver._addEnumProps
+        if (alphabetic == null) {
+            alphabetic = CLDRConfig.getInstance().getCollatorRoot();
+        }
+    }
+
```
srl295, line 2127: make it a [Bill Pugh singleton](https://en.wikipedia.org/wiki/Singleton_pattern#Lazy_initialization) instead?

srl295 (approved):
+1 ! Lots of good work here, data wrangling and paying down tech debt especially 🎉

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
btangmu: you mean like
```
public final class SingletonBillPugh {
	private static class SingletonHelper {
		private static final SingletonBillPugh INSTANCE = new SingletonBillPugh();
	}

	private SingletonBillPugh() {
	}

	public static SingletonBillPugh getInstance() {
		return SingletonHelper.INSTANCE;
	}
}
```
?

btangmu: Let's discuss -- in the meantime I'll merge the PR as-is -- thanks for the review!

PR #2413, opened 2022-10-03
CLDR-15034 kbd: spec: NFC as  the preferred form

srl295:
CLDR-15034

Summary:   `<info normalization=…>` already has the possible values `NFC`, `NFD`, and `other`. 

This notes that `NFC` is required for CLDR's repository, and that tooling would validate for `NFC` or `NFD` values.

If `other` is provided (this is like the 'mixed' @miloush  mentioned) then tooling would not be expected to check the output of the keyboard.

ALLOW_MANY_COMMITS=true
ALLOW_MULTI_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -477,9 +477,10 @@ _Attribute:_ `author`
 _Attribute:_ `normalization`
 
 > The `normalization` attribute describes the intended normalization form of the keyboard layout output. The valid values are `NFC`, `NFD` or `other`.
+>
 > An example use case is aiding a user to choose among the two same layouts with one outputting characters in the normalization form C and one in the normalization form D.
 >
-> All keyboards in the CLDR repository will be in `NFC` form.  However, users and implementations may produce and consume other normalization forms.
+> All keyboards in the CLDR repository will be in `NFC` or `NFD` forms.  However, users and implementations may produce and consume other normalization forms or mixed output, use the `other` value to indicate this case.
 >
 > When using `NFC` or `NFD`, tooling should verify that all possible keystrokes, gestures, and transforms on the keyboard only produce the specified normalization form, producing warnings if not.
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -478,6 +478,10 @@ _Attribute:_ `normalization`
 
 > The `normalization` attribute describes the intended normalization form of the keyboard layout output. The valid values are `NFC`, `NFD` or `other`.
 > An example use case is aiding a user to choose among the two same layouts with one outputting characters in the normalization form C and one in the normalization form D.
+>
+> All keyboards in the CLDR repository will be in `NFC` form.  However, users and implementations may produce and consume other normalization forms.
+>
+> Tooling should verify that all possible keystrokes and gestures on the keyboard only produce the specified normalization form.
 
 _Attribute:_ `layout`
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -481,7 +481,7 @@ _Attribute:_ `normalization`
 >
 > All keyboards in the CLDR repository will be in `NFC` form.  However, users and implementations may produce and consume other normalization forms.
 >
-> Tooling should verify that all possible keystrokes and gestures on the keyboard only produce the specified normalization form.
+> When using `NFC` or `NFD`, tooling should verify that all possible keystrokes, gestures, and transforms on the keyboard only produce the specified normalization form, producing warnings if not.
 
 _Attribute:_ `layout`
 


mcdurdin (approved):

docs/ldml/tr35-keyboards.md
```diff
@@ -478,6 +478,10 @@ _Attribute:_ `normalization`
 
 > The `normalization` attribute describes the intended normalization form of the keyboard layout output. The valid values are `NFC`, `NFD` or `other`.
 > An example use case is aiding a user to choose among the two same layouts with one outputting characters in the normalization form C and one in the normalization form D.
+>
+> All keyboards in the CLDR repository will be in `NFC` form.  However, users and implementations may produce and consume other normalization forms.
+>
+> Tooling should verify that all possible keystrokes and gestures on the keyboard only produce the specified normalization form.
```
mcdurdin, line 484: Thinking about this a bit more, this is a little ambiguous. We can only really verify that individual strings are in the specified normalization form, not that all text produced is going to be in the correct form. For example:

```xml
<key id="a" to="a" />
<key id="mark" to="\u0300" />
```

Each of those outputs is by itself NFC. But press <kbd>a</kbd> and then <kbd>mark</kbd>, and that's of course no longer NFC. The implementation is supposed to handle normalization.

So the only thing I think it makes sense to validate here is:

1. all attribute values are normalized

srl295: > Thinking about this a bit more, this is a little ambiguous. We can only really verify that individual strings are in the specified normalization form, not that all text produced is going to be in the correct form. For example:

That's a good point.

> The implementation is supposed to handle normalization.

That needs to be documented, then. 

> 1. all attribute values are normalized

well… but if we have:

```xml
<key id="c" to="a\u0300" />
```

couldn't the implementation *also* handle the normalization? just as it would for a+mark.

But I think it would be better/cleaner to say that all attribute values are in the specified form.

mcdurdin: > couldn't the implementation _also_ handle the normalization?

Yes it could. But it is messy as you say, and unnecessary, and leads to potentially confusing data for users.

miloush: I believe we agreed that the declared normalization is only informative for UI purposes, i.e. we will not expect platforms to change the normalization forms used in the file. I would even argue that implementation should not change the normalization "on the fly" going against wishes of the layout author.

mcdurdin: @miloush, I have a vague memory of that discussion now. Here's what I found in the [minutes from 2020-08-25/26](https://docs.google.com/document/d/1ZtrHfr85vnAXxgD6pwyMUGGrOMRoHRxpIBGLArpaDdI/edit#heading=h.lh4or466oeyh):

> Spec proposal discussion: https://unicode-org.atlassian.net/browse/CLDR-13273
> * TR35 is worded to assume and prefer NFC
> * Tooling that converts LDML to platform implementations may prefer to convert NFC to NFD
> * TR 35 implementation note: When a sequence is consumed in the keyboard state machine that output is expected to be NFC for CLDR. Other implementations may choose to use LDML files with NFD form or non-conforming and mixed sequences
> * Add a metadata property to allow a keyboard author to specify the normalization style of the layout ( example:  normalization=NFC, normalization=NFD, normalization=mixed )
> * It is not required for a keyboard to anticipate and normalize the output from a sequence of LDML keystrokes 

You wrote:
> I would even argue that implementation should not change the normalization "on the fly" going against wishes of the layout author.

That matches the expectations of Richard Wordingham per the CLDR ticket above. What about normalization for input to transforms? If we don't do this, we force keyboards to deal with multiple forms of context strings, because input to transforms can be arbitrary data not generated by the keyboard, during edit sessions.

Also. I know that some apps already normalize input as they receive it, so that is outside of the control of the implementation.

srl295: I would rather say that implementations will normalize to the specified format before processing the transform. It would add a lot of burden and inconsistencies on the transform to have to handle all possible forms. That also implies that the transform input strings can be validated as well, no need for  both `é` and `e´` in input.

srl295: @miloush 

> …we will not expect platforms to change the normalization forms used in the file.

so if a keyboard is declared as `NFC` we wouldn't expect the platform to send it NFD as transform input, and we wouldn't expect the keyboard to output NFD. Correct?

>  I would even argue that implementation should not change the normalization "on the fly" going against wishes of the layout author

I think the implementation should stick to the form declared, yes.

in the example above, are you (Jan) suggesting is that authors of an NFC keyboard with:

```xml
<key id="a" to="a" />
<key id="mark" to="\u0300" />
```

… would need to then add a transform from `a\u0300` -> `à` otherwise the keyboard will output non-NFC? That seems like it would add more burden.

miloush: > in the example above, are you (Jan) suggesting is that authors of an NFC keyboard with

Yeah I am suggesting that is the case. We could revisit the informativeness of the attribute, the feature I care about here is to have the option to not have the normalization being messed with, which could for example be indicated by "mixed" value, or perhaps if we do this, even better by value like "preserve".

We didn't really discuss the processing of inputs to transforms I think, but isn't that covered by "When a sequence is consumed in the keyboard state machine that output is expected to be NFC for CLDR."?

srl295:
This now represents the meeting consensus from last time.

Question: Should this note that NFC is *preferred* unless NFD is needed? Or perhaps it's best to just leave it how it is.

mcdurdin (approved):

mcdurdin:
> Question: Should this note that NFC is _preferred_ unless NFD is needed? Or perhaps it's best to just leave it how it is.

It is probably helpful to encourage use of NFC where possible.

miloush:
> It is probably helpful to encourage use of NFC where possible.

Nothing against that, but I am curious why in general NFC is preferred in documents over not NFC?

mcdurdin:
> Nothing against that, but I am curious why in general NFC is preferred in documents over not NFC?

The [Unicode Normalization FAQ](https://unicode.org/faq/normalization.html#2) says:

> [NFC](https://www.unicode.org/glossary/#nfc) is the best form for general text, since it is more compatible with strings converted from legacy encodings.

I lean towards NFC because it's more popular -- majority of systems and content appear to be NFC, whether for historical reasons or otherwise:

Mark Davis's [blog](https://www.macchiato.com/unicode-intl-sw/nfc-faq), referenced from the above FAQ says:

> This really depends on the kind of text involved. For example, according to data at Google:
> * ~99.98% of web HTML page content characters are definitely NFC.
>   * Content means after discarding markup, and doing entity resolution.

(data from 2009)

srl295:
NFC is 1:1 always the same whereas NFD doesn’t participate in reordering as
was noted. So NFD can have multiple binary forms for the same visual text.
Not good for databases etc. sometimes necessary due to the issues
mentioned.

srl295:
N.B. https://unicode-org.atlassian.net/browse/CLDR-16943 is for discussing follow up normalization

PR #2432, opened 2022-10-06
CLDR-16068 fix spec so generated html validates; make personNames image smaller

pedberg-icu:
CLDR-16068

1. Fix spec markdown so generated html passes validation (problems with angle brackets).
2. Make the personNamesFormatModel.png image smaller (use the same 72-dpi resolution as other spec images)


docs/ldml/tr35-keyboards.md
```diff
@@ -280,13 +280,11 @@ This is the top level element. All other elements defined below are under this e
...
-> Parents: _none_
-> Children: [version](#Element_version), [~~generation~~](#Element_generation), [info](#Element_info), [names](#Element_names), [settings](#Element_settings), [import](#Element_import), [keyMap](#Element_KeyMap), [displayMap](#Element_DisplayMap), [layer](#Element_layer), [vkeys](#Element_vkeys), [transforms](#Element_transforms), [reorders](#Element_reorder), [backspaces](#Element_backspaces)
-> Occurence: required, single
->
-> </small>
+> <small>Parents: _none_</small>
+
+> <small>Children: [version](#Element_version), [~~generation~~](#Element_generation), [info](#Element_info), [names](#Element_names), [settings](#Element_settings), [import](#Element_import), [keyMap](#Element_KeyMap), [displayMap](#Element_DisplayMap), [layer](#Element_layer), [vkeys](#Element_vkeys), [transforms](#Element_transforms), [reorders](#Element_reorder), [backspaces](#Element_backspaces)</small>
+
+> <small>Occurence: required, single</small>
```
pedberg-icu, line 287: The old markdown generated a single <p> for the block quote, inside a <small> element, but html does not allow <p> inside <small>, only the other way around. Also the old markdown appeared to be intended to generate 3 different lines of block quote for the three different parts (Parents, Children, Occurrence), but that was not happening. So add the empty lines to generate the three separate <p> in the block quote, hopefully with the <small> inside each of the <p>s.

docs/ldml/tr35-keyboards.md
```diff
@@ -280,13 +280,9 @@ This is the top level element. All other elements defined below are under this e
...
-> <small>
->
-> Parents: _none_
-> Children: [version](#Element_version), [~~generation~~](#Element_generation), [info](#Element_info), [names](#Element_names), [settings](#Element_settings), [import](#Element_import), [keyMap](#Element_KeyMap), [displayMap](#Element_DisplayMap), [layer](#Element_layer), [vkeys](#Element_vkeys), [transforms](#Element_transforms), [reorders](#Element_reorder), [backspaces](#Element_backspaces)
-> Occurence: required, single
->
-> </small>
+> <small>Parents: _none_<br/>
+> Children: [version](#Element_version), [~~generation~~](#Element_generation), [info](#Element_info), [names](#Element_names), [settings](#Element_settings), [import](#Element_import), [keyMap](#Element_KeyMap), [displayMap](#Element_DisplayMap), [layer](#Element_layer), [vkeys](#Element_vkeys), [transforms](#Element_transforms), [reorders](#Element_reorder), [backspaces](#Element_backspaces)<br/>
+> Occurence: required, single</small>
```
pedberg-icu, line 285: The old markdown generated a single <p> for the block quote, inside a <small> element, but html does not allow <p> inside <small>, only the other way around. Also the old markdown appeared to be intended to generate 3 different lines of block quote for the three different parts (Parents, Children, Occurrence), but that was not happening. So add <br/> to generate the three lines, and move the <small> so it hopefully gets put inside the <p>.

docs/ldml/tr35-dates.md
```diff
@@ -67,7 +67,7 @@ The LDML specification is divided into the following parts:
       * 4.5.1.2 [Variable periods](#Variable_periods)
       * 4.5.1.3 [Parsing Day Periods](#Parsing_Day_Periods)
 * 5 [Time Zone Names](#Time_Zone_Names)
-  * Table: [<timeZoneNames> Elements Used for Fallback](#timeZoneNames_Elements_Used_for_Fallback)
+  * Table: [&lt;timeZoneNames&gt; Elements Used for Fallback](#timeZoneNames_Elements_Used_for_Fallback)
```
srl295, line 70: again was this a manual change? because this line is generated by code automatically.  if it's a bug in the processor we can deal with that, but otherwise this would be overwritten next time someone manually runs ToC and it also would not have an effect on the final HTML, which always gets a ToC generation automatically.


It might be better just to drop the `<>` from the heading.

docs/ldml/tr35-keyboards.md
srl295: this is the way I restructured it in the proposed update (`keyboard-preview` branch) but actually I generate this line automatically now.  I'm going to approve this, but i'd recommend not bothering with the keyboard section because it's just going to create merge conflicts later.

srl295: This spec is largely rewritten in the branch i mentioned.

srl295 (approved):
LGTM although I would not change the keyboard spec

docs/ldml/tr35-dates.md
pedberg-icu: I will do that, thanks.

pedberg-icu:
> LGTM although I would not change the keyboard spec

Thanks, I will back out the keyboard spec changes, and fix the timeZoneNames heading to not use angle brackets.

pedberg-icu:
@srl295 OK I addressed you comments, thanks, this is ready to re-review

srl295 (approved):

PR #2437, opened 2022-10-10
CLDR-15034 kbd: dtd: tighten up the DTD for directions

srl295:
CLDR-15034

The DTD should specify the flick direction regex

### DIFF
--- a/keyboards/dtd/ldmlKeyboard.dtd
+++ b/keyboards/dtd/ldmlKeyboard.dtd
@@ -132,7 +132,7 @@ Please see CLDR-15034 for the latest information. -->
 
 <!ELEMENT flick EMPTY >
 <!ATTLIST flick directions NMTOKENS #REQUIRED >
-    <!--@MATCH:regex/(n|e|s|w|ne|nw|se|sw|[ ]+)+-->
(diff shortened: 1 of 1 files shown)


keyboards/dtd/ldmlKeyboard.dtd
```diff
@@ -132,7 +132,7 @@ Please see CLDR-15034 for the latest information. -->
 
 <!ELEMENT flick EMPTY >
 <!ATTLIST flick directions NMTOKENS #REQUIRED >
-    <!--@MATCH:any-->
+    <!--@MATCH:regex/(n|e|s|w|ne|nw|se|sw|[ ]+)+-->
```
mcdurdin, line 135: This seems like it would allow `nn` or `nsew` or other arbitrary using the letters `nesw` sequences. I am also unsure on the `@MATCH:regex` comment, whether that is matching the whole string, or a partial string match?

srl295: it matches on the whole string.

mcdurdin:
```suggestion
    <!--@MATCH:regex/(n|e|s|w|ne|nw|se|sw)([ ]+(n|e|s|w|ne|nw|se|sw))*-->
```

This disallows prefix and suffix spaces but I think that is tidy anyway

PR #2440, opened 2022-10-11
CLDR-14310 fix charts for romance plurals

macchiati:
CLDR-14310


macchiati:
Modified chart is at https://github.com/unicode-org/cldr-staging/blob/8673bee1d3f7828b7461335d7a1e543dc1143ea4/docs/charts/42/supplemental/language_plural_rules.html

srl295 (approved):

PR #2443, opened 2022-10-11
CLDR-16032 json: update for atTime

srl295:
CLDR-16032


srl295:
This is a large diff but it shows the impact of this change

https://github.com/unicode-org/cldr-json/commit/c05e58021c85dd56ddd74cf1116275bf65b6f036

Parallel to `dateTimeFormats` is now a key `dateTimeFormats-atTime`

```json
            "dateTimeFormats-atTime": {
              "standard": {
                "full": "{1} 'at' {0}",
                "long": "{1} 'at' {0}",
                "medium": "{1}, {0}",
                "short": "{1}, {0}"
              }
            }
```

and, if any other types come in besides atTime they will be handled automatically.

Also note the new nested `standard` this is also for future proofing,  if there's a different `<pattern type=…>` in the future it will create a different subkey.

pedberg-icu (approved):

PR #2447, opened 2022-10-12
CLDR-16023 In DTD delta charts, allow significant alt values

macchiati:
CLDR-16023

Allows the significant alt values (but kept the suppression of others, to cut down on the noise).
Also added a TSV chart for the data, since that was simple to do and I was right in the code.

The changes in the charts are visible at https://github.com/unicode-org/cldr-staging/pull/63


pedberg-icu (approved):

macchiati:
I couldn't merge this (waiting on license/cla). Steven, can you check on what is hanging it up, and then merge?

srl295 (approved):

PR #2449, opened 2022-10-12
CLDR-16082 personName spec format fixes 2: EBNF table, */_ outside of quotes, …

pedberg-icu:
CLDR-16082

A few more formatting fixes:
- For the EBNF table use `<span style="white-space:nowrap">` to prevent wrapping (instead of `&nbsp;`)
- A few more moving of italic/bold styles outside of quotes
- For table cell alignment use `style="text-align:center"` instead of the obsolete `align="center"`
- Some spacing and positioning adjustments


richgillam (approved):
The GitHub diff view does weird things to that table, but I believe you when you say it looks better in the real context, and the other changes all look good.

PR #2462, opened 2022-10-17
CLDR-15349 Refactoring related to user levels

btangmu:
-Move userlevels request from AddUser.vue to cldrUserLevels.js

-Use session header, not query param s, for api/userlevels

-Do not forbid api/userlevels access since that info is public

-Comments

CLDR-15349

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/UserLevels.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/UserLevels.java
@@ -1,9 +1,6 @@
 package org.unicode.cldr.web.api;
 
-import javax.ws.rs.GET;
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/UserLevels.java
```diff
@@ -45,13 +42,13 @@ public class UserLevels {
                             + "}"))),
         })
     public Response getLevels(
-        @QueryParam("s") @Schema(required = true, description = "Session String") String sessionString) {
+        @HeaderParam(Auth.SESSION_HEADER) String sessionString) {
```
srl295, line 45: good modernization

srl295 (approved):

PR #2466, opened 2022-10-17
CLDR-16094 spec: fix header for numbers part

pedberg-icu:
CLDR-16094

The markdown header for art 3 numbers was missing the header/body separation row


pedberg-icu:
@AEApple I noticed this additional problem (often these do not become evident until we fix other problems and then commit). Hopefully this is the last one before tagging.

rickmcgowan (approved):

AEApple (approved):

PR #2475, opened 2022-10-20
CLDR-16098 merge remaining maint/maint-42 changes to main

pedberg-icu:
CLDR-16098

DISABLE_JIRA_ISSUE_MATCH=true
ALLOW_MANY_COMMITS=true

This PR merges the remaining maint/maint-42 changes after Oct 3 back to main (i.e. the changes not already included in the previous https://github.com/unicode-org/cldr/pull/2420 which merged maint/maint-42 changes through Oct 3 to main).

**Note: This requires a MERGE COMMIT to merge; this is normally not enabled, so will require temporarily enabling it by going to the github admin panel, enabling it, using that option for the merge, then disabling the option again.**


pedberg-icu:
> @srl295 - Is there a way we can update it so that we can have dashes in the commit message title without them failing for things like branch names?

@AEApple The commit message error was due to a leading space. I just fixed that (amended the commit message and force pushed)

AEApple:
Ah, I've had issues with dashes in the past unless I'm misremembering

srl295:
> @srl295 - Is there a way we can update it so that we can have dashes in the commit message title without them failing for things like branch names?

I think it's best to keep it more strict, because it simplifies parsing later.

srl295 (approved):
RSLGTM

PR #2478, opened 2022-10-21
CLDR-15368 After switch to API, remove dead code

btangmu:
-Remove dead code (WHAT_GETROW, WHAT_SUBMIT, ...)

-Fix some compiler warnings

CLDR-15368

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -10,17 +10,7 @@
 
 package org.unicode.cldr.web;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Hashtable;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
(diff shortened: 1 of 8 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
```diff
@@ -2129,7 +2083,7 @@ private void populateFromThisXpath(String xpath, Set<String> extraXpaths, CLDRFi
             checkCldr.getExamples(xpath, isExtraPath ? null : ourValue, examplesResult);
         }
         if (ourValue != null && ourValue.length() > 0) {
-            addOurValue(ourValue, row, checkCldrResult, sourceLocaleStatus, xpath, setInheritFrom);
+            addOurValue(ourValue, row, checkCldrResult, sourceLocaleStatus, xpath);
```
btangmu, line 2086: IntelliJ pointed out that setInheritFrom was always null here and so had no effect in addOurValue. It's true though not very obvious. setInheritFrom is null if ourValueIsInherited is false. So for setInheritFrom to be non-null, ourValueIsInherited must be true. But then we would have returned already if (ourValueIsInherited && !isExtraPath), so isExtraPath would need to be true for setInheritFrom to be non-null here. But if isExtraPath is true, then ourValue is null, so addOurValue wouldn't be called, QED

srl295 (approved):
💯

PR #2479, opened 2022-10-22
CLDR-14943 cldr-splaining: add VoteResolver function to get a transcript

srl295:
- verbose explaination of vote resolution
- ⚠️ refactoring of some critical functions

CLDR-14943

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
@@ -78,6 +78,37 @@ public class VoteResolver<T> {
     private static final boolean DEBUG = false;
 
     /**
+     * This enables a prose discussion of the voting process.
+     */
+    private StringBuilder transcript = null;
+
+    void enableTranscript() {
+        if (transcript == null) {
+            transcript = new StringBuilder();
+        }
+    }
+
+    String getTranscript() {
+        if (transcript == null) {
+            return null;
+        } else {
+            return transcript.toString();
+        }
+    }
+
+    /**
+     * Add an annotation
+     * @param fmt
+     * @param args
+     */
+    private final void annotateTranscript(String fmt, Object... args) {
+        if (transcript == null) {
+            return;
+        }
+        transcript.append(String.format(fmt, args)).append("\n");
+    }
+
+    /**
      * A placeholder for winningValue when it would otherwise be null.
      * It must match NO_WINNING_VALUE in the client JavaScript code.
      */
@@ -557,6 +588,9 @@ public class VoteResolver<T> {
             totalVotes.clear();
             baileyValue = null;
             baileySet = false;
+            if (transcript != null) {
+                transcript = new StringBuilder();
+            }
         }
 
         /**
@@ -642,6 +676,7 @@ public class VoteResolver<T> {
             }
             totals.clear();
 
+            annotateTranscript("- Getting all totals by organization:");
             for (Map.Entry<Organization, MaxCounter<T>> entry : orgToVotes.entrySet()) {
                 Counter<T> items = entry.getValue();
                 if (items.size() == 0) {
@@ -653,11 +688,11 @@ public class VoteResolver<T> {
                 if (weight == 0) {
                     continue;
                 }
+                annotateTranscript("-- Considering %s which has %d item(s)", entry.getKey().getDisplayName(), items.size());
                 Organization org = entry.getKey();
                 if (DEBUG) {
                     System.out.println("sortedKeys?? " + value + " " + org.getDisplayName());
                 }
-
                 // if there is more than one item, check that it is less
                 if (iterator.hasNext()) {
                     T value2 = iterator.next();
@@ -665,6 +700,7 @@ public class VoteResolver<T> {
                     // if the votes for #1 are not better than #2, we have a dispute
                     if (weight == weight2) {
                         if (conflictedOrganizations != null) {
+                            annotateTranscript("--- There are conflicts due to different values by users of this same organization");
                             conflictedOrganizations.add(org);
                         }
                     }
@@ -691,6 +727,10 @@ public class VoteResolver<T> {
                     if (count > maxCount) {
                         maxCount = count;
                         maxtime = time;
+                        // tell the 'losing' item
+                        if (considerItem != null) {
+                            annotateTranscript("---- Org is not voting for '%s': there is a higher ranked vote", considerItem);
+                        }
                         considerItem = item;
                         if (DEBUG) {
                             System.out.println("count>maxCount: " + considerItem.toString() + ":" + new Timestamp(considerTime).toString() + " COUNT: "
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1111,7 +1164,9 @@ private void resolveVotes() {
         winningStatus = computeStatus(weights[0], weights[1]);
 
         // if we are not as good as the baseline (trunk), use the baseline
+        // TODO: how could baselineStatus be null here??
```
btangmu, line 1167: maybe a newly-added (and/or "extra"???) path with no baseline value/status? or maybe it's impossible -- better safe than sorry though

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1399,24 +1470,46 @@ private long[] setBestNextAndSameVoteValues(Set<T> sortedValues, HashMap<T, Long
 
     /**
      * Compute the status for the winning value.
+     * See: https://cldr.unicode.org/index/process
      *
-     * @param weight1 the weight (vote count) for the best value
-     * @param weight2 the weight (vote count) for the next-best value
+     * @param O the weight (vote count) for the best value
+     * @param N the weight (vote count) for the next-best value
```
btangmu, line 1476: per the link (which is a great addition),

   The first value is the optimal value (O).
   The second value (if any) is the next best value (N).

maybe worth including that in the @param -- I couldn't remember what O and N were abbreviations for

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -77,6 +77,37 @@ public VoteResolver(VoterInfoList vil) {
 
     private static final boolean DEBUG = false;
 
+    /**
+     * This enables a prose discussion of the voting process.
+     */
+    private StringBuilder transcript = null;
```
btangmu, line 83: suggestion: next step, move this to a new separate class and file -- maybe "VoteResolverTranscript.java"

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -653,18 +688,19 @@ public Counter<T> getTotals(EnumSet<Organization> conflictedOrganizations) {
                 if (weight == 0) {
                     continue;
                 }
+                annotateTranscript("-- Considering %s which has %d item(s)", entry.getKey().getDisplayName(), items.size());
```
btangmu, line 691: suggestion (if you're not already planning on it): don't hard-code the initial hyphens or their count, instead each VoteResolverTranscript should track its indentation level, and the conversion from 2 into "-- " should be in its own method

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1098,11 +1148,14 @@ private void resolveVotes() {
          * Adjust sortedValues and voteCount as needed for annotation keywords.
          */
         if (isUsingKeywordAnnotationVoting()) {
+            annotateTranscript("TODO: Annotation stuff is happening, but we can’t explain it right now.");
```
btangmu, line 1151: ah, the geometric mean calculation! what joy to explain!!

btangmu (approved):
awesome!

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: i meant "VoteResolverSplainer.java"

btangmu: seriously though, it should be a separate file for max encapsulation

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: really the conversion from 2 into "-- " should happen on the front end -- the back end should provide structured json with separate count and message

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1379,12 +1434,28 @@ private long[] setBestNextAndSameVoteValues(Set<T> sortedValues, HashMap<T, Long
...
+                        if (weightArray[0] > weightArray[1]) {
+                            annotateTranscript("- This is the winning value because it has the highest weight (voting score).");
+                        } else if(winningValue.equals(baselineValue)) {
+                            annotateTranscript("- This is the winning value because it is the same as the baseline value, though the weight was otherwise equal to the next-best."); // aka blue star
+                        } else if(winningValue.equals(CldrUtility.INHERITANCE_MARKER)) {
+                            annotateTranscript("- This is the winning value because it is the inheritance marker, though the weight was otherwise equal to the next-best."); // triple up arrow
+                        } else {
+                            annotateTranscript("- This is the winning value because it comes earlier than '%s' when the text was sorted, though the weight was otherwise equal to the next-best.", nValue);
+                        }
+                        annotateTranscript("The Next-best (N) value is '%s', with weight %d", nValue, valueWeight);
```
btangmu, line 1458: these 12 new lines could all go in a subroutine

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: It was checked above

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: Actually I'd rather it was right here with the code that produces the calculation.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: This is somewhat of a tech preview. I'd like to improve how it's explained beyond just a prose text blob. So this is a start so that we can see that the right things are being explained. That's a good refactoring but let's wait to see how this shapes up.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: I'm not sure what O and N are abbreviations for, possibly Optimal and Nextbest?  but they correspond to the algorithm in the link.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: a subroutine next to votesThenUcaCollator

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: it's trying to explain to humans what getWinningValue actually does.

PR #2480, opened 2022-10-24
CLDR-14943 cldrsplain: refactor VoteResolver code, add API

srl295:
- for now, include transcript with each row

CLDR-14943

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.js
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.js
@@ -575,6 +575,26 @@ function updateRowVoteInfo(tr, theRow) {
       cldrDom.createChunk(msg, "p", "alert alert-warning fix-popover-help")
     );
   }
+  if (theRow.voteTranscript) {
+    const transcriptBox = cldrDom.createChunk(
+      "",
+      "div",
+      "transcript-container"
+    );
+    const transcriptText = cldrDom.createChunk(
+      theRow.voteTranscript,
+      "pre",
+      "transcript-text"
+    );
+    transcriptBox.appendChild(transcriptText);
+    const transcriptNote = cldrDom.createChunk(
+      cldrText.get("transcript_note"),
+      "p",
+      "alert alert-warning"
+    );
+    transcriptBox.appendChild(transcriptNote);
+    tr.voteDiv.appendChild(transcriptBox);
+  }
 }
 
 /**
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
@@ -81,13 +81,13 @@ public class VoteResolver<T> {
      */
     private StringBuilder transcript = null;
 
-    void enableTranscript() {
+    public void enableTranscript() {
         if (transcript == null) {
             transcript = new StringBuilder();
         }
     }
 
-    String getTranscript() {
+    public String getTranscript() {
         if (transcript == null) {
             return null;
         } else {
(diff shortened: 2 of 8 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1073,6 +1078,29 @@ public int compare(T o1, T o2) {
...
+            annotateTranscript("- This is the winning value because it has the highest weight (voting score).");
+        } else if(winningValue.equals(baselineValue)) {
+            annotateTranscript("- This is the winning value because it is the same as the baseline value, though the weight was otherwise equal to the next-best."); // aka blue star
+        } else if(winningValue.equals(CldrUtility.INHERITANCE_MARKER)) {
+            annotateTranscript("- This is the winning value because it is the inheritance marker, though the weight was otherwise equal to the next-best."); // triple up arrow
+        } else {
+            annotateTranscript("- This is the winning value because it comes earlier than '%s' when the text was sorted, though the weight was otherwise equal to the next-best.", nValue);
+        }
+        annotateTranscript("The Next-best (N) value is '%s', with weight %d", nValue, N);
+    }
```
btangmu, line 1102: minor quibble: better if subroutine is below its caller

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -929,6 +929,10 @@ public void clear() {
         oValue = null;
         winningValue = null;
         nValue = null;
+
+        if(transcript != null) {
```
btangmu, line 933: minor minor quibble: java needs formatting

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: in this case however it's intentionally adjacent to votesThenUcaCollator.compare

ideally the annotateTranscript calls might be inside votesThenUcaCollator.compare itself?

tools/cldr-apps/js/src/esm/cldrInfo.js
```diff
@@ -569,6 +569,26 @@ function updateRowVoteInfo(tr, theRow) {
...
+    );
+    transcriptBox.appendChild(transcriptText);
+    const transcriptNote = cldrDom.createChunk(
+      cldrText.get("transcript_note"),
+      "p",
+      "alert alert-warning"
+    );
+    transcriptBox.appendChild(transcriptNote);
+    tr.voteDiv.appendChild(transcriptBox);
+  }
```
btangmu, line 591: soon to be a vue component?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: Right.  
It can't be inside the collator because actually we don't annotate _every_ comparison, just the top two.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: done

tools/cldr-apps/js/src/esm/cldrInfo.js
srl295: but as i mentioned this is a preview…  there may be ways to visualize the data better, however, i want to make sure i'm getting the right data first.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: the collator could call a method that has a boolean to turn on annotation; boolean false when the collator calls it, boolean true when this method calls it just for the top two

btangmu (approved):

PR #2483, opened 2022-10-25
CLDR-13665 Move Survey Tool translation hints to the right Info Panel

btangmu:
-New TranslationHints.java

-New data/translation-hints/hints.txt

-Delete en_ZZ.xml

-Change references to en_ZZ to en

-In http response, include translationHint separately from displayName

-Fix bug where popovers failed in comparison cells with both circled i and circled e

-Show translations hints in Info Panel as well as in circled i popover in comparison cells

-Suppress useless trivial example popovers where example was identical to value; new getNonTrivialExampleHtml

-Remove dead code

-Fix some compiler warnings

CLDR-13665


tools/cldr-apps/js/src/esm/cldrDeferHelp.js
```diff
@@ -5,12 +5,29 @@ const defaultEndpoint = "https://dbpedia.org/sparql/";
 const format = "JSON";
 const abstractLang = "en";
 
-function addDeferredHelpTo(fragment, helpHtml, resource) {
+const USELESS_IN_INFO_PANEL = "see info panel";
```
srl295, line 8: haha! nice name

tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -754,41 +754,19 @@ function updateRowCodeCell(tr, theRow, cell) {
...
-        (theRow.displayExample
-          ? theRow.displayExample.replace(/\[translation hint.*?\]/g, "")
-          : "");
-      theRow.displayName = theRow.displayName.substr(0, hintPos);
-    }
     cell.appendChild(
       cldrDom.createChunk(theRow.displayName, "span", "subSpan")
     );
-    const TRANS_HINT_ID = "en_ZZ"; // must match SurveyMain.TRANS_HINT_ID
+    const TRANS_HINT_ID = "en"; // expected to match SurveyMain.TRANS_HINT_ID
```
srl295, line 760: probably ideally this should go back to being the 'baseline language' wherever that is, should be `en`… and delete this constant

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -226,41 +216,42 @@ public void setCacheOnly(boolean cacheOnly) {
...
+    private final BestMinimalPairSamples bestMinimalPairSamples;
 
-    private ExampleCache exCache = new ExampleCache();
+    private final ExampleCache exCache = new ExampleCache();
 
-    private ICUServiceBuilder icuServiceBuilder = new ICUServiceBuilder();
+    private final ICUServiceBuilder icuServiceBuilder = new ICUServiceBuilder();
 
-    private PluralInfo pluralInfo;
+    private final PluralInfo pluralInfo;
```
srl295, line 231: Just a note, this refactoring would be better off in a separate commit or even better a separate PR. Makes it harder to see the actual semantic changes.

srl295 (approved):
looks GREAT to me. will file separate ticket about english

tools/cldr-apps/js/src/esm/cldrTable.js
btangmu: I've always wondered what affect "en_ZZ" actually had on the browser in this context... and now I wonder whether "en" does anything; might the comparison column otherwise display right-to-left or something depending on the locale?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
btangmu: yes -- I'm trying to get into a habit of doing a preliminary clean-up pr for each ticket that's likely to modify files that need it, but I failed to do that in this case

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1218,25 +1224,24 @@ public enum UnitLength {
...
-                throw new IllegalArgumentException();
+            switch (listPatternType) {
+                case "unit":
+                    return UnitLength.LONG;
+                case "unit-narrow":
+                    return UnitLength.NARROW;
+                case "unit-short":
+                    return UnitLength.SHORT;
+                default:
+                    throw new IllegalArgumentException();
```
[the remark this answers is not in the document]
btangmu: yes, it shows up as a compiler warning though in IntelliJ

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/translation-hints/hints.txt
```diff
@@ -0,0 +1,264 @@
+HINT: "day" here means a time duration of 24 hours, not a calendar day
```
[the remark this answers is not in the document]
btangmu: good idea, I'll follow up with that -- I wasn't sure if .txt was going to be the long-term solution, but even if it turns out to somewhat temporary, it should be commented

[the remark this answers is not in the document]
btangmu: Is there actually a need for "(c)" or "©"? Maybe better to skip it unless it's somehow mandated

PR #2488, opened 2022-10-29
CLDR-16105 Inheritance not recognized by PerXPathData

btangmu:
-Normalize Bailey value to INHERITANCE_MARKER in STFactory

-New method STFactory.PerLocaleData.processValue, called by voteForValue and loadVoteValues

-For consistency/encapsulation, use DAIP for the normalization

-Fix one test in TestSTFactory.xml that assumed hard vote would persist when bailey changed

CLDR-16105

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -29,6 +29,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
 import org.unicode.cldr.test.CheckCLDR;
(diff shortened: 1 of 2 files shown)


srl295 (approved):
lgtm…  probably could note that the DAIP is constructed in a synchronized block

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -707,8 +710,11 @@ private void loadVoteValues(DataBackedSource targetXmlSource, VoteLoadingContext
                         continue;
                     }
                     try {
+                        value = processValue(xpath, value);
```
srl295, line 713: Value is already DAIP-processed, see CLDR-13180

PR #2490, opened 2022-10-31
CLDR-6396 Combine seed and common

btangmu:
-Move all files from seed to common

-Add README.md in seed and its subdirectories

CLDR-6396


btangmu:
this is only a draft pr for reference

compare https://github.com/unicode-org/cldr/pull/2491

btangmu:
closing this to avoid confusion with https://github.com/unicode-org/cldr/pull/2491

PR #2493, opened 2022-10-31
CLDR-16089 Remove periods from ordinal number formatters in languages that don't have this convention

grhoten:
CLDR-16089


allashashkina:
We should wait for Polish. Most likely, it will also needed to be removed. Waiting for a confirmation.

grhoten:
Polish confirmed that "1250. raz" is an acceptable way to represent ordinals in digit form.

pedberg-icu (approved):
Thanks!

btangmu:
@pedberg-icu is this supposed to be included in push to production today?

PR #2506, opened 2022-11-01
CLDR-10384 Organization names follow-up

btangmu:
-Remove ?? from Lao Posts/Telecom??

-Add Utilika before Utilka Foundation, the latter is suspected typo

CLDR-10384


btangmu:
There is a new ticket for further work on organization names: https://unicode-org.atlassian.net/browse/CLDR-16107

rickmcgowan (approved):
Looks OK to me.

macchiati:
Note: we should separately look at a way of deprecating Organizations that
have not participated in (say) the last 5 years (that is, we would leave
them around, but skip them in various lists).

PR #2517, opened 2022-11-03
CLDR-15816 Forum posts: follow-up with simplified threadId

btangmu:
-Change threadId to be simply the root post id

-Update unit tests

-Comments

CLDR-15816


srl295 (approved):
that's better

PR #2523, opened 2022-11-07
CLDR-16126 fix transliteration problems

macchiati:
CLDR-16126

Data files

- 4 of these are completely new. Three were produced for the most part by using the romanizations at https://www.eki.ee/wgrs/, with some extra characters that were missing from the romanizations but show up in the data with the script (such as punctuation characters).
- Most of these are tweaks to existing transliterators for missing characters. Where possible, the https://www.eki.ee/wgrs/ romanizations are used.

Code
- I had to do a serious rewrite of TestTransforms and CLDRTransforms. The old mechanism for figuring out dependencies was fragile, and difficult to maintain. So I read the files twice, once to get all the dependencies and figure out the proper order to register (which is very important).

Note: the TestUnicodeToZawgyi111107 changes are spurious, due to reformatting; the diff is having trouble then distinguishing changes from the additional test case. (github.dev does a better job; hit a . in the "Files Changed" view for that)


pedberg-icu (commented):
Wow, a lot of nice additions here, there have been some other tickets about the missing Interindic things. Some unit tests are failing though...

macchiati:
Yup, not done yet. Ran into a problem with order of initialization, so
working on it.

macchiati:
Just figured out final problem with ordering that was causing a test failure.

pedberg-icu (approved):

PR #2529, opened 2022-11-09
CLDR-11228 ST: notify TC of forum responses; notify

btangmu:
-New method SurveyForum.gatherUsersInterestedInThread

-Rename gatherInterestedUsers to gatherUsersInterestedInLocale

-Parameter PostInfo in emailNotify, used for gatherUsersInterestedInThread, also reduces parameters

-Simplify and speed up by replacing getFirstPosterInThread with getUserId(postInfo.getRoot())

-Use logger.severe instead of deprecated SurveyLog.errln

-Fix bug in cldrAccount.js, needed return in getInterestLocalesUrl

-Fix bug in cldrForumPanel.js, in exceptional case content is not const

-Comments

CLDR-11228

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
@@ -172,7 +172,7 @@ public class SurveyForum {
         return num;
     }
 
-    private void gatherInterestedUsers(String forum, Set<Integer> cc_emails, Set<Integer> bcc_emails) {
+    private void gatherUsersInterestedInLocale(String forum, Set<Integer> cc_emails, Set<Integer> bcc_emails) {
         try {
             Connection conn = null;
             PreparedStatement pIntUsers = null;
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
```diff
@@ -213,27 +213,53 @@ private void gatherInterestedUsers(String forum, Set<Integer> cc_emails, Set<Int
...
+     * Add users interested in this particular thread
+     *
+     * If this is a reply, add the user who started the thread, if they are not the current poster,
+     * and they are a TC member
+     *
+     * @param postInfo the new post
+     * @param currentUser the poster of the new post
+     * @param cc_emails the set of emails to which we may add
+     */
+    private void gatherUsersInterestedInThread(PostInfo postInfo, User currentUser, Set<Integer> cc_emails) {
```
btangmu, line 251: this is the essential new method

PR #2532, opened 2022-11-09
CLDR-16141 use multiplication sign instead of X

srl295:
CLDR-16141

- two spots, use ✕ instead of X

Followup to CLDR-13837


srl295:
See the ticket for before and after images

btangmu (approved):

btangmu:
@srl ticket still new? I was recently surprised still to see X not ✕

PR #2535, opened 2022-11-10
CLDR-15187 Encapsulate locale name constants for mis, mul, root, und, zxx

btangmu:
-New class LocaleNames has constants MIS, MUL, ROOT, UND, ZXX

CLDR-15187

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/AddUser.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/AddUser.java
@@ -10,6 +10,7 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+import org.unicode.cldr.util.LocaleNames;
 import org.unicode.cldr.util.LocaleNormalizer;
 import org.unicode.cldr.util.Organization;
 import org.unicode.cldr.web.CookieSession;
@@ -123,7 +124,7 @@ public class AddUser {
     private String getNewUserLocales(String requestLocales) {
         requestLocales = LocaleNormalizer.normalizeQuietly(requestLocales);
         if (requestLocales.isEmpty()) {
-            return "mul";
+            return LocaleNames.MUL;
         }
         return requestLocales;
     }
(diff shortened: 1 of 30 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/AddUser.java
```diff
@@ -123,7 +124,7 @@ private AddUserResponse doAddNewUser(AddUserRequest request, CookieSession sessi
     private String getNewUserLocales(String requestLocales) {
         requestLocales = LocaleNormalizer.normalizeQuietly(requestLocales);
         if (requestLocales.isEmpty()) {
-            return "mul";
+            return LocaleNames.MUL;
```
srl295, line 127: This is a bug I'm working on separately. Can you remove it from this ticket?

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/AddUser.java
srl295: I guess its OK for now

mul is also used for the sandbox (which is the meaning here) . That meaning should be ideally encapsulated separately

btangmu: maybe this is at least a step in that direction? Would they both still be named "mul"?

macchiati (approved):

btangmu:
This PR had conflicts with main earlier today, for TestPersonNameFormatter.java

At this moment "This branch has no conflicts with the base branch", following https://github.com/unicode-org/cldr/pull/2554

Nevertheless I don't want to create difficulties for the work-in-progress on person names so I'll postpone merging this -- and might step back and do this in stages, postponing changes to files that have work-in-progress for other tickets...

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/AddUser.java
srl295: yes, they are both named mul

srl295:
> This PR had conflicts with main earlier today, for TestPersonNameFormatter.java
> 
> At this moment "This branch has no conflicts with the base branch", following #2554
> 
> Nevertheless I don't want to create difficulties for the work-in-progress on person names so I'll postpone merging this -- and might step back and do this in stages, postponing changes to files that have work-in-progress for other tickets...

actually if you rebase it it should be fine to merge

srl295 (approved):

PR #2536, opened 2022-11-10
CLDR-10264 forum post check; run in separate thread

btangmu:
-Get separate db connection from sm.dbUtils

CLDR-10264

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForumCheck.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForumCheck.java
@@ -21,11 +21,11 @@ public class SurveyForumCheck {
 
     private int errorCount = 0;
 
-    public SurveyForumCheck(Connection conn) {
+    public SurveyForumCheck(SurveyMain sm) {
         this.logger = SurveyLog.forClass(SurveyForumCheck.class);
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForumCheck.java
```diff
@@ -21,11 +21,11 @@ public class SurveyForumCheck {
 
     private int errorCount = 0;
 
-    public SurveyForumCheck(Connection conn) {
+    public SurveyForumCheck(SurveyMain sm) {
         this.logger = SurveyLog.forClass(SurveyForumCheck.class);
         // In spite of setting Level.INFO here, the Appender may discard anything less than Level.WARNING
         logger.setLevel(java.util.logging.Level.INFO);
-        this.conn = conn;
+        this.conn = sm.dbUtils.getAConnection();
```
srl295, line 28: does it need to be cleaned up?

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForumCheck.java
btangmu: what do you mean by "cleaned up"?

PR #2539, opened 2022-11-11
CLDR-16135 Add test for displaying names

macchiati:
CLDR-16135

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRTransforms.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRTransforms.java
@@ -924,6 +924,9 @@ public class CLDRTransforms {
         case "Japn":
             id = "Jpan-Latn";
             break;
+        case "Kore":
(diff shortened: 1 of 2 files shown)


richgillam (approved):
Looks okay, I think, but I'm not completely clear on what you're doing here.  I guess you're taking the sample names for each locale and transliterating them back to Latin and formatting that result with the English name formatter?  What does that get you, and how are you using that?  Is this being used to generate a report, or does it show up in the Survey Tool, or something else?

macchiati:
The transliteration just helps us to see where the names are simply
transliterations of English, rather than real native names. I wasn't
thinking of having this in the ST (would be possible, but mostly of benefit
to us rather than to the vetters).

richgillam:
> The transliteration just helps us to see where the names are simply transliterations of English, rather than real native names. I wasn't thinking of having this in the ST (would be possible, but mostly of benefit to us rather than to the vetters).
> […](#)

Okay, that's good enough for me.

PR #2540, opened 2022-11-11
CLDR-4903 simpler display of exemplar sets

macchiati:
CLDR-4903

The code and tests are now complete. I'd like to merge this before doing the DAIP step.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
@@ -3,6 +3,22 @@
 
 package org.unicode.cldr.test;
 
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+import com.google.myanmartools.ZawgyiDetector;
+import com.ibm.icu.lang.UCharacter;
+import com.ibm.icu.text.Collator;
+import com.ibm.icu.text.DateIntervalInfo;
+import com.ibm.icu.text.DateTimePatternGenerator;
+import com.ibm.icu.text.DecimalFormat;
+import com.ibm.icu.text.Normalizer;
+import com.ibm.icu.text.RuleBasedCollator;
+import com.ibm.icu.text.Transform;
+import com.ibm.icu.text.Transliterator;
+import com.ibm.icu.text.UnicodeSet;
+import com.ibm.icu.text.UnicodeSetIterator;
+import com.ibm.icu.util.Output;
+import com.ibm.icu.util.ULocale;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


macchiati:
Results are in https://docs.google.com/spreadsheets/d/1mqabnPh9CN0DVEep2Uqgw4RVI7tTvt89jl4fFkcXnts/edit#gid=0 for comparison.

DavidLRowe:
> Results are in https://docs.google.com/spreadsheets/d/1mqabnPh9CN0DVEep2Uqgw4RVI7tTvt89jl4fFkcXnts/edit#gid=0 for comparison.

I looked at main exemplar for "bal".
- The "source" lists the characters in order by Unicode value. The "formatted" version lists them in apparently random order. I would think that would make it harder to visually determine what characters are included.
-  The "source" has a range U+0631 to U+0634. The "formatted" changes the range into four individual characters. (Three other ranges were preserved.)
- The hyphen replacement does make it easier to see that a range is indicated.

The listing for "bal_Arab" is a duplicate of "bal" (since bal_Arab has no exemplar data and is falling back to bal, I suppose). Is it needed?

For "az", "source" has "iIİ" since both dotted-i and dotless-i are used. The "formatted" line has "i I İ ı", though I wonder if that should be "i İ ı I" (or "ı I i İ") to match the case mapping used in Azerbaijani.

macchiati:
Converting this to draft. I'm going to change the ticket fixVersion to 44. It turns out that more substantial surgery needs to be done on DAIP than I anticipated.

srl295 (approved):
LG so far

macchiati:
@DavidLRowe I regenerated the data. Re your comments:

- The "formatted" version lists them in apparently random order. I would think that would make it harder to visually determine what characters are included.
I think the issue is that bal is RTL. The order in this test is the default root order.

- The listing for "bal_Arab" is a duplicate of "bal" (since bal_Arab has no exemplar data and is falling back to bal, I suppose). Is it needed?
This is just a verbose listing from a test, so no need to abbreviate

- For "az", "source" has "iIİ" since both dotted-i and dotless-i are used. The "formatted" line has "i I İ ı", though I wonder if that should be "i İ ı I" (or "ı I i İ") to match the case mapping used in Azerbaijani.
The order for this test is just the root order.

srl295 (approved):
Looks reasonable.

srl295:
(for the record) so working on CLDR-13180, something broke with this.  It looks like perhaps the encoding is happening _backwards_ when users vote.  I'm going to re-review this PR.

Example: voted for `c q d x` results in a vote in the database for `[{\[cdqx\]}]` (as if, `[cqdx]` but then escaped as a string!) 

Also `c q d x` is shown as a 'red' losing candidate — it doesn't match up with the winning `[{\[cdqx\]}]` 

image

https://cldr-smoke.unicode.org/cldr-apps/v#/agq/Alphabetic_Information/2703e9d07ab2ef3a

srl295:
@macchiati  see https://unicode-org.atlassian.net/browse/CLDR-13180 discussion

PR #2544, opened 2022-11-15
CLDR-15747 New API for getting alt attribute values

btangmu:
-New api/xpath/alt/hexId gets array of alt attribute values for specified xpath

CLDR-15747

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAlt.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAlt.java
@@ -1,0 +1,94 @@
+package org.unicode.cldr.web.api;
+
+import java.util.*;
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAlt.java
```diff
@@ -0,0 +1,94 @@
...
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
+import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+import org.unicode.cldr.icu.LDMLConstants;
+import org.unicode.cldr.util.*;
+import org.unicode.cldr.web.CookieSession;
+
+@Path("/alt")
```
srl295, line 19:
```suggestion
@Path("/xpath/alt")
```

srl295 (changes requested):
LGTM i wouldkeep the APIs together with others in xpath

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAlt.java
btangmu: should that be true of /api/whatis as well? --> /api/xpath/whatis (then also need to revise in LookUp.vue)

btangmu: for now I've just fixed it for alt

btangmu:
I've made a bunch of further changes locally and it will be better to make a new PR from scratch.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/XPathAlt.java
srl295: probably good to move whatis also.

PR #2548, opened 2022-11-15
CLDR-16000 en name of TR, make Türkiye default; update comments/descriptions to match

pedberg-icu:
CLDR-16000

In en, swap the names for TR to make Türkiye the primary name and Turkey the variant. Update en derived data (emoji flag name) and test data. Update English comments and descriptions in CLDR files. But do *not* update "Turkey" used as the internal name (key) or alias for a timezone/metazone.


AEApple:
What about all the other languages who provided an alt value in 42? Are we going to flip the values for the vetters or rely on them to flip them due to 'English changed' status?

macchiati (approved):

pedberg-icu:
> What about all the other languages who provided an alt value in 42? Are we going to flip the values for the vetters or rely on them to flip them due to 'English changed' status?

We discussed in TC 2022-11-16. We cannot be sure in other locales that have TR alt="variant" that it corresponds to the Turkey/Türkiye difference and/or that they are ready to use any "Türkiye" equivalent as the default in their locale.

PR #2550, opened 2022-11-16
CLDR-16135 Cleanup for sample names

macchiati:
CLDR-16135

This cleans up the last part of the previous PR, and also enables the script check.

There was a small change in CheckPlaceHolders to allow surname2 if either the surname or surname-core is present (since we treat those as equivalent if there is no surname-prefix).

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9454,16 +9454,16 @@ annotations.
 			<nameField type="suffix">M.D. Ph.D.</nameField>
 		</sampleName>
 		<sampleName item="foreignG">
-			<nameField type="given">Zendaya</nameField>
+			<nameField type="given">Sinbad</nameField>
 		</sampleName>
 		<sampleName item="foreignGS">
-			<nameField type="given">Irene</nameField>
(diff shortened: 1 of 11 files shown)


pedberg-icu (approved):
Looks good, thanks

common/main/en.xml
```diff
@@ -9454,16 +9454,16 @@ annotations.
...
-			<nameField type="surname">Adler</nameField>
+			<nameField type="given">Käthe</nameField>
+			<nameField type="surname">Müller</nameField>
 		</sampleName>
 		<sampleName item="foreignGGS">
-			<nameField type="given">John</nameField>
+			<nameField type="given">Zäzilia</nameField>
 			<nameField type="given2">Hamish</nameField>
-			<nameField type="surname">Watson</nameField>
+			<nameField type="surname">Stöber</nameField>
```
richgillam, line 9466: "Zäzilia Hamish Stöber"?  You sure you don't want to change "Hamish" too?

richgillam (commented):
I know you've already merged, but I looked at this and it LGTM.

PR #2553, opened 2022-11-16
CLDR-16137 Replace prefix and suffix

macchiati:
CLDR-16137

Changes prefix, suffix to title, generation, credentials. Some high points.

ldml.dtd - necessary changes

most locales - just changed prefix and suffix, but left everything alone. In particular, did not add {generation}

en.xml
- Added {generation} to very formal 
- I checked formats on the web, and typically there is a , before credentials if they are present, eg Dr. John Smith, MD.

ja.xml 
- suffix was clearly being used where title was appropriate, so replaced those.

ExampleGenerator.java
Added lines for Native and Foreign, so people could distinguish them.

TestPersonNameFormatter.java
Lots of "expected" values changed.

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3219,7 +3219,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ELEMENT nameField ( #PCDATA ) >
     <!--@TECHPREVIEW-->
 <!ATTLIST nameField type CDATA #REQUIRED >
-    <!--@MATCH:literal/prefix, given, given-informal, given2, surname, surname-prefix, surname-core, surname2, suffix-->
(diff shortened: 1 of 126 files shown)


richgillam (dismissed):
This all LGTM.

macchiati:
Thanks for the review. There was a test failure so I had to do a bit more work. Once the test finishes, I'd like another review. 

The change was caused by a failure in the attribute value comparators. Having been hit by this before, 

1. I changed them to be driven from the enum, and 
2. added a test that the ordering matches the enum and the dtd @MATCH literals

macchiati:
Now ready for re-review.

richgillam (approved):
Yeah, I like that.

PR #2556, opened 2022-11-17
CLDR-16150 slack notification on cldr-staging

srl295:
CLDR-16150

image


srl295:
looks like the standard github slack bot might be sufficient?

PR #2558, opened 2022-11-18
CLDR-16137 replace prefix and suffix

macchiati:
CLDR-16137

2nd Try
Took #2553, rebased on main, added (minor) changes to ldml.dtd, ran tests, and made new PR. 

There was one problem: the nativeSample name in root.xml had the old prefix/suffix identifiers instead of title/credentials. That causes the somewhat opaque error message:
> java.lang.IllegalArgumentException: Missing Map Comparator value(s): surname2(7),	prefix(null),
[The map comparators for attribute value comparison are set to fail when an item is not included, rather than silently accept it. However the error message should be clearer.]

----
Changes prefix, suffix to title, generation, credentials. Some high points.

ldml.dtd 
- necessary replacements
- also made fix in lines like . Even in 42 they are no longer sets; that was left over from the development version.

en.xml
- Added {generation} to very formal
- I checked formats on the web, and typically there is a , before credentials if they are present, eg Dr. John Smith, MD.
- Changed the foreign names to be truely foreign

ja.xml
- suffix was clearly being used where title was appropriate, so replaced those.

most locales 
- just changed prefix and suffix, but left everything alone. In particular, did not add {generation}

ExampleGenerator.java
- Added lines for Native and Foreign, so people could distinguish them.

TestPersonNameFormatter.java
- Lots of "expected" values changed.

For review, go to Files changed, and hit ".". (Regular diff fails due to number/size of files.

ALLOW_MULTI_COMMITS=yes
ALLOW_MANY_COMMITS=yes

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3219,7 +3219,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ELEMENT nameField ( #PCDATA ) >
     <!--@TECHPREVIEW-->
 <!ATTLIST nameField type CDATA #REQUIRED >
-    <!--@MATCH:literal/prefix, given, given-informal, given2, surname, surname-prefix, surname-core, surname2, suffix-->
+    <!--@MATCH:literal/title, given, given-informal, given2, surname, surname-prefix, surname-core, surname2, generation, credentials-->
 <!ATTLIST nameField alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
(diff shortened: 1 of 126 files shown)


macchiati:
This passes all the unit tests and checks, so I hope there isn't anything standing in the way. Can you please review early in your morning, and if ok: squash&merge, verify deployed on smoketest, then notify Fredrik (cc: cldr core) that it is ready for sanity checking.

srl295:
Reviewed as best i could with github.dev and also compared to a rebase from before the other PR's revert, to get a sense of the true delta  vs #2553 

LGTM

macchiati:
Thanks!

PR #2559, opened 2022-11-18
CLDR-14943 explain UI improvements

srl295:
- fix wordwrap in explainer
- hide explainer unless you click the circled 'i'
- update VoteResolver annotations to not prematurely declare a winner
- improve comments around annotations

This adds a top level function to cldrBundle due to difficulties in event handlers for the info panel.

CLDR-14943

### DIFF
--- a/tools/cldr-apps/js/src/index.js
+++ b/tools/cldr-apps/js/src/index.js
@@ -12,9 +12,20 @@ import "ant-design-vue/dist/antd.css";
 import { showPanel, createCldrApp } from "./cldrVueRouter.js";
 import { runGui } from "./runGui.js";
 
+/**
+ * TODO Does not belong here. CLDR-14193
+ * Workaround (aka hack) due to flattening in the current info panel.
+ */
+function toggleTranscript() {
+  document
+    .getElementsByClassName("transcript-container")[0]
+    .classList.toggle("visible");
+}
+
 // The following will show up in the cldrBundle global
 export default {
   createCldrApp,
   runGui,
   showPanel,
+  toggleTranscript,
 };
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
@@ -703,7 +703,7 @@ public class VoteResolver<T> {
                     // if the votes for #1 are not better than #2, we have a dispute
                     if (weight == weight2) {
                         if (conflictedOrganizations != null) {
-                            annotateTranscript("--- There are conflicts due to different values by users of this same organization");
+                            annotateTranscript("--- There are conflicts due to different values by users of this organization.");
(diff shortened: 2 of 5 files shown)


srl295:
the link to 'show' the transcript:

image

the little X to 'hide' the transcript, also fixed wrapping and wording for annotations.

image

tools/cldr-apps/js/src/index.js
```diff
@@ -12,9 +12,20 @@ import "ant-design-vue/dist/antd.css";
...
+    .getElementsByClassName("transcript-container")[0]
+    .classList.toggle("visible");
+}
+
 // The following will show up in the cldrBundle global
 export default {
   createCldrApp,
   runGui,
   showPanel,
+  toggleTranscript,
```
srl295, line 30: @btangmu  yes, this function seems out of place in our top level exported bundle.  However, it seemed like the simplest/expedient way to add a toggle  accessible via a simple `a href="javascript:cldrBundle.toggleTranscript()"` since all javascript/events gets stripped out of the info panel until it gets rewritten.

btangmu: okay temporarily -- ideally we shouldn't need to refer to cldrBundle from inside any module

to get started modernizing info panel, the following dom elements should just be created once, and then never recreated from scratch, but the individual divs can be shown/hidden/cleared/repopulated individually, so that updating one doesn't have to destroy the others:

	itemInfo
		alertDiv
		voteDiv
		sidewaysDiv
		forumDiv
		xpathDiv

something like that is needed for https://unicode-org.atlassian.net/browse/CLDR-14045, and it will pave the way towards including vue components inside any of those divs

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1454,7 +1462,7 @@ private long[] setBestNextAndSameVoteValues(Set<T> sortedValues, HashMap<T, Long
                 winningValue = value;
                 weightArray[0] = valueWeight;
                 valuesWithSameVotes.add(value);
-                annotateTranscript("The winning value (O) is '%s', with a weight of %d", winningValue, valueWeight);
+                annotateTranscript("The optimal value (O) is '%s', with a weight of %d", winningValue, valueWeight);
```
btangmu, line 1465: "optimal" goes nicely with "(O)" anyway

btangmu (dismissed):

srl295:
@btangmu  100% agree.

btangmu:
@srl295 wrong ticket? "keyboard: sort transforms by codepoint"

srl295:
> @srl295 wrong ticket? "keyboard: sort transforms by codepoint"
> 
> should be https://unicode-org.atlassian.net/browse/CLDR-14943

Good catch! fixed

btangmu (approved):

PR #2565, opened 2022-11-22
CLDR-15717 Open in default coverage after 10 hours inactivity

btangmu:
-Add millisSinceAction to status response on back end SurveyAjax.java

-New cldrCoverageReset.js on front end, uses notification from ant-design-vue

-Refactor where cldrSurvey calls cldrCoverageReset, move adjacent code to new subroutines

CLDR-15717


tools/cldr-apps/js/src/esm/cldrCoverageReset.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+import * as cldrCoverage from "../esm/cldrCoverage.js";
+import * as cldrGui from "../esm/cldrGui.js";
+import * as cldrLoad from "../esm/cldrLoad.js";
+import * as cldrMenu from "../esm/cldrMenu.js";
+import * as cldrText from "../esm/cldrText.js";
+
+import { notification } from "ant-design-vue";
+
+const TEST_DANGER = false; // false for production, true only when testing
```
srl295, line 12:
```suggestion
const TEST_DANGER = false && cldrStatus.getIsUnofficial(); // false for production, true only when testing
```

then even if it was set to `true` it would be false for production. 

https://unicode-org.atlassian.net/browse/CLDR-15239 is a better solution for such flags.

srl295 (dismissed):

tools/cldr-apps/js/src/esm/cldrCoverageReset.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+import * as cldrCoverage from "../esm/cldrCoverage.js";
+import * as cldrGui from "../esm/cldrGui.js";
+import * as cldrLoad from "../esm/cldrLoad.js";
+import * as cldrMenu from "../esm/cldrMenu.js";
+import * as cldrText from "../esm/cldrText.js";
+
+import { notification } from "ant-design-vue";
+
+const TEST_DANGER = false && cldrStatus.getIsUnofficial(); // false for production, true only when testing
```
btangmu, line 12: @srl295 I accepted your suggestion here -- not likely we would leave it true, still it's good to be cautious!

srl295 (approved):

PR #2566, opened 2022-11-22
CLDR-16161 hook PersonNames chart up to ST

srl295:
- minor refactors in Charts to prefer using CLDRURLS instead of the hard coded URLs
- Java side API done, JS todo

CLDR-16161


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonName.java
```diff
@@ -66,10 +65,10 @@ public String getFileName() {
 
     enum Filter {Main, Sorting, Monogram}
     enum Source {NativeSamples, ForeignSamples}
-
+ 
     @Override
-    public void writeContents(Writer pw) throws IOException {
-        CLDRFile cldrFile = FACTORY.make(locale, true);
+    public void writeContents(Writer pw, Factory factory) throws IOException {
```
srl295, line 70: @macchiati  this is now the abstract function to override for adding a chart.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/Chart.java
```diff
@@ -165,7 +184,19 @@ public static String getTsvDir(String targetDir, String topicName) {
...
+     * Attempt to allocate the Chart that goes along with this report
+     */
+    public static Chart forReport(final ReportId report, final String locale) {
+        switch(report) {
+            case personnames:
+                return new ChartPersonName(locale);
+            default:
+                return null;
+        }
+    }
```
srl295, line 201: @macchiati  maps a `ReportId` enum to a Chart object.  That should be all that is needed to add a new report, once the JS side is updated.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
```diff
@@ -298,4 +309,64 @@ public ReportUpdate() {
...
+                } else {
+                    switch (report) {
+                    // "Old Three" reports
+                    case compact:
+                    case datetime:
+                    case zones:
+                        try (final Writer w = new OutputStreamWriter(output);) {
+                            SurveyAjax.generateReport("r_" + report.name(), w, sm, loc);
+                        }
+                        break;
```
srl295, line 362: the right thing to do here is clearly to migrate these to become Charts…

srl295 (commented):
PTAL work in progress.

btangmu:
About the code scanning failure: maybe sanitize the locale id by conversion to CLDRLocale

tools/cldr-code/src/main/java/org/unicode/cldr/tool/Chart.java
macchiati: Very cool!

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonName.java
```diff
@@ -18,7 +18,6 @@
 
 public class ChartPersonName extends Chart {
     private static final CLDRConfig CLDR_CONFIG = CLDRConfig.getInstance();
-    static final Factory FACTORY = CLDR_CONFIG.getCldrFactory();
     static final CLDRFile ENGLISH = CLDR_CONFIG.getEnglish();
```
macchiati, line 21: Hmmm. if the FACTORY isn't right when coming for CLDR_CONFIG, then is the getEnglish()?

macchiati:
Looking good!

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonName.java
srl295: yes, because en.xml is static. getEnglish within ST also loads the disk file.

srl295:
All ready for review. And/or a cldr-staging visit!

srl295:
adding a new report consists of (and these are somewhat linked together):

1. creating the `Chart` subclass and implementing `writeContents(Writer pw, Factory factory)`
2. adding an enum to `VoterReportStatus.ReportId` referencing that chart such as `yourchart`
3. updating `Chart.forReport()` to construct said Chart given the enum value, and finally
4. updating `cldrText.js` with `special_r_yourchart` with the English name for the chart and title

macchiati:
Sounds good. Did you load it up on cldr-staging?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
macchiati: Yup

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonName.java
macchiati: ok. I think I need to go back through the other chart subclasses to make sure they are all kosher. Can do later (no point in generating charts when people can't see v43)

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonName.java
srl295: I updated the subtask description with instructions.

srl295:
I updated https://unicode-org.atlassian.net/browse/CLDR-16161 with how-to

srl295:
should be on smoketest soon

macchiati:
I don't see it. When I go to https://st.unicode.org/cldr-apps/v#/de// and
go to the nav panel and pick Reports, only the old 3 are there.

macchiati:
whoops, wrong link; will check smoketest

macchiati:
Bingo, that worked. However, the CSS is missing, so it jumbles the text
together. The file is at
https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/43/personNames/index.css
(the Chart software copies it from source code).

srl295:
The reports only load a html chunk into the middle of the page. Probably
need to include these into the overall css

PR #2567, opened 2022-11-22
CLDR-15826 Fix PersonName Check warnings and errors

macchiati:
CLDR-15826

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAccessor.java

- Needed access to the Phase() in the CheckAccessor.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java

- Name Order locales were being changed for exemplars, but those are just ASCII

- Dropped the old (ugly) REQUIRED and REQUIRED_EMPTY. The info is improved, and moved to SampleType. So we can just test with sampleType.getOptionality(modifiedField);
- The logic in handleCheck is improved. We are about whether the MISSING values are in the locale either explicitly or by explicit inheritance, but don't show errors if we don't yet have a value added to the locale.
- We emit warnings in BUILD phase (so we can check in), but errors in Submission, etc.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java

- We emit warnings in BUILD phase (so we can check in), but errors in Submission, etc.

- Here is where we build the set of items that are required vs optional into the SampleType, along with a getter (getOptionality(…)) that we can use in unittests and Checks
- Rebuild the test for TestCheckPersonNames to use sampleType.getOptionality.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAccessor.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAccessor.java
@@ -1,5 +1,7 @@
 package org.unicode.cldr.test;
 
+import org.unicode.cldr.test.CheckCLDR.Phase;
+
 /**
  * Special interface for testing; allows test cases to avoid setting up a fake CLDR file.
  */
@@ -8,4 +10,5 @@ public interface CheckAccessor {
     public String getUnresolvedStringValue(String path);
     public String getLocaleID();
     public CheckCLDR getCause();
+    public Phase getPhase();
 }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCoverage.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCoverage.java
@@ -9,10 +9,19 @@ package org.unicode.cldr.test;
 import java.util.List;
 
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
-import org.unicode.cldr.util.*;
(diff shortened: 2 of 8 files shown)


macchiati:
I know everyone is off for the holiday, but if you happen to have some spare time before Monday's meeting, I'd appreciate a review.

btangmu (dismissed):

macchiati:
Only changes in rebase were adjusting import statements.

btangmu:
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 4.56 s <<< FAILURE! - in org.unicode.cldr.unittest.web.TestShim
Error:  TestAll  Time elapsed: 4.559 s  <<< ERROR!
com.google.common.util.concurrent.UncheckedExecutionException: com.ibm.icu.util.ICUException: Couldn't make resolved CLDR file for en
	at org.unicode.cldr.unittest.web.TestShim.TestAll(TestShim.java:23)
Caused by: com.ibm.icu.util.ICUException: Couldn't make resolved CLDR file for en
	at org.unicode.cldr.unittest.web.TestShim.TestAll(TestShim.java:23)
Caused by: java.lang.NullPointerException
	at org.unicode.cldr.unittest.web.TestShim.TestAll(TestShim.java:23)

macchiati:
That turned out to be a nasty order of static initialization, with some clunky file loading. Works locally, so we'll see what happens on the server.

macchiati:
Had to squash - all tests now passing.

btangmu (approved):

PR #2569, opened 2022-11-23
CLDR-15187 Encapsulate locale name constants for mis, mul, root, und, zxx

btangmu:
-New class LocaleNames has constants MIS, MUL, ROOT, UND, ZXX

CLDR-15187


srl295 (changes requested):
LGTM. Compared to the other PR.

srl295 (approved):
Wrong button… LGTM

PR #2571, opened 2022-11-24
CLDR-16136 Tweak FSR conditions

macchiati:
CLDR-16136

Found that the logic didn't work for Thai. Changes:

tools/cldr-code/src/main/java/org/unicode/cldr/tool/LikelySubtags.java

- Slight reworking of the code to allow access to methods that directly manipulate a LanguageTagParser, which is more convenient to use for what we want.

- added the formatter language and script, so that "is a foreign language" is not dependent on anything but the likely language and script. This was to take care of an old TODO, where the old code just did a raw comparison, and would miss that en and en-UK are not "foreign languages" for the purpose of picking foreign space replacements.
- added a nativeSpaceReplacement. For the locales  that don't need spaces, this is set to "", otherwise it is " ".
  - NOTE we might want to make this settable in the future by vetters

- Added an test of Thai formatting, to verify that it handles the problem

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/LikelySubtags.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/LikelySubtags.java
@@ -125,7 +125,17 @@ public class LikelySubtags {
         return maximize(ltp);
     }
 
-    private String maximize(LanguageTagParser ltp) {
+    /** Maximize to a string (modifying the LanguageTagParser in so doing) */
+    public String maximize(LanguageTagParser ltp) {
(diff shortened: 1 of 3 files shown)


macchiati:
I know everyone is off for the holiday, but if you happen to have some spare time before Monday's meeting, I'd appreciate a review.

(It hasn't completed tests, so I'll check again in the morning.)

btangmu (approved):
https://github.com/unicode-org/cldr/pull/2567 changes some of the same files; I hope there won't be a merge conflict

macchiati:
Thanks! Iot looks like#2567 already had some conflicts with the latest main, so will fix there.

PR #2574, opened 2022-11-25
CLDR-16136 Allow empty FSR

macchiati:
CLDR-16136

I realized that there is no way for vetters to enter an empty FSR. So we need to handle it like we did zxx for the nameOrderLocales.

That is done here. 

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
@@ -576,4 +576,17 @@ public class TestDisplayAndInputProcessor extends TestFmwk {
         val = daip.processInput(xpath, "a  ।  b", null);
         assertEquals("U+0964 DEVANAGARI DANDA with spaces becomes pipe", normVal, val);
     }
+
+    private static final String FSR_START_PATH = "//ldml/personNames/foreignSpaceReplacement";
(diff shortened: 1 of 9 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
```diff
@@ -576,4 +576,17 @@ public void TestSplitBar() {
         val = daip.processInput(xpath, "a  ।  b", null);
         assertEquals("U+0964 DEVANAGARI DANDA with spaces becomes pipe", normVal, val);
     }
+
+    private static final String FSR_START_PATH = "//ldml/personNames/foreignSpaceReplacement";
+    private static final String EMPTY_ELEMENT_VALUE = "❮EMPTY❯";
```
btangmu, line 581: is this better than defining it once as public?

btangmu (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
macchiati: Will look at that in followup

macchiati: I can combine them.

PR #2575, opened 2022-11-27
CLDR-16136 First cut at enabling ST to have empty values

macchiati:
CLDR-16136

Found the code that was checking for empty values. Needed to add a method and infrastructure to test a path and an element.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java

- Using new code on line 445 to verify that when val is empty we only register a problem when DtdData.getValueConstraint(cp) == DtdData.Element.ValueConstraint.nonempty
- NOTE: I'm unable to compile cldr-apps locally, or even check syntax, so its a bit of guesswork

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
- unify the approach for zxx and ❮EMPTY❯
- add some constants at the top
- at 355, add test for either path
- at 506 qualify the replacement for empty.
- at 581 replace the ❮EMPTY❯ in locale lists (but also remove zxx for now)

- Replace EMPTY_VALUE_ALLOWED_IN_PCDATA by a map, that depends on the DtdType.
- Add ValueConstraint to indicate the status of an element (whether it allows empty values or not)
- Test with HACK_PCDATA_ALLOWS_EMPTY to set the ValueConstraint on an element.
- Add a static method getValueConstraint that works on a path

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
- Fix description that talked about zxx

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
- Refine test to cover both cases

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
- Add test

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java
- Refine test

- Add test

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
@@ -6,7 +6,6 @@ import java.util.logging.Logger;
 
 import javax.json.bind.spi.JsonbProvider;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.Response.Status;
 
 import org.unicode.cldr.test.CheckCLDR;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus;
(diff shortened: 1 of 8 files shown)


btangmu (approved):

srl295 (commented):
PMLGTM

PR #2576, opened 2022-11-28
CLDR-16169 Reset namePattern and nameField to draft="provisional"

macchiati:
CLDR-16169

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -9305,130 +9305,130 @@ annotations.
 		<initialPattern type="initial">{0}.</initialPattern>
 		<initialPattern type="initialSequence">{0} {1}</initialPattern>
 		<personName order="givenFirst" length="long" usage="referring" formality="formal">
-			<namePattern draft="provisional">{title} {given} {given2} {surname} {generation}, {credentials}</namePattern>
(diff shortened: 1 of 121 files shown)


common/main/en.xml
```diff
@@ -9305,130 +9305,130 @@ annotations.
 		<initialPattern type="initial">{0}.</initialPattern>
 		<initialPattern type="initialSequence">{0} {1}</initialPattern>
 		<personName order="givenFirst" length="long" usage="referring" formality="formal">
-			<namePattern>{title} {given} {given2} {surname} {generation}, {credentials}</namePattern>
+			<namePattern draft="provisional">{title} {given} {given2} {surname} {generation}, {credentials}</namePattern>
 		</personName>
```
pedberg-icu, line 9309: We don't want to reset _`en`_ to provisional, do we??

macchiati: Right, realized that I'd not excluded en.xml part-way through the meeting!

macchiati: Fixed that and a couple of other issues.

AEApple (approved):
LGTM, but I only spot checked

macchiati:
Thanks! With this many files, spot-checking is exactly the right approach. (I think the tests would catch anything serious.)

PR #2577, opened 2022-11-28
CLDR-16172 support empty string values in the FE

srl295:
- had to do some work arounds for abstention as well, also for JSON serialization

CLDR-16172

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.js
+++ b/tools/cldr-apps/js/src/esm/cldrTable.js
@@ -35,6 +35,11 @@ const CLDR_TABLE_DEBUG = false;
 const NO_WINNING_VALUE = "no-winning-value";
 
 /**
+ * Special input.value meaning an empty value as opposed to an abstention
+ */
+const EMPTY_ELEMENT_VALUE = "❮EMPTY❯";
+
+/**
  * Prepare rows to be inserted into the table
  *
  * @param theDiv the division (typically or always? with id='DynamicDataSection') that contains, or will contain, the table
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
```diff
@@ -858,7 +858,7 @@ public String getWinningVHash() {
         }
 
         public String getVoteVHash() {
-            String voteVhash = "";
+            String voteVhash = null;  // abstension
```
btangmu, line 861: comment -- spelling should be "abstention"

tools/cldr-apps/js/src/esm/cldrTable.js
```diff
@@ -419,7 +424,7 @@ function reallyUpdateRow(tr, theRow) {
   tr.rawValueToItem = {}; // hash:  string value to item (which has a div)
   for (let k in theRow.items) {
     const item = theRow.items[k];
-    if (item.value) {
+    if (item.value || item.value === "") {
```
btangmu, line 427: ideally encapsulate this as a boolean method -- maybe `cldrVote.hasValue(item.value)`  -- as it is the logic is duplicated half a dozen times

btangmu (approved):
I'm concerned these changes may have missed other places where the front or back end, or their http requests/responses, make inconsistent assumptions about null/empty/undefined values... For example, are Abstain and empty-string votes distinguished in the database, and for Import Old Votes?

srl295:
Abstain is already handled as null in the DB. Null vs empty string seems to transit fine.  I think that these changes are fairly contained.

PR #2578, opened 2022-11-28
CLDR-16136 The examples are not updating properly.

macchiati:
CLDR-16136

The example cache needs to have the dependencies updated. In particular:

The namePattern examples depend on:

1. nameSamples
2. initial patterns
3. nameOrderLocales
4. foreignSpaceReplacement

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleDependencies.java
- Update the general Example generators with changes to the above.

- There is also a special cache to keep from recreating the personname samples and formatter, so update that with changes to the above.

- There was also a bug in the examples, where personnames needs to know about the background/forground flags. It has to remove then re-add them in the case of generated initials, etc.

- Added test for the initials.


macchiati:
Ready for review

btangmu (approved):

srl295 (commented):
PMLGTM

PR #2579, opened 2022-11-29
CLDR-16174 Generate Births

macchiati:
CLDR-16174


AEApple (approved):
Is there anything else I should look out for on this since per Steven's comment on PR #2055 it is relying on tests.

Is there a way to spot check in the survey tool?

macchiati:
The tests all completed ok (You can see after the merge if you look at the Checks tab above). 
Good question about ST. There is a way, but we don't usually do it. For the record I'll go through it. (we might put this on #2 belwo)

1. https://www.google.com/search?q=cldr+generatebirth
2. https://cldr.unicode.org/development/cldr-development-site/updating-englishroot
3. The log output is on https://github.com/unicode-org/cldr-staging/tree/main/births/43.0
4. Pick your favorite locale, like https://github.com/unicode-org/cldr-staging/blob/main/births/43.0/nl.txt
5. Pick a path, like //ldml/localeDisplayNames/languages/language[@type="hi_Latn"]
6. Go to smoketest (or the cldr-staging server if doing beforehand - see Steven's guide) and look up the item.
7. Make sure that there is a warning on it.

PR #2583, opened 2022-11-30
CLDR-16181 TZ database 2022g updates, accurate Mexican exemplar city …

yumaoka:
…names for Mexican zones

CLDR-16181


macchiati (changes requested):
To allow Juarez to be voted on in v43, you will need to also modify the regex in:

SubmissionLocales.PATHS_ALLOWED_IN_LIMITED

yumaoka:
@macchiati 
> To allow Juarez to be voted on in v43, you will need to also modify the regex in:
> 
> SubmissionLocales.PATHS_ALLOWED_IN_LIMITED

I don't know whether my change done in the code is correct, because I don't know about this part much. Please confirm.

common/bcp47/timezone.xml
```diff
@@ -280,6 +280,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="mwblz" description="Blantyre, Malawi" alias="Africa/Blantyre"/>
             <type name="mxchi" description="Chihuahua, Mexico" alias="America/Chihuahua"/>
             <type name="mxcun" description="Cancún, Mexico" alias="America/Cancun"/>
+            <type name="mxcjs" description="Ciudad Juárez, Mexico" alias="America/Ciudad_Juarez"/>
```
markusicu, line 283:
```suggestion
            <type name="mxcjz" description="Ciudad Juárez, Mexico" alias="America/Ciudad_Juarez"/>
```
There is no 's' in the name of the city.

markusicu (commented):
one change please

FYI @gnrunge

common/bcp47/timezone.xml
yumaoka: @markusicu We use LOCODE by default - https://service.unece.org/trade/locode/mx.htm

MX  CJS | Ciudad Juárez | Ciudad Juarez | CHH | ---4---- | AI | 0212 |   | 3144N 10629W
-- | -- | -- | -- | -- | -- | -- | -- | --

So "mxcjs" is correct.

markusicu: ok tnx never mind... but then the LOCODE is wrong :-P

markusicu (approved):
lgtm

macchiati (approved):
Looks good. Please verify with smoke test (or cldr staging) that a vetter can vote on the item before closing the ticket.

PR #2592, opened 2022-12-02
CLDR-16137 Add missing generation field for English

macchiati:
CLDR-16137

The generation field was overlooked in English for foreignFull, which causes vetters to forget it for their languages.


richgillam (approved):
LGTM

PR #2596, opened 2022-12-02
CLDR-16181 Added 'since' attribute to the new zone Ciudad Juarez

yumaoka:
CLDR-_____


yumaoka:
I forgot to add `since` attribute to the new zone.

macchiati (approved):

macchiati:
Could you please also file a ticket to create a test that verifies that there is a "since" on every new item?

macchiati:
I assume this is ok to Merge now?

PR #2599, opened 2022-12-06
CLDR-16192 Admin email encoding failure for non-ASCII text

btangmu:
-Simply return parameter unchanged for WebContext.decodeFieldString

-Comments

CLDR-16192


btangmu:
Per today's infrastructure meeting, we'll postpone merging this until we've split a maintenance branch for production. Don't need this for the rest of v43 and there may be nonzero risk that some legacy code (as in JSPs) still relies on old behavior of decodeFieldString.

btangmu:
switched back to draft since we're waiting for maint. branch

srl295 (approved):

PR #2601, opened 2022-12-07
CLDR-16205 Swedish Person Names report; avoid StreamingOutput

btangmu:
-Get the complete report as a String before making a Response

CLDR-16205


btangmu:
@macchiati @pedberg-icu this PR does not completely fix the problem, but fixing this bug helps to reveal where a deeper bug is, in SimpleNameObject.java.

pedberg-icu (approved):

btangmu:
Just to clarify, the reason this is a bug --

`return Response.ok(new StreamingOutput() {...})`

-- is that in reality a failure, such as an exception, may occur during the streaming output, and when that happens, we're not telling the truth with Response.ok -- it's not OK

PR #2602, opened 2022-12-07
CLDR-16190 Don't have error values for Basic Locales

macchiati:
CLDR-16190

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
- Disallow the markers for empty string

- No longer check for hasRootParent

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
- Restructure a bit so that the constant paths with null root values are in a static set
- Add the sample names to that set
- This will keep the sample names showing up even if not in root
- Add pointer to TestPaths.extraPathAllowsNullValue for future additions

common/main/root.xml
 - Remove ∅∅∅ from the root values for sample names

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCLDRFile.java
- Add sampleNames to the PlaceholderStatus.DISALLOWED set

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java
- Add a workaround for test failures because of ^^^ in the current data.
- Filed ticket and added logKnowIssue for commenting out the workaround in Resolution.


btangmu:
Running tests with this code locally, TestVettingDataDriven fails on az.xml as it does here in the pr.

Running the server with this code locally, it seems to run fine, and there are no missing items in the bho dashboard at basic.

macchiati:
What I would like to have is a copy of the file that fails to be read.

btangmu:
> What I would like to have is a copy of the file that fails to be read

I've just produced that this morning, and I'll attach a copy to the ticket. In a nutshell, here's the problem:

```
		<sampleName item="nativeGGS">
			<nameField type="given" draft="provisional">Con</nameField>
			<nameField type="given2" draft="provisional">Hamiş</nameField>
			<nameField type="surname" draft="provisional">Vatson</nameField>
		</sampleName>
			<nameField type="surname-prefix" draft="provisional">van den</nameField>
			<nameField type="surname-core" draft="provisional">Volf</nameField>
			<nameField type="surname2" draft="provisional">Becker Schmidt</nameField>
			<nameField type="credentials" draft="provisional">M.D. Ph.D.</nameField>
		</sampleName>
	</personNames>
```

btangmu:
@macchiati locally, all tests pass with this change:
```
diff --git a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrXmlWriter.java b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrXmlWriter.java
index 62dfc4e655..ec29ba6d42 100644
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrXmlWriter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrXmlWriter.java
@@ -138,6 +138,9 @@ public class CldrXmlWriter {
                 continue;
             }
             String v = cldrFile.getStringValue(xpath);
+            if (v == null) {
+                continue;
+            }
             if (suppressInheritanceMarkers && CldrUtility.INHERITANCE_MARKER.equals(v)) {
                 continue;
             }
```
Details are in comments in the ticket itself https://unicode-org.atlassian.net/browse/CLDR-16190

btangmu (approved):

macchiati:
Looks like everything is working on Smoketest!

PR #2605, opened 2022-12-08
CLDR-16201 Add initial unitPreferenceData for rainfall-rate, snowfall-rate

pedberg-icu:
CLDR-16201

Add initial unitPreferenceData for rainfall-rate, snowfall-rate. Can refine if we get more input on this.

### DIFF
--- a/common/supplemental/units.xml
+++ b/common/supplemental/units.xml
@@ -443,6 +443,15 @@ For terms of use, see http://www.unicode.org/copyright.html
             <unitPreference regions="001">kilometer-per-hour</unitPreference>
             <unitPreference regions="GB US">mile-per-hour</unitPreference>
         </unitPreferences>
+        <unitPreferences category="speed" usage="rainfall-rate">
+            <unitPreference regions="001">millimeter-per-hour</unitPreference>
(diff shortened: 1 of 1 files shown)


markusicu (approved):
looks plausible to me

with the usual caveat that when we write regions="... US ..." we should probably include most of the US territories/possessions (PR, VI, ...) but that's for another day

richgillam (approved):
Looks good to me...

common/supplemental/units.xml
```diff
@@ -443,6 +443,15 @@ For terms of use, see http://www.unicode.org/copyright.html
             <unitPreference regions="001">kilometer-per-hour</unitPreference>
             <unitPreference regions="GB US">mile-per-hour</unitPreference>
         </unitPreferences>
+        <unitPreferences category="speed" usage="rainfall-rate">
```
sffc, line 446: Does it need to be "rainfall-rate", or could it just be "rainfall"?

macchiati:
I agree. The usage parameters are already specific to the quantity (length,
area, etc) so we don't need additional qualifiers.

PR #2433, opened 2022-10-06
CLDR-15054 Drop hard inheritance

btangmu:
-Add option to enable DAIP to replace Bailey with INHERITANCE_MARKER

-Never replace if localeWhereFound is root or code-fallback

-Call daip.enableInheritanceReplacement for CLDRModify (-fp), WHAT_SUBMIT, VoteAPI

-Update numerous xml files due to running revised CLDRModify -fp

-VoteResolver and DataRow change all votes for Bailey to votes for INHERITANCE_MARKER

-Revise unit test testAlias, en.xml now has sideways inheritance for test path

-Revise unit test testBaileyVotes, Bailey can no longer win, always changes to INHERITANCE_MARKER

-Revise unit test TestExampleGeneratorConsistency per recent DAIP NNBSP change

-Remove unused message voteInfo_votesForSpecificValue in cldrText.js

-Divide large method daip.processInput with new subroutines

-Fix miscellaneous warnings by IntelliJ in DisplayAndInputProcessor.java

-Fix an unrelated info panel help link with new CLDRURLS.COMPOUND_UNITS_HELP

-Remove dead code

-Comments

CLDR-15054

- [ ] This PR completes the ticket.

PR #2391, opened 2022-09-27
CLDR-14481 Coalesce multiple interior spaces

btangmu:
-Include NNBSP in WHITESPACE_AND_NBSP_TO_NORMALIZE and related patterns

-New unit tests for NNBSP

-Comments

CLDR-14481

- [x] This PR completes the ticket.

PR #2609, opened 2022-12-08
CLDR-16210 Revert some changes involving NNBS U+202F

btangmu:
-Revert changes in eight xml files in common/main

-Revert changes in DAIP  and unit test

-Comments

CLDR-16210

### DIFF
--- a/common/main/brx.xml
+++ b/common/main/brx.xml
@@ -1231,7 +1231,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						<dateFormatItem id="H">HH</dateFormatItem>
 						<dateFormatItem id="hm">a h:mm</dateFormatItem>
 						<dateFormatItem id="Hm">HH:mm</dateFormatItem>
-						<dateFormatItem id="hms">a h:mm:ss</dateFormatItem>
+						<dateFormatItem id="hms">a h:mm:ss</dateFormatItem>
 						<dateFormatItem id="Hms">HH:mm:ss</dateFormatItem>
 						<dateFormatItem id="M">L</dateFormatItem>
 						<dateFormatItem id="Md">d/M</dateFormatItem>
--- a/common/main/cv.xml
+++ b/common/main/cv.xml
@@ -456,17 +456,17 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<dateFormats>
 					<dateFormatLength type="full">
 						<dateFormat>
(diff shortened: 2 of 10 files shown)


btangmu:
This effectively reverts all the changes (except comments) to DisplayAndInputProcessor.java that were made in https://github.com/unicode-org/cldr/pull/2391

It also effectively reverts all the 46 spacing changes (from U+202F to either U+0020 or U+00A0) that would have been made if CLDRModify -fp had been run after https://github.com/unicode-org/cldr/pull/2391

Those same 46 changes would have been made if CLDRModify -fp had been run just before https://github.com/unicode-org/cldr/pull/2433

macchiati (approved):

macchiati:
Ah, clicked too soon!

btangmu:
also skip the 4 failing unit test items for now

macchiati (approved):

btangmu:
there were 2 things done in https://github.com/unicode-org/cldr/pull/2391

* Coalesce multiple interior spaces
* Recognize U+202F as one kind of space in various patterns, including path-specific normalization related to wantsRegularSpace, wantsNBSP, wantsNNBSP

the second one turns out to have been problematic; as we discussed in email and slack, the strategy here is first to "Revert the changes in DAIP that changed spaces" without distinguishing the above two things; subsequently we can restore the beneficial changes...

macchiati:
What we need to do in #2 is distinguish the paths that need/allow an NNBSP
/ NSP (U+202F / U+2009) from those that didn't, so we don't mash some
characters. (And then of course there is the ideographic space)

I think we need some categories of paths, so that we can answer the
following questions for each set:

   1. What set of the 5 spaces is allowed?
   2. Which of the allowed ones is preferred (this might also vary by
   locale)? Might be an ordering like : Nonbreaking > breaking then full >
   narrow
   3. What do we do if we hit a disallowed space?
   4. What do we do if we have a sequence of multiple spaces?

Note: the answer for #4 I think we can derive from the first two, something
like the following (for [SPACE, NBSP, NNBSP, SPACE]

   1. If a specific space occurs later in the sequence, we remove it. Thus
   we only have unique cases, such as [SPACE, NBSP, NNBSP]
   2. If there are only disallowed ones, we do #3.
   3. If there is at least one allowed space, we remove all disallowed
   ones. Eg [NBSP, NNBSP]
   4. The only case left is multiple allowed ones. If there is a preferred
   one, pick it; otherwise take the first in codepoint order

macchiati:
Thanks Tom!

PR #2614, opened 2022-12-14
CLDR-16199 ckb_IR needs H time formats for calendars inherited from ckb

pedberg-icu:
CLDR-16199

Per supplemental data, IR prefers time cycle H.
So ckb_IR needs to provide standard time formats explicitly using H for all calendars inherited from ckb (which defaults to IQ and time cycle h).

### DIFF
--- a/common/main/ckb_IR.xml
+++ b/common/main/ckb_IR.xml
@@ -63,26 +63,26 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<timeFormats>
 					<timeFormatLength type="full">
 						<timeFormat>
-							<pattern>↑↑↑</pattern>
-							<datetimeSkeleton>↑↑↑</datetimeSkeleton>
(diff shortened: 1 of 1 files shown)


AEApple (commented):
One question though, is there a reason why fa doesn't also use HH since it also would be IR?

See https://github.com/unicode-org/cldr/blob/main/common/main/fa.xml#L2103

pedberg-icu:
> > One question though, is there a reason why fa doesn't also use HH since it also would be IR?
> 
> I don't know for a fact that ckb_IR should use H, but if I had to guess, I would guess it should also use H (like fa_IR) and not HH. Zero-leading numbers (in hours, dates, etc) is pretty rare in Iran (except for things like minutes and seconds of course).

Well, ckb_IR uses HH for Gregorian so I just copied that for Persian. The supplemental timeData does not specify formatting (HH vs H), only cycle (like H vs h), so that is the main thing to get right here.

PR #2617, opened 2022-12-15
CLDR-16239 Always change bailey to inheritance marker in VoteResolver

btangmu:
-New method setWinningValue ensures winningValue never gets bailey

-This makes VoteResolver more consistent

-Fix unit tests that failed to set bailey

CLDR-16239

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
@@ -937,7 +937,7 @@ public class VoteResolver<T> {
         // TODO: clear these out between reuse
         // Are there other values that should be cleared?
         oValue = null;
-        winningValue = null;
+        setWinningValue(null);
         nValue = null;
 
         if (transcript != null) {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1475,7 +1475,7 @@ private long[] setBestNextAndSameVoteValues(Set<T> sortedValues, HashMap<T, Long
             ++i;
             long valueWeight = voteCount.get(value);
             if (i == 0) {
-                winningValue = value;
+                setWinningValue(value);
```
btangmu, line 1478: this is the typical case when there are enough votes -- it was already the case that we would return inheritance marker here rather than bailey

when setWinningValue is called from here, changeBaileyToInheritance in setWinningValue should always be superfluous, since that's already done before vote resolution -- nevertheless we always use the setter for encapsulation

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1136,12 +1136,12 @@ private void resolveVotes() {
          */
         if (sortedValues.size() == 0) {
             if (baselineValue != null) {
-                winningValue = baselineValue;
+                setWinningValue(baselineValue);
```
btangmu, line 1139: this is the change that fixes the bug I was encountering

notice that it's a fallback for when there are no (unconflicted) votes

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
srl295: thanks for updating the transcript annotations!

srl295 (commented):
PMLGTM

PR #2618, opened 2022-12-15
CLDR-16239 Change some inheritance markers to specific values in en_001.xml and en_GB.xml

btangmu:
-Changes are limited to a few paths with short, duration-hour

-The values are derived by reverting to before drop-hard-inheritance, then running cldrmodify one locale at a time, parent before child

CLDR-16239


btangmu:
> Looks good, but this is just the first step, right? Only the English revisions.

That's right. Now I'm working on the next steps.

srl295 (commented):
post merge LGTM

PR #2562, opened 2022-11-22
CLDR-16135 Clean up examples for PersonNames

macchiati:
CLDR-16135

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
- Make the labels be italic
- Add section for foreign name & script

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPersonNameFormatter.java
- Adjust tests

- [ ] This PR completes the ticket.

PR #2620, opened 2022-12-16
CLDR-16246 update PR template to allow multi commits by default

srl295:
CLDR-16246

Many PRs are landed which fail the 'single commit check'.  It would be better to quell this error so that we can see when there are real errors.

Examples: 
#2616 #2575 #2578 #2562

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,3 +16,5 @@ license before the PR is accepted.
 - sign: https://cla-assistant.io/unicode-org/cldr
 - license: https://www.unicode.org/copyright.html#License
 -->
+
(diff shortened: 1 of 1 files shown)


macchiati (approved):

btangmu (approved):
OK, though I still prefer to squash "manually" to see the results before merging

srl295:
> OK, though I still prefer to squash "manually" to see the results before merging

it doesn't prevent that.

srl295:
@btangmu  in fact you can still remove the ALLOW line from a PR if you prefer,this just changes the default

btangmu:
that's good to know!

PR #2621, opened 2022-12-16
CLDR-16239 Un-drop dropped hard inheritance

btangmu:
-New option -fU for CLDRModify

-Change INHERITANCE_MARKER back to hard value of 2022-10-07

-Revert to hard values of 80029f1d4dc0309acaaf9e9f4d005675c75a7163

-They were dropped in 02fbf4b88bfa95f52850bca6659f0c756733759d or later

CLDR-16239

- [ ] This PR completes the ticket.

PR #2622, opened 2022-12-18
CLDR-16239 Partially un-drop hard inheritance for common/main

btangmu:
-Change INHERITANCE_MARKER back to hard value of 2022-10-07 prior to PR 2433

-Revert to hard values of 80029f1d4dc0309acaaf9e9f4d005675c75a7163; they were dropped in 02fbf4b88bfa95f52850bca6659f0c756733759d or later

-Revert TestAlias to 2022-10-07 prior to PR 2433; change other back to one

CLDR-16239


btangmu:
This is a slight improvement over https://github.com/unicode-org/cldr/pull/2621

The change to CLDRModify for producing the output (not included in the PR, possible one-time operation) is:

```
fixList.add('U', "Un-drop inheritance", new CLDRFilter() {
    final private String baseDir = "/Users/tbishop/Documents/WenlinDocs/Organizations/Unicode/CLDR_job/2022_10_07_pre/";
    final File[] list = new File[]{
        new File(baseDir + "common/main/"),
        new File(baseDir + "common/annotations/")
    };
    final Factory preFactory = SimpleFactory.make(list, ".*");
    CLDRFile preFile;

    @Override
    public void handleStart() {
        String localeID = cldrFileToFilter.getLocaleID();
        try {
            preFile = preFactory.make(localeID, false /* not resolved */);
        } catch (Exception e) {
            System.out.println("Skipping " + localeID + " due to " + e);
            e.printStackTrace();
            preFile = null;
        }
    }

    @Override
    public void handlePath(String xpath) {
        if (preFile == null) {
            return;
        }
        if (xpath.contains("personName")) {
            return;
        }
        synchronized(this) {
            String value = cldrFileToFilter.getStringValue(xpath);
            if (CldrUtility.INHERITANCE_MARKER.equals(value)) {
                String preValue = preFile.getStringValue(xpath);
                if (!CldrUtility.INHERITANCE_MARKER.equals(preValue)) {
                    String fullXPath = cldrFileToFilter.getFullXPath(xpath);
                    replace(fullXPath, fullXPath, preValue);
                }
            }
        }
    }
```

common/main/ajp.xml
```diff
@@ -1,39 +1,39 @@
-<?xml version="1.0" encoding="utf-8"?>
```
btangmu, line 1: something strange here

btangmu: for this file the only differences are "utf-8" versus "UTF-8", and whitespace

btangmu:
@pedberg-icu for the current ICU-integration, would this data possibly work as-is?

the plan is to start changing some of these hard values back to ^^^, parent locales before child locales, and see what happens -- specifically, do the resolved values change or not...

macchiati:
I'd like to confirm what you're proposing:

1. Revert the data that was changed from <explicit> to ^^^,  back to <explicit> to get us past the integration problems.
2. After we are ok, take a gradual approach to converting <explicit> to ^^^, checking that we don't convert something that causes problems.

Is that what you are saying?

btangmu:
@macchiati yes, that's what I'm suggesting. If this data is good for the ICU integration, then let's use it as-is for that purpose.

As soon as we start changing hard values to ^^^ again, it's going to be an experiment with some risk that ^^^ doesn't mean what we think it means. The plan seems good, to try dropping hard inheritance gradually, first with parent locales and then with child locales, and testing whether it changes the resolved values. It's still not entirely clear to me how best to implement that plan, and I wouldn't hazard a guess how it will turn out. I'm eager to get started. If that experiment isn't needed before the ICU integration, however, let's not needlessly delay the ICU integration.

btangmu:
I've just made this ready for review. We might want several reviews before actually merging it. I'd like to get it running on cldr-staging as another test. Important: the Survey Tool user interface is still expected to treat the "hard" values as though they were ^^^, and that's the best-case scenario. The worst-case scenario is that ST will have bugs due to not handling hard values correctly.

macchiati:
The good thing is that the data that we are collecting this round doesn't have any sideways inheritance. So we have some freedom to act, given that we can revert data that had hard values and went to ^^^. 

What I'd requested earlier was a test for comparing fully resolved files before and after a hard vs soft inheritance change. That would allow us to have confidence that the change was not causing problems.

I think a longer-term change would be to integrate the "immediate sideways inheritance" fully into the XML source, but we need to discuss that.

btangmu:
"a test for comparing fully resolved files before and after a hard vs soft inheritance change" -- I'll work on that next

Question: does sideways inheritance ever occur for annotations? If not, we can limit our attention to common/main as far as this issue is concerned.

macchiati:
No, it doesn't. Sideways is limited to certain attributes; alt, count,
grammar,...

btangmu:
@macchiati fyi since "modify -fp" includes inheritance substitution, it can't normalize the data in this pr without changing to ^^^ again; as we've seen, that's messes up the resolved values. However, locally I've made a variant "modify -fP" (capital P) that's like -fp but without inheritance substitution. (There's also new "-fI" that does only inheritance substitution.) Running -fP locally on the data in this PR results in the following changes. I don't understand why "A B C D ..." is changed to "A B T S ..."; is that change correct? If these changes are correct it would be good to include them soon to get back a valid, stable baseline common/main.

%aa	Replacing	-	«[A B C D E F G H I K L M N O Q R S T U W X Y]»	→	«[A B T S E C K X I D Q R F G O L M N U W H Y]»	//ldml/characters/exemplarCharacters[@type="index"]
%aa	Replacing	-	«[a b c d e f g h i k l m n o q r s t u w x y]»	→	«[a b t s e c k x i d q r f g o l m n u w h y]»	//ldml/characters/exemplarCharacters
%ajp	Replacing	-	«[\u064B \u064C \u064D \u064E \u064F \u0650 \u0651 \u0652 \u0670 ء أ ؤ إ ئ ا آ ب ة ت ث ج ح خ د ذ ر ز س ش ص ض ط ظ ع غ ف ق ك ل م ن ه و ى ي]»	→	«[\u064B \u064C \u064D \u064E \u064F \u0650 \u0651 \u0652 \u0670 ء آ أ ؤ إ ئ ا ب ة ت ث ج ح خ د ذ ر ز س ش ص ض ط ظ ع غ ف ق ك ل م ن ه و ى ي]»	//ldml/characters/exemplarCharacters
%ajp	Replacing	-	«[\- ‐ ‑ – — ، ؛ \: ! ؟ . … ' " \& « » ( ) \[ \]]»	→	«[\- ‐ ‑ – — ، ؛ \: ! ؟ . … ' " « » ( ) \[ \] \&]»	//ldml/characters/exemplarCharacters[@type="punctuation"]
%bal	Replacing	-	«[آ ا {ائی} {ای} {اێ} ب پ ت ٹ ج چ د ڈ ر ز ڑ ژ س ش ف ک گ ل م ن ه و ی]»	→	«[آ ا {ای} {اێ} {ائی} ب پ ت ٹ ج چ د ڈ ر ڑ ز ژ س ش ف ک گ ل م ن و ه ی]»	//ldml/characters/exemplarCharacters[@type="index"][@draft="contributed"]
%bal	Replacing	-	«[\u064E \u064F \u0650 \u0651 \u0652 آ ئ ا ب پ ت ٹ ج چ د ڈ ر ز ڑ ژ س ش ک گ ل م ن ه و ۆ ی ێ ے]»	→	«[\u064E \u064F \u0650 \u0651 \u0652 آ ا ئ ب پ ت ٹ ج چ د ڈ ر ڑ ز ژ س ش ک گ ل م ن و ۆ ه ی ێ ے]»	//ldml/characters/exemplarCharacters[@draft="contributed"]
%bal_Latn	Replacing	-	«[A Á B {Ch} D {Dh} E É F G H I J K L M N O Ó P R {Rh} S {Sh} T {Th} U V W Y Z {Zh}]»	→	«[Á A B {Ch} D {Dh} É E F G H I J K L M N Ó O P R {Rh} S {Sh} T {Th} U V W Y Z {Zh}]»	//ldml/characters/exemplarCharacters[@type="index"][@draft="provisional"]
%bal_Latn	Replacing	-	«[a á b c d e é g h i j k l m n o ó p r s t u v w y z]»	→	«[á a b c d é e g h i j k l m n ó o p r s t u v w y z]»	//ldml/characters/exemplarCharacters[@draft="provisional"]
%cu	Replacing	-	«[\u0487\uA67D \u0483 ҂ а б \u2DE0 в \u2DE1 г \u2DE2 д \u2DE3 е є ж \u2DE4 з \u2DE5 ꙁ ѕ и і ї й к \u2DE6 л \u2DE7 м \u2DE8 н \u2DE9 о \u2DEA п р \u2DEC с \u2DED т у ꙋ ф х \u2DEF ѡ ѿ ꙍ ѽ ѻ ц ч \u2DF1 ш щ ⸯ ꙿ ъ ы ь ѣ ю ꙗ ѧ ѫ ѯ ѱ ѳ \u2DF4 ѵ ѷ]»	→	«[\u0487\uA67D \u0483 ҂ а б \u2DE0 в \u2DE1 г \u2DE2 д \u2DE3 е є ж \u2DE4 \u2DE5 ѕ з ꙁ и й і ї к \u2DE6 л \u2DE7 м \u2DE8 н \u2DE9 ѻ о \u2DEA п р \u2DEC с \u2DED т у ꙋ ф х \u2DEF ѡ ѿ ꙍ ѽ ц ч \u2DF1 ш щ ⸯ ꙿ ъ ы ь ѣ ю ѫ ꙗ ѧ ѯ ѱ ѳ ѵ ѷ \u2DF4]»	//ldml/characters/exemplarCharacters
%cu	Replacing	-	«[А Б В Г Д Є Ж З Ѕ И І К Л М Н О П Р С Т Ꙋ Ф Х Ѡ Ѿ Ѻ Ц Ч Ш Щ Ъ Ы Ь Ѣ Ю Ꙗ Ѧ Ѫ Ѯ Ѱ Ѳ Ѵ]»	→	«[А Б В Г Д Є Ж Ѕ З И І К Л М Н Ѻ О П Р С Т Ꙋ Ф Х Ѡ Ѿ Ц Ч Ш Щ Ъ Ы Ь Ѣ Ю Ѫ Ꙗ Ѧ Ѯ Ѱ Ѳ Ѵ]»	//ldml/characters/exemplarCharacters[@type="index"]
%lmo	Replacing	-	«[- — , ; \: ! ? . … ' ’ " “ ” « » ( ) \[ \] \{ \} @ /]»	→	«[\- ‑ — , ; \: ! ? . … ' ’ " “ ” « » ( ) \[ \] \{ \} @ /]»	//ldml/characters/exemplarCharacters[@type="punctuation"]
%pap	Replacing	-	«[, ; \: ! ? . … ‘ ' ’ ′ ″ “ " ” ( ) \[ \] / @ \& # § \- %]»	→	«[\- ‑ , ; \: ! ? . … ' ‘ ’ " “ ” ( ) \[ \] § @ / \& # % ′ ″]»	//ldml/characters/exemplarCharacters[@type="punctuation"]
%pap	Replacing	-	«[a b c d e è f g h i j k l m n ñ o ò p q r s t u ü ù v w x y z]»	→	«[a b c d e è f g h i j k l m n ñ o ò p q r s t u ù ü v w x y z]»	//ldml/characters/exemplarCharacters
%pap	Replacing	-	«[á é í ó ú ]»	→	«[á é í ó ú]»	//ldml/characters/exemplarCharacters[@type="auxiliary"]
%rif	Replacing	-	«[\- ‐ ‑ – — , ; \: ! ? . … ` ' ‘ ’ " " “ ” « » ( ) \[ \] \{ \} § @ * / \& #]»	→	«[\- ‐ ‑ – — , ; \: ! ? . … ' ‘ ’ " “ ” « » ( ) \[ \] \{ \} § @ * / \& # `]»	//ldml/characters/exemplarCharacters[@type="punctuation"]
%ssy	Replacing	-	«[A B C D E F G H I K L M N O Q R S T U W X Y]»	→	«[A B T S E C K X I D Q R F G O L M N U W H Y]»	//ldml/characters/exemplarCharacters[@type="index"]
%ssy	Replacing	-	«[a b c d e f g h i k l m n o q r s t u w x y]»	→	«[a b t s e c k x i d q r f g o l m n u w h y]»	//ldml/characters/exemplarCharacters

macchiati:
Those appear to be reorderings within the UnicodeSets. It appears that they
haven't had a basic pass applied. There is a utility to check if 2
UnicodeSets are identical, eg

https://util.unicode.org/UnicodeJsps/unicodeset.jsp?a=[%C3%81%20A%20B%20{Ch}%20D%20{Dh}%20%C3%89%20E%20F%20G%20H%20I%20J%20K%20L%20M%20N%20%C3%93%20O%20P%20R%20{Rh}%20S%20{Sh}%20T%20{Th}%20U%20V%20W%20Y%20Z%20{Zh}]&b=[%C3%81%20A%20B%20{Ch}%20D%20{Dh}%20%C3%89%20E%20F%20G%20H%20I%20J%20K%20L%20M%20N%20%C3%93%20O%20P%20R%20{Rh}%20S%20{Sh}%20T%20{Th}%20U%20V%20W%20Y%20Z%20{Zh}]

DavidLRowe:
The reordering to A B T S is consistent with common/collation/aa.xml

btangmu:
"...haven't had a basic pass applied..." -- "basic" meaning "modify" without "-fp"? -- I'll try that, thanks!

srl295:
> This is a slight improvement over #2621
> 
> The change to CLDRModify for producing the output (not included in the PR, possible one-time operation) is:
> 
> `fixList.add('U', "Un-drop inheritance", new CLDRFilter() {`

Maybe worth including in case someone needs to rerun this in the future ?

srl295 (commented):
Spot check data LGTM

I know this is in progress but… seems like if the code isn't reverted, the change will come back again automatically?

btangmu:
"a test for comparing fully resolved files before and after a hard vs soft inheritance change" -- as discussed in infrastructure meeting and emails, testing was done, as follows:

- New “modify -fI” does bailey -> ^^^ only
- Limit to locales that are exactly one step from root
- Produce resolved xml with “modify -z”
- Delete draft=”...” with perl
- Compare resolved xml from before and after "modify -fI"

Result: a huge number of differences in values (and draft status). Changing value in parent from hard to ^^^ definitely causes changed resolved value in child.

pedberg-icu:
@btangmu This looks good for integration, thanks!

pedberg-icu (approved):

PR #2624, opened 2022-12-20
CLDR-16224 st: updates to loading and retry

srl295:
- handleDisconnect no longer tries to load the 'loading' page.  That page only should show up if the ST is not up yet.
- on loading rows, process the error code properly and display the error, such as E_BAD_SECTION
- update the errCodeMsg box to have a Reload and a Home button

CLDR-16224

### on an incorrect section (i.e. bad URL) 

image

### On an incorrect locale (bad URL) 

image

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLoad.js
+++ b/tools/cldr-apps/js/src/esm/cldrLoad.js
@@ -273,6 +273,7 @@ function verifyJson(json, subkey) {
     );
     loadingChunk.appendChild(cldrDom.createChunk(msg_fmt));
     flipper.flipTo(pages.loading, loadingChunk);
+    loadingChunk.appendChild(cldrDom.createChunk("", "br"));
     var retryButton = cldrDom.createChunk(
       cldrText.get("loading_reload"),
       "button"
@@ -280,6 +281,14 @@ function verifyJson(json, subkey) {
     loadingChunk.appendChild(retryButton);
     retryButton.onclick = function () {
       window.location.reload(true);
+    };
+    var homeButton = cldrDom.createChunk(
+      cldrText.get("loading_home"),
+      "button"
+    );
+    loadingChunk.appendChild(homeButton);
+    homeButton.onclick = function () {
+      window.location.href = cldrStatus.getContextPath();
     };
     return false;
   } else if (json.err) {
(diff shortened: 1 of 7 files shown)


tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -779,15 +788,29 @@ function loadAllRows(itemLoadInfo, theDiv) {
 function loadAllRowsFromJson(json, theDiv) {
   isLoading = false;
   cldrSurvey.showLoader(cldrText.get("loading2"));
-  if (!json) {
+  if (!json || json.code) {
```
btangmu, line 791: what is json.code? does it imply failure to load?

I don't understand this -- if it's right, there should be an explanatory comment

btangmu (changes requested):
json.code?

tools/cldr-apps/js/src/esm/cldrLoad.js
srl295: Yes it's an error code from the payload. I can add a comment. 

I updated the loader so that rather than showing a generic error on 404 it comes here 

{ code: E_BAD_SECTION }

tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -871,14 +900,18 @@ function trimNull(x) {
...
-    console.log("Error: " + err);
-    notification.error({
-      message: `Could not fetch ${message}`,
-      description: `Error: ${err.toString()}`,
-      duration: 8,
-    });
-    handler(null);
+  const errorHandler = function (err, request) {
+    if (request.status === 404 && request.response?.code) {
+      handler(request.response); // pass through
```
srl295, line 905: @btangmu this is why the handler is called with `json.code` set in case of a 404

tools/cldr-apps/js/src/views/WaitingPanel.vue
```diff
@@ -1,9 +1,7 @@
 <template>
   <div id="home">
     <!-- if $specialPage == retry then we have reached this page 'directly' -->
-    <h1 v-if="$specialPage != 'retry'" class="hang">
-      Waiting for the SurveyTool to start up…
```
srl295, line 5: It's not always 'waiting to start up' at this line, so fixed the message

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPI.java
```diff
@@ -98,7 +98,9 @@ public Response getRow(
                 description = "Forbidden, no access to this data"),
             @APIResponse(
                 responseCode = "404",
-                description = "Row or Locale does not exist"),
+                description = "Row or Locale does not exist. More details in the 'code' field.",
+                content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = STError.class))),
```
srl295, line 103: an oversight in the API doc earlier.

tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -779,15 +788,35 @@ function loadAllRows(itemLoadInfo, theDiv) {
 function loadAllRowsFromJson(json, theDiv) {
   isLoading = false;
   cldrSurvey.showLoader(cldrText.get("loading2"));
-  if (!json) {
-    // Failed to load
+  if ( !json || json.code ) {
+    // Either no JSON was returned,
+    // or a STError was returned (json.code)
```
srl295, line 793: Added some comments , does this help ? @btangmu

tools/cldr-apps/js/src/esm/cldrLoad.js
srl295: I added a comment.

tools/cldr-apps/js/src/esm/cldrLoad.js
btangmu: so in this case "json" is not really the json returned by the server, but json = request.response -- ? that's confusing for me

maybe room for improvement in the protocol, or maybe I'm still not getting it?

btangmu (approved):

tools/cldr-apps/js/src/esm/cldrLoad.js
srl295: it is the json returned by the server.  request.response is from the server.

srl295: @btangmu  if you fetched `https://st.unicode.org/cldr-apps/api/voting/wxyz/page/BADPAGE` (but with the right surveytool-session header) you would get this 404 response:

```
{
  "code": "E_BAD_SECTION",
  "err": true,
  "message": "Error: E_BAD_SECTION"
}
```

btangmu: I think I get it now -- this code is `STError.code` filled in like this:

`return Response.status(404).entity(new STError(ErrorCode.E_BAD_SECTION)).build();`

That may seem obvious. Still, for easily-confused programmers like me, a comment relating `json.code` to `STError.code` would help, and even better would be to encapsulate that in cldrAjax -- something like `cldrAjax.getErrorCode(json)` which function could have a comment explaining the connection to STError on the back end.

It's not obvious in `if (!json || json.code)` that json.code is an error code -- after all, in general, a successful response could have `json.code`  meaning something completely different, right? For a while I was confusing it with `VoteAPI.RowResponse.Row.code`...

srl295: the surrounding code now refers to STError.

srl295: it could check `if (!json || json.err)` because err is always true in an STerror. maybe that would be clearer

btangmu: `myLoad` is called with many different handlers playing the role of `handler` -- will all of them grok the "pass through"?

srl295: I think this line is only for this load

srl295: OK, i didn't realize, `myLoad` is exported

> encapsulate that in cldrAjax

I think we should have less encapsulation here rather than more. It's already too many layers deep.

PR #2626, opened 2022-12-20
CLDR-16224 revert changes to myLoad, use fetch for loading rows

srl295:
CLDR-16224

- revert #2624 changes to `myLoad()` 
- update loading rows to use fetch instead

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLoad.js
+++ b/tools/cldr-apps/js/src/esm/cldrLoad.js
@@ -780,17 +780,26 @@ function loadAllRows(itemLoadInfo, theDiv) {
   );
   const url = cldrTable.getPageUrl(curLocale, curPage, curId);
   $("#nav-page").show(); // make top "Prev/Next" buttons visible while loading, cf. '#nav-page-footer' below
-  myLoad(url, "page", function (json) {
-    loadAllRowsFromJson(json, theDiv);
-  });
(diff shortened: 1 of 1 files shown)


srl295:
Now uses the `ferrbox` class for row errors, it looks a little more like an error and less like a headline.

image

tools/cldr-apps/js/src/esm/cldrLoad.js
```diff
@@ -780,17 +780,26 @@ function loadAllRows(itemLoadInfo, theDiv) {
...
   const url = cldrTable.getPageUrl(curLocale, curPage, curId);
   $("#nav-page").show(); // make top "Prev/Next" buttons visible while loading, cf. '#nav-page-footer' below
-  myLoad(url, "page", function (json) {
-    loadAllRowsFromJson(json, theDiv);
-  });
+  cldrAjax
+    .doFetch(url)
+    .then((response) => response.json())
+    .then((json) => loadAllRowsFromJson(json, theDiv))
+    .catch((err) => {
```
srl295, line 787: this will catch any error, even if the `loadAllRowsFromJson()` function throws.

btangmu: excellent modernization!

btangmu (approved):

PR #2627, opened 2022-12-21
CLDR-16254 update voting progress spreadsheet

srl295:
- add org coverage level
- add %

UI:
- disable button while generating

CLDR-16254

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
+++ b/tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
@@ -11,6 +11,7 @@ import * as cldrRetry from "./cldrRetry.js";
 import * as cldrStatus from "./cldrStatus.js";
 import * as cldrSurvey from "./cldrSurvey.js";
 import * as cldrText from "./cldrText.js";
+import * as cldrXlsx from "./cldrXlsx.js";
 import * as XLSX from "xlsx";
 
 let nf = null; // Intl.NumberFormat initialized later
@@ -114,10 +115,10 @@ async function downloadVettingParticipation(opts) {
       "Code",
       "Level",
       "Votes",
-      "CldrCovCount",
-      "ProgressVote",
-      "ProgressCount",
-      "ProgressPercent",
+      "Cldr Coverage Count",
+      "Progress Vote",
+      "Progress Count",
+      "Progress Percent",
       "Coverage",
       "Vetter#",
       "Email",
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
```diff
@@ -115,6 +117,8 @@ async function downloadVettingParticipation(opts) {
       "CldrCovCount",
       "ProgressVote",
       "ProgressCount",
+      "ProgressPercent",
```
macchiati, line 120: Minor: for spreadsheets it is better to have spaces in titles so that they wordwrap. So

"Progress Vote",
"Progress Count",
"Progress Percent",

Also, we should have somewhere a key to what these columns are. Can be a Sites page. For example, I don't know off-hand what "Votes" means or what "CldrCovCount" means. I'm guessing:

Votes = total number of path values in this locale that have a vote from this vetter, possibly including paths that are above the organization's coverage target for the locale (such as comprehensive)

CldrCovCount = total number of paths that are in your organization's coverage target for this locale

srl295: Ok and I can add a spreadsheet comment also with more info 

Thanks for the feedback on the wording. 

The percent column isn't working yet but otherwise this is close.

btangmu (dismissed):

tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
srl295: @macchiati I think  it's:

> CldrCovCount = total number of paths that are in *CLDR's* coverage target for this locale

hence the CLDR in the name.

srl295: done

srl295:
Seems to be working PTAL

tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
```diff
@@ -112,9 +115,11 @@ async function downloadVettingParticipation(opts) {
       "Code",
       "Level",
       "Votes",
-      "CldrCovCount",
-      "ProgressVote",
-      "ProgressCount",
+      "Cldr Coverage Count",
```
macchiati, line 118: In that case this should be changed to just Cov Count, and code chanege to be the org's, not CLDR 's. (Whole point of this sheet now is being from the org's perspective)

Also, the votes should be restricted to the org's coverage

macchiati: That is, if a vetter votes for a modern path, but the org's coverage is moderate, then that vote isn't included in Votes.

btangmu (approved):

tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
srl295: @macchiati I don't think it works that way now, so I'll take "Should" as meaning "please change it to…"

In any event, merged this because it makes two improvements as already requested, and I'll add your request to the ticket.

PR #2631, opened 2022-12-24
CLDR-16239 Add CLDRModify options related to inheritance

btangmu:
-Process: -fP like -fp but without inheritance replacement

-Inheritance: -fI like -fp but only inheritance replacement

-Un-drop: -fU revert INHERITANCE_MARKER to pre-drop-hard-inheritance values

CLDR-16239

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
@@ -30,38 +30,15 @@ import org.unicode.cldr.test.CLDRTest;
 import org.unicode.cldr.test.CoverageLevel2;
 import org.unicode.cldr.test.DisplayAndInputProcessor;
 import org.unicode.cldr.test.QuickCheck;
-import org.unicode.cldr.util.Annotations;
-import org.unicode.cldr.util.CLDRConfig;
-import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.*;
 import org.unicode.cldr.util.CLDRFile.DraftStatus;
 import org.unicode.cldr.util.CLDRFile.ExemplarType;
 import org.unicode.cldr.util.CLDRFile.NumberingSystem;
 import org.unicode.cldr.util.CLDRFile.WinningChoice;
-import org.unicode.cldr.util.CLDRLocale;
-import org.unicode.cldr.util.CLDRPaths;
-import org.unicode.cldr.util.CLDRTool;
-import org.unicode.cldr.util.CldrUtility;
-import org.unicode.cldr.util.DateTimeCanonicalizer;
 import org.unicode.cldr.util.DateTimeCanonicalizer.DateTimePatternType;
-import org.unicode.cldr.util.DtdData;
-import org.unicode.cldr.util.DtdType;
-import org.unicode.cldr.util.Factory;
-import org.unicode.cldr.util.FileProcessor;
-import org.unicode.cldr.util.LanguageTagParser;
-import org.unicode.cldr.util.Level;
 // import org.unicode.cldr.util.Log;
-import org.unicode.cldr.util.LogicalGrouping;
-import org.unicode.cldr.util.PathChecker;
-import org.unicode.cldr.util.PatternCache;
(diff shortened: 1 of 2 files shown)


btangmu:
This PR only adds new options for CLDRModify, without changing the behavior of any existing options. The new options will facilitate the ongoing "blocks progress" work on inheritance.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -1259,6 +1236,130 @@ public void handlePath(String xpath) {
...
+        });
+
+        // use DAIP for one thing only: replaceBaileyWithInheritanceMarker
+        fixList.add('I', "Inheritance-substitution", new CLDRFilter() {
+            private DisplayAndInputProcessor inputProcessor;
+            private final int STEPS_FROM_ROOT = 1; // only process if locale's level matches; root = 0, en = 1, ...
+
+            @Override
+            public void handleStart() {
+                int steps = stepsFromRoot(cldrFileToFilter.getLocaleID());
```
macchiati, line 1272: If we fix the inheritance, we may not have to proceed in levels. But we can decide that later, so this is ok.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -1259,6 +1236,130 @@ public void handlePath(String xpath) {
...
+            }
+
+            @Override
+            public void handleEnd() {
+                inputProcessor = null; // clean up, just in case
+            }
+
+            @Override
+            public void handlePath(String xpath) {
+                if (inputProcessor == null) {
```
macchiati, line 1302: probably want to throw exception here instead of just returning. That is, we sould always have an input processor for each locale.

macchiati (commented):
All looks good. The comments I made can be postponed until later.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
btangmu: actually for -fU we get inputProcessor == null for a few locales including ajp that don't occur in the pre-drop data -- details are in one of the draft PRs

also we get inputProcessor == null if steps != STEPS_FROM_ROOT

these are cases where the current data will be kept without modification

srl295: OK if this is a TODO - fix missing inputProcessor for some locales.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -1259,6 +1236,130 @@ public void handlePath(String xpath) {
...
+        });
+
+        // Un-drop hard inheritance: revert INHERITANCE_MARKER to pre-drop-hard-inheritance values
+        fixList.add('U', "Un-drop inheritance", new CLDRFilter() {
+            // baseDir needs to be the "pre-drop" path of an existing copy of old common/main
+            // For example, 2022_10_07_pre folder gets xml from pull request 2433, commit 80029f1
+            // Also ldml.dtd is required; for example:
+            //   mkdir ../2022_10_07_pre/common/dtd
+            //   cp common/dtd/ldml.dtd ../2022_10_07_pre/common/dtd
+            final private String baseDir = "../2022_10_07_pre/";
```
srl295, line 1322: Ideally would be relative to the Generated dir, less arbitrary than `..` but not a blocker

srl295 (approved):

btangmu:
@srl "TODO - fix missing inputProcessor for some locales" -- actually the locales (ajp, ...) didn't exist yet in the old (pre-drop) data, so there doesn't seem to be anything to fix

srl295:
> @SRL "TODO - fix missing inputProcessor for some locales" -- actually the locales (ajp, ...) didn't exist yet in the old (pre-drop) data, so there doesn't seem to be anything to fix

OK, so there's not a problem with DAIP for those locales, it's that they don't exist yet. Sounds good then.

PR #2641, opened 2023-01-03
CLDR-16272 limit some reports during limited submission

srl295:
CLDR-16272

- add a limited set of available reports in SubmissionLocales
- only show available reports in the sidebar
- don't count unavailable reports against completion in Dashboard
- throw 403 Forbidden if user tries to navigate to or vote on un-available reports

ALLOW_MANY_COMMITS=true


srl295:
also see PR #2640

macchiati (approved):

srl295:
@macchiati great, will merge when built.  Or feel free to do the honors if you want.

macchiati:
merged, thanks!

PR #2642, opened 2023-01-05
CLDR-15034 kbd: add @ALLOWS_UESC DTD annotation

srl295:
_(Note: Keyboard PR- not for v43!)_

CLDR-15034

Add a `<!--@ALLOWS_UESC-->` DTD annotation noting where `\u{…}` escaping is allowed on an attribute.
This does not address changes to UnicodeSet escaping, because there are still open issues around UnicodeSets in keyboard

CLDRFile is not affected, and the escaping is optional.

Currently this annotation is only on Keyboard attributes, so escaping is not permitted on any non-keyboard data items

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -243,18 +243,28 @@ If it becomes necessary in the future, the format could extend the ISO layout to
 
 When explicitly specified, attributes can contain escaped characters. This specification uses two methods of escaping, the _UnicodeSet_ notation and the `\u{...}` notation.
 
+### UnicodeSet Escaping
+
 The _UnicodeSet_ notation is described in [UTS #35 section 5.3.3](tr35.md#Unicode_Sets) and allows for comprehensive character matching, including by character range, properties, names, or codepoints. Currently, the following attributes allow _UnicodeSet_ notation:
 
 * `from` or `before` on the `<transform>` element
 * `from` or `before` on the `<reorder>` element
 * `chars` on the [`<repertoire>`](#test-element-repertoire) test element.
 
+### UTS18 Escaping
+
 The `\u{...}` notation, a subset of hex notation, is described in [UTS #18 section 1.1](https://www.unicode.org/reports/tr18/#Hex_notation). It can refer to one or multiple individual codepoints. Currently, the following attributes allow the `\u{...}` notation:
 
-* `to`, `longPress`, `multiTap`, `hint` on the `<map>` element
+* `to`, `longPress`, `multiTap`, and `longPressDefault` on the `<key>` element
+* `to` on the `<flick>` element
 * `to` on the `<transform>` element
+* `to` and `display` on the `<display>` element
+* `baseCharacter` on the `<displayOptions>` element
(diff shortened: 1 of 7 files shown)


macchiati (changes requested):
I think we need more information about the impact of this statement. If the attribute foo has it:

- if the rawValue is \u{61}b, then the resolved value is 'ab', right?
- if we read the values, eg with CLDRFile, is the software expected to resolve the value?
- if we write the values, eg with CLDRFile, which characters are expected to be escaped?

Some of these are gaps in the original explanation of \u in keyboards, but we should take the time to nail it down.

srl295:
> I think we need more information about the impact of this statement. If the attribute foo has it:
> 
> 
> 
> - if the rawValue is \u{61}b, then the resolved value is 'ab', right?

Yes 

> - if we read the values, eg with CLDRFile, is the software expected to resolve the value?
> 
> - if we write the values, eg with CLDRFile, which characters are expected to be escaped?

Right now these are only on keyboard attributes. I don't make any statement here about CLDRFile but potentially it could do as you say, with some unicode set that defines what needs to be escaped. 



> 
> 
> Some of these are gaps in the original explanation of \u in keyboards, but we should take the time to nail it down.

Right. Currently the spec for \u is prose on certain elements and this makes it metadata. 

An alternate approach would be to use a VALUE subtype indicating that escaping is acceptable.   VALUE any/escaped perhaps.

macchiati:
Hmmm. It doesn't quite fit the VALUE use case. But maybe if we changed the name for clarity, eg @ALLOWS_UESC, to indicate that the attribute value allows \u syntax (but doesn't require it)?

mcdurdin (commented):
I concur with the `@ALLOWS_UESC` rename -- otherwise afaict the changes are correct.

docs/ldml/tr35-keyboards.md
```diff
@@ -243,19 +243,29 @@ If it becomes necessary in the future, the format could extend the ISO layout to
...
+* `to`, `longPress`, `multiTap`, and `longPressDefault` on the `<key>` element
+* `to` on the `<flick>` element
 * `to` on the `<transform>` element
+* `to` and `display` on the `<display>` element
+* `baseCharacter` on the `<displayOptions>` element
+* Some attributes on [Keyboard Test Data](#Keyboard Test Data) subelements
 
 Characters of general category of Combining Mark (M), Control characters (Cc), Format characters (Cf), and whitespace other than space should be encoded using one of the notation above as appropriate.
 
+Attributes escaped in this manner are annotated with the `<!--@ALLOWS_UESC-->` DTD annotation, see [DTD Annotations](tr35.md#57-dtd-annotations)
```
macchiati, line 267: Attribute => Attribute values

macchiati: Multiple places below.

macchiati (changes requested):

docs/ldml/tr35-keyboards.md
srl295: good catch, thanks

macchiati (approved):

PR #2643, opened 2023-01-06
CLDR-16199 remove draft="provisional" from root <namePattern> entries

pedberg-icu:
CLDR-16199

Remove the 'draft="provisional"' that seems to have been accidentally added to all of the root <namePattern> entries with real data (as opposed to being aliases). This must have been a side effect of resetting the personName data for CLDR 43 Survey Tool submission; but data in root and en should never be provisional. This was causing resource load errors in the ICU4J PersonName code.


srl295 (approved):

Paaplo9 (commented):
Duplicate of ### 

> 

- [ ] _**#``**_

PR #2645, opened 2023-01-13
CLDR-16302 Adding core data for cho (Choctaw)

ashashkina:
CLDR-16302

https://unicode-org.atlassian.net/browse/CLDR-16302
@pedberg-icu

### DIFF
--- a/common/main/cho.xml
+++ b/common/main/cho.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2021 Unicode, Inc.
+<!-- Copyright © 1991-2023 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-DFS-2016
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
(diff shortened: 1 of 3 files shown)


DavidLRowe:
The common/main/cho.xml file has an error on lines 25 and 31, where there are an extraneous set of <characters> and </characters> tags.

I expect that the existing CLDR file exemplars/main/cho.xml will need to be removed.

pedberg-icu:
Hmm, the one unt test failure has nothing to do with this PR:
```
2023-01-14T00:57:01.9763997Z     TestCurrency
2023-01-14T00:57:01.9774385Z ##[error] (TestSupplementalInfo.java:1426)  Error: Codes in ISO 4217 but not current tender in CLDR (may need to update https://cldr.unicode.org/development/updating-codes/update-currency-codes ): [HRK]
```
But it is probably related to the fact that after 2023-01-01 , HRK may no longer be legal tender (Croatia switched to Euro). I think someone else should file a separate PR about that, then you can rebase on main once that is merged. I will send an e-mail to the list...

srl295:
> Hmm, the one unt test failure has nothing to do with this PR:
> 
> ```
> 2023-01-14T00:57:01.9763997Z     TestCurrency
> 2023-01-14T00:57:01.9774385Z ##[error] (TestSupplementalInfo.java:1426)  Error: Codes in ISO 4217 but not current tender in CLDR (may need to update https://cldr.unicode.org/development/updating-codes/update-currency-codes ): [HRK]
> ```
> 
> But it is probably related to the fact that after 2023-01-01 , HRK may no longer be legal tender (Croatia switched to Euro). I think someone else should file a separate PR about that, then you can rebase on main once that is merged. I will send an e-mail to the list...

This is now fixed in #2646, so if you rebase that test should pass

pedberg-icu (approved):

PR #2652, opened 2023-01-19
CLDR-15034 kbd: spec: move hardware type into 'form'

srl295:
thanks to @mcdurdin suggestion below, this changes the format to the below. Dropping the `hardware=` attribute, and changing the meaning of `form` to be the hardware type or otherwise `touch`

- more than one hardware layout is allowed per file. This is pursuant to the 'single file' goal. Otherwise we will need one `fr.xml` for US and one `fr.xml` for ISO, etc.
- only one `layers` allowed per 'form'.  So, illegal to have two with `form="iso"`.

```xml
<keyboard>
    <layers form="touch" />
    <layers form="touch" minDeviceWidth="100" /> <!-- distinguished by device width -->
    <layers form="iso" /> <!-- hardware -->
    <layers form="us" /> <!-- hardware -->
</keyboard>
```

-----

<details><summary>previous format:</summary>

Previously, `form=touch` or `form=hardware`, and a separate attribute for `hardware=iso` etc.  Also previously, only one `hardware` layers was allowed.

```xml
<keyboard>
    <layers form="touch" /> <!-- etc… -->
    <layers form="touch" minDeviceWidth="100" /> <!-- etc… -->
    <layers form="hardware" hardware="iso" /> <!-- etc… -->
</keyboard>
```

</details>

CLDR-15034

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1151,7 +1151,7 @@ hardware or touch layout.
 - At least one `layers` element is required.
 - Multiple `<layers form="touch">` elements are allowed with distinct `minDeviceWidth` values.
 - At most one `<layers>` element is allowed per hardware `form`. In other words,
-at most one `<layers form="jis">` element is allowed.
+  at most one `<layers form="jis">` element is allowed.
 
 _Attribute:_ `form` (required)
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1149,19 +1149,24 @@ hardware or touch layout.
 > </small>
 
 - At least one `layers` element is required.
-- Multiple `<layers form="touch">` elements are allowed with distinct `minDeviceWidth` values.
-- At most one `<layers>` element is allowed per hardware `form`. In other words,
-  at most one `<layers form="jis">` element is allowed.
 
 _Attribute:_ `form` (required)
 
 > This attribute specifies the physical layout of a hardware keyboard,
-> or that the form is a touch layout.
+> or that the form is a `touch` layout.
 >
 > It is recommended to always have at least one hardware (non-touch) form.
 > If there is no `hardware` form, the implementation may need
 > to choose a different keyboard file, or use some other fallback behavior when using a
 > hardware keyboard.
+>
+> Multiple `<layers form="touch">` elements are allowed with distinct `minDeviceWidth` values.
+> At most one `<layers>` element is allowed per hardware `form`. In other words, at most one `<layers form="jis">` element is allowed.
+>
+> In most situations, only one hardware `layers` element will be included. Note that the typical keyboard author will be designing a keyboard based on their circumstances and the hardware that they are using. So, for example, if they are in South East Asia, they will almost certainly be using an 101 key hardware keyboard with US key caps. So we want them to be able to reference that (`<layers form="us">`) in their design, rather than having to work with an unfamiliar form. An implementation should be able to translate the keyboard designer's chosen form to the other hardware forms.
+
+
+
 >
 > When using an on-screen keyboard, if there is not a `<layers form="touch">`
 > element, the hardware elements can be used for on-screen use.
@@ -1175,7 +1180,7 @@ _Attribute:_ `form` (required)
 
 _Attribute:_ `minDeviceWidth`
 
-> This attribute specifies the minimum required width, in millimeters (mm), of the touch surface.  The `layers` entry with the greatest matching width will be selected. This attribute is intended for `form="touch"`, but is supported for hardware forms`.
+> This attribute specifies the minimum required width, in millimeters (mm), of the touch surface.  The `layers` entry with the greatest matching width will be selected. This attribute is intended for `form="touch"`, but is supported for hardware forms.
 
 ### <a name="Element_layer" href="#Element_layer">Element: layer</a>
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1152,9 +1152,9 @@ _Attribute:_ `form` (required)
 
 > `form` may be either `hardware` or `touch`.
 >
-> There may only be a single `layers` with `form="hardware"`.
+> There may only be a single (at most one) `layers` element with `form="hardware"` for each `hardware=…` type.
 >
-> It is recommended to always have one
+> It is recommended to always have at least one
 > `<layers form="hardware">` element.
 > If there is no `hardware` form, the implementation may need
 > to choose a different keyboard file, or use some other fallback behavior when using a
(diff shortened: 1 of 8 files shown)


srl295:
Where this is going is that we should be able to actually drop `form="hardware | touch"` completely…… and just add a `hardware="touch"` indicating what `form=touch` used to.

In other words here is what we have now:

```xml
<keyboard>
    <layers form="touch" />
    <layers form="hardware" hardware="iso" />
    <layers form="hardware" hardware="us" /> <!-- not allowed by current spec -->
</keyboard>
```

but we could change it to:

```xml
<keyboard>
    <layers hardware="touch" />
    <layers hardware="iso" />
    <layers hardware="us" />
</keyboard>
```

mcdurdin:
Agree that we don't need both `hardware` and `form`. I feel like using `form` instead of `hardware` for the attribute name may be clearer, because `hardware="touch"` may be confusing?

> ```
> <keyboard>
>     <layers form="touch" />
>     <layers form="iso" />
>     <layers form="us" />
> </keyboard>
> ```

srl295:
> Agree that we don't need both `hardware` and `form`. I feel like using `form` instead of `hardware` for the attribute name may be clearer, because `hardware="touch"` may be confusing?
> 
> > ```
> > <keyboard>
> >     <layers form="touch" />
> >     <layers form="iso" />
> >     <layers form="us" />
> > </keyboard>
> > ```

That makes sense.

srl295:
OK PTAL all … this now drops the `hardware` attribute as discussed.

mcdurdin:
> * only one `layers` allowed per 'form'.  So, illegal to have two with `form="iso"`.

We allow multiple `layers` for `form='touch'`, distinguished by `minDeviceWidth`.

docs/ldml/tr35-keyboards.md
```diff
@@ -1130,29 +1130,34 @@ hardware or touch layout.
 >
 > </small>
 
+- At least one `layers` element is required.
+- Multiple `<layers form="touch">` elements are allowed with distinct `minDeviceWidth` values.
+- At most one `<layers>` element is allowed per hardware `form`. In other words,
+at most one `<layers form="jis">` element is allowed.
```
mcdurdin, line 1136: formatnit

```suggestion
  at most one `<layers form="jis">` element is allowed.
```

mcdurdin (approved):

srl295:
@miloush I think this was the one you had questions about, can you add any concerns here? thanks

miloush:
What I am gathering from the description is that the idea is to have multiple hardware layouts based on the physical keyboard arrangement, and the attribute values refer to the physical layout, i.e. us means ANSI104, iso means ISO105. Is that correct? If so, the attribute values are not clear enough.

I imagine there will be a lot of duplication between these layouts and only the various keys like OEM keys would differ. Is there any sharing of the data? Is there a set of hardware layouts we recommend authors to include, or can the missing ones be automatically inferred?

Since you allow multiple hardware layouts I am assuming the attribute is "that's what applies when ISO105 is connected" rather than "this layout is a ISO105 layout". How would you know which physical arrangement is connected to the device? What happens if ISO105 is connected but the file only contains ANSI104 one?

mcdurdin:
@miloush, thank you for the helpful feedback. I agree that the enumeration values are somewhat imprecise at present, and I am grateful for any thoughts on how to tighten them up.

In terms of intention: the typical keyboard author will be designing a keyboard based on their circumstances and the hardware that they are using. So, for example, if they are in South East Asia, they will almost certainly be using an 101 key hardware keyboard with US key caps. So we want them to be able to reference that in their design, rather than having to work with an unfamiliar form.

The keyboard processor should be able to translate the keyboard designer's chosen form to the other hardware forms. (AFAIK, the key count in the various standards differs by only 1 or 2, apart from frame keys, which are currently excluded, and scan codes are generally stable?)

Generally, I expect there will only be a single hardware layout in a file, but there is no specific reason to block having multiple. I could see edge cases where people working in places such as parts of Africa, where there are multiple hardware standards in use, might want to tweak the hardware layouts for each form. But I imagine it would be rare.

mcdurdin:
> What happens if ISO105 is connected but the file only contains ANSI104 one?

In general, unspecified keys should probably 'inherit' system behaviour -- it's either that or no-op, which is probably less helpful. We _could_ make it a layout option but that probably isn't all that useful either.

miloush:
Thank you @mcdurdin. The scenario/use case is what I was missing most (also the "single file" note could mean all platform layouts in single file).

However, isn't the idea of the `<layer>` element to represent a mapping from spatial arrangement to output and then have the platform figure out how the input maps to the spatial arrangement?

I have been recently contacted by a colleague having Japanese layout on his laptop (e.g. something like OADG109A). He is using Swiss French layout on Windows and some of the keys that the Swiss French layout (or any other western layout) outputs are unreachable, because the 109-key keyboard produces different SC for some of the keys from what ANSI104 layout would, some of which are not covered by the "standard" SC to VK mapping at all. (FYI @xadxura)

I am hoping that we would have a solution to this. Ideally, the spatial arrangement in `<layer>` would remain the same and the platform would figure out which SC is the key next to backspace based on whether the hardware is 109- or 101-key (determining that might be the tricky bit).

But I think having a separate layer in LDML for 109- and 101- key would already be an improvement enabling people to fix this scenario at least manually - so in principle I welcome this direction. It's just that this PR did not come with a spec update about what and why is going on. It sounds like the attribute value is mostly intended to mean "this LDML file was created using ISO105 layout".

mcdurdin:
> However, isn't the idea of the `<layer>` element to represent a mapping from spatial arrangement to output and then have the platform figure out how the input maps to the spatial arrangement?

Yes, exactly.

In vast majority of cases, I hope to see only 1 `<layer form='<hardware>'>` element in a file, and the platform can transform that to any other forms it supports. But hopefully allowing keyboard designers to specify their layout according to their own hardware layout rather than an arbitrarily chosen one will be helpful to them.

> It's just that this PR did not come with a spec update about what and why is going on.

Indeed. @srl295 can we add some wording on the why to the spec details? Hopefully we can start from what I wrote in https://github.com/unicode-org/cldr/pull/2652#issuecomment-1403012988 and @miloush's response and expansion?

@miloush, do you have some clear standard names we should use for the enumeration? Of course, so long as they are unique within the context of this spec and unambiguous, we don't necessarily need to use a long and unwieldly reference name in the enum itself, just in the docs for it. (That is, I'd prefer `'us'` to '[INCITS 154-1988 [S2009]](https://blog.ansi.org/standardization-of-keyboard-layouts/)', and `'iso'` to 'ISO/IEC 9995-2:2009'.)

srl295:
There's also a challenge I can see, and that is what if the xml has both a `us` and an `iso` `<layers>` element, but. you plugin a  JIS keyboard.  This could be a similar issue to the one @miloush mentioned.

srl295:
@miloush @mcdurdin  I added some text from comments into the PR.

mcdurdin:
> There's also a challenge I can see, and that is what if the xml has both a `us` and an `iso` `<layers>` element, but. you plugin a JIS keyboard. This could be a similar issue to the one @miloush mentioned.

I think we can leave this detail to implementers -- my inclination would be to use `iso` in this case as it defines an extra key.

srl295:
> > There's also a challenge I can see, and that is what if the xml has both a `us` and an `iso` `<layers>` element, but. you plugin a JIS keyboard. This could be a similar issue to the one @miloush mentioned.
> 
> I think we can leave this detail to implementers -- my inclination would be to use `iso` in this case as it defines an extra key.

That seems like it could be prone to incompatibilities. Maybe this should just be noted as a known issue.

miloush:
We could say that if none of the layouts matches the hardware (or the OS does not know the hardware layout), the implementation just uses the first one defined.

srl295:
- rebased

mcdurdin (approved):

srl295:
I'm going to address the test failures separately.

PR #2654, opened 2023-01-20
CLDR-16239 New boolean constant VoteResolver.DROP_HARD_INHERITANCE

btangmu:
-DROP_HARD_INHERITANCE = true: no change in behavior

-DROP_HARD_INHERITANCE = false: partial reversion to pre-drop-hard-inheritance behavior

-Making it false reduces changes from baseline when generating VXML

CLDR-16239

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -582,7 +582,7 @@ public class DataPage {
             if (value == null) {
                 return null;
             }
-            if (value.equals(inheritedValue) && !inheritsFromRootOrFallback()) {
+            if (VoteResolver.DROP_HARD_INHERITANCE && value.equals(inheritedValue) && !inheritsFromRootOrFallback()) {
                 value = CldrUtility.INHERITANCE_MARKER;
             }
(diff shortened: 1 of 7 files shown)


btangmu:
In this PR, the new boolean VoteResolver.DROP_HARD_INHERITANCE is true, meaning no change in behavior. In the long run, the expectation is that it will be true, or will be removed (effectively permanently true). In the short run, for v43, setting it to false (locally) results in a huge reduction in changes from baseline when generating VXML.

As a crude measurement, with DROP_HARD_INHERITANCE = true,

diff -ICopyright common/main ../vetdata-2023-01-17T18-15-14-791738Z/vxml/common/main | wc -l
  811555

With DROP_HARD_INHERITANCE = false,
diff -ICopyright common/main ../vetdata-2023-01-20T18-29-48-707269Z/vxml/common/main | wc -l
   45935

45935 / 811555 = 5.6%

srl295 (approved):
Reviewed but forgot to click approve before!

macchiati (approved):

macchiati:
This looks good. I'm assuming this doesn't incorporate the algorithm in the
doc for the v43 solution, so it is just to get us the VXML, to which you'll
apply another tool to do the algorithm.

Is that correct?

PR #2659, opened 2023-01-23
CLDR-16239 Generated VXML, plain (no post-process), DROP_HARD_INHERITANCE = false

btangmu:
CLDR-16239

ALLOW_MANY_COMMITS=true


macchiati (approved):

btangmu:
Observations:

- many values change to ^^^, surprising since this is based on "drop=false" vxml
- problem remains: "{0} swaartekrag" changed to "{0} G", no reversion to baseline since vxmlValue is not ^^^

PR #2661, opened 2023-01-24
CLDR-15034 kbd: change modifiers= to be NMTOKENS, doc

srl295:
- update docs
- update mt.xml

CLDR-15034

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1236,6 +1236,8 @@ _Attribute:_ `modifier` (required for `hardware`)
 > - `shift`
 >
 > Note that `alt` is sometimes referred to as _opt_ or _option_.
+>
+> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
 
 > For hardware layouts, the use of `@modifier` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1238,6 +1238,8 @@ _Attribute:_ `modifier` (required for `hardware`)
 > Note that `alt` is sometimes referred to as _opt_ or _option_.
 >
 > Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
+>
+> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
 
 > For hardware layouts, the use of `@modifier` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1217,9 +1217,28 @@ _Attribute_ `id` (required for `touch`)
 
 _Attribute:_ `modifier` (required for `hardware`)
 
-> This has two roles. It acts as an identifier for the `layer` element for hardware keyboards (in the absence of the id= element) and also provides the linkage from the hardware modifiers into the correct `layer`. To indicate that no modifiers apply, the reserved name of "none" can be used. For the purposes of fallback vkey mapping, the following modifier components are reserved: "shift", "ctrl", "alt", "caps", "cmd", "opt" along with the "L" and "R" optional single suffixes for the first 3 in that list.
-
-For hardware layouts, the use of `@modifier` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
+> This has two roles. It acts as an identifier for the `layer` element for hardware keyboards (in the absence of the id= element) and also provides the linkage from the hardware modifiers into the correct `layer`.
+>
+> To indicate that no modifiers apply, the reserved name of `none` can be used.
+> The following modifier components can be used, separated by spaces.
+> Note that `L` or `R` indicates a left- or right- side modifier only (such as `shiftL`)
+> whereas `shift` indicates _either_ left or right shift.
+>
+> - `none` (no modifier, may not be combined with others)
+> - `alt`
+> - `altL`
+> - `altR`
+> - `caps`
+> - `cmd`
+> - `ctrl`
+> - `ctrlL`
+> - `ctrlR`
+> - `opt`
+> - `shift`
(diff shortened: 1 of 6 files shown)


keyboards/dtd/ldmlKeyboard.dtd
```diff
@@ -149,7 +149,7 @@ Please see CLDR-15034 for the latest information. -->
 <!ELEMENT layer ( row+, special* ) >
 <!ATTLIST layer id NMTOKEN #IMPLIED >
     <!--@MATCH:any-->
-<!ATTLIST layer modifier NMTOKEN #IMPLIED >
+<!ATTLIST layer modifier NMTOKENS #IMPLIED >
     <!--@MATCH:any-->
```
mcdurdin, line 153: We should be able to restrict the set of possible modifiers here now?

keyboards/dtd/ldmlKeyboard.xsd
```diff
@@ -299,7 +299,7 @@
         <xs:element minOccurs="0" maxOccurs="unbounded" ref="special"/>
       </xs:sequence>
       <xs:attribute name="id" type="xs:NMTOKEN"/>
-      <xs:attribute name="modifier" type="xs:NMTOKEN"/>
+      <xs:attribute name="modifier" type="xs:NMTOKENS"/>
```
mcdurdin, line 302: Can we list the possible modifiers here?

mcdurdin (changes requested):

keyboards/dtd/ldmlKeyboard.dtd
srl295: Yea with a `@MATCH` though is it a closed set ?

mcdurdin: We could list the allowable combos, e.g. Right + Left modifiers together is not sensible; Alt + Option not helpful (query whether Option should be included since it is roughly equivalent to Alt anyway). There's only really a smallish set of combos possible. This would also help us to keep the order of modifiers consistent.

srl295:
I plan to make these changes to the PR also:

- [ ] drop `opt` - just use `alt` instead.  These keys are synonymous, in fact the option key on apple keyboards is often labelled `alt` as well.
- [ ] drop `cmd`  - out of scope for these keyboards, just as super (aka windows key) is out of scope.
- [ ] drop left vs right distinction for shift - just a single `shift`.

keyboards/dtd/ldmlKeyboard.xsd
srl295: probably not *here* right now, but will list them in annotations

srl295: CLDR-15990 is the general ticket for dtd -> xsd and covers converting the @ -annotations mapping to more precise XSD.

srl295:
> We could list the allowable combos, e.g. Right + Left modifiers together is not sensible; Alt + Option not helpful (query whether Option should be included since it is roughly equivalent to Alt anyway). There's only really a smallish set of combos possible. This would also help us to keep the order of modifiers consistent.
 

take a look at this update

srl295:
> I plan to make these changes to the PR also:

- [X] drop `opt` - just use `alt` instead.  These keys are synonymous, in fact the option key on apple keyboards is often labelled `alt` as well.
- [X] drop `cmd`  - out of scope for these keyboards, just as super (aka windows key) is out of scope.
- [X] drop left vs right distinction for shift - just a single `shift` 

done

keyboards/dtd/ldmlKeyboard.dtd
```diff
@@ -149,8 +149,8 @@ Please see CLDR-15034 for the latest information. -->
 <!ELEMENT layer ( row+, special* ) >
 <!ATTLIST layer id NMTOKEN #IMPLIED >
     <!--@MATCH:any-->
-<!ATTLIST layer modifier NMTOKEN #IMPLIED >
-    <!--@MATCH:any-->
+<!ATTLIST layer modifier NMTOKENS #IMPLIED >
+    <!--@MATCH:regex/(none|((alt[LR]?|ctrl[LR]?|caps|shift)[ ]?)+)-->
```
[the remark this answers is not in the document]
srl295: I'm going to come back to the regex once there's consensus on the actual direction.

docs/ldml/tr35-keyboards.md
```diff
@@ -1183,9 +1183,27 @@ _Attribute_ `id` (required for `touch`)
...
+> - `alt`
+> - `altL`
+> - `altR`
+> - `caps`
+> - `ctrl`
+> - `ctrlL`
+> - `ctrlR`
+> - `shift`
+>
+> Note that `alt` is sometimes referred to as _opt_ or _option_.
```
mcdurdin, line 1204:
```suggestion
> Note that `alt` is sometimes referred to as _opt_ or _option_.
>
> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
```

srl295:
also rebased

keyboards/dtd/ldmlKeyboard.dtd
```diff
@@ -155,8 +155,8 @@ Please see CLDR-15034 for the latest information. -->
 <!ELEMENT layer ( row+, special* ) >
 <!ATTLIST layer id NMTOKEN #IMPLIED >
     <!--@MATCH:any-->
-<!ATTLIST layer modifier NMTOKEN #IMPLIED >
-    <!--@MATCH:any-->
+<!ATTLIST layer modifier NMTOKENS #IMPLIED >
+    <!--@MATCH:regex/^(none|(\b(alt[LR]?)?\b[ ]?\b(ctrl[LR]?)?\b[ ]?\b(shift)?\b[ ]?\b(caps)?\b))$-->
```
mcdurdin, line 159: OK, so in the Keyman repo, you sorted the tokens alphabetically in the string before processing. Should we consider doing that here too? (`"alt[LR] caps ctrl[LR] shift"` -- case against: it feels awkward throwing caps in the middle)

mcdurdin (approved):

keyboards/dtd/ldmlKeyboard.dtd
miloush: didn't we talk about this being potentially open ended set for cases with extra shift states?

srl295: i'll make it open ended

srl295: done

PR #2663, opened 2023-01-25
CLDR-16348 Add Gregorian fallback eras in Japanese calendar

sffc:
CLDR-16348

I wanted to open a PR for this small subpart of CLDR-16185 to get early feedback.

ALLOW_MANY_COMMITS=true

CC @Manishearth @FrankYFTang @yumaoka

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -4411,6 +4411,8 @@ XXX Code for transations where no currency is involved
         <calendar type="japanese">
           <calendarSystem type="solar" />
           <eras>
+              <era type="237" end="0-12-31"/>
+              <era type="238" start="1-1-1"/>
               <era type="0" start="645-6-19"/>
               <era type="1" start="650-2-15"/>
               <era type="2" start="672-1-1"/>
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

srl295:
@sffc sorry. i don't see this listed in https://unicode-org.atlassian.net/browse/CLDR-16185 and in any event, it seems like it's as you say only a very small part.

I would suggest that you open a subtask of CLDR-16185 (you can go ahead and accept a subtask without extra permission) for the 'small part', including justification, and target this PR there.

Manishearth (approved):

macchiati (changes requested):
This is not what the committee agreed to.

sffc:
> I would suggest that you open a subtask (hit 'create subtask' in jira) of [CLDR-16185](https://unicode-org.atlassian.net/browse/CLDR-16185) (you can go ahead and accept a subtask without extra permission) for the 'small part', including justification, and target this PR there.

Done. The subtask is New, not Accepted. Should I accept it?

> This is not what the committee agreed to.

Updated according to the discussion last week.

macchiati (approved):

macchiati:
Note: the tests might hiccup on the negative numbers, and need fixing. 
We will also need a ticket on the ICU side to adjust the CLDR data conversion to skip these (or maybe this would be done as a part of the integration work, @pedberg-icu ?)

Manishearth (approved):

sffc:
I folded this in with #2665.

PR #2665, opened 2023-01-26
CLDR-16185 Add era codes to supplementalData

sffc:
CLDR-16185

ALLOW_MANY_COMMITS=true

@Manishearth @yumaoka @FrankYFTang

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -250,6 +250,11 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST era end CDATA #IMPLIED >
     <!--@MATCH:time/yyyy-MM-dd-->
     <!--@VALUE-->
+<!ATTLIST era code NMTOKEN #IMPLIED >
+    <!--@MATCH:regex/[a-z0-9]{3,8}(\-[a-z0-9]{3,8})*-->
+    <!--@VALUE-->
+<!ATTLIST era aliases NMTOKENS #IMPLIED >
+    <!--@VALUE-->
 <!ATTLIST era named (true | false) #IMPLIED >
     <!--@VALUE-->
 
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -4402,8 +4402,8 @@ XXX Code for transations where no currency is involved
         <calendar type="gregorian">
            <calendarSystem type="solar" />
            <eras>
-               <era type="0" end="0-12-31" />
-               <era type="1" start="1-01-01" />
+               <era type="0" end="0-12-31" code="gregory-inverse" aliases="bc bce"/>
+               <era type="1" start="1-01-01" code="gregory" aliases="ad ce"/>
            </eras>
         </calendar>
         <calendar type="generic"/>
@@ -4411,6 +4411,8 @@ XXX Code for transations where no currency is involved
         <calendar type="japanese">
           <calendarSystem type="solar" />
           <eras>
+              <era type="-2" end="0-12-31" code="japanese-inverse" aliases="gregory-inverse bc bce"/>
+              <era type="-1" start="1-01-01" code="japanese" aliases="gregory ad ce"/>
               <era type="0" start="645-6-19"/>
               <era type="1" start="650-2-15"/>
               <era type="2" start="672-1-1"/>
@@ -4643,101 +4645,101 @@ XXX Code for transations where no currency is involved
               <era type="229" start="1861-2-19"/>
               <era type="230" start="1864-2-20"/>
               <era type="231" start="1865-4-7"/>
-              <era type="232" start="1868-9-8"/>
-              <era type="233" start="1912-7-30"/>
-              <era type="234" start="1926-12-25"/>
-              <era type="235" start="1989-1-8"/>
-              <era type="236" start="2019-5-1"/>
+              <era type="232" start="1868-9-8" code="meiji"/>
+              <era type="233" start="1912-7-30" code="taisho"/>
+              <era type="234" start="1926-12-25" code="showa"/>
+              <era type="235" start="1989-1-8" code="heisei"/>
+              <era type="236" start="2019-5-1" code="reiwa"/>
           </eras>
         </calendar>
         <calendar type="islamic">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15"/>
+              <era type="0" start="622-7-15" code="islamic" aliases="ah"/>
           </eras>
         </calendar>
         <calendar type="islamic-civil">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-16"/>
+              <era type="0" start="622-7-16" code="islamic-civil" aliases="islamicc ah"/>
           </eras>
         </calendar>
         <calendar type="islamic-rgsa">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15"/>
+              <era type="0" start="622-7-15" code="islamic-rgsa" aliases="ah"/>
           </eras>
         </calendar>
         <calendar type="islamic-tbla">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15"/>
+              <era type="0" start="622-7-15" code="islamic-tbla" aliases="ah"/>
           </eras>
         </calendar>
         <calendar type="islamic-umalqura">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15"/>
+              <era type="0" start="622-7-15" code="islamic-umalqura" aliases="ah"/>
           </eras>
         </calendar>
         <calendar type="chinese">
           <calendarSystem type="lunisolar"/>
           <eras>
-              <era type="0" start="-2636-01-01"/> <!-- 'sequential' year -->
+              <era type="0" start="-2636-01-01" code="chinese"/> <!-- 'sequential' year -->
           </eras>
         </calendar>
         <calendar type="hebrew">
           <calendarSystem type="lunisolar"/>
           <eras>
-              <era type="0" start="-3760-10-7"/>
+              <era type="0" start="-3760-10-7" code="hebrew" aliases="am"/>
           </eras>
         </calendar>
         <calendar type="buddhist">
           <calendarSystem type="solar" />
           <eras>
-              <era type="0" start="-542-01-01"/>
+              <era type="0" start="-542-01-01" code="buddhist" aliases="be"/>
           </eras>
         </calendar>
         <calendar type="coptic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="284-08-28"/>
-              <era type="1" start="284-08-29"/>
+              <era type="0" end="284-08-28" code="coptic-inverse"/>
+              <era type="1" start="284-08-29" code="coptic"/>
           </eras>
         </calendar>
         <calendar type="persian">
           <calendarSystem type="solar"/>
           <eras>
-              <era type="0" start="622-01-01"/>
+              <era type="0" start="622-01-01" code="persian" aliases="ap"/>
           </eras>
         </calendar>
         <calendar type="dangi">
           <calendarSystem type="lunisolar"/>
           <eras>
-              <era type="0" start="-2332-01-01"/> <!-- 'sequential' year -->
+              <era type="0" start="-2332-01-01" code="dangi"/> <!-- 'sequential' year -->
           </eras>
         </calendar>
         <calendar type="ethiopic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="8-08-28"/>
-              <era type="1" start="8-08-29"/>
+              <era type="0" end="8-08-28" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
+              <era type="1" start="8-08-29" code="ethiopic" aliases="incar"/>
           </eras>
         </calendar>
         <calendar type="ethiopic-amete-alem">
           <eras>
-              <era type="0" end="-5492-08-29"/>
+              <era type="0" end="-5492-08-29" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
           </eras>
         </calendar>
         <calendar type="indian">
           <eras>
-              <era type="0" start="79-01-01"/>
+              <era type="0" start="79-01-01" code="indian" aliases="saka"/>
           </eras>
         </calendar>
         <calendar type="roc">
           <eras>
-              <era type="0" end="1911-12-31"/>
-              <era type="1" start="1912-01-01"/>
+              <era type="0" end="1911-12-31" code="roc-inverse" aliases="before-roc"/>
+              <era type="1" start="1912-01-01" code="roc" aliases="minguo"/>
           </eras>
         </calendar>
 
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDtdData.java
@@ -492,7 +492,7 @@ public class TestDtdData extends TestFmwk {
                 || (elementName.equals("matchVariable") && attribute.equals("id"))
                 || attribute.equals("iso4217")
                 || attribute.equals("iso3166")
-                || attribute.equals("code")
+                || attribute.equals("code") && elementName.equals("telephoneCountryCode")
                 || (attribute.equals("type") && !elementName.equals("calendarSystem") && !elementName.equals("mapZone")
                     && !elementName.equals("numberingSystem") && !elementName.equals("variable"))
                 || attribute.equals("id") && elementName.equals("variable")


Manishearth (dismissed):

common/supplemental/supplementalData.xml
```diff
@@ -4402,342 +4402,344 @@ XXX Code for transations where no currency is involved
...
-              <era type="0" end="8-08-28"/>
-              <era type="1" start="8-08-29"/>
+              <era type="0" end="8-08-28" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
+              <era type="1" start="8-08-29" code="ethiopic" aliases="incar"/>
           </eras>
         </calendar>
         <calendar type="ethiopic-amete-alem">
           <eras>
-              <era type="0" end="-5492-08-29"/>
+              <era type="0" end="-5492-08-29" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
```
FrankYFTang, line 4731: the type in this calendar is "ethiopic-amete-alem" should the code in here be also "ethiopic-amete-alem" and the aliases be "ethioaa mundi" instead?

FrankYFTang (commented):
about consistency between code of era and type of calendar

common/supplemental/supplementalData.xml
FrankYFTang: I understand this now.

FrankYFTang:
We may also need to amend https://www.unicode.org/reports/tr35/tr35-dates.html#Calendar_Data

FrankYFTang:
Should we also change https://github.com/unicode-org/cldr/blob/main/docs/ldml/tr35-dates.md#Calendar_Data in this PR?

common/supplemental/supplementalData.xml
```diff
@@ -4402,342 +4402,344 @@ XXX Code for transations where no currency is involved
...
+              <era type="227" start="1854-11-27" code="ansei-1854"/>
+              <era type="228" start="1860-3-18" code="manen-1860"/>
+              <era type="229" start="1861-2-19" code="bunkyu-1861"/>
+              <era type="230" start="1864-2-20" code="genji-1864"/>
+              <era type="231" start="1865-4-7" code="keio-1865"/>
+              <era type="232" start="1868-9-8" code="meiji" aliases="m"/>
+              <era type="233" start="1912-7-30" code="taisho" aliases="t"/>
+              <era type="234" start="1926-12-25" code="showa" aliases="s"/>
+              <era type="235" start="1989-1-8" code="heisei" aliases="h"/>
+              <era type="236" start="2019-5-1" code="reiwa" aliases="r"/>
```
sffc, line 4652: @justingrant suggests removing the aliases here. I agree.

macchiati: We're also discussed a different approach for the codes.

macchiati (changes requested):

common/supplemental/supplementalData.xml
sffc: Yes. For context, the suggestion was to use ordinals in insertion order with post-meiji elision, resulting in era codes such as `showa-001` for "the first Showa era in the period before Meiji". I floated the idea by @Manishearth, who reminded me of his position from when we previously discussed this last year, which is that `showa-001` doesn't tell you anything about the era: was it in the first or second millennium? Before or after the Edo Period?

The counter-proposal is to use `showa-131x`. I'm currently waiting on feedback from @yumaoka.

common/dtd/ldmlSupplemental.dtd
```diff
@@ -250,9 +250,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST era end CDATA #IMPLIED >
     <!--@MATCH:time/yyyy-MM-dd-->
     <!--@VALUE-->
-<!ATTLIST era code NMTOKEN #REQUIRED >
+<!ATTLIST era code NMTOKEN #IMPLIED >
     <!--@MATCH:regex/[a-z0-9]{3,8}(\-[a-z0-9]{3,8})*-->
-    <!--@VALUE-->
```
macchiati: Don't remove the @VALUE. That marks that the era code is not metadata and not distinguishing, and instead is additional data about the type. Just like the start/end data

common/supplemental/supplementalData.xml
```diff
@@ -4411,245 +4411,245 @@ XXX Code for transations where no currency is involved
...
-              <era type="228" start="1860-3-18" code="manen-1860"/>
-              <era type="229" start="1861-2-19" code="bunkyu-1861"/>
-              <era type="230" start="1864-2-20" code="genji-1864"/>
-              <era type="231" start="1865-4-7" code="keio-1865"/>
-              <era type="232" start="1868-9-8" code="meiji" aliases="m"/>
-              <era type="233" start="1912-7-30" code="taisho" aliases="t"/>
-              <era type="234" start="1926-12-25" code="showa" aliases="s"/>
-              <era type="235" start="1989-1-8" code="heisei" aliases="h"/>
-              <era type="236" start="2019-5-1" code="reiwa" aliases="r"/>
+              <era type="-2" start="645-6-19" code="japanese-inverse" aliases="gregory-inverse bc bce"/>
```
macchiati, line 4414: Question; wouldn't the japanese-inverse end on that date (or actually the day before)?

macchiati (changes requested):

common/dtd/ldmlSupplemental.dtd
sffc: I tried, but for some reason if I have the `@VALUE`, I get the following error. I DON'T get this error if I remove it.

> Error: (TestDtdData.java:313) : isDistinguished supplementalData: elementName.equals("era") && attribute.equals("code"): expected true, got false

Manishearth (approved):

common/dtd/ldmlSupplemental.dtd
macchiati: I'll have to look it over tomorrow morning

common/supplemental/supplementalData.xml
macchiati: ?

macchiati (changes requested):
It is clear that eras with different start/end dates must have different codes. For those that have identical start/end dates, **but different fundamental semantics**, we also need different era codes. However, the following don't make sense to me:

```
<era type="0" start="622-7-15" code="islamic" aliases="ah"/>
<era type="0" start="622-7-15" code="islamic-rgsa" aliases="ah"/>
<era type="0" start="622-7-15" code="islamic-tbla" aliases="ah"/>
<era type="0" start="622-7-15" code="islamic-umalqura" aliases="ah"/>

```
I believe these should all be:
<era type="0" start="622-7-15" code="islamic" aliases="ah"/>

macchiati (changes requested):
(see other comment)

common/supplemental/supplementalData.xml
sffc: Fixed.

common/dtd/ldmlSupplemental.dtd
sffc: Thanks @macchiati for your investigation! It is working now

justingrant:
> eras with different start/end dates must have different codes. For those that have identical start/end dates, but different fundamental semantics, we also need different era codes.

Aren't era names scoped to each calendar?  If not&mdash;meaning we're expecting any era to be used in any calendar&mdash;then it seems like we're adding an unnecessary future compatibility risk.  What is the benefit of allowing era codes to be free-standing entities, as opposed to being scoped to each calendar like AFAICT real-world eras are?

common/supplemental/supplementalData.xml
sffc: The codes for pre-Meiji eras were removed in order to allow for more time to investigate the quality of the data.

sffc:
I still need to write up the spec, but the understanding is that era codes are globally unique, and era aliases are scoped to the calendar. This allows for all the flavors of Islamic to use an "ah" era alias, as well as for Gregorian, Julian, and Juliogregorian to all use "ad" and "bc".

@macchiati I'm not sure what else is missing? The four Islamic eras have different codes and shared aliases.

sffc:
@macchiati Are you suggesting that the four Islamic eras all share the same era code? I gave them different codes primarily because they have different behavior.

macchiati:
All of those calendars are closely related variants of the islamic
calendar, and share the fundamental semantics and start/end dates. I don't
see the value in having separate codes; only user confusion.

sffc:
@macchiati My [proposal](https://github.com/tc39/proposal-intl-era-monthcode/issues/9) suggests that an era code defines two things:

> 1. An epoch day
> 2. A formula for converting from the (eraYear, monthCode, day) triple to a number of days relative to that epoch

The reasons for constraint 2 are:

1. Allows for eras with the same epoch day to coexist, like `gregory` and `julian` may need to do in Juliogregory.
2. Allows for unique identifiers during date conversion.

Example to illustrate point 2: suppose that a user keeps dates in the Umm al-Qura for bookkeeping purposes, but wants to render them in the Tabular Islamic calendar based on moon sightings. Having separate codes for the two eras allows for the following code to be written: `{ calendar: "islamic-tbla", era: "islamic-umalqura", eraYear: 1444, monthCode: "M07", day: 12 }`. This is _similar_ to how a date could be expressed in the `ethioaa` era but then converted at the call site to the `ethiopic` one.

In addition to this, each calendar has an era of the same name as the calendar so that it can be the anchor era for the relativeYear / arithmetic year.

macchiati:
#2 seems very hacky to me, convoluting the era names just for a slightly
more compact form in some cases than explicitly including the calendar:

The full description appears to be that you want to convert from
{calendar:"islamic-umalqura" era: "islamic", eraYear: 1444, monthCode:
"M07", day: 12} to calendar: "islamic-tbla"
Just like I might want to convert from
{calendar:"japanese" era: "showa", eraYear: 3, monthCode: "M07", day: 12}
 to calendar: "islamic-tbla"

Because the eras are often distinct for calendars, you are forcing all of
them to be fully distinct so you can save a parameter. But although I find
it screwy, I won't stand in the way.

justingrant:
> suppose that a user keeps dates in the Umm al-Qura for bookkeeping purposes, but wants to render them in the Tabular Islamic calendar based on moon sightings. Having separate codes for the two eras allows for the following code to be written: `{ calendar: "islamic-tbla", era: "islamic-umalqura", eraYear: 1444, monthCode: "M07", day: 12 }`.

I'm not sure I understand this example. My understanding was that the way to move dates between calendars was to use `withCalendar()`, like this:

```js
date = Temporal.PlainDate.from({ calendar: "islamic-umalqura", year: 1444, monthCode: "M07", day: 12 });
tabluarDate = date.withCalendar('islamic-tbla');

date = Temporal.PlainDate.from({ calendar: "japanese", eraYear: 10, era: 'meiji', monthCode: "M07", day: 12 });
gregorianDate = date.witihCalendar('gregory');
```

I think the core issue here is that it's confusing to apply an era from one calendar to a different calendar where that era isn't used.  This confuses the concepts of calendars and eras. These concepts are _already_ confusing on their own. Mixing them up makes it worse. 

I think this is even more confusing in single-era calendars like Islamic calendars, Chinese, Buddhist, etc. where eras aren't generally used in normal life.  In those cases, the era and the calendar are essentially the same thing, which makes it confusing when they're mixed. And even more confusing if the (single) era has the same name as the calendar. 

That said, it's an important question how to define and name eras in calendars that _do_ use eras, like:
* Japanese which (unlike all other CLDR calendars) there's no obvious "primary era" to pick, so we want to use AD as the anchor era which determines `year`
* Ethiopian calendars that are only differentiated by eras
* Gregorian and Julian-Gregorian, which will likely share same-named eras.

Some calendars like the ones above may share same-named and/or same-semantics eras with other calendars. But these special cases don't require a generalized ability to use any era across to any calendar.  I would argue that this kind of many-to-many mapping between calendars and eras makes both concepts harder to understand.

Overall, I think that eras will be easier to explain to programmers if:
* Eras are scoped to each calendar, so there's no expectation they can be used in calendars where they don't belong.
* Different calendars may have eras with the same name, without this being a requirement for all eras to be shared across all calendars
* Each era's canonical name (the name that shows up in the debugger) is a name like `bc` or `bce` that's familiar to users of that calendar, instead of making up names like `gregory-inverse` which require a trip to the docs for programmers to understand and that doesn't improve discoverability because its name simply matches the calendar's name.

Note that my concerns are about the public API, not CLDR data. If there's some reason for the two to differ, that's OK with me, as long as it doesn't make it harder to have an understandable public API.

common/supplemental/supplementalData.xml
```diff
@@ -4643,101 +4645,101 @@ XXX Code for transations where no currency is involved
...
           <eras>
-              <era type="0" start="79-01-01"/>
+              <era type="0" start="79-01-01" code="indian" aliases="saka"/>
           </eras>
         </calendar>
         <calendar type="roc">
           <eras>
-              <era type="0" end="1911-12-31"/>
-              <era type="1" start="1912-01-01"/>
+              <era type="0" end="1911-12-31" code="roc-inverse" aliases="broc"/>
```
FrankYFTang, line 4741: I suggest you keep it simple and remove aliases="broc" . It is not a known code and there are no need for such alias.

sffc: When writing this PR, I adopted all narrow abbreviations in the root and english locales as aliases.

FrankYFTang: but four letter "broc" is nowhere to be found in CLDR files anywhere now. How is that related to "all narrow abbreviations in the root and english locales" ?  Which file has that?

I did a 
```
find|xargs egrep -s   broc
```
but none of them are broc as a token and related to a era.

Manishearth: Yes, but I'm not 100% sure if that is attested. As Frank found there is definitely 民國前 in the wild, but I'm not sure about "B. ROC." or similar.

A thing I've repeatedly come across when it comes to CLDR non-gregorian calendar data is that a lot of the symbols aren't necessarily attested and may just be made up.

sffc: https://github.com/unicode-org/cldr/blob/abdaabaef38362e3cef2b20b4938c0a5f16cd902/common/main/en.xml#L2825

```xml
			<calendar type="roc">
				<eras>
					<eraAbbr>
						<era type="0">B.R.O.C.</era>
						<era type="1">Minguo</era>
					</eraAbbr>
				</eras>
			</calendar>
```

FrankYFTang: to be clear. I am not asking you to change to a different alias for "roc-inverse" I am just saying there are no need for any alias for it. Keep it empty until someone have a strong case to lobby to add one (I do NOT).

sffc: OK, removed.

FrankYFTang: oh I see. I suggest you to use
"before-roc" instead. At least that is already used in several  API

Java https://docs.oracle.com/javase/8/docs/api/java/time/chrono/MinguoEra.html
Android https://developer.android.com/reference/java/time/chrono/MinguoEra#BEFORE_ROC
Joda Time https://cljdoc.org/d/clojure-interop/java.time/1.0.5/api/jdk.time.chrono.MinguoEra#BEFORE_ROC
HarmonyOS https://developer.harmonyos.com/en/docs/documentation/doc-references/minguoera-0000001054518588#ZH-CN_TOPIC_0000001054518588__BEFORE_ROC
Three Ten https://www.threeten.org/threetenbp/apidocs/org/threeten/bp/chrono/MinguoEra.html#BEFORE_ROC

FrankYFTang: I am also fine if you do not add "before-roc"

sffc: Added "before-roc" as an alias.

macchiati (approved):

macchiati:
@sffc if you don't have merge rights, let me know.

FrankYFTang (approved):
Thanks for working on this!!!

FrankYFTang:
I think he need to squashed it first.

FrankYFTang (approved):

robertbastian:
I agree with Mark's and Justin's sentiment that calendars should be able to share era names. In particular

* Why do the islamic calendars not all call their era `islamic`? They all allow calling it `ah`, so the calendarless validation that Shane seems to want to do does not always work.
* Why does the Japanese calendar call its Gregorian eras `japanese-inverse` and `japanese` instead of `gregory-inverse` and `gregory`? "The Japanese era" is not a thing that exists, these names seem to also have been chosen for the purpose of guessing calendar from era.

Both of these instances break the property that an identifier is either an era name, or an alias to another era name: in the Islamic case because `ah` is an alias to four different era names, in the Japanese case because `gregory` is an alias to `japanese` in the Japanese calendar, but an era in the Gregorian calendar. This makes it impossible to normalise an era name without knowing the calendar to normalize for.

I don't understand why this ticket introduces the requirement that era codes need to be unique across calendars. The previous era numbers certainly weren't.

sffc:
With the landed CLDR proposal, calendars _can_ share era names. It's only when you "canonicalize" an era that it produces a code that is globally unique. I stated my original motivation in https://github.com/unicode-org/cldr/pull/2665#issuecomment-1414577550, but I acknowledge the points in this thread and agree that there's likely more iteration that could be done. One suggestion I support is being proposed by @ptomato in https://github.com/tc39/proposal-intl-era-monthcode/pull/20

PR #2672, opened 2023-01-30
CLDR-16122 v43 BRS: skip golden value PN tests

srl295:
CLDR-16122

- skip some tests in TestPersonNameFormatter that are too data dependent
- reference CLDR-16345 to fix these later

ALLOW_MANY_COMMITS=true


macchiati:
I was working on this at the same time. I did a bit of surgery to make them
less data-dependent, though it keeps (and adjusts) the japanese & thai
cases.

PR #2675, opened 2023-01-31
CLDR-16294 Add 986 chars from TGH-2013 Level 1 to zh aux exemplars

pedberg-icu:
CLDR-16294

Add 986 chars from TGH-2013 Level 1 to the 479 chars already in zh aux exemplars.

ALLOW_MANY_COMMITS=true


markusicu (approved):
rslgtm
FYI @kenlunde

PR #2676, opened 2023-01-31
CLDR-16308 DAIP Latin-script intervalFormat adjustments should handle preceding literals

pedberg-icu:
CLDR-16308

Adjusted DisplayAndInputProcessor (DAIP) so that the intervalFormat normalizations it performs for Latin-script locales (replacing hyphen or en-dash surrounded by space with en-dash \u2013 surrounded by thin space \u2009) work even if those characters are preceded by literal text such as a period or a quoted literal. It turn out that when DateTimePatternGenerator.FormatParser.getItems returns the tokens for the first part of the intervalFormat pattern, the last token - which ends with the spaces + hyphen/dash - also includes any preceding literal text which came after the last pattern character in the first part.

Then ran the updated DAIP to actually fix the problem intervalFormats. The differences are just changing regular space to thin space around an en-dash, to be consistent with the other intervalFormats in each of these locales.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/bs.xml
+++ b/common/main/bs.xml
@@ -4682,7 +4682,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 							<greatestDifference id="m">h:mm – h:mm B</greatestDifference>
 						</intervalFormatItem>
 						<intervalFormatItem id="d">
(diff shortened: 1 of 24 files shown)


markusicu (approved):
lgtm tnx

PR #2679, opened 2023-01-31
CLDR-14884 change ajp_JO to apc_SY

DavidLRowe:
CLDR-14884

ALLOW_MANY_COMMITS=true


srl295 (approved):
Looks good.
A question, were the likely and default content lines manual or automatic generation? 

Manual is probably fine for now but will need to be sorted out eventually.

DavidLRowe:
The values in
"likelySubtags" element in likelySubtags.xml
"languageData" element in supplementalData.xml
"territoryInfo" element in supplementalData.xml
were changed manually, which IIUC is how to do it, as of v42 cycle.

DavidLRowe:
The following entry was already in languageInfo.xml file and I think it satisfies the request that this locale (originally agp, now apc) fall back to standard Arabic.
`<languageMatch desired="apc" supported="ar" distance="10" oneway="true"/>	<!-- North Levantine Arabic -->`

PR #2687, opened 2023-02-01
CLDR-16320 hi_Latn: drop compact numbers

srl295:
- keep draft status but change to triple up arrow

CLDR-16320

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/hi_Latn.xml
+++ b/common/main/hi_Latn.xml
@@ -5361,58 +5361,58 @@ annotations.
 			</decimalFormatLength>
 			<decimalFormatLength type="long">
 				<decimalFormat>
-					<pattern type="1000" count="one">0 hazaar</pattern>
-					<pattern type="1000" count="other">0 hazaar</pattern>
-					<pattern type="10000" count="one">00 hazaar</pattern>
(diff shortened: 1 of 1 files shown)


common/main/hi_Latn.xml
```diff
@@ -5361,58 +5361,58 @@ annotations.
 			</decimalFormatLength>
 			<decimalFormatLength type="long">
 				<decimalFormat>
```
btangmu, line 5363: do these paths under "long" count as "compact"? I imagine they do given the ticket name... I'm ignorant about what "compact" means here...

btangmu (approved):
looks fine except for my ignorance about how "compact" relates to "long" and "short" here

common/main/hi_Latn.xml
srl295: that's a good question.  I think they are still compact in that they aren't spelled out.   (`123 thousand`,  not `one hundred twenty three thousand`) .  Maybe get @macchiati  and @AEApple comment?

AEApple: Yes, it's that they aren't spelled out as a number so 1 million would be the compact number form vs. 1,000,000.

PR #2688, opened 2023-02-01
CLDR-16323 Update code to handle patterns without given, fix charts

macchiati:
CLDR-16323

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonNames.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonNames.java
@@ -15,7 +15,7 @@ public class ChartPersonNames extends Chart {
     private static final Factory CLDR_FACTORY = CLDRConfig.getInstance().getCldrFactory();
     static final String MAIN_HEADER = "<p>These charts shows the sample person names for different locales, formatted according to the locale's pattern.</p>";
     private static final boolean DEBUG = false;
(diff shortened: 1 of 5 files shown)


macchiati:
Including @richgillam so he can see the changes to the person name formatting. Turned out to be pretty straightforward.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonNames.java
```diff
@@ -15,7 +15,7 @@ public class ChartPersonNames extends Chart {
     private static final Factory CLDR_FACTORY = CLDRConfig.getInstance().getCldrFactory();
     static final String MAIN_HEADER = "<p>These charts shows the sample person names for different locales, formatted according to the locale's pattern.</p>";
     private static final boolean DEBUG = false;
-    static final String DIR = CLDRPaths.CHART_DIRECTORY + "personNames/";
+    static final String DIR = CLDRPaths.CHART_DIRECTORY + "verify/personNames/";
```
srl295, line 18: what is `verify/` ?

srl295 (approved):
looks reasonable with a question…

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartPersonNames.java
macchiati: It is one of the chart subfolders.

richgillam:
The changes to the person name formatter look good to me.

PR #2692, opened 2023-02-01
CLDR-15071 update to langtags.json

DavidLRowe:
CLDR-15071

This commit is only to update the langtags.json submission made in 2021 with newer data.

ALLOW_MANY_COMMITS=true


DavidLRowe:
Hopefully it won't be hard to verify that this new data doesn't cause you any problems. Most of the changes are to fields (such as language names) that I don't think you use, but there are a few changes in tags where we have newer information.

You always have to the option to deny this PR and proceed with the older data.

macchiati (approved):

PR #2694, opened 2023-02-02
CLDR-16323 Fix patterns with given-initial

macchiati:
CLDR-16323

The previous PR fixed most problems, but one remained.

That code handles the case where the name data is a monogram (no surname) and there's no {given} in the pattern. It does that by using a filtered NameObject that returns the given name instead of the surname. 

1. That mostly works. Where it doesn't is in cases of "Z" in the nativeG column on https://unicode-org.github.io/cldr-staging/charts/43/verify/personNames/en_GB.html, where we would not expect just an initial (if there were a surname, it would be a full name).
2. The problem comes when we have a monogram, and the pattern is {given-initial} {surname} (perhaps with additional items like {title}. We get just "Z" (or "Z." in https://unicode-org.github.io/cldr-staging/charts/43/verify/personNames/en.html).
3. We could of course address this by adding extra patterns in the data. However, we'd have to add it everywhere for a very common case. And the extra patterns are so very cumbersome for vetters (and us).

My proposal is: 

In the filtered name object, change it to return null for 'given'. That has no effect on 2.1 above.
We then add an additional condition:

- where the name data is a monogram (no surname) and there's no {given} or {given-initial} in the pattern

Spot-checking, this fixed the problems with Z (or Z.)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -659,7 +659,7 @@ public class PersonNameFormatter {
             StringBuilder literalTextAfter = new StringBuilder();
 
             // Check that we either have a given value in the pattern or a surname value in the name object
-            if (!getFields().contains(Field.given) && !nameObject.getAvailableFields().contains(Field.surname)) {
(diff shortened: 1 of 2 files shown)


richgillam (approved):
Yeah, this seems reasonable.

richgillam (approved):
Test change looks fine too.

PR #2697, opened 2023-02-06
CLDR-15995 fixes to charts

macchiati:
CLDR-15995

1. Modifies GenerateSidewaysView and ShowData to avoid generating files too large for github.
2. Modifies ShowLocaleCoverage to clean up the Key, drop the column that showed common vs seed, added links to the tsv files.
3. Fixed a missing index line (that effectively hid all the person name charts.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLocaleCoverage.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLocaleCoverage.java
@@ -678,7 +678,7 @@ public class ShowLocaleCoverage {
                 + "The Core values are described on <a target='_blank' href='https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales'>Core Data</a>. "
                 + "</td></tr>\n"
                 + "<tr><th>Missing Features</th><td>These are not single items, but rather specific features, such as plural rules or unit grammar info. "
-                + "They are listed if missing at the computed level.<br>"
+                + "They are listed if missing at the computed level. For more information, see <a href='https://cldr.unicode.org/index/locale-coverage'>Missing Features</a><br>"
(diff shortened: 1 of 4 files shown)


AEApple:
Where is the best place to preview the updated charts?

macchiati:
Directly at https://unicode-org.github.io/cldr-staging/charts/43/ (They are linked from https://cldr.unicode.org/index/downloads/cldr-43)

Two main files affected are:

1. https://unicode-org.github.io/cldr-staging/charts/43/supplemental/locale_coverage.html
2. https://unicode-org.github.io/cldr-staging/charts/43/verify/index.html

Also switched the generation to CLDR v43.0β for all charts.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLocaleCoverage.java
```diff
@@ -76,6 +76,9 @@
 import com.ibm.icu.util.VersionInfo;
 
 public class ShowLocaleCoverage {
+
+    private static final String TSV_BASE = "https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/43/tsv/";
+
```
pedberg-icu, line 81: For the future, can we get the "43" from a constant somewhere else (or create such if it does not exist)?

pedberg-icu (approved):

AEApple (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLocaleCoverage.java
macchiati: Yes, I'll include that in another chart ticket.

PR #2295, opened 2022-08-18
CLDR-15939 Amharic Gemination Support

dyacob:
CLDR-15939

This is a revised and current branch and PR that replaces: https://github.com/unicode-org/cldr/pull/1674 . This PR completes the new ticket [CLDR-15939](https://unicode-org.atlassian.net/browse/CLDR-15939).

Note that a previous PR (https://github.com/unicode-org/cldr/pull/2284) was created for this branch, the branch itself was _**renamed**_ and a new PR created.  Hopefully not leading to any process weirdness.

- [X] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

[CLDR-15939]: https://unicode-org.atlassian.net/browse/CLDR-15939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2701, opened 2023-02-07
CLDR-15939 Amharic Gemination Support

btangmu:
CLDR-15939

ALLOW_MANY_COMMITS=true


btangmu:
This is intended to be identical to https://github.com/unicode-org/cldr/pull/2295 but squashed and free from commit-message problems.

pedberg-icu (approved):

pedberg-icu:
Went ahead and merged...

PR #2702, opened 2023-02-08
CLDR-15071 Fix dtd tests

macchiati:
CLDR-15071

I noticed that the tests were not catching a mismatch between the sil1 value in the .xml file and the sil22 value in the dtd. I looked into it, and found some problems in the test, where paths not being tested correctly. Fixed those, and the cases in the DTD that they exposed.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -68,6 +68,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@METADATA-->
     <!--@DEPRECATED:true, false-->
 <!ATTLIST language references CDATA #IMPLIED >
+    <!--@MATCH:any-->
     <!--@METADATA-->
 
 <!ELEMENT script ( #PCDATA ) >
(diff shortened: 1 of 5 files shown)


common/dtd/ldml.dtd
```diff
@@ -2825,10 +2826,11 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT cr ( #PCDATA ) >
 <!ATTLIST cr alt NMTOKENS #IMPLIED >
-    <!--@MATCH:literal/variant-->
+    <!--@MATCH:literal/variant, proposed-->
```
srl295, line 2829: ouch… that's not recent

srl295 (approved):
good fix…

common/dtd/ldml.dtd
macchiati: Yes, the changes in the test exposed some gaps.

srl295 (approved):

srl295 (approved):

PR #2703, opened 2023-02-08
CLDR-16239 replace lateral inheritance markers in trunk

macchiati:
CLDR-16239

I would suggest looking at the CLDRModify changes first, then spot-checking some of the generated files. Note that "constructed" is the path returned for fallback constructed language items, like en_US.

Some files have many changes, so you might prefer to use the magic . option in the Files view to use the other diff mechanisms.

The key issues are:
1. nothing but inheritance markers (↑↑↑) should be affected (code inspection should be sufficient for this)
2. only markers that inherit laterally should be affected (also code inspection), but with sanity check.
3. constructed values (only languages with types having _) are not affected
4. ↑↑↑ may be left in where they inherit from root, with no lateral.
5. L2+ locales (eg fr_CA) won't be affected. This is a bit of a short-cut; it depends on the parent locale's being complete. I think that is enough for this release, but we might revisit.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/be.xml
+++ b/common/main/be.xml
@@ -835,7 +835,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<territory type="TN">Туніс</territory>
 			<territory type="TO">Тонга</territory>
 			<territory type="TR">Турцыя</territory>
-			<territory type="TR" alt="variant">↑↑↑</territory>
+			<territory type="TR" alt="variant">Турцыя</territory>
 			<territory type="TT">Трынідад і Табага</territory>
 			<territory type="TV">Тувалу</territory>
 			<territory type="TW">Тайвань</territory>
@@ -1014,13 +1014,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<calendar type="buddhist">
 				<eras>
 					<eraNames>
-						<era type="0">↑↑↑</era>
+						<era type="0">б.э.</era>
 					</eraNames>
 					<eraAbbr>
 						<era type="0">б.э.</era>
 					</eraAbbr>
 					<eraNarrow>
-						<era type="0">↑↑↑</era>
+						<era type="0">б.э.</era>
 					</eraNarrow>
 				</eras>
 				<dateFormats>
@@ -1092,18 +1092,18 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 							<month type="12">12</month>
 						</monthWidth>
 						<monthWidth type="narrow">
-							<month type="1">↑↑↑</month>
-							<month type="2">↑↑↑</month>
-							<month type="3">↑↑↑</month>
-							<month type="4">↑↑↑</month>
-							<month type="5">↑↑↑</month>
-							<month type="6">↑↑↑</month>
-							<month type="7">↑↑↑</month>
-							<month type="8">↑↑↑</month>
-							<month type="9">↑↑↑</month>
-							<month type="10">↑↑↑</month>
-							<month type="11">↑↑↑</month>
-							<month type="12">↑↑↑</month>
+							<month type="1">1</month>
+							<month type="2">2</month>
+							<month type="3">3</month>
+							<month type="4">4</month>
+							<month type="5">5</month>
+							<month type="6">6</month>
+							<month type="7">7</month>
+							<month type="8">8</month>
+							<month type="9">9</month>
+							<month type="10">10</month>
+							<month type="11">11</month>
+							<month type="12">12</month>
 						</monthWidth>
 					</monthContext>
 					<monthContext type="stand-alone">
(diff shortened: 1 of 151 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2402,6 +2431,32 @@ public void handleEnd() {
                 }
             }
         });
+
+        fixList.add('V', "Fix values that would inherit laterally", new CLDRFilter() {
+            boolean skip;
+            @Override
+            public void handleStart() {
+                skip = !XMLSource.ROOT_ID.equals(LocaleIDParser.getParent(getLocaleID()));
```
srl295, line 2439:
```suggestion
                // skip if not an immediate child of root
                skip = !XMLSource.ROOT_ID.equals(LocaleIDParser.getParent(getLocaleID()));
```

srl295 (dismissed):

macchiati:
Note: fixed errors exposed by tests. I retracted the bs_Cyrl changes because there was a problem with logical groups. I'd like to unblock by getting the rest of the changes in first, then do a second PR for bs_Cyrl

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
macchiati: added in latest commit

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2402,6 +2431,33 @@ public void handleEnd() {
...
+                    return;
+                }
+                String value = cldrFileToFilter.getStringValue(xpath);
+                if (CldrUtility.INHERITANCE_MARKER.equals(value)) {
+                    Output<String> pathWhereFound = new Output<>();
+                    Output<String> localeWhereFound = new Output<>();
+                    String baileyValue = getResolved().getBaileyValue(xpath, pathWhereFound, localeWhereFound);
+                    if (baileyValue != null
+                        && !xpath.equals(pathWhereFound.value)
+                        && !"constructed".equals(pathWhereFound.value)) {
```
btangmu, line 2454: instead of hard-coded "constructed" this really should be GlossonymConstructor.PSEUDO_PATH

GlossonymConstructor.java:
    public static final String PSEUDO_PATH = "constructed";

btangmu (commented):
The Java addition looks fine (and harmless unless -V is actually run)

From a quick look at the xml it seems to be all ↑↑↑ replaced by hard values as expected

Whether the resolved values would be the same, I have no idea; before we commit this PR, we should test it, right? I hope to do this shortly per email: “given two directories (main and production), verify that the corresponding fully resolved CLDRFiles for each locale are identical (and listing differences if they are not)”

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
macchiati: I think one problem we have is that these magic constants are scattered around, making it difficult to find them (or know that they are there). And that includes utilities like LocaleIDParser.getParent (I always have to search for the one class that has the right method. CLDRUtility, no, LanguageTagParser, no, CLDRLocale, no ...). We might want to have a techdebt cleanup for those little annoyances.

macchiati:
I agree with Tom. I think we need to wait on the comparison tool before proceeding with this.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
srl295: It's definitely an issue.  I guess my point is that the clients who will encounter the constant are working at the level of CLDRLocale.getBaileyValue() so a 'higher level' location for the constant (CLDRUtility would make sense).  Most clients who encounter this value aren't going to call GlossonymConstructor directly.

srl295: All of the XPath related constants should go in one place

common/main/be.xml
```diff
@@ -835,7 +835,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<territory type="TN">Туніс</territory>
 			<territory type="TO">Тонга</territory>
 			<territory type="TR">Турцыя</territory>
-			<territory type="TR" alt="variant">↑↑↑</territory>
+			<territory type="TR" alt="variant">Турцыя</territory>
```
btangmu, line 838: This is strange since the resulting file be.xml has Турцыя twice, as non-alt and as alt

btangmu: When I run "modify -fF" I get:

Resolved values differ: vxml [Турцыя] differs from baseline [Турэччына] for be_TARASK //ldml/localeDisplayNames/territories/territory[@type="TR"][@alt="variant"]

See https://github.com/unicode-org/cldr/pull/2708

macchiati:
see comments on that PR

macchiati:
The alt values will often be the same as the unaltered value, if the
language doesn't have the distinction that English does.

common/main/be.xml
```diff
@@ -10147,126 +10147,126 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
 			</unit>
 			<unit type="graphics-pixel-per-inch">
-				<displayName>↑↑↑</displayName>
-				<unitPattern count="one">↑↑↑</unitPattern>
-				<unitPattern count="few">↑↑↑</unitPattern>
-				<unitPattern count="many">↑↑↑</unitPattern>
-				<unitPattern count="other">↑↑↑</unitPattern>
+				<displayName>піксель/цалю</displayName>
+				<unitPattern count="one">{0} піксель/цалю</unitPattern>
+				<unitPattern count="few">{0} пікселі/цалю</unitPattern>
```
btangmu, line 10208: "modify -fF" reports:

Resolved values differ: vxml [{0} пікселі/цалю] differs from baseline [{0} піксэлі/цалю] for be_TARASK //ldml/units/unitLength[@type="narrow"]/unit[@type="graphics-pixel-per-inch"]/unitPattern[@count="few"]

Note, the difference is "е" versus "э"... replacement of soft with hard value here evidently changes what gets inherited by be_TARASK

```
$ egrep "піксэлі/цалю" common/main/be_TARASK.xml
				<unitPattern count="few" draft="provisional">{0} піксэлі/цалю</unitPattern>
```

btangmu: if it's already hard-coded in be_TARASK.xml, how can inheritance make any difference?

btangmu: likely answer to my own question above, is that it's already hard-coded for at least one path but other paths in the same child locale inherit from it laterally; but if ^^^ changes to hard value in the parent then the child inherits vertically instead of laterally

btangmu:
@macchiati a possible next step is to change approx. 1951 ^^^ (or missing) to hard values in the child locales (hard values being fully-resolved common/main) (See comments in https://github.com/unicode-org/cldr/pull/2711 and see the zip file attached to this ticket for the complete output.) If that change leads in turn to side-effects in the grandchildren, then also change some ^^^ (or missing) to hard values in the grandchildren, and so forth. The fully-resolved values in all cases would be based on what is now in the main branch common/main. The end result would be data with more hard values yet producing the same fully-resolved values.

PR #2704, opened 2023-02-08
CLDR-16284 Remove Thai from no-space name scripts

macchiati:
CLDR-16284

Removing Thai from the list of scripts that don't need spaces in names.

ALLOW_MANY_COMMITS=true


richgillam (approved):
LGTM...

pedberg-icu:
One of the commits has the wrong JIRA ticket number, maybe this can be fixed when you squash?

pedberg-icu (approved):

btangmu (approved):

richgillam (approved):
I take it you're getting the list of scripts that don't use spaces from some other resources in CLDR and then modifying it to exclude Thai?  Because if this is the only place where that information lives, you could just take Thai out of the list and be done with it.

macchiati:
Right.

PR #2706, opened 2023-02-09
CLDR-16377 update CLDRFileTransformer to overwrite in place

srl295:
- don't write to Generated, but update files - if actual changes
- update TempPrintWriter to support this

CLDR-16377

(supports CLDR-14120 )

ALLOW_MANY_COMMITS=true


macchiati (approved):
I think this is safe right now. (I was looking to see if the order of evaluation mattered)

macchiati:
It should be safe, so you can merge

PR #2710, opened 2023-02-10
CLDR-16346 Automatically handle adjacent duplicate literals in person names

macchiati:
CLDR-16346

common/main/ar.xml

- retracted the special pattern

- I had to understand what the code was doing, so went ahead and added comments for the future.
- The main change was to coalesceLiterals. It is not just duplicates, but rather anytime we would get a repeat at the end of a combined literal

- Added test cases

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
@@ -652,11 +652,27 @@ public class PersonNameFormatter {
 
         public String format(NameObject nameObject, FormatParameters nameFormatParameters, FallbackFormatter fallbackInfo) {
             StringBuilder result = new StringBuilder();
-            boolean seenLeadingField = false;
-            boolean seenEmptyLeadingField = false;
-            boolean seenEmptyField = false;
-            StringBuilder literalTextBefore = new StringBuilder();
-            StringBuilder literalTextAfter = new StringBuilder();
+            /*
+             * We have a series of literals and placeholders.
+             * • The literals are never "",
+             * • while the placeholders may have a value or may be missing (null).
+             * If we have a missing placeholder at the start, we discard the literal (if any) before it.
+             *      1. Effectively, that means that we don't append it to result until we've seen the following field.
+             * If we have a missing placeholder at the end, we discard a literal (if any) after it.
+             *      2. Effectively, that means that we don't append it to result if the last placeholder was missing.
+             * If we have adjacent missing placeholders, then we discard the literal between them.
+             *
+             * We also coalesce literals A and B. This can only happen if we had one or more empty placeholders:
+             *      3. If A.endsWith(B) then we discard it.
+             *      4. Any sequence of multiple whitespace is reduced to the first.
+             * The following booleans represent the state we are in:
+             */
+            boolean seenLeadingField = false; // set to true with not missing (so we have at least 1 non-missing field)
+            boolean seenEmptyLeadingField = false; // set to false with not missing; set to true with missing and !seenLeadingField
+            boolean seenEmptyField = false; // set to false with seenEmptyField & not missing; set to true with missing & seenLeadingField
+
+            StringBuilder literalTextBefore = new StringBuilder(); // literal right after a non-missing placeholder
+            StringBuilder literalTextAfter = new StringBuilder();  // literal right after a missing placeholder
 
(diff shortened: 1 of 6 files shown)


srl295:
The general theory makes sense.. Question, will it bring user surprise? Could `B.endsWith(A)` cause trouble if the literals are somehow words `"fred".endsWith("ed")`?

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -773,6 +793,21 @@ private String coalesceLiterals(StringBuilder l1, StringBuilder l2) {
...
         }
 
+        private boolean endsWith(StringBuilder l1, StringBuilder l2) {
+            final int l2Length = l2.length();
+            final int delta = l1.length() - l2Length;
+            if (delta < 0) {
+                return false;
+            }
+            for (int i = 0; i < l2Length; ++i) {
+                // don't have to worry about unpaired surrogates.
```
srl295, line 803: why, out of curiosity?

btangmu: DAIP does "normalize all input to NFC" so how could we get an unpaired surrogate?

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
srl295: denormalized would be a different issue, that would be something like `"é".endsWith("\u0301")`

btangmu: I was imagining that com.ibm.icu.text.Normalizer#compose() would never return a string containing an unpaired surrogate; I could be wrong...

btangmu: I was wrong! Just tested like this with an unpaired surrogate:
```
        String xValue = Normalizer.compose("\uDC04", false);
        System.out.println("xValue = " + xValue + " " + xValue.contains("\uDC04"));
```
[INFO] xValue = ? true

It may not be likely in practice, still it's worrisome ... we should add a check to DAIP for well-formed strings!

btangmu: it could be tested like this:
```
        CharsetEncoder encoder = Charset.forName("UTF-16BE").newEncoder();
        try {
            char[] data = value.toCharArray();
            encoder.encode(CharBuffer.wrap(data));
        } catch (Exception e) {
            internalException[0] = e;
            return null;
        }
```

macchiati: Right, as long as values are well-formed, it doesn't matter. We could file a ticket to raise an error for vetters for any value that was ill-formed. I wouldn't make it a high priority, and if we do this we can do it more efficiently than firing up an encoder, because the rules are so simple (ICU might have a method already).

srl295: OK so it's really:

```suggestion
                // not currently worrying about unpaired surrogates.
```

anyway, not a problem

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java
```diff
@@ -773,6 +793,21 @@ private String coalesceLiterals(StringBuilder l1, StringBuilder l2) {
...
+                if (l1.charAt(i+delta) != l2.charAt(i)) {
+                    return false;
+                }
+            }
+            return true;
```
richgillam, line 808: I'm a little curious about A.endsWith(B) as the algorithm you're using here.  So we'd only coalesce if A and B are equal or B is a substring of A.  Intuitively, it seems like you'd either want to require them to be exactly equal, or you'd want to look for substrings of _any_ (reasonable) length that end A and begin B.  Why the middle-ground approach here?  It seems like you're accounting for a corner case here I'm not seeing.

My first thought was that it had to do with whitespace, but you're not actually changing any of the whitespace-handling logic in this function.  Is it whitespace anyway, or are you accounting for some other situation?

PR #2714, opened 2023-02-13
CLDR-7674 Fix a few locales, adjust charts, test, and utility

macchiati:
CLDR-7674
*.xml files
Added names as per ticket

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLocaleCoverage.java
- Minor change to set default to basic for the last column

- Significant revamp here. 
- Added 3 items at Basic level (we had moved up from Core). 
- Added static sets for quick access to the CoreItems at each level (and below)
- Added a sublevel, to distinguish the ones that should be present at the start. 
- Moved ordinals to Modern (discussed earlier, since we don't depend on them in CLDR itself)
- Restructured the main function getCoreCoverageInfo
    - Now compute result at the end, simplifying the code and making less error-prone
    - Better tests of edge cases like ZZ and 001
    - Handles the autonym (own language) and denonym (own region)
 
tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
- Cleaned up the logic
- Emits warnings for missing items

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CoreCoverageInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CoreCoverageInfo.java
@@ -3,7 +3,6 @@ package org.unicode.cldr.util;
 import java.io.File;
 import java.util.Arrays;
 import java.util.EnumMap;
-import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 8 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/CoreCoverageInfo.java
```diff
@@ -30,107 +33,163 @@ public class CoreCoverageInfo {
...
+        String bestRegion = region.isEmpty() ? maxRegion : region;
+
+        String languagePath = CLDRFile.getKey(CLDRFile.LANGUAGE_NAME, baseLanguage);
+        String languageName = resolvedFile.getStringValue(languagePath);
+        if (languageName == null) {
+            detailedErrors.put(CoreItems.own_language, languagePath);
+        } else {
+            String localeWhereFound = resolvedFile.getSourceLocaleID(languagePath, null);
+            if ("root".equals(localeWhereFound)
+                || "code-fallback".equals(localeWhereFound)) {
```
srl295, line 171:
```suggestion
            if (XMLSource.ROOT_ID.equals(localeWhereFound)
                || XMLSource.CODE_FALLBACK_ID.equals(localeWhereFound)) {
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
```diff
@@ -84,47 +86,43 @@ public void TestSelected() {
     public void TestLocales() {
         long start = System.currentTimeMillis();
         logln("Status\tLocale\tName\tLevel\tCount" + showColumn(all)
-            + "\tError Messages");
-        LanguageTagParser ltp = new LanguageTagParser();
+        + "\tError Messages");
         Multimap<CoreItems, String> errors = LinkedHashMultimap.create();
-        Set<String> toTest = new HashSet(
-            Arrays.asList("ky mn ms uz az kk pa sr zh lo".split(" ")));
```
srl295, line 91: I'm always glad to see hard coded lists removed.

srl295 (approved):

PR #2715, opened 2023-02-14
CLDR-16239 fix main lateral inheritance

macchiati:
CLDR-16239

- added option to retain unchanged files (this allows simple generation of a complete set) with -R
- for -f, added printing of number of changes to each file.
- changed fixList.add('V', "Fix values that would inherit laterally", which changes the following paths, but *just* those that inherit laterally:
  - if we have an L1 locale, change each ↑↑↑ value
  - if we don't have an L1 locale
    - change each ↑↑↑ value where its parent bailey value is different.
    - ALSO change each null value where its parent bailey value is different. (pulled my hair out for this one). This is done in a new handleEnd()
- Also added a new utility to get the parent chain. After the dust settles, this should probably be moved to a central location.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CompareResolved.java
- Added a new parameter 
- Made more verbose printout, but with abbreviated paths.
- Printed counts if either in verbose mode or there are diffs.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
- Minor optimization (we could do more to clean this up, later)

The resulting data was:

36,990 | fix ↑↑↑ lateral; L1
-- | --
87 | fix ↑↑↑ lateral; L2+
171 | fix null lateral; L2+

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
@@ -26,16 +26,46 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.unicode.cldr.draft.FileUtilities;
-import org.unicode.cldr.test.*;
-import org.unicode.cldr.util.*;
+import org.unicode.cldr.test.CLDRTest;
+import org.unicode.cldr.test.CoverageLevel2;
+import org.unicode.cldr.test.DisplayAndInputProcessor;
+import org.unicode.cldr.test.QuickCheck;
+import org.unicode.cldr.test.SubmissionLocales;
+import org.unicode.cldr.util.Annotations;
+import org.unicode.cldr.util.CLDRConfig;
+import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRFile.DraftStatus;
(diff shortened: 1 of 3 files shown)


btangmu:
@macchiati why still draft, not ready for review yet?

btangmu:
@macchiati if you'd like me to work today on "Harden all the values we need to, in order to prevent lateral inheritance AND retain all resolved values", then I suggest that you make this ready for review and we merge, it so we can all use the same updated comparison tool and avoid merge conflicts

btangmu (approved):
rubber stamp for now, will study more soon

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2404,33 +2454,144 @@ public void handleEnd() {
...
+            localeID = parent;
+        }
+    }
+
+    static {
+        System.out.println("en_DE" + "\t" + getParentChain("en_DE"));
+        System.out.println("fr_CA" + "\t" + getParentChain("fr_CA"));
+        System.out.println("fr" + "\t" + getParentChain("fr"));
+        System.out.println("root" + "\t" + getParentChain("root"));
+    }
```
srl295, line 2593: Debugging, belongs in a test if it's for verifying something.

srl295:
PMLGTM

PR #2716, opened 2023-02-14
CLDR-16386 Update windowsZones.xml for 2022g Mexican time zones

mattjohnsonpint:
CLDR-16386

ALLOW_MANY_COMMITS=true


common/supplemental/windowsZones.xml
```diff
@@ -70,13 +70,13 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<mapZone other="US Mountain Standard Time" territory="ZZ" type="Etc/GMT+7"/>
 
 			<!-- (UTC-07:00) Chihuahua, La Paz, Mazatlan -->
```
mattjohnsonpint, line 72: Note: I'm not sure if the display name in comments should change, as Microsoft has yet to make an update for this.

yumaoka: I think it's OK to update the comment later.

PR #2720, opened 2023-02-15
CLDR-16376 v43 BRS: en_GB to en_001 promotion

srl295:
CLDR-16376

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_GB.xml
+++ b/common/main/en_GB.xml
@@ -10680,7 +10680,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<nameField type="surname">↑↑↑</nameField>
 		</sampleName>
 		<sampleName item="nativeFull">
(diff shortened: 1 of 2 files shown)


common/main/en_GB.xml
```diff
@@ -10680,7 +10680,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<nameField type="surname">↑↑↑</nameField>
 		</sampleName>
 		<sampleName item="nativeFull">
-			<nameField type="title">Mr</nameField>
```
macchiati, line 10683: Should change to ↑↑↑ in both cases; keeps vetters from having to confirm next time.

macchiati (changes requested):

macchiati (approved):

PR #2722, opened 2023-02-16
CLDR-16239 Revised data with hardened values

macchiati:
CLDR-16239

1. Generated revised data using the CLDRModify -fV option from https://github.com/unicode-org/cldr/pull/2715. This replaces lateral inheritance when necessary. Almost all of these change the inheritance marker to the equivalent bailey value. < 100 of them changed missing values (getStringValue() == null) to the equivalent bailey value.
2. Verified the result with CompareResolved, to make sure that the resolved results from the old main and the generated main are identical.
3. Then copied the results into main to create this PR. Note that many of the file changes are "empty", just a copyright change.
4. Had to make changes in https://github.com/unicode-org/cldr/pull/2722/commits/c727a5fb1529c56fc475484c06f0b0012443f715 to fix logical group errors in bs.xml and bs_Cyrl.xml. It appears that the logical group check doesn't some but not all items to have inheritanceMarker, so fixed that (and filed ticket).

See also email.

Note that a bunch of the changes are just copyright updates, which would happen later anyway.

ALLOW_MANY_COMMITS=true


macchiati:
PS. This isn't final because of ka.xml, but I can get that later, after https://github.com/unicode-org/cldr/pull/2715 is merged.

common/main/aa_DJ.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2022 Unicode, Inc.
+<!-- Copyright © 1991-2023 Unicode, Inc.
```
srl295, line 3: I'd been not checking in these copyright-only changes. Would be good to skip these, regardless of how CLDR-14239 shakes out.

common/main/ar_SA.xml
```diff
@@ -4496,6 +4496,30 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<compoundUnit type="times">
 				<compoundUnitPattern>↑↑↑</compoundUnitPattern>
 			</compoundUnit>
+			<unit type="angle-revolution">
```
srl295, line 4499: not a ^^^ but it is a missing (aka null)

srl295 (approved):
spot check looks OK. Would be good if copyright-year-only were omitted.

PR #2718, opened 2023-02-14
CLDR-16346 Fix initials 2

macchiati:
CLDR-16346

Detect and fix inconsistencies between initial pattern and literals between initial fields. Handles the "J. R.R. Tolkien" case.

From Person Names WG agenda

Overall Recommendations AGREED

1. For en, en_GB: Switch to J.R.R. Tolkien.
  a. In en, drop space between initial fields, drop space in initial sequence pattern, 
  b. In en_GB, en_001, change to ^^^.
2. For ar, move period to ‘initial’. Was:
  a. <initialPattern type="initial">{0}</initialPattern>  R Gillam→R. Gillam
  b. <initialPattern type="initialSequence">{0}. {1}.</initialPattern> R. B. Gillam (same)
3. For nl, drop the space in the adjacent fields. 
  a. Will look like en, en_GB in (i) above
4. For yue, the sorting forms need spaces, eg
  a. <namePattern>{surname-core}{given-initial}{given2-initial}</namePattern>
  b. The FSR will cause them to be removed or replaced by ·
  c. <namePattern>{surname-core} {given-initial} {given2-initial}</namePattern>

Also adds Check and Test

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #2723, opened 2023-02-16

btangmu:

-Resulting xml

CLDR-16239

ALLOW_MANY_COMMITS=true


btangmu:
This is more different than I expected from https://github.com/unicode-org/cldr/pull/2722

There are differences not only in which values got "hardened", but also in what they got as hardened values. Quite possibly there are mistakes in how I generated this, being distracted and out of office. To be safe, I suggest we study further before merging any PR with big changes to data, if we can afford a little more time.

Here, common/main is from "gh pr checkout 2722", and the other directory is the source of this PR (2723).

```
$ diff -ICopyright common/main ../vetdata-2023-02-16-modW2/vxml/common/main
Only in common/main: .DS_Store
diff -ICopyright common/main/ar.xml ../vetdata-2023-02-16-modW2/vxml/common/main/ar.xml
14436,14437c14436,14437
< 		<initialPattern type="initial">{0}.</initialPattern>
< 		<initialPattern type="initialSequence">{0} {1}</initialPattern>
---
> 		<initialPattern type="initial">{0}</initialPattern>
> 		<initialPattern type="initialSequence">{0}. {1}.</initialPattern>
diff -ICopyright common/main/bs.xml ../vetdata-2023-02-16-modW2/vxml/common/main/bs.xml
5097c5097
< 						<era type="0" alt="variant">P. n. e.</era>
---
> 						<era type="0" alt="variant">P.n.e.</era>
diff -ICopyright common/main/bs_Cyrl.xml ../vetdata-2023-02-16-modW2/vxml/common/main/bs_Cyrl.xml
1861c1861
< 						<era type="0" alt="variant">п. н. е.</era>
---
> 						<era type="0" alt="variant">п.н.е</era>
1863c1863
< 						<era type="1" alt="variant">н. е.</era>
---
> 						<era type="1" alt="variant">н.е.</era>
...
```

btangmu:
Evidently I wasn't up to date with Mark's PR from Wednesday: https://github.com/unicode-org/cldr/pull/2718

That accounts for the differences in common/main/ar.xml and maybe others.

@macchiati given that https://github.com/unicode-org/cldr/pull/2722 was already merged, maybe it would be a waste of time for me to continue investigating this. If there's anything more I should do for this ticket (v43), just let me know.

srl295:
> Evidently I wasn't up to date with Mark's PR from Wednesday: #2718
> 
> That accounts for the differences in common/main/ar.xml and maybe others.
> 
> @macchiati given that #2722 was already merged, maybe it would be a waste of time for me to continue investigating this. If there's anything more I should do for this ticket (v43), just let me know.


at least the conflicts will need resolution.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CompareResolved.java
```diff
@@ -156,7 +156,10 @@ public static void main(String[] args) {
                 // fell through, so record diff
 
                 ++diffCount;
-                System.out.println(path + "\t" + sourceValue + "\t" + compareValue);
+                if (sourceValue == null || compareValue == null || sourceValue.isEmpty() || compareValue.isEmpty()) {
+                    throw new RuntimeException("Null or empty sourceValue or compareValue: [" + sourceValue + "/" + compareValue + "]");
+                }
+                System.out.println("≠DIFF\t" + localeID + "\t" + path + "\t" + sourceValue + "\t" + compareValue);
```
srl295, line 162: seems like a good improvement, maybe a separate PR?

macchiati (changes requested):
I will need time to study this

common/main/de.xml
```diff
@@ -125,7 +125,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="chy">Cheyenne</language>
 			<language type="ckb">Zentralkurdisch</language>
 			<language type="ckb" alt="menu">Kurdisch (Sorani)</language>
-			<language type="ckb" alt="variant">↑↑↑</language>
+			<language type="ckb" alt="variant">Zentralkurdisch</language>
```
macchiati, line 128: When I look at main, I see that it already has this change.

https://github.com/unicode-org/cldr/blob/main/common/main/de.xml#L128

I suggest you rebase and we can see what the differences are.

macchiati: Or what would be simpler, is to make a copy of the tool changes, make a new branch from main, and run those changes again. That way you don't need to muck with the conflicts.

**What we need to to is make sure that the logic in the CLDRModify and the CompareResolved is correct.**

macchiati (changes requested):

common/main/de.xml
btangmu: I'll get started on that!

btangmu:
see https://github.com/unicode-org/cldr/pull/2732 for rebased equivalent (without re-running the tools!)

btangmu:
> at least the conflicts will need resolution

@srl295 conflict resolution might not be necessary -- as far as I know we're good with what's already committed to main branch (which already passes the CompareResolved requirement) -- the point of this alternative method of "hardening" is only for comparison as a way of double-checking the data that's already merged

most differences will probably be "harmless" and simply due to the fact that there's more than one way to "harden" the data, and one way isn't necessarily better than the other

so "resolution" might just consist in satisfying ourselves that the differences are all harmless

my latest attempt is at https://github.com/unicode-org/cldr/pull/2734

PR #2724, opened 2023-02-16
CLDR-16397 v43 BRS: update various codes

srl295:
CLDR-16397

- add `CQ` to territories in RFC 4646 but not in CLDR. I think this is OK, it's in the same boat as `IC` being exceptionally  reserved.
- add Turkey as an alias for TR
- make alternate_country_names.txt support comments
- make failure messages clearer
- clean up dead code
- update ISO-639-3 to v20230123
- update langauge-subtag-registry to 2023-02-14
- update tlds-alpha to 2023021501

ALLOW_MANY_COMMITS=true


srl295:
I stumbled on other tickets regarding CQ / Sark https://unicode-org.atlassian.net/browse/CLDR-10923

DavidLRowe:
I didn't see any red flags, but I'm not qualified to review Java code and only have a vague idea of other stuff. The change for apc/ajp looks reasonable.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/draft/ScriptMetadata.java
```diff
@@ -101,25 +101,9 @@ public enum Shaping {
...
-        // addSynonym(NAME_TO_LANGUAGE_CODE, "Ancient Greek", "Ancient Greek (to 1453)");
-        // //addSynonym(NAME_TO_LANGUAGE_CODE, "Khmer", "Cambodian");
-        // addSynonym(NAME_TO_LANGUAGE_CODE, "Old Irish", "Old Irish (to 900)");
-
-        // addNameToCode("region", NAME_TO_REGION_CODE);
-        // // NAME_TO_REGION_CODE.put("UNKNOWN", "ZZ");
-        // // NAME_TO_REGION_CODE.put("", "ZZ");
-        // NAME_TO_REGION_CODE.put("N/A", "ZZ");
-        // addSynonym(NAME_TO_REGION_CODE, "Laos", "Lao People's Democratic Republic");
-    }
```
srl295, line 122: dead code

tools/cldr-code/src/main/java/org/unicode/cldr/tool/AddPopulationData.java
```diff
@@ -487,17 +487,17 @@ public boolean handle(String line) {
...
+                if (gdp == 0) {
                     myErrors.append("\n" + territory + ";" + sc.getData("territory", territory) + ";gdp-ppp;0;reason");
                 }
                 if (literacy == 0) {
                     myErrors.append("\n" + territory + ";" + sc.getData("territory", territory) + ";literacy;0;reason");
                 }
-                if (population == 0) {
-                    myErrors.append("\n" + territory + ";" + sc.getData("territory", territory)
-                        + ";population;0;reason");
-                }
```
srl295, line 500: the advice given was incorrect, had rows in the wrong order

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
```diff
@@ -815,7 +818,7 @@ public String toString(boolean b) {
 
         static boolean MARK_OUTPUT = false;
 
-        public String getRickLanguageCode() {
+        public String getLanguageCode() {
```
srl295, line 821: @rickmcgowan  you OK with this change?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateEnums.java
```diff
@@ -385,6 +385,7 @@ public void loadCLDRData() throws IOException {
         missing.remove("IC");
         missing.remove("QU");
         missing.remove("UN");
+        missing.remove("CQ");
```
srl295, line 388: this may not be needed if validity is updated

rickmcgowan (commented):
Viewed the changes, and they look reasonable, but I'm not really qualified on most of them.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
rickmcgowan: Heh heh... Sure... :-)

macchiati (changes requested):
Looks good, except that we need to decide about Sark. It is exceptionally reserved, so we have the choice of whether to have it be translated or not. 
Can you see what we do with the other exceptionally reserved, and then we can decide?

srl295:
OK it's off and running. May be more tests to fix but ptal

tools/cldr-code/src/main/java/org/unicode/cldr/util/Iso3166Data.java
```diff
@@ -0,0 +1,80 @@
...
+            }
+            this.isoStatus = Collections.unmodifiableMap(isoStatus);
+            this.isoDescription = Collections.unmodifiableMap(isoDescription);
+        }
+
+        public static final Iso3166Helper INSTANCE = new Iso3166Helper();
+    }
+
+    private static final List<String> DELETED3166 = Collections.unmodifiableList(Arrays
+    .asList(new String[] { "BQ", "BU", "CT", "DD", "DY", "FQ", "FX", "HV",
```
macchiati, line 53: Not blocker: the "new String[] {" can now be omitted (with the trailing "}").

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CompareIso3166_1Status.java
```diff
@@ -21,35 +20,21 @@
...
-        Splitter semi = Splitter.on(';').trimResults();
-        for (String line : FileUtilities.in(StandardCodes.class, "data/external/iso_3166_status.txt")) {
-            if (line.startsWith("#")) continue;
-            List<String> parts = semi.splitToList(line);
-            // AC ; Exceptionally reserved  ; Refers to the United Nations and reserved by the ISO 3166 Maintenance Agency.
-            final String regionCode = parts.get(0);
-            isoStatus.put(regionCode, Iso3166Status.valueOf(parts.get(1).toLowerCase(Locale.ROOT).replace(' ', '_')));
-            isoDescription.put(regionCode, parts.get(1));
-        }
+        final Map<String, Iso3166Status> isoStatus = Iso3166Data.getIsoStatus();
```
macchiati, line 36: Good change, to refactor that out.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/Iso3166Data.java
srl295: makes sense…

PR #2728, opened 2023-02-18
CLDR-15111 Full set of test data

macchiati:
CLDR-15111

This now has a full set of test data, and a draft version of a test to go into ICU.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/CheckPersonNamesTest.java
The ICU test is still hacky, because I couldn't get a regular test to run. Look at this file. At the top of that file are additional items that should be fixed in the ICU implementation.

When you run the Check program you'll get an output of failure cases.
- Many are with SORTING, so there is an abbreviated listing of those.
- ICU might not be computing the order of the locale of the name properly (I through in a spurious region to check that). I put in a function in the tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/PersonNameFormatter.java called     getOrderFromLocale(ULocale inputLocale). You can take a look at that for how I think ICU ought to do it.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/testData/personNameTest/af.txt
+++ b/common/testData/personNameTest/af.txt
@@ -1,4 +1,4 @@
-# CLDR person name formatting test data for: ja
+# CLDR person name formatting test data for: af
 #
 # Test lines have the following structure:
 #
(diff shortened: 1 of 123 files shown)


macchiati:
ping

richgillam (approved):
Didn't have time for an extensive review, but a spot-check of everything looks good.  I'll try and figure out what's going on on the ICU side when I get to a good stopping place in my day job.

macchiati:
Thanks. It looks like most of the differences are in space / literal handling, as we'd expect.

PR #2730, opened 2023-02-20
CLDR-16377 v43 BRS: cldrModify, derived annotations, algo

srl295:
CLDR-16377

## What's inside? 📦 
- re-ran cldr-modify on annotationsDerived - maybe this was missed before.
- regenerated derived annotations.  Some surprises:
   - U+00A0 added in `cs.xml`
   - RLE mark removed from `ar.xml`
   - Many updates due to `TR` name change
- algorithmic locales
   -  only common/main changes due to inheritance changes

ALLOW_MANY_COMMITS=true


srl295:
- annotationsDerived/en_GB derived annotations removed!  The original file is 100% ↑↑↑ 
- annotationsDerived/ha_NE dropped maybe same reason
- several new other annotationsDerived/ such as quc

macchiati:
The en_GB is ok, since all the annotations get pushed up into en_001.

macchiati (approved):

PR #2732, opened 2023-02-21

btangmu:

-Resulting xml

CLDR-16239

ALLOW_MANY_COMMITS=true


btangmu:
this is the same as https://github.com/unicode-org/cldr/pull/2723 except that it's rebased on current main (which includes  https://github.com/unicode-org/cldr/pull/2722 )

IMPORTANT: this PR is flawed since it the result of running CLDRModify and CompareResolved on outdated data -- I didn't have time yet to re-run the tools -- so this is just to get a rough idea of how this approach compares with  https://github.com/unicode-org/cldr/pull/2722

btangmu:
see https://github.com/unicode-org/cldr/pull/2734

PR #2736, opened 2023-02-22
CLDR-16410 Fix GenerateProductionData

macchiati:
CLDR-16410

The following has a solution for v43, but we will need follow-on work in v44.

The problem is that resolved(main) != resolved(GenerateProductionData(main))

common/main/be_TARASK.xml ...
- Added a few items that need to be retained in the production data. 

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocaleIDParser.java
tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestInheritance.java
- Refactored to move getParentChain to a reasonable location, because I needed it at one point but didn't want spaghetti import chains. 
- Turned out to not be necessary, but it is a good refactoring anyway. 
- Ended up using     public static boolean isL1(String localeId)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CompareResolved.java
- Improved the printout to make it easier to find the bad cases with a spreadsheet

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
- added a number of privates to expose possible dead code
- removed localeToSubdivisionsToMigrate so that the 3 subdivisions were not handled differently. Code that used it was also removed. This handled one set of problems.
- pathEqualsOrIsOkAltVariantOf
  - renamed to clarify usage
  - added extra parameters needed to do a good job
  - Tried a number of ways to fix the problem, but it requires more substantive changes than we have time for in v43
  - So added LOCALE_TO_PATH_EXCEPTIONS, to contain the few items remaining that need to be handled.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/be_TARASK.xml
+++ b/common/main/be_TARASK.xml
@@ -48,6 +48,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="awa" draft="provisional">авадгіская</language>
 			<language type="ay" draft="provisional">аймарская</language>
 			<language type="az" draft="provisional">азэрбайджанская</language>
+			<language type="az" alt="short" draft="provisional">азербайджанская</language>
 			<language type="ba" draft="provisional">башкірская</language>
 			<language type="ban" draft="provisional">балійская</language>
 			<language type="bas" draft="provisional">басайская</language>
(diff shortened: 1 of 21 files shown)


pedberg-icu (approved):

Jackpoot:
Go to

PR #2740, opened 2023-02-22
CLDR-16400 v43 BRS: lang pop, ConvertLanguageData, Likely…

srl295:
CLDR-16400

- [X] fixes to country language population data, regen supplemental data
- [X] add links to CLDR-16380 and warnings to not drop sil1 data
- [x] regen likely subtags

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -988,8 +988,7 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Hmong Njua; ?; Suriname } => { Hmong Njua; Lao; Suriname }-->
 		<likelySubtag from="hnj_TH" to="hnj_Laoo_TH"/>
 		<!--{ Hmong Njua; ?; Thailand } => { Hmong Njua; Lao; Thailand }-->
-		<likelySubtag from="hnj_US" to="hnj_Laoo_US"/>
-		<!--{ Hmong Njua; ?; United States } => { Hmong Njua; Lao; United States }-->
+		<likelySubtag from="hnj_US" to="hnj_Hmnp_US"/> <!-- manual update -->
 		<likelySubtag from="hnj_VN" to="hnj_Laoo_VN"/>
 		<!--{ Hmong Njua; ?; Vietnam } => { Hmong Njua; Lao; Vietnam }-->
 		<likelySubtag from="hnj_Laoo" to="hnj_Laoo_LA"/>
@@ -3438,6 +3437,8 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ ?; Coptic; ? } => { Coptic; Coptic; Egypt }-->
 		<likelySubtag from="und_Cpmn" to="und_Cpmn_CY"/> <!-- TODO: manually re-added -->
  		<!--{ ?; Cypro-Minoan; ? } => { ?; Cypro-Minoan; Cyprus }-->
+		<likelySubtag from="und_Cpmn_CY" to="und_Cpmn_CY"/>  <!-- TODO: manually re-added -->
+ 		<!--{ ?; Cypro-Minoan; Cyprus } => { ?; Cypro-Minoan; Cyprus }-->
 		<likelySubtag from="und_Cprt" to="grc_Cprt_CY"/>
 		<!--{ ?; Cypriot; ? } => { Ancient Greek; Cypriot; Cyprus }-->
 		<likelySubtag from="und_Cyrl" to="ru_Cyrl_RU"/>
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE supplementalData SYSTEM "../../common/dtd/ldmlSupplemental.dtd">
 <!--
-Copyright © 1991-2022 Unicode, Inc.
+Copyright © 1991-2023 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-DFS-2016
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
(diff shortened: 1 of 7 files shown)


DavidLRowe (approved):
changes seem reasonable

common/supplemental/likelySubtags.xml
```diff
@@ -1928,8 +1950,6 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Nzima; ?; ? } => { Nzima; Latin; Ghana }-->
 		<likelySubtag from="oc" to="oc_Latn_FR"/>
 		<!--{ Occitan; ?; ? } => { Occitan; Latin; France }-->
-		<likelySubtag from="oc_ES" to="oc_Latn_ES"/>
-		<!--{ Occitan; ? ; Spain } => { Occitan; Latin; Spain }-->
```
srl295, line 1932: I think this removal is OK because the script is already picked up from the previous line… correct?

common/supplemental/likelySubtags.xml
```diff
@@ -2085,11 +2105,7 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="ria" to="ria_Latn_IN"/>
 		<!--{ Riang (India); ?; ? } => { Riang (India); Latin; India }-->
 		<likelySubtag from="rif" to="rif_Latn_MA"/>
-		<!--{ Riffian; ?; ? } => { Riffian; Latn; Morocco }-->
-		<likelySubtag from="rif_Tfng" to="rif_Tfng_MA"/>
-		<!--{ Riffian; Tifinagh; ? } => { Riffian; Tifinagh; Morocco }-->
-		<likelySubtag from="rif_NL" to="rif_Latn_NL"/>
-		<!--{ Riffian; ?; Netherlands } => { Riffian; Latin; Netherlands }-->
+		<!--{ Riffian; ?; ? } => { Riffian; Latin; Morocco }-->
```
srl295, line 2107: @DavidLRowe I think the remainder of these likely tags are also implied.

srl295 (commented):
ended up with some manual updates to likelySubtags.xml to keep the status quo.

common/supplemental/likelySubtags.xml
macchiati: Yes. Changes like this should be the result of a population increase that tips the scales to ES. So you need to check that that is the case.

common/supplemental/likelySubtags.xml
macchiati: These replacement happens if the Tfng doesn't change the country, and NL doesn't change the script.

common/supplemental/likelySubtags.xml
```diff
@@ -2882,6 +2897,8 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Zaza; ?; ? } => { Zaza; Latin; Türkiye }-->
 		<likelySubtag from="und" to="en_Latn_US"/>
 		<!--{ ?; ?; ? } => { English; Latin; United States }-->
+		<likelySubtag from="und_001" to="en_Latn_US"/>
```
macchiati, line 2900: This shouldn't be showing up, because it duplicates line 2897 (adding 001 makes no difference). Just in case, let's hand-delete for now.

common/supplemental/likelySubtags.xml
```diff
@@ -3418,10 +3436,8 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ ?; Chorasmian; ? } => { Chorasmian; Chorasmian; Uzbekistan }-->
 		<likelySubtag from="und_Copt" to="cop_Copt_EG"/>
 		<!--{ ?; Coptic; ? } => { Coptic; Coptic; Egypt }-->
-		<likelySubtag from="und_Cpmn" to="und_Cpmn_CY"/>
-		<!--{ ?; Cypro-Minoan; ? } => { ?; Cypro-Minoan; Cyprus }-->
-		<likelySubtag from="und_Cpmn_CY" to="und_Cpmn_CY"/>
-		<!--{ ?; Cypro-Minoan; Cyprus } => { ?; Cypro-Minoan; Cyprus }-->
+		<likelySubtag from="und_Cpmn" to="und_Cpmn_CY"/> <!-- TODO: manually re-added -->
+ 		<!--{ ?; Cypro-Minoan; ? } => { ?; Cypro-Minoan; Cyprus }-->
```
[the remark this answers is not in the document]
macchiati: Yes, there is a known problem with the SIL data that Peter ran into; should never have X_Y_Z map to X_Y_Z.

common/supplemental/likelySubtags.xml
pedberg-icu, line 2900: Hmm, I do not offhand know of another case in which likelySubtags alters an explicit subtag other than und or ZZ. This seems a bit odd. Why not und_001 -> en_Latn_001 (since we have a locale for en_001)?

macchiati (changes requested):
Just the change noted.

common/supplemental/likelySubtags.xml
srl295: There are other nnn to TT (number to territory ) cases. But yeah en 101 makes sense here

common/supplemental/likelySubtags.xml
srl295: This is already ES

common/supplemental/likelySubtags.xml
macchiati: The other ones are continents; 001 is the whole world.

common/supplemental/likelySubtags.xml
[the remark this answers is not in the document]
srl295: This isn't from SIL, this was hand added a long time ago. 

Without the X-Y-Z mapping to X-Y-Z we get the following error:

      Error: (LikelySubtagsTest.java:199) : checkAdding: max(und_Cpmn)->und_Cpmn_CY, however max(und_Cpmn_CY)->: expected "und_Cpmn_CY", got "el_Cpmn_CY"
      Error: (LikelySubtagsTest.java:199) : checkAdding: max(und_Cpmn)->und_Cpmn_CY, however max(und_Cpmn_CY)->: expected "und_Cpmn_CY", got "el_Cpmn_CY"

common/supplemental/likelySubtags.xml
srl295: this is already in _ES so no scale tipping needed.   Again this was a manual addition (from years back) and I think it's just not needed.

common/supplemental/likelySubtags.xml
srl295: makes sense

macchiati (approved):

srl295:
thanks. will merge when tests pass

PR #2741, opened 2023-02-22
CLDR-16419 CLDRModify changes: space fixes for newly added items, exemplar reorder for geresh/ayim

pedberg-icu:
CLDR-16419

CLDRModify passes none, -fQ, -fP before alpha2 integration & tag.

The changes are:
1. Space normalization for recently-added items, mostly or all from https://github.com/unicode-org/cldr/pull/2722 (inheritance hardening)
2. Exemplar reordering in `he`/`yi` due to the recent collation changes for geresh/gershayim etc.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/sc.xml
+++ b/common/main/sc.xml
@@ -1363,11 +1363,11 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 					<intervalFormats>
 						<intervalFormatFallback>{0} – {1}</intervalFormatFallback>
 						<intervalFormatItem id="Bh">
-							<greatestDifference id="B">h B – h B</greatestDifference>
+							<greatestDifference id="B">h B – h B</greatestDifference>
 							<greatestDifference id="h">h–h B</greatestDifference>
 						</intervalFormatItem>
 						<intervalFormatItem id="Bhm">
-							<greatestDifference id="B">h:mm B – h:mm B</greatestDifference>
+							<greatestDifference id="B">h:mm B – h:mm B</greatestDifference>
 							<greatestDifference id="h">h:mm–h:mm B</greatestDifference>
 							<greatestDifference id="m">h:mm–h:mm B</greatestDifference>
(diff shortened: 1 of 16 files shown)


macchiati (approved):

common/main/sc.xml
```diff
@@ -1363,19 +1363,19 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 					<intervalFormats>
 						<intervalFormatFallback>{0} – {1}</intervalFormatFallback>
 						<intervalFormatItem id="Bh">
-							<greatestDifference id="B">h B – h B</greatestDifference>
+							<greatestDifference id="B">h B – h B</greatestDifference>
```
btangmu, line 1366: this wouldn't have been necessary if the value in root were normalized; is there a reason that root isn't normalized?

btangmu: this is "hard inheritance" (from root) except for normalization; the lack of root normalization may make it harder for us to see what's really going in our v44 inheritance work

macchiati:
Hmmm. We should probably include root in the DAIP pass.

common/main/sc.xml
pedberg-icu: The space normalization (to \u2009\u2013\u2009) is only for locales in scripts Latn, Cyrl, Grek; if we normalized root then it would have the wrong fallback for all other locales.

pedberg-icu: (And locales in scripts Latn, Cyrl, Grek ar more likely to have vetters providing explicit values, and not rely on root fallback)

pedberg-icu:
> Hmmm. We should probably include root in the DAIP pass.
I think not, for the reasons I mention in the responses to Tom's comment above.

macchiati:
Good point. If we change root, we need to 'harden' the inheritance from it.

PR #2744, opened 2023-02-22
CLDR-16418 Add spec text for parent locales by component

sffc:
CLDR-16418

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -1682,6 +1682,7 @@ _Examples:_
 
 ```xml
 <!ELEMENT parentLocales ( parentLocale* ) >
+<!ATTLIST parentLocales component NMTOKEN #IMPLIED >
 <!ELEMENT parentLocale EMPTY >
 <!ATTLIST parentLocale parent NMTOKEN #REQUIRED >
 <!ATTLIST parentLocale locales NMTOKENS #REQUIRED >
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35.md
```diff
@@ -1721,9 +1730,8 @@ There may be specific exceptions to these for certain closely-related languages
 
 There are certain invariants that must always be true:
 
-3. The parent must either be the root locale or have the same script as the child.
```
pedberg-icu, line 1724: This requirement is still true for the non-component parentLocales data. And for the component ones the only cross-script parents are related, i.e. both Hans and Hant are subsets of Hani, so we only allow inheritance across scripts when they are such subsets.

sffc: Collator has an inheritance in Serbian that breaks this rule

pedberg-icu: Ah OK. But we can still keep this condition for the non-component parentLocales.

docs/ldml/tr35.md
```diff
@@ -1706,7 +1707,15 @@ For case 2, the children and parent share the same primary language, but the reg
...
+There are certain components that require addenda to the common parent fallback rules. For example, in collation, for a locale like `zh_Hant` in the example above, the `parentLocale` element would dictate the parent as `root` when referring to main locale data, but for collation data, the parent locale would still be `zh`, even though the `parentLocale` element is present for that locale. To address this, components can have their own fallback rules that inherit from the common rules and add additional parents that supplement or override the common rules:
+
+```xml
+<parentLocales component="segmentations">
+  <parentLocale parent="zh" locales="zh_Hant"/>
+</parentLocales>
+```
+
+The above data would mean that in the "segmentations" component, `zh_Hant` should inherit from `zh` instead of from `root`.
```
macchiati, line 1718: I think we have to be more explicit about "add additional parents"

Suppose that the normal rules (component="") are

parentLocale parent="X" locales="Y Z"
parentLocale parent="A" locales="B C"


1. What effect does having the following make with component="collations"?

parentLocale parent="A" locales="Y B"

I'd propose that we treat the rules logically as if they were all decomposed into pairs:

component=""
(X, Y)
(X, Z)
(A, B)
(A, C)
raw component="collations"
(A, Y)
(A, B)

resolved component="collations" (italic means removed)
_(X, Y)_
(X, Z)
_(A, B)_
(A, C)
(A, Y)
(A, B)

When merging to produce the resolved collation parentLocales, when we are adding a pair from a component, then we remove any pair from the original where the second of the pair matches. That also has the effect of removed duplicate pairs. 
Thus the parentLocale parent="A" locales="Y B" doesn't have any effect on (A, C).  That also means that if we want to remove a parent child relationship, we have to specify where it goes (which is fine).

macchiati (changes requested):

docs/ldml/tr35.md
macchiati: There are other ways to phrase this that might be more direct. Maybe something like:

To produce the resolved relations from the no-component rules plus component=X
1. Let S be the set of all no-component parentLocale elements.
2. For each component=X element E
  1. Remove all locales NMTOKENS from all parentLocale elements in S.
  2. If there isn't a parentLocale element E2 in S with the same parent, add E to S
  3. If there is a parentLocale element E2 in S with the same parent, add the locale NMTOKENS from E to E2
3. At the end, remove each element from S with an empty locales attribute value.

We should also add that there is a constraint that for a given component, for any two distinct parentLocale elements E and E2:
1. E's parent ≠ E2's parent
2. E's locales are disjoint with E2's locales.

docs/ldml/tr35.md
sffc: Restored the rule.

docs/ldml/tr35.md
sffc: I tried explaining it in terms of key/value maps. PTAL.

docs/ldml/tr35.md
```diff
@@ -1707,15 +1707,71 @@ For case 2, the children and parent share the same primary language, but the reg
...
+There are certain components that require addenda to the common parent fallback rules. For a locale like `zh_Hant` in the example above, the `parentLocale` element would dictate the parent as `root` when referring to main locale data, but for collation data, the parent locale should still be `zh`, even though the `parentLocale` element is present for that locale. To address this, components can have their own fallback rules that inherit from the common rules and add additional parents that supplement or override the common rules:
 
 ```xml
 <parentLocales component="segmentations">
   <parentLocale parent="zh" locales="zh_Hant"/>
 </parentLocales>
 ```
 
-The above data would mean that in the "segmentations" component, `zh_Hant` should inherit from `zh` instead of from `root`.
+Logically, component-specific parent locales should be merged with the common parent locales as if merging maps with children as the keys and parents as the values. For example, consider the following XML:
```
macchiati, line 1718: I think this can work, but needs to be explicit about the order of the merge. If I'm merging normal into collations as a map, then usually the later one wins. But that's not what you want in this case.

macchiati (changes requested):

docs/ldml/tr35.md
macchiati: > as if merging maps with children as the keys and parents as the values. 

Maybe: 

as if merging maps with children as the keys and parents as the values, and retaining the component-specific mapping whenever there are duplicate keys.

macchiati:
This will need to be retargeted at v43.

macchiati (dismissed):
Works for me now, but should let Peter weigh in also.

macchiati:
@pedberg-icu

pedberg-icu (approved):
Much clearer, thanks.

pedberg-icu (approved):

PR #2745, opened 2023-02-22
CLDR-16417 Add spec text for era codes

sffc:
CLDR-16417

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -1022,6 +1022,8 @@ As in other cases, **narrow** may be ambiguous out of context.
 <!ATTLIST era type NMTOKENS #REQUIRED>
 <!ATTLIST era start CDATA #IMPLIED>
 <!ATTLIST era end CDATA #IMPLIED>
+<!ATTLIST era code NMTOKEN #IMPLIED >
+<!ATTLIST era aliases NMTOKENS #IMPLIED >
 ```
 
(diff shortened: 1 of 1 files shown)


pedberg-icu (approved):
looks good, thanks

docs/ldml/tr35-dates.md
```diff
@@ -1042,6 +1044,12 @@ For a sequence of eras with specified start dates, the end of each era need not
 …
 ```
 
+Some eras have additional `code` and `aliases` attributes that define invariant strings for identifying the eras. The `code` is a single globally unique identifier, and `aliases` are space-separated identifiers unique within the calendar. The code and aliases follow the following rules:
+
+1. Every calendar has an era with a `code` that is the same as the BCP-47 name of that calendar. This era should be used for anchoring the "extended year" in the calendar (`u` in the date format pattern).
+2. Eras that count backwards (larger numbers for older years) are suffixed with `-inverse`.
+3. If the same era code is used in multiple calendars, the arithmetic for that era must be the same in all calendars in which it is used. For example, the `ethioaa` era is used in two calendar systems.
```
macchiati, line 1051: the arithmetic => 
the calculations for year, month, and day

macchiati (changes requested):

macchiati (approved):

pedberg-icu (approved):

srl295:
no discussion of -1 and -2 eras

PR #2746, opened 2023-02-23
CLDR-15861 Add additional spec text around resource inheritance

sffc:
CLDR-15861

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -1477,7 +1477,7 @@ If a language has more than one script in customary modern use, then the CLDR fi
 lang
 lang_script
 lang_script_region
-lang_region (aliases to lang_script_region)
+lang_region (aliases to lang_script_region based on likely subtags)
 ```
 
 #### <a name="Bundle_vs_Item_Lookup" href="#Bundle_vs_Item_Lookup">Bundle vs Item Lookup</a>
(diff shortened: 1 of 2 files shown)


sffc:
I focused my PR on the new region-priority inheritance. However, it does not contain the following pieces from my [Flexible Vertical Fallback](https://docs.google.com/document/d/1Mp7EUyl-sFh_HZYgyeVwj88vJGpCBIWxzlCwGgLCDwM/edit#) doc:

1. The clean algorithm for language-priority fallback. UTS 35 has the rules for how to handle language-based fallback scattered around the section, with no single step-by-step algorithm, but as far as I can tell, the expected behavior is the same.
2. My doc has a section on fallback with extensions, which I didn't include because it seems rather niche for the CLDR spec. I anticipate that there will be ICU4X-specific documentation about this part.
3. I also did not include the part about the subdivision tag. I could include this if you like.

macchiati (dismissed):

macchiati (approved):

sffc:
This is ready to merge after another approval

docs/ldml/tr35.md
```diff
@@ -1474,9 +1474,11 @@ If a language has more than one script in customary modern use, then the CLDR fi
 lang
 lang_script
 lang_script_region
-lang_region (aliases to lang_script_region)
+lang_region (aliases to lang_script_region based on likely subtags)
 ```
 
+Note: When the requested locale is `lang_script_region` and `lang_region` implies the `script` according to likely subtags, then resource lookup should start at `lang_region`. This applies whether or not the language has more than one script in customary modern use. For example, `zh-Hant-TW` should start at `zh-TW` (since `zh-TW` implies `Hant`), and `de-Latn-LI` should start at `de-LI` (since `de` implies `Latn` and `de-LI` does not have its own entry in likely subtags).
```
macchiati, line 1480: Add another note:

Note: Item lookup also requires use of the parentLocales, as described in <a href="#Parent_Locales">Parent Locales</a>.

docs/ldml/tr35.md
```diff
@@ -1725,6 +1727,37 @@ There are certain invariants that must always be true:
 4. There must never be cycles, such as: X parent of Y ... parent of X.
 5. Following the inheritance path, using parentLocale where available and otherwise truncating the locale, must always lead eventually to the root locale.
 
+#### 4.1.4 <a name="Region_Priority_Inheritance" href="#Region_Priority_Inheritance">Region-Priority Inheritance</a>
+
+Certain data may be more appropriate to store with the region as the primary key instead of language. This is often needed for regional user preferences, such as week info, calendar system, and measurement system.
+
+The default search chain for region-priority inheritance removes the language subtag before the region subtag, and it restores the variant subtag after doing so, as follows:
```
macchiati, line 1734: We have never done reattachment of variants before, for normal item lookup. The variant is really more closely attached to the base language than to the region, so I think the "US_someVariant" step should be removed. 

----
Longer term, I think we need a rewrite of inheritance to have clean separation, and treat resource bundle lookup and item lookup in separate sections, and use distinct terminology in each, and always use the same terminology.

macchiati (changes requested):

docs/ldml/tr35.md
macchiati: For this to be usable, we have to be precise about exactly which items use region inheritance, and which don't. I'd propose that we say that this applies to all item in rgScope (found in rgScope.xml). Note that some of the attributes that are named "region" take regions or locales, eg

<!ATTLIST hours regions NMTOKENS #REQUIRED >

macchiati (changes requested):

macchiati:
@pedberg-icu , You might look over this PR as well, since it affects inheritance.

macchiati:
This will need to be retargeted at maint/maint-43

docs/ldml/tr35.md
sffc: There is already a statement below the example block that says:

> _Both the resource bundle inheritance and the inherited item inheritance use the parentLocale data, where available, instead of simple truncation._

docs/ldml/tr35.md
sffc: Thanks for the reference to rgScope. How does this look? I added a link to the existing section on rgScope. Do we need to say anything about attributes named "region" taking regions or locales?

macchiati (dismissed):
Looks ok to me now. @pedberg-icu ?

pedberg-icu (approved):
Looks good!

pedberg-icu (approved):

PR #2747, opened 2023-02-23
CLDR-16122 v43 BRS: rename CompareFilesBetweenReleases2  to CompareFilesBetweenReleases

srl295:
The '2' is superfluous.

CLDR-16122

ALLOW_MANY_COMMITS=true


macchiati:
I filed a ticket to do more substantive renaming, folders, etc in v43. We
might wait on this minor item until then.

srl295:
> I filed a ticket to do more substantive renaming, folders, etc in v43. We might wait on this minor item until then.

can you reference it here? Can pick this up in that release.

Also we should stop looking in seed (or actually error if present)

macchiati:
https://unicode-org.atlassian.net/browse/CLDR-16383

And we need to look in seed for when the code looks at old versions

macchiati (approved):

PR #2748, opened 2023-02-23
CLDR-16122 v43 BRS: update languageGroup.xml

srl295:
CLDR-16122

- update to data file
- also, update FormattedFileWriter to print the path of files as they are written (docs were wrong on outputted file path for this task)

ALLOW_MANY_COMMITS=true


srl295:
Q: does cldr-staging need to be updated with this? or will that be picked up 'next time charts are run'?

macchiati:
> Q: does cldr-staging need to be updated with this? or will that be picked up 'next time charts are run'?

It will be picked up with the next charts. However, I usually build that chart just to check differences. (I don't see any yet in this PR)

srl295:
> > Q: does cldr-staging need to be updated with this? or will that be picked up 'next time charts are run'?
> 
> It will be picked up with the next charts. However, I usually build that chart just to check differences. (I don't see any yet in this PR)

the built HTML file had diffs vs. cldr-staging, reflecting the data changes.

btangmu:
This is new to me. I see the "Language Containment" item in BRS, and this documentation:
https://cldr.unicode.org/development/updating-codes/updating-language-groups

@srl295 I guess you already ran TestLanguageGroup?

Any suggestions on what I should do for review?

I see "mub" was removed, and "dep" was added; neither of those is in CLDR anyway.

How does CLDR use these files? For example, childToParentWithCodes.tsv? I see where in the code the file is created, but I don't see where it's used...

srl295:
> This is new to me. I see the "Language Containment" item in BRS, and this documentation: https://cldr.unicode.org/development/updating-codes/updating-language-groups
> 
> @srl295 I guess you already ran TestLanguageGroup?
> 
> Any suggestions on what I should do for review?
> 
> I see "mub" was removed, and "dep" was added; neither of those is in CLDR anyway.
> 
> How does CLDR use these files? For example, childToParentWithCodes.tsv? I see where in the code the file is created, but I don't see where it's used...

The .tsv files are just checked in for auditing the process.  I think the data itself is used for testing and analysis of the relationship between languages.

macchiati (approved):
It's really hard to see the differences because the did is to costs and doesn't have names. Maybe gen the chart?

srl295:
OK. I'll merge and then open a PR to gen the chart.

srl295:
@macchiati PR with new chart https://github.com/unicode-org/cldr-staging/pull/83

PR #2749, opened 2023-02-23
CLDR-16425 v43 BRS: JSON

srl295:
- fixes for CoverageLevel reader
- add `und` to coverageLocales.txt for convenience 
- maybe a new command line option ✨ 

CLDR-16425

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
@@ -43,6 +43,7 @@ import com.ibm.icu.number.LocalizedNumberFormatter;
 import com.ibm.icu.number.NumberFormatter;
 import com.ibm.icu.number.Precision;
 import com.ibm.icu.util.NoUnit;
+import com.ibm.icu.util.ULocale;
 
 /**
  * Utility methods to extract data from CLDR repository and export it in JSON
@@ -60,7 +61,7 @@ public class Ldml2JsonConverter {
     enum RunType {
         all,
         main,
-        supplemental(false, false), // aka 'core'
+        supplemental(false, false), // aka 'cldr-core'
         segments, rbnf(false, true), annotations, annotationsDerived, bcp47(false, false);
 
         private final boolean isTiered;
@@ -145,7 +146,8 @@ public class Ldml2JsonConverter {
                 .add("identity", 'i', "(true|false)", "true",
                     "Whether to copy the identity info into all sections containing data")
                 .add("konfig", 'k', ".*", null, "LDML to JSON configuration file")
-                .add("pkgversion",  'V', ".*", getDefaultVersion(), "Version to be used in writing package files");
+                .add("pkgversion",  'V', ".*", getDefaultVersion(), "Version to be used in writing package files")
+                .add("Modern", 'M', "(true|false)", "true", "Whether to include the -modern tier");
 
     public static void main(String[] args) throws Exception {
         options.parse(args, true);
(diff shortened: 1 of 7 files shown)


srl295:
This causes `-full` to be basic level or better, and the optional (offby default) `-E` option adds `*-everything` packages that have everything.

tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
```diff
@@ -54,7 +54,8 @@
 @CLDRTool(alias = "ldml2json", description = "Convert CLDR data to JSON")
 public class Ldml2JsonConverter {
     private static final String CLDR_PKG_PREFIX = "cldr-";
-    private static final String FULL_TIER_SUFFIX = "-full";
```
macchiati, line 57: Can you explain more about this change and why you're doing it?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CalculatedCoverageLevels.java
```diff
@@ -13,6 +13,11 @@
 
 
 public class CalculatedCoverageLevels {
+    /**
+     * Assumed level for root. CLDR-16420
+     */
+    final private static Level DEFAULT_ROOT_LEVEL = Level.MODERN;
```
macchiati, line 19: This is very strange. Why modern? It's it so it is always included?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CalculatedCoverageLevels.java
```diff
@@ -26,6 +31,30 @@ public Map<String, Level> getLevels() {
...
+        return getEffectiveCoverageLevel(CLDRLocale.getInstance(locale));
+    }
+
+    public Level getEffectiveCoverageLevel(CLDRLocale locale) {
+        // per spec, assumed level for the explicit root locale
+        if (locale == CLDRLocale.ROOT) {
+            return DEFAULT_ROOT_LEVEL;
+        }
+        // See if there is an explicit entry
+        final Level level = levels.get(locale.getBaseName());
```
macchiati, line 44: It should follow the locale idparser getparent chain

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCalculatedCoverageLevels.java
```diff
@@ -21,4 +33,45 @@ void TestBasic() {
...
+        "sr_Cyrl, modern",
+        "sr_Cyrl_XK, modern",
+        "sr, modern",
+    })
+    void TestEffective(final String loc, String l) {
+        Level level = null;
+        if (!l.equals("null")) {
+            level = Level.fromString(l.toUpperCase());
+        }
+        CalculatedCoverageLevels ccl = CalculatedCoverageLevels.getInstance(); // TODO: data sensitivity.  Ideally would use a private instance of CCL from a static file.
```
macchiati, line 74: We're should move all the logic to CalculatedCoverageLevels

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/CalculatedCoverageLevels.java
srl295: Yes, so that applications can just filter on the effective level.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CalculatedCoverageLevels.java
srl295: It does via CLDRLocale

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCalculatedCoverageLevels.java
srl295: ? This is a unit test

tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
srl295: I think this is an outdated commit

macchiati (approved):

PR #2758, opened 2023-03-01
CLDR-15034 kbd: update DTD, import should have distinguishing attrs

srl295:
CLDR-15034

a few fixes:

- [X] update the DTD:  the attributes on `<import>` should be distinguishing, not 'value'.  Symptom: 'duplicate' errors from CLDR tests
- [x] update DTD with `@TECHPREVIEW` marker
- [x] update fr azerty so it builds. 
- …
- 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/keyboards/3.0/fr-t-k0-azerty.xml
+++ b/keyboards/3.0/fr-t-k0-azerty.xml
@@ -78,7 +78,6 @@
 		<key id="super-2" to="²" longPress="₂" />
 		<key id="hyphen" to="-" /> <!-- 6 -->
 		<key id="underscore" to="_" /> <!-- 8 -->
-		<key id="umlaut" to="¨" />
 		<key id="a" flicks="a" to="a" longPress="à â á ä ã å ā" />
 		<flicks id="a">
 			<flick directions="nw" to="\u1234" />
--- a/keyboards/3.0/fr-t-k0-azerty.xml
+++ b/keyboards/3.0/fr-t-k0-azerty.xml
@@ -5,6 +5,8 @@
     This is a sample data file.
     This file is subject to change.
     Please see https://cldr.unicode.org/index/keyboard-workgroup for the latest information.
+
+	Also NOTE: this is really a test keyboard.  CLDR-12026 will be for the real new azerty keyboard
 -->
 <keyboard locale="fr-t-k0-azerty" conformsTo="techpreview">
 	<locales>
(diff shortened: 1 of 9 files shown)


srl295:
at some point i need to fork off this `fr` keyboard and just call it "TEST" or something, and then implement the real azerty, https://unicode-org.atlassian.net/browse/CLDR-12026

keyboards/3.0/fr-t-k0-azerty.xml
```diff
@@ -52,19 +52,32 @@
...
+		<!-- switch keys -->
+		<key id="shift" switch="shift" />
+		<key id="numeric" switch="numeric" />
+		<key id="symbol" switch="symbol" />
+		<key id="base" switch="base" />
+
+
+		<key id="bksp" gap="true" /> <!-- TODO -->
+		<key id="extra" gap="true" /> <!-- TODO -->
+		<key id="enter" to="\u{000A}" /> <!-- TODO - implied key? -->
```
mcdurdin, line 67: The Enter key should not be generating a character -- that's an application-level decision. And certainly should not be defined in each keyboard; I can't see any need for Enter key to ever produce anything keyboard-specific (unlike spacebar).

keyboards/3.0/fr-t-k0-azerty.xml
```diff
@@ -52,19 +52,32 @@
...
+		<import base="cldr" path="techpreview/keys-Zyyy-currency.xml" />
+
+		<!-- switch keys -->
+		<key id="shift" switch="shift" />
+		<key id="numeric" switch="numeric" />
+		<key id="symbol" switch="symbol" />
+		<key id="base" switch="base" />
+
+
+		<key id="bksp" gap="true" /> <!-- TODO -->
```
mcdurdin, line 65: Why is bksp being defined as a gap? I don't think we should be able to redefine bksp or enter.

mcdurdin:
I note we have nothing in the spec at present about `bksp`, `enter`, or `extra` being predefined key identifiers. Something for discussion perhaps?

srl295:
Already on the agenda

keyboards/3.0/fr-t-k0-azerty.xml
srl295: il'l take these keys out for now

srl295:
merging this so i can make progress on tests. Feel free to post review or add comments here

PR #2759, opened 2023-03-01
CLDR-16399 en_CA revert short date format to 41

srl295:
- per TC 2023-02-27

CLDR-16399

ALLOW_MANY_COMMITS=true


common/main/en_CA.xml
```diff
@@ -1883,10 +1883,10 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						<dateFormatItem id="MMMMW" count="other">↑↑↑</dateFormatItem>
 						<dateFormatItem id="ms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="y">↑↑↑</dateFormatItem>
-						<dateFormatItem id="yM">↑↑↑</dateFormatItem>
-						<dateFormatItem id="yMd">↑↑↑</dateFormatItem>
+						<dateFormatItem id="yM">MM/y</dateFormatItem>
```
macchiati, line 1886: We should always have the order: y then M then d

Would be good to have a test for this, whenever the pattern contains M or MM.

macchiati (changes requested):

common/main/en_CA.xml
srl295: i'm not quite following.  the modified `y`'s are:

```xml
						<dateFormatItem id="y">↑↑↑</dateFormatItem>
						<dateFormatItem id="yM">MM/y</dateFormatItem>
						<dateFormatItem id="yMd">y-MM-dd</dateFormatItem>
						<dateFormatItem id="yMd" alt="variant">d/M/y</dateFormatItem>
						<dateFormatItem id="yMEd">E, y-MM-dd</dateFormatItem>
						<dateFormatItem id="yMEd" alt="variant">E, d/M/y</dateFormatItem>
```

Which one is out of order?

btangmu: Mark, do you mean MM/y should be y/MM?

srl295: ptal

btangmu (dismissed):

srl295:
@macchiati  look ok?

AEApple:
I thought we were waiting for @pedberg-icu to also review?

macchiati:
Right

common/main/en_CA.xml
```diff
@@ -1869,11 +1869,11 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						<dateFormatItem id="hmv">↑↑↑</dateFormatItem>
```
pedberg-icu, line 1869: Don't we also need to change the patterns for GyMd above (and also add patterns from GyM and GyMED which we are currently inheriting from en[_US])? And what about the intervalFormats for short dates, need to fix those too.

pedberg-icu (changes requested):
Need to fix more availableFormats items (GyM, GyMd, GyMEd), also intervalFormats for short dates. For thoroughness we should also fix the generic calendar formats but lower priority.

macchiati:
Note: 
There are already a small number of variants for dMy AND for yMd!!

Given this information, I suggest that we complete the set of variants for these paths, so we have:

⟪Value⟫ | Path
-- | --
⟪y/M/d⟫ | //ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/availableFormats/dateFormatItem[@id="yMd"]
⟪M/d/y⟫ | //ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/availableFormats/dateFormatItem[@id="yMd"][@alt="Mdy"]
⟪d/M/y⟫ | //ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/availableFormats/dateFormatItem[@id="yMd"][@alt="dMy"]
-- | --

srl295:
> Need to fix more availableFormats items (GyM, GyMd, GyMEd), also intervalFormats for short dates. For thoroughness we should also fix the generic calendar formats but lower priority.

OK. Thought I reverted all of them. but will check again

macchiati:
I have a complete list in the spreadsheet (
https://github.com/unicode-org/cldr/pull/2759#issuecomment-1452776336). You
can run SearchCLDR locally to see the outstanding items after your changes.

As to https://github.com/unicode-org/cldr/pull/2759#issuecomment-1452805154,
we would have to take that to the committee. Probably too late to do this
for v43, though.

srl295:
> Note: There are already a small number of variants for dMy AND for yMd!!
> 
> Given this information, I suggest that we complete the set of variants for these paths, so we have for each relevant path the 3 forms:

I'm not sure how much utility there would be in the variants. We should probably try to get consensus around a single variant besides the common value.  but, for later.

common/main/en_CA.xml
srl295: `GyMd` et al. wasn’t set in 41 and prior, so there was nothing to revert here. But I can revert that also.  The interval formats don't work as well because they already use dashes for ranges, thought we said we would push that to next time.

srl295:
the query above didn't catch this inherited item, that Peter noted

```xml
<dateFormatItem id="GyMd">M/d/y G</dateFormatItem>
```

or these

```xml
						<dateFormatItem id="GyMMMd">MMM d, y G</dateFormatItem>
						<dateFormatItem id="GyMMMEd">E, MMM d, y G</dateFormatItem>
```

common/main/en_CA.xml
srl295: @pedberg-icu  PTAL - see if this addresses it

srl295: en.xml for reference:

```xml
						<dateFormatItem id="Gy">y G</dateFormatItem>
						<dateFormatItem id="GyMd">d MMM y G</dateFormatItem>
						<dateFormatItem id="GyMMM">MMM y G</dateFormatItem>
						<dateFormatItem id="GyMMMd">d MMM y G</dateFormatItem>
						<dateFormatItem id="GyMMMEd">E, d MMM y G</dateFormatItem>
```

srl295:
OK.  I've updated to try to keep consistent y…M…d order, please take a look and comment.

- I thought from prior discussion that intervalFormats and other calendars should wait till next time
- I also left variants alone

common/main/en_CA.xml
```diff
@@ -1854,10 +1854,11 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						<dateFormatItem id="Ehms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="EHms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Gy">↑↑↑</dateFormatItem>
-						<dateFormatItem id="GyMd">↑↑↑</dateFormatItem>
-						<dateFormatItem id="GyMMM">↑↑↑</dateFormatItem>
-						<dateFormatItem id="GyMMMd">↑↑↑</dateFormatItem>
-						<dateFormatItem id="GyMMMEd">↑↑↑</dateFormatItem>
+						<dateFormatItem id="GyMd">y-MM-dd G</dateFormatItem>
+						<dateFormatItem id="GyM">y-MM G</dateFormatItem>
+						<dateFormatItem id="GyMMM">y MMM G</dateFormatItem>
```
pedberg-icu, line 1859: What are we changing the medium-style formats here? I thought those were supposed to inherit from en. It is only the short (numeric) dates that should have y MM dd order.

pedberg-icu (changes requested):
We should leave the medium-style forms (with MMM) alone and just change the short (numeric) formats to yMd order.

common/main/en_CA.xml
macchiati: Good catch, we only want to change the MM or M patterns.

My list (https://github.com/unicode-org/cldr/pull/2759#issuecomment-1452934163) just had the numeric formats. Not sure where the MMM's snuck in.

macchiati: FYI: there are other languages that have different order for numeric dates than they do for non numeric (eg MMM+), like Swedish.

srl295: OK, will drop the MMM.  I thought they also needed to be in the same order.

common/main/en_CA.xml
srl295: GyMd now changed. Did not find GyM or GyMED

macchiati:
Did you check en.xml and root.xml?

srl295:
For the Gregorian calendar, yes, did not see any other patterns that needed to be updated

pedberg-icu (approved):
OK, this looks good now, thanks

AEApple:
@macchiati - Can you re-approve or confirm that we should dismiss your earlier review since @pedberg-icu has already approved. It's not clear to me if anything further is needed?

srl295:
@macchiati ping?

srl295:
> Did you check en.xml and root.xml?

For the record,

I checked all of the `calendar="gregorian"` availableFormats for en.xml and root.xml, there were no remaining items which would result in slashes or hypens in the wrong order (y-M-d).  All problematic items have been overridden as shown in the diff.

macchiati:
> We should leave the medium-style forms (with MMM) alone and just change the short (numeric) formats to yMd order.

Has this been done?

srl295:
> > We should leave the medium-style forms (with MMM) alone and just change the short (numeric) formats to yMd order.
> 
> 
> 
> Has this been done?

I believe so

macchiati (approved):

PR #2761, opened 2023-03-06
CLDR-16438 document filtering out unneeded data

macchiati:
This is for tickets https://unicode-org.atlassian.net/browse/CLDR-16438 and https://unicode-org.atlassian.net/browse/CLDR-16340.

CLDR-16438

ALLOW_MANY_COMMITS=true


docs/ldml/tr35.md
```diff
@@ -3170,6 +3173,41 @@ Processes sometimes encounter invalid number or date patterns, such as a number
...
+
+Locale data can be _sliced_ to exclude data not needed by a particular implementation. This can be _vertical slicing_: excluding a locale and all the locales inheriting from them, or _horizontal slicing_: excluding particular types of data from all locales. For example:
+  * A vertical slice could retain only those locales used in a particular set of markets, such as EU locales.
+  * A horizontal slice could remove all data in the emoji/ directory, which are annotations for emoji and symbols.
+  
+Of course, both of these techniques can be applied.
+
+### 8.1 <a name="Vertical_Slicing" href="#Vertical_Slicing">Vertical Slicing</a>
+
+It is important to note that if a particular locale is in a vertical slice, then all of its parents should be as well, because of inheritance. This is not a factor if the data is fully resolved, as in Javascript.
```
srl295, line 3187:
```suggestion
It is important to note that if a particular locale is in a vertical slice, then all of its parents should be as well, because of inheritance. This is not a factor if the data is fully resolved, as in the JSON format data.
```

correct name

docs/ldml/tr35.md
```diff
@@ -3170,6 +3173,41 @@ Processes sometimes encounter invalid number or date patterns, such as a number
...
+
+2. Filter out the locales as per #1, except also retain also those locales representing the official languages for for countries, based on data in supplementalData.xml. For a human-readable chart, see [Territory-Language Information](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_language_information.html).
+
+### 8.2 <a name="Horizontal_Slicing" href="#Horizontal_Slicing">Horizontal_Slicing</a>
+
+The Comprehensive coverage level includes data that is above the Modern coverage level. This data is over and above what is typically needed by many applications. That data can be filtered out, at considerable savings in data size. The easiest way to do that is to use the CLDR tooling to filter the data before generating the implementation's data format. That way allows the implementation to have direct access to the CoverageLevel code that can determine the coverage level, for a given locale and path. Once the data is transformed, such as to the JSON format, the CoverageLevel code is no longer accessible. For example, here is a code snippet:
```
srl295, line 3198:
```suggestion
Paths at the Comprehensive [coverage level](tr35-info.md#Coverage_Levels) include data that is above the Modern coverage level. This data is over and above what is typically needed by many applications. That data can be filtered out, at considerable savings in data size. The easiest way to do that is to use the CLDR Java tooling (the `cldr-code` package) to filter the data before generating the implementation's data format. That way allows the implementation to have direct access to the CoverageLevel code that can determine the coverage level, for a given locale and path. Once the data is transformed, such as to the JSON format, the CoverageLevel code is no longer accessible. For example, here is a code snippet:
```

We use “coverage” to mean a couple of things (either a horizontal or vertical slice).  By mentioning _paths_ it clarifies that this is a horizontal slice.

Also, we don't talk a lot about the Java tools in the spec. That should perhaps be a link.

srl295 (approved):
Definite improvement.

At some point we should make sure the wording here is harmonized with https://cldr.unicode.org/index/cldr-spec/coverage-levels for example.

Also the Coverage level section I linked to says:

> That structure is primarily intended for internal use in CLDR tooling — it is not anticipated that users of CLDR data would need it.

May need to revise the above if they are no longer internal.

docs/ldml/tr35.md
```diff
@@ -3170,6 +3173,41 @@ Processes sometimes encounter invalid number or date patterns, such as a number
...
+
+Slicing can also remove related supplemental data. For example, the likely subtags data includes a large number of languages that may not be of interest for all implementations. Here are some sample techniques can be used to filter the data so that the number of mappings is reduced:
+
+1. Only keep likely subtags data with for locales that are used in the implementation.
+  * For example, if the implementation only includes the CLDR locales at Basic coverage in [Unicode CLDR - Coverage Levels](https://cldr.unicode.org/index/cldr-spec/coverage-levels) (and locales inheriting from them), then filter out all likely subtag data that doesn’t match. 
+
+2. Filter out the locales as per #1, except also retain also those locales representing the official languages for for countries, based on data in supplementalData.xml. For a human-readable chart, see [Territory-Language Information](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_language_information.html).
```
sffc, line 3194:
```suggestion
2. Filter out the locales as per #1, except also retain also those locales representing the official languages for countries, based on data in supplementalData.xml. For a human-readable chart, see [Territory-Language Information](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_language_information.html).
```

srl295 (approved):
LGTM.  I expect we may have some bumps and questions but they can shake out.

docs/ldml/tr35.md
sffc, line 3194: Can you link to a section of the spec that discusses how official languages are encoded?

For example, is the recommendation to include _all_ official languages in supplementalData.xml, or is there an additional hierarchy of which official languages are most important to include from there? For example, many of the official languages are listed with fewer than 1% of speakers in their respective region, and some with 0 speakers.

In other words, I see three tiers of likely subtags applications may want to include:

1. Minimal set required for rendering the application (CLDR locales).
2. Larger set for better resolution of regional preferences and language matching.
3. Full set for maximal coverage in applications that have no size constraints.

I think when we are talking about including official languages, we're talking about use case 2. What precisely is the best way to define what set is most useful for that typical use case?

docs/ldml/tr35.md
```diff
@@ -3170,6 +3173,41 @@ Processes sometimes encounter invalid number or date patterns, such as a number
...
+
+### 8.2 <a name="Horizontal_Slicing" href="#Horizontal_Slicing">Horizontal_Slicing</a>
```
sffc, line 3196: The main reason to perform horizontal slicing is when an application wants to remove data required for powering some feature. If an app isn't performing date formatting, it can remove all transitive date formatting data, except if it is formatting currencies, it needs to retain some number formatting data. I don't understand how coverage levels play into this.

sffc (changes requested):

docs/ldml/tr35.md
macchiati: Good question. 

It might be better for now to just replace #2 (because it is more complicated) with a broader note that:

What is chosen to filter depends very much on the application. Some information that might be useful for that is the [Supplemental Territory Information](https://unicode.org/reports/tr35/tr35-info.html#Supplemental_Territory_Information), which provides information on the use of languages in different countries/regions.

macchiati: Revamped this part

docs/ldml/tr35.md
macchiati: Good point. Incorporated it, and also clarified the use of coverage information.

sffc (dismissed):
This looks good. Definitely good to explicitly state this in the spec.

I think there's room to improve this section further (more emphasis on explicit algorithms rather than left-to-the-reader guidelines), but that needs more thought and doesn't need to be done right away.

macchiati:
Sorry, had wrong target. Can you restamp?

macchiati:
Messed up in github, so created new PR https://github.com/unicode-org/cldr/pull/2766

macchiati:
Good. I agree that we can flesh this out over time. We might even want to
break out a new "Part" that has implementation techniques, where we can
gather them all together.

srl295:
> Good. I agree that we can flesh this out over time. We might even want to break out a new "Part" that has implementation techniques, where we can gather them all together.

It's a trade off. That was the goal of the implementers guide ( [stub](https://github.com/unicode-org/cldr-implementers-guide/blob/main/FAQ.md) | [toc](https://github.com/unicode-org/cldr-implementers-guide/issues/5#issuecomment-1458520271) ) … I'm not sure the TR is the right place for too much implementation guidance vs. some more informal cookbook.  However, no other vehicles have 'stuck' in the meantime.

macchiati:
We could have an additional github document, eg LDML Implementers Guide

PR #2765, opened 2023-03-07
CLDR-16433 Document changes in person names

macchiati:
CLDR-16433

1. Removed prefixed section numbers from anchors (unfortunately, this caused many lines to change)
2. Dropped that this is tech preview
3. Changed prefix and suffix fields to title, generation, and credentials
4. Changed 6.5 to cover 2 topics, and made them 6.5.1 and 6.5.2
5. Described how to coalesce literals (and revamped the rest of the description)
6. Fixed the sample name categories
7.  Deleted some examples in 8.2 that aren't needed or are not explained clearly.
8. Fix the second on locales with scripts that don't need spaces between words.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-personNames.md
+++ b/docs/ldml/tr35-personNames.md
@@ -2,7 +2,7 @@
 
 # Unicode Locale Data Markup Language (LDML)<br/>Part 8: Person Names
 
-|Version|43 (draft)              |
+|Version|43 (draft)                      |
 |-------|------------------------|
 |Editors|Mark Davis, Peter Edberg,  Rich Gillam, Alex Kolisnychenko, Mike McKenna and <a href="tr35.md#Acknowledgments">other CLDR committee members</a>|
 
@@ -96,8 +96,6 @@ There is a wide variety in the way that people’s names appear in different lan
 
 This document provides the [LDML](http://www.unicode.org/reports/tr35/) specification for formatting of personal names, using data, structure, and examples. 
 
-> **This is a technology preview; thus not intended for production software (except where itself marked as a technology preview). We have gathered a first round of data, and intend to refine the way in which we gather data. We are looking for additional feedback on the tech preview so that we can make improvements. For example, there are a few areas where we intend enhancements: handling native vs foreign names (in the native script); handling prefix and suffix fields; and so on.**
-
 The CLDR functionality is targeted at formatting names for typical usage on computers (e.g. contact names, automated greetings, etc.), rather than being designed for special circumstances or protocol, such addressing royalty. However, the structure may be enhanced in the future when it becomes clear that additional features are needed for some languages.
 
 This addition to CLDR is based on review of current standards and practices that exist in LDAP, hcard, HTML and various other international standards and commercial implementations.
@@ -110,8 +108,7 @@ The following features are currently out of scope for Person Names formating:
 
 * Grammatical inflection of formatted names.
 * Context-specific cultural aspects, such as when to use “-san” vs “-sama” when addressing a Japanese person. 
-* Providing lists of prefixes and suffixes (Mr, Ms., Mx., Dr., etc.).
-* Distinctions among prefixes and suffixes, such as title (Dr., Prof., Sir), gender-honorifics, generation (Jr., IV), accreditations (MBA, Esq.).
+* Providing lists of titles, generation terms, and credential (Mr, Ms., Mx., Dr., Jr., M.D., etc.).
 * Validation of input, such as  which fields are required, and what characters are allowed.
 * Combining alternative names, such as multicultural names in Hong Kong "[Jackie Chan Kong-Sang](https://en.wikipedia.org/wiki/Jackie_Chan)”, or ‘Dwayne “The Rock” Johnson’.
 * More than two levels of formality for names.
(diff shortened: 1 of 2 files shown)


docs/ldml/tr35-personNames.md
```diff
@@ -346,9 +344,9 @@ The order attribute is used for patterns with different orders of fields. The or
 
 For example, when the display language is Japanese, it is customary to use _surnameFirst_ for names of people from Japan and Hungary, but use _givenFirst_ for names of people from the United States and France. Although the English pattern for sorting is distinct from the other patterns (except for unusual names), that is not necessarily the case in other languages.
 
-### 4.2 <a name="4-2-length" href="#4-2-length">length</a>
+### 4.2 <a name="length" href="#length">length</a>
 
-The `length` attribute specifies the relative length of a formatted name depending on context. For example, a `long` formal name in English would include prefix, given, given2, surname plus suffix; whereas a `short` informal name may only be the given name.
+The `length` attribute specifies the relative length of a formatted name depending on context. For example, a `long` formal name in English might include title, given, given2, surname plus generation and credentials; whereas a `short` informal name may only be the given name.
```
richgillam, line 349: Would this read better if you had the formality in the code typeface too (i.e., `long formal` and `short informal`)?

docs/ldml/tr35-personNames.md
```diff
@@ -746,35 +755,43 @@ The following process is used to produce initials when they are not supplied by
...
 
-2. If a locale **doesn’t** require spaces between words, there are two cases, based on whether the foreign name is written in the locale’s script, or the foreign name is left in its native script. In both cases, patterns from the **locale of the name** are used. For example, the formatting locale might be Japanese, and the locale of the PersonName object might be de_CH, German (Switzerland), such as Albert Einstein.
+2. If a locale **doesn’t** require spaces between words, there are two cases, based on whether the name is foreign or not (based on the PersonName objects explicit or calculated locale's language subtag). For example, the formatting locale might be Japanese, and the locale of the PersonName object might be de_CH, German (Switzerland), such as Albert Einstein.
 
-    1. **The foreign name is written in the formatting locale’s script.** In that case, the **foreignSpaceReplacement** is substituted for each space in the patterns from the _locale of the name_. Here are examples for Albert Einstein in Japanese and Chinese:
+    1. **The name is native** In that case, the patterns are used as is.
```
richgillam, line 782: What happened to `nativeSpaceReplacement`?

Also, I think the way I implemented cross-script formatting differs a bit from what you have here.

I think what I've got right now is:
1. If the formatter locale uses spaces, don't do anything special.
2. If the formatter locale doesn't use spaces, the name locale does use spaces, and the actual text of the name is in a script that is not native to the formatter locale, use the name locale's formatting patterns instead of the formatter locale's patterns.
3. Otherwise, use the formatter's patterns to format the name.  Afterwards, if the name locale and formatter locale agree, apply the `nativeSpaceReplacement`, and if they don't, apply the `foreignSpaceReplacement`.

docs/ldml/tr35-personNames.md
```diff
@@ -746,35 +755,43 @@ The following process is used to produce initials when they are not supplied by
...
-        * [Albert Einstein](https://de.wikipedia.org/wiki/Albert_Einstein) 
+
 
 In both cases, the ordering may be changed according to the **Name Order for Locales** settings that each locale provides. If the PersonName object does not supply a locale for a name, then a default locale will be derived based on other information (such as the script of the characters in the name fields).
 
-> **Note** In the tech preview, the structure isn't yet powerful enough to handle cases with `foreignSpaceReplacement` where the formatting locale doesn’t need spaces between words, but the name locale has the same ordering as the formatting locale. 
-> For example, consider where the formatting locale is Thai, and the name is in English, but transliterated into Thai.
+Remember that **a name in a different script** will use a different locale for formatting. 
+For example, when formatting a name for Japanese, if the name is in the Latin script, a Latin based locale will be used to format it, such as when “Albert Einstein” appears in Latin characters on [Albert Einstein](https://ja.wikipedia.org/wiki/Albert_Einstein)
```
richgillam, line 791: See my comments above.  I think this text needs to be incorporated into the description above, and I don't think you can separate out the separate-script case as cleanly as you're doing here.

richgillam (approved):
Looks really good.  I have a few comments on the handling of foreign names.

docs/ldml/tr35-personNames.md
macchiati: good idea. Might be better to do this in a follow-on PR. (I think it was a mistake to fix the malformed anchors in this PR because it makes it harder to see the material changes)

docs/ldml/tr35-personNames.md
macchiati: I agree that it needs to be reworked. Here's my trial.

1. If the name locale's script is different than the formatter locale's script, then use a formatter for the name locale. So when a Hindi formatter hits a name in Cyrillic, the code creates and uses a Cyrillic formatter to format that name. So we can assume for the rest of this discussion that the Hindi formatter will only need to handle names in Hindi script.

2. There are then two characteristics that need to be established.
* what is the nameOrder? — affects which patterns are chosen
* what is the spaceReplacement? —  affects post-processing of the composed string

3. The nameOrder is based on a lookup of the name locale in the name order data in the formatting locale the. (It isn't a simple lookup, because we follow the parent chain until there is a map.

4. The spaceReplacement is a choice between the nativeSpaceReplacement and the foreignSpaceReplacement, also based on the locale of the formatter and locale of the name. Most locales have " " = nativeSpaceReplacement = foreignSpaceReplacement, so they are easy.

* The foreignSpaceReplacement is provided by data in the locale.
*The nativeSpaceReplacement is currently calculated, based on the LineBreak information for the script. It is either " " or "".

The choice of which to use for the spaceReplacement is calculated: if the language subtag of the formatter's locale equals the language subtag of the name's locale, then the nativeSpaceReplacement is used, otherwise the foreignSpaceReplacement.
 
(Ideally we should add some more structure to make spaceReplacement more data-driven — filed https://unicode-org.atlassian.net/browse/CLDR-16457)

macchiati:
Rich, since you ok'd this, I'm going to go ahead and merge. Will be easier to apply changes then on top of what is done.

docs/ldml/tr35-personNames.md
richgillam: This isn't what I have implemented on the ICU side, and I think it bears some discussion.  I also don't think step 1 is quite right, because it doesn't draw a distinction between the name **locale's** script and the **name's** script.  What we originally had was that we'd use a different formatter if the name locale had a different script **AND** and name itself was using that script (or something other than the formatter locale's script).  If you had a foreign name that had been transliterated into the native script, you kept the current formatter and used the foreignSpaceReplacement.  And is the CLDR code doing this at all right now?  I didn't see anything, but maybe I missed it.  I also think we need more test cases around this, on both the ICU and CLDR sides.

docs/ldml/tr35-personNames.md
```diff
@@ -48,37 +48,37 @@ The LDML specification is divided into the following parts:
...
-* 7 [Sample Name](#7-sample-name)
-  * 7.1 [Syntax](#7-1-syntax)
-  * 7.2 [Expected values](#7-2-expected-values)
-* 8 [PersonName Data Interface Examples](#8-personname-data-interface-examples)
+  * 6.7 [Deriving initials](#deriving-initials)
+  * 6.8 [Handling foreign names](#handling-foreign-names)
+* 7 [Sample Name](#sample-name)
+  * 7.1 [Syntax](#syntax)
+  * 7.2 [Expected values](#expected-values)
+* 8 [PersonName Data Interface Examples](#personname-data-interface-examples)
```
srl295, line 81: @macchiati  it looks like the ToC was perhaps updated 'by hand' when it should have been regenerated using the script, I'l open a PR

macchiati: I didn't realize it was generated. That would have saved time trying (and failing) to keep it updated.

PR #2768, opened 2023-03-07
CLDR-16433 Clean up language around spaceReplacement

macchiati:
CLDR-16433

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-personNames.md
+++ b/docs/ldml/tr35-personNames.md
@@ -479,25 +479,31 @@ The patterns are in personName elements, which are themselves in a personNames c
 
 The details of the XML structure behind the data referenced here are in [XML Structure](#xml-structure).
 
-### 6.1 <a name="derive-the-name-locale" href="#derive-the-name-locale">Derive the name locale</a>
+### 6.1 <a name="derive-the-name-locale" href="#derive-the-name-locale">Derive the name locale information</a>
 
-Create a **full name locale** as follows.
+Let the **full formatting locale** be the fully-fleshed-out formatting locale, using likely subtags.
 
-1. First, let the **full formatting locale** be the fully-fleshed-out formatting locale using likely subtags.
-2. If there is a name locale available via the PersonName data interface, obtain the full name locale from the name locale using likely subtags. Thus de ⇒ de_Latn_DE.
-3. Otherwise the full name locale is created based on the characters in the name and the full formatting locale, as follows:
-    1. Find the predominant script for the name in the following way.
-        1. For each character in the given and surname, find the script(s) of the character using the Script_Extensions property.
-        2. For each of those scripts, increment a counter for that script, and record the position of the first character encountered in that script.
-    2. The predominant script is the script with the highest counter value.
-        1. In the rare case that there are multiple counters with the highest counter value, take the one with the lowest first position.
-        2. In the even rarer case that there is still more than one, use the script whose script code is alphabetically lowest. (These two steps are simply to guarantee a determinant result.)
-    3. If the predominant script is the same as the script of the full formatting locale, then let the full name locale be identical to the full formatting locale.
-    4. Otherwise, find the likely locale for the predominant script, as specified by the likely subtags. (This will add a language and region.) Let the full name locale be that likely locale.
+Note that a few script values from likely subtags indicate a set of scripts, such as Jpan = {Hani, Kana, Hira}. Two script codes are said to _match_ when they are either identical, or their corresponding sets match. For example, Hani and Jpan match.
+
+Let the **full name locale** be the fully-fleshed out name locale obtained via the PersonName data interface, then applying likely subtags
+
+Find the **name script** in the following way.
+1. Create a counter (a map from scripts to integers)
+2. Iterate through the characters of the surname, then through the given name.
+    1. Find the script set of that character using the Script_Extensions property.
+    2. If the script set has one element, and it is not Common, Inherited, nor Unknown, return that element as the name script
+3. If nothing is found during the iteration:
+    1. The name script is set to the full name locale's script, if there is one.
+    2. Otherwise, name script is set to the full formatting locale's script. 
+
+<!-- not sure we will need this, after consulting with Rich -->
+Find the **name language** in the following way.
+1. If there is a full name locale, return its language.
+2. Otherwise, find the likely locale for name script, and return its language.
 
 ### 6.2 <a name="derive-the-formatting-locale" href="#derive-the-formatting-locale">Derive the formatting locale</a>
 
-If the full name locale has a different script than that of the full formatting locale, then the name formatted with a different formatter than originally requested. For example, when a Hindi formatter is called upon to format a name with the Ukrainian (Cyrillic) locale, under the covers a Ukrainian (Cyrillic) formatter should be instantiated to format that name.
+If the name script doesn't match the script of the full formatting locale, then the name formatted with a different formatter than originally requested. For example, when a Hindi formatter is called upon to format a name with the Ukrainian (Cyrillic) locale, under the covers a Ukrainian (Cyrillic) formatter should be instantiated to format that name.
 
 So in all of the processing below the "name locale" is the full name locale, and the "formatting locale" is the full formatting locale, one that is fully-fleshed out using likely subtags. Moreover the name locale has the same script as the formatting locale.
 
--- a/docs/ldml/tr35-personNames.md
+++ b/docs/ldml/tr35-personNames.md
@@ -612,7 +612,16 @@ For example:
 3. Pattern C is discarded, because it has the least number of populated name fields.
 4. Out of the remaining patterns A and B, pattern B wins, because it has only 3 unpopulated fields compared to pattern A.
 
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-personNames.md
```diff
@@ -763,29 +762,33 @@ There are two main challenges in dealing with foreign name formatting that needs
...
+
+1. The foreignSpaceReplacement is provided by the value for the `foreignSpaceReplacement` element; the default value is " ".
+
+2. The nativeSpaceReplacement is determined by the following algorithm, chosing between " " and "".
+
+1. Get the script of the formatting locale
+2. If the likely script is Thai, let nativeSpaceReplacement = " "
+3. Otherwise let nativeSpaceReplacement = "" if either of the following applies:
+    1. The script is Jpan, Hant, or Hans
+    2. The script has the script metadata property lbLetters = YES (this can also be algorithmically derived from the LineBreak property data).
```
richgillam, line 786: I think we have this language in two places.  It might be nice to have a separate set of steps for "does this locale uses spaces in names?"

docs/ldml/tr35-personNames.md
```diff
@@ -763,29 +762,33 @@ There are two main challenges in dealing with foreign name formatting that needs
...
+5. Otherwise, let nativeSpaceReplacement = ""
+
+3. If the language subtag of the formatter's locale equals the language subtag of the name's locale, then let spaceReplacement = nativeSpaceReplacement, otherwise let spaceReplacement = foreignSpaceReplacement.
```
richgillam, line 789: That's not what I implemented, and I'm not sure it's what we want.  See my comments in my PR.

docs/ldml/tr35-personNames.md
```diff
@@ -763,29 +762,33 @@ There are two main challenges in dealing with foreign name formatting that needs
...
+4. Replace all sequences of space in the resolved pattern string by the spaceReplacement.
```
richgillam, line 790: Is that replace all _sequences_ of spaces in the resolved pattern with a _single_ spaceReplacement?  I'm not doing that right now, either, but if we make sure the locale data is properly vetted, it won't matter.

richgillam (commented):
I'm approving this, but if you can hold it until the meeting, I think we should talk about it first.

macchiati:
We can definitely discuss in the meeting!

docs/ldml/tr35-personNames.md
macchiati: I think it is only in one place (but my eyes might be missing it!)

docs/ldml/tr35-personNames.md
macchiati: Right, it would only matter for non-properly vetted patterns, so no need to change your implementation.

docs/ldml/tr35-personNames.md
richgillam: Okay, I've gone back and updated https://github.com/unicode-org/icu/pull/2356 based on our discussion today.  I think it's a lot simpler.  I changed a bunch of stuff in my unit tests to match the new logic, but I think the changes make sense, and everything in the `ExhaustivePersonNameFormatterTest` that passed before still passes.

The one thing I'm not happy with is the logic to decide whether to use the foreign or native space replacement character.  I decided you were right that looking for common scripts between the locales was a bad idea, so I'm just looking at the language, but then I had to put a hack in there to get Japanese and Chinese to format correctly in each other's locales, and I really don't like that hack.  If you've got a smarter idea, I'm all ears.

docs/ldml/tr35-personNames.md
```diff
@@ -616,16 +619,16 @@ If the “winning” namePattern still has fields that are unpopulated in the Pe
...
-  2. If there is a single empty field, it is removed.
+    1. If there are two or more empty fields separated only by literals, the fields and the literals between them are removed.
+    2. If there is a single empty field, it is removed.
 4. If the processing from step 3 results in two adjacent literals (call them A and B), they are coalesced into one literal as follows:
-  1. If either is empty the result is the other one.
-  3. If B matches the end of A, then the result is A. So xyz + yz ⇒ xyz, and xyz + xyz ⇒ xyz.
-  4. Otherwise the result is A + B, but with any sequence of two or more white space characters replaced by the first whitespace character. 
+    1. If either is empty the result is the other one.
+    2. If B matches the end of A, then the result is A. So xyz + yz ⇒ xyz, and xyz + xyz ⇒ xyz.
+    3. Otherwise the result is A + B, further modified by replacing with any sequence of two or more white space characters by the first whitespace character.
```
richgillam, line 627: I think you have an extra "with" in here.  I'm guessing this should read "...further modified by replacing any sequence of two or more white space characters by the first whitespace character."

docs/ldml/tr35-personNames.md
```diff
@@ -780,19 +781,18 @@ Here are examples for Albert Einstein in Japanese and Chinese:
...
-    2. The script has the script metadata property lbLetters = YES (this can also be algorithmically derived from the LineBreak property data).
-5. Otherwise, let nativeSpaceReplacement = ""
-
-3. If the formatter base language equals the name base language, then let spaceReplacement = nativeSpaceReplacement, otherwise let spaceReplacement = foreignSpaceReplacement.
+    1. Get the script of the formatting locale
+    2. If the likely script is Thai, let nativeSpaceReplacement = " "
+    3. Otherwise let nativeSpaceReplacement = "" if either of the following applies:
+        1. The script is Jpan, Hant, or Hans
+        2. The script has the script metadata property lbLetters = YES (this can also be algorithmically derived from the LineBreak property data).
+    4. Otherwise, let nativeSpaceReplacement = ""
```
richgillam, line 790: I'm pretty sure you mean "Otherwise, let nativeSpaceReplacement = `" "`.  As it stands, the nativeSpaceReplacement character is the empty string unless the script is Thai.

richgillam (dismissed):
A couple typos, but otherwise looks good.

docs/ldml/tr35-personNames.md
```diff
@@ -607,23 +612,23 @@ For example:
 3. Pattern C is discarded, because it has the least number of populated name fields.
 4. Out of the remaining patterns A and B, pattern B wins, because it has only 3 unpopulated fields compared to pattern A.
 
-#### 6.5.1 <a name="process-the-namepattern" href="#process-the-namepattern">Choose a namePattern</a>
+#### 6.5.2 <a name="process-the-namepattern" href="#process-the-namepattern">Process a namePattern</a>
 
 If the “winning” namePattern still has fields that are unpopulated in the PersonName object, we process the pattern algorithmically with the following steps:
```
macchiati, line 617: Just spotted another problem. 

If the PersonName P has no surname and the pattern has no given name or only a given with initial modifiers, we (logically) create a new PersonName P2 that returns P's given name when asked for the surname, and nothing when asked for the given. (use better wording)

richgillam: Yup.  Totally forgot about that.  Good catch.

macchiati: See how it looks. I inserted a new 6.5.2, and renumbered the old.

macchiati: Forgot to update the TOC; so doing that now.

docs/ldml/tr35-personNames.md
```diff
@@ -612,7 +612,16 @@ For example:
 3. Pattern C is discarded, because it has the least number of populated name fields.
 4. Out of the remaining patterns A and B, pattern B wins, because it has only 3 unpopulated fields compared to pattern A.
 
-#### 6.5.2 <a name="process-the-namepattern" href="#process-the-namepattern">Process a namePattern</a>
+#### 6.5.2 <a name="missing-surname" href="#missing-surname">Handle missing surname</a>
+
+All PersonName objects will have a given name (for mononyms the given name is used). However, there may not be a surname. In that case, the following process is followed so that formatted patterns produce reasonable results.
+
+1. If there is no surname from a PersonName P1 _and_ there is no non-initialized given name in the pattern, then
```
richgillam, line 619: "non-initialized" reads weird to me, because I want to use the standard computer-science meaning of "initialize".  I'd suggest using more words here: "If there is no surname from a PersonName P1 and the pattern either doesn't include the given name or only shows an initial for the given name, then..."

richgillam (approved):
One nit.  Otherwise looks good.

richgillam (approved):
Sold!

docs/ldml/tr35-personNames.md
macchiati: I only found one place where there is an algorithm for using foreignSpaceReplacement

docs/ldml/tr35-personNames.md
macchiati: I kept the hack in the spec; for a data-driven solution we'll have to wait for v44

richgillam (approved):

macchiati:
Made one more change, to fix the wording of non-initialized. If it all looks ok, then let's merge this and then take further comments.

richgillam (approved):

docs/ldml/tr35-personNames.md
srl295: see #2770  (but merge this first). 
I'll drop the numbering once the dust settles on these

PR #2770, opened 2023-03-09
CLDR-16122 v43 spec: regenerate ToC

srl295:
CLDR-16122

- also drop some extra whitespace
- regen ToC using `cd tools/scripts/tr-archive ; npm ci ; npm run fix-tocs`
- fix the generator tool to not add extra junk on top of the tr35 ToC

Note that this tool runs automatically before the HTML is published.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-personNames.md
+++ b/docs/ldml/tr35-personNames.md
@@ -78,7 +78,7 @@ The LDML specification is divided into the following parts:
   * 6.7 [Process a namePattern](#process-the-namepattern)
     * 6.7.1 [Handling foreign names](#handling-foreign-names)
     * 6.7.2 [Setting the spaceReplacement](#spaceReplacement)
-    * 6.7.3 [Examples of space replacement](#examples-examples)
+    * 6.7.3 [Examples of space replacement](#spaceReplacement-examples)
   * 6.8 [Formatting examples](#formatting-examples)
 * 7 [Sample Name](#sample-name)
   * 7.1 [Syntax](#syntax)
@@ -710,7 +710,7 @@ For the purposes of this algorithm, two base languages are said to __match__ whe
 Remember that **a name in a different script** will use a different locale for formatting, as per <a href="#switch-formatting-locales">Switch the formatting locale if necessary</a>.
 For example, when formatting a name for Japanese, if the name is in the Latin script, a Latin based locale will be used to format it, such as when “Albert Einstein” appears in Latin characters on [Albert Einstein](https://ja.wikipedia.org/wiki/Albert_Einstein)
 
-#### 6.7.3 <a name="examples-examples" href="#spaceReplacement-examples">Examples of space replacement</a>
+#### 6.7.3 <a name="spaceReplacement-examples" href="#spaceReplacement-examples">Examples of space replacement</a>
 
 To illustrate how foreign space replacement works, consider the following name data. For illustration, the name locale is given in the maximized form: in practice, `ja` would be used instead of `ja_Jpan_JP`, and so on.: For more information, see [Likely Subtags](tr35.html#Likely_Subtags).
 
--- a/docs/ldml/tr35-personNames.md
+++ b/docs/ldml/tr35-personNames.md
@@ -67,31 +67,33 @@ The LDML specification is divided into the following parts:
 * 6 [Formatting Process](#formatting-process)
   * 6.1 [Derive the name locale](#derive-the-name-locale)
   * 6.2 [Derive the formatting locale](#derive-the-formatting-locale)
-      * 6.2.1 [Switch the formatting locale if necessary]("switch-formatting-locales)
+    * 6.2.1 [Switch the formatting locale if necessary](#switch-formatting-locales)
   * 6.3 [Derive the name order](#derive-the-name-order)
   * 6.4 [Choose a personName](#choose-a-personname)
-  * 6.5 [Choose a namePattern](#choose-a-namepattern")
-  * 6.6 [Access PersonName object](#access-personname")
-      * 6.6.1 [Handle missing surname](#missing-surname")
-      * 6.6.2 [Handle core and prefix](#core-and-prefix)
-      * 6.6.3 [Derive initials](#derive-initials)
+  * 6.5 [Choose a namePattern](#choose-a-namepattern)
+  * 6.6 [Access PersonName object](#access-personname-object)
+    * 6.6.1 [Handle missing surname](#missing-surname)
+    * 6.6.2 [Handle core and prefix](#core-and-prefix)
+    * 6.6.3 [Derive initials](#derive-initials)
   * 6.7 [Process a namePattern](#process-the-namepattern)
-      * 6.7.1 [Handling foreign names](#handling-foreign-names)
-      * 6.7.2 [Setting the spaceReplacement](#spaceReplacement)
-      * 6.7.3 [Examples of space replacement](#spaceReplacement-examples)
+    * 6.7.1 [Handling foreign names](#handling-foreign-names)
+    * 6.7.2 [Setting the spaceReplacement](#spaceReplacement)
+    * 6.7.3 [Examples of space replacement](#examples-examples)
   * 6.8 [Formatting examples](#formatting-examples)
 * 7 [Sample Name](#sample-name)
   * 7.1 [Syntax](#syntax)
   * 7.2 [Expected values](#expected-values)
 * 8 [PersonName Data Interface Examples](#personname-data-interface-examples)
+  * 8.1 [Example 1](#example-1)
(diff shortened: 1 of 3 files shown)


docs/ldml/tr35.md
```diff
@@ -3174,48 +3174,48 @@ Processes sometimes encounter invalid number or date patterns, such as a number
...
-Slicing can also remove related supplemental data. 
-For example, the likely subtags data includes a large number of languages that may not be of interest for all implementations. 
-Where an the implementation only includes (say) the CLDR locales at Basic coverage in [Unicode CLDR - Coverage Levels](https://cldr.unicode.org/index/cldr-spec/coverage-levels) 
+Slicing can also remove related supplemental data.
+For example, the likely subtags data includes a large number of languages that may not be of interest for all implementations.
+Where an the implementation only includes (say) the CLDR locales at Basic coverage in [Unicode CLDR - Coverage Levels](https://cldr.unicode.org/index/cldr-spec/coverage-levels)
 (and locales inheriting from them), the likely subtag data that doesn’t match can be filtered out.
 
-### 8.2 <a name="Horizontal_Slicing" href="#Horizontal_Slicing">Horizontal_Slicing</a>
+### 8.2 <a name="Horizontal_Slicing" href="#Horizontal_Slicing">Horizontal Slicing</a>
```
srl295, line 3204: assume this was a typo?

docs/ldml/tr35-personNames.md
```diff
@@ -607,7 +611,7 @@ For example:
 3. Pattern C is discarded, because it has the least number of populated name fields.
 4. Out of the remaining patterns A and B, pattern B wins, because it has only 3 unpopulated fields compared to pattern A.
 
-#### 6.5.1 <a name="process-the-namepattern" href="#process-the-namepattern">Choose a namePattern</a>
+#### 6.5.1 <a name="process-the-namepattern" href="#process-the-namepattern">Process the namePattern</a>
```
srl295, line 614: This is a good reason to drop the section numberings, it would just become:

```markdown
#### Process the namePattern
```

srl295: … which i plan to do soon

docs/ldml/tr35.md
macchiati: Yup!

docs/ldml/tr35-personNames.md
macchiati: Hmmm. We should discuss that first. The downside of that is that:
1. In the unprocessed view, there wouldn't be any links.
2. It would be very easy for someone to break external links, by inadvertently editing the title.

macchiati: I felt ok about changing the links here, because this was in tech preview. But once it 'gradulates'...

docs/ldml/tr35-personNames.md
```diff
@@ -979,7 +983,7 @@ The `nameField` values and their modifiers are described in the [Person Name Obj
 
 ## 8 <a name="personname-data-interface-examples" href="#personname-data-interface-examples">PersonName Data Interface Examples</a>
 
-### 8.1 <a name="8-1-example-1" href="#8-1-example-1">Example 1</a>
+### 8.1 <a name="example-1" href="#example-1">Example 1</a>
```
macchiati, line 986: Great. Those were a pain to fix manually.

docs/ldml/tr35-personNames.md
macchiati: Sorry, my comments were directed to "dropping the <a name=...>", not dropping the section header.

srl295: @macchiati to discuss in CLDR-15473  

1. there would be.  consider:

```markdown
- [foo](#foo)
…
# Foo
```

image

The headings are their own links. Environments can navigate them appropriately without needing HTML.

2. That can already happen, but I plan to keep a stabilized set of headings and make sure they don't get removed over releases.

macchiati: that works.

srl295:
Scrambled merge. I'll update this tomorrow with a cleaned up diff.

docs/ldml/tr35-personNames.md
srl295: the `<a name… ` in tr35-personNames.md look like they were calculated _from_ the header names, as if someone added the section numbering and `<a name…` after the fact.  So it will be easy to move it back.  They may need to be retained on other sections for stable links.

macchiati (changes requested):
This appears to be on a previous version of tr35-personNames.md. The TOC is now (sorry, loses formatting.

6.5 [Choose a namePattern](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#choose-a-namepattern%22)
6.6 [Access PersonName object](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#access-personname%22)
6.6.1 [Handle missing surname](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#missing-surname%22)
6.6.2 [Handle core and prefix](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#core-and-prefix)
6.6.3 [Derive initials](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#derive-initials)
6.7 [Process a namePattern](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#process-the-namepattern)
6.7.1 [Handling foreign names](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#handling-foreign-names)
6.7.2 [Setting the spaceReplacement](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#spaceReplacement)
6.7.3 [Examples of space replacement](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#spaceReplacement-examples)
6.8 [Formatting examples](https://github.com/unicode-org/cldr/blob/maint/maint-43/docs/ldml/tr35-personNames.md#formatting-examples)

srl295:
> This appears to be on a previous version of tr35-personNames.md. The TOC is now (sorry, loses formatting.

Crossed wires. I moved this back to draft because it's incomplete, will fix tomorrow.

macchiati:
NP

srl295:
@macchiati  PTAL (but the en_CA PR is more important) … a fresh look.  Had to fix some indent levels to match the section numbers.

docs/ldml/tr35-personNames.md
```diff
@@ -67,39 +67,41 @@ The LDML specification is divided into the following parts:
...
+    * 6.6.1 [Handle missing surname](#missing-surname)
+    * 6.6.2 [Handle core and prefix](#core-and-prefix)
+    * 6.6.3 [Derive initials](#derive-initials)
   * 6.7 [Process a namePattern](#process-the-namepattern)
-      * 6.7.1 [Handling foreign names](#handling-foreign-names)
-      * 6.7.2 [Setting the spaceReplacement](#spaceReplacement)
-      * 6.7.3 [Examples of space replacement](#spaceReplacement-examples)
+    * 6.7.1 [Handling foreign names](#handling-foreign-names)
+    * 6.7.2 [Setting the spaceReplacement](#spaceReplacement)
+    * 6.7.3 [Examples of space replacement](#examples-examples)
```
macchiati, line 81: Generally good fixes. This link is odd; probably from a mistake further down in the file, but best to fix here.

macchiati: Should be spaceReplacement-examples in 2 places

macchiati (changes requested):
Could you make the fix for example-example? as part of this?

docs/ldml/tr35-personNames.md
srl295: By the way, it's actually somewhat inverted.  Headings  such as `#### Name` are better supported, parsing the `<a name… ` chokes quite a few tools.  Markdown validation fails with "no such header: X" because the HTML isn't parsed.  `<a name` is in spec and works fine for HTML generation, but not as well supported as native markdown.

docs/ldml/tr35-personNames.md
srl295: fixed PTAL

macchiati (approved):

docs/ldml/tr35-personNames.md
macchiati: Related (but different). We should decide this week whether we're going to remove all the section numbers in the parts or not. Don't know how much time that would take.

PR #2777, opened 2023-03-13
CLDR-15850 Collect type of vote: direct, auto/manual import, bulk upload

btangmu:
-New enum VoteType in its own file VoteType.java

-New vote_type column in db table cldr_vote_value_44 (and later)

-New methods for submitting vote with vote type; default to VoteType.DIRECT

-Remove BallotBox.VOTE_IS_AUTO_IMPORTED, use VoteType.AUTO_IMPORT instead

-For VoteType.MANUAL_IMPORT (anonymous) do not trigger automatic forum posts

CLDR-15850

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -137,7 +137,7 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
      * Names of some columns in DBUtils.Table.VOTE_VALUE
      */
     private static final String VOTE_OVERRIDE = "vote_override";
-    private static final String VOTE_TYPE = "type";
+    private static final String VOTE_TYPE = "vote_type";
 
     private class DataBackedSource extends DelegateXMLSource {
         PerLocaleData ballotBox;
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -1200,7 +1200,9 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
             }
             SurveyLog.debug(et);
 
-            if (sm.fora != null && voteType == VoteType.DIRECT) {
+            // Voting can trigger adding a forum post (agree/decline) and/or closing a forum thread.
+            // AUTO_IMPORT votes are excluded; DIRECT, BULK_UPLOAD, and MANUAL_IMPORT are not excluded.
+            if (sm.fora != null && voteType != VoteType.AUTO_IMPORT) {
                 sm.fora.doForumAfterVote(locale, user, distinguishingXpath, xpathId, value, didClearFlag);
             }
         }
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -1200,7 +1200,7 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
             }
             SurveyLog.debug(et);
 
-            if (sm.fora != null && voteType == VoteType.AUTO_IMPORT) {
+            if (sm.fora != null && voteType == VoteType.DIRECT) {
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -1209,14 +1193,16 @@ private void saveVoteToDb(final User user, final String distinguishingXpath, fin
...
-                DBUtils.close(saveOld, ps, ps2, conn);
+                DBUtils.close(saveOld, ps, conn);
             }
             SurveyLog.debug(et);
 
-            if (sm.fora != null && !voteIsAutoImported) {
+            // Voting can trigger adding a forum post (agree/decline) and/or closing a forum thread.
+            // AUTO_IMPORT and MANUAL_IMPORT votes are excluded; DIRECT and BULK_UPLOAD are not excluded.
+            if (sm.fora != null && (voteType != VoteType.AUTO_IMPORT && voteType != VoteType.MANUAL_IMPORT)) {
                 sm.fora.doForumAfterVote(locale, user, distinguishingXpath, xpathId, value, didClearFlag);
```
btangmu, line 1206: The exclusion of MANUAL_IMPORT is new -- I'm not sure whether such forum posts have been made, but they would be very strange given that such votes are done in the name of an anonymous "user", and so the forum posts would also be from anonymous users (whose votes have zero weight)

btangmu: I checked the db -- there has been exactly one forum post by an anonymous user:
```
mysql> select id,poster,loc,xpath,last_time,version,type,is_open from cldr_forum_posts where poster >= 2055 and poster <= 2074 limit 10;
+-------+--------+------+--------+---------------------+---------+------+---------+
| id    | poster | loc  | xpath  | last_time           | version | type | is_open |
+-------+--------+------+--------+---------------------+---------+------+---------+
| 53706 |   2055 | ps   | 687913 | 2020-07-06 04:29:12 | 38      |    3 |       0 |
+-------+--------+------+--------+---------------------+---------+------+---------+
1 row in set (0.01 sec)
```

It's an "Agree" post (type 3).

btangmu: https://st.unicode.org/cldr-apps/v#/ps/EnergyPower/376496946a9c5f0b

srl295: I don't know how an anonymous user would post.  Almost sounds like an error?

srl295 (approved):
somehow my earlier comment got lost.

- LGTM
- would be good to actually have a migration strategy for prior tables otherwise this can't be used. It could be done in Java code or in a stored procedure that's just run every time. https://stackoverflow.com/questions/24571611/mysql-alter-table-if-column-not-exists

btangmu:
@srl295 "how an anonymous user would post": when a vetter manually imports old losing votes, the votes get imported as zero-weight votes by an anonymous user. As an accidental side-effect, if such a vote "agrees" with a vote that has a "request" post, then an "agree" post is created automatically -- my change prevents that

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -1152,13 +1160,8 @@ private void saveVoteToDb(final User user, final String distinguishingXpath, fin
...
                 if (DBUtils.db_Mysql) {
                     add0 = "IGNORE";
-                    // add1="ON DUPLICATE KEY IGNORE";
                 } else {
-                    add2 = "and not exists (select * from " + DBUtils.Table.VOTE_VALUE_ALT
-                        + " where " + DBUtils.Table.VOTE_VALUE_ALT + ".locale="
-                        + DBUtils.Table.VOTE_VALUE + ".locale and "
-                        + DBUtils.Table.VOTE_VALUE_ALT + ".xpath=" + DBUtils.Table.VOTE_VALUE + ".xpath and "
-                        + DBUtils.Table.VOTE_VALUE_ALT + ".value=" + DBUtils.Table.VOTE_VALUE + ".value )";
+                    throw new RuntimeException("Unexpected db type, expected " + DBUtils.db_Mysql);
```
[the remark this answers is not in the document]
btangmu: Good!

PR #2780, opened 2023-03-13
CLDR-16470 BRS v43: update readme for beta1

pedberg-icu:
CLDR-16470

This was supposed to update both the IU4J libs in CLDR to the latest ICU version, and update the readme in CLDR for beta1. Unfortunately it only does the latter; a recent change in ICU4J per https://github.com/unicode-org/icu/pull/2326 eliminates from core the classes UnicodeMap and ElapsedTimer that a lot of CLDR code depends upon, so we cannot use the current ICU4J libs. 

ALLOW_MANY_COMMITS=true


srl295 (approved):
LGTM

PR #2788, opened 2023-03-15
CLDR-16469 Change -2/-1 eras to inheritEras

sffc:
CLDR-16469

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #2790, opened 2023-03-15
CLDR-16469 Add inheritEras and remove eras with -2/-1

sffc:
CLDR-16469

Replaces #2788


pedberg-icu (approved):

macchiati:
Peter, Shane may not be able to merge, so you should go ahead.

PR #2792, opened 2023-03-17
CLDR-15850 Add vote_type column if not present in current table

btangmu:
-Use existing method DBUtils.tableHasColumn, and new ALTER TABLE statement

CLDR-15850

ALLOW_MANY_COMMITS=true


srl295 (approved):
looks great

PR #2794, opened 2023-03-17
CLDR-16122 v43 remove noise from test run

srl295:
CLDR-16122

very very minor. But running tests was complaining because there were two test annotations on this test.  


tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCalculatedCoverageLevels.java
```diff
@@ -40,7 +40,6 @@ public static Stream<Arguments> allLocales() {
 
     @ParameterizedTest(name = "{0}")
```
srl295, line 41: ^ this was the correct annotation

btangmu (approved):

PR #2795, opened 2023-03-17
CLDR-16389 v43 update analytics tags

srl295:
CLDR-16389

- we had at least 2 different versions of the tags, consolidated
- dropped standardFooter because the FormattedFileWriter already uses chart-template.html which now includes the footer and analytics (no longer at the foot per guidance)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/Chart.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/Chart.java
@@ -133,34 +133,27 @@ public abstract class Chart {
        throw new IllegalArgumentException("Not implemented yet");
     }
 
-    public void writeFooter(FormattedFileWriter pw) throws IOException {
-        standardFooter(pw, AnalyticsID.CLDR);
+    private static final class AnalyticsHelper {
+        private static final AnalyticsHelper INSTANCE = new AnalyticsHelper();
+
+        public final String str;
+
+        AnalyticsHelper() {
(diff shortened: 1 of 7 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/Chart.java
```diff
@@ -133,34 +133,27 @@ public void writeContents(Writer pw, Factory factory) throws IOException {
...
+
+        AnalyticsHelper() {
+            str = ToolUtilities.getUTF8Data("analytics.html").lines().collect(Collectors.joining("\n"));
+        }
     }
 
-    private enum AnalyticsID {
-        CLDR("UA-7672775-1"), ICU("UA-7670213-1"), ICU_GUIDE("UA-7670256-1"), UNICODE("UA-7670213-1"), UNICODE_UTILITY("UA-8314904-1");
+    public enum AnalyticsID {
+        CLDR("G-BPN1D3SEJM"), ICU("G-06PL1DM20S"), ICU_GUIDE("UA-7670256-1"), UNICODE("G-GC4HXC4GVQ"), UNICODE_UTILITY("G-0M7Q5QLZPV");
```
srl295, line 147: i don't know why the other tags are here, but i updated them

rickmcgowan (approved):
Only checked the GA IDs, which look OK.

macchiati (approved):

PR #2801, opened 2023-03-23
CLDR-16122 v43 spec: fix bad link, update wording

srl295:
CLDR-16122

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -3189,7 +3189,8 @@ Of course, both of these techniques can be applied.
 ### <a name="Vertical_Slicing" href="#Vertical_Slicing">Vertical Slicing</a>
 
 The choice of locales to include depends very much upon particular implementations.
-Some information which might be useful for determining that that is the [Supplemental Territory Information](tr35-info.md#Supplemental_Territory_Information),
+Some information that might be useful for determining the choice is found in the
+ [Supplemental Territory Information](tr35-info.md#Supplemental_Territory_Information),
 which provides information on the use of languages in different countries/regions.
 (For a human-readable chart, see [Territory-Language Information](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_language_information.html).)
 
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -3188,8 +3188,8 @@ Of course, both of these techniques can be applied.
 
 ### <a name="Vertical_Slicing" href="#Vertical_Slicing">Vertical Slicing</a>
 
-Locales chosen to filter out depend very much on the particular implementation.
-Some information that might be useful for determining that that is the [Supplemental Territory Information](https://unicode.org/reports/tr35/tr35-info.html#Supplemental_Territory_Information),
+The choice of locales to include depends very much upon particular implementations.
+Some information which might be useful for determining that that is the [Supplemental Territory Information](tr35-info.md#Supplemental_Territory_Information),
 which provides information on the use of languages in different countries/regions.
 (For a human-readable chart, see [Territory-Language Information](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_language_information.html).)
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1266,7 +1266,7 @@ This element must have the `transforms` element as its parent. This element repr
 
 _Attribute:_ `from` (required)
 
-> The `from` attribute consists of a sequence of elements. Each element matches one character and may consist of a codepoint or a UnicodeSet (both as defined in [UTS#35 section 5.3.3](https://www.unicode.org/reports/tr35/#Unicode_Sets)).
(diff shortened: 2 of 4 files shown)


richgillam (dismissed):
LGTM

docs/ldml/tr35.md
```diff
@@ -3188,8 +3188,8 @@ Of course, both of these techniques can be applied.
 
 ### <a name="Vertical_Slicing" href="#Vertical_Slicing">Vertical Slicing</a>
 
-Locales chosen to filter out depend very much on the particular implementation.
-Some information that might be useful for determining that that is the [Supplemental Territory Information](https://unicode.org/reports/tr35/tr35-info.html#Supplemental_Territory_Information),
+The choice of locales to include depends very much upon particular implementations.
+Some information which might be useful for determining that that is the [Supplemental Territory Information](tr35-info.md#Supplemental_Territory_Information),
```
macchiati, line 3192: Should be
Some information which might be useful for determining that that is the
=>
Some information that might be useful for determining the choice is found in the

macchiati (changes requested):
Thanks! 

However, these are not the only problems I pointed to.

We shouldn't ever have:
https://unicode.org/reports/tr35/tr35-.*/.(html|md)?#
or 
https://unicode.org/reports/tr35/#
or variants of these.

That is, we should always have **relative** links within the parts of the spec, except for the (few) instances that must point to specific versions.
----

tr35.md (9 matches)
3,192: Some information that might be useful for determining that that is the [Supplemental Territory Information](https://unicode.org/reports/tr35/tr35-info.html#Supplemental_Territory_Information), 
tr35-general.md (2 matches)
1,269: > The `from` attribute consists of a sequence of elements. Each element matches one character and may consist of a codepoint or a UnicodeSet (both as defined in [UTS#35 section 5.3.3](https://www.unicode.org/reports/tr35/#Unicode_Sets)). 
tr35-personNames.md (4 matches)
104: This document provides the [LDML](http://www.unicode.org/reports/tr35/) specification for formatting of personal names, using data, structure, and examples. 
262: * `#PCDATA `is a space delimited list of one or more [unicode_locale_id](https://unicode.org/reports/tr35/#unicode_locale_id)s. Normally each locale is limited to language, script, and region. The _und_ locale ID may only occur once, either in _surnameFirst_ or _givenFirst_, but not both, and matches all base locales not explicitly listed. 
543: These nameOrderLocales will match any locale with a zh or ja [unicode_language_subtag](https://unicode.org/reports/tr35/#unicode_language_subtag) and any locale with a CN, TW, SG, HK MO, HU, or JP [unicode_region_subtag](https://unicode.org/reports/tr35/#unicode_region_subtag).

docs/ldml/tr35.md
srl295: > Thanks!
> 
> However, these are not the only problems I pointed to.
> 
> We shouldn't ever have: [https://unicode.org/reports/tr35/tr35-.*/.(html|md)?#](https://unicode.org/reports/tr35/tr35-.*/.(html%7Cmd)?#) or https://unicode.org/reports/tr35/# or variants of these.
> 
> ## That is, we should always have **relative** links within the parts of the spec, except for the (few) instances that must point to specific versions.
> tr35.md (9 matches) 3,192: Some information that might be useful for determining that that is the [Supplemental Territory Information](https://unicode.org/reports/tr35/tr35-info.html#Supplemental_Territory_Information), tr35-general.md (2 matches) 1,269: > The `from` attribute consists of a sequence of elements. Each element matches one character and may consist of a codepoint or a UnicodeSet (both as defined in [UTS#35 section 5.3.3](https://www.unicode.org/reports/tr35/#Unicode_Sets)). tr35-personNames.md (4 matches) 104: This document provides the [LDML](http://www.unicode.org/reports/tr35/) specification for formatting of personal names, using data, structure, and examples. 262: * `#PCDATA `is a space delimited list of one or more [unicode_locale_id](https://unicode.org/reports/tr35/#unicode_locale_id)s. Normally each locale is limited to language, script, and region. The _und_ locale ID may only occur once, either in _surnameFirst_ or _givenFirst_, but not both, and matches all base locales not explicitly listed. 543: These nameOrderLocales will match any locale with a zh or ja [unicode_language_subtag](https://unicode.org/reports/tr35/#unicode_language_subtag) and any locale with a CN, TW, SG, HK MO, HU, or JP [unicode_region_subtag](https://unicode.org/reports/tr35/#unicode_region_subtag).

will work on the match.  I thought i hit all of these except for ones which pointed to specific versions.

srl295: OK. i definitely was not looking for `https://unicode.org/reports/tr35/#`  - that's an odd form.

srl295: fixed

srl295:
I added another regex to catch the unversioned `reports/tr35/#` case.

macchiati:
Does it also catch .../tr35# ?

srl295:
> Does it also catch .../tr35# ?

yes (as noted above)

macchiati (approved):

PR #2803, opened 2023-03-24
CLDR-16123 add spotless

srl295:
CLDR-16123

- update docs
- to re-run yourself: `mvn -f tools/pom.xml spotless:apply`
- The First commit is docs and tooling, the next commit is the actual code changes

ALLOW_MANY_COMMITS=true


macchiati (changes requested):
Just to confirm: we won't be doing this for real until maint-43 is merged back in, right?

srl295:
> Just to confirm: we won't be doing this for real until maint-43 is merged back in, right?

Correct.

srl295:
> Just to confirm: we won't be doing this for real until maint-43 is merged back in, right?

Any specific concern or just wanting to verify the timing ?

macchiati:
Just verifying.

macchiati:
Results look pretty reasonable, eg
https://github.com/unicode-org/cldr/pull/2803/files?diff=unified&w=1#diff-fc6d4246f61ecdae85f649ac17c04630d85fcac0b439a0259ef3dfd35359d792

And of course, passing all the tests gives us a lot of assurance.

srl295:
for vscode, i added these lines to `.vscode/settings.json` on my local copy, but I am still testing this.

```json
{
    "java.format.settings.url": "https://raw.githubusercontent.com/aosp-mirror/platform_development/master/ide/eclipse/android-formatting.xml",
    "java.format.enabled": true
}
```

srl295:
@echeran

srl295:
@echeran  seems like this is sensitive not just to the Style version but to the maven plugin version, [here](https://github.com/unicode-org/cldr/compare/unicode-org:79848be..unicode-org:7b0c61a) is what happened when trying to go from 2.35.0 to 2.36.0 (current PR is 2.35.0 for parity with UnicodeTools)

echeran:
@srl295 FYI: Even though the [Unicode Tools instructions](https://unicode-org.github.io/unicodetools/build) talk about getting the formatter to work locally in IDEs, in practice, I couldn't get any of them to work. If you can get the formatter to work in VS Code or IntelliJ, then I want to use that too, and we probably want to update the Unicode Tools instructions, as well.

srl295:
I'm merging with the spotless check *nonblocking*.  So it won't block PRs. 

I can do smaller fixups to main, and i'll work on the IDE settings.  We may just need our own config file, I think the ones linked diverged from what spotless actually does somehow.

srl295:
I meant to merge this as two commits!

srl295:
@echeran a thought is it might be simplest for now to just invoke mvn spotless:apply from the IDE.

PR #2811, opened 2023-03-27
CLDR-16437 use mocha-headless-chrome

srl295:
CLDR-16437

- also see CLDR-14408 CLDR-14409

ALLOW_MANY_COMMITS=true


srl295:
@btangmu  go ahead and merge too if you want and i'm not around, this will unbreak JS tests

btangmu (approved):

PR #2814, opened 2023-03-28
CLDR-16511 flesh out modification section

macchiati:
CLDR-16511

ALLOW_MANY_COMMITS=true


macchiati:
There is more to come, but we could go ahead and merge this much.

srl295 (approved):

pedberg-icu (approved):

PR #2821, opened 2023-03-29
CLDR-16048 New forum post window sometimes disappears

btangmu:
-Use LRU cache in cldrForumPanel.loadInfo and cldrSideways.loadMenu to avoid disappearance/reload

-New cldrCache.mjs for simple front-end LRU cache, with unit test TestCldrCache.mjs

-New Refresh button in Info Panel, clears caches and reloads

-Clear forum cache when user submits a post

-Remove cldrForumPanel.appendForumStuff, merge into loadInfo instead

-Encapsulate FORUM_DIV_CLASS (forumDiv) and SIDEWAYS_AREA_CLASS (sidewaysArea)

-Refactor/modernize some related code; reduce cruft; new subroutines to shorten long methods

-Remove dead code (buttonTitle, theForum, replyStub, postUrl, tr.forumDiv.forumPosts, ...)

-Comments

CLDR-16048

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrCache.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrCache.mjs
@@ -1,0 +1,42 @@
+/*
+ * cldrCache: a simple least-recently-used cache for the Survey Tool front end
+ *
+ * Based partly on: https://stackoverflow.com/questions/996505/lru-cache-implementation-in-javascript
+ * post by odinho - Velmont
+ */
+export class LRU {
+  constructor(max = 100) {
+    this._max = max;
+    this._cache = new Map();
+  }
+
+  clear() {
+    this._cache.clear();
+  }
+
+  get(key) {
+    const val = this._cache.get(key);
+    if (val) {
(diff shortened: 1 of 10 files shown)


btangmu:
> The info panel should become a vue component… that'd be relatively low hanging fruit. could still be done incrementally

actually it already is, though only the title bar! beyond that we can start making other bits of it into vue components -- since a legacy DOM element can contain a vue component, but not (I think) vice-versa

srl295:
> > The info panel should become a vue component… that'd be relatively low hanging fruit. could still be done incrementally
> 
> actually it already is, though only the title bar!

That's right!  Maybe the title bar could grow in scope and vertical size!

> …beyond that we can start making other bits of it into vue components -- since a legacy DOM element can contain a vue component, but not (I think) vice-versa

Actually I think it can.  Looking at https://vuejs.org/guide/essentials/template-refs.html#template-refs it shows how to get a DOM reference in Vue.  From there I think a component could call the usual cldr functions to manipulate the dom.

PR #2772, opened 2023-03-10
CLDR-15473 drop section numbering

srl295:
CLDR-15473

Note: more radical changes to personNames as it is newer, simplified its header structure.

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #2822, opened 2023-03-29
CLDR-16513 v43 BRS: keep anchors stable

srl295:
CLDR-16513

- See ticket for some discussion
- see https://www.unicode.org/~srloomis/tmp/tr3520894/tr35.html for a generated example (this affects the HTML output)
- revert of https://github.com/unicode-org/cldr/commit/f8141b9563470ab3581b708a270b6f8574ece03d  which was a component of https://github.com/unicode-org/cldr/pull/2772 landed in https://github.com/unicode-org/cldr/commit/43fa8529c2c6935ccae331231e15d27461e327c6 - this was the commit which broke the stable links
- add some new files such as `docs/ldml/tr35.anchors.json` which track anchors to make sure they don't get dropped accidentally.
- Add a tool that generates the above anchors files.
- update the archiver so that when we have the situation `<h4 id="generated-name"><a name="Explicit_Name" …` (where markdown generates an explicit anchor) we swap the `name` with the replacement `id`. This causes the auto linker to use the Explicit_Name in the little link icon where present.

## Please review https://github.com/unicode-org/cldr/pull/2822/commits/bc76d61b80d10be003b6aaaab991203e92335bd3  - this shows which anchors are being dropped (or added) in v43. This will become a BRS task.

ALLOW_MANY_COMMITS=true


macchiati (approved):
I can't tell whether all of the anchors are fixed, but I presume that your test does that.

srl295:
> I can't tell whether all of the anchors are fixed, but I presume that your test does that.

Commit https://github.com/unicode-org/cldr/commit/bc76d61b80d10be003b6aaaab991203e92335bd3 shows that the only valuable  anchors removed were autogenerated ones with section numbers.

PR #2823, opened 2023-03-30
CLDR-16522 deps add .node-version file @ v18.15.0

srl295:
- add a .node-version file, defacto format defined at https://github.com/shadowspawn/node-version-usage
- mvn/ant code to use the .node-version file within frontend-maven-plugin, workaround for https://github.com/eirslett/frontend-maven-plugin/issues/1047
- use the nodenv/actions/node-version@v3 action which reads .node-version in relevant places
- turn on webpack's persistent cache, and cache it across builds
- also, move some maven plugin dependencies up to tools/pom.xml I'm keeping package dependencies (such as mysql) in cldr-apps/pom.xml

CLDR-16522

ALLOW_MANY_COMMITS=true


srl295:
I use this with nvm and `alias nvmi='if [ -f .node-version ]; then nvm install $(cat .node-version); else nvm use default; fi'`

btangmu (approved):

PR #2828, opened 2023-03-31
CLDR-16316 test: make it a warning if a currency code recently expired

srl295:
- give 1 year for the ISO 4217 to be still in CLDR if it recently expired
- for example, the notice of extension for SLL only came the day after expiration
- also improve the output to include the actual expiration date

CLDR-16316

ALLOW_MANY_COMMITS=true

-----

example output:

> `Warning: (TestSupplementalInfo.java:1445) WARNING: Codes in ISO 4217 and until-recently legal tender in CLDR. (may need to update https://cldr.unicode.org/development/updating-codes/update-currency-codes ): SLL end=Thu Mar 30 17:00:00 PDT 2023`

Previous error:

> `(TestSupplementalInfo.java:1428)  Error: Codes in ISO 4217 but not current tender in CLDR (may need to update https://cldr.unicode.org/development/updating-codes/update-currency-codes ): [SLL]`

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateConstants.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateConstants.java
@@ -1,0 +1,23 @@
+package org.unicode.cldr.util;
+
+import java.util.Date;
+
+/**
+ * Some date related constants
+ */
+public class DateConstants {
+    /**
+     * Right now. A single constant so that it is consistent across callers.
+     */
+    public static final Date NOW = new Date();
+    public static final long MILLIS_PER_SECOND  = 60 * 1000;
+    public static final long MILLIS_PER_MINUTE  = MILLIS_PER_SECOND * 60;
+    public static final long MILLIS_PER_HOUR    = MILLIS_PER_MINUTE * 60;
+    public static final long MILLIS_PER_DAY     = MILLIS_PER_HOUR * 24;
+    public static final long MILLIS_PER_MONTH   = MILLIS_PER_DAY * 30;
+
+    /**
+     * A date a little less than 8 months ago.
+     */
+    public static final Date RECENT_HISTORY = new Date(NOW.getTime() - (MILLIS_PER_MONTH * 8));
+}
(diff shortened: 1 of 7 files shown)


DavidLRowe:
Sorry, I'm not qualified to review java code.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -1424,10 +1439,22 @@ public void TestCurrency() {
...
+                .collect(Collectors.joining(", ")));
+            missing.removeAll(recentMissing); // not errors
+        }
         if (missing.size() != 0) {
             errln("Codes in ISO 4217 but not current tender in CLDR " +
                 "(may need to update https://cldr.unicode.org/development/updating-codes/update-currency-codes ): " +
-                missing.toString());
+                nonModernCurrencyCodes.entrySet().stream().filter(p -> missing.contains(p.getKey()))
+                .map(p -> (p.getKey() + " ended=" + p.getValue().getSecond().getEnd()))
+                .collect(Collectors.joining(", ")));
```
btangmu, line 1457: the url and the filter code ideally might be encapsulated rather than repeated

btangmu (dismissed):

srl295:
> We tend to ignore warnings, so I think it should cause an error (but after a decent interval, like 7 months).

It still causes an error after 1 year. I can shorten it to 7 months.

macchiati:
I'd make it specifically less than 1 year, so we make sure to address it after at most 1 release. 7 or 8 months.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -1347,7 +1346,7 @@ public void TestCompleteness() {
      *
      * @param args
      */
-    public void TestCurrency() {
+    public void TestSupplementalCurrency() {
```
srl295, line 1349: renamed this to avoid ambiguity

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateConstants.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Date;
+
+/**
+ * Some date related constants
+ */
+public class DateConstants {
+    /**
+     * Right now. A single constant so that it is consistent across callers.
+     */
+    public static final Date NOW = new Date();
```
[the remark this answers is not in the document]
srl295: it's more for tests

macchiati (approved):

PR #2834, opened 2023-04-04
CLDR-16532 Ignore missing hidden notifications table if db is too old

btangmu:
-Avoid throwing exception with db before version 42, for testing other features than hidden notifications

-Earlier versions had versioned tables like cldr_dash_hide_41 or cldr_review_hide_38

CLDR-16532

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReviewHide.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReviewHide.java
@@ -103,8 +103,11 @@ public class ReviewHide {
                 this.hiddenNotifications.put(subtype, xpstrid, val);
             }
         } catch (SQLException sqe) {
-            SurveyLog.logException(sqe, "Getting hidden notifications for uid#" + userId + " in " + localeId, null);
-            throw new InternalError("Error getting hidden notifications: " + sqe.getMessage());
+            // first version with cldr_dash_hide table is 42; silently ignore exceptions for earlier versions
+            if (Integer.parseInt(SurveyMain.getNewVersion()) >= 42) {
+                SurveyLog.logException(sqe, "Getting hidden notifications for uid#" + userId + " in " + localeId, null);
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReviewHide.java
```diff
@@ -103,8 +103,11 @@ private void getData() {
                 this.hiddenNotifications.put(subtype, xpstrid, val);
             }
         } catch (SQLException sqe) {
-            SurveyLog.logException(sqe, "Getting hidden notifications for uid#" + userId + " in " + localeId, null);
-            throw new InternalError("Error getting hidden notifications: " + sqe.getMessage());
+            // first version with cldr_dash_hide table is 42; silently ignore exceptions for earlier versions
+            if (Integer.parseInt(SurveyMain.getNewVersion()) >= 42) {
```
srl295, line 107: I'd suggest something like `ReviewHide.FIRST_DB_VERSION`. 
In fact, even better might be:

`DBUtils.throwIfAtLeast(sqe, ReviewHide.FIRST_DB_VERSION, () -> "Getting hidden notifications for uid#" + userId + " in " + localeId)`

where that function checks `SurveyMain.getNewVersion()` (or via CLDRConfig) and throws an `InternalError` if it's a new version, otherwise does a `SurveyLog.warnOnce()`

the third arg to `throwIfAtLeast()` is a `Producer<String>` so only called when needed

I say this because I think we have and will have more of these.

But just externalizign the version number or doing nothing is fine too at your discretion

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ReviewHide.java
btangmu: I'll do a follow-up PR per your suggestions

PR #2835, opened 2023-04-04
CLDR-16532 Follow-up: define constant FIRST_DB_VERSION for magic number 42

btangmu:
CLDR-16532

ALLOW_MANY_COMMITS=true


btangmu:
@srl your suggestion in https://github.com/unicode-org/cldr/pull/2834 for a new method DBUtils.throwIfAtLeast seems promising, however maybe beyond the scope (and time constraints) of this ticket -- if and when more cases arise we can search for calls to getNewVersion() where such encapsulation would be helpful

srl295:
like i said i only mention it because i think we may encounter more of these cases, if adding such a function saves time because it unencumbers making DB changes, then it’s worth it.  Otherwise it can be an idea for later.

srl295 (approved):

PR #2836, opened 2023-04-04
CLDR-16131 fix glych in empty value

macchiati:
CLDR-16131

Change the way that an empty value is represented: have it be the collection of Status=value attributes. That preserves a fallback for strange cases, but also lets an empty value be empty.

ALLOW_MANY_COMMITS=true


macchiati:
See https://github.com/unicode-org/cldr-staging/pull/88/files for differences in person name chart

srl295 (approved):

pedberg-icu (approved):

PR #2838, opened 2023-04-04
CLDR-16449 Retain sr_Latn in coverage chart 

macchiati:
CLDR-16449

And in the coverageLevels.txt file

ALLOW_MANY_COMMITS=true


macchiati:
See https://github.com/unicode-org/cldr-staging/pull/88/files for differences in coverage chart

srl295 (approved):

pedberg-icu (approved):

PR #2841, opened 2023-04-05
CLDR-16543 Fix exemplar cities

robertbastian:
CLDR-16543

I didn't move `Kostanay` and `Uzhhorod` to `root`, as these seem to be English names (the strict romanization is different).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -3462,15 +3462,6 @@ annotations.
 			<zone type="Etc/Unknown">
 				<exemplarCity>Unknown City</exemplarCity>
 			</zone>
-			<zone type="Antarctica/DumontDUrville">
-				<exemplarCity>Dumont d’Urville</exemplarCity>
-			</zone>
-			<zone type="America/St_Barthelemy">
-				<exemplarCity>St. Barthélemy</exemplarCity>
-			</zone>
-			<zone type="America/Curacao">
-				<exemplarCity>Curaçao</exemplarCity>
-			</zone>
 			<zone type="Europe/London">
 				<long>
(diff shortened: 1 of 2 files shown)


common/main/en.xml
```diff
@@ -3462,15 +3462,6 @@ annotations.
 			<zone type="Etc/Unknown">
 				<exemplarCity>Unknown City</exemplarCity>
 			</zone>
-			<zone type="Antarctica/DumontDUrville">
-				<exemplarCity>Dumont d’Urville</exemplarCity>
```
srl295, line 3466: Why?

srl295:
I would not change `en.xml` in this PR.

macchiati:
We don't treat en.xml in the same way as the other locales; if the values
are the same as inherited, we don't repeat them.
In other locales, we require confirmation that the values are the same. We
could, however, "flesh out" the en.html values in the same way as we do
other locales; you could file a ticket for that.

robertbastian:
That makes sense.

> We could, however, "flesh out" the en.html values in the same way as we do other locales; you could file a ticket for that.

This would need to be done for https://unicode-org.atlassian.net/browse/CLDR-16546

common/main/en.xml
macchiati: en.xml doesn't need to be 'complete', except for sideways inheritance. We require other locales to be complete so that we are sure that they have considered each data item.

robertbastian:
@pedberg-icu who is assigned as reviewer on the CLDR issue

macchiati (approved):

PR #2845, opened 2023-04-06
CLDR-16550 BRS43 beta5 CLDRModify passes: -fQ reorders annotations w/ quotes; -fP reorders geresh/gershayim in punct exemplars

pedberg-icu:
CLDR-16550

CLDRModify passes before beta5:
- `-fQ` reorders annotations with quotes (in common/annotations);
- `-fP` reorders geresh/gershayim in punctuation exemplars (in common/main)

This all seems as expected due to the quotation mark collation order changes (note that no annotations were removed, and no spaces in punctation exemplars were removed, as in the discarded previous CLDRModify passes which had various quotes as equal).

ALLOW_MANY_COMMITS=true


srl295 (approved):

macchiati (approved):
Looks good. The split between ׳ and ״ is a bit surprising, so I'll report that back to Markus, but nothing that should prevent the release.

macchiati:
Never mind my comments. The bidi ordering on the screen mixed me up.

markusicu (approved):
rslgtm

PR #2846, opened 2023-04-07
CLDR-15850 Replace single Vote Type column with five specific columns

btangmu:
-New const COLUMN_TITLE_VOTES_DIRECT, ...; new const VOTE_TYPES; new method getVoteTypes

-Revise COLUMNS

-Revise the wording of some column comments

-Include column title with comment, for convenience if column is narrower than title

-Warn in console if more than 26 columns A-Z

-Change some console.log to console.warn

CLDR-15850

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
+++ b/tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
@@ -23,12 +23,16 @@ const COLUMN_TITLE_COVERAGE_COUNT = "Cldr Coverage Count";
 const COLUMN_TITLE_VOTED_PATH_COUNT = "Progress Vote";
 const COLUMN_TITLE_VOTABLE_PATH_COUNT = "Progress Count";
 const COLUMN_TITLE_PROGRESS_PERCENT = "Progress Percent";
+const COLUMN_TITLE_VOTES_DIRECT = "Direct Votes";
+const COLUMN_TITLE_VOTES_AUTO_IMPORT = "Auto-imported Votes";
+const COLUMN_TITLE_VOTES_MANUAL_IMPORT = "Manually-imported Votes";
+const COLUMN_TITLE_VOTES_BULK_UPLOAD = "Bulk-uploaded Votes";
+const COLUMN_TITLE_VOTES_UNKNOWN = "Unknown-type Votes";
 const COLUMN_TITLE_USER_ID = "Vetter#";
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
```diff
@@ -448,16 +476,37 @@ function getDefaultRow(id, user, columnIndex) {
 function addColumnComments(ws) {
   let unicode = "A".charCodeAt(0);
   for (let col of COLUMNS) {
-    const letter = String.fromCharCode(unicode++); // A, B, C, ... assume max 26 columns!
+    const letter = String.fromCharCode(unicode++); // A, B, C, ...
     const cell = letter + "1"; // A1, B1, C1, ...
```
srl295, line 480: use the utility for this

```js
const cell = XLSX.utils.encode_cell({r: 1, c: column_number});
```

instead of `unicode` 

see https://docs.sheetjs.com/docs/csf/general#cell-addresses-1

tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
```diff
@@ -448,16 +476,37 @@ function getDefaultRow(id, user, columnIndex) {
...
-    cldrXlsx.pushComment(ws, cell, col.comment);
+    // include title with comment, for convenience if column is narrower than title
+    cldrXlsx.pushComment(ws, cell, col.title + ": " + col.comment);
+    if (letter === "Z") {
+      if (COLUMNS.length > 26) {
+        console.warn("Comments are not supported for more than 26 columns A-Z");
+      }
+      return;
+    }
```
srl295, line 488: see above

srl295 (approved):
lgtm but comments

tools/cldr-apps/js/src/esm/cldrVettingParticipation.js
btangmu: it's always fun to reinvent the wheel!

I'll change it to encode_cell in the next PR

btangmu: Actually "r: 0" not "r: 1"

The documentation is misleading

srl295: sorry, yes r:0 my mistake

PR #2847, opened 2023-04-10
CLDR-15850 Vetting table xlsx use encode_cell

btangmu:
-Support more than 26 columns using pre-existing library method

CLDR-15850

ALLOW_MANY_COMMITS=true


srl295 (approved):

srl295:
we have lots of spreadsheet features and features in the wings, so i think good patterns are worth it here.

PR #2851, opened 2023-04-11
CLDR-16538 B Revise the likely subtags chart

macchiati:
CLDR-16538

Notes for reviewers

- I did a bit of refactoring so that I could do more rapid turnaround. That involved inlining private static StringWriter printLanguageData(...), so that disappeared (lines 104...), plus a flag to exit early.
- The only substantive changes are at line 936 and following. I reordered the columns, reset the sorting priority, and dropped the setSpanRows(false).
- At line 958, I disabled the normal handling of 'und' for the target language, so that it would be in an appropriate place.
- At line 962, I reordered the data in the cells to match the column headers.

I uploaded to the the chart to cldr-staging so that the results can be checked. It should show up soon (check that the date says 2023-04-11) at https://cldr-smoke.unicode.org/staging-dev/charts/43/supplemental/likely_subtags.html .

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
@@ -85,6 +85,7 @@ import com.ibm.icu.util.ULocale;
 @CLDRTool(alias = "showlanguages", description = "Generate Language info charts")
 public class ShowLanguages {
     private static final boolean SHOW_NATIVE = true;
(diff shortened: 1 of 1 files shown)


macchiati:
The plan is to merge this after release (unless something else in the release needs to change).

macchiati:
Why was this assigned to me? It's been merged.

srl295:
> Why was this assigned to me? It's been merged.

Just updating metadata

PR #2853, opened 2023-04-12
CLDR-16464 v44: bump version numbers to v44

srl295:
CLDR-16464

ALLOW_MANY_COMMITS=true


srl295:
it's hard to tell what this error is actually saying. @macchiati or @pedberg-icu ?

`Error: (TestValidity.java:219) subdivision:cqzzzz:unknown => null — missing in new data vs. cldr-43.0`

I added the 43.0 part. I think it's complaining that v44 doesn't have `cqzzzz` as valid and v43 did, due to the merge from branch not happening yet. So perhaps that's a downside of bumping the version number too quickly?  Maybe we should merge from the branch at the betas?

srl295:
@btangmu i didn't see your review comments before i updated and now i can't see them… what was the comment?

btangmu:
I don't think I made a review comment -- just approved

PR #2854, opened 2023-04-12
CLDR-16560 Requirements for storing inheritance marker

btangmu:
-New CLDRFile.reviseInheritanceAsNeeded and STFactory.DataBackedSource.reviseInheritanceAsNeeded

-Revise CLDRFile.add and STFactory.DataBackedSource.setValueFromResolver

-Throw exception if CLDRFile.reviseInheritanceAsNeeded called with isResolved false

CLDR-16560

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -230,9 +230,20 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
             }
             delegate.removeValueAtDPath(path);
             if (value != null) {
+                value = reviseInheritanceAsNeeded(path, value);
                 delegate.putValueAtPath(fullPath, value);
             }
             return resolver;
+        }
+
+        public String reviseInheritanceAsNeeded(String path, String value) {
+            if (value != null) {
+                CLDRFile cldrFile = getDiskFile(ballotBox.locale);
+                if (cldrFile != null) {
+                    value = cldrFile.reviseInheritanceAsNeeded(path, value);
+                }
+             }
+            return value;
         }
 
         private String getFullPathWithResolver(String path, VoteResolver<String> resolver) {
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -238,7 +238,7 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
 
         public String reviseInheritanceAsNeeded(String path, String value) {
             if (value != null) {
-                CLDRFile cldrFile = getDiskFile(ballotBox.locale);
+                CLDRFile cldrFile = make(ballotBox.locale.getBaseName());
                 if (cldrFile != null) {
                     value = cldrFile.reviseInheritanceAsNeeded(path, value);
                 }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -431,26 +431,13 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
      */
     @Override
     public String getStringValue(String xpath) {
-        boolean debug = "//ldml/localeDisplayNames/languages/language[@type=\"ko\"]".equals(xpath);
-        if (debug) {
-            System.out.println("getStringValue 0 debug = " + debug);
-        }
         try {
             String result = dataSource.getValueAtPath(xpath);
-            if (debug) {
-                System.out.println("getStringValue 1 result = " + result); // null
-            }
             if (result == null && dataSource.isResolving()) {
                 final String fallbackPath = getFallbackPath(xpath, false, true);
                 // often fallbackPath equals xpath -- in such cases, isn't it a waste of time to call getValueAtPath again?
-                if (debug) {
-                    System.out.println("getStringValue 2 fallbackPath = " + fallbackPath);
-                }
                 if (fallbackPath != null) {
                     result = dataSource.getValueAtPath(fallbackPath);
-                    if (debug) {
-                        System.out.println("getStringValue 3 result = " + result);
-                    }
                 }
             }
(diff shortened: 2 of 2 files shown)


btangmu:
Testing locally on my macBook, I get 4 errors in TestSTFactory (and I expect to get the same errors for this PR)

```
getStringValue 0 debug = true
getStringValue 1 result = null
getStringValue 4 returning null
reviseInheritanceAsNeeded 0 debug = true path = //ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraNarrow/era[@type="0"] locale = mul_AQ
reviseInheritanceAsNeeded pathWhereFound = //ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraAbbr/era[@type="0"] localeWhereFound = root
reviseInheritanceAsNeeded CHANGED got ↑↑↑ returned BCE
reviseInheritanceAsNeeded 0 debug = true path = //ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraNarrow/era[@type="0"] locale = mul_AQ
reviseInheritanceAsNeeded pathWhereFound = //ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraAbbr/era[@type="0"] localeWhereFound = root
reviseInheritanceAsNeeded CHANGED got ↑↑↑ returned BCE
reviseInheritanceAsNeeded 0 debug = true path = //ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraNarrow/era[@type="0"] locale = mul_AQ
reviseInheritanceAsNeeded pathWhereFound = //ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraAbbr/era[@type="0"] localeWhereFound = root
reviseInheritanceAsNeeded CHANGED got ↑↑↑ returned BCE
 {
      Error: (TestSTFactory.java:497) 58b Expected AAA at mul_AQ://ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraNarrow/era[@type="0"] got BCE
```

why is mul_AQ inheriting from root instead of mul?

We really need the inheritance explainer: https://unicode-org.atlassian.net/browse/CLDR-16390

macchiati:
I think 'mul' might somehow get special treatment. Run this by @srl295

btangmu:
As expected:

Error summary:
CLDR.Web/TestSTFactory/TestVettingDataDriven

<< 4 TEST(S) FAILED >>

srl295:
might be a problem with the test factory itself

btangmu:
> might be a problem with the test factory itself

maybe the CLDRFile returned by STFactory.getDiskFile(ballotBox.locale) doesn't get the same votes as the one produced by TestSTFactory.getFactory?

btangmu:
as I wrote in email, I'd like to  accept https://unicode-org.atlassian.net/browse/CLDR-16390 as “blocks progress”, and make it a subtask of https://unicode-org.atlassian.net/browse/CLDR-16537 
blocking https://unicode-org.atlassian.net/browse/CLDR-16560

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -632,6 +647,51 @@ public CLDRFile add(String currentFullXPath, String value) {
...
+     * 1. If the path changes when getting bailey, then we are inheriting sideways. We need to use a hard value.
+     * 2. If the value is different from the bailey value, can't use inheritance; we need a hard value.
+     * 3. Otherwise we use inheritance marker.
+     *
+     * @param path the path
+     * @param value the input value
+     * @return the possibly modified value
+     */
+    public String reviseInheritanceAsNeeded(String path, String value) {
+        if (isResolved()) {
```
macchiati, line 663: If someone calls this on a non-resolved cldrfile, it should throw; otherwise people will think it is producing a valid answer when it isn't.

btangmu: OK, in that case the caller CLDRFile.add needs to check isResolved before calling reviseInheritanceAsNeeded -- since CLDRFile.add does get sometimes get called with isResolved false

macchiati:
Ah, I hadn't noticed that you were planning to do this in CLDRFile add(..), rather than at a higher level. I'm a leery of that approach since I think we might need to have more control at a low level (plus there will be extra processing cost).

I think it would be cleaner to use a new method (eg addWithInheritance(..)) and then just call that where the SurveyTool is adding elements.

btangmu:
> Ah, I hadn't noticed that you were planning to do this in CLDRFile add(..), rather than at a higher level. I'm a leery of that approach since I think we might need to have more control at a low level (plus there will be extra processing cost).
> 
> I think it would be cleaner to use a new method (eg addWithInheritance(..)) and then just call that where the SurveyTool is adding elements.

I don't know who would call addWithInheritance() in that case. Maybe such a method isn't needed, and we could just omit the call to reviseInheritanceAsNeeded() from add(). However, add() is called a lot, though not when a user votes the ordinary way in Survey Tool. Intuitively, it seems safest always to enforce the rule consistently. I don't understand what you mean by "more control at a low level" but I would defer to your judgment if you think it's best for add() not to enforce the rule. Maybe DataBackedSource.setValueFromResolver is the only place it's necessary to enforce the rule.

btangmu:
@macchiati I've just tried reverting CLDRFile.add so it doesn't call reviseInheritanceAsNeeded, and the tests do pass. So, the tests pass regardless of whether CLDRFile.add calls reviseInheritanceAsNeeded..

Should I revise the PR so that CLDRFile.add doesn't call reviseInheritanceAsNeeded?

macchiati (commented):
Looks good. Are you going to add some tests?

btangmu:
See https://github.com/unicode-org/cldr/pull/2855

btangmu:
> Looks good. Are you going to add some tests?

That's a good idea!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -230,11 +230,22 @@ private VoteResolver<String> setValueFromResolver(String path, VoteResolver<Stri
...
             if (value != null) {
+                value = reviseInheritanceAsNeeded(path, value);
                 delegate.putValueAtPath(fullPath, value);
             }
             return resolver;
         }
 
+        public String reviseInheritanceAsNeeded(String path, String value) {
+            if (value != null) {
+                CLDRFile cldrFile = make(ballotBox.locale.getBaseName());
```
srl295, line 241: I know this is closed ,but yes, this will return a resolving CLDRFile for the specific locale (the result is cached).

PR #2855, opened 2023-04-13
CLDR-16560 Requirements for storing inheritance marker

btangmu:

-New method STFactory.DataBackedSource.reviseInheritanceAsNeeded called by setValueFromResolver

-Throw exception from reviseInheritanceAsNeeded if CLDRFile has isResolved false

-Only change STFactory; no changes to CLDRFile

CLDR-16560

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -18,6 +18,7 @@ import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
+import com.ibm.icu.util.Output;
 import org.unicode.cldr.test.CheckCLDR;
 import org.unicode.cldr.test.DisplayAndInputProcessor;
 import org.unicode.cldr.test.TestCache;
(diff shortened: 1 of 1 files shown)


srl295:
OK i was just trying to review this, but i have questions about the goals here, updated ticket filed at https://unicode-org.atlassian.net/browse/CLDR-16560

I'll take a look now.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -230,11 +231,41 @@ private VoteResolver<String> setValueFromResolver(String path, VoteResolver<Stri
             }
             delegate.removeValueAtDPath(path);
             if (value != null) {
+                value = reviseInheritanceAsNeeded(path, value);
```
srl295, line 234: This is actually an odd place to update.  because this is the bottleneck for basically READING from the DB, not for STORING.  

I'm not sure this does what you want.  This won't affect the database. 

But again, I'm not sure of the goals.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -230,11 +231,41 @@ private VoteResolver<String> setValueFromResolver(String path, VoteResolver<Stri
...
+                    Output<String> pathWhereFound = new Output<>();
+                    Output<String> localeWhereFound = new Output<>();
+                    String baileyValue = cldrFile.getBaileyValue(path, pathWhereFound, localeWhereFound);
+                    if (baileyValue != null && (CldrUtility.INHERITANCE_MARKER.equals(value) || baileyValue.equals(value))) {
+                        return pathWhereFound.value.equals(path) ? CldrUtility.INHERITANCE_MARKER : baileyValue;
+                    }
+                }
+             }
+            return value;
+        }
```
srl295, line 267: I don't think this belongs here.  It should go as a static function on CLDRFile so it can be unit tested separately.

macchiati: good point

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
srl295: Given the update on the ticket. this is not in the right place.  This will cause the 'revised' value to vary over time instead of being stored as such in the DB.  It will affect the reading of the DB, but will affect it differently each time.

Also, this is a recursive problem.   Calling STFactory.make() indirectly calls `setValueFromResolver()`! 

As noted in the ticket, the `BallotBox.voteForValueWithType()` override on STFactory would be the right place to make this change (with the worker function pushed into a utility).

btangmu: @srl295 "this is the bottleneck for basically READING from the DB, not for STORING"

-- that's not clear to me. As noted in the ticket, the call stack when a user votes is:

putValueAtPath:272, XMLSource (org.unicode.cldr.util)
setValueFromResolver:233, STFactory$DataBackedSource (http://org.unicode.cldr.web/)
voteForValueWithType:1088, STFactory$PerLocaleData (http://org.unicode.cldr.web/)
voteForValueWithType:369, STFactory$PerLocaleData (http://org.unicode.cldr.web/)
handleVote:377, VoteAPIHelper (org.unicode.cldr.web.api)
vote:270, VoteAPI (org.unicode.cldr.web.api)

"set" and "put" both look like STORING, not READING -- so where does STORING really happen?

btangmu: "STFactory.make() indirectly calls setValueFromResolver()"

even if the CLDRFile is already cached?

srl295: that cache is invalidated if the value changes.

PR #2860, opened 2023-04-14
CLDR-16390 Tooling and functions for Inheritance Explaining

srl295:
CLDR-16390

- new tool PathInfo
- Option.getHelp() now public and useful
- new class LocaleInheritanceInfo to track inheritance traces
- added documentation to With for using forEach
- updated XMLSource.getPathLocation mechanism to track LocaleInheritanceInfo details
- added code to CLDRFile.getName to track glossonymn construction

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
@@ -1,0 +1,149 @@
+package org.unicode.cldr.tool;
+
+import java.io.BufferedReader;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Set;
+
+import org.unicode.cldr.tool.Option.Options;
+import org.unicode.cldr.util.*;
+import org.unicode.cldr.util.CLDRFile.Status;
+import org.unicode.cldr.util.XMLSource.SourceLocation;
+
+@CLDRTool(alias = "pathinfo", description = "Get information about paths such as inheritance")
+public class PathInfo {
+    final static Options options = new Options(PathInfo.class)
+    .add("locale", ".*", "und", "Locale ID for the path info")
+    .add("infile", ".*", null, "File to read paths from or '-'' for stdin");
+
+    public static void main(String args[]) throws IOException {
+        final Set<String> paths = options.parse(args, true);
+
+        final String inPath = options.get("infile").getValue();
+        if (inPath != null) {
+            if (inPath.equals("-")) {
+                addLines(System.in, paths);
+            } else {
+                addLines(inPath, paths);
+            }
+        }
+
+        if (paths.isEmpty()) {
+            System.err.println("Error: No paths were specified.");
+            System.out.println(options.getHelp());
+            System.exit(1);
+        }
+
+        CLDRFile file = null;
+
+        final String locale = options.get("locale").getValue();
+        try {
+            file = CLDRConfig.getInstance().getCLDRFile(locale, true);
+        } catch (Throwable t) {
+            System.err.println("Could not load locale " + locale + " due to " + t);
+        }
+
+        showPaths(paths, file);
+        if (file == null) {
+            System.err.println("To enable inheritance lookup, use a different path.");
+        }
+    }
+
+    private static void showPaths(final Set<String> paths, CLDRFile file) {
+        boolean prePopulated = false;
+        boolean hadStringMisses = false;
+
+        for (final String path : paths ) {
+            if (path.startsWith("//")) {
+                showPath(path, file);
+            } else {
+                if (!prePopulated && file != null) {
+                    // scan all possible XPaths
+                    System.err.println("INFO: Scanning all possible xpaths for hex IDs...");
+                    // TODO: include fullpath, etc etc...
+                    file.fullIterable().forEach(x -> StringId.getHexId(x));
+                    prePopulated = true;
+                }
+                final String xpath = StringId.getStringFromHexId(path);
+                if (xpath == null) {
+                    System.err.println("• ERROR: could not find hex id " + path + " - may not have been seen yet.");
+                    hadStringMisses = true;
+                }
+                showPath(xpath, file);
+            }
+        }
+        if (hadStringMisses) {
+            System.err.println("ERROR: One or more XPaths could not be converted from hex form.");
+            if(file == null) {
+                System.err.println("Tip: Set a locale ID with -l when using hex ids");
+            }
+            System.exit(1);
+        }
+    }
+
+    private static void showPath(final String path, CLDRFile file) {
+        System.out.println("-------------------\n");
+        System.out.println("• " + path);
+        System.out.println("• " + StringId.getHexId(path));
+        // TODO: PathHeader, etc.
+        if (file == null) {
+            return;
+        }
+        final String dPath = CLDRFile.getDistinguishingXPath(path, null);
+        if (dPath != null && !dPath.equals(path)) {
+            System.out.println("• Distinguishing: " + dPath);
+            System.out.println("• Distinguishing:" + StringId.getHexId(dPath));
+        }
+        System.out.println("• Value: " + file.getStringValue(dPath));
+        // File lookup
+        SourceLocation source = file.getSourceLocation(path);
+        if (source != null) {
+            System.out.println(source + " XML Source Location");
+        }
+        // Inheritance lookup
+        final String fullPath = file.getFullXPath(path);
+        if (!fullPath.equals(path)) {
+            System.out.println("• Full path: " + fullPath);
+            System.out.println("• Full path: " + StringId.getHexId(fullPath));
+        }
+
+        Status status = new Status();
+        if (false) {
+            // For debugging: compare the below to calling getSourceLocaleIdExtended directly.
+            final String xlocale = file.getSourceLocaleIdExtended(dPath, status, true, null);
+            System.out.println("• SLIE = " + xlocale+":"+status.pathWhereFound);
+        }
+        System.out.println("• Inheritance chain:");
+        for (final LocaleInheritanceInfo e : file.getPathsWhereFound(dPath)) {
+            System.out.println("\n  • " + e); // reason : locale + xpath
+            if (e.getLocale() != null && e.getPath() != null && !e.getLocale().equals("code-fallback")) {
+                final CLDRFile subFile = CLDRConfig.getInstance().getCLDRFile(e.getLocale(), false);
+                SourceLocation subsource = subFile.getSourceLocation(e.getPath());
+                if (subsource != null) {
+                    System.out.println("    " +subsource + " XML Source");
+                }
+            }
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private static void addLines(final InputStream in, Set<String> set) throws IOException {
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8))) {
+            With.in(new FileReaders.ReadLineSimpleIterator(br))
+            .forEach(str -> {
+                if (!str.isBlank()) {
+                    set.add(str.trim());
+                }
+            });
+        }
+    }
+
+    private static void addLines(final String path, Set<String> set) throws IOException {
+        try (FileInputStream fis = new FileInputStream(path)) {
+            addLines(fis, set);
+        }
+    }
+}
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
@@ -12,6 +12,7 @@ import org.unicode.cldr.tool.Option.Options;
 import org.unicode.cldr.util.*;
 import org.unicode.cldr.util.CLDRFile.Status;
 import org.unicode.cldr.util.XMLSource.SourceLocation;
+import static org.unicode.cldr.util.XMLSource.CODE_FALLBACK_ID;
 
 @CLDRTool(alias = "pathinfo", description = "Get information about paths such as inheritance")
 public class PathInfo {
@@ -119,7 +120,7 @@ public class PathInfo {
         System.out.println("• Inheritance chain:");
         for (final LocaleInheritanceInfo e : file.getPathsWhereFound(dPath)) {
             System.out.println("\n  • " + e); // reason : locale + xpath
-            if (e.getLocale() != null && e.getPath() != null && !e.getLocale().equals("code-fallback")) {
+            if (e.getLocale() != null && e.getPath() != null && !e.getLocale().equals(CODE_FALLBACK_ID)) {
                 final CLDRFile subFile = CLDRConfig.getInstance().getCLDRFile(e.getLocale(), false);
                 SourceLocation subsource = subFile.getSourceLocation(e.getPath());
                 if (subsource != null) {
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
@@ -18,7 +18,7 @@ import static org.unicode.cldr.util.XMLSource.CODE_FALLBACK_ID;
 public class PathInfo {
     final static Options options = new Options(PathInfo.class)
     .add("locale", ".*", "und", "Locale ID for the path info")
-    .add("infile", ".*", null, "File to read paths from or '-'' for stdin");
+    .add("infile", ".*", null, "File to read paths from or '--infile=-' for stdin");
 
     public static void main(String args[]) throws IOException {
         final Set<String> paths = options.parse(args, true);
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -502,9 +502,9 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
      */
     public List<LocaleInheritanceInfo> getPathsWhereFound(String xpath) {
         List<LocaleInheritanceInfo> list = new LinkedList<>();
-        // first, call getBaileyValue
+        // first, call getSourceLocaleIdExtended to populate the list
         Status status = new Status();
-        final String locale1 = getSourceLocaleIdExtended(xpath, status, true, list);
+        getSourceLocaleIdExtended(xpath, status, true, list);
         final String path1 = status.pathWhereFound;
         // For now, the only special case is Glossonym
         if (path1.equals(GlossonymConstructor.PSEUDO_PATH)) {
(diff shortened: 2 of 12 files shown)


macchiati:
In the description, can you give a summary of what you are doing in each file and why?

srl295:
Yes. This is very early

srl295:
Sample runs:

### `PathInfo -l en_AU 774230f76f27a54e 3f16ed8804cebb7d 72a282be0770ccdf`

```
• //ldml/localeDisplayNames/languages/language[@type="lv"]
• 774230f76f27a54e
• Value: Latvian
/Users/srl295/src/cldr/common/main/en.xml:389:42:  XML Source Location
• Inheritance chain:

  • inheritancemarker: en_AU://ldml/localeDisplayNames/languages/language[@type="lv"]
    /Users/srl295/src/cldr/common/main/en_AU.xml:379:38:  XML Source

  • parent: en_001://ldml/localeDisplayNames/languages/language[@type="lv"]

  • value: en://ldml/localeDisplayNames/languages/language[@type="lv"]
    /Users/srl295/src/cldr/common/main/en.xml:389:42:  XML Source
-------------------

• //ldml/localeDisplayNames/languages/language[@type="zh_Hans"]
• 3f16ed8804cebb7d
• Value: Simplified Chinese
/Users/srl295/src/cldr/common/main/en_AU.xml:684:58:  XML Source Location
• Inheritance chain:

  • value: en_AU://ldml/localeDisplayNames/languages/language[@type="zh_Hans"]
    /Users/srl295/src/cldr/common/main/en_AU.xml:684:58:  XML Source
-------------------

• //ldml/localeDisplayNames/languages/language[@type="chp"]
• 72a282be0770ccdf
• Value: Chipewyan
/Users/srl295/src/cldr/common/main/en.xml:129:45:  XML Source Location
• Inheritance chain:

  • inheritancemarker: en_AU://ldml/localeDisplayNames/languages/language[@type="chp"]
    /Users/srl295/src/cldr/common/main/en_AU.xml:123:39:  XML Source

  • parent: en_001://ldml/localeDisplayNames/languages/language[@type="chp"]

  • value: en://ldml/localeDisplayNames/languages/language[@type="chp"]
    /Users/srl295/src/cldr/common/main/en.xml:129:45:  XML Source
```

### `PathInfo -l cu 774230f76f27a54e 3f16ed8804cebb7d 72a282be0770ccdf`

```
• //ldml/localeDisplayNames/languages/language[@type="lv"]
• 774230f76f27a54e
• Value: латві́йскїй
/Users/srl295/src/cldr/common/main/cu.xml:49:66:  XML Source Location
• Full path: //ldml/localeDisplayNames/languages/language[@type="lv"][@draft="unconfirmed"]
• Full path: e1468dc2de48eec
• Inheritance chain:

  • value: cu://ldml/localeDisplayNames/languages/language[@type="lv"]
    /Users/srl295/src/cldr/common/main/cu.xml:49:66:  XML Source
-------------------

• //ldml/localeDisplayNames/languages/language[@type="zh_Hans"]
• 3f16ed8804cebb7d
• Value: ᲂу҆проще́нный хи́нскїй
/Users/srl295/src/cldr/common/main/cu.xml:60:82:  XML Source Location
• Full path: //ldml/localeDisplayNames/languages/language[@type="zh_Hans"][@draft="unconfirmed"]
• Full path: 6c6ff86c669f6bb1
• Inheritance chain:

  • value: cu://ldml/localeDisplayNames/languages/language[@type="zh_Hans"]
    /Users/srl295/src/cldr/common/main/cu.xml:60:82:  XML Source
-------------------

• //ldml/localeDisplayNames/languages/language[@type="chp"]
• 72a282be0770ccdf
• Value: chp
• Inheritance chain:

  • parent: cu://ldml/localeDisplayNames/languages/language[@type="chp"]

  • parent: root://ldml/localeDisplayNames/languages/language[@type="chp"]

  • codefallback: code-fallback://ldml/localeDisplayNames/languages/language[@type="chp"]
```

###  `PathInfo -l agq 3f16ed8804cebb7d`

```

• //ldml/localeDisplayNames/languages/language[@type="zh_Hans"]
• 3f16ed8804cebb7d
• Value: Chàenê (Hans)
• Inheritance chain:

  • constructed: agq://ldml/localeDisplayNames/languages/language[@type="zh"]
    /Users/srl295/src/cldr/common/main/agq.xml:58:41:  XML Source

  • constructed: root://ldml/localeDisplayNames/localeDisplayPattern/localePattern
    /Users/srl295/src/cldr/common/main/root.xml:21:44:  XML Source

  • constructed: code-fallback://ldml/localeDisplayNames/scripts/script[@type="Hans"]
```

srl295:
@macchiati @btangmu  any other interesting cases I should try?

- [ ] need to add some unit tests, etc

srl295:
CLI tool for dev but would work very well within surveytool

btangmu:
> any other interesting cases

Some "one" versus "other" paths, as in https://docs.google.com/document/d/1F_WODFRIgBgn2CGHVyPpDi2iMMqhFzHL3JCIaaPHurI/edit#

Related comment in https://unicode-org.atlassian.net/browse/CLDR-16239

Locale = en_001
Path = //ldml/units/unitLength[@type="short"]/unit[@type="angle-revolution"]/unitPattern[@count="one"]

Scenario One: en has hard values
	en_001 getPathLocation finds it present in en, done
 	good "{0} rev" singular value for "one" path

Scenario Two: en has ^^^
	en_001 getPathLocation skips past ^^^ in en
		finds null in root
		looks for an alias in root, doesn't fine one
		sees that path doesn't contain "@alt"
		sees that path does contain "@count"
		changes “count=one" to “count=other", done
	bad "{0} revs" plural value for "one” path

tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
```diff
@@ -0,0 +1,149 @@
...
+        Status status = new Status();
+        if (false) {
+            // For debugging: compare the below to calling getSourceLocaleIdExtended directly.
+            final String xlocale = file.getSourceLocaleIdExtended(dPath, status, true, null);
+            System.out.println("• SLIE = " + xlocale+":"+status.pathWhereFound);
+        }
+        System.out.println("• Inheritance chain:");
+        for (final LocaleInheritanceInfo e : file.getPathsWhereFound(dPath)) {
+            System.out.println("\n  • " + e); // reason : locale + xpath
+            if (e.getLocale() != null && e.getPath() != null && !e.getLocale().equals("code-fallback")) {
```
btangmu, line 122: "code-fallback" must have a constant somewhere...

CLDRResolver.java:
```
    public static final String CODE_FALLBACK = "code-fallback"
```

also XMLSource.java:
```
    public static final String CODE_FALLBACK_ID = "code-fallback";
```

srl295: somewheres… ok, will use that.

I think for this API it would be better to actually omit the (pseudo) locale, because `codefallback` is already the reason code.

srl295:
> > any other interesting cases
> 
> Some "one" versus "other" paths, as in https://docs.google.com/document/d/1F_WODFRIgBgn2CGHVyPpDi2iMMqhFzHL3JCIaaPHurI/edit#
> 
> Related comment in https://unicode-org.atlassian.net/browse/CLDR-16239
> 
> Locale = en_001 Path = //ldml/units/unitLength[@type="short"]/unit[@type="angle-revolution"]/unitPattern[@count="one"]
> 
> Scenario One: en has hard values en_001 getPathLocation finds it present in en, done good "{0} rev" singular value for "one" path
> 
> Scenario Two: en has ^^^ en_001 getPathLocation skips past ^^^ in en finds null in root looks for an alias in root, doesn't fine one sees that path doesn't contain "@alt" sees that path does contain "@count" changes “count=one" to “count=other", done bad "{0} revs" plural value for "one” path

thanks. 

For unit testing I can definitely feed this completely static xml files.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
srl295: OK, I still need this test, because the constructed path has to have a locale.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -488,6 +494,36 @@ public String getBaileyValue(String xpath, Output<String> pathWhereFound, Output
...
+     * values can have multiple contributory paths.
+     * @return map from xpath to locale of one or more contributing paths in the case of glossonymns or other
+     * constructed values.
+     */
+    public List<LocaleInheritanceInfo> getPathsWhereFound(String xpath) {
+        List<LocaleInheritanceInfo> list = new LinkedList<>();
+        // first, call getBaileyValue
+        Status status = new Status();
+        final String locale1 = getSourceLocaleIdExtended(xpath, status, true, list);
+        final String path1 = status.pathWhereFound;
```
btangmu, line 508: The comment says "first, call getBaileyValue" -- and yet, getBaileyValue isn't called here at least not directly

The relationship between getSourceLocaleIdExtended and getBaileyValue is complicated

Note that CLDRFile.getBaileyValue calls getSourceLocaleID

So I'm in doubt what the PathInfo class/tool is really showing, or should be showing -- my impression is that CLDRFile.getBaileyValue is the single most important function for us to be able to explain

Note that XMLSource.ResolvingSource.getBaileyValue is very distinct from CLDRFile.getBaileyValue and the latter is generally the final decider of inheritance at least as far as Survey Tool is concerned...

srl295:
### angle-revolution one

```
• //ldml/units/unitLength[@type="short"]/unit[@type="angle-revolution"]/unitPattern[@count="one"]
• 298a9846ec06970b
• Value: {0} rev
• Inheritance chain:

  • parent: root://ldml/units/unitLength[@type="short"]/unit[@type="angle-revolution"]/unitPattern[@count="one"]

  • value: root://ldml/units/unitLength[@type="short"]/unit[@type="angle-revolution"]/unitPattern[@count="other"]
    /Users/srl295/src/cldr/common/main/root.xml:4437:53:  XML Source
```

- [ ] helpful, but doesn't explain yet. TODO

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
srl295: it's a leftover. I actually didn't want to call getBaileyValue here.

srl295: This sounds like you're wanting a _Bailey_ explainer. The ticket was originally setup to attempt to explain where an actual value's inheritance came from.

srl295:
The reason for the infile parameter is that the XPaths were getting mangled being passed through maven.  With this I can do:

```shell
$ echo '//ldml/units/unitLength[@type="short"]/unit[@type="angle-revolution"]/unitPattern[@count="one"]' | cldr-tool PathInfo '-l root --infile=-'
```

… and it will process a list of xpaths

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
srl295: anyway, i think a Bailey explainer could probably work similarly

btangmu: yes, I think we need a Bailey explainer, above all

could it be, that an inheritance explainer and a bailey explainer are exactly the same thing if the value is inherited

and if the value isn't inherited, there's nothing to explain?

srl295: let's discuss tomorrow… in any event, right now if you run this where there's an exact value it will tell you there's a value… if it's inherited, it will explain the inheritance.

btangmu: that has been my (mis-?)understanding...

now we need a bailey explainer explainer, and a bailey explainer explainer explainer, ad infinitum

btangmu: seriously, though ... even when a value is not inherited, we still need to know the bailey value, and explain the bailey value

srl295: So look at this comment => https://github.com/unicode-org/cldr/pull/2860#issuecomment-1512108214 Is that what you'd expect from this tool or something different? (Assuming it explains _how_ the inheritance got that way, which it doesn't right now)

srl295: Can we start with an inheritance explainer and add Bailey in the next round?

srl295:
I plan to add some unit tests and to improve the wording when aliases are encountered, but i think this should be ready to review tomorrow.

btangmu:
> angle-revolution one

it would be ideal to show the specified locale and path explicitly, as well as the locale-where-found, path-where-found, and the value

the explanation ideally should be like a breadcrumb trail, telling what happened at each key step of the inheritance algorithm -- whether it went vertically/horizontally/etc, and why -- very much like the vote-resolution explainer

srl295:
> > angle-revolution one
> 
> 
> 
> it would be ideal to show the specified locale and path explicitly, as well as the locale-where-found, path-where-found, and the value
> 
> 
> 
> the explanation ideally should be like a breadcrumb trail, telling what happened at each key step of the inheritance algorithm -- whether it went vertically/horizontally/etc, and why -- very much like the vote-resolution explainer

The locale is shown here in this case root. It doesn't yet say why the sideways Alias happened but that's coming. 

This is the CLI tool showing what will be available to the front end. 

I think this structure will serialize well.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
macchiati: BTW, it is too hard to find constants like this, which makes it less likely they'll be used. Suggest we have a single folder for files that just have constants in them (eg no dependencies on any code).

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
macchiati: They are _very_ closely related. The bailey value is precisely what would be inherited if a value didn't exist.
The only difference would be that inheritance stops when the first hard value is found.

macchiati:
> @macchiati @btangmu any other interesting cases I should try?
> 
> * [ ]  need to add some unit tests, etc

Paths in calendars represent good cases, since they end up with long inheritance chains.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/PathInfo.java
srl295: or even a single file, and then static import them. We have LDMLConstants.java in the icu package, we could move and repurpose it.  and move all constants there.

srl295: a better name would just be Constants.java

btangmu:
If the value is inherited, I wonder whether this code always agrees with CLDRFile.getBaileyValue -- that is, ends up with the same value, locale-where-found, and path-where-found

It's not clear how it could always agree, except by some lucky coincidence

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -3623,8 +3736,14 @@ private String getLstrFallback(String codeType, String code) {
...
                 if (matcher.lookingAt()) {
-                    return matcher.group(1).trim();
+                    temp = matcher.group(1).trim();
                 }
+                // TODO: this code is not called anywhere where 'paths' is used at present.
+                // if (temp != null) {
+                //     if (paths != null) {
+                //         paths.add("//lstreg/"+codeType+"[@type=\"" + code + "\"]"); // pseudo-xpath
+                //     }
+                // }
```
btangmu, line 3746: The TODO is vague -- would the commented-out code suffice supposing 'paths' were used? if so, just uncomment now? Otherwise, could throw an exception if paths isn't null?

I've been trying to avoid TODO comments (except in draft PR) unless they're accompanied by ticket numbers -- otherwise they tend to become wishful thinking that just clutters the code

macchiati:
I agree about the TODOs, we should have a bug attached to each one. Maybe
we can have a test to verify that?

srl295:
> I agree about the TODOs, we should have a bug attached to each one. Maybe we can have a test to verify that?
> […](#)

The JIRA plugin for VSCode suggests filing a ticket every time it sees TODO.  I think there's a plugin on Eclipse that does similar.

srl295:
Rebasing and adding tests. Then I'd like to merge this so it can get some testing.

srl295:
PTAL. I think this is ready to review and merge.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
srl295: done

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
```diff
@@ -1,1331 +1,3288 @@
 <!--
-Copyright © 2003-2010 Unicode, Inc. and others. All rights reserved. Distributed under the Terms of Use in http://www.unicode.org/copyright.html.
-
-Permission is hereby granted, free of charge, to any person obtaining a copy of the Unicode data files and any associated documentation (the "Data Files") or Unicode software and any associated documentation (the "Software") to deal in the Data Files or Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, and/or sell copies of the Data Files or Software, and to permit persons to whom the Data Files or Software are furnished to do so, provided that (a) the above copyright notice(s) and this permission notice appear with all copies of the Data Files or Software, (b) both the above copyright notice(s) and this permission notice appear in associated documentation, and (c) there is clear notice in each modified Data File or in the Software as well as in the documentation associated with the Data File(s) or Software that the data or software has been modified.
+Copyright © 1991-2023 Unicode, Inc.
```
btangmu, line 2: this whole file is test data, right? Why so many changes?

btangmu: does this match another ldml.dtd elsewhere?

srl295: It's an exact update of the production ldml.dtd.

it's frozen here for testing purposes.

srl295: `- update the 'testdata' common tree, with a new DTD`

btangmu (approved):
This looks very promising!

About getBaileyValue

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
srl295: @btangmu it also shows how we can work with a very controlled data tree, might be needed for inheritance tests.

PR #2862, opened 2023-04-15
CLDR-15372 Support Popular Ethiopic Transliteration Systems (v44)

dyacob:
CLDR-15372

* Fixes for a variable name misspelling where 2 letters were transposed (ራዕብ → ራብዕ).  
* The `gz-Ethi-t-und-sarb.xml` file has enhancements for the forward transliteration of Musnad numeral sequences.
* The `byn-Ethi-t-byn-latn-m0-xaleget.xml` file has substantial enhancements to support reverse transliteration following experience from a recent project.

- [ X] This PR completes the ticket.

ALLOW_MANY_COMMITS=true


common/testData/transforms/byn-Latn-t-byn-ethi-m0-xaleget.txt
```diff
@@ -1,1091 +1,1091 @@
...
-ሕጩይ	xchuy
-ሕጪት	xchit
-ሕጩፍ	xchuf
-ሕፍት	xft
-ሖጀብ	xojeb
-ሖግ	xog
+ሁለቱም	huletum
+ሁሉ	hulu
+ሁሉም	hulum
+ሁኔታ	huneeta
```
btangmu, line 4: @dyacob the change from hunéta to huneeta could use some clarification. Per the ticket, "Xalaget" = "Transliteration from the Latin convention as defined by the government of Eritrea for education in the Blin language". Is there a reference for Xalaget that clarifies hunéta vs huneeta? And why this file had hunéta previously?

dyacob: @btangmu I was surprised by this and had to wonder about it also. A limitation was that the information that I had to work with did not come directly from the Eritrean Ministry of Education, but from secondary sources. They in turn did not address vowels clearly enough, or I may have misunderstood.  One source clearly was in error, indicating that "ie" should be used.

Converting a full book that came from the education ministry brought the problem to light.  I've attached the introduction to the book here as a reference.
  
[BlinChildrensBook-Intro.pdf](https://github.com/unicode-org/cldr/files/11255006/BlinChildrensBook-Intro.pdf)

btangmu (approved):

PR #2866, opened 2023-04-18
CLDR-16560 Requirements for inher. marker: STFactory

btangmu:

-New method STFactory.PerLocaleData.reviseInheritanceAsNeeded called by setValueFromResolver

-Only change STFactory; no changes to CLDRFile

-TODO comments, work in progress

CLDR-16560

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -18,6 +18,7 @@ import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
+import com.ibm.icu.util.Output;
 import org.unicode.cldr.test.CheckCLDR;
 import org.unicode.cldr.test.DisplayAndInputProcessor;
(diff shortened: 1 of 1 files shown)


btangmu:
This looks similar to https://github.com/unicode-org/cldr/pull/2855 (which was reverted)

Key differences:

- STFactory.PerLocaleData.reviseInheritanceAsNeeded not STFactory.DataBackedSource.reviseInheritanceAsNeeded
- STFactory.PerLocaleData.getFile not STFactory.make
- This PR succeeds in changing the value stored in the db (changed from hard to soft or vice-versa as needed)

Please note that I didn't change CLDRFile since Steven has outstanding changes in that file for the explainer, and also this is PR is only intended as a draft, the ticket still being in design

Also no tests yet -- if we agree on this general strategy then unit test and integration test should follow; unit test just for the new method could be fairly trivial; integration test might use TestSTFactory.xml though that code might need enhancement to distinguish bailey from inheritance marker

srl295 (approved):
great approach

btangmu:
A better alternative, though still not quite right, is: https://github.com/unicode-org/cldr/pull/2880

PR #2871, opened 2023-04-19
CLDR-16390 UI for Inheritance Explainer

srl295:
CLDR-16390

- also refactors some of the component .js code
- also reformats the Info panel a little

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/InfoPanel.vue
+++ b/tools/cldr-apps/js/src/views/InfoPanel.vue
@@ -10,7 +10,6 @@
         >
           ✕
         </a-button>
-        <span class="i-am-info-panel">Info Panel</span>
         <a-button
           shape="circle"
           class="cldr-nav-btn"
@@ -19,10 +18,11 @@
         >
           ↻
         </a-button>
+        <span class="i-am-info-panel">Info Panel</span>
         <a-button
           shape="circle"
           class="cldr-nav-btn"
-          title="Reload"
+          title="Explain"
           @click="explain"
         >
           ↑
--- a/tools/cldr-apps/js/src/views/InfoPanel.vue
+++ b/tools/cldr-apps/js/src/views/InfoPanel.vue
@@ -1,25 +1,54 @@
 <template>
-  <section id="InfoPanelSection">
-    <header class="sidebyside-column-top">
-      <button
-        class="cldr-nav-btn info-panel-closebox"
-        title="Close"
-        @click="closeInfoPanel"
-      >
-        ✕
-      </button>
-      <span class="i-am-info-panel">Info Panel</span>
-      <button class="cldr-nav-btn" title="Reload" @click="reloadInfoPanel">
-        ↻
-      </button>
-    </header>
-  </section>
+  <div>
+    <section id="InfoPanelSection">
+      <header class="sidebyside-column-top">
+        <a-button
+          shape="circle"
+          class="cldr-nav-btn info-panel-closebox"
+          title="Close"
+          @click="closeInfoPanel"
+        >
+          ✕
+        </a-button>
+        <span class="i-am-info-panel">Info Panel</span>
+        <a-button
+          shape="circle"
+          class="cldr-nav-btn"
+          title="Reload"
+          @click="reloadInfoPanel"
+        >
+          ↻
+        </a-button>
+        <a-button
+          shape="circle"
+          class="cldr-nav-btn"
+          title="Reload"
+          @click="explain"
+        >
+          ↑
+        </a-button>
+      </header>
+    </section>
+    <InheritanceExplainer ref="inheritanceExplainer" />
+  </div>
 </template>
 
 <script>
+import { ref } from "vue";
 import * as cldrInfo from "../esm/cldrInfo.js";
+import * as cldrStatus from "../esm/cldrStatus.js";
+import InheritanceExplainer from "./InheritanceExplainer.vue";
 
 export default {
+  setup() {
+    let inheritanceExplainer = ref(null);
+    return {
+      inheritanceExplainer,
+    };
+  },
+  components: {
+    InheritanceExplainer,
+  },
   methods: {
     closeInfoPanel() {
       cldrInfo.closePanel();
@@ -27,6 +56,13 @@ export default {
 
     reloadInfoPanel() {
       cldrInfo.clearCachesAndReload();
+    },
+
+    explain() {
+      this.inheritanceExplainer.explain(
+        cldrStatus.getCurrentLocale(),
+        cldrStatus.getCurrentId()
+      );
     },
   },
 };
@@ -59,5 +95,6 @@ header {
 .i-am-info-panel {
   font-weight: bold;
   margin-right: 1ex;
+  flex-grow: 1;
 }
 </style>
--- a/tools/cldr-apps/js/src/views/InheritanceExplainer.vue
+++ b/tools/cldr-apps/js/src/views/InheritanceExplainer.vue
@@ -7,10 +7,16 @@
   >
     <template #content>
       <a-spin size="large" :delay="500" v-if="visible && !inheritance" />
-      <p></p>
       <a-timeline v-if="visible && inheritance">
         <a-timeline-item
-          v-for="{ locale, newLocale, reason, xpath, xpathFull } in inheritance"
+          v-for="{
+            locale,
+            newLocale,
+            newXpath,
+            reason,
+            xpath,
+            xpathFull,
+          } in inheritance"
           v-bind:color="colorForReason(reason)"
         >
           <p v-bind:title="reason">{{ reasons[reason] || reason }}</p>
@@ -19,19 +25,13 @@
             Locale: <b>{{ locale }}</b>
           </p>
           <!-- <b v-if="locale" v-bind:title="Locale ID">{{ locale }}</b> -->
-          <p v-if="xpath" class="xpath">
+          <p v-if="xpath && newXpath" class="xpath">
             {{ xpathFull }}
-            <!-- show the Go button, only if we have locale AND xpath, and if it's actually not where we already  -->
-            <a-button
-              v-if="
-                locale &&
-                xpath &&
-                (xpath !== itemXpath || locale !== itemLocale)
-              "
-              @click="go(locale, xpath)"
-              >Go</a-button
-            >
           </p>
+          <!-- show the Go button, only if we have locale AND xpath. Show always, so we can navigate   -->
+          <a-button v-if="locale && xpath" @click="go(locale, xpath)"
+            >Go</a-button
+          >
         </a-timeline-item>
       </a-timeline>
     </template>
@@ -93,14 +93,14 @@ export default {
     colorForReason(reason) {
       return (
         {
-          novalue: "gray",
+          none: "gray",
           value: "green",
-          itemalias: "teal",
-          codefallback: "red",
+          itemAlias: "teal",
+          codeFallback: "red",
           constructed: "purple",
-          alt: "blue",
-          count: "yellow",
-          inheritancemarker: "orange",
+          removedAttribute: "blue",
+          changedAttribute: "yellow",
+          inheritanceMarker: "orange",
         }[reason] || null
       );
     },
--- a/tools/cldr-apps/js/src/views/InheritanceExplainer.vue
+++ b/tools/cldr-apps/js/src/views/InheritanceExplainer.vue
@@ -7,32 +7,39 @@
   >
     <template #content>
       <a-spin size="large" :delay="500" v-if="visible && !inheritance" />
-      <a-timeline v-if="visible && inheritance">
+      <a-timeline id="inheritanceTimeline" v-if="visible && inheritance">
         <a-timeline-item
           v-for="{
+            attribute,
             locale,
             newLocale,
             newXpath,
             reason,
+            showReason,
             xpath,
             xpathFull,
-            attribute,
           } in inheritance"
           v-bind:color="colorForReason(reason)"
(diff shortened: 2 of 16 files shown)


tools/cldr-apps/js/src/views/InfoPanel.vue
```diff
@@ -1,25 +1,54 @@
...
+          class="cldr-nav-btn"
+          title="Reload"
+          @click="reloadInfoPanel"
+        >
+          ↻
+        </a-button>
+        <a-button
+          shape="circle"
+          class="cldr-nav-btn"
+          title="Reload"
```
btangmu, line 25: "Reload" is wrong here -- should be something like "Explain inheritance"

btangmu: When this is extended to include "bailey explainer", there should be an explanation of the bailey value, as well as an explanation of the winning value (if they're not the same) -- I guess a single control (button) to show both is better than separate controls attached to the winning and bailey values

tools/cldr-apps/js/src/views/InheritanceExplainer.vue
```diff
@@ -0,0 +1,115 @@
+<template>
+  <a-popover
+    v-model:visible="visible"
+    placement="bottomLeft"
+    title="Inheritance"
+    trigger="click"
+  >
+    <template #content>
+      <a-spin size="large" :delay="500" v-if="visible && !inheritance" />
+      <p></p>
```
btangmu, line 10: empty paragraph?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private String getXPathByHex(String hexId) {
+        String xpath = null;
+        try {
+            xpath = sm.xpt.getByStringID(hexId);
+        } catch (RuntimeException e) {
+            /*
+             * Don't report the exception. This happens when it simply wasn't found.
+             * Possibly getByStringID, or some version of it, should not throw an exception.
+             */
```
btangmu, line 111: this isn't good karma -- exceptions shouldn't be ignored, especially not generic ones like RuntimeException

better would be to add a new method in XPathTable that simply returns null instead of throwing an exception -- it could wrap getByStringID with a catch (or add a catch within getByStringID itself if that doesn't break compatibility with other callers)

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
```diff
@@ -0,0 +1,149 @@
...
+            return STError.badLocale(localeId);
+        }
+        String xpath = getXPathByHex(hexId);
+        if (xpath == null) {
+            return STError.badPath(hexId);
+        }
+
+        final CLDRFile f = sm.getSTFactory().make(locale, true);
+        final List<LocaleInheritanceInfo> paths = f.getPathsWhereFound(xpath);
+        return Response.ok(new InheritanceResponse(paths)).build();
```
btangmu, line 100: if there's the slightest chance that `new InheritanceResponse(paths))` could throw an exception, it would be best to call it in a try block and return 500 if it throws

safest not to call ok until it's confirmed ok

btangmu (changes requested):

btangmu:
Suggestion: the Reload button should be adjacent to "Info Panel" on the left, for consistency with Dashboard reload button

the up-arrow button for inheritance could stay on the right, or move somewhere else... unlike Reload it doesn't apply in the same way to the entire info panel

tools/cldr-apps/js/src/views/InfoPanel.vue
srl295: > When this is extended to include "bailey explainer", there should be an explanation of the bailey value, as well as an explanation of the winning value (if they're not the same) -- I guess a single control (button) to show both is better than separate controls attached to the winning and bailey values

I'd like some more clarity on what that looks like. We don't surface the Bailey value explicitly as such however it's always one of the candidates. Maybe clicking on that candidate and then clicking this gets you there.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
srl295: This is a copy and paste from elsewhere. But agreed

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
srl295: Throwing here should throw a 500 automatically

srl295:
> Suggestion: the Reload button should be adjacent to "Info Panel" on the left, for consistency with Dashboard reload button
> 
> 
> 
> the up-arrow button for inheritance could stay on the right, or move somewhere else... unlike Reload it doesn't apply in the same way to the entire info panel

Well, it's partly here because it's a vue component

tools/cldr-apps/js/src/views/InfoPanel.vue
btangmu: "We don't surface the Bailey value explicitly as such however it's always one of the candidates" -- I'd say that does count as surfacing -- we just don't surface the term "Bailey"

I still don't really get the point of "inheritance explainer" for a value that's not inherited -- there's not much to explain other than the obvious fact that since it's not inherited it must be present in the current locale+path

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
btangmu: what we've seen in other similar situations, is that the response is mangled

better something like this:

```
         try {
             InheritanceResponse response = new InheritanceResponse(paths)
             return Response.ok(response).build();
         } catch (Exception e) {
             return Response.status(500, "An exception occurred").entity(e).build();
         }
```

See https://unicode-org.atlassian.net/browse/CLDR-16223 which BTW is assigned to you as major for v44, but still "new"?

btangmu:
> it's partly here because it's a vue component

understood -- soon it should be easier to add vue components inside the info panel

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
btangmu: by "mangled" I mean for example "Error: could not load: Status 200 OK; URL: /cldr-apps/api/voting/reports/locales/sv/reports/personnames.html?&cacheKill=1670013698549 Missing response; responseType = text" -- as reported in https://unicode-org.atlassian.net/browse/CLDR-16205

srl295: > what we've seen in other similar situations, is that the response is mangled
> 
> 
> 
> better something like this:
> 
> 
> 
> ```
> 
>          try {
> 
>              InheritanceResponse response = new InheritanceResponse(paths)
> 
>              return Response.ok(response).build();
> 
>          } catch (Exception e) {
> 
>              return Response.status(500, "An exception occurred").entity(e).build();
> 
>          }
> 
> ```
> 
> 
> 
> See https://unicode-org.atlassian.net/browse/CLDR-16223 which BTW is assigned to you as major for v44, but still "new"?

The error handler is supposed to do exactly that already for all paths. May be a bug.

btangmu: "may be a bug" -- could be; anyway safest to create a 500 response explicitly

srl295: So i tried this and any exception does create a 500 automatically, with a stack trace (instead of just 'an exception occurred) and also error logging. I'm not sure what "200 OK" above is, but sounds like a different bug.

Error handling is great if we're going to actually do something with it.  

What might be better instead is to define an ExceptionMapper across our APIs… that uses a standardized response (we can use STError) to return something like `{"err": "An exception occurred"}` with the 500 instead of the HTML page.   Something like [this](https://www.ibm.com/docs/en/was-nd/8.5.5?topic=applications-defining-exception-mappers-resource-exceptions-errors) would do that.  Otherwise, `Response.status(500, "An exception occurred").entity(e).build()` returns less information than the standard mapper and doesn't log. 

image

srl295:
image
Updated wording and ui

srl295:
image

macchiati:
Drop "The value was not found in this locale". That is the default for
everything but the last, so no need to repeat.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
btangmu: I guess it depends on what exactly goes on in the response-returning code, whether an exception there results in a well-formed and informative 500 response, or a bogus 200 response, as in the old person-names report, when it looked like this:

```
  return Response.ok(new StreamingOutput() {
        @Override
        public void write(OutputStream output) throws IOException, WebApplicationException {
            if (chart != null) {
                chart.writeContents(output, stf);
            } else {
                switch (report) {
                // "Old Three" reports
                case compact:
                case datetime:
                case zones:
                    try (final Writer w = new OutputStreamWriter(output);) {
                        SurveyAjax.generateReport("r_" + report.name(), w, sm, loc);
                    }
                    break;
                default:
                    try (final Writer w = new OutputStreamWriter(output);) {
                        w.write("Could not load report: " + report.name());
                    }
                }
            }
        }
    }).build();
```

All I know is that was a pain to debug, and I'd feel less apprehensive if we only called `Response.ok(response)` after successfully obtaining the response object.

srl295: > All I know is that was a pain to debug, and I'd feel less apprehensive if we only called Response.ok(response) after successfully obtaining the response object.

if there's an exception then `Response.ok` isn't called because the exception breaks out.

srl295:
> Drop "The value was not found in this locale". That is the default for everything but the last, so no need to repeat.

Can do, but note that the original plan for this (as an inheritance explainer) was to show the parent locale hierarchy.  For example,  items in `en_FI`  show the path through `en_150` and `en_001` and `en` which might not be as obvious otherwise.

srl295:
@macchiati  and @btangmu  how about this - drop mentioning "none" except the first time, and tighten up the layout.


image

macchiati:
It's inconsistent.
The status is on one Kabyle, but not the other.
The status is after Kabyle, but not after the first root.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
srl295: i added comments to CLDR-16205 let's discuss there

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
srl295: Fixed now.

srl295 (commented):
@btangmu  PTAL i think i resolved the issues

srl295:
>  It's inconsistent.
> The status is on one Kabyle, but not the other.
> The status is after Kabyle, but not after the first root.

I'll drop `none`'s status everywhere then. Then it will be consistent and less noisy.

srl295:
@macchiati  dropped:

image

macchiati:
Looks good. The last root has the message on the same line as locale. For
consistency (and tightness) can be done in other cases. EG

Kabyle [Jump]

Root     Changed attribute: count [Jump]
//ldml/units/...

Kabyle [Jump}

Root    Found: explicit value

srl295:
> Looks good. The last root has the message on the same line as locale. For consistency (and tightness) can be done in other cases. EG Kabyle [Jump] Root Changed attribute: count [Jump] //ldml/units/... Kabyle [Jump} Root Found: explicit value
> […](#)

I see that, but the 'changed attribute' is actually indicating two things:  

1. Root Jump — this is where there isn't a value
2. Changed attribute value — this is _after_ the root value has been considered, _then_ the attribute value changes.  That's why the circle is in yellow and doesn't have a locale. 

I'd propose to keep these separate, since we're trying to trace the process here.  These would correspond to separate steps in lookup.

macchiati:
Makes sense. Separate line for any time the path changes (with reason)

srl295:
@macchiati @btangmu  comments? can we merge this and keep working on the UI?

srl295:
Mark says:
> I think we will need a ℹ️ link to an explainer.

this is specifically the inheritance explainer though… can we merge this so it can be tried out, and iterate on the ui?

macchiati:
Yes, sounds good. BTW, I think we can expand this into a full bailey
explainer pretty easily, by looping. That is, whenever the explainer hits a
value, continue the process by checking for the bailey value at that point.

btangmu:
@macchiati "expand this into a full bailey explainer pretty easily, by looping" -- I suspect that the result wouldn't always agree with CLDRFile.getBaileyValue, and the only way to explain that correctly is by calling CLDRFile.getBaileyValue and deriving the explanation from that function and its callees

btangmu (approved):
"merge this so it can be tried out, and iterate on the ui" -- yes, and also to iterate on the implementation -- I still strongly suspect we need to call CLDRFile.getBaileyValue and trace what it does, in order to have a full-fledged inheritance explainer

PR #2876, opened 2023-04-24
CLDR-16579 VoteAPIHelper.DROP_HARD_INHERITANCE_FROM_RESPONSE

btangmu:
-New boolean VoteAPIHelper.DROP_HARD_INHERITANCE_FROM_RESPONSE

-This results in only soft inheritance in server response

-This is simpler than making VoteResolver.DROP_HARD_INHERITANCE true (maybe not better)

CLDR-16579

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
@@ -34,6 +34,9 @@ import org.unicode.cldr.web.api.VoteAPI.VoteResponse;
 public class VoteAPIHelper {
     private static final boolean DEBUG_SERIALIZATION = false;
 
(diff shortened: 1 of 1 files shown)


btangmu:
This PR is one way we could solve this particular ticket, by only changing the response, entirely limited to VoteAPIHelper and not affecting the rest of the back end's operation.

We probably shouldn't do this. (We can keep it in mind as an alternative if needed.) Instead, we should change VoteResolver.DROP_HARD_INHERITANCE to true, see https://github.com/unicode-org/cldr/pull/2877

Based on local testing both ways (PRs 2876 and 2877), the results mostly look the same on the front end. However, for example, explanations for vote resolution may pointlessly/confusingly distinguish hard/soft votes if VoteResolver.DROP_HARD_INHERITANCE stays false.

btangmu:
I'm closing this PR with expectation we should instead change VoteResolver.DROP_HARD_INHERITANCE to true

PR #2877, opened 2023-04-24
CLDR-16579 Make VoteResolver.DROP_HARD_INHERITANCE true

btangmu:
-Change VoteResolver.DROP_HARD_INHERITANCE to true

CLDR-16579

ALLOW_MANY_COMMITS=true


btangmu:
This is best done in the same PR as https://unicode-org.atlassian.net/browse/CLDR-16560

See, for example, https://github.com/unicode-org/cldr/pull/2880

So I'll close this PR. Possibly this ticket will get closed as "fixed by other ticket"

PR #2878, opened 2023-04-25
CLDR-16590 Coverage Menu Woes

srl295:
- Fix Coverage menu, use responsive array instead of overwriting the coverage menu array

Also for CLDR-16390 Inheritance UI:
- Don't popup the explainer if no locale is selected.

CLDR-16590

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #2879, opened 2023-04-25
CLDR-16590 ui: remove $cldrOpts, add reference variables

srl295:
CLDR-16590

- support 'refs' object in cldrStatus exposing reactive values
- remove $cldrOpts, clean up some references and linkage

Inheritance UI improvements
- show jump button only if it goes somewhere
- show inheritance explainer button only if a row is selected

ALLOW_MANY_COMMITS=true


srl295:
this means anywhere in Vue we can use `$locale` (or `this.$locale` within `<script>` ) to get the current, live locale, et.

btangmu:
I'm really skeptical  whether global refs are better than the existing methods such as cldrStatus.getCurrentLocale or (even better) explicitly providing data such as with method parameters

global data is often considered bad, as in https://embeddedartistry.com/fieldatlas/the-problems-with-global-variables/ 

the idea of a "current locale" is ill-defined for something like Priority Items Summary which refers to multiple locales

granted, ST does have a current locale as far as the main vetting table and dashboard are concerned, but then the current locale is also part of the URL, and it's an explicit parameter in http requests etc.

a "current path" is even more ill-defined since we're typically viewing rows for many paths at the same time, and although a particular row/path may be currently selected, nevertheless many of the views on the screen -- ideally Vue components -- are for different paths

btangmu:
another reference: https://spin.atomicobject.com/2015/09/08/global-data-drawbacks-alternatives/

srl295:
> I'm really skeptical whether global refs are better than the existing methods such as cldrStatus.getCurrentLocale or (even better) explicitly providing data such as with method parameters

getCurrentLocale isn't reactive.  It would mean that, for the case of the "Jump" button, or the "Login" button (which already does this) a listener has to be setup and torn down on each UI component, in order to respond to changes in data.

> global data is often considered bad, as in https://embeddedartistry.com/fieldatlas/the-problems-with-global-variables/

No argument there.  However, I don't think there's a good alternative prior to restructuring the entire page in Vue (which is where I think we will want to go eventually).  Ideally,  cldrStatus, cldrLoad, cldrGui, etc would be basically replaced with a `<SurveyTool/>` app component, which then manages its own UI state, and we move all of the UI state out of the `esm` directory.  

In that model, probably the current locale would be a property in the parent app, and then it would be accessible in children as need be.   See for example https://dev.to/michaelthiessen/vue-tips-22-where-do-you-put-shared-state-g43 (from the excellent book).  

> the idea of a "current locale" is ill-defined for something like Priority Items Summary which refers to multiple locales

It's defined to be cldrStatus.getCurrentLocale, but reactive. It's the current locale the user has navigated to.

> granted, ST does have a current locale as far as the main vetting table and dashboard are concerned, but then the current locale is also part of the URL, and it's an explicit parameter in http requests etc.

Yes, it is. This exposes that to the UI, in a way that reduces non-UI code needed in the components.

> a "current path" is even more ill-defined since we're typically viewing rows for many paths at the same time, and although a particular row/path may be currently selected, nevertheless many of the views on the screen -- ideally Vue components -- are for different paths

Yes, however for example the Jump button is affected by which path is selected.

srl295:
> They are just available on all components

Would you be OK to remove $locale / $xpath as conveniences but keep `cldrRef`, so that components that want to access these refs can explicitly access `cldrRef.locale`?

This would add some small boilerplate in each such component, but have the advantage that we can track which components are accessing cldrRef.  ( I guess searching for $locale would do that also)

srl295:
Anyway, I think this PR is probably the best and most responsive way to implement the improvements to the Jump button and other issues in that class.  But I would be open to an alternative suggestion

btangmu:
There's a problem currently with the inheritance explainer, as shown in this screenshot:
image

I got that by changing "fr" to "de" in the address bar; the explanation in the info panel doesn't get updated

The component is surprisingly persistent, and too persistent in this case

I can see that's true also -- no update -- when there are two buttons for switching between child and parent -- and in fact the persistence seems necessary, otherwise, starting from the parent, how would you know which child or children to show -- so even though you jump to the parent, the info panel is still describing the child

More robust might be to ensure the explainer data gets refreshed whenever the info panel gets refreshed for a new locale/path -- and add in a "most recent sublocale for explainer" variable somewhere to enable the Jump button from parent to child

If you'd like, I could help work on the front end while you work on the back end for getBaileyValue explainer...

srl295:
#2878 is the main PR for the current fixes to the inheritance explainer

> There's a problem currently with the inheritance explainer, as shown in this screenshot: image
> 
> I got that by changing "fr" to "de" in the address bar; the explanation in the info panel doesn't get updated

No, but your first click on anything will dismiss the popover. 

> The component is surprisingly persistent, and too persistent in this case
> 
> I can see that's true also -- no update -- when there are two buttons for switching between child and parent -- and in fact the persistence seems necessary, otherwise, starting from the parent, how would you know which child or children to show -- so even though you jump to the parent, the info panel is still describing the child

The info panel itself doesn't always respond to URL changes either.

Fixed in one line once this PR is in:

```diff
             <a-button
               size="small"
               shape="round"
-              v-if="locale && xpath"
+              v-if="(locale && xpath) && ((locale != $locale.value)||(xpath != $xpath.value))"
               @click="go(locale, xpath)"
               >Jump</a-button
             >
```

srl295:
to motivate discussion, I added a commit that uses this change.  It conflicts with #2878 , so it's just for discussion.

srl295:
> Anyway, I think this PR is probably the best and most responsive way to implement the improvements to the Jump button and other issues in that class. But I would be open to an alternative suggestion

Rebased, and I think this is ready for review on its own now.

tools/cldr-apps/js/src/esm/cldrRef.mjs
```diff
@@ -0,0 +1,13 @@
...
+ * The current user locale. Read only.
+ */
+const locale = ref(null);
+
+/**
+ * The current XPath (hex id). Read only.
+ */
+const xpath = ref(null);
+
+export { locale, xpath };
```
btangmu, line 13: I don't see any need/benefit for this whole mechanism

anyway, if we had to do it, getters/setters would be far preferable to exporting and importing variables

tools/cldr-apps/js/src/esm/cldrStatus.js
```diff
@@ -125,6 +126,7 @@ function setCurrentId(id) {
     id = id.toString();
     currentId = idIsAllowed(id) ? id : "";
   }
+  cldrRef.xpath.value = currentId;
```
btangmu, line 129: a setter would obviate including Vue implementation concerns like ".value"

btangmu: what if "id" is not an xpath at all? sometimes it identifies a user or a forum post

tools/cldr-apps/js/src/views/InfoPanel.vue
```diff
@@ -20,6 +20,7 @@
         </a-button>
         <span class="i-am-info-panel">Info Panel</span>
         <a-button
+          v-if="($locale.value) && ($xpath.value)"
```
btangmu, line 23: this could be implemented without globals, simply by calling a function here that in turn calls getCurrentLocale and getCurrentId

for example, see the function canBeHidden in DashboardWidget.vue

tools/cldr-apps/js/src/views/InheritanceExplainer.vue
```diff
@@ -26,17 +26,16 @@
             <span class="reason" v-if="showReason">{{
               getReason(reason, attribute)
             }}</span>
-            <!-- show the Go button, only if we have locale AND xpath. Show always, so we can navigate   -->
+            <!-- show the Go button, only if we have locale AND xpath. Hide if we're already there.  -->
             <a-button
               size="small"
               shape="round"
-              v-if="locale && xpath"
+              v-if="(locale && xpath) && ((locale != $locale.value)||(xpath != $xpath.value))"
```
btangmu, line 33: again, could just call a function here that in turn calls getCurrentLocale and getCurrentId

btangmu (changes requested):
global variables don't seem to be needed, though of course I may be missing something

please see my comments about using v-if with function instead

tools/cldr-apps/js/src/views/InfoPanel.vue
srl295: It's actually different.  $locale here is reevaluated dynamically when the locale changes.   Calling getCurrentLocale() would not have the same behavior.

tools/cldr-apps/js/src/views/InheritanceExplainer.vue
srl295: see above

tools/cldr-apps/js/src/esm/cldrStatus.js
srl295: I thought this particular currentId was always an xpath.

tools/cldr-apps/js/src/views/InfoPanel.vue
srl295: details: https://vuejs.org/guide/essentials/reactivity-fundamentals.html

tools/cldr-apps/js/src/esm/cldrStatus.js
btangmu: unfortunately not

tools/cldr-apps/js/src/views/InfoPanel.vue
btangmu: reactivity is great; reactive global variables, not so great, and not necessary

the info panel gets regenerated anyway whenever the locale or path changes, or at least it always has been that way until now

the persistence of a component in the info panel when the locale or path changes is an interesting idea and I'm not opposed to it, if it's implemented without global variables

getCurrentLocale and getCurrentId will work -- at most, a few lines of code might be needed to make the component refresh when the info panel is regenerated -- I'd be happy to implement it

another alternative is for cldrInfo to call setters for locale/path in the component when the info panel is regenerated -- that way locale/path can be reactive (and not global) -- theoretically that could prevent some flicker in the UI, though realistically it's unlikely to be a perceptible difference given that the rest of the window is also updating to show different data for the new locale/path -- again, I'd be happy to implement it

btangmu (approved):
Removing $cldrOpts is good. I don't actually believe cldrStatus.refs is necessary or the best solution. It would be better for each reactive data item to be encapsulated in the component that uses it, and only be visible externally through a setter in the component. We should decrease, not increase, dependency on the idea of "current" locale and path for the entire app. It would be better for the scope to be cldrInfo instead of cldrStatus. As long as we have only one Info Panel, it's meaningful to speak of the current path and locale that are described in the Info Panel, though it would be more ideal to assume that eventually there could be more than one Info Panel visible at the same time, each with its own current locale/path. So, locale/path should be properties of the Info Panel, not global properties. I'm approving the PR anyway, since the ticket is blocking and I don't want to impede progress.

srl295:
> Removing $cldrOpts is good. I don't actually believe cldrStatus.refs is necessary or the best solution. It would be better for each reactive data item to be encapsulated in the component that uses it, and only be visible externally through a setter in the component. We should decrease, not increase, dependency on the idea of "current" locale and path for the entire app. It would be better for the scope to be cldrInfo instead of cldrStatus. As long as we have only one Info Panel, it's meaningful to speak of the current path and locale that are described in the Info Panel, though it would be more ideal to assume that eventually there could be more than one Info Panel visible at the same time, each with its own current locale/path. So, locale/path should be properties of the Info Panel, not global properties. I'm approving the PR anyway, since the ticket is blocking and I don't want to impede progress.

Understood. 

It does give us a single place for reactivity rather than scattered. Right now the header also has its own reactivity.

PR #2880, opened 2023-04-25
CLDR-16560 Requirements for storing inher. marker: STFactory/VoteResolver

btangmu:
-Change VoteResolver.DROP_HARD_INHERITANCE to true

-New method STFactory.PerLocaleData.reviseInheritanceAsNeeded called by setValueFromResolver

-New method VoteResolver.reviseInheritanceAsNeeded called by STFactory.PerLocaleData.reviseInheritanceAsNeeded

CLDR-16560

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -1072,6 +1072,7 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
                 throws BallotBox.InvalidXPathException, BallotBox.VoteNotAcceptedException {
             makeSureInPathsForFile(distinguishingXpath, user, value);
             value = processValue(distinguishingXpath, value);
+            value = reviseInheritanceAsNeeded(distinguishingXpath, value);
             SurveyLog.debug(
                     "V4v: "
                             + locale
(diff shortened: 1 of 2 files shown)


btangmu:
This is very similar to https://github.com/unicode-org/cldr/pull/2866 -- key differences:

- the essential new method is in VoteResolver rather than STFactory
- DROP_HARD_INHERITANCE is true

This doesn't work right yet. For this path,

http://localhost:9080/cldr-apps/v#/fr_CA/Alphabetic_Information/5d9dd084fd590ee

Running the main branch locally, Jump to Original goes to "fr" ( = theRow.inheritedLocale)

Running this branch locally, Jump to Original goes to "root" ( = theRow.inheritedLocale); this is bad since we generally don't want to inherit from root (even if it matches bailey). We check for that in DataRow.inheritsFromRootOrFallback...

This is an example of where a full-fledged "inheritance explainer" (or "bailey explainer") would be extremely helpful!

btangmu:
The problem (described with an example in my previous comment) involves lack of consistency in the following:

1. DataPage.DataRow.inheritsFromRootOrFallback is used to prevent converting hard value to inheritance marker if the inheritance would be from root/fallback; it only affects the candidate items sent to the front end
2. VoteResolver.reviseInheritanceAsNeeded converts inheritance marker to hard value, or vice-versa, to enforce "requirements" involving sideways inheritance; it only affects what gets stored in CLDRFile and the database
3. VoteResolver.changeBaileyToInheritance converts the hard value to inheritance marker unconditionally, making no distinction whether inheritance is from root/fallback

btangmu:
See https://github.com/unicode-org/cldr/pull/2889

PR #2881, opened 2023-04-28
CLDR-16390 Bailey explainer

srl295:
- update the implementations behind getPathsForValue to keep going when a hard value is hit. Won't affect getSourceLocale, etc, return values,
but will populate the tracing list with additional values.

- [ ] This PR completes the ticket

CLDR-16390

ALLOW_MANY_COMMITS=true

-----
Example output:

```
• //ldml/localeDisplayNames/territories/territory[@type="PM"]
• 576f1c82a03a22a8
• Value: Saint-Pierre-et-Miquelon
• Inheritance chain:
  • value: en_CA://ldml/localeDisplayNames/territories/territory[@type="PM"]
  • value: en://ldml/localeDisplayNames/territories/territory[@type="PM"]
  • none: root://ldml/localeDisplayNames/territories/territory[@type="PM"]
  • codeFallback: //ldml/localeDisplayNames/territories/territory[@type="PM"]
```

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -570,10 +570,8 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         LinkedList<LocaleInheritanceInfo> list = new LinkedList<>();
         // first, call getSourceLocaleIdExtended to populate the list
         Status status = new Status();
-        final String locale1 = // for debugging
-                getSourceLocaleIdExtended(xpath, status, false, list);
+        getSourceLocaleIdExtended(xpath, status, false, list);
         final String path1 = status.pathWhereFound;
-        System.out.println("Path: " + path1 + ", loc: " + locale1);
         // For now, the only special case is Glossonym
         if (path1.equals(GlossonymConstructor.PSEUDO_PATH)) {
             // it's a Glossonym, so as the GlossonymConstructor what the paths are.  Sort paths in
@@ -712,6 +710,11 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         if (result == XMLSource.CODE_FALLBACK_ID && dataSource.isResolving()) {
             final String fallbackPath = getFallbackPath(distinguishedXPath, false, true);
             if (fallbackPath != null && !fallbackPath.equals(distinguishedXPath)) {
+                if (list != null) {
+                    list.add(
+                            new LocaleInheritanceInfo(
+                                    getLocaleID(), distinguishedXPath, Reason.fallback, null));
+                }
                 result =
                         dataSource.getSourceLocaleIdExtended(
                                 fallbackPath, status, skipInheritanceMarker, list);
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -554,32 +554,47 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
     }
 
     /**
-     * Similar to {@link #getBaileyValue(String, Output, Output)} except returns a map, as
-     * constructed values can have multiple contributory paths.
+     * Return a list of all paths which contributed to the value, as well as all bailey values. This
+     * is used to explain inheritance and bailey values. The list must be interpreted in order. When
+     * {@link LocaleInheritanceInfo.Reason#isTerminal()} return true, that indicates a successful
+     * lookup and partitions values from subsequent bailey values.
      *
-     * @return map from xpath to locale of one or more contributing paths in the case of glossonymns
-     *     or other constructed values.
+     * @see #getBaileyValue(String, Output, Output)
+     * @see #getSourceLocaleIdExtended(String, Status, boolean)
      */
     public List<LocaleInheritanceInfo> getPathsWhereFound(String xpath) {
-        List<LocaleInheritanceInfo> list = new LinkedList<>();
+        if (!isResolved()) {
+            throw new IllegalArgumentException(
+                    "getPathsWhereFound() is only valid on a resolved CLDRFile");
+        }
+        LinkedList<LocaleInheritanceInfo> list = new LinkedList<>();
         // first, call getSourceLocaleIdExtended to populate the list
         Status status = new Status();
-        getSourceLocaleIdExtended(xpath, status, true, list);
+        final String locale1 = // for debugging
+                getSourceLocaleIdExtended(xpath, status, false, list);
         final String path1 = status.pathWhereFound;
+        System.out.println("Path: " + path1 + ", loc: " + locale1);
         // For now, the only special case is Glossonym
         if (path1.equals(GlossonymConstructor.PSEUDO_PATH)) {
-            list.clear(); // Rebuild this
-            // it's a Glossonym, so dig in
+            // it's a Glossonym, so as the GlossonymConstructor what the paths are.  Sort paths in
+            // reverse order.
             final Set<String> xpaths =
-                    new GlossonymConstructor(this).getPathsWhereFound(xpath, new TreeSet<String>());
+                    new GlossonymConstructor(this)
+                            .getPathsWhereFound(
+                                    xpath, new TreeSet<String>(Comparator.reverseOrder()));
             for (final String subpath : xpaths) {
-                // recurse
                 final String locale2 = getSourceLocaleIdExtended(subpath, status, true);
                 final String path2 = status.pathWhereFound;
-                list.add(new LocaleInheritanceInfo(locale2, path2, Reason.constructed));
+                // Paths are in reverse order (c-b-a) so we insert them at the top of our list.
+                list.addFirst(new LocaleInheritanceInfo(locale2, path2, Reason.constructed));
             }
-        } else {
-            return list;
+
+            // now the list contains:
+            // constructed: a
+            // constructed: b
+            // constructed: c
+            // (none) - this is where the glossonym was
+            // (bailey value(s))
         }
         return list;
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
@@ -1115,11 +1115,17 @@ public abstract class XMLSource implements Freezable<XMLSource>, Iterable<String
                         if (list == null) {
                             return new AliasLocation(xpath, localeID);
                         }
-                        list.add(new LocaleInheritanceInfo(localeID, xpath, Reason.value));
+                        if (CldrUtility.INHERITANCE_MARKER.equals(value)) {
+                            list.add(
+                                    new LocaleInheritanceInfo(
+                                            localeID, xpath, Reason.inheritanceMarker));
+                        } else {
+                            list.add(new LocaleInheritanceInfo(localeID, xpath, Reason.value));
+                        }
                         // Now, keep looping to add additional Bailey values.
                         // Note that we will typically exit the recursion (terminal state)
                         // with Reason.codeFallback or Reason.none
-                        if (firstValue != null) {
+                        if (firstValue == null) {
                             // Save this, this will eventually be the function return.
                             firstValue = new AliasLocation(xpath, localeID);
                             // Everything else is only for Bailey.
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
@@ -590,6 +590,7 @@ public abstract class XMLSource implements Freezable<XMLSource>, Iterable<String
                 list.add(
                         new LocaleInheritanceInfo(
                                 locale, path, LocaleInheritanceInfo.Reason.value));
+                // Since we’re not resolving, there’s no way to look for a Bailey value here.
             } else {
                 list.add(
                         new LocaleInheritanceInfo(
@@ -987,7 +988,7 @@ public abstract class XMLSource implements Freezable<XMLSource>, Iterable<String
              * There is no caching problem with skipFirst, since that is always false here -- though
              * getBaileyValue could use a cache if there was one for skipFirst true.
              *
-             * Also skip caching if the list (tracing) is enabled.
+             * Also skip caching if the list is non-null, for tracing.
              */
             if (!skipInheritanceMarker || !cachingIsEnabled || (list != null)) {
                 return getPathLocation(xpath, false /* skipFirst */, skipInheritanceMarker, list);
@@ -1089,6 +1090,11 @@ public abstract class XMLSource implements Freezable<XMLSource>, Iterable<String
                 boolean skipFirst,
                 boolean skipInheritanceMarker,
                 List<LocaleInheritanceInfo> list) {
+
+            //   When calculating the Bailey values, we track the final
+            //   return value as firstValue. If non-null, this will become
+            //   the function's ultimate return value.
+            AliasLocation firstValue = null;
             for (XMLSource source : sources.values()) {
                 if (skipFirst) {
                     skipFirst = false;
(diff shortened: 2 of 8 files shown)


srl295:
test needs update

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
```diff
@@ -590,6 +590,7 @@ public String getSourceLocaleIdExtended(
                 list.add(
                         new LocaleInheritanceInfo(
                                 locale, path, LocaleInheritanceInfo.Reason.value));
+                // Since we’re not resolving, there’s no way to look for a Bailey value here.
```
macchiati, line 593: We should throw an exception if we try for a bailey value with a non-resolving source.

srl295: Good point. and i'll add an exception up in `getPathsWhereFound()`, certainly.  This code, however is two levels deeper, in the XMLSource override.

macchiati: ok. Are you going to update the test as part of this PR?

srl295: Yes, I'll probably work on that Monday.

srl295: Test found some bugs !

srl295:
should pass tests now

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
srl295: throws exception, passes tests

macchiati:
I don't think it is quite working yet. I loaded it up on cldr-staging, and looked at https://cldr-staging.unicode.org/cldr-apps/v#/de/CompoundUnits/3f5664e23b5fc783
What I see is the following:
Screenshot 2023-05-01 at 13 37 31

srl295:
> I don't think it is quite working yet. I loaded it up on cldr-staging, and looked at https://cldr-staging.unicode.org/cldr-apps/v#/de/CompoundUnits/3f5664e23b5fc783
> What I see is the following:
> Screenshot 2023-05-01 at 13 37 31

What should it do?

Some of the Jump buttons don't work because they go to non-existent paths.

macchiati:
Some issues

1. The original path is not shown. That should be at the top (it was, in an earlier iteration). Without that, it is hard to follow.

- //ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="masculine"][@case="accusative"]

2. Some jumps don't work, because there is nothing to jump to. All of the following are really one transaction; picking up an alias in Root, changing the path, then accessing German. Only the **last** [Jump] (for German) has any meaning.

- Root [Jump]
- An alias was found at this location [Jump]
- //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="masculine"][@case="accusative"]
- German [Jump]

3. The first two should be combined (aliases are only in Root)

- Root — An alias changes the path to:
- //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="masculine"][@case="accusative"]

3a. It would be *much* easier to follow if we bolded the difference between the paths. Eg **[@count="one"]**. That can follow later, however.

4. We are not seeing the subsequent changes to the path (before code fallback) (I think) because part of the fallback is done in of the fallback is done in CLDRFile when it should be in XMLFile, namely by
    public String getCountPathWithFallback(String xpath, Count count, boolean winning) {
[mea culpa, by the way]

We should move that logic down into XMLSource eventually. But in the meantime you might be able to instrument that part also.

srl295:
Generally, could we consider merging this when the lower level API is getting the right data, and look at UI improvements separately as a lot of low level code is touched here? (item #1 is now solved)

> Some issues
> 
> 1. The original path is not shown. That should be at the top (it was, in an earlier iteration). Without that, it is hard to follow.

I've restored that in a commit just now.

> * //ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="masculine"][@case="accusative"]
> 
> 2. Some jumps don't work, because there is nothing to jump to. All of the following are really one transaction; picking up an alias in Root, changing the path, then accessing German. Only the **last** [Jump] (for German) has any meaning.
> 
> * Root [Jump]
> * An alias was found at this location [Jump]
> * //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="masculine"][@case="accusative"]
> * German [Jump]

Yes, that’s exactly why I opened #2879, so that the Jump buttons could be made reactive. 

> 3. The first two should be combined (aliases are only in Root)
> 
> * Root — An alias changes the path to:
> * //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="masculine"][@case="accusative"]
> 
> 3a. It would be _much_ easier to follow if we bolded the difference between the paths. Eg **[@count="one"]**. That can follow later, however.

That's a good idea, show the edit difference between them.

Generally speaking, XPaths aren't what we actually show in most of the UI.  We could perhaps show PathHeader instead with an option to show the XPath. 

> 4. We are not seeing the subsequent changes to the path (before code fallback) (I think) because part of the fallback is done in of the fallback is done in CLDRFile when it should be in XMLFile, namely by
>    public String getCountPathWithFallback(String xpath, Count count, boolean winning) {
>    [mea culpa, by the way]
>
> We should move that logic down into XMLSource eventually. But in the meantime you might be able to instrument that part also.

I can instrument where the xpath changes as a result of getFallbackPaths, that yields the following: (note the `| fallback` entry marked with a `»`)

```
• //ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
• 2c01af8246eb4f09
• Value: {0}²
• Inheritance chain:
Path: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"], loc: cs
  • none: cs://ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
  • none: root://ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
  | alias: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
  • none: cs://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
  • none: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
  | removedAttribute: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"]
    attribute=alt
  • none: cs://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"]
  • none: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"]
  | changedAttribute: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"][@gender="feminine"][@case="locative"]
    attribute=count
  • none: cs://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"][@gender="feminine"][@case="locative"]
  • none: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"][@gender="feminine"][@case="locative"]
  • codeFallback: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"][@gender="feminine"][@case="locative"]
» | fallback: cs://ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@gender="feminine"][@case="locative"][@alt="variant"]
  • none: cs://ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@alt="variant"]
  • none: root://ldml/units/unitLength[@type="long"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@alt="variant"]
  | alias: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@alt="variant"]
  • none: cs://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@alt="variant"]
  • none: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"][@alt="variant"]
  | removedAttribute: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"]
    attribute=alt
  • value: cs://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"]
    /Users/srl295/src/cldr/common/main/cs.xml:16807:66:  XML Source
  • none: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="one"]
  | changedAttribute: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"]
    attribute=count
  • value: cs://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"]
    /Users/srl295/src/cldr/common/main/cs.xml:16810:68:  XML Source
  • value: root://ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"]
    /Users/srl295/src/cldr/common/main/root.xml:4830:68:  XML Source
  • codeFallback: //ldml/units/unitLength[@type="short"]/compoundUnit[@type="power2"]/compoundUnitPattern1[@count="other"]
```

srl295:
> > Generally speaking, XPaths aren't what we actually show in most of the UI. We could perhaps show PathHeader instead with an option to show the XPath.
> 
> Yes, that should be the end goal for non-power users. We could have a hover with the path, that would be useful for us.
> 
> > I can instrument where the xpath changes as a result of getFallbackPaths, that yields the following: (note the | fallback entry marked with a...
> > Excellent! That shows us what is there — _**and**_ what we need to fix!

OK, I've committed this. I just call it `fallback` but it could probably have a better name.

macchiati:
I put together a chart showing what is happening (there might be a missing step in your output), and what I think ought to happen (PROPOSED)
https://docs.google.com/spreadsheets/d/1ho6La6QlfqpSdC9kNBaWvKmZYRA_9QUXaDQXEggnZl4

srl295:
> I put together a chart showing what is happening (there might be a missing step in your output), and what I think ought to happen (PROPOSED)
> https://docs.google.com/spreadsheets/d/1ho6La6QlfqpSdC9kNBaWvKmZYRA_9QUXaDQXEggnZl4

OK. it's not tracing through getFallbackPaths() just noting the final output.

macchiati:
That's ok for it not to trace through getFallbackPaths: I think we know what is happening there. What do you think about the PROPOSED? (Will of course have to run it by others, and it shouldn't block this PR).

srl295:
> That's ok for it not to trace through getFallbackPaths: I think we know what is happening there. What do you think about the PROPOSED? (Will of course have to run it by others, and it shouldn't block this PR).

It makes some sense, I would have to look at the data to see what would be impacted.   Dropping the alt at first is definitely a big change in some sense.

btangmu (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
```diff
@@ -1089,6 +1090,11 @@ private AliasLocation getPathLocation(
                 boolean skipFirst,
                 boolean skipInheritanceMarker,
                 List<LocaleInheritanceInfo> list) {
+
+            //   When calculating the Bailey values, we track the final
+            //   return value as firstValue. If non-null, this will become
+            //   the function's ultimate return value.
```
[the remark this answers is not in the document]
srl295: I'm not sure, looks like spotless did it

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -554,32 +554,47 @@ public String getBaileyValue(
...
+                    "getPathsWhereFound() is only valid on a resolved CLDRFile");
+        }
+        LinkedList<LocaleInheritanceInfo> list = new LinkedList<>();
         // first, call getSourceLocaleIdExtended to populate the list
         Status status = new Status();
-        getSourceLocaleIdExtended(xpath, status, true, list);
+        final String locale1 = // for debugging
+                getSourceLocaleIdExtended(xpath, status, false, list);
         final String path1 = status.pathWhereFound;
+        System.out.println("Path: " + path1 + ", loc: " + locale1);
```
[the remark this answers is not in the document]
srl295: I think I've removed that now, thanks

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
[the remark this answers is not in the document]
srl295: when in Rome (or when in XMLSource) … see the lines just prior:
image

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -554,32 +554,47 @@ public String getBaileyValue(
...
+     * is used to explain inheritance and bailey values. The list must be interpreted in order. When
+     * {@link LocaleInheritanceInfo.Reason#isTerminal()} return true, that indicates a successful
+     * lookup and partitions values from subsequent bailey values.
      *
-     * @return map from xpath to locale of one or more contributing paths in the case of glossonymns
-     *     or other constructed values.
+     * @see #getBaileyValue(String, Output, Output)
+     * @see #getSourceLocaleIdExtended(String, Status, boolean)
      */
     public List<LocaleInheritanceInfo> getPathsWhereFound(String xpath) {
```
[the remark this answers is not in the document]
srl295: Correct, however, `ResolvingSource.getBaileyValue()` actually just calls `ResolvingSource.getPathLocation(…,  /*skipFirst*/ true)` to skip the first value.  That same `getPathLocation()` is the one that does the heavy lifting to build up the list.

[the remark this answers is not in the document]
srl295: getFallbackPath() should be called from XMLSource - that's mentioned in https://unicode-org.atlassian.net/browse/CLDR-13266 - when that is so and the glossonym stuff cleaned up, CLDRFile.getBaileyValue() can just call XMLSource.getBaileyValue() which will be the same code path.

We could also add tests to show comparable results. `getPathsWhereFound()` does a lot more work that `getBaileyValue()` might not want or need.

macchiati:
SGTM

macchiati (approved):

srl295:
@btangmu can we approve and do testing in a follow on?

srl295:
Api doesn't lend itself to cross testing versus Bailey especially for constructed.

btangmu (approved):
a follow-up is needed to ensure getPathsWhereFound and getBaileyValue actually implement the same algorithm and always find the same localeWhereFound and pathWhereFound; one way would be for getBaileyValue to be a short wrapper around getPathsWhereFound

srl295:
> "can we approve and do testing in a follow on?" -- yes, though more than just testing is needed -- it's problematic to have two separate functions getPathsWhereFound and getBaileyValue that supposedly implement the same crucial algorithm. If there's a performance problem with the extra explanatory work done by getPathsWhereFound, that extra work could be contingent on a boolean or something. Sooner or later -- probably sooner, since there are progress-blocking inheritance bugs -- the algorithm is going to change, and it's going to be a burden to keep the two functions consistent

Actually, getPathsWhereFound does not add any new lookup code.  It adds tracing to the existing implementation of getSourceLocaleIdExtended, the same implementation that is called by getBaileyValue. 

> btw why no "This PR completes the ticket" checkbox for this PR? I'm approving the PR with the understanding that if there were a checkbox it wouldn't be checked

It must have been accidentally removed. I've re-added it, unchecked. Thanks for the catch.

btangmu:
"Actually, getPathsWhereFound does not add any new lookup code. It adds tracing to the existing implementation of getSourceLocaleIdExtended, the same implementation that is called by getBaileyValue."

OK, still, what needs to be explained is getBaileyValue, which is 37 lines long and doesn't actually call getSourceLocaleIdExtended directly.

What we need is a "getBaileyValue explainer", not a "getSourceLocaleIdExtended explainer" :-)

PR #2891, opened 2023-05-03
CLDR-15034 kbd: clarify identifier restrictions per DTD

srl295:
CLDR-15034

- also, re-add `display id=` to the DTD which was in-spec but missing from DTD

FYI @eggrobin — this makes the spec match the DTD

ALLOW_MANY_COMMITS=true

DISABLE_JIRA_ISSUE_MATCH=true


srl295:
merging to make progress

PR #2892, opened 2023-05-03
CLDR-14050 annotation fix mxclaus

srl295:
CLDR-14050

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/annotations/en.xml
+++ b/common/annotations/en.xml
@@ -1437,7 +1437,7 @@ annotations.
 		<annotation cp="🎅" type="tts">Santa Claus</annotation>
 		<annotation cp="🤶">celebration | Christmas | claus | mother | Mrs. | Mrs. Claus</annotation>
 		<annotation cp="🤶" type="tts">Mrs. Claus</annotation>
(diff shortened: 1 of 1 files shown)


common/annotations/en.xml
```diff
@@ -1437,7 +1437,7 @@ annotations.
 		<annotation cp="🎅" type="tts">Santa Claus</annotation>
 		<annotation cp="🤶">celebration | Christmas | claus | mother | Mrs. | Mrs. Claus</annotation>
 		<annotation cp="🤶" type="tts">Mrs. Claus</annotation>
-		<annotation cp="🧑‍🎄">Claus, christmas | mx claus</annotation> <!-- 1F9D1 200D 1F384 -->
+		<annotation cp="🧑‍🎄">claus | christmas | Mx. Claus</annotation> <!-- 1F9D1 200D 1F384 -->
```
pedberg-icu, line 1440: Shouldn't it be mx claus without period to match the tts form? (Otherwise doesn't the tts form get automatically added by CLDRModify?)

srl295: makes sense. fixed

pedberg-icu (approved):

PR #2898, opened 2023-05-04
CLDR-13180 st: use DAIP for English, Root, Others

srl295:
- add rawEnglish field to DataPage for what was formerly the displayName
- displayName in the voting data is now DAIP processed
- remove DAIP normalization from STFactory, it should be done by callers, and not done on DB values. CLDR-16105
- PathInfo CLI tooling to test DAIP

Still some rough edges:

1. ST doesn't always recognize the display value as the one you were voting for, so vote 'jumps' unnecessarily.
2. raw UnicodeSet in the info panel. 

CLDR-13180

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.js
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.js
@@ -513,7 +513,7 @@ function updateRowVoteInfo(tr, theRow) {
        * theRow.inheritedValue can be undefined here; then do not append
        */
       if (theRow.inheritedValue) {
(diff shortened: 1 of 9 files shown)


macchiati (approved):
Not familiar with the cldr-apps side, but it looks good. And the rest looks good too.

srl295:
@btangmu this fixes some definitely regressions, and STFactory tests pass.  Going to merge it  but please review if there are any issues.

PR #2900, opened 2023-05-04
CLDR-16623 V2 make specified changes to the PNF spec

macchiati:
CLDR-16623

Includes both the changes for the PNF spec, and the finalization for 43.1:

- change the version to 43.1 in all files
- change the date to today
- change from proposed draft to final

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):
Looks good, thanks for retargeting!

PR #2901, opened 2023-05-05
CLDR-16636 removed

srl295:
removed for now


srl295:
Exemplars from Wikipedia and pending check.

PR #2905, opened 2023-05-05
CLDR-16627 remove FFFF and FEFF characters

DavidLRowe:
CLDR-16627

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ab.xml
+++ b/common/main/ab.xml
@@ -919,7 +919,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<parseLenient sample="’" draft="unconfirmed">['ʼ՚᾽᾿’＇]</parseLenient>
 			<parseLenient sample="%" draft="unconfirmed">[%٪﹪％]</parseLenient>
 			<parseLenient sample="‰" draft="unconfirmed">[؉‰]</parseLenient>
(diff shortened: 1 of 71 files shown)


DavidLRowe:
For 71 files (the 70 files in the CLDR ticket plus oc_ES.xml), a "$" is removed from the end of a `parseLenient` element. (This is a reserved code for FFFF and should not be in this element.)

In addition four occurrences of FEFF (zero width no break space / byte order mark) were removed from the ends of four strings in am.xml.

srl295:
> For 71 files (the 70 files in the CLDR ticket plus oc_ES.xml), a "$" is removed from the end of a `parseLenient` element. (This is a reserved code for FFFF and should not be in this element.)
> 
> In addition four occurrences of FEFF (zero width no break space / byte order mark) were removed from the ends of four strings in am.xml.

do you think they were meant to be `$` or perhaps `\$` (escaped dollarsign)?

macchiati (approved):
Looks good, thanks!

macchiati:
I think they were meant to be $, but people didn't realize that needed
escaping. I think in each case there was an literal $ in the set anyway.

common/main/ab.xml
limck613, line 1: Understand Cc

limck613 (approved):

limck613 (approved):

PR #2907, opened 2023-05-08
CLDR-4903 Correct ordering etc

macchiati:
CLDR-4903
This is to handle the 3 changes listed in https://unicode-org.atlassian.net/browse/CLDR-4903?focusedCommentId=169687

Some more notes:

This clearly separates the three different string forms of UnicodeSet
- the plain form coming from calling unicodeSet.toString() or .toPattern(), and produced by new UnicodeSet()
- the prettier form used internally to XML. This is valid UnicodeSet, but is formatted to group the elements differently, eg in sorting order according to the locale. This is handled by UnicodeSetPrettyPrinter, with parse and format methods. (The parse method just calls new UnicodeSet, since the prettier form is still valid UnicodeSet input.)
- a new display form for vetters, that simplifies the syntax to help prevent them from making mistakes, handled by SimpleUnicodeSetFormatter with parse and format methods.

SimpleUnicodeSetFormatter.parseLenient(value) is a new method that tries parsing the value in two ways:
1. If the format […], then with new UnicodeSet()
2. Otherwise with SimpleUnicodeSetFormatter.parse()

Example:
DAIP.processInput() uses SimpleUnicodeSetFormatter.parseLenient(value) to parse the value coming from the vetter, and UnicodeSetPrettyPrinter.format() to then put it in the right format for internal use. This might throw an exception if the content is not valid according to UnicodeSet, _and_ not valid according  to SimpleUnicodeSetFormatter. The exceptions have messages that are suitable for showing to vetters.

DAIP.formatForDisplay() displays the value using SimpleUnicodeSetFormatter, with a simpler format.

Various places in the code needed to change to use SimpleUnicodeSetFormatter.parseLenient

------
This was #2906; I had to change that PR because the branch name was illegal (had a comma in it). 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
@@ -9,11 +9,13 @@ import com.ibm.icu.text.UnicodeSet;
 import com.ibm.icu.text.UnicodeSetIterator;
 import com.ibm.icu.util.ULocale;
 import java.util.BitSet;
+import java.util.Comparator;
 import java.util.List;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.Factory;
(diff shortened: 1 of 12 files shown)


macchiati:
Note, the naming is not quite consistent, but I'd rather deal with that in a followup PR.

PR #2912, opened 2023-05-10
CLDR-16457 nativeSpaceReplacement and so on

macchiati:
CLDR-16457

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -3161,7 +3161,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST featureName draft (approved | contributed | provisional | unconfirmed) #IMPLIED >
     <!--@METADATA-->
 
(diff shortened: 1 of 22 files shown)


richgillam (approved):
I guess I'm okay with all of this.  A few questions:
- Did we actually agree on all of this in the Person Names group meeting?  I don't remember the discussion.  Was it while I was out?
- I don't see anything in here dealing with the proposed `spaceLocales` or `useFormatter` stuff listed in the ticket.  Are those coming in a future PR?
- There'll need to be changes in the spec for all this stuff.  Is that coming in another PR?
- We'll need an ICU ticket to make the necessary changes in ICU.

macchiati:
We triaged this in the PNF group; this is just the first of the changes. I'll float some ideas for the rest to the PNF.
And the doc changes, and vetter guide changes will need to follow, and the ICU update ticket.

PR #2914, opened 2023-05-10
CLDR-16643 Fix kh, add more comprehensive tests

macchiati:
CLDR-16643

Most of the work was in fleshing out the test to try to prevent breakage before it hits ICU. Didn't have time to port the ICU tests, so just did a quick test.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/transforms/Cyrillic-Latin.xml
+++ b/common/transforms/Cyrillic-Latin.xml
@@ -33,7 +33,12 @@ $descender = ˌ;
 $umlaut = \u0308 ;
 $macron = \u0304 ;
 $breveBelow = \u032E;
-# move up so not masked
+
+$ignoreForCase = [‧[:Mn:][:Me:]] ;
+$lower = [[:Latin:][:Cyrillic:] &amp; [:Ll:]];
+$beforeLower = $ignoreForCase * $lower ;
+
+# Move up so not masked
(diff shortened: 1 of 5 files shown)


pedberg-icu (approved):
Nice to have more transform testing on the CLDR side!

pedberg-icu:
Hmm. The ticket is for 43.1 but this PR is against the main branch. Guess I will cherry-pick it to maint/maint-43 branch (easier with this commit than with the spec one) when I cherry pick the relevant maint/maint-43 fixes in the other direction, to main.

srl295 (approved):

common/transforms/Cyrillic-Latin.xml
```diff
@@ -33,7 +33,12 @@ $descender = ˌ;
 $umlaut = \u0308 ;
 $macron = \u0304 ;
 $breveBelow = \u032E;
```
markusicu, line 35: Is this unused now?

markusicu (commented):
thanks! one question...

common/transforms/Cyrillic-Latin.xml
macchiati: Just checked, and it isn't used. (I don't think this is worth a new PR; we haven't been concerned about details like that in the files.

PR #2921, opened 2023-05-12
CLDR-16503 test for consistent English units

srl295:
CLDR-16503

Currently fails, 

> Inconsistent units, missing from en: volume-dessert-spoon-imperial(short) volume-quart-imperial(short) volume-dessert-spoon(short) volume-pinch(short) light-candela(short) mass-grain(short) light-lumen(short) volume-drop(short) volume-dram(short) temperature-generic(narrow, short) volume-jigger(short) length-earth-radius(short) graphics-dot(short) 

ALLOW_MANY_COMMITS=true


macchiati:
We haven't guaranteed that English has values for all root values, so I think that might be what you are seeing. But that is a broader issue, not just units.

macchiati:
Also there is already a tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUnits.java (different directory). We shouldn't add duplicate names even in different directories.

srl295:
> Also there is already a tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUnits.java (different directory). We shouldn't add duplicate names even in different directories.

it's the non-JUnit version. 
- [ ] I can disambiguate the name.

macchiati:
Disambiguating the name sounds good. But should they be in the same
directory, or is that because of our historical structure didn't match?

srl295:
this ticket is closing, so if desired this should move to a new ticket?

PR #2923, opened 2023-05-12
CLDR-15034 super basic keyboard charts

srl295:
CLDR-15034

image

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -32,6 +32,8 @@ jobs:
       - name: Run TR archiver
         # Note: will update ToC if out of date
         run: 'cd tools/scripts/tr-archive/ && npm ci && npm run build'
+      - name: Run Kbd Charts
+        run: 'cd docs/charts/keyboard && npm ci && npm run build'
       - name: 'Run TR link extractor (experimental)'
         # Note: see CLDR-16526 for making this an error
         run: 'cd tools/scripts/tr-archive/ && npm run extract-link-targets || (echo Warning, please fix these ; true)'
--- a/.github/workflows/keyboard.yml
+++ b/.github/workflows/keyboard.yml
@@ -86,3 +86,5 @@ jobs:
           -DCLDR_GITHUB_ANNOTATIONS=true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run Kbd Charts
+        run: 'cd docs/charts/keyboard && npm ci && npm run build'
--- a/docs/charts/keyboard/.gitignore
+++ b/docs/charts/keyboard/.gitignore
@@ -1,0 +1,2 @@
+/node_modules
+/static/data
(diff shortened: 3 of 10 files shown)


srl295:
i posted it up live https://www.unicode.org/~srloomis/tmp/kbd-chart/

miloush:
Maybe use different colors for gap?

I think we should also have a SGCaps keyboard amongst simple examples.

DavidLRowe:
For the AZERTY keyboard, shift layer, the key to the left of the "1" is empty. On the chart the key disappears and the remaining keys on the top row are shifted left. (But the Tab and CapsLock keys aren't there either, so maybe this isn't relevant to this stage in the charts.)

srl295:
> Maybe use different colors for gap?

good idea. 

> I think we should also have a SGCaps keyboard amongst simple examples.

what is SGCaps?

miloush:
Keyboard like Czech or [S]wiss [G]erman for which Shift keys are not the same as Caps keys: https://kbdlayout.info/features/SGCAPS. But maybe it's not that important, it's like 5% of keyboard in Windows that use it, it just wasn't obvious from the examples how this is solved.

srl295:
> For the AZERTY keyboard, shift layer, the key to the left of the "1" is empty. On the chart the key disappears and the remaining keys on the top row are shifted left. (But the Tab and CapsLock keys aren't there either, so maybe this isn't relevant to this stage in the charts.)

Good catch.

Actually, there's no key there in the layout file. But looking at some screenshots, I think it's a bug and there should be a gap there… or perhaps the superscript-2 is repeated in the unshifted layer.

Anyway the chart is accurate, in fact helpful in spotting the issue!

srl295:
> Keyboards like Czech or [S]wiss [G]erman for which Shift keys are not the same as Caps keys: https://kbdlayout.info/features/SGCAPS. But maybe it's not that important, it's like 5% of keyboard in Windows that use it, it just wasn't obvious from the examples how this is solved.

got it… filed https://unicode-org.atlassian.net/browse/CLDR-16653

DavidLRowe:
On a standard AZERTY layout, the key above the Tab key produces a superscript 2 on the unshifted layer and nothing on the shifted layer. Do you have a way to represent a key that does nothing?!

srl295:
> 

yes, it would be a gap.

srl295:
Charts are now auto generated and pushed… https://cldr-smoke.unicode.org/spec/keyboard-preview/charts/keyboard/static/

Yes, i'm fixing the encoding issue, so embarassing 🙈

macchiati:
Best to have the cells have a constant width, with the actual character in
them. The name can be in a mouse-hover. A bit larger font also. Agree about
different color for the gap.

srl295:
> Best to have the cells have a constant width, with the actual character in them. The name can be in a mouse-hover. A bit larger font also. Agree about different color for the gap.
> […](#)

All done except for the font size and width. I'll capture that.

mcdurdin:
> for which Shift keys are not the same as Caps keys

I think this is a special case of a more general issue -- Shift is not the same as Caps for English either, e.g. punctuation and digits. I think that separate Caps and Shift layers solves both of these.

PR #2925, opened 2023-05-13
CLDR-16194 add locale ie Interlingue

DavidLRowe:
CLDR-16194

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv
```diff
@@ -1448,6 +1448,7 @@ United States	US	"329,256,465"	99%	"19,490,000,000,000"		Lojban	jbo	"10"
 United States	US	"329,256,465"	99%	"19,490,000,000,000"		Ido	io	"100"
 Unknown Region	ZZ	0	0%	0		Ido	io	0	99%
 Unknown Region	ZZ	0	0%	0		Interlingua	ia	0	99%
+Unknown Region	ZZ	0	0%	0		Interlingue	ie	0	99%
```
srl295, line 1451: Thanks! Were the other files generated from this?

srl295 (approved):

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv
DavidLRowe: Other changes were based on this, but entered by hand rather than being generated.

PR #2926, opened 2023-05-15
CLDR-16654 Add tool and code for downgrading <locale, path, value> matches

macchiati:
CLDR-16654

- add a new option, -fD for downgrade

tools/cldr-code/src/main/java/org/unicode/cldr/util/DowngradePaths.java
- simple API for calling from CLDRModify, and later while importing votes

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
- regex matcher for <locale, path, value>
- doesn't need to be highly optimized, because it isn't on critical paths. 

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
- a data file read by DowngradePaths.java

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathLookup.java
- adds basic unit tests for LocalePathValueListMatcher and Downgrade

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
@@ -57,6 +57,7 @@ import org.unicode.cldr.util.CLDRTool;
 import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.DateTimeCanonicalizer;
 import org.unicode.cldr.util.DateTimeCanonicalizer.DateTimePatternType;
+import org.unicode.cldr.util.DowngradePaths;
 import org.unicode.cldr.util.DtdData;
 import org.unicode.cldr.util.DtdType;
 import org.unicode.cldr.util.Factory;
@@ -2945,6 +2946,32 @@ public class CLDRModify {
                                 handlePath(path);
                             }
                         }
+                    }
+                });
+
+        fixList.add(
+                'D',
+                "Downgrade paths",
+                new CLDRFilter() {
+                    @Override
+                    public void handlePath(String xpath) {
+                        String value = cldrFileToFilter.getStringValue(xpath);
+                        if (!DowngradePaths.lookingAt(getLocaleID(), xpath, value)) {
+                            return;
+                        }
+                        String fullPath = cldrFileToFilter.getFullXPath(xpath);
+                        XPathParts fullParts = XPathParts.getFrozenInstance(fullPath);
+                        String oldDraft = fullParts.getAttributeValue(-1, "draft");
+                        if (oldDraft != null) {
+                            DraftStatus oldDraftEnum = DraftStatus.forString(oldDraft);
+                            if (oldDraftEnum == DraftStatus.provisional
+                                    || oldDraftEnum == DraftStatus.unconfirmed) {
+                                return;
+                            }
+                        }
+                        fullParts = fullParts.cloneAsThawed();
+                        fullParts.setAttribute(-1, "draft", "provisional");
+                        replace(fullPath, fullParts.toString(), value, "Downgrade to provisional");
                     }
                 });
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
@@ -1,0 +1,89 @@
+package org.unicode.cldr.util;
+
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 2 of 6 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2947,6 +2948,32 @@ public void handleEnd() {
...
+                    public void handlePath(String xpath) {
+                        String value = cldrFileToFilter.getStringValue(xpath);
+                        if (!DowngradePaths.lookingAt(getLocaleID(), xpath, value)) {
+                            return;
+                        }
+                        String fullPath = cldrFileToFilter.getFullXPath(xpath);
+                        XPathParts fullParts = XPathParts.getFrozenInstance(fullPath);
+                        String oldDraft = fullParts.getAttributeValue(-1, "draft");
+                        if (oldDraft != null) {
+                            DraftStatus oldDraftEnum = DraftStatus.forString(oldDraft);
```
srl295, line 2966:
```suggestion
                        DraftStatus oldDraftEnum = DraftStatus.forXpath(fullPath);
                        if (oldDraftEnum != null) {
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
```diff
@@ -0,0 +1,89 @@
...
+
+        public LocalePathValueMatcher(List<String> parts) {
+            localePattern =
+                    parts.size() < 1 || parts.get(0).isEmpty()
+                            ? null
+                            : Pattern.compile(parts.get(0));
+            pathPattern =
+                    parts.size() < 2 || parts.get(1).isEmpty()
+                            ? null
+                            : Pattern.compile(parts.get(1).replace("[@", "\\[@"));
```
srl295, line 31: Ideally this should go in a more generic utility class. Or, alternately, `XPathParts x.toPatternString()` where it could handle any other escaping needed

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
```diff
@@ -0,0 +1,11 @@
+# Format is 
+#    locale-regex ; path-regex ; value-regex
+# Where a field is missing, it matches all values
```
srl295, line 3: some comments, not blocking

- `.*` might be less err prone than missing

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
```diff
@@ -0,0 +1,11 @@
...
+# The match uses 'looking at', equivalent to starting with ^ — so don't start with that
+# If you need to match to the end, follow with $
+# For the path, [@ is automatically escaped
+
+# The following will match anything starting with graphics-dot, thus include graphics-dot-per-centimeter, etc.
+ ; //ldml/units/unitLength[@type="long"]/unit[@type="graphics-dot
```
srl295, line 9:
```suggestion
.* ; ^//ldml/units/unitLength[@type="long"]/unit[@type="graphics-dot.*$
```

in fact, the `^` and `.*$` and end may be best implied unless the pattern ends with a $.

Might be best encapsulated in an XPathMatcher  class

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
```diff
@@ -0,0 +1,11 @@
+# Format is
```
srl295, line 1:
```suggestion
# Read by tools/cldr-code/src/main/java/org/unicode/cldr/util/DowngradePaths.java
# Format is 
```

srl295 (approved):
lots of comments but nothing blocking

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
macchiati: I thought of this, but I think most of the time it is not going to be draft, so because we'll need the fullParts anyway below it...

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
macchiati: Agreed.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
macchiati: I was debating whether to make it be a match or not. We already do this processing in RegexLookup, so makes sense to centralize

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
macchiati: Good idea.

PR #2927, opened 2023-05-15
CLDR-13180 st: use DAIP view in Info panel

srl295:
CLDR-13180

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -521,7 +521,7 @@ function updateRowVoteInfo(tr, theRow) {
     } else {
       cldrVote.appendItem(
         valdiv,
-        value,
+        item.value, // get display value not raw
         value === theRow.winningValue ? "winner" : "value"
       );
     }


macchiati (dismissed):

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -790,7 +790,7 @@ function showItemInfoFn(theRow, item) {
       displayValue = theRow.inheritedDisplayValue;
     }
 
-    cldrVote.appendItem(h3, displayValue, item.pClass);
+    cldrVote.appendItem(h3, displayValue, item.pClass); // @@
```
btangmu, line 793: What's @@ here?

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -521,7 +521,7 @@ function updateRowVoteInfo(tr, theRow) {
     } else {
       cldrVote.appendItem(
         valdiv,
-        value,
+        item.value, // get display value not raw
```
btangmu, line 524: is appendItem consistently called with display value elsewhere?

btw I happened to notice just now, there's a line in cldrVote.mjs (not due to this PR) that has no effect:

`var span = appendItem(h3, value, "value");`

srl295: > is appendItem consistently called with display value elsewhere?

Everywhere except i'm not sure about cldrOldVotes.mjs, but I haven't checked that API.

> btw I happened to notice just now, there's a line in cldrVote.mjs (not due to this PR) that has no effect:
> 
> `var span = appendItem(h3, value, "value");`

The `var span = ` could be removed, but it _does_ have an effect. The span will be appended to `h3` even though the return type is not captured.

btangmu: my impression is the answer is "yes" -- appendItem is consistently called with display value elsewhere

this instance was exceptional since here value is from vr.value_vote

srl295: @btangmu ok to merge then?

btangmu: about `var span = ...` we should delete `var span` (but keep` appendItem...`) next time we touch that file

btangmu (approved):

PR #2929, opened 2023-05-15
CLDR-16604 kbd: scancodeLayouts.xml

srl295:
CLDR-16604

Draft of scancode layout format

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1203,7 +1203,7 @@ This element represents a specific `form` element which defines the layout of a 
 
 > *Note:* Most keyboards will not need to use this element directly, and the CLDR repository will not accept keyboards which define a custom `form` element.  This element is provided for two reasons:
 
-1. To document the standard hardware layouts used with CLDR for implementations. Implementations can verify the layout, and validate keyboards against the number of items per row.
+1. To formally specify the standard hardware arrangements used with CLDR for implementations. Implementations can verify the arrangement, and validate keyboards against the number of rows and the number of keys per row.
 
 2. To allow a way to customize the scancode layout for keyboards not intended to be included in the common CLDR repository.
 
@@ -1246,9 +1246,9 @@ There is an implied set of `<form>` elements corresponding to the default forms,
 
 Here is a summary of the implied form elements. Keyboards included in the CLDR Repository must only use these `form=` values and may not override the scanCodes.
 
-> - `touch` - Touch (non-hardware) layout. 
-> - `abnt2` - Brazilian 103 key ABNT2 layout (iso + extra key left of right shift)
-> - `iso` - European 102 key layout (extra key right of left shift)
+> - `touch` - Touch (non-hardware) layout.
+> - `abnt2` - Brazilian 103 key ABNT2 layout (iso + extra key near right shift)
+> - `iso` - European 102 key layout (extra key near left shift)
 > - `jis` - Japanese 109 key layout
 > - `us` - ANSI 101 key layout
 
@@ -1414,7 +1414,7 @@ _Attribute:_ `keys` (required)
 
 > This is a string that lists the id of [`key` elements](#Element_key) for each of the keys in a row, whether those are explicitly listed in the file or are implied.  See the `key` documentation for more detail.
 >
-> For non-`touch` forms, the number of keys in each row may not exceed the number of scan codes defined for that row. See [`scanCodes`](#Element_scanCodes);
+> For non-`touch` forms, the number of keys in each row may not exceed the number of scan codes defined for that row, and the number of rows may not exceed the defined number of rows for that form. See [`scanCodes`](#Element_scanCodes);
 
 **Example**
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -85,6 +85,10 @@ The LDML specification is divided into the following parts:
   * [Element: displays](#Element_displays)
   * [Element: display](#Element_display)
   * [Element: displayOptions](#Element_displayOptions)
+  * [Element: forms](#Element_forms)
+  * [Element: form](#Element_form)
+    * [Implied Form Values](#implied-form-values)
+  * [Element: scanCodes](#Element_scanCodes)
   * [Element: layers](#Element_layers)
   * [Element: layer](#Element_layer)
   * [Element: row](#Element_row)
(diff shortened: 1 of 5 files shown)


miloush:
Can we refer to this from the spec text? So that it gets visibility for PRI

srl295:
I can make CLDR-16604 a sub task of the pri feedback. Then it will show up
when they click feedback.

srl295:
As well the pri page can get an update with a list of known issues etc.

keyboards/supplemental/README.md
```diff
@@ -0,0 +1,3 @@
+All of the files in this directory can be thought of as merged together.
```
mcdurdin, line 1: I'm not sure that this makes much sense? It would be helpful to explain here why we have scan code files, and what the differences are between the various files.

Also would note that the scan code definitions exclude 'frame' keys.

keyboards/supplemental/form-us.xml
```diff
@@ -0,0 +1,15 @@
...
+    <!-- A file could have multiple 'form' definition, but by convention, each form
+        is in its own file. -->
+    <form id="us"> <!-- Describe the form here: One line enter, etc. -->
+        <scanCodes>
+            <!-- One element per row. Space separated hex bytes. -->
+            <row codes="29 02 03 04 05 06 07 08 09 0A 0B 0C 0D" /> <!-- 13 keys -->
+            <row codes="10 11 12 13 14 15 16 17 18 19 1A 1B 2B" /> <!-- 13 keys -->
+            <row codes="1E 1F 20 21 22 23 24 25 26 27 28" />  <!-- 11 keys -->
+            <row codes="2C 2D 2E 2F 30 31 32 33 34 35" /> <!-- 10 keys -->
+            <row codes="39" /> <!-- 1 key -->
```
mcdurdin, line 12:
```suggestion
            <row codes="29 02 03 04 05 06 07 08 09 0A 0B 0C 0D" /> <!-- 13 keys, ` to = -->
            <row codes="10 11 12 13 14 15 16 17 18 19 1A 1B 2B" /> <!-- 13 keys, Q to \ -->
            <row codes="1E 1F 20 21 22 23 24 25 26 27 28" />  <!-- 11 keys, A to ' -->
            <row codes="2C 2D 2E 2F 30 31 32 33 34 35" /> <!-- 10 keys, Z to / -->
            <row codes="39" /> <!-- 1 key, spacebar -->
```

srl295:
@miloush I neglected to refer to this in the spec in time for PRI (now closed of course). Does this structure substantially address your concerns?

miloush:
I would rather if this was importable, otherwise we basically created another platform file, not allowing to have a self-contained single file. However, if the question is whether this formalizes the physical arrangement sufficiently in a specification, then provided this is referred to in the specification I believe the answer is yes.

srl295:
> I would rather if this was importable, otherwise we basically created another platform file, not allowing to have a self-contained single file. However, if the question is whether this formalizes the physical arrangement sufficiently in a specification, then provided this is referred to in the specification I believe the answer is yes.

How about something where there's an implied import of standard layouts (iso/us/abnt2/jis)? 

That might work… i'll sketch it up.

miloush:
Sure implied import works for me.

srl295:
- [ ] Still need to document in the spec.  Trying to see if this structure is OK or not.

with this structure the implied import would look like this…

```xml
<keyboard>
  … 
 <layers>
    …
    <forms>
        <import base="cldr" path="techpreview/forms-Zzzz-implied.xml" />
    </forms>
 </layers>
```

but you could override with

```xml
<keyboard>
  … 
 <layers>
    <layer form="jan"> <!-- could be 'us' otherwise -->
        …
    </layer>
    <forms>
        <form id="jan">
           <scanCodes …/>
        </form>
    </forms>
 </layers>
```

srl295:
@miloush does this XML look OK? If so i'll rebase and document.

srl295:
> It is better than nothing, although it would make sense the forms to precede layers. Since this is a lookup, wouldn't it be easier if forms were all in their own element? I.e. what benefit do you get by having `forms` inside `layers` rather than inside `keyboard`?

probably no benefit… makes sense, I will move forms out to it's own top level item preceding layers. 

> I wanted to bring up the same issue with flicks - since we are not trying to maintain compatibility with the previous version and it is basically a dictionary with its ids, why are we mixing flicks and keys in one element, and presumably having them share a set of ids?

can you file a ticket on flicks?

srl295:
@miloush here is the current proposed structure:
```xml
<keyboard>
  … 
    <forms>
        <!-- this is an implied import - so most keyboards wouldn't need this -->
        <import base="cldr" path="techpreview/forms-Zzzz-implied.xml" />
    </forms>
 <layers>
    …
 </layers>
```

but you could override with

```xml
<keyboard>
  … 
    <forms>
        <form id="jan"> <!-- or us -->
           <scanCodes …/>
        </form>
    </forms>
 <layers>
    <layer form="jan"> <!-- could be 'us' otherwise -->
        …
    </layer>
 </layers>
```

srl295:
rebased on main

srl295:
@miloush  does the XML look OK?

miloush (approved):

docs/ldml/tr35-keyboards.md
```diff
@@ -1163,6 +1167,119 @@ This attribute may be escaped with `\u` notation, see [Escaping](#Escaping).
...
+        <!-- ... -->
+    </form>
+</forms>
+```
+
+* * *
+
+### <a name="Element_form" href="#Element_form">Element: form</a>
+
+This element represents a specific `form` element which defines the layout of a particular hardware form.
```
miloush, line 1202: The word "layout" is a bit loaded, maybe we could use "arrangment" for the physical layout of keys?

docs/ldml/tr35-keyboards.md
```diff
@@ -1163,6 +1167,119 @@ This attribute may be escaped with `\u` notation, see [Escaping](#Escaping).
...
+    <forms>
+        <import base="cldr" path="techpreview/scanCodes-implied.xml" /> <!-- the version will match the current conformsTo of the file -->
+    </forms>
+</keyboard>
+```
+
+Here is a summary of the implied form elements. Keyboards included in the CLDR Repository must only use these `form=` values and may not override the scanCodes.
+
+> - `touch` - Touch (non-hardware) layout. 
+> - `abnt2` - Brazilian 103 key ABNT2 layout (iso + extra key left of right shift)
```
miloush, line 1250: "iso" is not introduced yet, consider rearranging these lines

docs/ldml/tr35-keyboards.md
```diff
@@ -1163,6 +1167,119 @@ This attribute may be escaped with `\u` notation, see [Escaping](#Escaping).
...
+
+```xml
+<keyboard>
+    <forms>
+        <import base="cldr" path="techpreview/scanCodes-implied.xml" /> <!-- the version will match the current conformsTo of the file -->
+    </forms>
+</keyboard>
+```
+
+Here is a summary of the implied form elements. Keyboards included in the CLDR Repository must only use these `form=` values and may not override the scanCodes.
```
miloush, line 1247: Do you want to mention the actual standards here? (ANSI INCITS 154-1998 (R1999), ISO/IEC 9995-2:2009 etc.)

docs/ldml/tr35-keyboards.md
```diff
@@ -1163,6 +1167,119 @@ This attribute may be escaped with `\u` notation, see [Escaping](#Escaping).
...
+
+> - `touch` - Touch (non-hardware) layout. 
+> - `abnt2` - Brazilian 103 key ABNT2 layout (iso + extra key left of right shift)
+> - `iso` - European 102 key layout (extra key right of left shift)
```
miloush, line 1251: Pick either right or left, that is important for the order of elements. I would suggest left.

srl295: i'm going to change it to 'near'.

srl295:
@miloush good comments, take a look now

PR #2934, opened 2023-05-17
CLDR-16322 Make SurveyTool personName pattern ordering as in personName report

pedberg-icu:
CLDR-16322

The idea here is to have the SurveyTool show personName patterns with a  grouping and tagging that matches the personName report. So this puts the patterns into 5 groups, in the following order:

1. Main patterns for order givenFirst
2. Main patterns for order surnameFirst
3. Sorting patterns (order is sorting, usage is referring)
4. Monogram patterns for order givenFirst (usage is monogram)
5. Monogram patterns for order surnameFirst (usage is monogram)

Within the first two sections, individual items are tagged with length-usage-formality[-alt], but ordered by length, formality then usage as in the report.

In the other sections, items are only tagged with length-formality[-alt], since usage is implied by the section (for sorting it is referring, for monogram it is monogram).

Will need to verify all of this on smoketest after PR is merged.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
@@ -2051,17 +2051,6 @@ public class PathHeader implements Comparable<PathHeader> {
                             // sampleName item values in desired sort order
                             final List<String> itemValues =
                                     PersonNameFormatter.SampleType.ALL_STRINGS;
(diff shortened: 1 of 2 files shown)


richgillam (approved):
I don't understand this code well enough to say anything intelligent, but I'll give you the approval anyway...

pedberg-icu:
Checked in smoketest, seems to be working as intended

PR #2868, opened 2023-04-18
CLDR-15296 Add NIST units

macchiati:
CLDR-15296

This is preparation for adding more units (eg Japanese units of length), and adding NIST quantities.

common/supplemental/units.xml
- added -it for BTU, calorie variants
- changed comment to description
- fixed 'dose' to be more general
- added some english units and NIST units
- ignore the commented-out abhenry unit alias

common/validity/unit.xml
- added new long units

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
- Changed the MapComparator<String> unitOrder to be data-driven. 
- Ignore the commented out old stuff for this PR
- Changed raw access to unitOrder to be a function getUnitOrder(), allowing for lazy loading (needed to avoid circularity!)

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
- Changed raw access to unitOrder to be a function getUnitOrder()

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
- Minor changes to allow for longer repeating format

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
- Better exception message
- Added pow4 (needed for NIST)
- Added kilogram-force to the SKIP_PREFIX set of exceptions

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitParser.java
- added method to get left-over items (used in test)
- fixed problem with type of value

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/ExternalUnitConversionData.java
tools/cldr-code/src/test/java/org/unicode/cldr/unittest/NistUnits.java
- Added information about the source ('from')
- Added handling of BTU IT and calorie IT

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/GenerateNewUnits.java
- Added tool to make it easier to add new units based on data files. For now, focused on NIST
- This replaces a hack in TestUnits at around 211

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUnits.java
- Broke out the debugging flags to make more granular
- Added simple TestAUnits() for displaying information (if isVerbose)
- Around line 330, just formatting differences. Easiest to Hide Whitespace in the diff.
- Added quantityToCategory and CATEGORY_EXCEPTIONS, so that we could do a more granular test, and get rid of a very ugly set of switches, at around 767 below.
- Added BTU IT and calorie IT to STILL_RECOGNIZED_SIMPLES
- Added new test TestValidityAgainstUnitFile
[...]

PR #2939, opened 2023-05-19
CLDR-15296 Restore previous XML ordering for units

macchiati:
CLDR-15296

A previous change modified the DTD attribute order to be driven from the Validity data. This causes the ordering of units to change in all the locale files when CLDRModify is run. This reverts that change. It also runs CLDRModify (no args) to ensure that unit order doesn't change.

Note: In the process of fixing that, I developed code that does a regularized order that should allow use to use the same order in the Survey Tool and in XML. But that was too disruptive for now; but I want to keep the tool in the source for later refinement.

Longer term I want to restore the use of validity, instead of a hard-coded list. But we need to get this restored right away so we can use CLDRModify.

common/supplemental/units.xml
- added missing conversion exposed by tests

common/validity/unit.xml
- restored list, with new items.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/UnitValidityCanonicalizer.java
- tool to use validity (eventually)

- restored use of hard coded list, for now

tools/cldr-code/src/main/java/org/unicode/cldr/util/Units.java
- utility for new tool

- small fixes for test

ALLOW_MANY_COMMITS=true


btangmu:
There are some compiler errors including:

Error:  /home/runner/work/cldr/cldr/tools/cldr-code/src/main/java/org/unicode/cldr/tool/UnitValidityCanonicalizer.java:[22,35] org.unicode.cldr.util.Units.TypeAndCore is not public in org.unicode.cldr.util.Units; cannot be accessed from outside package

And a formatting error, which should be fixable by running `mvn --file=tools/pom.xml spotless:apply`

I don't see any other problems.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/UnitValidityCanonicalizer.java
```diff
@@ -0,0 +1,155 @@
...
+                                    "pressure",
+                                    "temperature",
+                                    "digital",
+                                    "short",
+                                    "angle",
+                                    "concentr",
+                                    "consumption",
+                                    "force",
+                                    "light",
+                                    "torque"))
```
srl295, line 54: why is this a static list? Should this go into supplemental metadata?

srl295 (approved):
changes seem OK, though we should be removing rather than adding static lists.

Was the "prior change" #2868 ?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Units.java
```diff
@@ -81,6 +82,32 @@ public static String combinePattern(
...
+    }
+    /**
+     * Returns the type and core for a unit, be it long or short
+     *
+     * @param longOrShortUnit
+     * @param core
+     * @return
+     * @return
+     */
+    public static TypeAndCore splitUnit(String longOrShortUnit, TypeAndCore typeAndCore) {
```
btangmu, line 97: is splitUnit part of determining the order of items, as in the xml? if so, ideally the order could depend less on special-purpose java methods that may be tightly coupled to many other parts of the code -- something simpler and more data-driven, and loosely coupled, might be more robust and maintainable

macchiati: We have unfortunately both long ids and short ids for units: length-kilometer and kilometer. The former are in our locale files, and used in the ST ordering, the latter are "real" and used in the code.
There is a ticket to drop the "length-" etc subtag, but for now we have to work with it.

PR #2940, opened 2023-05-22
CLDR-16490 Using "Kanton" as the exemplary city for `Pacific/Enderbury`

robertbastian:
CLDR-16490

ALLOW_MANY_COMMITS=true

This is achieved by using the current exemplar cities of `Pacific/Kanton` (we already have this data) for `Pacific/Enderbury` (the canonical but deprecated tzdb name that we're stuck with).


robertbastian:
Actually I'm going to close this PR, it's not clear to me whether this will fix the issue. There are still over 1000 occurences of `[^/]Enderbury`.

PR #2941, opened 2023-05-22
CLDR-14380 Remove db backup from production after rsync to corp

btangmu:
-Add --remove-source-files option to rsync; only removes if successful

-Also change -q to equivalent --quiet for clarity

CLDR-14380

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/scripts/ansible/templates/cldrbackup/backup_sh.j2
+++ b/tools/scripts/ansible/templates/cldrbackup/backup_sh.j2
@@ -17,4 +17,4 @@ mysqldump ${DATABASE_NAME} --add-drop-table --create-options --default-character
 
 # Wait until mysqldump is finished before running xz, to minimize db disruption (don't use | pipe)
 xz ${FNAME}.sql
-rsync -q ${FNAME}.sql.xz ${RSYNC_DEST}
+rsync --quiet --remove-source-files ${FNAME}.sql.xz ${RSYNC_DEST}


btangmu:
The "testprovision" check still says "Waiting for a runner to pick up this job..." after 5 hours?

btangmu:
> Depends on a Mac .. yeah, maybe out of runners

@srl295 I don't know what that means... I see a reference to "testprovision" at https://github.com/unicode-org/cldr/actions/runs/5047011697/workflow?pr=2941 -- see below -- but, what's it all about? Is this something new? How can my PR get unblocked?

```
...
  testprovision:
    runs-on: macos-10.15  # see https://github.com/actions/virtual-environments/issues/4060
    steps:
    - uses: actions/checkout@v3
      with:
        lfs: false # not needed for this job, as we don’t currently do a Java build
    - name: Install Vagrant and Ansible
      run: brew install ansible
    - name: Prepare the provision
      run: (cd tools/scripts/ansible && ansible-galaxy install -r requirements.yml && ln -sv test-local-vars local-vars && vagrant validate)
    - name: Try the provision (ignored, see CLDR-14757)
      run: (cd tools/scripts/ansible && vagrant up) || ( echo Failing But Not Blocking CLDR-14757 ; true)
``

srl295 (approved):

srl295:
@btangmu  It's not new. 

A "runner" is what runs Github Actions.

Maybe there's a lack of Mac runners for some reason.

Ansible Lint doesn't block, you can merge

PR #2942, opened 2023-05-22
CLDR-13962 Announcements: show in header if unread; fix db; fix locales

btangmu:
-New pulsating balloon 🎈 emoji link in main header with unread count, if nonzero

-Remove the word Menu adjacent to ☰ icon in main header to make more room

-New cldrAnnounce.getUnreadCount, etc.

-Fix db bugs: no DB_SQL_IDENTITY for ANNOUNCE_READ; skip add/deleteCheckRow when superfluous

-Add primary key for ANNOUNCE_READ on announce_id,user_id

-New api/locales/combine-variants converts fr_CA to fr (etc), and normalizes

-Form for entering locales calls combine-variants automatically

CLDR-13962

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/MainHeader.vue
+++ b/tools/cldr-apps/js/src/views/MainHeader.vue
@@ -3,7 +3,13 @@
     <ul>
       <li>{{ stVersionPhase }}</li>
       <li>
-        <a href="#menu///"><span class="main-menu-icon">☰</span> Menu</a>
+        <a href="#menu///"><span class="main-menu-icon">☰</span></a>
+      </li>
+      <li v-if="unreadAnnouncementCount">
+        <a href="#announcements///" v-bind:title="announcementsTitle"
+          ><span class="attention-icon">🎈</span>
+          {{ unreadAnnouncementCount }}</a
+        >
       </li>
       <li v-if="coverageLevel">
         <label for="coverageLevel">Coverage:</label>
@@ -71,6 +77,7 @@
 </template>
 
 <script>
+import * as cldrAnnounce from "../esm/cldrAnnounce.mjs";
 import * as cldrCoverage from "../esm/cldrCoverage.mjs";
 import * as cldrMenu from "../esm/cldrMenu.mjs";
 import * as cldrStatus from "../esm/cldrStatus.mjs";
@@ -80,6 +87,7 @@ import * as cldrVote from "../esm/cldrVote.mjs";
 export default {
   data() {
     return {
+      announcementsTitle: null,
       coverageLevel: null,
       coverageMenu: [],
       coverageTitle: null,
@@ -89,6 +97,7 @@ export default {
       sessionMessage: null,
       specialHeader: null,
       stVersionPhase: null,
+      unreadAnnouncementCount: 0,
       userName: null,
       voteCountMenu: null,
       voteLevelChanged: 0,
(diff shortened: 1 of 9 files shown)


btangmu:
This might complete the ticket except that it needs more testing, and a follow-up PR is needed for new TODO comments in SurveyForum.java

tools/cldr-apps/js/src/views/MainHeader.vue
```diff
@@ -160,6 +170,13 @@ export default {
     setVoteLevel() {
       cldrVote.setVoteLevelChanged(this.voteLevelChanged);
     },
+
+    setUnreadCount(n) {
+      this.unreadAnnouncementCount = n;
+      this.announcementsTitle = n
+        ? "You have " + n + " unread announcement(s)"
```
srl295, line 177: someday we should use plurals…

srl295: and MFWG2.0

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleList.java
```diff
@@ -87,8 +88,43 @@ public Response normalize(
...
+                                "Combined/normalized response; e.g., given zh fr_BE fr_CA, return fr zh",
+                        content =
+                                @Content(
+                                        mediaType = "application/json",
+                                        schema =
+                                                @Schema(
+                                                        implementation =
+                                                                LocaleNormalizerResponse.class))),
+            })
+    public Response combineRegionalVariants(
```
srl295, line 114: should this belong on the normalize api? if it's not really connected to forums per se?

srl295 (approved):

tools/cldr-apps/js/src/views/MainHeader.vue
DavidLRowe: Or cheat and use:
"Unread announcements: " + n

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyForum.java
```diff
@@ -88,7 +90,7 @@ private static String HTMLUnsafe(String s) {
 
     private synchronized int getForumNumber(CLDRLocale locale) {
         String forum = localeToForum(locale);
-        if (forum.length() == 0) {
+        if (forum.length() == 0 || LocaleNames.ROOT.equals(forum)) {
```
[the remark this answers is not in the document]
btangmu: This isn't needed yet, but the current behavior depends on a difference between ULocale and CLDRLocale -- which return "" (empty) versus "root" for the root locale -- the only time length would be zero here is for root

srl295:
Good

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/LocaleList.java
btangmu: I'm not sure what you mean -- this is the same file as the normalize api and it shares some code with it -- and you're right, it's not connected to forums

srl295: Sorry, i misread, sounds good!

PR #2943, opened 2023-05-23
CLDR-16382 Add default parameters for PNF

macchiati:
CLDR-16382

common/dtd/ldml.dtd
common/main/en.xml
common/main/root.xml
tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
- Standard stuff when adding new structure:

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
- Standard stuff to test new structure

- load new structure (just to make sure we can; not needed in CLDR examples)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
@@ -40,7 +40,7 @@ public class ToolConstants {
                             .map(x -> VersionInfo.getInstance(x))
                             .collect(Collectors.toList()));
 
-    public static final String DEV_VERSION = "43";
+    public static final String DEV_VERSION = "44";
     public static final VersionInfo DEV_VERSION_VI = VersionInfo.getInstance(DEV_VERSION);
 
     public static final Set<String> CLDR_RELEASE_VERSION_SET =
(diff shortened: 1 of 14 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
```diff
@@ -40,7 +40,7 @@ public enum ChartStatus {
                             .map(x -> VersionInfo.getInstance(x))
                             .collect(Collectors.toList()));
 
-    public static final String DEV_VERSION = "43";
+    public static final String DEV_VERSION = "44";
```
srl295, line 43: oops

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
macchiati: I needed to generate the charts ... and that required bumping the version.

PR #2951, opened 2023-05-24
CLDR-16534 Downgrade paths for islamic calendar display names (renaming to use Hijri)

pedberg-icu:
CLDR-16534

One part of changing the islamic calendar display names to use Hijri: Downgrade the relevant paths.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/downgrade.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/downgrade.txt
@@ -12,6 +12,10 @@
 
 # The rest of the paths from https://unicode-org.atlassian.net/browse/CLDR-16503 need to be included
 
+# CLDR-16534 Downgrade paths for renaming Islamic Calendar to Hijri Calendar
+
+	;	 //ldml/localeDisplayNames/types/type[@key="calendar"][@type="islamic
+
 # CLDR-16228 Add the paths where spaces changed in cjk+ person names
 # CLDR-16568 Require checking of questionable PNF data
 


tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
```diff
@@ -12,6 +12,9 @@
 
 # CLDR-16228 Add the paths where spaces changed in cjk+ person names
 # CLDR-16568 Require checking of questionable PNF data
+# CLDR-16534 Downgrade paths for renaming Islamic Calendar to Hijri Calendar
```
macchiati, line 15: I think we want to move this up, so the comment about PNF data is just before the PNF paths

macchiati (changes requested):

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
btangmu: I've just merged the commit that moves this file from util to util/data

pedberg-icu: > I think we want to move this up, so the comment about PNF data is just before the PNF paths

@macchiati I rearranged the lines as requested (I think)

btangmu: yes, different topic, however it's the same file so potential merge conflict

pedberg-icu: @macchiati OK I have resolved conflicts, please approve, thanks

PR #2952, opened 2023-05-24
CLDR-16670 downgrade

macchiati:
CLDR-16670
Downgrade paths as in downgrade.txt

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -6475,9 +6475,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<unitPattern count="other">{0} kalorieë</unitPattern>
 			</unit>
 			<unit type="energy-foodcalorie">
-				<displayName>Kalorieë</displayName>
-				<unitPattern count="one">{0} kalorie</unitPattern>
-				<unitPattern count="other">{0} kalorieë</unitPattern>
+				<displayName draft="provisional">Kalorieë</displayName>
+				<unitPattern count="one" draft="provisional">{0} kalorie</unitPattern>
+				<unitPattern count="other" draft="provisional">{0} kalorieë</unitPattern>
 			</unit>
 			<unit type="energy-kilojoule">
 				<displayName>kilojoule</displayName>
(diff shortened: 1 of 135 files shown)


pedberg-icu:
Hmm, this does not include the downgrade in https://github.com/unicode-org/cldr/pull/2951 which is not yet approved or merged. Should it wait for that?

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
```diff
@@ -70,20 +82,33 @@ public static LocalePathValueListMatcher load(Stream<String> lines) {
...
     public boolean lookingAt(String locale, String path, String value) {
         for (LocalePathValueListMatcher.LocalePathValueMatcher lpv : matchData) {
-            if (!lpv.lookingAt(locale, path, value)) {
-                return false;
+            if (lpv.lookingAt(locale, path, value)) {
+                return true;
             }
         }
-        return true;
+        return false;
```
btangmu, line 89: this is the key fix I had in mind

tools/cldr-code/src/main/resources/org/unicode/cldr/util/downgrade.txt
```diff
@@ -5,8 +5,10 @@
 # If you need to match to the end, follow with $
```
btangmu, line 5: If this file were in util/data rather than in util, then maybe methods like CldrUtility.getUTF8Data would be easier to use -- is there a reason it's not in util/data?

macchiati:
> #2951

We don't need to wait; the processing is idempotent, so we can run CLDRModify -fD after your changes are accepted. In the future, it is probably simplest to run the -fD when the new paths are added to downgrade.txt; that way the changes are in separate commits, and easier to review.

macchiati:
So if you could review this, @pedberg-icu , then I can merge — and then follow up with your additional changes.

pedberg-icu (approved):

pedberg-icu:
> So if you could review this, @pedberg-icu , then I can merge — and then follow up with your additional changes.

Looks good!

PR #2953, opened 2023-05-24
CLDR-16654 Revise the downgrade tool

macchiati:
CLDR-16654

Fix the tool to:

- Have the right polarity on the test
- Ignore en and root
- Have a fast path to ignore locales that can't match any line in downgrade.txt

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
@@ -849,7 +849,7 @@ public class CLDRModify {
             } else if (oldValueNewPath.equals(newValue)) {
                 showAction(
                         reason,
-                        "Redundant Value",
+                        "Unchanged Value",
                         oldValueOldPath,
                         oldValueNewPath,
                         newValue,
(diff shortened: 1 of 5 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
```diff
@@ -70,20 +82,33 @@ public static LocalePathValueListMatcher load(Stream<String> lines) {
...
+            }
+        }
+        return false;
     }
 
     private static void load(
             String line, List<LocalePathValueListMatcher.LocalePathValueMatcher> _matchData) {
         line = line.trim();
-        if (line.startsWith("#")) {
+        if (line.startsWith("#") || line.isEmpty()) {
```
btangmu, line 105: this explains those "null\tnull\tnull" I was seeing!

btangmu (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocalePathValueListMatcher.java
macchiati: Yup! When I first tested I had no blank lines!

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2967,8 +2967,25 @@ public void handleEnd() {
                 'D',
                 "Downgrade paths",
                 new CLDRFilter() {
+
+                    boolean skipLocale = false;
+
+                    @Override
+                    public void handleStart() {
+                        // TODO Auto-generated method stub
```
btangmu, line 2975: TODO Auto-generated method stub

btangmu (approved):

PR #2955, opened 2023-05-24
CLDR-16661 Use /util/DowngradePaths.java when importing votes

btangmu:
-Filter by DowngradePaths.lookingAt when importing old winning or losing votes

-Revise DowngradePaths to use getUTF8Data(downgrade.txt); move downgrade.txt to util/data

-Make sure vote table exists before importing old votes, call STFactory.setupDB

-Make STFactory.setupDB public (there should be a better way, required for table access)

CLDR-16661

ALLOW_MANY_COMMITS=true


macchiati (commented):
Looks good to me, but Steven should review the ST stuff.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -643,6 +643,7 @@ private void processRequest(
                         send(r, out);
                     } else if (what.equals(WHAT_OLDVOTES)) {
                         mySession.userDidAction();
+                        CookieSession.sm.getSTFactory().setupDB();
```
srl295, line 646: Probably ok for now

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1992,7 +2009,7 @@ private void doAutoImportOldWinningVotes(SurveyJSONWrapper r, User user, SurveyM
             return;
         }
         alreadyAutoImportedVotes(user.id, "set");
-
+        CookieSession.sm.getSTFactory().setupDB();
```
srl295, line 2012: For the record could just skip if not existing. But ok.

tools/cldr-code/src/main/java/org/unicode/cldr/util/DowngradePaths.java
```diff
@@ -1,15 +1,9 @@
...
-import org.unicode.cldr.draft.FileUtilities;
-
 public class DowngradePaths {
 
     static LocalePathValueListMatcher data =
-            LocalePathValueListMatcher.load(
-                    Paths.get(
-                            FileUtilities.getRelativeFileName(
-                                    DowngradePaths.class, "downgrade.txt")));
+            LocalePathValueListMatcher.load(CldrUtility.getUTF8Data("downgrade.txt").lines());
```
srl295, line 6: +1 this is what I expected here

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
btangmu: I don't think "could just skip": if the new table doesn't exist yet, the old votes won't get imported unless the table gets created here.

PR #2958, opened 2023-05-25
CLDR-16384 add grammatical modifiers

macchiati:
CLDR-16384

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
- In setCldrFileToCheck, compute the Modifiers that are allowed in this locale (depending on grammar) and cache.
-  checkPersonNamePatterns adds some parameters, and now checks for the allowed Modifiers
- changed \" to «, etc. Makes dealing with test easier.

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
- Added the capability to have personNames as a scope for grammar. Unused right now, but will add in release.

- added constants and utilities to Modifier. For now, just two grammatical categories; may add more later (We need to tell vetters that if they need more than these, they need to communicate with us.
- have 'fake' fallbacks to let people know that that sample names would be modified to be genitive or vocative

- add fake fallbacks to test
- modify the CheckAccessorStub so we can get the allowed modifiers
- TestCheckPatterns: added two tests that fr disallows -vocative and ru allows it.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
@@ -22,6 +22,7 @@ import java.util.regex.Pattern;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Type;
 import org.unicode.cldr.util.CLDRConfig;
+import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.LocaleNames;
 import org.unicode.cldr.util.Pair;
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
```diff
@@ -94,7 +105,8 @@ public CheckCLDR handleCheck(
                     checkSampleNames(this, parts, value, result);
                     return this;
                 case "personName":
-                    checkPersonNamePatterns(this, path, parts, value, result);
+                    checkPersonNamePatterns(
+                            this, allowedModifiers, getLocaleID(), path, parts, value, result);
```
btangmu, line 109: best practice is max 4 or 5 parameters per method

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckPlaceHolders.java
```diff
@@ -313,9 +325,32 @@ public static void checkSampleNames(
...
+            List<CheckStatus> result) {
+        checkPersonNamePatterns(
+                checkAccessor, allowedModifiers, "fr", path, pathParts, value, result);
+    }
+
+    /**
+     * Verify the that personName patterns are clean.
+     *
+     * @param locale TODO
+     * @param path TODO
```
btangmu, line 348: omitting @param would be better than "TODO"

btangmu (approved):
LGTM other than minor quibbles

pedberg-icu (approved):

PR #2959, opened 2023-05-25
CLDR-16670 Downgrade name for islamic calendar to force revisit

macchiati:
CLDR-16670

Peter made the change to downgrade.txt; this runs the tool

ALLOW_MANY_COMMITS=true


DavidLRowe (approved):
LGTM (spot check only)

pedberg-icu (approved):

PR #2960, opened 2023-05-25
CLDR-16385 Added a new "retain" modifier that, when used with "initia…

richgillam:
…l", preserves any punctuation that might go

between the initials when a field contains multiple words (i.e., "Jean-Michel" would initialize to "J.-M." rather than "J. M.").

CLDR-16385

ALLOW_MANY_COMMITS=true


macchiati (approved):
Looks great!

macchiati:
Merging, since Rich is away

PR #2012, opened 2022-05-16
CLDR-13963 show placeholders in info panel

srl295:
- also add to Row payload
- also add to ("new") Vote API

CLDR-13963

Image example:

image

- [X] This PR completes the ticket.

(though, we can tweak the style if desired, in this or a future PR)

PR #2967, opened 2023-05-26
CLDR-13916 Update help link for unitLength...gender; make clickable

btangmu:
-Add CLDRURLS.GRAMMATICAL_INFLECTION for .../unitLength.../gender paths

-Make helpHtml links clickable in Info Panel

-Remove dead code from PathDescription.java

CLDR-13916

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
@@ -8,6 +8,9 @@ const abstractLang = "en";
 const USELESS_IN_INFO_PANEL = "see info panel";
 
 function addDeferredHelpTo(fragment, helpHtml, resource, translationHint) {
+  if (helpHtml) {
+    helpHtml = linkifyHttpsUrls(helpHtml);
+  }
   // Always have help (if available).
   const theHelp = $("<div/>", {
     class: "alert alert-info fix-popover-help",
@@ -43,6 +46,16 @@ function addDeferredHelpTo(fragment, helpHtml, resource, translationHint) {
   }
 
   $(fragment).append(theHelp);
+}
+
+function linkifyHttpsUrls(text) {
+  const regex = /\b(https:\/\/\S+)/g;
+  const s = text.replace(regex, "<a target='_blank' href='$1'>$1</a>");
+  // In principle, a URL can end in a period;
+  // in practice, a final period in PathDescription.java is not part of the URL
+  const periodQuoteEndBracketRegex = /\.'>/g;
+  const quoteEndBracket = "'>";
+  return s.replace(periodQuoteEndBracketRegex, quoteEndBracket);
 }
 
 function subloadAbstract(resource) {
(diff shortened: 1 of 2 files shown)


macchiati (commented):
Looks ok, but Steven should look at the mjs

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
```diff
@@ -8,6 +8,9 @@ const abstractLang = "en";
 const USELESS_IN_INFO_PANEL = "see info panel";
 
 function addDeferredHelpTo(fragment, helpHtml, resource, translationHint) {
+  if (helpHtml) {
+    helpHtml = linkifyHttpsUrls(helpHtml);
```
srl295, line 12: fine for now, pending a better solution in https://unicode-org.atlassian.net/browse/CLDR-15708

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
```diff
@@ -45,6 +48,16 @@ function addDeferredHelpTo(fragment, helpHtml, resource, translationHint) {
...
+function linkifyHttpsUrls(text) {
+  const regex = /\b(https:\/\/\S+)/g;
+  const s = text.replace(regex, "<a target='_blank' href='$1'>$1</a>");
+  // In principle, a URL can end in a period;
+  // in practice, a final period in PathDescription.java is not part of the URL
+  const periodQuoteEndBracketRegex = /\.'>/g;
+  const quoteEndBracket = "'>";
+  return s.replace(periodQuoteEndBracketRegex, quoteEndBracket);
+}
+
```
srl295, line 60: lots of problems here.  OK for now, pending https://unicode-org.atlassian.net/browse/CLDR-15708

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
```diff
@@ -1025,32 +1023,6 @@ public String getDescription(String path, String value, Object context) {
         return description;
     }
 
-    /**
-     * Creates an escaped HTML string of placeholder information.
-     *
-     * @param path the xpath to specify placeholder information for
-     * @return a HTML string, or an empty string if there was no placeholder information
-     */
-    public String getPlaceholderDescription(String path) {
```
srl295, line 1034: huh, i wonder when this got dropped?  Probably accidental.

This was added in CLDR-4544, and then unknowingly replicated in https://github.com/unicode-org/cldr/pull/2012

srl295 (approved):
It's OK for now. I'm not at all a fan of auto-linkification.

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
btangmu: I agree about auto-linkification not being the best way, still I think it's better than expecting ST users to copy and paste links. Perfect is the enemy of good.

srl295: Yes, and I'm glad to approve this and improve things for users.

PR #2969, opened 2023-05-26
CLDR-13916 Follow up on auto-linkification

btangmu:
-Add TODO comment referencing ticket for better solution

-Do not linkify if text already contains href (be cautious even if maybe not needed currently)

CLDR-13916

ALLOW_MANY_COMMITS=true


srl295 (approved):
Even better would be to mention CLDR-15708 in the commit so it can be found. but minor. I've already referenced CLDR-13916 from CLDR-15708

btangmu:
I avoid referencing other tickets in commit messages, to prevent spurious detection of PRs on tickets that don't really have PRs

PR #2971, opened 2023-05-27
CLDR-16680 Fix bug preventing voting for aliased values

btangmu:
-In DataRow.getVoteVHash inherited value matches inheritance marker and vice-versa

CLDR-16680

ALLOW_MANY_COMMITS=true


btangmu:
on the back end, the "hard" inherited value is used instead of inheritance marker in special cases involving lateral inheritance, while for the front end there is no such distinction, and inheritance marker is sent to the front end uniformly

this bug resulted from failing to take that fact into account in the response data indicating which item the current user voted for (voteVHash)

the solution here is to convert voteVHash as needed to match the convention used in the rest of the response (which currently uses inheritance marker -- however, this bug fix would also work if the convention were reversed)

macchiati (approved):

macchiati:
Thanks! Makes sense.

PR #2913, opened 2023-05-10
CLDR-16629 Account for macroregions in likely subtags

macchiati:
CLDR-16629

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #2975, opened 2023-05-29
CLDR-16629 Account for macroregions in likely subtags (#2913)

pedberg-icu:
(cherry picked from commit 5f36fda0dea159aa5856a7d27d019062693df312)

CLDR-16629

The fix for 43.1 issue [CLDR-16629](https://unicode-org.atlassian.net/browse/CLDR-16629) (Many (almost most) "und_" + 3 digits entry in likelySubtags.xml will produced suprising results.) was committed to main, cherry-pick it to the maint/maint-43 branch

ALLOW_MANY_COMMITS=true

[CLDR-16629]: https://unicode-org.atlassian.net/browse/CLDR-16629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -2165,8 +2165,7 @@ To look up data in the table, see if a locale matches one of the `from` attribut
 So looking up "zh_TW" returns "zh_Hant_TW", while looking up "zh" returns "zh_Hans_CN".
 
 In more detail, the data is designed to be used in the following operations.
-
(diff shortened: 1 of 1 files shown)


macchiati (approved):

FrankYFTang:
Mark- is there a table in CLDR define which region is "macroregion"?

macchiati:
https://github.com/unicode-org/cldr/blob/main/common/validity/region.xml

FrankYFTang:
Dear Mark- I do not believe this change fix the issue 
You change the text from
"Let x<sub>r</sub> = x<sub>s</sub> if x<sub>s</sub> is not empty, and x<sub>m</sub> otherwise."
to
"Let x<sub>r</sub> = x<sub>s</sub> if x<sub>s</sub> is not empty or x<sub>s</sub> is a macroregion, and x<sub>m</sub> otherwise."

but for every x<sub>s</sub> which is a macroregion, it is for sure "not empty" already. so this change of algorithm will produce no changes of the outcome, right?

macchiati:
You're right; I'll file another PR.

PR #2980, opened 2023-05-31
CLDR-16600 Improve UI for Explain Vote Counts

btangmu:
-Position the button in the same place as the transcript, below the vote div

-Use the same button to reveal and hide (toggle)

-Move the code to a new subroutine makeVoteExplainerDiv

-Make the feedback link clickable

CLDR-16600

ALLOW_MANY_COMMITS=true


btangmu:
@macchiati this is how it looks before clicking the blue "i" button:
image

-- and this is how it looks after clicking the button and scrolling down:
image

btangmu:
a further improvement would be to make the "Feedback" link clickable

srl295 (approved):

PR #2983, opened 2023-05-31
CLDR-16146 Downgrade main path for BIOT

macchiati:
CLDR-16146

Followup to downgrade main path

ALLOW_MANY_COMMITS=true


macchiati:
Also updated the Info Hub; see the ticket.

pedberg-icu (approved):

PR #2984, opened 2023-06-01
CLDR-16482 Add core data for mic, change English name to Mi'kmaw

pedberg-icu:
CLDR-16482

Add core data for `mic`. Change the English name from Mi'kmaq to Mi'kmaw (note that English uses a straight quote, while `mic` itself uses 02BC for instance in its autonym Lʼnuiʼsuti). Data from Alla Shashkina and Dragan Bešević (with input from Mi'kmaw Kina'matnewey), I just made the CLDR PR.

Also note:
- We already included the display name for `mic` in modern coverage.
- It was also already in attributeValueValidity.xml
- We also already had territoryInfo data for `mic` in Canada.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -408,7 +408,7 @@ annotations.
 			<language type="mgo">Metaʼ</language>
 			<language type="mh">Marshallese</language>
 			<language type="mi">Māori</language>
-			<language type="mic">Mi'kmaq</language>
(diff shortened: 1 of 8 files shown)


macchiati (approved):

ashashkina (approved):
Thanks @pedberg-icu !

PR #2999, opened 2023-06-01
CLDR-7536 Info Panel not for specials; preliminary refactoring

btangmu:
-Do not use the Info Panel for specials; instead, show guidance inline for Bulk Close Posts, Flagged Items, List Emails, Vetting Participation

-Remove Forum Participation guidance, was useless: This is the Forum Participation page

-Usage of Info Panel for specials needlessly complicated the code; simplification facilitates modernization

-Refactor cldrInfo.mjs slightly to facilitate further changes and modernization

-Rename pucontent to itemInfoElement for clarity

-In cldrLoad, use notification.error instead of cldrInfo.showMessage to report problem loading data

-Fix bug in cldrForumParticipation.mjs discovered while testing: no json in error handler

-Fix bugs in cldrListEmails.mjs discovered while testing: import cldrRetry; no json in error handler

-Remove dead code in cldrVote.mjs (!value is impossible here)

-TODO comments

CLDR-7536

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -47,12 +47,8 @@ function initialize(cid, nid, bclass) {
 
 function insertWidget() {
   try {
-    const fragment = document.createDocumentFragment();
-    createCldrApp(InfoPanel).mount(fragment); // returns App object "wrapper", currently not saved
     const containerEl = document.getElementById(containerId);
-    const vueEl = document.createElement("section");
-    containerEl.appendChild(vueEl);
-    vueEl.replaceWith(fragment);
+    insertVueElement(containerEl);
     insertLegacyElement(containerEl);
   } catch (e) {
     console.error("Error loading InfoPanel vue " + e.message + " / " + e.name);
@@ -62,6 +58,14 @@ function insertWidget() {
       duration: 0,
     });
   }
+}
+
+function insertVueElement(containerEl) {
+  const fragment = document.createDocumentFragment();
+  createCldrApp(InfoPanel).mount(fragment); // returns App object "wrapper", currently not saved
+  const vueEl = document.createElement("section");
+  containerEl.appendChild(vueEl);
+  vueEl.replaceWith(fragment);
 }
 
 /**
@@ -78,6 +82,10 @@ function insertLegacyElement(containerEl) {
   nonVueEl.className = "sidebyside-scrollable";
   nonVueEl.id = "itemInfo";
   containerEl.appendChild(nonVueEl);
+  // TODO: here, create divs inside itemInfo, each containing a part of the Info Panel
+  // which may be either a legacy div or a Vue component; never destroy these
+  // divs wholesale by calling removeAllChildNodes on itemInfo.
+  // Reference: https://unicode-org.atlassian.net/browse/CLDR-7536
 }
 
 function openPanel() {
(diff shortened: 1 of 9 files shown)


tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -243,63 +287,66 @@
   }
 
   if (str) {
-    // If a simple string, clone the string
-    var div2 = document.createElement("div");
-    div2.innerHTML = str;
-    fragment.appendChild(div2);
+    const div = document.createElement("div");
+    div.innerHTML = str;
```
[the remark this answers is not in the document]
btangmu: This isn't new, and str is rarely from the user -- the main (only?) exception is showWithRow, and per comments in this PR, the plan is to replace showWithRow with notifications.error

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -64,6 +60,14 @@ function insertWidget() {
   }
 }
 
+function insertVueElement(containerEl) {
+  const fragment = document.createDocumentFragment();
+  createCldrApp(InfoPanel).mount(fragment); // returns App object "wrapper", currently not saved
+  const vueEl = document.createElement("section");
+  containerEl.appendChild(vueEl);
+  vueEl.replaceWith(fragment);
+}
```
srl295, line 69: could go in cldrVueRouter.js

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -78,6 +82,10 @@ function insertLegacyElement(containerEl) {
   nonVueEl.className = "sidebyside-scrollable";
   nonVueEl.id = "itemInfo";
   containerEl.appendChild(nonVueEl);
+  // TODO: here, create divs inside itemInfo, each containing a part of the Info Panel
+  // which may be either a legacy div or a Vue component; never destroy these
+  // divs wholesale by calling removeAllChildNodes on itemInfo.
+  // Reference: https://unicode-org.atlassian.net/browse/CLDR-7536
```
srl295, line 88: Right, ideally the info panel would stick around and just get updated data, never rebuilt from scratch.

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -150,6 +161,22 @@ function updateOpenPanelButtons() {
  * @param {Node} theObj to listen to, a.k.a. "hideIfLast"
  * @param {Function} fn the draw function
  */
+// TODO: simplify this and related code to enable a cleaner and better organized
+// appearance using modern components. The fn parameter (a.k.a. showFn) shouldn't
+// need to be preconstructed for each item in each row, attached to DOM elements,
+// and passed around as a parameter in such a complicated way.
```
srl295, line 167: yes, it wanted to be Vue components. It's a home grown framework. All of the complexity with little of the benefit.

srl295 (approved):
It's a good incremental improvement and documentation, which will be helpful when, as it should be, the raw DOM work is replaced with components.

tools/cldr-apps/js/src/esm/cldrInfo.mjs
btangmu: yes eventually, if there were multiple callers, the component (in this case InfoPanel) could be a parameter...

PR #3001, opened 2023-06-02
CLDR-15296 Add Japanese units

macchiati:
CLDR-15296

This is a first pass at adding extra units, using Japanese units. These need review; they are used internally, but if those of only historical usage we may want to omit.

common/main/ja.xml
- the format would be normalized by CLDRModify but I wanted the simpler grouping for review.

common/supplemental/units.xml
- Some of the units have suffixes
- For accuracy, constants are used to map to a 'pivot' unit, which will have a conversion to a standard base unit.
- Again, the units are all grouped together for now; the normalized format will intersperse them

common/validity/unit.xml
- Again, grouped for now.

- add a jpsystem, like us and uk

- minor changes

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -1476,6 +1476,12 @@ public Set<String> baseUnits() {
         PREFIXES = ImmutableMap.copyOf(temp);
     }
 
+    public static final Set<String> METRIC_TAKING_PREFIXES =
+            ImmutableSet.of(
+                    "bit", "byte", "liter", "tonne", "degree", "celcius", "kelvin", "calorie",
```
pedberg-icu, line 1481: Celsius seems wrong here

pedberg-icu: @macchiati ping

macchiati:
Fixed the misspelling, and adjusted the tests to allow for 6 letter SI prefixes (and adjusted ticket CLDR-16711).

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: Fixed that typo

pedberg-icu (approved):

PR #3018, opened 2023-06-05
CLDR-7536 New cldrNotify to replace showWithRow and encapsulate notification

btangmu:
-Get rid of cldrInfo.showWithRow to simplify cldrInfo and use more modern notifications

-New cldrNotify.mjs with three methods: show, warning, and error

-In JavaScript, always throw new Error instead of throw string so e.message is defined

-Also throw new Error instead of throw Error, for consistency (although they are equivalent)

CLDR-7536

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrAddAlt.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrAddAlt.mjs
@@ -3,13 +3,12 @@
  */
 import * as cldrAjax from "./cldrAjax.mjs";
 import * as cldrLoad from "./cldrLoad.mjs";
+import * as cldrNotify from "./cldrNotify.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 
 import AddAlt from "../views/AddAlt.vue";
 
 import { createCldrApp } from "../cldrVueRouter.mjs";
-
-import { notification } from "ant-design-vue";
 
 function addButton(containerEl, xpstrid) {
   try {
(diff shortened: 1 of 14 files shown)


tools/cldr-apps/js/src/esm/cldrAddAlt.mjs
```diff
@@ -21,11 +20,11 @@ function addButton(containerEl, xpstrid) {
     addAltWrapper.setXpathStringId(xpstrid);
   } catch (e) {
     console.error("Error loading Add Alt vue " + e.message + " / " + e.name);
-    notification.open({
-      message: `${e.name} while loading AddAlt.vue`,
-      description: `${e.message}`,
-      duration: 0,
-    });
+    cldrNotify.open(
+      `${e.name} while loading AddAlt.vue`,
```
srl295, line 24:
```suggestion
      `Internal Error: ${e.name} while loading AddAlt.vue`,
```

AddAlt.vue isn't going to make any sense to users, so I think it's best to preface it with "Internal Error" which is the most important part for them to know.

srl295 (approved):

srl295 (approved):

btangmu:
@srl295 fyi still testing this locally and not seeing what I expect when server returns Response.Status.NOT_ACCEPTABLE

btangmu:
@srl it turns out there's already a bug in main: if I change the server to return Response.Status.NOT_ACCEPTABLE in VoteAPIHelper.handleVote, there's no error message shown except in the browser console, and the pop-up window for entering a value just stays visible:
image

So I made a new ticket for that, blocking this one: https://unicode-org.atlassian.net/browse/CLDR-16721

btangmu:
I'll go ahead and merge this PR, given that https://unicode-org.atlassian.net/browse/CLDR-16721 is a pre-existing bug. There's no indication that this PR makes that bug any worse. Leaving these changes in limbo would only be an obstacle to progress on both tickets. I'll prioritize fixing the bug to make it possible to confirm cldrNotify.error works correctly as a substitute for cldrInfo.showWithRow in the context of voting.

PR #3019, opened 2023-06-05
CLDR-14721 add KGS symbol

srl295:
CLDR-14721

note: character is in Windows 11 but not on mac yet?

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -4355,6 +4355,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<currency type="KYD">
 				<symbol alt="narrow">$</symbol>
 			</currency>
+			<currency type="KGS">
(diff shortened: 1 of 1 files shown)


roozbehp:
Do we also need to add the character to the list of characters with fallbacks?

srl295:
@roozbehp that would be ( per https://unicode-org.atlassian.net/browse/CLDR-14721?focusedCommentId=161189 )  by @pedberg-icu 

> (combining form would be U+0421 СCYRILLIC CAPITAL LETTER ES followed by U+0332 ◌̲ COMBINING LOW LINE)

?

pedberg-icu (approved):

pedberg-icu:
> Do we also need to add the character to the list of characters with fallbacks?

> @roozbehp that would be ...

But the fallback addition can (should) be in a different PR.

PR #3020, opened 2023-06-05
Cldr 16675 generate charts v44

macchiati:
CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
@@ -187,7 +187,7 @@ public class ChartGrammaticalForms extends Chart {
                 if (multiline) {
                     for (GrammaticalScope usage : GrammaticalScope.values()) {
                         values = scopeToValues.get(usage);
-                        if (values.isEmpty()) {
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
```diff
@@ -966,12 +967,16 @@ public static String getBestBaseUnit(
             Rational bestFactor = null;
             Rational inputBoundary = Rational.of(2).multiply(info.factor);
             for (Entry<Rational, String> entry : factorToUnit.entrySet()) {
+                final String possibleUnit = entry.getValue();
+                if (possibleUnit.equals("cup-jp")) {
+                    continue; // skip odd unit
+                }
                 final Rational currentFactor = entry.getKey();
```
srl295, line 974: does this need a ticket for followup?

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartGrammaticalForms.java
macchiati: Good question. This particular line doesn't, but there should be a change to favor the regular metric units. Will do.

PR #3023, opened 2023-06-06
CLDR-13552 Remove alt values for styleName in root

macchiati:
CLDR-13552

The alt values are unnecessary, and for most languages just confusing, because they seem to want people to supply alternative names that don't exist in their languages. So removing from root so they don't propagate further.

ALLOW_MANY_COMMITS=true


srl295 (approved):
Should they actually go into en.xml? as alternative English names?

macchiati:
Good suggestion

macchiati:
Checked, and en.xml already has those.

macchiati:
Pinging for review

stenshamn (approved):

srl295 (commented):
PMLGTM

PR #3024, opened 2023-06-06
CLDR-16688 unbreak locale USER

srl295:
- fix LocaleCompletion to return 404 instead of stack trace on locale USER
- fix cldrProgress to not attempt to query locale USER

CLDR-16688

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrProgress.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.mjs
@@ -366,7 +366,7 @@ function fetchLocaleData(unlessLoaded) {
     return;
   }
   const locale = cldrStatus.getCurrentLocale();
-  if (!locale) {
+  if (!locale || locale === "USER") {
     return; // no locale
   }
   if (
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/js/src/esm/cldrProgress.mjs
```diff
@@ -366,7 +366,7 @@ function fetchLocaleData(unlessLoaded) {
     return;
   }
   const locale = cldrStatus.getCurrentLocale();
-  if (!locale) {
+  if (!locale || locale === "USER") {
```
btangmu, line 369: possible future enhancement, reject if getLocaleInfo returns null

btangmu (approved):

tools/cldr-apps/js/src/esm/cldrProgress.mjs
srl295: basically centralize this code… yeah

PR #3025, opened 2023-06-07
CLDR-15952 Changes to tool for checking coverage

macchiati:
CLDR-15952

Also a fix for Pts (unicode change in property)

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -228,7 +228,7 @@ public static UnicodeSet getAllRgiNoES() {
                 "punctuation",
                 "§ † ‡ \\u0020  , 、 ، ; : ؛ ! ¡ ? ¿ ؟ ¶ ※ / \\ & # % ‰ ′ ″ ‴ @ * ♪ ♭ ♯ ` ´ ^ ¨ ‐ ― _ - – — • · . … 。 ‧ ・ ‘ ’ ‚ ' “ ” „ » « ( ) [ ] { } 〔 〕 〈 〉 《 》 「 」 『 』 〖 〗 【 】"
             },
-            {"currency", "€ £ ¥ ₹ ₽ $ ¢ ฿ ₪ ₺ ₫ ₱ ₩ ₡ ₦ ₮ ৳ ₴ ₸ ₲ ₵ ៛ ₭ ֏ ₥ ₾ ₼ ₿ ؋"},
+            {"currency", "€ £ ¥ ₹ ₽ $ ¢ ฿ ₪ ₺ ₫ ₱ ₩ ₡ ₦ ₮ ৳ ₴ ₸ ₲ ₵ ៛ ₭ ֏ ₥ ₾ ₼ ₿ ؋ ₧ ¤"},
```
srl295, line 231: why is this static data? missing the Kyrgyz Som. Should just extract it from root.

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
macchiati: These are just to get the categories for the characters.

To add any new character is different, since the list is curated; we need to make a conscious decision as to which to translate.

PR #3026, opened 2023-06-07
CLDR-16721 Show a notification if there is an error for a newly submitted value

btangmu:
-Call new cldrNotify.errorWithCallback, with cldrLoad.reloadV to avoid leaving the submission pop-up visible

-This change only affects the behavior when the back end does not return OK (200)

CLDR-16721

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrRetry.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrRetry.mjs
@@ -2,7 +2,6 @@
  * cldrRetry: encapsulate notification (errors, etc.) functions for Survey Tool
  */
 import * as cldrEvent from "./cldrEvent.mjs";
-import * as cldrGenericVue from "./cldrGenericVue.mjs";
 import * as cldrLoad from "./cldrLoad.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/js/src/esm/cldrRetry.mjs
```diff
@@ -2,7 +2,6 @@
  * cldrRetry: encapsulate notification (errors, etc.) functions for Survey Tool
  */
 import * as cldrEvent from "./cldrEvent.mjs";
-import * as cldrGenericVue from "./cldrGenericVue.mjs";
```
btangmu, line 5: I happened to notice this import was unused -- not related to the ticket except that I was investigating cldrRetry since the old code called cldrRetry.handleDisconnect -- VS Code helpfully indicates unused imports

BTW there are other places in cldrVote that still call cldrRetry.handleDisconnect and I don't get the rationale for that

srl295: eslint can given an error on unused imports.

btangmu: eslint wants a config file -- we don't have one yet, do we?

$ eslint tools/cldr-apps/js/src/esm/*.mjs

Oops! Something went wrong! :(

ESLint: 8.42.0

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    npm init @eslint/config

ESLint looked for configuration files in /Users/tbishop/Documents/WenlinDocs/Organizations/Unicode/CLDR_job/cldr/tools/cldr-apps/js/src/esm and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint Discord server: https://eslint.org/chat

PR #3027, opened 2023-06-07
CLDR-16669 Update collation root from unihan-index

echeran:
CLDR-16669

Update the collation root.xml file with the unihan-index.txt contents which has changes for Unicode 15.1 (new radicals).
@markusicu forgot to update this file in PR #3007.

ALLOW_MANY_COMMITS=true


macchiati (approved):

markusicu (approved):
lgtm

PR #3029, opened 2023-06-08
CLDR-7536 Revise cldrNotify

btangmu:
-New method cldrNotify.exception for caught exceptions

-Make duration automatic: no time-out for errors/exceptions, 8 seconds for show/warning

-Minor refactoring and comments

CLDR-7536

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrProgress.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.mjs
@@ -122,11 +122,7 @@ function insertWidget(spanId) {
     console.error(
       "Error loading ProgressMeters vue " + e.message + " / " + e.name
     );
-    cldrNotify.error(
-      `${e.name} while loading ProgressMeters.vue`,
-      e.message,
-      cldrNotify.NO_TIMEOUT
-    );
+    cldrNotify.exception(e, "while loading ProgressMeters.vue");
   }
 }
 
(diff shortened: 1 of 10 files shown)


btangmu:
enough preliminary stuff! after this I'll add more vue components to the info panel...

tools/cldr-apps/js/src/esm/cldrProgress.mjs
```diff
@@ -122,11 +122,7 @@ function insertWidget(spanId) {
     console.error(
       "Error loading ProgressMeters vue " + e.message + " / " + e.name
     );
-    cldrNotify.error(
-      `${e.name} while loading ProgressMeters.vue`,
-      e.message,
-      cldrNotify.NO_TIMEOUT
-    );
+    cldrNotify.exception(e, "while loading ProgressMeters.vue");
```
srl295, line 125: Same. I wonder if we can refactor some functions around the mount() calls.   But in any event, this is great progress.

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrProgress.mjs
btangmu: Yes, I'll do some refactoring around mount() and also prefix the exception() message with "Internal error"... who knows what e.name would even be if mount() ever throws an exception; I'll try to force that to happen in testing, to see what it looks like

btangmu: @srl295 Here's how cldr.exception looks with added "Internal error":

image

OK?

PR #3033, opened 2023-06-09
CLDR-16390 inh explainer: add 'hidden' property

srl295:
- routine on STFactory to calculate if an xpath is hidden
- also, add generated XPaths to the XPathTable

CLDR-16390

Half (back end) implementation for the issue of the broken jump button.  Adds a `hidden: true/false` property to the API. 

Update:  including Front End  also, see image

image

for path `/cldr-apps/v#/de/Duration/1aab94512d97c3fd`

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
@@ -23,6 +23,8 @@ import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.LocaleInheritanceInfo;
 import org.unicode.cldr.util.StringId;
+import org.unicode.cldr.web.CookieSession;
+import org.unicode.cldr.web.STFactory;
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Inheritance.java
```diff
@@ -122,38 +124,45 @@ public Response getInheritanceByHex(
...
+        public boolean hidden = false;
+
+        public LocaleInheritance(LocaleInheritanceInfo info, final STFactory stf) {
             this.reason = info.getReason();
             this.locale = info.getLocale();
             this.attribute = info.getAttribute();
             final String x = info.getPath();
             if (x != null) {
                 this.xpath = StringId.getHexId(x);
+                CookieSession.sm.xpt.getByXpath(x); // make sure it's in the XPT!
```
btangmu, line 154: this is a little mysterious to me; supposing bogus http requests contained bogus xpaths, would they get added to the db?

btangmu (approved):

PR #3035, opened 2023-06-13
CLDR-16741 Update languages/codes: World DataBank

btangmu:
-Update the file world_bank_data.csv from databank.worldbank.org/reports.aspx?source=world-development-indicators

CLDR-16741

ALLOW_MANY_COMMITS=true


btangmu:
After making this PR I realized there's more at https://cldr.unicode.org/development/updating-codes/updating-population-gdp-literacy that I should do first. So, I've changed it to a draft PR.

I've done "Load the World DataBank" but not "Load UN Literacy Data" or "Load CIA Factbook" since I ran into problems with outdated links. It seems that I should still try to run AddPopulationData and ConvertLanguageData before making a PR.

PR #3036, opened 2023-06-13
CLDR-15034 kbd: remove errant reference to final transforms

srl295:
CLDR-15034

Somehow I missed this.  I moved the obsolete `<transforms type="final"/>` examples to become post-reorder examples.

ALLOW_MANY_COMMITS=true


mcdurdin (approved):
LGTM

PR #3039, opened 2023-06-13
CLDR-14883 drop + for unit-gender

srl295:
- for the 'gender' element in units, show fixed candidate items where applicable
- hide the + (add) as it's not possible to add a new value.
- capability to add more fixed candidates in the future

CLDR-14883

German
image

Czech
image

ALLOW_MANY_COMMITS=true


btangmu (approved):
brilliant!!!

PR #3040, opened 2023-06-14
CLDR-16728 Callers should always use .from

macchiati:
CLDR-16728 Callers should always use .from(...) instead of .valueOf(...). Sadly we can't prevent use of valueOf.

Test case is https://st.unicode.org/cldr-apps/v#/gd/PersonNameFormats/14f7e5b0cb052d25

CLDR-16728

ALLOW_MANY_COMMITS=true


srl295:
it's not clear what the specific issue is (.valueOf() does X, .from() does Y) but changes LGTM

macchiati:
valueOf for an enum requires an exact match, or it throws an exception. This is particularly needed for Length, because the enum value does not match the attribute value ('short' can't be an enum value in Java).

PR #3041, opened 2023-06-14
CLDR-16704 markdown in infopanel

srl295:
- use marked (same as in tr-archiver)
- update html to use links in new window
- update 1 path into markdown as a test (exemplars)

CLDR-16704

ALLOW_MANY_COMMITS=true


srl295:
image
### 

image


image

srl295:
image
`<a target="_blank" …` on all links

srl295:
This could be a good candidate to put on staging if anyone wants to take a closer look at the generated HTML.

I also wonder if we need a new structure for the markdown source, it may not work in Java as easily

macchiati:
Up until v40, it was data-driven, in a text file. Placeholders.txt still is. I think the change was to have redirection for the paths.

We should consider reverting to data-driven, but use a bit more structure. Eg

v40
```^//ldml/units/unitLength\[@type="([^"]*)"]/unit\[@type="area-acre"]/displayName ; Display name ({1} form) for “area-acre”, referring specifically to an English acre. Note: before translating, be sure to read http://cldr.org/translation/units-1/units.
```
v43
```
 + "^//ldml/units/unitLength\\[@type=\"([^\"]*)\"]/unit\\[@type=\"area-acre\"]/displayName"
+ RegexLookup.SEPARATOR
+ "Display name ({1} form) for “area-acre”, referring specifically to an English acre. Note: before translating, be sure to read "
+ CLDRURLS.UNITS_HELP
+ ".\n"
```

=> ?? v44
PathDescriptions.txt

```
path=^//ldml/units/unitLength\[@type="([^"]*)"]/unit\[@type="area-acre"]/displayName
message=Display name ({1} form) for “area-acre”, referring specifically to an English acre.
link=UNITS_HELP
```

And a separate file:

HelpLinks.txt

```
UNITS_HELP=http://cldr.org/translation/units-1/units
```

Possibly a json format. (Not suggesting XML; that's overkill)

macchiati:
> This could be a good candidate to put on staging if anyone wants to take a closer look at the generated HTML.

Yes, please do

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
```diff
@@ -143,9 +143,9 @@ public class PathDescription {
                     + ".\n"
                     + "^//ldml/characters/exemplarCharacters$"
                     + RegexLookup.SEPARATOR
-                    + "Defines the set of characters used in your language. You may not edit or vote on this item at this time. Before filing any tickets to request changes, be sure to read "
-                    + CLDRURLS.EXEMPLAR_CHARACTERS
-                    + ".\n"
+                    + String.format(
```
macchiati, line 146: I don't think this was necessary (filed https://unicode-org.atlassian.net/browse/CLDR-16755) but also not a blocker.

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
```diff
@@ -1,6 +1,7 @@
 /*
  * cldrDeferHelp: encapsulate code related to showing language descriptions in the Info Panel
  */
+import { marked } from "./cldrMarked.mjs";
```
btangmu, line 4: @srl295 this is failing for me locally

ERROR in ./src/esm/cldrDeferHelp.mjs 13:15-21
export 'marked' (imported as 'marked') was not found in './cldrMarked.mjs' (module has no exports)

btangmu: never mind, it's OK with clean rebuild

PR #3042, opened 2023-06-14
CLDR-???? reduce DB opens

srl295:
CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
@@ -351,7 +351,23 @@ public class DBUtils {
         return getInstance().tableExists(table);
     }
 
+    private static final ConcurrentHashMap<String, Boolean> existenceCache =
+            new ConcurrentHashMap<String, Boolean>();
+
     public boolean tableExists(String table) {
+        // once exists, return true
+        // if false, try again.
+        boolean cachedValue = existenceCache.getOrDefault(table, false);
+        if (cachedValue) return cachedValue;
+        cachedValue = _tableExists(table);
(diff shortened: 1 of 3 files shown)


btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
```diff
@@ -351,7 +351,23 @@ public static boolean hasTable(String table) {
...
 
+    private static final ConcurrentHashMap<String, Boolean> existenceCache =
+            new ConcurrentHashMap<String, Boolean>();
+
     public boolean tableExists(String table) {
+        // once exists, return true
+        // if false, try again.
+        boolean cachedValue = existenceCache.getOrDefault(table, false);
+        if (cachedValue) return cachedValue;
+        cachedValue = _tableExists(table);
```
macchiati, line 362: Is this supposed to be thread-safe, because it doesn't appear to be.

srl295: The results only go from false to true, shouldn't ever go back to false… It's OK if two threads actually hit the database to check, that's what's happening right now… This just reduces the calls directly to the database

srl295: In other words, two threads, trying to create the same database, already needed to be synchronized outside of this code

macchiati: Ok, makes sense.

srl295:
Note, this is just a draft with a few ideas and anomalies that I found, not completely baked yet

PR #3046, opened 2023-06-15
CLDR-16757 Core data for bew

srl295:
CLDR-16757

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/bew.xml
+++ b/common/main/bew.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2023 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="bew"/>
+	</identity>
+	<characters>
+		<exemplarCharacters>[a b c d eéèĕ f g h i j k l m n o p q r s t u v w x y z]</exemplarCharacters>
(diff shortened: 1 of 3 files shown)


pedberg-icu (approved):
Looks good. I see we already have a name for it in `en`, an entry in likelySubtags, and territoryInfo data in ID for it.

srl295:
> Looks good. I see we already have a name for it in `en`, an entry in likelySubtags, and territoryInfo data in ID for it.

Yes, this one came ready to go!

macchiati:
I believe this is incomplete; it doesn't have the name of the language bew.

srl295:
> I believe this is incomplete; it doesn't have the name of the language bew.

The English name was already in en.xml, the autonym is not needed in https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales nor in the core data form.  The autonym is needed for Basic.

macchiati:
You're right; we don't require (or want) that anymore in core.

srl295:
> You're right; we don't require (or want) that anymore in core.

Tests ought to verify that all locales are at core. Should we have a Ddl ticket for double checking ?

grcspace311:
> CLDR-16757
> 
> - [ ] This PR completes the ticket.
> 
> ALLOW_MANY_COMMITS=true
> 

>echo [] CLASSPATH=%APP_HOME%\gradle\wrapper\

common/main/bew.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2023 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
```
grcspace311, line 6: > CLASSPATH=%APP_HOME%\gradle\wrapper\ 'add'

grcspace311 (approved):
Grcspace311

PR #3047, opened 2023-06-16
CLDR-16763 Reduce deadlocks

srl295:
CLDR-16763

Initial work:

- use a real cache instead of homegrown
- initialize the non-resolved XMLSource / CLDRFile in the Per-Locale c'tor instead of deferring

I think initializing the resolved XMLSource might be more challenging because it's reentrant, needs to construct other PerLocaleDatas. 

But perhaps, when, say, French or French in Canada is hit, ST should grab a lock  and initialize ALL `fr*` sublocales right there. That way if  say `fr_FR` and `fr_CA` both got hit at the same time, they wouldn't deadlock on adding `fr` . 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/LocaleMaxSizer.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/LocaleMaxSizer.java
@@ -1,0 +1,70 @@
+package org.unicode.cldr.web;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import org.unicode.cldr.util.CLDRLocale;
+
+/**
+ * This class tracks the expected maximum size of strings in the locale.
+ *
+ * @author srl
+ */
+public class LocaleMaxSizer {
+    public static final int EXEMPLAR_CHARACTERS_MAX = 8192;
+
+    public static final String EXEMPLAR_CHARACTERS = "//ldml/characters/exemplarCharacters";
+
+    Map<CLDRLocale, Map<String, Integer>> sizeExceptions;
+
+    TreeMap<String, Integer> exemplars_prefix = new TreeMap<>();
+    Set<CLDRLocale> exemplars_set = new TreeSet<>();
+
+    /** Construct a new sizer. */
+    public LocaleMaxSizer() {
+        // set up the map
+        sizeExceptions = new TreeMap<>();
+        exemplars_prefix.put(EXEMPLAR_CHARACTERS, EXEMPLAR_CHARACTERS_MAX);
+        String[] locs = {"ja", "ko", "zh", "zh_Hant" /*because of cross-script inheritance*/};
+        for (String loc : locs) {
+            exemplars_set.add(CLDRLocale.getInstance(loc));
+        }
+    }
+
+    /**
+     * It's expected that this is called with EVERY locale, so we do not recurse into parents.
+     *
+     * @param l
+     */
+    public void add(CLDRLocale l) {
+        if (l == null) return; // attempt to add null
+        CLDRLocale hnr = l.getHighestNonrootParent();
+        if (hnr == null) return; // Exit if l is root
+        if (exemplars_set.contains(hnr)) { // are we a child of ja, ko, zh?
+            sizeExceptions.put(l, exemplars_prefix);
+        }
+    }
+
+    /**
+     * For the specified locale, what is the expected string size?
+     *
+     * @param locale
+     * @param xpath
+     * @return
+     */
+    public int getSize(CLDRLocale locale, String xpath) {
+        Map<String, Integer> prefixes = sizeExceptions.get(locale);
+        if (prefixes != null) {
+            for (Map.Entry<String, Integer> e : prefixes.entrySet()) {
+                if (xpath.startsWith(e.getKey())) {
+                    return e.getValue();
+                }
+            }
+        }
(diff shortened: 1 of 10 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -1650,10 +1667,20 @@ public BallotBox<User> ballotBoxForLocale(CLDRLocale locale) {
     }
 
     /** Per locale map */
-    private final Map<CLDRLocale, Reference<PerLocaleData>> locales = new HashMap<>();
-
-    private final Cache<CLDRLocale, PerLocaleData> rLocales =
-            CacheBuilder.newBuilder().softValues().build();
+    private final LoadingCache<CLDRLocale, PerLocaleData> locales =
+            CacheBuilder.newBuilder()
+                    .concurrencyLevel(5)
```
srl295, line 1672: OK, i should set softValues  and perhaps some duration timeout.

srl295:
this is looking promising. 
Needs a longer cache time. But the reentrancy seems to be fixed.

macchiati (dismissed):
Looks good to me.

btangmu (dismissed):
I looked through the changes quickly and no problems were obvious to me, however, this is a huge amount of change and I didn't understand a lot of this code in the first place, so basically a rubber-stamp LGTM

srl295:
> I looked through the changes quickly and no problems were obvious to me, however, this is a huge amount of change and I didn't understand a lot of this code in the first place, so basically a rubber-stamp LGTM

It's hopefully more understandable with smaller units and more comments.

srl295:
just one more commit since approval, https://github.com/unicode-org/cldr/pull/3047/commits/d5b9870d42c69c90fc03fd75134d48e798bb4b92 for productization

Pushing this to cldr-staging

tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -908,6 +691,11 @@ public VoteResolver<String> getResolver(String path) {
             return getResolver(peekXpathData(path), path, null);
         }
 
+        @Override
+        public VoteResolver<String> getResolver(String path, VoteResolver<String> r) {
+            return getResolver(peekXpathData(path), path, r);
```
btangmu, line 696: minor quibble: I prefer to put caller before callee -- easier to read with higher-level methods higher in the file

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -2516,36 +2511,19 @@ public void close() {
...
-                resolvedFile = cldrfile;
-            }
         }
 
-        public void clear() {
-            // TODO: try just kicking these instead of clearing?
-            cldrfile = null;
-            dbSource = null;
-        }
+        public void clear() {}
```
btangmu, line 2526: is `clear()` still called anywhere? could just remove it?

btangmu (approved):
lgtm though a lot of it is still mysterious to me

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
srl295: I'm not sure, I didn't add it in this PR

btangmu: no callers, not even from .jsp

srl295: I think UserLocaleStuff was for convenience of JSPs and other very old flows… to make it easier to get the 'current' CLDRFile.  Probably most of the class is obsolete.

srl295:
@btangmu PTAL. Let's discuss merging tomorrow if no issues show up

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/LocaleMaxSizer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.TreeMap;
+import java.util.TreeSet;
+import org.unicode.cldr.util.CLDRLocale;
+
+/**
+ * This class tracks the expected maximum size of strings in the locale.
+ *
+ * @author srl
+ */
+public class LocaleMaxSizer {
```
macchiati, line 14: Question. By maximum size, do you mean the maximum **length** of any element value in the main locale file? It looks like the code assumes that, and also assumes that the longest one is the exemplar value in the CJK locales which is 8192, and that every other locale has a max of 4096. And we don't care that it is not exact numbers. If these are true, would be nice to add to the javadoc. (but not a blocker)

macchiati (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/LocaleMaxSizer.java
srl295: Yes length

srl295: Created a ticket https://unicode-org.atlassian.net/browse/CLDR-16777 to document.

PR #3049, opened 2023-06-17
CLDR-16766 Voting for inheritance, error message, Dashboard disappears

btangmu:
-Take into account that the back end may return multiple rows even when only one row was requested

-Include the requested xpath string ID in the response to avoid confusion

-Increase error checking on the front end to facilitate future debugging; more try/catch and notifications

-Comments; TODO comment for returning only the requested row

CLDR-16766

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
@@ -137,6 +137,13 @@ public class VoteAPIHelper {
                 if (xp == null) {
                     return Response.status(404).entity(new STError(ErrorCode.E_BAD_XPATH)).build();
                 }
+                // TODO: only return data for the single row the front end requested!
+                // This matcher can match "alt" paths related to the given path, so for example,
+                // if the requested path is
+                // //ldml/localeDisplayNames/territories/territory[@type="FK"]
+                // then currently we return not only data for that path, but also
+                // for //ldml/localeDisplayNames/territories/territory[@type="FK"][@alt="variant"]
+                // Reference: https://unicode-org.atlassian.net/browse/CLDR-16766
                 matcher = XPathMatcher.getMatcherForString(xp); // single string
             } else {
                 // Should not get here.
@@ -148,6 +155,9 @@ public class VoteAPIHelper {
             final DataPage pageData = DataPage.make(pageId, mySession, locale, xp, matcher);
             pageData.setUserForVotelist(mySession.user);
             r.page = new RowResponse.Page();
+            if (args.xpstrid != null) {
+                r.setOneRowPath(args.xpstrid);
+            }
 
             // don't return default content
             CLDRLocale dcParent =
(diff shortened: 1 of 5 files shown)


btangmu:
This fixes the bug. I added a TODO comment for making the back end return only one row if only one row was requested, which I'll do with a subsequent PR on the same ticket if that's OK

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
```diff
@@ -137,6 +137,13 @@ private static Response handleGetRows(ArgsForGet args) {
...
                     return Response.status(404).entity(new STError(ErrorCode.E_BAD_XPATH)).build();
                 }
+                // TODO: only return data for the single row the front end requested!
+                // This matcher can match "alt" paths related to the given path, so for example,
+                // if the requested path is
+                // //ldml/localeDisplayNames/territories/territory[@type="FK"]
+                // then currently we return not only data for that path, but also
+                // for //ldml/localeDisplayNames/territories/territory[@type="FK"][@alt="variant"]
+                // Reference: https://unicode-org.atlassian.net/browse/CLDR-16766
                 matcher = XPathMatcher.getMatcherForString(xp); // single string
```
srl295, line 147: I see, it's not an exact match.  Maybe need an exactMatcherForString(). or just add `$` to the end of the xpath?

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
btangmu: exactMatcherForString sounds good -- might be more efficient than a regex ending with $

this is the only place getMatcherForString is called

BaseAndPrefixMatcher is deprecated, and seems not to be used anywhere other than getMatcherForString so it looks like we could replace getMatcherForString with a new exactMatcherForString and remove BaseAndPrefixMatcher

srl295: > BTW any idea what DataPage.CONTINENT_DIVIDER "~" is about? I put a breakpoint here after the "if" and tried voting in all the timzezone paths but I never hit it
> 
> ```
>    public static XPathMatcher getMatcherForString(String str) {
>         if (str.contains(DataPage.CONTINENT_DIVIDER)) {
>             // just return prefix
>             return BaseAndPrefixMatcher.getInstance(
>                     XPathTable.NO_XPATH, str.substring(0, str.indexOf(DataPage.CONTINENT_DIVIDER)));
>         } else {
>             return BaseAndPrefixMatcher.getInstance(XPathTable.NO_XPATH, str);
>         }
>     }
> ```

CONTENT_DIVIDER… isnt' that the rocky mountains?

srl295: OK seriously. getMatcherForString()  was added in 2012 may have had other callers. 

I would suspect that it may have been used before there was a 'fetch by pathheader section' api.  But don't remember exactly.   @macchiati ?

btangmu: ... where the rivers change directions ...

PR #3051, opened 2023-06-20
CLDR-16741 Update languages/codes: World DataBank

btangmu:
-Update the file world_bank_data.csv from databank.worldbank.org/reports.aspx?source=world-development-indicators

-Update the file supplementalData.xml produced by ConvertLanguageData

CLDR-16741

ALLOW_MANY_COMMITS=true


common/supplemental/supplementalData.xml
```diff
@@ -1396,7 +1396,7 @@ XXX Code for transations where no currency is involved
 		<language type="bku" scripts="Latn"/>
 		<language type="bku" scripts="Buhd" alt="secondary"/>
 		<language type="bla" scripts="Latn"/>
-		<language type="blo" scripts="Latn" territories="BJ TG"/>
```
srl295, line 1399: this may have been added, manually, since Anii (blo) is in CLDR. Was there any warning when you ran the generator?

btangmu: Yes, as described in the ticket:
```
# Problems in language_script.tsv

* DETAIL *	changing <languageData>:	blo	Anii	[BJ:Benin]  REMOVED!
* DETAIL *	changing <languageData>:	blo	Anii	[TG:Togo]  REMOVED!
* DETAIL *	changing <languageData>:	blo	Anii	{primary=[primary; scripts=Latn; territories=]}
```

macchiati: That shouldn't cause a problem. blo is still present.

macchiati (approved):

DavidLRowe:
I guess since all the tests are passing, that language element for "blo" isn't needed (and presumably will reappear when the generated data is redone later in the cycle). AFAIK it seems okay.

PR #3052, opened 2023-06-20
CLDR-10985 add Tuvan tyv

srl295:
CLDR-10985

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/tyv.xml
+++ b/common/main/tyv.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2023 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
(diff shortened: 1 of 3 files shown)


DavidLRowe:
Are we sure about the exemplar characters?

I would have expected
- `а б` (U+0430 U+0431), rather than `{\[б}` at the beginning, and
- `я` (U+44F), rather than `{я|а\]\]}` at the end

It seems that the ends of the main exemplar got scrambled a bit.

Everything else looks fine.

srl295:
> Are we sure about the exemplar characters?
> 
> I would have expected
> 
> * `а б` (U+0430 U+0431), rather than `{\[б}` at the beginning, and
> * `я` (U+44F), rather than `{я|а\]\]}` at the end
> 
> It seems that the ends of the main exemplar got scrambled a bit.
> 
> Everything else looks fine.

thanks, yes it had an extra [] around it.

srl295:
fyi @ftyers

pedberg-icu (approved):

PR #3053, opened 2023-06-20
CLDR-16758 fix 'import old votes'

srl295:
CLDR-16758

ALLOW_MANY_COMMITS=true


btangmu (approved):

srl295:
fixed version 

image

btangmu (approved):
excellent!

pedberg-icu (approved):

macchiati:
Great. Sounds like people don't have to re-import

btangmu:
@macchiati there's also a major bug in auto-import, so after a fix is committed (I should have it ready shortly), I think we'll need to reset everyone's auto-import flag, causing auto-import to run again the next time they log in

PR #3054, opened 2023-06-20
CLDR-16758 Too many abstained votes; fix bug in import votes

btangmu:
-Move equalsOrInheritsCurrentValue from XMLSource to CLDRFile

-The callers had non-resolving XMLSource for which getBaileyValue is always null

CLDR-16758

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
@@ -7,6 +7,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.logging.Logger;
 import org.json.JSONException;
+import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.XMLSource;
 
 public class SurveyBulkClosePosts {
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
```diff
@@ -110,10 +111,12 @@ private void updateList(ResultSet rs) throws SQLException {
     }
 
     private boolean matchesWinning(String loc, Integer xpath, String value) {
-        XMLSource diskData = sm.getDiskFactory().makeSource(loc).freeze();
+        STFactory stf = sm.getSTFactory();
+        XMLSource diskData = stf.makeSource(loc).freeze();
```
srl295, line 115: This isn't diskData! This will give live (vetted) data 

Also  I think it's wrong to call freeze() on the STF's source, that will freeze out other callers—why?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
```diff
@@ -110,10 +111,12 @@ private void updateList(ResultSet rs) throws SQLException {
...
         String xpathString = sm.xpt.getById(xpath);
         String curValue = diskData.getValueAtDPath(xpathString);
-        return diskData.equalsOrInheritsCurrentValue(value, curValue, xpathString);
+        CLDRFile cldrFile = stf.make(loc, true, true);
+        return cldrFile.equalsOrInheritsCurrentValue(value, curValue, xpathString);
```
srl295, line 119: as above, will use vetted data not disk data

srl295 (commented):
you can just call `make(localeId, /* resolved */ true)` on any of the DiskFactories

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyBulkClosePosts.java
srl295: `sm.getDiskFactory().make(loc, true)` will give you a resolved CLDRFile

srl295: the disk files ought to be be pre-frozen, not sure i they are or not

PR #3055, opened 2023-06-21
CLDR-16758 Too many abstained votes; fix bug in import votes

btangmu:
-Move equalsOrInheritsCurrentValue from XMLSource to CLDRFile

-The callers had non-resolving XMLSource for which getBaileyValue is always null

-Make sure CLDRFile.equalsOrInheritsCurrentValue is called with a CLDRFile made using getDiskFactory

-New method CLDRFile.getResolvingDataSource

-In addRowToImportTable, do not throw exception if row already exists

-Fix warning, ServletException is never thrown

-Fix warnings for invalid @param comments

CLDR-16758

ALLOW_MANY_COMMITS=true


btangmu:
Note: I tried a different strategy, just making sure to use a XMLSource.ResolvingDataSource, not a plain XMLSource, for equalsOrInheritsCurrentValue. That strategy, however, was unsuccessful, since equalsOrInheritsCurrentValue calls getBaileyValue, and ResolvingDataSource.getBaileyValue very often returns null where CLDRFile.getBaileyValue returns non-null (such as for some lateral inheritance). As a consequence, far fewer votes were auto-imported that way, even though the vetter had voted for inheritance and the inherited value (per CLDRFile.getBaileyValue) was winning. The conclusion is that ResolvingDataSource.getBaileyValue is not a reliable substitute for CLDRFile.getBaileyValue -- possibly its name should be changed to avoid confusion.

btangmu:
Error summary:
CLDR/LikelySubtagsTest/testGetResolvedScriptVsExemplars
?

macchiati:
To unblock, copy the "got" value into the test case for Arab. I'll run down the reason for the problem later

macchiati:
Or just disable the test... They are only extra sample examples.

srl295 (approved):

macchiati (approved):

PR #3057, opened 2023-06-23
CLDR-16787 exclude typographicNames from same-as-code check

srl295:
CLDR-16787

ALLOW_MANY_COMMITS=true


srl295:
image

macchiati (approved):

PR #3058, opened 2023-06-23
CLDR-16786 make languages in CLDR locales' ids available

macchiati:
CLDR-16786

Note: this adds the missing languages (such as aa and osa). I left some duplication in the attributeValueValidity file to make review easier (while ensuring that nothing else changes).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
@@ -1,27 +1,38 @@
 package org.unicode.cldr.unittest;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
+import com.ibm.icu.impl.Row;
+import com.ibm.icu.impl.Row.R4;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 import org.unicode.cldr.test.CoverageLevel2;
+import org.unicode.cldr.tool.LikelySubtags;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.CoreCoverageInfo;
 import org.unicode.cldr.util.CoreCoverageInfo.CoreItems;
 import org.unicode.cldr.util.Factory;
+import org.unicode.cldr.util.LanguageTagParser;
 import org.unicode.cldr.util.Level;
 import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.Organization;
 import org.unicode.cldr.util.PathHeader;
 import org.unicode.cldr.util.StandardCodes;
 import org.unicode.cldr.util.SupplementalDataInfo;
+import org.unicode.cldr.util.SupplementalDataInfo.PopulationData;
 import org.unicode.cldr.util.XMLSource;
+import org.unicode.cldr.util.XPathParts;
 
 public class TestCoverage extends TestFmwkPlus {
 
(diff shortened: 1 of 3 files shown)


macchiati:
PS, I'm not going to have time to do any substantial work on this today. If the review is ok, please merge.

srl295 (commented):
If i understand, this *will* fail if a locale is added without updating attributeValueValidity, so it's not the 'dynamic' aspect mentioned in the ticket.  SO we can think of this as an interim step perhaps?

macchiati:
It does add an extra step for new locales, so we want to look at automating
later (and some of the code could be re-purposed for that end). But I don't
want to make that change now.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
```diff
@@ -157,10 +169,245 @@ private String showColumn(Set items) {
...
+                    + stringForm(languageLevel)
+                    + "\t"
+                    + stringForm(level)
+                    + "\t"
+                    + displayName;
+        }
+
+        private String stringForm(Level level2) {
+            if (level == null) {
+                return "υnd";
```
srl295, line 215:
```suggestion
                level = Level.UNDETERMINED;
```

maybe this should go on the Level object though?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
```diff
@@ -157,10 +169,245 @@ private String showColumn(Set items) {
...
+                default:
+                    return level2.toString();
+            }
+        }
+    }
+
+    public void testLSR() {
+        SupplementalDataInfo supplementalData = testInfo.getSupplementalDataInfo();
+        Factory factory = testInfo.getCldrFactory();
+        CLDRFile root = factory.make("root", true);
```
srl295, line 231:
```suggestion
        CLDRFile root = factory.make(LocaleNames.ROOT, true);
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
```diff
@@ -157,10 +169,245 @@ private String showColumn(Set items) {
...
+        CoverageLevel2 coverageLevel =
+                CoverageLevel2.getInstance(supplementalData, "yyy"); // non-existant locale
```
srl295, line 233:
```suggestion
                CoverageLevel2.getInstance(supplementalData, "qtz"); // non-existant locale
```

should this use a private use code instead?

srl295 (approved):
all minor comments

srl295:
```
Error:  (TestCoverageLevel.java:795)  Error: Comprehensive & no exception for path =>	//ldml/localeDisplayNames/languages/language[@type="skr"]

Error:  (TestCoverageLevel.java:795)  Error: Comprehensive & no exception for path =>	//ldml/localeDisplayNames/languages/language[@type="blo"]

Error:  (TestCoverageLevel.java:795)  Error: Comprehensive & no exception for path =>	//ldml/localeDisplayNames/languages/language[@type="xnr"]

Error:  (TestCoverageLevel.java:795)  Error: Comprehensive & no exception for path =>	//ldml/localeDisplayNames/languages/language[@type="vmw"]

Error:  (TestCoverageLevel.java:795)  Error: Comprehensive & no exception for path =>	//ldml/localeDisplayNames/languages/language[@type="kxv"]
```

so it's "working" and notes that kxv, vmw, xnr, blo, skr should be added to comprehensive at least (if not better)

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
macchiati: Normally we don't need it (a null indicates an anomaly), so probably best to keep just in the test.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
macchiati: ok

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
macchiati: ok

macchiati:
BTW, moved the test into a different file. See https://unicode-org.atlassian.net/browse/CLDR-13465

srl295 (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
```diff
@@ -1077,4 +1086,239 @@ private Multimap<String, String> delta(Set<String> expected, Set<String> groupin
...
+        }
+        Level old = codeToBestLevel.get(code);
+        if (old == null) {
+            codeToBestLevel.put(code, level);
+        } else if (level.compareTo(old) > 0) {
+            codeToBestLevel.put(code, level);
+        } else if (level != old) {
+            int debug = 0;
+        }
+    }
```
srl295, line 1323: `if(code.isEmpty()) return;` to avoid having to call `regions.remove("")` later

PR #3060, opened 2023-06-27
CLDR-16795 Add Zhuang (za) locale

Dringsim:
CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/za.xml
+++ b/common/main/za.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="za"/>
+	</identity>
+	<characters>
+		<exemplarCharacters>[a b c d e f g h i j k l m n o p q r s t u v w x y z]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[]</exemplarCharacters>
+		<exemplarCharacters type="index">[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z]</exemplarCharacters>
+		<exemplarCharacters type="numbers">[\- ‑ , . % ‰ + 0 1 2 3 4 5 6 7 8 9]</exemplarCharacters>
+		<exemplarCharacters type="punctuation">[\- ‐‑ – — , ; \: ! ? . … '‘’ &quot;“” ( ) \[ \] § @ * / \&amp; # † ‡ ′ ″]</exemplarCharacters>
+	</characters>
+</ldml>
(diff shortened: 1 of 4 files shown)


DavidLRowe:
@Dringsim Apparently adding `za` to genderList.xml is causing a test failure. Could you try removing that change from the pull request? After the locale is successfully added, we can see what else is needed in the way of supplemental data.

DavidLRowe:
Also, the normal workflow would not include adding `za` to AttributeValueValidity.xml. (That would occur later if/when the locale is added to the list by the CLDR TC.) Could you remove `za` from that file?

Dringsim:
Done.

DavidLRowe:
Still has error:
```
Error:  (TestCoverageLevel.java:1285)  Error: : Language subtags that are in a CLDR locale's ID are in root (1): expected "", got "za"
```
@srl295 Any clues on why this is failing?

macchiati:
To catch the cases where a language is added but doesn't appear at the comprehensive level, za needs to be added to AttributeValueValidity.xml. 
It shouldn't be added to genderList.xml until it reaches Basic.

srl295:
> To catch the cases where a language is added but doesn't appear at the comprehensive level, za needs to be added to AttributeValueValidity.xml. 

Right, until it can be added automatically without needing a file change.

We should improve the message to clarify the situation also. 

> It shouldn't be added to genderList.xml until it reaches Basic.

DavidLRowe:
@Dringsim My apologies for asking you to change the AttributeValueValidity.xml file.

DavidLRowe (approved):

DavidLRowe:
The Jira ticket has been Accepted and all the tests pass. The only problem is that the commit message doesn't start with the Jira ticket number. If you click on "Details" link opposite the jira-ticket test:
image
then use the "Squash branch" button to set the commit message to `CLDR-16795 Add Zhuang (za) locale`. 
You'll need to tick the box and "Perform squash".

Once the commit is correct and the tests all pass, we can merge the PR and get the new locale into the system. At that point you can request a Survey Tool account using https://corp.unicode.org/reporting/cldr-id.html and specifying `za` for the "Language(s)" field. Unfortunately there is less than a week remaining in this submission cycle. Hopefully you'll be able to submit the data for Basic level (approximately 70 items) during that time.

Thanks for your contribution!

Dringsim:
Done.
I've requested a Survey Tool account before (dringsim [at] qq.com). Can I use that account to submit the data for `za`?

DavidLRowe:
@rickmcgowan When the next push to production happens, this locale (za) should be available. Can you then add `za` to the list of authorized locales for @Dringsim ? Let me know if there's anything else needed.

DavidLRowe:
@rickmcgowan With the push to production that Steven just made, `za` is now available. Can you update Dringsim please?

PR #3061, opened 2023-06-27
CLDR-16313 update Report API to include vote results

srl295:
- update some VoteAPIs as well to be reusable

CLDR-16313

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/ReportAPI.java
@@ -6,6 +6,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.HeaderParam;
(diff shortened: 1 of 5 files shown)


pedberg-icu (dismissed):

srl295:
OK it looks like I missed a formatter step on the last fix

srl295:
ready for review

pedberg-icu (approved):

macchiati:
Can you capture a static HTML page with the results of this, with the current ST data, so we can see what it would look like.

srl295:
> Can you capture a static HTML page with the results of this, with the current ST data, so we can see what it would look like.

There's no HTML, this is only the API, the JSON is given above.

Here's an example from smoketest for `de`

```json
        {
          "acceptability": "notAcceptable",
          "notAcceptableScore": 50,
          "report": "personnames",
          "resolvedVoteCounts": {
            "notAcceptable": 50
          },
          "status": "approved",
          "votersForAcceptable": 0,
          "votersForNotAcceptable": 1,
          "votes": {
            "2869": "notAcceptable"
          },
          "votingResults": {
            "nameTime": {
              "#2869": 1687891825000
            },
            "orgs": {
              "unaffiliated": {
                "conflicted": false,
                "orgVote": "notAcceptable",
                "status": "ok",
                "votes": {
                  "notAcceptable": 50
                }
              }
            },
            "requiredVotes": 4,
            "valueIsLocked": false,
            "value_vote": [
              "notAcceptable",
              "50"
            ]
          }
        },
```

macchiati:
I'm confused. I thought the goal was to be able to get the information like in https://unicode.org/cldr/charts/43/supplemental/locale_coverage.html, but for a single locale, and using ST data rather than XML data.

I don't see anything in the above that corresponds to, or can be used to build the data in, https://unicode.org/cldr/charts/43/supplemental/locale_coverage.html#af; for example

srl295:
> I'm confused. I thought the goal was to be able to get the information like in https://unicode.org/cldr/charts/43/supplemental/locale_coverage.html, but for a single locale, and using ST data rather than XML data.
> 
> 
> 
> I don't see anything in the above that corresponds to, or can be used to build the data in, https://unicode.org/cldr/charts/43/supplemental/locale_coverage.html#af; for example

I think you're thinking of a different ticket. This PR is on a subtask of https://unicode-org.atlassian.net/browse/CLDR-16297 "reports need to work more like other votes"

srl295:
I just filed https://unicode-org.atlassian.net/browse/CLDR-16807 for the issue you mentioned mark

PR #3062, opened 2023-06-27
CLDR-16797 Fix deadlock in the CLDR Tools library

mihnita:
CLDR-16797

ALLOW_MANY_COMMITS=true


srl295 (changes requested):
please see https://github.com/unicode-org/icu/pull/2502#issuecomment-1610323474 instead

PR #3066, opened 2023-06-30
CLDR-16815 fix vote-for-inheritance showing up

srl295:
- CandidateItem.rawValue (sic!) can contain INHERITANCE_MARKER, thus failing to retrieve a list of voters.
- Probably only affects UI. It was showing the vote result for inheritance but no voters.
- Also note that the vote transcript was still correct in this case.

CLDR-16815

ALLOW_MANY_COMMITS=true


srl295:
Before:
image


After:
image

macchiati (approved):

PR #3068, opened 2023-06-30
CLDR-16816 Apple should cover modern for mi

stenshamn:
CLDR-16816

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
@@ -282,7 +282,7 @@ Apple	;	kn	;	modern
 Apple	;	ko	;	modern
 Apple	;	lt	;	modern
 Apple	;	lv	;	modern
-Apple	;	mi	;	basic
+Apple	;	mi	;	modern
 Apple	;	mk	;	modern
 Apple	;	ml	;	modern
 Apple	;	mr	;	modern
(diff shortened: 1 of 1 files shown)


AEApple (dismissed):

srl295 (dismissed):
Change LGTM, but SMH on the err

> _Error:  (TestCLDRLocaleCoverage.java:476)  Error: : cldr level = max for mi (Māori): expected org.unicode.cldr.util.Level<basic>, got org.unicode.cldr.util.Level<modern>_

Too high, Māori! Reduce it!

AEApple:
> Error: (TestCLDRLocaleCoverage.java:476) Error: : cldr level = max for mi (Māori): expected org.unicode.cldr.util.Level, got org.unicode.cldr.util.Level

Does that mean the coverage level in the CLDR org also needs to be updated?

srl295:
Be less cryptic anyone? 

I THINK it's complaining about this line:

```
Cldr  ; mi  ; basic
```

Something like:  "the `Cldr` line is the maximum of `apple`, `google`, `microsoft`, and  `special`" (yes the code is that specific)


Probably changing Cldr to `modern` also could fix it

Please file a ticket for the error message, it should explain what it's looking for and what to change.

stenshamn:
Do I need to change the `Cldr` level too or are we OK?

AEApple:
That's super cryptic. Could we update the error message if that's what it turns out to be?

srl295:
> Do I need to change the `Cldr` level too or are we OK?

needs change

AEApple (approved):

stenshamn:
Filed https://unicode-org.atlassian.net/browse/CLDR-16817 for the error message.

PR #3075, opened 2023-07-01
CLDR-16815 fix vote-for-inheritance showing up again

srl295:
- need to include both triple up arrow (soft) AND explicit (hard) votes

CLDR-16815

URLS for testing:

with this PR, these should both show users under the info panel voting for all values

https://cldr-smoke.unicode.org/cldr-apps/v#/mt//6e373b70f56b2bde (now broken!)
https://cldr-smoke.unicode.org/cldr-apps/v#/tk//e8ba06720b375aa (now fixed!)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -228,15 +228,31 @@ public class DataPage {
              * @return the set of votes
              */
             public Set<UserRegistry.User> getVotes() {
-                if (!checkedVotes) {
-                    String valueToCheck = rawValue;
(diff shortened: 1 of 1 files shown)


macchiati (approved):

srl295:
Thanks. Maybe push this Monday if smoke looks ok

macchiati:
Or beforehand, since otherwise the visible vote counts are wrong for Monday
for most of the world

srl295:
The visible counts are correct. It just doesn’t show who voted for
inheritance in some casss unless you show the transcript. If smoke looks ok
feel free to push.

srl295:
I’m also not sure there aren’t other latent bugs here since a common case
was already broken.

PR #3082, opened 2023-07-03
CLDR-16815 fix voter info

srl295:
- revert VoteEntry to just have userid

CLDR-16815

ALLOW_MANY_COMMITS=true


stenshamn:
FWIW, personally I think it's fine to display the org votes publicly, just not individual names and email addresses  (can't remember what we decided on that - maybe for smaller orgs that would be practically the same thing).

srl295:
it's supposed to show the org votes just not any other info.

srl295:
image

^ when logged out

btangmu (approved):

PR #3083, opened 2023-07-05
CLDR-16811 fix unicode set variables

macchiati:
CLDR-16811

The main change is to add variables as an optional part of the semantics. That is reflected in the EBNF table, and in a new section below that. In so doing, a few other clarifications were included, such as the syntax and semantics of '$', and that [: is an error if not terminated with :], but that : can otherwise occur within a set without escaping.

Following https://sembr.org/, also added linebreaks to paragraphs touched in this section.

ALLOW_MANY_COMMITS=true


macchiati:
Done by other PR

PR #3085, opened 2023-07-06
CLDR-16717 Fix a few more issues in the Persian to IPA transform

roozbehp:
CLDR-16717 Fix a few more issues in the Persian to IPA transform

ALLOW_MANY_COMMITS=true


srl295 (approved):
≈

PR #3086, opened 2023-07-07
CLDR-16841 add examples for units

macchiati:
CLDR-16841

- Adds the example to unit/unitPattern and /displayName
- The related units are fetched from the UnitConverter, and displayed with each separate example.
- The format is like **1 ounce-troy (US/UK) 🟰 ~82.943 fun (JP)**, so it shows whether a unit is in a country-system (UK, US, or JP), and what the equivalent amount would be in a related unit

- Add a shorter, simpler way to show units. 
- currently if numerator or denominator are < 20, shows as fraction (or just numerator, if denominator is 1)
- Uses 5 significant digits, and adds ~ if the displayed value is approximate.

- Adds some utilities
- Displays the country systems
- getRelatedExamples picks some samples. Currently
  - gets all the simple units that are interconvertable with the source unit
  - adds a few others, and removes some unusual ones
  - always adds metric
  - checks all the systems (except si, and except jp outside of Japanese) to find the closest units, above and below, where possible. Skips units that are far away (current > 10000x or < 1/10000x)
  - skips some more unusual units
  - removes the source unit, if it got added.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
- fixes current tests, and adds a few more

- tests the getRelatedExamples codes. Currently the only test is to make sure they don't throw exceptions

Current limitations
- It just handles the simple units, so not, say, square-kilometer
- When it doesn't handle a unit, it should still probably show an example if it is in a country-system
- getExampleUnitSystems should probably be moved to the UnitConverter
- It might be showing too many units, or not the best ones, but I didn't want to spend too much time tuning right now.
- FYI, just filed the related (but independent) https://unicode-org.atlassian.net/browse/CLDR-16851
- Especially if we add more units, we probably want to cache the related units. Right now it filters by locale, but that could be done at runtime.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -2,7 +2,6 @@ package org.unicode.cldr.test;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Sets;
 import com.ibm.icu.impl.Row.R3;
 import com.ibm.icu.impl.Utility;
 import com.ibm.icu.impl.number.DecimalQuantity;
@@ -1232,8 +1231,10 @@ public class ExampleGenerator {
         }
         // add related units
         Map<Rational, String> relatedUnits =
-                UNIT_CONVERTER.getRelatedExamples(shortUnitId, getExampleUnitSystems());
+                UNIT_CONVERTER.getRelatedExamples(
+                        shortUnitId, UnitConverter.getExampleUnitSystems(cldrFile.getLocaleID()));
         String unitSystem = null;
+        boolean first = true;
         for (Entry<Rational, String> relatedUnitInfo : relatedUnits.entrySet()) {
             if (unitSystem == null) {
                 Set<UnitSystem> systems = UNIT_CONVERTER.getSystemsEnum(shortUnitId);
@@ -1242,12 +1243,23 @@ public class ExampleGenerator {
             Rational relatedValue = relatedUnitInfo.getKey();
             String relatedUnit = relatedUnitInfo.getValue();
             Set<UnitSystem> systems = UNIT_CONVERTER.getSystemsEnum(relatedUnit);
+            String relation = "≡";
+            String relatedValueDisplay = relatedValue.toString(FormatStyle.basic);
+            if (relatedValueDisplay.startsWith("~")) {
+                relation = "≈";
+                relatedValueDisplay = relatedValueDisplay.substring(1);
+            }
+            if (!examples.isEmpty() && first) {
+                examples.add(""); // add blank line
+                first = false;
+            }
             examples.add(
                     String.format(
-                            "1 %s%s 🟰 %s %s%s",
+                            "1 %s%s %s %s %s%s",
                             shortUnitId,
                             unitSystem,
-                            relatedValue.toString(FormatStyle.basic),
+                            relation,
+                            relatedValueDisplay,
                             relatedUnit,
                             UnitSystem.getSystemsDisplay(systems)));
         }
@@ -1337,18 +1349,6 @@ public class ExampleGenerator {
                 }
             }
         }
-    }
-
-    /**
-     * Customize the systems according to the locale
-     *
-     * @return
-     */
-    public Set<UnitSystem> getExampleUnitSystems() {
-        LanguageTagParser ltp = new LanguageTagParser().set(cldrFile.getLocaleID());
-        return ltp.getLanguage().equals("ja")
-                ? UnitSystem.ALL
-                : Sets.difference(UnitSystem.ALL, Set.of(UnitSystem.jpsystem));
     }
 
     private String getConstrastingCase(
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -2,6 +2,7 @@ package org.unicode.cldr.test;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Sets;
 import com.ibm.icu.impl.Row.R3;
 import com.ibm.icu.impl.Utility;
 import com.ibm.icu.impl.number.DecimalQuantity;
@@ -70,6 +71,8 @@ import org.unicode.cldr.util.Level;
 import org.unicode.cldr.util.PathDescription;
 import org.unicode.cldr.util.PatternCache;
 import org.unicode.cldr.util.PluralSamples;
+import org.unicode.cldr.util.Rational;
+import org.unicode.cldr.util.Rational.FormatStyle;
 import org.unicode.cldr.util.ScriptToExemplars;
 import org.unicode.cldr.util.SimpleUnicodeSetFormatter;
 import org.unicode.cldr.util.SupplementalDataInfo;
@@ -78,6 +81,7 @@ import org.unicode.cldr.util.SupplementalDataInfo.PluralInfo.Count;
 import org.unicode.cldr.util.SupplementalDataInfo.PluralType;
 import org.unicode.cldr.util.TransliteratorUtilities;
 import org.unicode.cldr.util.UnitConverter;
+import org.unicode.cldr.util.UnitConverter.UnitSystem;
 import org.unicode.cldr.util.Units;
 import org.unicode.cldr.util.XListFormatter.ListTypeLength;
 import org.unicode.cldr.util.XPathParts;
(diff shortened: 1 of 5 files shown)


macchiati:
Here's what the examples look like:
Screenshot 2023-07-07 at 13 15 24

macchiati:
@AEApple @stenshamn , included you not so much for the code, but for feedback on the format/information in the examples. (Although we might have to postpone any extensive changes to next release, because of timing.)

stenshamn:
Overall I like it. 
I think it could be better if we visually separated the first, localized line (that shows an example of what the localized format would look like) from the other, non-localized lines (that shows conversion examples to the vetter knows what unit this is). Otherwise it looks a bit jumbled and confusing ("why aren't the other values localized?")
Also, I don't know if we need conversion to five other units, seems a bit much?

srl295:
UI comment: maybe this should go in the info panel, not in the examples. It's more of information about the row than an example of the value.

srl295:
would it be better as

> 1 ounce-troy (US/UK) ≈ 82.943 fun (P)

?

stenshamn:
> UI comment: maybe this should go in the info panel, not in the examples. It's more of information about the row than an example of the value. that is, in PathDescription.

Oh, I thought this _was_ for the info panel...

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1294,7 +1337,18 @@ private String handleFormatUnit(XPathParts parts, String unitPattern) {
...
-        return formatExampleList(examples);
+    }
+
+    /**
+     * Customize the systems according to the locale
+     *
+     * @return
+     */
+    public Set<UnitSystem> getExampleUnitSystems() {
+        LanguageTagParser ltp = new LanguageTagParser().set(cldrFile.getLocaleID());
```
srl295, line 1348: could just be `CLDRLocale.getInstance(…).getLanguage()` but that's another discussion

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1294,7 +1337,18 @@ private String handleFormatUnit(XPathParts parts, String unitPattern) {
...
+        return ltp.getLanguage().equals("ja")
+                ? UnitSystem.ALL
+                : Sets.difference(UnitSystem.ALL, Set.of(UnitSystem.jpsystem));
```
srl295, line 1351: seems like this should be territory based ,not language based.  So if the maximal id's territory is `JP`, etc.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
```diff
@@ -334,10 +336,29 @@ public String toString() {
...
+                    return nf.format(numerator) + "/" + nf.format(denominator);
+                } else {
+                    String result =
+                            nf2.format(numerator.doubleValue() / denominator.doubleValue())
+                                    .toString();
+                    Rational roundtrip = Rational.of(result);
+                    if (roundtrip.equals(this)) {
+                        return result;
+                    }
+                    return "~" + result;
```
srl295, line 359: ah ok, the `~` comes from here.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
```diff
@@ -350,8 +371,7 @@ public String toString(FormatStyle style) {
             case repeating:
                 limit = 30;
             case repeatingAll:
-                String result =
-                        toRepeating(limit); // limit of 30 on the repeating length, so we don't get
+                String result = toRepeating(limit);
                 // unreasonable
```
srl295, line 375:
```suggestion
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -1617,6 +1620,26 @@ public static Set<UnitSystem> fromStringCollection(Collection<String> stringUnit
...
+                    case ussystem:
+                        result.add("US");
+                        break;
+                    case uksystem:
+                        result.add("UK");
+                        break;
+                    case jpsystem:
+                        result.add("JP");
+                        break;
+                }
```
srl295, line 1639: don't we already have localization for this field?

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -1884,4 +1907,118 @@ public Multimap<String, Continuation> getContinuations() {
...
+                        "dunam",
+                        "rod",
+                        "chain",
+                        "furlong",
+                        "fortnight",
+                        "rankine",
+                        "kelvin",
+                        "calorie-it",
+                        "british-thermal-unit-it",
+                        "foodcalorie"));
```
srl295, line 1950: should be driven by preferences? or other metadata?

srl295 (approved):
code lgtm, ui comments…

srl295:
> > UI comment: maybe this should go in the info panel, not in the examples. It's more of information about the row than an example of the value. that is, in PathDescription.
> 
> Oh, I thought this _was_ for the info panel...

It's in the info panel and mouse-over on each item.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Yup.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Except that the choice of which units to even show for translation are language-based, not region based. Eg outside of Japan, people aren't really going to care much about 1.5 寸

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
macchiati: right

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
macchiati: Will do.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: These and the unit equalities are unlocalized (like the Code field)

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: We don't have any of that information available.

Remember, these are just examples to help establish the meaning of the current unit, so the goal was to quickly eliminate units that would not be very familiar to lots of people. I did, however, notice a problem above with bushel, bushel, barrel so will fix.

macchiati:
General issue:  with more time we could do it, but we don't have much time. So I'm looking for small (in terms of dev time) changes that would make a signficant impace.

> I think it could be better if we visually separated the first, localized line (that shows an example of what the localized format would look like) from the other, non-localized lines (that shows conversion examples to the vetter knows what unit this is). Otherwise it looks a bit jumbled and confusing ("why aren't the other values localized?")

The easiest is to leave a blank line (that would be 5 minutes); second would be to have a different background (somewhat more time).

> Also, I don't know if we need conversion to five other units, seems a bit much?

I thought of that, but it is harder to winnow the number down without missing some important comparisons. One possible easy change is if it isn't an Anglophone locale we could skip comparisons to UK specific measures. 

> UI comment: maybe this should go in the info panel, not in the examples. It's more of information about the row than an example of the value. that is, in PathDescription.

I think we generally describe the entire left panel as the Info Panel. I think you mean the blue pane at the top. Yes, the information could go there, but not without a lot of restructuring.

> 1 ounce-troy (US/UK) ≈ 82.943 fun (P)

I could relatively easily make the following change if we think it is worth it.

🟰 ~XXX 
to 
≈ XXX

probably also change for consistency:

🟰 XXX
to
≡ XXX

I was going with the emoji 🟰 just to set it off more from the translation examples, but sounds like that was of little value.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
srl295: > language based, not region based
> outside of Japan, people aren't really going to care much about 1.5 寸

somehow this doesn't compute… isn't Japan a region?

srl295:
> Blue panel … not without a lot of restructuring

it's just the PathDescription class.  Anyway, this will show up in the info panel AND on mouseover of every value.  Maybe that's OK


high bit: this brings value. I think we should move forward with it and iterate.

srl295 (approved):

macchiati:
Here is the screenshot after changes:

Screenshot 2023-07-07 at 16 52 35

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Sorry, I meant "outside of Japanese, people aren't really going to care much about 1.5 寸"

PR #3088, opened 2023-07-10
CLDR-16857 add SLE to 80, move SLL to 100

srl295:
- coverage

CLDR-16857

ALLOW_MANY_COMMITS=true


stenshamn (approved):
Looks right to me.

AEApple:
Can we add a test that requires coverage to be added or explicitly opted out whenever we add a new currency?

srl295:
@AEApple did just that

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestLevel.java
```diff
@@ -47,4 +55,49 @@ void testMax(final String a, final String b, final String c) {
...
+    }
+
+    @ParameterizedTest()
+    @MethodSource("modernCurrencies")
+    public void testModernCurrencies(final String code) {
+        Level l =
+                sdi.getCoverageLevel(
+                        String.format(
+                                "//ldml/numbers/currencies/currency[@type=\"%s\"]/symbol", code),
+                        "mt");
```
srl295, line 97: proxy here because a language needs to be added, maybe it should be `und`?

pedberg-icu (approved):
Looks good to me

PR #3089, opened 2023-07-11
CLDR-16811 Update UTS35 UnicodeSet syntax

skius:
[CLDR-16811](https://unicode-org.atlassian.net/browse/CLDR-16811)

This PR combines the work of #3083 and skius/cldr#1.

There are still questions where a conclusion needs to be reached, I have opened comments in this PR referencing them:
1. https://github.com/unicode-org/cldr/pull/3089/files#r1259921837
2. https://github.com/unicode-org/cldr/pull/3089/files#r1259924343
3. https://github.com/unicode-org/cldr/pull/3089/files#r1259927642

It is related to multiple CLDR issues:
* [CLDR-16811](https://unicode-org.atlassian.net/browse/CLDR-16811) - Regularize use of variables in UnicodeSets
* [CLDR-16772](https://unicode-org.atlassian.net/browse/CLDR-16772) - Bring the syntax of the specification up to speed with changes in ICU and resolve ambiguities
* [CLDR-16852](https://unicode-org.atlassian.net/browse/CLDR-16852) - ~`propName`~ `charName` for `\N` escapes

A summary of the changes:
1. Allow empty UnicodeSets and empty strings
4. Add `$variable` support, backwards-incompatibly demoting `$` from being a valid `char` in all cases 
5. Specify the meaning of multi-codepoint-escapes (`\x{61 62}`)
6. Clarify validity of string ranges (`{ax}-{bz}`)
7. Reflecting ICU supported syntax in the spec (e.g., `[^-]` being valid, separate syntax characters inside strings)
8. Explicitly defining parse behavior for the ambiguous POSIX property sets
9. Adding examples for especially confusing syntax edge cases

Required changes in ICU to reflect the new syntax:
1. Variable replacing must be done with the new 'typed' variable map. The current replacement mechanism works equivalently on CLDR data, so this won't become an immediate compatibility issue with CLDR.
2. Spaces before `^` in negation syntax are supported in ICU, not anymore in the new syntax. CLDR data is again unaffected.
3. Switch variables from IDS/IDC to XIDS/XIDC
4. ... potentially more, needs more investigation.

ALLOW_MANY_COMMITS=true

(cc @eggrobin @robertbastian @younies)

[CLDR-16811]: https://unicode-org.atlassian.net/browse/CLDR-16811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDR-16811]: https://unicode-org.atlassian.net/browse/CLDR-16811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDR-16772]: https://unicode-org.atlassian.net/browse/CLDR-16772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDR-16852]: https://unicode-org.atlassian.net/browse/CLDR-16852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `quoted`       | <pre>= 'u' \(hex\{4\} \| bracketedHex\)<br/>\| 'x' \(hex\{2\} \| bracketedHex\)<br/>\| 'U00' \('0' hex\{5\} \| '10' hex\{4\}\)<br/>\| 'N\{' charName '\}'<br/>\| \[\[\\u0000\-\\U00010FFFF\]\-\[uxUN\]\]</pre> | n, U0000FFFE, \{, $, \] <br/> _note: lengths are exact_ |
+| `charName`     | <pre>= s \[A\-Za\-z0\-9\] \[\-A\-Za\-z0\-9\_\\x20\]\* s</pre>                | TIBETAN LETTER \-A                       |
+| `bracketedHex` | <pre>= '\{' s hexCodePoint \(s hexCodePoint\)\* s '\}'</pre>        | \{61 2019 62\}  <br/> *note: the whitespace between hexCodePoints is required here*                         | <!-- TODO: Multi escape semantics? If so, add note explaining here or somewhere else -->
+| `hexCodePoint` | <pre>= hex\{1,5\} \| '10' hex\{4\}</pre>                           |                                         |
+| `hex`          | <pre>= \[0\-9A\-Fa\-f\]</pre>                                       |                                         |
+| `var`          | <pre>= '$' \[:ID_Start:\] \[:ID_Continue:\]\*</pre>                 | $a, $elt5  (optional support)            | <!-- TODO: XID* or ID*? -->
```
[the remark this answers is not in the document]
eggrobin: There is no strong reason to switch to XIDS/XIDC; the difference is motivated by NFKC, see https://www.unicode.org/reports/tr31/#NFKC_Modifications.
We use case-sensitive identifiers, so we should use NFC for identifier equivalence, not NFKC(_CF), thus that motivation is not relevant.
If compatibility is harmless, we should be compatible; IDS/IDC sounds good to me. See also the compatibility considerations from https://www.unicode.org/reports/tr55/tr55-2.html#Evolution-Unicode-3, which recommend IDS/IDC or XIDS/XIDC depending on where you are coming from.

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `prop`         | <pre>= '\\' \[pP\] '\{' propName \(\[≠=\] s value1\+\)? '\}'<br/>\| '\[:' '^'? propName \(\[≠=\] s value2\+\)? ':\]'</pre> | \\p\{x=y\}, \[:x=y:\]<br/> |
+| `propName`     | <pre>= s \[A\-Za\-z0\-9\] \[A\-Za\-z0\-9\_\\x20\]\* s</pre>                | General\_Category,<br/>General Category  |
+| `value1`       | <pre>= \[^\\\}\]<br/>\| '\\' quoted</pre>                         | Lm,<br/>\\n,<br/>\\\}                    |
+| `value2`       | <pre>= \[^:\]<br/>\| '\\' quoted</pre>                           | Lm,<br/>\\n,<br/>\\:                    |
+| `element`      | <pre>= char \| string \| var </pre> | %, b, \{hello\}, \{\}, \\x\{61 62\} |
+| `string`      | <pre>= '\{' \(s charInString\)\* s '\}' </pre> | \{hello\} |
+| `char`         | <pre>= \[^ \\^ \\& \\\- \\\[ \\\] \\\\ \\\{ \\$ \[:Pat_WS:\]\]<br/>\| '\\' quoted</pre> | a, b, c, \\n, \\\{, \\$ |
+| `charInString` | <pre>= \[^ \\\\ \\\} \[:Pat_WS:\]\]<br/>\| '\\' quoted</pre> | a, b, c, \\n, \{, $ <br/> *note: only quoted items that resolve to one code point are allowed* | <!-- TODO: Multi escape semantics? If flattening semantics, this note is unnecessary -->
```
macchiati, line 2890: I don't understand "note: only quoted items that resolve to one code point are allowed", because resolving to multiple characters is certainly allowed.

They are only disallowed in ranges, like {ab}-{cd}

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `quoted`       | <pre>= 'u' \(hex\{4\} \| bracketedHex\)<br/>\| 'x' \(hex\{2\} \| bracketedHex\)<br/>\| 'U00' \('0' hex\{5\} \| '10' hex\{4\}\)<br/>\| 'N\{' charName '\}'<br/>\| \[\[\\u0000\-\\U00010FFFF\]\-\[uxUN\]\]</pre> | n, U0000FFFE, \{, $, \] <br/> _note: lengths are exact_ |
+| `charName`     | <pre>= s \[A\-Za\-z0\-9\] \[\-A\-Za\-z0\-9\_\\x20\]\* s</pre>                | TIBETAN LETTER \-A                       |
+| `bracketedHex` | <pre>= '\{' s hexCodePoint \(s hexCodePoint\)\* s '\}'</pre>        | \{61 2019 62\}  <br/> *note: the whitespace between hexCodePoints is required here*                         | <!-- TODO: Multi escape semantics? If so, add note explaining here or somewhere else -->
```
macchiati, line 2893: > note: the whitespace between hexCodePoints is required here
syntactically it is not. Maybe:
note: the whitespace between hexCodePoints is needed for separate code points. \u{2345} ≠ \u{23 45} = \u{23}\u45}

docs/ldml/tr35.md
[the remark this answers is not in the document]
macchiati: We're encouraging XID over ID partly for uniformity; less to explain to people.

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `hexCodePoint` | <pre>= hex\{1,5\} \| '10' hex\{4\}</pre>                           |                                         |
+| `hex`          | <pre>= \[0\-9A\-Fa\-f\]</pre>                                       |                                         |
+| `var`          | <pre>= '$' \[:ID_Start:\] \[:ID_Continue:\]\*</pre>                 | $a, $elt5  (optional support)            | <!-- TODO: XID* or ID*? -->
+| `s`            | <pre>= \[:Pattern_White_Space:\]\*</pre>                          | optional whitespace                     |
+
+Some constraints on UnicodeSet syntax are not captured by this EBNF.
+Notably:
+1. Property names and values are restricted to those supported by the implementation, and have additional constraints imposed by [[UAX44](https://www.unicode.org/reports/tr41/#UAX44)].
+2. Quoted characters in strings must resolve to exactly one character. That is, **\[\{\x{61\}\}\]** is valid and equivalent to **\[\{a\}\]**, but **\[\{\\x\{61 62\}\}\]** is invalid and in particular not equivalent to **\[\\x\{61 62\}\]**, which is equivalent to **\[\{ab\}\]**. <!-- TODO: Multi escape semantics? -->
```
macchiati, line 2902: Ok, I see where you are coming from earlier.

if ICU permits [\u{23 45}] but does not permit [{\u{23 45}}] then that is probably a problem with ICU's implementation. That is, where supported, \u{23 45} should always be equivalent to \u{23}\u{45}

NOTE: you can't use the online jsp tools for this, because they do some extra unescaping beyond what UnicodeSet supports.

macchiati: Also, you're missing the following from my PR.
> If […] starts with [:, then it begins a prop, and must terminate with :]. Thus [:di:] is a valid property expression, [di:] is a 3 code-point set, and [:di] raises an error.

(You might get to this later on.)

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+3. Ranges (**\[X-Y\]**) are only supported in the case that X and Y resolve to single code points. That is, **\[a-b\]** and **\[\{a\}-\{b\}\]** are supported, while **\[a-{bz}\]** and **\[\{ax\}-\{bz\}\]** are not. <!-- TODO: Support for string ranges of length != 1? -->
```
macchiati, line 2903: maybe 
Ranges (**X-Y**) are only supported in the case that elements X and Y

docs/ldml/tr35.md
```diff
@@ -2904,12 +2918,59 @@ The syntax characters are listed in the table below:
...
+* \- is equivalent to the literal character \\- when occuring at the very beginning of a set, after a \^ at the beginning of a set, or at the very end of a set, in all cases with or without whitespace (**[-abc]**, **[ ^ -abc]**, **[abc-]**), and used as the set difference or range operator elsewhere (**[[abc]-[bc]]**, **[a-z]**)
+* \: initiates a POSIX property set when directly after a \[ without whitespace inbetween (**[:L:]**), ends a POSIX property set when directly before a \] without whitespace inbetween (**[:L:]**), and is equivalent to the literal character \\\: in any other place (**[ \:]**, **[L\:]**)
+* \} ends a string when occurring inside a string (**[{hello}]**), and is equivalent to the literal character \\\} in any other place (**[}a]**)
+
+###### Syntax Special Case Examples
+In the following, a list of examples including common sources of confusion concerning the UnicodeSet syntax:
```
macchiati, line 2929: Use a table for better organization. One column for the expression, and a second for the description (eg either 
the set of 'a' and '^'
or 
a syntax error

docs/ldml/tr35.md
```diff
@@ -2904,12 +2918,59 @@ The syntax characters are listed in the table below:
...
 
-A string with multiple code points is represented in a list by being surrounded by curly braces, such as in **[a-z \{ch}]**. It can be used with the range notation, as described in _Section String Range](#String_Range)_ . There is an additional restriction on string ranges in a UnicodeSet: the number of codepoints in the first string of the range must be identical to the number in the second. Thus [\{ab}-\{c}] and [\{ab}-c] are invalid.
+A string with multiple code points is represented in a list by being surrounded by curly braces, such as in **[a-z \{ch}]**. It can be used with the range notation, as described in _Section String Range](#String_Range)_ . There is an additional restriction on string ranges in a UnicodeSet: the number of codepoints in the first string of the range must be identical to the number in the second. Thus [\{ab}-\{c}] and [\{ab}-c] are invalid. A string consisting of a single code point is equivalent to that code point, that is, **[\{a}-c]** is valid and equivalent to **[a b c]**.  <!-- TODO: Support for string ranges of length != 1? -->
```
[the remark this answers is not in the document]
macchiati: Let's not.

docs/ldml/tr35.md
```diff
@@ -2972,6 +3033,40 @@ The binary operators '&', '-', and the implicit union have equal precedence and
 
 **One caution:** the '&' and '-' operators operate between sets. That is, they must be immediately preceded and immediately followed by a set. For example, the pattern **[[:Lu:]-A]** is illegal, since it is interpreted as the set **[:Lu:]** followed by the incomplete range **-A**. To specify the set of upper case letters except for 'A', enclose the 'A' in brackets: **[[:Lu:]-[A]]**.
 
+##### <a name="Variables_in_UnicodeSets" href="#Variables_in_UnicodeSets">Variables in UnicodeSets</a>
+
+Support for variable identifiers (var) is optional.
```
macchiati, line 3038: Add: But when they are used, they are defined as follows:

docs/ldml/tr35.md
```diff
@@ -2972,6 +3033,40 @@ The binary operators '&', '-', and the implicit union have equal precedence and
...
 **One caution:** the '&' and '-' operators operate between sets. That is, they must be immediately preceded and immediately followed by a set. For example, the pattern **[[:Lu:]-A]** is illegal, since it is interpreted as the set **[:Lu:]** followed by the incomplete range **-A**. To specify the set of upper case letters except for 'A', enclose the 'A' in brackets: **[[:Lu:]-[A]]**.
 
+##### <a name="Variables_in_UnicodeSets" href="#Variables_in_UnicodeSets">Variables in UnicodeSets</a>
+
+Support for variable identifiers (var) is optional. 
+They are used in certain contexts such as in [https://cldr-smoke.unicode.org/spec/main/ldml/tr35-general.html#Transforms](Transforms).
+
+UnicodeSets may contain variables (`$my_char`, `$the_set`, ...) in place of full UnicodeSets and strings/characters.
+
+This introduces ambiguities when parsing. For example, it is unclear whether **[$a-$b]** is a range of strings/characters, or a set difference, because that depends on the values of the variables.
```
macchiati, line 3043: We have to define it so that it **doesn't produce ambiguities**. Then this first line is false. That is, implementations that support variables must supply the set of variables and their values at the time that a UnicodeSet is parsed. (you could then explain why this is needed in a note, if you want).

docs/ldml/tr35.md
```diff
@@ -2972,6 +3033,40 @@ The binary operators '&', '-', and the implicit union have equal precedence and
...
+In particular in the case of **$a-$b**, it must be that **$a** and **$b** are of the same kind, i.e., both are strings/characters or both are UnicodeSets. A variable map with any other combination (e.g., `{ a => '0', b => [:L:] }`) is incompatible with the syntax **$a-$b**, and would return an error at parse-time.
+Thus, implementations that support variables may want to require the value of variables to be known at UnicodeSet parse time.
+Parsing a UnicodeSet with variables support requires referenced variables to exist, in other words, parsing a UnicodeSet referring to an unknown variable leads to a parse error.
+
+Variables are replaced by value, that is, **[a \$minus z]** with a variable map `{ minus => '-' }` is equivalent to **[-az]**, not **[a-z]** (i.e., cardinality of 3 instead of 26).
```
macchiati, line 3048: I believe that this is not done by ICU in transforms, so this would be needed to add to the list of changes that we would want in ICU.

docs/ldml/tr35.md
```diff
@@ -2985,7 +3079,7 @@ The following table summarizes the syntax that can be used.
 | [[pat1]&[pat2]]      | The intersection of sets specified by pat1 and pat2 |
 | [[pat1]-[pat2]]      | The asymmetric difference of sets specified by pat1 and pat2 |
 | [a \{ab} \{ac}]      | The code point 'a' and the multi-code point strings "ab" and "ac" |
-| [x\\u\{61 2019 62}y] | Equivalent to [x\\u0061\\u2019\\u0062y] (= [xa’by]) |
+| [x\\u\{61 2019 62}y] | Equivalent to [x\\u0061\\u2019\\u0062y] (= [xa’by]) | <!-- TODO: Multi escape semantics? -->
```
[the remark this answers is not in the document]
macchiati: The way that multi-code-point escapes are defined, they are exactly equivalent to a series of single code point escapes. Having them magically convert to a string would break that.

macchiati (changes requested):
First pass

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
-| `charName`     | <pre>= s [A-Za-z0-9] [-A-Za-z0-9_\x20]* s</pre>                | TIBETAN LETTER -A                       |
-| `bracketedHex` | <pre>= '{' s hexCodePoint (s hexCodePoint)* s '}'</pre>        | \{61 2019 62}                           |
-| `hexCodePoint` | <pre>= hex{1,5} \| '10' hex{4}</pre>                           |                                         |
-| `hex`          | <pre>= [0-9A-Fa-f]</pre>                                       |                                         |
-| `s`            | <pre>= [:Pattern_White_Space:]*</pre>                          | optional whitespace                     |
-
-Some constraints on UnicodeSet syntax are not captured by this EBNF. Notably, property names and values are restricted to those supported by the implementation, and have additional constraints imposed by [[UAX44](https://www.unicode.org/reports/tr41/#UAX44)]. In addition, quoted values that resolve to more than one code point are disallowed in ranges of the form `char '-' char`.
+| `root`         | <pre>= prop<br/>\| '\[' '^'? s '-'? s seq\* \[\\$ \\-\]? s '\]' <br/>\| var</pre> | \\p\{x=y\},<br/>[abc],<br/>$myset                     |
```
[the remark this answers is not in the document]
skius: To be clear about the handling of whitespace in the `root` nonterminal: ICU was the main inspiration for this, i.e., the previously UTS35-unsupported cases `[^-]`, `[^ -]`, `[ -]` *are* supported in ICU, and *are* supported in this new syntax.

One caveat: Per your suggestion I disallowed whitespace in front of the `^` for negation, even though ICU currently supports `[ ^]` (same as `[^]`).

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `seq`          | <pre>= root \(s \[\\&\\\-\] s root\)\* s<br/>\| range s</pre>        | \[abc\]\-\[cde\], a                          |
+| `range`        | <pre>= element \('\-' element\)?       | a, a\-c, \{abc\}, \{ax\}\-\{bz\}                          | <!-- TODO: Support for string ranges of length != 1? -->
```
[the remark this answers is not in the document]
skius: Thank you, changed!

docs/ldml/tr35.md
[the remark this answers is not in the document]
skius: Thank you, changed!

docs/ldml/tr35.md
skius: This was a left-over from my interpretation that multi-escapes are strings, not flattened characters. Changed it now to reflect the flattened meaning.

docs/ldml/tr35.md
skius: I do get to that later on, but I think this is an important enough addition to add it to the notes directly underneath the EBNF. Added it here.

There is some slight duplication between the constraint section underneath the EBNF, and the constraint section underneath the table of syntax characters, however I think the following way to think about that split makes sense:
* The section underneath the EBNF talks about _global_ constraints about the syntax and some semantics
* The section underneath the table of syntax chars talks about _context-sensitive_ constraints on syntax characters

If you disagree, we can try to combine the two constraint sections somehow.

docs/ldml/tr35.md
skius: Thank you, changed!

docs/ldml/tr35.md
skius: Thank you, added!

docs/ldml/tr35.md
skius: Thank you, agreed. I switched the order around such that I first require variables' values be known, and then I state the problems of the alternative in a conditional phrase.

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `value1`       | <pre>= \[^\\\}\]<br/>\| '\\' quoted</pre>                         | Lm,<br/>\\n,<br/>\\\}                    |
+| `value2`       | <pre>= \[^:\]<br/>\| '\\' quoted</pre>                           | Lm,<br/>\\n,<br/>\\:                    |
+| `element`      | <pre>= char \| string \| var </pre> | %, b, \{hello\}, \{\}, \\x\{61 62\} |
+| `string`      | <pre>= '\{' \(s charInString\)\* s '\}' </pre> | \{hello\} |
+| `char`         | <pre>= \[^ \\^ \\& \\\- \\\[ \\\] \\\\ \\\{ \\$ \[:Pat_WS:\]\]<br/>\| '\\' quoted</pre> | a, b, c, \\n, \\\{, \\$ |
+| `charInString` | <pre>= \[^ \\\\ \\\} \[:Pat_WS:\]\]<br/>\| '\\' quoted</pre> | a, b, c, \\n, \{, $ <br/> *note: only quoted items that resolve to one code point are allowed* | <!-- TODO: Multi escape semantics? If flattening semantics, this note is unnecessary -->
+| `quoted`       | <pre>= 'u' \(hex\{4\} \| bracketedHex\)<br/>\| 'x' \(hex\{2\} \| bracketedHex\)<br/>\| 'U00' \('0' hex\{5\} \| '10' hex\{4\}\)<br/>\| 'N\{' charName '\}'<br/>\| \[\[\\u0000\-\\U00010FFFF\]\-\[uxUN\]\]</pre> | n, U0000FFFE, \{, $, \] <br/> _note: lengths are exact_ |
```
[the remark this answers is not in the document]
skius: We are in an already-escaped context here, so `n` is a valid production of `quoted`. Yes, it will be used as `\n`, because `quoted` always appears after a `\` , but that's not actually a concern of `quoted` itself.

docs/ldml/tr35.md
eggrobin: The definition is fine, and indeed unambiguous when everything is taken together; what this is trying to get at is that if you are solely doing parsing based on the grammar given above, you get a conflict which you can only resolve by looking at what the variables actually are.
This is not uncommon (e.g. `a * b;` in C could be declaring a variable `b` of tye `a*` or multiplying `a` by `b` and discarding that, depending on whether `a` is a type), but noteworthy.

```suggestion
The types of the variables need to be known in order to determine how to parse some UnicodeSets.
For example, the UnicodeSet **[$a-$b]** could formally be either a range if $a and $b are characters, or a set difference if they are sets.
In practice, it is expected that parsing would be done with the values of the variables already known. 
```

docs/ldml/tr35.md
skius: You are right, I added a (probably incomplete in minor ways) section to the PR description: "Required changes in ICU to reflect the new syntax".

In this case, CLDR data actually works the same with the ICU variable behavior and this draft's variable behavior. There are edge cases that work differently in current ICU and this draft, e.g., the example from the draft: `[a $minus z]` is `a, z, -` in this draft, and `a, b, c, ..., z` in ICU.

docs/ldml/tr35.md
```diff
@@ -2861,38 +2861,52 @@ Element content whose display may be affected in this way should include an expl
...
+| `root`         | <pre>= prop<br/>\| '\[' '^'? s '-'? s seq\* \[\\$ \\-\]? s '\]' <br/>\| var</pre> | \\p\{x=y\},<br/>[abc],<br/>$myset                     |
+| `seq`          | <pre>= root \(s \[\\&\\\-\] s root\)\* s<br/>\| range s</pre>        | \[abc\]\-\[cde\], a                          |
+| `range`        | <pre>= element \('\-' element\)?       | a, a\-c, \{abc\}, \{ax\}\-\{bz\}                          | <!-- TODO: Support for string ranges of length != 1? -->
+| `prop`         | <pre>= '\\' \[pP\] '\{' propName \(\[≠=\] s value1\+\)? '\}'<br/>\| '\[:' '^'? propName \(\[≠=\] s value2\+\)? ':\]'</pre> | \\p\{x=y\}, \[:x=y:\]<br/> |
+| `propName`     | <pre>= s \[A\-Za\-z0\-9\] \[A\-Za\-z0\-9\_\\x20\]\* s</pre>                | General\_Category,<br/>General Category  |
+| `value1`       | <pre>= \[^\\\}\]<br/>\| '\\' quoted</pre>                         | Lm,<br/>\\n,<br/>\\\}                    |
```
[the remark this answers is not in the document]
skius: I opted for the slightly longer `pValuePerl` and `pValuePosix` to avoid confusion (does pvalueCurly _contain_ Curly or _exclude_ Curly?)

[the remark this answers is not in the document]
skius: I avoided touching the whole `prop` in general, because I don't quite understand that.

In a perfect world, there would be:
* An alternative nonterminal specifying the case where `propName` is left out (e.g., `[:L:]`
  * Reason being, what is the meaning of `[:≠:]`? Is it a syntax error, or `[:General_Category=≠:]`?
* An unambiguous set for `value1/2`:
  * whitespace is not excluded in `value`, but still valid syntax after `=/≠`
  * `\` is not excluded in `value`, so in theory both `[:name=\:]` and `[:name=\::]` are valid

I assume there's a reason for the value nonterminals to be so unrestrictive in the first place, so I feel underqualified to change them.

docs/ldml/tr35.md
skius: Thanks, changed!

docs/ldml/tr35.md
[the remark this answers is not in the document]
eggrobin: > I assume there's a reason for the value nonterminals to be so unrestrictive in the first place

I am sure there is a reason, but I would not think there is a good one.

> what is the meaning of `[:≠:]`?

It is an error in ICU, as it should be, though not a syntax error. It might as well be a syntax error.

Note however that `[:meow:]` is also an error, but cannot be a syntax error (if we were to add a binary property `Meow`, a script `Meow`, or a General_Category alias `Meow`, it would become well-formed).

docs/ldml/tr35.md
skius: @eggrobin I think we wrote over each other - Is the new version okay with you or too strict? https://github.com/unicode-org/cldr/pull/3089/files#diff-549880ab02f2fc27e87768d55cd7030b3afb0f5b18db21ef4297c2cddf9c985bR3047

docs/ldml/tr35.md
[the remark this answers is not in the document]
skius: Are there (supported) properties with "non-propName/charName" values? 
Is it valid for me to just search through https://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt and  https://www.unicode.org/Public/UCD/latest/ucd/PropertyAliases.txt, or is there a more complete list?

I agree that we should not restrict UnicodeSet property _syntax_ to all currently-valid expressions of properties (`letter`, `letteR`, `lette r`, ...).

docs/ldml/tr35.md
```diff
@@ -2972,6 +3036,43 @@ The binary operators '&', '-', and the implicit union have equal precedence and
...
+When they are used, they are defined as follows:
+
+UnicodeSets may contain variables (`$my_char`, `$the_set`, ...) in place of full UnicodeSets and strings/characters.
+
+Implementations that support variables must require the value of variables to be known at UnicodeSet parse time.
+Parsing a UnicodeSet with variables support requires referenced variables to exist, in other words, parsing a UnicodeSet referring to an unknown variable leads to a parse error.
+
+If the values of variables were unknown at parse time, this would introduce ambiguities when parsing, thus the values must be known.
+For example, it would be unclear whether **[$a-$b]** is a range of strings/characters, or a set difference, because that depends on the values of the variables.
+In particular in the case of **$a-$b**, it must be that **$a** and **$b** are of the same kind, i.e., both are strings/characters or both are UnicodeSets. A variable map with any other combination (e.g., `{ a => '0', b => [:L:] }`) is incompatible with the syntax **$a-$b**, and would return an error at parse-time.
```
eggrobin, line 3052: Reading this again it feels a bit odd, because usually we don’t expose a separate parsing operation, only « turn this string in UnicodeSet syntax into an actual set of characters and strings », which obviously requires the values of the variables.

How about this, trimming the discussion of the reduce-reduce conflict and keeping it a note, and putting the emphasis on the fact that variable assignments can be invalid, which is what really matters to the user, as opposed to what might matter to the implementer:
```suggestion
Not all variable maps are valid for a given expression in UnicodeSet syntax.
For instance, consider **[$a-$b]**; this may be a range of characters if both **$a** and **$b** are characters,
or a difference of sets if they are both sets; but given the map `{ a => '0', b => [:L:] }`, it is invalid.

> **Note:** In particular, the variable map is needed not just to compute the actual set of characters and strings represented by the UnicodeSet,
but also to parse the UnicodeSet syntax: if **$a** and **$b** are unknown, the parsing of **[$a-$b]** is ambiguous.
```

skius: Thanks! Tried something in https://github.com/unicode-org/cldr/pull/3089/commits/8f504903fc089f4b72d67c3f5acd6662bfad57f8 :)

docs/ldml/tr35.md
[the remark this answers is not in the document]
macchiati: What you need to look at is the behavior with "free-spacing" mode (?x).
See https://gist.github.com/CMCDragonkai/6c933f4a7d713ef712145c5eb94a1816#:~:text=(%3Fx)%20(-,free%2Dspacing%20mode,-)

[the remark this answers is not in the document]
skius: Got it, thanks! I think we might even need to talk about `(?xx)`, because `(?x)` in PCRE still treats whitespace in character classes significantly - https://www.regular-expressions.info/freespacing.html

With `(?x)`: The behavior I described in the previous comment
With `(?xx)`: whitespace is ignored, for both `^` and `-`. `(?xx)[ ^ a- ]` matches everything except `a` and `-`.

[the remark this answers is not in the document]
eggrobin: > With (?xx): whitespace is ignored, for both ^ and -. (?xx)[ ^ a- ] matches everything except a and -.

In that case we should keep the current ICU behaviour (`[ ^ a ]` = `[^a]`), since compatibility agrees with consistency with PCRE.

docs/ldml/tr35.md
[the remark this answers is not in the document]
macchiati: - Perl and Posix are fine.
- If there is no propName it is a syntax error, so [::], [:=:], [:=L:], are all syntax errors (same for Perl syntax). As Robin says, we don't restrict the syntax to currently-valid properties. 

- Whitespace, _, case variants are allowed arbitrarily in propNames so [: d _ I _ :] = [:di:]
- As for the values, it depends. Most properties that behave like propName. [:gc=Lc:] = [: g c = L _ C :], [: n v = 1 _ 0 _ :] = [:nv=10:]
- String property values are different; whitespace (and _, and case differences) **are** meaningful. [:bmg=(:] ≠ [:bmg= ( :]
- We should reference the [the Regex spec](https://unicode.org/reports/tr18/) and in Section 5.9 Matching Rules of [[UAX44](https://www.unicode.org/reports/tr18/proposed.html#UAX44)] for propNames and pValues

I realized that there is another optional feature that we should describe here as well: implementations can add additional propNames and pValues.

docs/ldml/tr35.md
macchiati: The types of the variables need to be known in order to determine how to parse some UnicodeSets.

Not just "in practice". And that needs to be the first point (before 'types of')

docs/ldml/tr35.md
macchiati: That's what I was thinking, that ICU just does a variable substitution before parsing (the simplest implementation)

docs/ldml/tr35.md
```diff
@@ -2904,17 +2918,62 @@ The syntax characters are listed in the table below:
...
 
-A string with multiple code points is represented in a list by being surrounded by curly braces, such as in **[a-z \{ch}]**. It can be used with the range notation, as described in _Section String Range](#String_Range)_ . There is an additional restriction on string ranges in a UnicodeSet: the number of codepoints in the first string of the range must be identical to the number in the second. Thus [\{ab}-\{c}] and [\{ab}-c] are invalid.
+A string with multiple code points is represented in a list by being surrounded by curly braces, such as in **[a-z \{ch}]**. It can be used with the range notation, with the restriction that each string contains exactly one code point. Thus **\[\{ab\}-\{c\}\]**, **\[\{ax\}-\{bz\}\]**, and **\[\{ab\}-c\]** are invalid. A string consisting of a single code point is equivalent to that code point, that is, **[\{a}-c]** is valid and equivalent to **[a b c]**.
 
-In UnicodeSets, there are two ways to quote syntax code points:
-
-<a name="Backslash_Escapes"></a>
-Outside of single quotes, certain backslashed code point sequences can be used to quote code points:
+##### <a name="Backslash_Escapes" href="#Backslash_Escapes">Backslash Escapes</a>
```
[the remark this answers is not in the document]
macchiati: ok

docs/ldml/tr35.md
skius: Since we define 'types' of variables inside variable maps, I think that approach is not a valid implementation anymore. If we really wanted to allow implementations to pass in arbitrary strings of syntax characters, we'd need to define a new 'type' "SyntaxStream" or something like that. 

I am strongly against that (I believe @eggrobin is as well), but I understand that forcing implementation work onto ICU needs to be justified well. Theoretically speaking, until ICU decides that they do want proper variable support in UnicodeSets, they could just state that their UnicodeSet implementation _does not support variables as defined in UTS35 UnicodeSets_. As long as CLDR data adheres to UTS35, and both implementations of variables result in the same interpretation, that might be an option?

docs/ldml/tr35.md
eggrobin: (That suggested text is outdated.)

docs/ldml/tr35.md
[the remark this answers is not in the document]
skius: Are we defining how escaping should be handled here? For example, are we saying that `[:bmg=\::]` is referring to all code points with property `bmg` set to `:` (i.e., the `\:` is escaped)?

I haven't found a supported string property in ICU, so I don't know what the precedent is here.

My suggestion (assuming we want to handle escaping on the UnicodeSet side) would be something like:

```
prop = '[:' '^'? propName ([≠=] pValuePosix+)? ':]' ;
propName = [-A-Za-z0-9_[:Pattern_White_Space]]+ ; # Note: must contain at least one char in [A-Za-z0-9]
pValuePosix = [^ : \\ ] | '\' quoted ;
```

(and correspondingly for Perl property expressions)

This is compatible with [UAX44-LM3](https://www.unicode.org/reports/tr44/#UAX44-LM3) - implementations parse a propName that may contain whitespace, underscores, hyphens, and do the loose matching according to UAX44 themselves, if they want to. A similar thing for pValuePosix, except that escaping is supported, and whitespace may or may not be significant, hence the lack of a `Note` as in propName (it is significant in cases like `[:bmg= :]`, I assume).

(A side note on handling the escaping part in UnicodeSets: The playground does not escape, so `[:bmg=::]` is valid and results in `[\:]`, as is `[:bmg=\:]` which results in `[\\]`.)

I can try adding that (and your other comments about referencing UAX44 and custom propNames/pValues), if you agree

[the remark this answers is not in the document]
macchiati: https://util.unicode.org/UnicodeJsps/properties.jsp lists the properties with the value type (UCD, ICU,...)

You don't need to quote the : in your example, eg [[:Case_Folding=::]](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ACase_Folding%3D%3A%3A%5D&g=&i=)

But you do need a quote in \p{Case_Folding=\}}

**Syntactically**, I don't think UnicodeSet in ICU currently cares what the character repertoire is for the propName or the pValues, other than the exclusion of =, ≠, } or :]. Although it is unlikely that the UCD would ever have non-ASCII, we might not want to limit it to ASCII. Please check the code for UnicodeSet to see what the constraints are.

docs/ldml/tr35.md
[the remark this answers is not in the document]
macchiati: yup, my oversight.

docs/ldml/tr35.md
macchiati: Let's mull this over a bit. Since it is optional, we could also say that there are two optional profiles for extending US for variables: string substitution or syntactic substitution. We can encourage the latter, but allow for the former.

(Note: there is a ticket for formalizing conformance clauses in LDML, but we don't want to wait on that for your ticket.)

docs/ldml/tr35.md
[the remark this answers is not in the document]
skius: This belongs here: While investigating properties, I noticed that ICU also ignores whitespace between `:` and `^` in e.g. `[:  ^ L :]`

docs/ldml/tr35.md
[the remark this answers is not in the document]
skius: > https://util.unicode.org/UnicodeJsps/properties.jsp lists the properties with the value type (UCD, ICU,...)

`toNFM` is listed with source ICU, yet ICU throws an `Invalid name: toNFM` exception: https://github.com/unicode-org/icu/blob/70d308731ae01a4187fe49f7ae1c8e1ad1570be6/icu4j/main/classes/core/src/com/ibm/icu/lang/UCharacter.java#L5159

> You don't need to quote the : in your example, eg [[:Case_Folding=::]](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ACase_Folding%3D%3A%3A%5D&g=&i=)

Exactly - this is invalid according to the `pValuePosix`/`value2` nonterminal. Do we want to support unquoted `:` characters in POSIX property sets, and add a worded-out note saying the sequence `:]` ends the expression eagerly?

> Please check the code for UnicodeSet to see what the constraints are.

`=` splits anything existing before `}` or `:]` into two. So yes, ICU currently does not care about the character repertoire. Interestingly, `≠` is not supported (neither is `!=`). Also, no escaping is happening. The code lives here: 
https://github.com/unicode-org/icu/blob/70d308731ae01a4187fe49f7ae1c8e1ad1570be6/icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java#L3797

docs/ldml/tr35.md
skius: Sounds good to me! 

Can I assume this current draft will (at least roughly) be a supported variant in the future? This is relevant for my implementation work in ICU4X. Small changes like whitespace significance here and there are no problem of course, but I'd like to confirm that this draft's typed variable approach will most likely _not_ conflict with any spec changes in the near future, and is safe to implement.

eggrobin: > we could also say that there are two optional profiles for extending US for variables: string substitution or syntactic substitution

This is a terrible idea. The whole point of putting them in the grammar (and of this proposal) is to avoid this kind of interaction between the syntax and the variables; conversely, you cannot have variables within the grammar if they may do arbitrary syntactic things.

As Niels points out, if someone feels like doing string substitution, they can always do so at a higher level; the UnicodeSet syntax should not standardize higher-level syntaxes. After all, an implementation can always do `UnicodeSet(std::format("[a{}z]", "-"))`, and this is none of LDML’s business.

If you really want a profile, that profile should be in the transform rules syntax (that is the higher level syntax here), where instead of using UnicodeSet variables, an implementation could do string substitution.
But there, UnicodeSet is just the tip of the iceberg; if we start defining every weird edge case that Niels and I found playing with ICU’s implementation of the transform rules syntax, we’ll be writing profiles for another year.

eggrobin: > if we start defining every weird edge case that Niels and I found playing with ICU’s implementation of the transform rules syntax, we’ll be writing profiles for another year.

(Off-topic for UnicodeSet, but fun example: https://util.unicode.org/UnicodeJsps/transform.jsp?a=%24meow+%3D+%28meo%2B%29w%3B%0D%0A%0D%0A%24meow%2B+%3E+%24meow%3B&b=meoooowmeoow%3Bmeowmeow%2Fmeow.)

docs/ldml/tr35.md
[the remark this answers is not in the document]
macchiati: > yet ICU throws an Invalid name
The jsps add some custom properties. The ones that are supported in ICU (even if not in ICU's version of UnicodeSet) are still marked "ICU" in that list.

Re [[:Case_Folding=::]](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ACase_Folding%3D%3A%3A%5D&g=&i=), I get what you mean not. Yes, that should be supported; probably best do with an explicit constraint note. 

Ok, ICU doesn't care about the repertoire. So there are three options, roughly:

1. XID based (but limited to ASCII), eg UCD style
2. XID based, eg adds other languages/scripts for future possible extension
3. Anything goes, eg current ICU

We could live with any of these (if not #3, we can always have a profile for that usage; if not #1, we can note that current props are always ASCII.) I guess I favor #2 as the cleanest, but it is not a strong preference. Others?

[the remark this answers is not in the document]
eggrobin: No strong opinions from me here; I think I agree that 2 is cleaner, and it probably leads to better error messages (saying `[:]:]` or `[:cf!=fi:]` is a syntax error seems more helpful than complaining about `]` not being a known property or property value alias, or `cf!` not being a known property).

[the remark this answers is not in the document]
eggrobin: > if not 3, we can always have a profile for that usage

Note that there is no need for a profile for 3, because the specification does not require distinguishing syntax errors from other errors: `[:≠:]` is just a bad UnicodeSet; ICU4C/J will tell you that is because `≠` is neither a binary property nor a GC or Script alias, and ICU4X might tell you that it is syntactically incorrect, but it is bad either way.

docs/ldml/tr35.md
```diff
@@ -2904,17 +2920,67 @@ The syntax characters are listed in the table below:
 |  }   | U+007D | RIGHT CURLY BRACKET  | Strings in set; Perl property syntax       |
 |      | U+0020 U+0009..U+000D U+0085<br/>U+200E U+200F<br/>U+2028 U+2029 | ASCII whitespace,<br/>LRM, RLM,<br/>LINE/PARAGRAPH SEPARATOR | Ignored except when escaped |
 
+Note that some syntax characters only have a special meaning in a certain context. In particular:
+* Out of all above syntax characters, only \\, \}, and whitespace have a special meaning inside strings (**\[\{\[a-z\]\}\]** is the set of the string '\[a-z\]', **\[\{\$blah\}\]** is the set of the string '\$blah').
+* \$ is equivalent to \uFFFF when appearing at the very end of a set with or without trailing whitespace (**[a-z\$]**, **[a-z\$ ]**), and used as starting indicator for a variable reference elsewhere, in which case the variable name will be the longest match on the `var` nonterminal (**[\$my_set]**).
```
macchiati, line 2925: > on the var nonterminal ([$my_set]).
This is a bit mysterious

docs/ldml/tr35.md
```diff
@@ -2904,17 +2920,67 @@ The syntax characters are listed in the table below:
...
+* \- is equivalent to the literal character \\- when occuring at the very beginning of a set, after a \^ at the beginning of a set, or at the very end of a set, in all cases with or without whitespace (**[-abc]**, **[ ^ -abc]**, **[abc-]**), and used as the set difference or range operator elsewhere (**[[abc]-[bc]]**, **[a-z]**)
+* \: initiates a POSIX property set when directly after a \[ without whitespace inbetween (**[:L:]**), ends a POSIX property set when directly before a \] without whitespace inbetween (**[:L:]**), and is equivalent to the literal character \\\: in any other place (**[ \:]**, **[L\:]**)
+* \} ends a string when occurring inside a string (**[{hello}]**), and is equivalent to the literal character \\\} in any other place (**[}a]**)
+
+###### Syntax Special Case Examples
+In the following, a table of examples including common sources of confusion concerning the UnicodeSet syntax:
```
macchiati, line 2931: Had an idea for reorging this for clarity, see https://docs.google.com/spreadsheets/d/1xbLH2pTFHR9AG0H3Ybl3aNsoxivzOr2Y_OcVUlYn_8M/edit

docs/ldml/tr35.md
macchiati: Ok, makes sense. BTW, I found that ICU implements the variables via SymbolTable, so it is not just text replacement. If we wanted to make ICU conform, we could (but possibly breaking compatibility with some external cliant).

macchiati (changes requested):

docs/ldml/tr35.md
skius: Which part is mysterious? (In this section I've followed the convention that all special cases have a UnicodeSet example in parentheses that makes use of that special case)

docs/ldml/tr35.md
skius: > If we wanted to make ICU conform, we could (but possibly breaking compatibility with some external cliant).

Yes, from what I can tell, a minimal compliance update would "just" have to update the UnicodeSet parser to make use of the SymbolTable in a less opaque way.

> BTW, I found that ICU implements the variables via SymbolTable, so it is not just text replacement.

Almost, it's a mix. In my opinion, that's even worse.

If you're interested, here's my understanding of how it works currently (in the context of transforms, using classes and names from ICU4J, though after a quick glance over ICU4C there's not much difference):

1. The SymbolTable that the transform rule parser uses is reused, in other words, variables are mapped to strings that are encodings of literal characters or some `UnicodeMatcher`s (`UnicodeSet` is a `UnicodeMatcher`), also encoded to characters (in the PUA)
2. The UnicodeSet parser uses a `RuleCharacterIterator`, which handles variables, to parse a UnicodeSet pattern.
3. A RuleCharacterIterator handles variable references, and *opaquely passes the string value of that variable on, character by character, without doing any decoding*
   * See [`next` that sets `this.buf`](https://github.com/unicode-org/icu/blob/2e45e6ec0e84a1c01812015a254ea31b286316fb/icu4j/main/classes/core/src/com/ibm/icu/impl/RuleCharacterIterator.java#L154)
   * See [`_current()` that uses `this.buf`](https://github.com/unicode-org/icu/blob/2e45e6ec0e84a1c01812015a254ea31b286316fb/icu4j/main/classes/core/src/com/ibm/icu/impl/RuleCharacterIterator.java#L338)
4. Back to UnicodeSet parser, this just receives characters from the iterator. Whenever it receives one, it tries to decode the character into a `UnicodeMatcher` that is specifically a `UnicodeSet`.
   * See [here](https://github.com/unicode-org/icu/blob/2e45e6ec0e84a1c01812015a254ea31b286316fb/icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java#L2699)
5. If the decoding fails, this must be a regular character, so it is treated as a literal syntax character (**there is no distinction in RuleCharacterIterator's API whether a passed character comes from a variable or from the source string**). If it is a UnicodeSet, the UnicodeSet case is handled.

The takeaway from this should be that while variables pointing to UnicodeSets do have a (this-draft-compliant) special case handling, variables pointing to actual strings do not. The input streams `"x␣␣-␣b"` and `"x␣<variable pointing to "␣-␣b">"` are **indistinguishable** from the POV of the UnicodeSet parser (␣ as space).

A few fun examples:
* [Probably unrealistic](https://util.unicode.org/UnicodeJsps/transform.jsp?a=%24sample_uset+%3D+%27This+is+a+sample+UnicodeSet+%5Ba-z0-9%5D%27+%3B%0D%0A%0D%0A%23+I+think+this+should+replace+either+the+sample+text+or+the+character+%27%7C%27+with+MATCHED%2C+right%3F%0D%0A%5B%7C+%24sample_uset%5D+%3E+MATCHED+%3B&b=The+following+should+be+matched%3A+This+is+a+sample+UnicodeSet+%5Ba-z0-9%5D%0D%0AThe+following+should+be+matched%3A+%7C%0D%0A%0D%0AThis+should+not+match%3A+45458976)
* [Quite realistic, in my opinion](https://util.unicode.org/UnicodeJsps/transform.jsp?a=%24gamma_suff+%3D+%27-gamma%27+%3B%0D%0A%24alpha_suff+%3D+%27-alpha%27+%3B%0D%0A%0D%0A%23+remove+all+version+suffixes%2C+right%3F%0D%0A%5B%24alpha_suff+%24gamma_suff%5D+%3E+%3B%0D%0A%0D%0A%23+explanation%3A+above+is+equivalent+to+%5B-alpha-gamma%5D%2C+so+%5B%5C-+a-g+h+l+m+p%5D%0D%0A%0D%0A%23+%28working+version%3A%29%0D%0A%23+%24alpha_suff+%3E+%3B%0D%0A%23+%24gamma_suff+%3E+%3B&b=best_application-gamma)

docs/ldml/tr35.md
macchiati: I think it would be clearer in this instance to say (such as [...)

docs/ldml/tr35.md
skius: Thanks, implemented the changes!

macchiati (approved):

macchiati:
If there are any further changes, we can do in additional PRs

PR #3090, opened 2023-07-11
CLDR-16841 Tweek the formatting

macchiati:
CLDR-16841

The formatting of some values lead to overly long numbers, and sometimes hard to read, so modified to avoid that. 
Also modified the chart generation to use that, and use a better sort. The tests were modified to suit, and also simplified to make them easier to read and update.

Examples:

Screenshot 2023-07-11 at 16 17 35

(This is just interim improvements, and don't yet move the examples into the blue pane.)

ALLOW_MANY_COMMITS=true


macchiati:
Steven, if you have time to review this before the meeting tomorrow, I'd like to push this together with the other merges to production. Cf
https://github.com/unicode-org/cldr/compare/844246fa6520cf6275e5622df345411ab431fdc6...main

srl295 (approved):
Output is much improved.

PR #3091, opened 2023-07-11
CLDR-16717 Fix six more Persian to IPA transform issues

roozbehp:
CLDR-16717 Fix six more Persian to IPA transform issues

ALLOW_MANY_COMMITS=true


srl295 (approved):
LGTM! thanks for the clear docs. It made sense going commit by commit

PR #3092, opened 2023-07-13
CLDR-16862 Upgrade sw and zh_Hant_HK to 8-vote minimum

AEApple:
CLDR-16862

Upgrade sw and zh_Hant_HK to 8-vote minimum

It looks like bg and is had already been upgraded in https://github.com/unicode-org/cldr/pull/2098

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
@@ -243,6 +243,7 @@ Google	;	yo	;	modern	;	Yoruba
 Google	;	yue	;	modern	;	T5	Cantonese
 Google	;	zh	;	modern	;	T1	Simplified Chinese
 Google	;	zh_Hant	;	modern	;	T1	Traditional Chinese
+Google  ; zh_Hant  ; modern 
 Google	;	zu	;	modern	;	T5	Zulu
 Google ; * ; moderate ; All Others, provisionally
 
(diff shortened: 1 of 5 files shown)


AEApple:
I'm actually a bit confused as to why these weren't picked up as a part of [CLDR-15687](https://unicode-org.atlassian.net/browse/CLDR-15687)

Also per the discussion, should we file a new ticket that changes the test to be 3+ TC orgs at modern?

srl295:
for the record, 
- one of the tests simply had a hard coded list. added zh_Hant_HK
- one of the tests had a hard coded list of orgs (actually separate variables for each org!) - filed CLDR-16869 and refactored to `Organization.getTcOrgs()` - that fixed `sw`

remaining issue:
- `zh_Hant_HK` only has one TC org at at least modern, so the test is saying it should NOT have a lower bar (8) vote.  Is the test wrong or should there be an exception?

AEApple:
Well I don't think that's right since at least 2 other orgs are contributing at modern for zh_HK  but not listed in locales.txt and all 4 list contributing at modern in the sheet, so somethings wrong.

I think the orgs have been contributing for so many years, and we've just been updating based on the diff YoY which is why they're missing.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
```diff
@@ -243,6 +243,7 @@ Google	;	yo	;	modern	;	Yoruba
 Google	;	yue	;	modern	;	T5	Cantonese
 Google	;	zh	;	modern	;	T1	Simplified Chinese
 Google	;	zh_Hant	;	modern	;	T1	Traditional Chinese
+Google  ; zh_Hant  ; modern
```
stenshamn, line 246: Don't you mean to add `zh_Hant_HK` here?

AEApple: argh

macchiati (approved):

PR #3094, opened 2023-07-13
CLDR-16846 xlsx to xml converter for bulk upload

srl295:
- move bulk upload jsp link into a new 'upload' panel
- add a converter from spreadsheet (xlsx/csv/etc)  (the Dashboard Download) to bulk-upload XML
- CLDRFile: add a 'leniency' option, disable validation for XML upload (it gets triple checked anyway)

CLDR-16846

FYI, draft

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrBulkConverter.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrBulkConverter.mjs
@@ -46,9 +46,9 @@ function xlsHeaders(wb) {
 }
 
 function xlsGenerateXml(locale, wb, xColumn, vColumn) {
-    if (!wb) throw Error(`No workbook`);
-    const sheet = wb.Sheets[wb.SheetNames[0]];
-    const headers = xlsHeaders(wb);
+  if (!wb) throw Error(`No workbook`);
+  const sheet = wb.Sheets[wb.SheetNames[0]];
+  const headers = xlsHeaders(wb);
   const xNumber = headers.findIndex((v) => v === xColumn); // col # for XPath
   const vNumber = headers.findIndex((v) => v === vColumn); // col # for Value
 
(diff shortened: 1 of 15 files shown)


tools/cldr-apps/js/src/esm/cldrBulkConverter.mjs
```diff
@@ -0,0 +1,243 @@
...
+        <types/>
+    </localeDisplayNames>
+    <units>
+        <unitLength type="long"/>
+    </units>
+    <annotations/>
+    <characters/>
+    <numbers>
+        <miscPatterns numberSystem="latn"/>
+    </numbers>
```
srl295, line 97: realized this skeleton needs to be here to keep the XML in order… almost 100% from client!

macchiati:
looks good to me. Still draft?

srl295:
Thanks. Yes. It doesn’t have Dtd ordering so fails.

tools/cldr-apps/js/src/esm/cldrBulkConverter.mjs
srl295: fixed w/ leniency on cldrfile

srl295:
@macchiati ptal, i'm posting it to staging now

UI: 
image

image_720


image

srl295:
image
 updated wording slightly

macchiati (approved):

macchiati:
BTW, I think it would have been easier to adjust the bulk upload to take a <xpath, value> TSV file as well as an XML file. That would be useful for many other reasons, and we wouldn't have to have the complications of generating compliant XML in javascript (eg the ordering of items etc.)

srl295:
> BTW, I think it would have been easier to adjust the bulk upload to take a <xpath, value> TSV file as well as an XML file. That would be useful for many other reasons, and we wouldn't have to have the complications of generating compliant XML in javascript (eg the ordering of items etc.)

That could be a way to do that, except that the bulk upload is a JSP, was trying to not modify that (it needs a full modernize). Didn't end up having to do anything for ordering other than change the bulk upload to not validate XML, which will help anyone else doing a bulk upload also.

This way in the future we can make the bulk upload interactive - the client side can (via voting api) check each item and verify what you want to do with it.  When you throw the whole file over to the server, then it's harder to make a UI to work with certain parts.

PR #3099, opened 2023-07-17
CLDR-16683 if English is a sideways inherit, omit it rom ST

srl295:
- tweak UI for omitted English

CLDR-16683

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -560,7 +560,18 @@ public class DataPage {
             baselineStatus = resolver.getBaselineStatus();
 
             rawEnglish = comparisonValueFile.getStringValue(xpath);
-            displayName = getBaselineProcessor().processForDisplay(xpath, rawEnglish);
+
+            Output<String> pathWhereFound = new Output<String>(),
+                    localeWhereFound = new Output<String>();
+            comparisonValueFile.getStringValueWithBailey(xpath, pathWhereFound, localeWhereFound);
+            final boolean samePath = xpath.equals(pathWhereFound.value);
+            if (!samePath) {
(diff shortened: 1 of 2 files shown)


srl295:
image

also, https://st.unicode.org/cldr-apps/v#/fa/Fields/1d45310cbcf1b2e5 uses the same UI (missing english display string)

pedberg-icu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
```diff
@@ -560,7 +560,18 @@ public DataRow(String xpath) {
             baselineStatus = resolver.getBaselineStatus();
 
             rawEnglish = comparisonValueFile.getStringValue(xpath);
-            displayName = getBaselineProcessor().processForDisplay(xpath, rawEnglish);
+
+            Output<String> pathWhereFound = new Output<String>(),
```
macchiati, line 564: If you are not replacing the getStringValue, then it is slightly faster/cleaner to replace

```
Output<String> pathWhereFound = new Output<String>(),
     localeWhereFound = new Output<String>();
comparisonValueFile.getStringValueWithBailey(xpath, pathWhereFound, localeWhereFound);
final boolean samePath = xpath.equals(pathWhereFound.value);
```
with
```
final CLDRFile.Status status = new Status();
ENGLISH.getSourceLocaleID(path, status);
 final boolean samePath2 = path.equals(status.pathWhereFound);
```
But approving in the interests of getting this done fast.

macchiati (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
srl295: That makes more sense, I didn't understand why we needed the Bailey value at all

PR #3106, opened 2023-07-19
CLDR-16886 add test for PathDescription

srl295:
- broken in CLDR-16704

CLDR-16886

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
@@ -143,9 +143,9 @@ public class PathDescription {
                     + ".\n"
                     + "^//ldml/characters/exemplarCharacters$"
                     + RegexLookup.SEPARATOR
-                    + String.format(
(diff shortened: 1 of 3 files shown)


srl295:
Extremely bizarre issue.

The first commit fails for me locally.   More discussion and minimum test case at https://gist.github.com/srl295/733e6436bad7d170e7f1a00275acab48?permalink_comment_id=4634496#gistcomment-4634496 

Looks like a java compiler bug from the excessive string concatenation, though I can't repro it consistently (outside of VSCode!) 

This puts in:
1. a workaround that shouldn't be necessary in PathDescription.java
3. removes the silent swallowing of parse errs in RegexLookup
4. a test to fail hard if there's a problem in the future

It's another argument for CLDR-16755 - moving this data OUT of Java.

macchiati (approved):

macchiati:
It's another argument for [CLDR-16755](https://unicode-org.atlassian.net/browse/CLDR-16755) - moving this data OUT of Java.
!!

PR #3108, opened 2023-07-21
CLDR-15990 Generate .xsd and .md from .dtd

srl295:
CLDR-15990

- Updates `GenerateDtd`, which is normally run automatically as part of BRS, to do two things:
1. Generate a `.xsd` file for each `.dtd` with [XML Schema](https://www.w3.org/TR/xmlschema) for the DTD files. These are checked into the repository and intended to be released with CLDR.
2. Generate a `.md` file in `docs/ldml/dtd2md/` with a shell that can be copied into the specification for each element.  This directory is `.gitignored` and the resulting files are not added to the repository.

- Updates `XMLValidator` to add a `--schema=` option allowing it to be used to validate against a schema.

- Note that DTD @-annotations such as `@MATCH` are not supported by the converter at present, so such comments are removed from the .xsd

- https://github.com/unicode-org/cldr/pull/2350 is the original PR

-----

Example of  the generated `.md` content:

```md
### <a name="Element_languages" href="#Element_languages">Element: languages</a>

> <small>
>
> Parents: [localeDisplayNames](#Element_localeDisplayNames)
>
> Children: [alias](#Element_alias), [language](#Element_language), [_special_](tr35.md#special)
>
> Occurrence: ?, ?
> </small>

_Attribute:_ `draft`

_Attribute:_ `references`

_Attribute:_ `standard`

_Attribute:_ `validSubLocales`

* * *
```

ALLOW_MANY_COMMITS=true


common/dtd/ldml.xsd
```diff
@@ -0,0 +1,8624 @@
...
+            <xs:enumeration value="true"/>
+            <xs:enumeration value="false"/>
+          </xs:restriction>
+        </xs:simpleType>
+      </xs:attribute>
+    </xs:complexType>
+  </xs:element>
+  <!-- @METADATA -->
+  <!-- @DEPRECATED -->
+  <!-- @METADATA -->
```
pedberg-icu, line 59: Out of curiosity, what does it mean to have the two successive `@METADATA/@DEPRECATED` annotations here?

common/dtd/ldmlSupplemental.xsd
```diff
@@ -0,0 +1,2922 @@
...
+            <xs:enumeration value="provisional"/>
+            <xs:enumeration value="unconfirmed"/>
+            <xs:enumeration value="true"/>
+            <xs:enumeration value="false"/>
+          </xs:restriction>
+        </xs:simpleType>
+      </xs:attribute>
+    </xs:complexType>
+  </xs:element>
+  <!-- @DEPRECATED -->
```
pedberg-icu, line 1972: Still confused about how these annotations are being used. Is there a mapping from the sequence of annotations to items inside the previous block?

pedberg-icu (approved):
It all looks reasonable, but I did not check the xsd files in detail, I assume they are generated. I have a a general question about what the annotations like @DEPRECATED mean in these xsd files and how they are used, but that is not blocking.

macchiati (changes requested):
I looks like the xsd is tossing in all the annotations way after what they apply to. Also, the description for the PR needs much more information about what is being done and why.

srl295:
This conversion doesn’t handle annotations yet. So it’s not suitable for
replacing CLDR validation. However it is useful for clients checking data
that’s already from CLDR. And I can keep iterating it. I’d need to mutate
the xsd based on dtd data.

srl295:
> I looks like the xsd is tossing in all the annotations way after what they apply to. 

They are just comments. It's probably grouping them to the item they follow. Perhaps a step would be to just remove the annotation comments. 

> Also, the description for the PR needs much more information about what is being done and why.

Makes sense. Thanks.

common/dtd/ldmlSupplemental.xsd
srl295: The annotations aren't used, I've now removed them.

common/dtd/ldml.xsd
srl295: These are now removed.

srl295:
ping… can i get a review on this? it now removes the annotation comments.

macchiati (approved):

PR #3109, opened 2023-07-21
CLDR-15034 DTD @MATCH: add validity/bcp47 and semver match types

srl295:
For the Keyboard spec, we needed to match two additional `@MATCH` types to the DTD.

- `@MATCH:semver` matches a semantic version ([semver.org](https://semver.org)) such as `1.0.0` or `1.2.3-BETA`  - this is used for the keyboard version   
- `@MATCH:validity/bcp47` matches any valid bcp47 id, such as `nod-Lana` or `de-CH-t-k0-windows-extended-var` - this is used for the keyboard locale id
- plus tests

(cherry picked from commit 632c286d8abd7c20a082e749cf67c3b0b264816a)

CLDR-15034

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
@@ -114,6 +114,7 @@ public abstract class MatchValue implements Predicate<String> {
         }
     }
 
+    /** Check that a bcp47 locale ID is well-formed */
     public static class BCP47LocaleMatchValue extends MatchValue {
         static final UnicodeSet basechars = new UnicodeSet("[A-Za-z0-9_]");
 
@@ -131,10 +132,14 @@ public abstract class MatchValue implements Predicate<String> {
             try {
                 ULocale l = ULocale.forLanguageTag(item);
                 if (l == null || l.getBaseName().isEmpty()) {
-                    return false;
+                    return false; // failed to parse
                 }
+
+                // check with lstr parser
+                LanguageTagParser ltp = new LanguageTagParser();
+                ltp.set(item);
             } catch (Throwable t) {
-                return false;
+                return false; // string failed
             }
             return true;
         }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
@@ -122,7 +122,7 @@ public abstract class MatchValue implements Predicate<String> {
 
         @Override
         public String getName() {
-            return "validity/bcp47";
+            return "validity/bcp47-wellformed";
         }
 
         @Override
@@ -325,7 +325,7 @@ public abstract class MatchValue implements Predicate<String> {
             if (typeName.equals("locale")) {
                 return new LocaleMatchValue();
             }
-            if (typeName.equals("bcp47")) {
+            if (typeName.equals("bcp47-wellformed")) {
                 return new BCP47LocaleMatchValue();
             }
             int slashPos = typeName.indexOf('/');
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
@@ -114,11 +114,11 @@ public abstract class MatchValue implements Predicate<String> {
         }
     }
 
-    /** Check that a bcp47 locale ID is well-formed */
-    public static class BCP47LocaleMatchValue extends MatchValue {
+    /** Check that a bcp47 locale ID is well-formed. Does not check validity. */
+    public static class BCP47LocaleWellFormedMatchValue extends MatchValue {
         static final UnicodeSet basechars = new UnicodeSet("[A-Za-z0-9_]");
 
-        public BCP47LocaleMatchValue() {}
+        public BCP47LocaleWellFormedMatchValue() {}
 
         @Override
         public String getName() {
@@ -326,7 +326,7 @@ public abstract class MatchValue implements Predicate<String> {
                 return new LocaleMatchValue();
             }
             if (typeName.equals("bcp47-wellformed")) {
-                return new BCP47LocaleMatchValue();
+                return new BCP47LocaleWellFormedMatchValue();
             }
             int slashPos = typeName.indexOf('/');
             Set<Status> statuses = null;
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
@@ -13,6 +13,9 @@ import com.ibm.icu.text.UnicodeSet;
 import com.ibm.icu.text.UnicodeSet.SpanCondition;
 import com.ibm.icu.util.ULocale;
 import com.ibm.icu.util.VersionInfo;
+import com.vdurmont.semver4j.Semver;
+import com.vdurmont.semver4j.Semver.SemverType;
+import com.vdurmont.semver4j.SemverException;
 import java.text.ParseException;
 import java.util.Date;
 import java.util.EnumSet;
@@ -79,6 +82,9 @@ public abstract class MatchValue implements Predicate<String> {
                 case "regex":
                     result = RegexMatchValue.of(subargument);
                     break;
+                case "semver":
+                    result = SemverMatchValue.of(subargument);
+                    break;
                 case "metazone":
                     result = MetazoneMatchValue.of(subargument);
                     break;
@@ -105,6 +111,37 @@ public abstract class MatchValue implements Predicate<String> {
             return result;
         } catch (Exception e) {
             throw new IllegalArgumentException("Problem with: " + originalArg, e);
+        }
+    }
+
+    public static class BCP47LocaleMatchValue extends MatchValue {
+        static final UnicodeSet basechars = new UnicodeSet("[A-Za-z0-9_]");
+
+        public BCP47LocaleMatchValue() {}
+
+        @Override
+        public String getName() {
+            return "validity/bcp47";
+        }
+
+        @Override
+        public boolean is(String item) {
+            if (item.equals("und")) return true; // special case because of the matcher
+            if (item.contains("_")) return false; // reject any underscores
+            try {
+                ULocale l = ULocale.forLanguageTag(item);
+                if (l == null || l.getBaseName().isEmpty()) {
+                    return false;
+                }
+            } catch (Throwable t) {
+                return false;
+            }
+            return true;
+        }
+
+        @Override
+        public String getSample() {
+            return "de-u-nu-ethi";
         }
     }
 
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -13,6 +13,9 @@
 import com.ibm.icu.text.UnicodeSet.SpanCondition;
 import com.ibm.icu.util.ULocale;
 import com.ibm.icu.util.VersionInfo;
+import com.vdurmont.semver4j.Semver;
```
pedberg-icu, line 16: Hmm, any license issues for this code?

pedberg-icu (approved):

pedberg-icu:
Looks good pending any license issues with the "vdurmont" code which I do not remember being used before in CLDR

macchiati (changes requested):
The description for the PR needs much more information about what is being done and why.

miloush:
The https://semver.org/ at the bottom suggests regular expression to check the validity of these version numbers. Wouldn't be better to use the suggested regex rather than introducing a new dependency that relies on throwing exceptions?

srl295:
> The description for the PR needs much more information about what is being done and why.

Sorry. I will work more on these PR descriptions.

srl295:
MIT license

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
srl295: no, it's MIT licensed, and pulled in as a dependency.

srl295:
> The description for the PR needs much more information about what is being done and why.

Done

srl295:
> The https://semver.org/ at the bottom suggests regular expression to check the validity of these version numbers. Wouldn't be better to use the suggested regex rather than introducing a new dependency that relies on throwing exceptions?

- The `@MATCH` check is only used during unit tests of the file.
- We might want to parse the version, using the parser
- We could use the regex instead.

srl295:
@macchiati is the PR clear now?

macchiati (changes requested):
1. Can you explain what semver is doing?
2. The test on bcp47 is not even well-formedness. We have code in CLDR to check for validity of bcp47 language tags, and should use that instead.

srl295:
> * Can you explain what semver is doing?

Semantic Versioning ( https://semver.org ) is a commonly used standard for version numbers. It is used for the version number of keyboards.

> * The test on bcp47 is not even well-formedness. We have code in CLDR to check for validity of bcp47 language tags, and should use that instead.

OK. I will look for that, thanks.

srl295:
> One of the things I wanted to point out is that Semantic Version does not allow traditional 4-part version/build numbers, so things like 10.0.14393.1000 are invalid.

right, although you could use 10.0.14393-BUILD1000 … it is a well known system though, which seems to make sense for files that will be interchanged.

miloush:
Interchange is not a convincing reason to me. You can interchange other syntaxes too.

A better reason is we want an implementation to be able to pick from or sort a list of versions for the user in an order from newest to latest, and symver provides a defined order widely used in industry.

Not that this is not true for the 4 part numbers, but I think anytime we are trying to restrict syntax on something, we should have a good a reason to create such complexity/annoyance and be able to demonstrate that the chosen solution is the least disruptive one to achieve the goal. This includes restricting NMTOKENs to ASCII range and other potential restrictions we currently have in the draft.

In the case of versioning, the way I see it is that there are two reasonably established systems in different parts of the industry and whatever we pick will inconvenient one or the other. So provided we actually need a system that e.g. provides well-defined order, I don't mind which one is picked.

srl295:
Semver also defines not just ordering, but assertions, such as greater than or equal to, and so on. 

I agree that these are both well-established, in fact, ICU and CLDR use the 4-part number as well (four bytes), but there are many variations there.

Most newer software components that I see are using semver, which has worked out quite well especially in package managers and such (i include maven which is almost-semver).

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestMatchValue.java
```diff
@@ -0,0 +1,60 @@
...
+        ImmutableSet<String> is_semver = ImmutableSet.of("0.0.0", "1.0.0", "42.0.0-BETA3");
+        assertAll(
+                "is=true",
+                is_semver.stream().map(v -> () -> assertTrue(m.is(v), v + ": Should be a semver")));
+
+        ImmutableSet<String> not_semver = ImmutableSet.of("0.0", "v1.0", "Some Version", "");
+        assertAll(
+                "is=true",
+                not_semver.stream()
+                        .map(v -> () -> assertFalse(m.is(v), v + ": Should NOT be a semver")));
```
macchiati, line 34: this works, but forEach should be used where you aren't using the results of the match.

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -108,6 +114,37 @@ public static MatchValue of(String command) {
         }
     }
 
+    public static class BCP47LocaleMatchValue extends MatchValue {
```
macchiati, line 117: You should at least check for validity of the LSRV components using the calls as LocaleMatchValue; see 

            lang = new ValidityMatchValue(LstrType.language, statuses, false);

and following.


For the extensions, I checked, and it turns out that we would need to do a bit of work to pull some pieces together to do a full validity check. So that can be a follow-on.

macchiati: I'll file a ticket for the extension validity checking

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
srl295: @macchiati this checks for well-formed bcp47. If we check for valid bcp47, it would not be forwards-compatible with codes assigned that this version of CLDR doesn't know about.

Perhaps in the future we could add more parameters to switch validity on as well?

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestMatchValue.java
srl295: actually the map is to an assertion (an Executable), and they are all rolled up into the `assertAll` call.  This is very handy, because if two out of five fail, it will say so rather than stopping at the first failure.

srl295:
```
org.opentest4j.MultipleFailuresError: is=true (2 failures)
        org.opentest4j.AssertionFailedError: mt-Latn: Should be bcp47 ==> expected: <true> but was: <false>
        org.opentest4j.AssertionFailedError: und-US-u-rg-ustx-tz-uschi: Should be bcp47 ==> expected: <true> but was: <false>
     …
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
macchiati: We update each version of CLDR, and the chances of someone wanting to submit a keyboard for a language between the time that ISO approves a language code and CLDR releases is very small. And a validity check will help prevent malformed locale IDs, a recurrent problem.

srl295: That's true, but tooling from CLDR version, say, v44 _could_ be used to validate a keyboard for a language code not assigned until 2025. So the DTD should not be defined in terms of "bcp47 codes valid at the time of CLDR release".

srl295: Perhaps what we should have is a separate check, that verifies that all keyboards _in CLDR_ have valid codes. That would be separate from passing a DTD `@MATCH` test

srl295: @macchiati  I've added a subtask, https://unicode-org.atlassian.net/browse/CLDR-16950 to enforce valid locale IDs or keyboards which are (or are contemplating being!) included in CLDR's common data. 

Can we merge this or perhaps rename to bcp47-wellformed or something?

macchiati: That works.

srl295: updated to `@MATCH validity/bcp47-wellformed`

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -108,6 +114,42 @@ public static MatchValue of(String command) {
         }
     }
 
+    /** Check that a bcp47 locale ID is well-formed */
+    public static class BCP47LocaleMatchValue extends MatchValue {
+        static final UnicodeSet basechars = new UnicodeSet("[A-Za-z0-9_]");
+
+        public BCP47LocaleMatchValue() {}
```
macchiati, line 121: Let's rename this also, for clarity

macchiati (changes requested):
Other than that, lgtm

macchiati (approved):

srl295:
@macchiati (and all) PTAL. I've also verified that the matcher works within DTDs, evidence below.  ( The actual keyboard DTD update will follow #3114)

```
      Error: (TestPaths.java:608) osx/ar-t-k0-osx.xml/keyboard@locale, expected match to: ⟪validity/bcp47-wellformed⟫ actual: «aa-BB-CCC-DDDD-EEEEE-u-u»
```

given a temporary patch below:

```diff
diff --git a/keyboards/dtd/ldmlKeyboard.dtd b/keyboards/dtd/ldmlKeyboard.dtd
index 847885d76f..47827a0c28 100644
--- a/keyboards/dtd/ldmlKeyboard.dtd
+++ b/keyboards/dtd/ldmlKeyboard.dtd
@@ -14,7 +14,7 @@ Please see CLDR-15034 for the latest information. -->
 
 <!ELEMENT keyboard ( version, generation?, info?, names, settings?, import*, keyMap+, displayMap?, layer*, vkeys*, transforms*, reorders?, backspaces? ) >
 <!ATTLIST keyboard locale CDATA #REQUIRED >
-    <!--@MATCH:any/TODO-->
+    <!--@MATCH:validity/bcp47-wellformed-->
 
 <!ELEMENT version EMPTY >
 <!ATTLIST version platform CDATA #REQUIRED >
diff --git a/keyboards/osx/ar-t-k0-osx.xml b/keyboards/osx/ar-t-k0-osx.xml
index 5ace031716..7af2ede731 100644
--- a/keyboards/osx/ar-t-k0-osx.xml
+++ b/keyboards/osx/ar-t-k0-osx.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE keyboard SYSTEM "../dtd/ldmlKeyboard.dtd">
-<keyboard locale="ar-t-k0-osx">
+<keyboard locale="aa-BB-CCC-DDDD-EEEEE-u-u">
        <version platform="10.9" number="$Revision$"/>
        <names>
                <name value="Arabic"/>
@@ -363,4 +363,3 @@
                <map iso="A03" to=" "/> <!-- space -->
        </keyMap>
 </keyboard>
```

srl295:
@macchiati thanks

PR #3110, opened 2023-07-21
CLDR-15034 add @ALLOWS_UESC DTD Annotation

srl295:
This adds the `@ALLOWS_UESC` DTD annotation, used in the proposed Keyboard DTD.

This annotation on an attribute value indicates that an attribute value MAY contain the `\u{0127}` Unicode escaped format in the attribute.  For example,

```xml
<!ATTLIST key to CDATA #IMPLIED >
    <!--@MATCH:any-->
    <!--@VALUE-->
    <!--@ALLOWS_UESC-->
```

and then:

```xml
<key id="quote" to="\u{22}"/>
```

- Originally #2642 734b93461ff567458d58e0cd1cb9b679d8cbc506

(cherry picked from commit 38d093341c5205a8523aa3882a800a275072a387)

CLDR-15034

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
@@ -160,6 +160,7 @@ public class DtdData extends XMLFileReader.SimpleHandler {
         private final Set<String> commentsPre;
         private Set<String> commentsPost;
         private boolean isDeprecatedAttribute;
+        private boolean isUEscapedAttribute = false;
         public AttributeStatus attributeStatus =
                 AttributeStatus.distinguished; // default unless reset by annotations, or for xml:
         // attributes
@@ -294,6 +295,10 @@ public class DtdData extends XMLFileReader.SimpleHandler {
                     case "@DEPRECATED":
                         isDeprecatedAttribute = true;
                         break;
+                    case "@ALLOWS_UESC":
+                        isUEscapedAttribute = true;
+                        break;
+
                     default:
                         int colonPos = commentIn.indexOf(':');
                         if (colonPos < 0) {
@@ -374,6 +379,10 @@ public class DtdData extends XMLFileReader.SimpleHandler {
 
         public boolean isDeprecated() {
             return isDeprecatedAttribute;
+        }
+
+        public boolean isUEscaped() {
+            return isUEscapedAttribute;
         }
 
         public boolean isDeprecatedValue(String value) {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -160,6 +160,7 @@ public static class Attribute implements Named {
         private final Set<String> commentsPre;
         private Set<String> commentsPost;
         private boolean isDeprecatedAttribute;
+        private boolean isUEscapedAttribute = false;
```
macchiati, line 163: I think this needs to be isUEscap**able**Attribute instead, unless I misunderstand. That is, you are not *requiring* \\u escapes to be used in the field.

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -1402,6 +1414,17 @@ public boolean isDeprecated(String elementName, String attributeName, String att
                 attributeValue); // don't need special test for "*"
     }
 
+    public boolean isUEscaped(String elementName, String attributeName, String attributeValue) {
```
macchiati, line 1417: Same here.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
miloush: If it says "escapable", as a consumer you don't know whether to unescape it or not. If it is "escaped", then it is clear you need to unescape it and that works regardless there is any \\u sequences in the string or not.

srl295: I think the api should be revised to match the annotation we landed on, that is "ALLOWS_UESC".

srl295: Perhaps attributeAllowsUescape?

macchiati: That works (and in function names etc)

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
srl295: done

srl295:
Ping… I think this now follows the review feedback (and this is blocking #3114 )

pedberg-icu (approved):

srl295:
@macchiati  any further issue here?

macchiati (approved):

PR #3121, opened 2023-07-27
CLDR-15296 Fix gasoline-equivalent

macchiati:
CLDR-15296

There was no `@MATCH` for the description attributes

common/supplemental/units.xml
The report was that the unit was wrong. At first I thought that it was the case. Wrote a test to verify, and rechecked the conversion factor, which was off slightly from EPA definition. Added an explanation why the unit for `gasoline-equivalent` seems so odd. 

description="Constructed so that 1 gallon-gasoline-equivalent = 33.705 kWh as per https://www3.epa.gov/otaq/gvg/learn-more-technology.htm"

Added some utilities for testing

Write test to verify that gasoline-equivalent has the right value and unit, and adjust the conversion slightly to match EPA instead of wikipedia. 

The test file contains some additional code for testing units, with the class UnitEquivalence and the method testRelations.

The testRelations takes the "basic" units like newton, watt, meter, second, and find out which of those are related to two others by multiplication or division. It tests a few of them against known cases, but the full printout is available via -v, which can be useful in debugging relationships.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -415,6 +415,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@MATCH:regex/simple-->
     <!--@VALUE-->
 <!ATTLIST unitQuantity description CDATA #IMPLIED >
+    <!--@MATCH:any-->
     <!--@METADATA-->
 
 <!ELEMENT convertUnits ( convertUnit* ) >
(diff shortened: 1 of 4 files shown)


markusicu:
Is CLDR-15296 the right ticket? It's related to units, but it looks like this unit wasn't added in the previous PRs here, or maybe I just missed it.

markusicu (commented):
changes look mostly plausible, except i lack too much context for understanding testRelations(); i could rubber-stamp this but would be happier letting someone approve who knows the ins & outs here

macchiati:
@pedberg-icu I added more information about what testRelations does. Could you take a look at this?

pedberg-icu (approved):

PR #3123, opened 2023-07-30
CLDR-15034 kbd: move pcm keyboard to 3.0 format

srl295:
CLDR-15034

Conversion of the `pcm` keyboard from pre-3.0 to new format.

image

The prior format file is here: https://github.com/unicode-org/cldr/blob/maint/maint-43/keyboards/und/pcm-t-k0-windows.xml

ALLOW_MANY_COMMITS=true


keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,70 @@
...
+			<transform from="O\u{323}" to="Ọ" />
+			<transform from="''" to="\u{323}" /> <!-- Quick way to add dot below -->
+			<transform from="A\u{301}" to="Á" />
+			<transform from="e\u{323}" to="ẹ" />
+			<transform from="O\u{301}" to="Ó" />
+			<transform from="e\u{301}" to="é" />
+			<transform from="e\u{300}" to="è" />
+			<transform from="E\u{301}" to="É" />
+			<transform from="o\u{300}" to="ò" />
+		</transformGroup>
```
miloush, line 68: This is a peculiar order of transforms (and with the `''` thrown in the middle), is that deliberate? It is difficult to see whether all base chars or all combining marks are covered.

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,70 @@
...
+				keys="grave 1 2 3 4 5 6 7 8 9 0 hyphen equal" />
+			<row keys="acute w e r t y u i o p open-square close-square" />
+			<row keys="a s d f g h j k l odot edot slash" />
+			<row keys="slash z c v b n m comma period semi-colon apos" />
+			<row keys="space" />
+		</layer>
+
+		<layer modifier="shift">
+			<row
+				keys="grave bang at hash dollar naira percent amp asterisk open-paren close-paren underscore plus" />
```
miloush, line 38: why new row? (also in the layer above)

srl295: Why the new line? Just standard redhat xml formatter. I'll fix it.

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE keyboard SYSTEM '../dtd/ldmlKeyboard.dtd'>
+<keyboard locale="pcm" conformsTo="techpreview">
+	<version number="1.0.0" />
+	<names>
+		<name value="Nigerian Pidgin" />
+		<name value="Naijíriá Píjin" />
```
miloush, line 7: Yeah even though we said the implementation can pick a name it wants (or the first one), I feel like it would be useful to somehow note down what these values are, or how they differ, similar to `rel` attribute in `link`

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,70 @@
...
+		</layer>
+
+		<layer modifier="shift">
+			<row
+				keys="grave bang at hash dollar naira percent amp asterisk open-paren close-paren underscore plus" />
+			<row keys="acute W E R T Y U I O P open-curly close-curly" />
+			<row keys="A S D F G H J K L Odot Edot" />
+			<row keys="question Z C V B N M open-angle close-angle colon double-quote" />
+			<row keys="space" />
+		</layer>
```
miloush, line 43: Shouldn't there also be a caps layer? I don't see in the spec preview that that would happen automatically.

srl295: Yes, there should, and there is one in the original file

srl295: Good catch

miloush: Maybe we should say something like unless there is an explicit caps layer, caps case-maps everything to uppercase

srl295: It could be a tool feature too though, generate a caps layer based on unicode casing.

srl295:
Can you file a ticket on that ?

keyboards/3.0/pcm.xml
mcdurdin: Leave it to tooling I think -- auto caps layer adds complexity to implementations and necessitates a bunch of orthography assumptions which would be problematic and could lead to frustrating implementation inconsistencies and bugs. Turkish i/ı/İ/I is just the beginning.

keyboards/3.0/pcm.xml
mcdurdin: :+1: `rel` attribute would be very helpful.

keyboards/3.0/pcm.xml
mcdurdin: These transforms (apart from `''`) are just normalization, and should not be included as part of the keyboard -- it should be automatic per the spec.

keyboards/3.0/pcm.xml
miloush: Well the layout does specify a locale that could be used for the case mapping, although we also support noting down alternative locales. The conflicting goal here is to have minimal files. I am perfectly happy with leaving this to tooling rather than the implementation.

keyboards/3.0/pcm.xml
miloush: Didn't we say the specified normalization is descriptive and does not change the output in any way?

srl295: Correct, my understanding is that the normalization is descriptive.

keyboards/3.0/pcm.xml
srl295: great idea, file a ticket ? :)

keyboards/3.0/pcm.xml
srl295: > locale used for case mapping

That's very much locale data, unicode version specific.  yes, that does make the files a little less minimal, but once the file is there it's completely independent of all Unicode versions and all locale data situations, which is best for less user surprise. As probably a bad example but Cherokee and Georgian didn't originally have bicamerality and now they do, so caps behavior could change dramatically.

keyboards/3.0/pcm.xml
mcdurdin: Ticket relating to this: https://unicode-org.atlassian.net/browse/CLDR-16943

keyboards/3.0/pcm.xml
srl295: @miloush do you want to file a ticket on this one?

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,58 @@
...
+      <row keys="A S D F G H J K L Odot Edot slash" />
+      <row keys="slash Z C V B N M comma period semi-colon apos" />
+      <row keys="space" />
+    </layer>
+
+  </layers>
+
+  <transforms type="simple">
+    <transformGroup>
+      <transform from="''" to="\u{323}" /> <!-- Quick way to add dot below -->
```
miloush, line 55: Is this apos? I wonder how intuitive it is to have all these as implicit import

miloush:
Is there are ground truth for this particular layout, like a standard or community website or something that could be referenced/checked against?

keyboards/3.0/pcm.xml
srl295: This isn't an implicit import. this is a double apostrophe, as in the original file.

srl295:
this will have to be rebased once #3274  lands.

@miloush any other concerns with it?  Please compare to the original file: https://github.com/unicode-org/cldr/blob/maint/maint-43/keyboards/und/pcm-t-k0-windows.xml

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,58 @@
...
+      <row keys="acute w e r t y u i o p open-square close-square" />
+      <row keys="a s d f g h j k l odot edot slash" />
+      <row keys="slash z c v b n m comma period semi-colon apos" />
+      <row keys="space" />
+    </layer>
+
+    <layer modifier="shift">
+      <row keys="grave bang at hash dollar naira percent amp asterisk open-paren close-paren underscore plus" />
+      <row keys="acute W E R T Y U I O P open-curly close-curly" />
+      <row keys="A S D F G H J K L Odot Edot" />
```
miloush, line 38: The original file has a `question` at the end of this line, like the other states have `slash`
```suggestion
      <row keys="A S D F G H J K L Odot Edot question" />
```

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,58 @@
...
+      <row keys="question Z C V B N M open-angle close-angle colon double-quote" />
+      <row keys="space" />
+    </layer>
+
+    <layer modifier="caps">
+      <row keys="grave 1 2 3 4 5 6 7 8 9 0 hyphen equal" />
```
miloush, line 44: As far as I can tell, the original file has combining grave in _none_ and _shift_ states and non-combinign one here. In this PR all seem to be the same.

keyboards/3.0/pcm.xml
```diff
@@ -0,0 +1,58 @@
...
+      <row keys="acute W E R T Y U I O P open-square close-square" />
```
miloush, line 45: The original file has Q for _caps_ state
```suggestion
      <row keys="Q W E R T Y U I O P open-square close-square" />
```

miloush:
Let me know if these were intentional differences.

srl295:
@miloush i will get back to you on these… for now, this will need revamping also because it's now out of date due to the DTD

keyboards/3.0/pcm.xml
srl295: good catch, fixed

srl295:
@miloush  and all PTAL

miloush (approved):

PR #3124, opened 2023-07-30
Update ko_KP.xml

Avstraliya:
Added Germany and Russia in the Pyongyang Korean xml

CLDR-_____

ALLOW_MANY_COMMITS=true


AEApple:
Needs a CLDR ticket with justification per https://github.com/unicode-org/cldr/blob/main/docs/requesting_changes.md

PR #3127, opened 2023-07-31
CLDR-16848 fix bulk import

srl295:
- use negative lookbehind in split to preserve slashes in attributes

Example:

//ldml/dates/timeZoneNames/zone[@type="America/Guyana"]/exemplarCity

CLDR-16848

ALLOW_MANY_COMMITS=true


srl295:
@AEApple reports this fixes the issue.

macchiati (approved):

PR #3133, opened 2023-08-01
CLDR-16807 vetting viewer for unaffiliated TC+ gets all locales

srl295:
- to allow error counts for all locales

CLDR-16807

ALLOW_MANY_COMMITS=true


srl295:
takes an extremely long time

srl295:
Mark pointed out that we could _skip_ TC and HC locales

PR #3134, opened 2023-08-01
CLDR-16851 more unit systems

macchiati:
CLDR-16851

- Added the new systems (si_acceptable, metric_adjacent, astronomical, person_age, prefixable), and reordered

common/supplemental/units.xml

- Added the systems as appropriate. There is a worksheet in https://docs.google.com/spreadsheets/d/1V2uSCqvAKdPFgrfw9ahgEYX5V5_pcFgUwI6kGaAR8vk/edit#gid=909900283 with changes that were refined over the course of doing the ticket. 
- Column M was used to eliminate some si measures from being prefixable.
- There was a fix to the conversion factor for steradian

tools/cldr-code/src/main/java/org/unicode/cldr/test/BestMinimalPairSamples.java

- Change the logic to intersect with the si & metric (and variants)

- Added utility for approximatelyEqual

- Changed sourceToSystems to use the enum UnitSystem
- Added the values to UnitSystem
- Added a utility for resolving units
- Change SiOrMetric to encompass (mostly) the old values, to make the best examples computed for minimal pairs be more or less what they were.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/external/nistAcceptedUnits.txt

- Added this file from NIST to get the SI-acceptable units from them.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/ExternalUnitConversionData.java

- Rationalized the structure of this to move all the NIST-related gorp into tools/cldr-code/src/test/java/org/unicode/cldr/unittest/NistUnits.java

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/GenerateNewUnits.java

- Expanded this into a better tool for looking at CLDR vs NIST units

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/NistUnits.java

- Moved other code here, and rationalized.
- Now also reads the nistAcceptedUnits.txt file

- Added testUnitSystems()
- Some other cleanup

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/external/nistAcceptedUnits.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/external/nistAcceptedUnits.txt
@@ -1,4 +1,4 @@
-# From https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-4-two-classes-si-units-and-si-prefixes
+# From https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-5-units-outside-si
 
 # The term "acceptable units," which is introduced in this Guide for convenience, is used to mean the SI units plus 
 # (a) those non-SI units accepted for use with the SI (see Tables 6 - 9); and 
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/external/nistAcceptedUnits.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/external/nistAcceptedUnits.txt
@@ -1,0 +1,69 @@
+# From https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-4-two-classes-si-units-and-si-prefixes
+
+# The term "acceptable units," which is introduced in this Guide for convenience, is used to mean the SI units plus 
+# (a) those non-SI units accepted for use with the SI (see Tables 6 - 9); and 
+# (b) appropriate multiples and submultiples of such accepted non-SI units.
+
+# Modifications
+
+# minute—time, etc => minute
+# degree—plane angle(a) => degree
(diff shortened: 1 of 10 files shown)


macchiati:
Let me know if you need more details about the changes.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/external/nistAcceptedUnits.txt
```diff
@@ -0,0 +1,69 @@
+# From https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-5-units-outside-si
```
srl295, line 1: is this something we need a BRS to keep up to date?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/NistUnits.java
```diff
@@ -27,36 +31,139 @@
...
+            Multimap<String, String> _idChanges,
+            Set<String> _skipping,
+            Map<String, ExternalUnitConversionData> _unitToData,
+            Set<String> _siAcceptable) {
         try {
-            Multimap<String, String> _idChanges = LinkedHashMultimap.create();
-            Set<String> _skipping = new LinkedHashSet<>();
+            // Get the SI acceptable units
+            // Unfortunately, this page has inconsistent formats, so we just mine it for
+            // the systems
```
srl295, line 123: might be good feedback…

srl295 (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/NistUnits.java
macchiati: Not sure how to get it to them... Ideas?

PR #3135, opened 2023-08-01
CLDR-14316 doc: document SurveyTool setup

srl295:
CLDR-14213

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/README.md
+++ b/tools/cldr-apps/README.md
@@ -6,17 +6,49 @@ The `cldr-apps` subproject builds `cldr-apps.war` which contains the Survey Tool
 packaged for deployment. The Survey Tool is used to collect and confirm translations
 for CLDR.
 
+
+## DB setup
+
+- Setup MySQL or MariaDB
+- Create a database (schema), named `cldrdb` with codepage `latin1` and collation `latin1_bin` (Yes, ironic)
+
+```sql
+CREATE SCHEMA `cldrdb` DEFAULT CHARACTER SET latin1 COLLATE latin1_bin ;
+```
+
+- Create a user `surveytool`
+
+```sql
+create user 'surveytool'@'localhost' IDENTIFIED BY 'your_strong_password';
+```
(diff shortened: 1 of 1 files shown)


srl295:
wip…

btangmu:
see also: https://docs.google.com/document/d/1wkJsoQXPcGN_dlCzXzwhfjQoq9Mes7VezIRmNrHWV8c/edit

srl295:
@btangmu I'm planning to leave this open until after the merge (no reason to put it into 44)

srl295:
@btangmu I'd like to merge this in v45  so moving out of draft

srl295:
@btangmu PTAL.  If this is an improvement let's merge it otherwise let me know what you'd like to see.  You had asked about logging for example.

btangmu (approved):
LGTM -- FYI my own local setup has "root" instead of "surveytool" for the mysql user -- the username doesn't matter as long as you're consistent

also, I always import an existing db from a dump/backup rather than creating a new db, and that might be worthwhile to suggest as an alternative in this doc

My server.env also has these two lines:

WLP_DEBUG_SUSPEND=n
WLP_DEBUG_ADDRESS=7777

-- I'm not sure if they're needed, and I don't remember when/why I added them

A link to this doc might also be helpful here, even though it's also outdated: https://docs.google.com/document/d/1wkJsoQXPcGN_dlCzXzwhfjQoq9Mes7VezIRmNrHWV8c/edit#heading=h.ozmz4163qnxu

srl295:
you can run it as root, but i wouldn't recommend it.  particularly, we might miss if a SQL action requires elevated privileges. some of the forum report code manipulates stored procedures for example.

is that document owned by the U+CLDR-TC drive? If not would you consider moving it there?

I don't know what the suspend is, but https://www.ibm.com/docs/en/was-liberty/core?topic=liberty-customizing-environment says that 7777 is the default port.

PR #3140, opened 2023-08-01
CLDR-16945 BRS v44: fix unit tests

srl295:
CLDR-16945

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
@@ -272,6 +272,10 @@ public class TestAnnotations extends TestFmwkPlus {
     }
 
     public void TestUniqueness() {
+        if (logKnownIssue(
+                "CLDR-16947", "skip duplicate TestUniqueness in favor of CheckDisplayCollisions")) {
+            return;
+        }
         Set<String> locales = new TreeSet<>();
         locales.add("en");
         locales.addAll(Annotations.getAvailable());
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
```diff
@@ -272,6 +272,9 @@ public void TestCategories() {
     }
 
     public void TestUniqueness() {
+        if (logKnownIssue("CLDR-16945", "skip duplicate TestUniqueness in favor of CheckDisplayCollisions")) {
+            return;
+        }
```
srl295, line 277: @macchiati this test seems to just duplicate CheckDisplayCollisions, so disabled for now.

pedberg-icu (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
srl295: CLDR-16947 to track

macchiati: I think these were added because the CheckDisplayCollisions wasn’t catching them; so don’t delete them until we can check.

srl295: @macchiati it's not deleted, just logKnownIssue skipped so we can make progress. Unit test failures block seeing any ConsoleCheck results, currently.

PR #3142, opened 2023-08-01
CLDR-16945 BRS v44: update ShowLocaleCoverage

srl295:
- update ShowLocaleCoverage to use TempPrintWriter, so it doesn't zap the file before completion
- also makes sure it shows if there's a difference at the end.

CLDR-16945

ALLOW_MANY_COMMITS=true

Just a little improvement to ShowLocaleCoverage …


srl295:
Example output:

```
…
Found difference in : /Users/srl295/src/cldr-b/common/properties/coverageLevels.txt, /Users/srl295/src/cldr-b/common/properties/38004-coverageLevels.txt
 File1: 'b', 'n ;       modern ;        Bangla'
 File2: 'b', 'lo ;      moderate ;      Anii'
```

Without this change, coverageLocales.txt gets truncated _before_ the tool completes.  If it aborted in the middle, we'd end up with corrupt files.

macchiati (approved):

PR #3144, opened 2023-08-02
CLDR-16945 BRS v44: re-import of data from ST

srl295:
- skipped: aa, mt (would reintroduce errs)

CLDR-16945

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/km.xml
+++ b/common/main/km.xml
@@ -4640,17 +4640,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<symbols numberSystem="khmr">
 			<decimal>,</decimal>
 			<group>.</group>
-			<list draft="contributed">↑↑↑</list>
-			<percentSign draft="contributed">↑↑↑</percentSign>
-			<plusSign draft="contributed">↑↑↑</plusSign>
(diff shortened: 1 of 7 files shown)


common/main/km.xml
```diff
@@ -4640,17 +4640,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<symbols numberSystem="khmr">
 			<decimal>,</decimal>
 			<group>.</group>
-			<list draft="contributed">↑↑↑</list>
```
AEApple, line 4643: So I didn't vote for explicit inheritance here. Should we revert the km import, or should I go in and vote for explicit inheritance?

srl295: That i'm not sure, but i think for now we should just fix it in XML if needed?

AEApple: Wouldn't it just automatically inherit the default numbering system formatting even if it's not explicitly present?

srl295: i'd think so. it'll just be missing next time.

srl295:
@AEApple  OK to merge this?

AEApple (approved):

AEApple:
Yeah, we can always fix km in xml later.

PR #3148, opened 2023-08-03
CLDR-16894 Expand values for contextTransformUsage type, pattern numbers, dateFormatItem/intervalFormatItem id

pedberg-icu:
CLDR-16894

Expand the set of values (as specified in the dtd) for some element attributes:
- contextTransformUsage type attribute, add more of the values specified in TR35 that seem useful (including some already used in Apple data we hope to add)
- pattern numbers attribute, add more numbering systems used for days (at least in Apple data we hope to add)
- dateFormatItem/intervalFormatItem id attribute, use a regex for valid values to include a larger set. I did this instead of adding a special MatchValue function for dateSkeleton, since that would have just used a regex anyway, and since there are differences in the types of skeletons that are valid for intervalFormatItem vs dateFormatItem.

ALLOW_MANY_COMMITS=true


macchiati (dismissed):
Excellent

pedberg-icu:
@macchiati  I made a slight adjustment in the second commit, can you re-stamp please?

macchiati (approved):

PR #3149, opened 2023-08-03
CLDR-16945 BRS v44: additional fixes

srl295:
Most of the deleted items seem to be due to the CLDR-16946 bug

CLDR-16945

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ks_Deva.xml
+++ b/common/main/ks_Deva.xml
@@ -293,7 +293,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 							<day type="tue">बुवार</day>
 							<day type="wed">↑↑↑</day>
 							<day type="thu">↑↑↑</day>
-							<day type="fri">जुम्मा</day>
+							<day type="thu">↑↑↑</day>
 							<day type="sat">↑↑↑</day>
 						</dayWidth>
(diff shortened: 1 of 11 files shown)


common/main/ks_Deva.xml
```diff
@@ -293,7 +293,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 							<day type="tue">बुवार</day>
 							<day type="wed">↑↑↑</day>
 							<day type="thu">↑↑↑</day>
-							<day type="fri">जुम्मा</day>
+							<day type="thu">↑↑↑</day>
 							<day type="sat">↑↑↑</day>
```
btangmu, line 297: is there a logical-group test applicable here?

srl295: duplicate thu here!

srl295: no, the vetters actually shortened the long one so it's shorter than this.  THIS one was a "copy inheritance to hard values" 1-off that we did in xml.   So i'm just dropping this one so that the shorter value from the vetters is inherited.

srl295:
fixed duplicate thursday

PR #3152, opened 2023-08-04
CLDR-16945 Fix systems issue that caused check failures

srl295:
(cherry picked from commit 727ca04c350f2cf3569c98483a433c476d1ae3fc)

CLDR-16945

Note: #3151  but based on main.

ALLOW_MANY_COMMITS=true


srl295:
@macchiati rebased on main (bs no longer changed).  if this works it could merged first, and THEN the other PR should succeed.  Worst case re-run the other PR's failing job.

srl295:
bs and am pass vs. main

srl295:
the bs.xml change may need to be re-applied to the vxml branch

PR #3154, opened 2023-08-04
CLDR-7277 Parallel ConsoleCheck

srl295:
about 2x improvement, could be improved even more especially when targetting batch (non-interactive) mode

CLDR-7277

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
@@ -14,11 +14,9 @@ import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
```diff
@@ -455,30 +432,18 @@ public static void main(String[] args) throws IOException {
         Factory cldrFactory =
                 SimpleFactory.make(sourceDirectories, factoryFilter)
                         .setSupplementalDirectory(new File(CLDRPaths.SUPPLEMENTAL_DIRECTORY));
-        CompoundCheckCLDR checkCldr = CheckCLDR.getCheckAll(cldrFactory, checkFilter);
-        if (checkCldr.getFilteredTestList().size() == 0) {
-            throw new IllegalArgumentException("The filter doesn't match any tests.");
-        }
-        System.out.println("filtered tests: " + checkCldr.getFilteredTests());
+        TestCache testCache = new TestCache(); // for parallelism
+        testCache.setFactory(cldrFactory, checkFilter);
```
srl295, line 436: this is the core change: use TestCache (same as ST! ) to manage the CheckCLDR instances.

This does have an added benefit that we're running more similar code in ConsoleCheck and ST… I don't anticipate divergence otherwise, but it does mean the environ is similar.

srl295:
should be feature-complete. i'm not sure how to use the html and example features, but it should call the same code as before.

srl295:
Hrm, this might conflict with CLDR-17289 somehow.

btangmu:
"might conflict with [CLDR-17289](https://unicode-org.atlassian.net/browse/CLDR-17289) somehow" -- it shouldn't be a problematic merge conflict if that's what you mean -- only one file in common, CheckCLDR.java, and only a few lines

or do you mean a conflict in some other sense?

srl295:
> "might conflict with [CLDR-17289](https://unicode-org.atlassian.net/browse/CLDR-17289) somehow" -- it shouldn't be a problematic merge conflict if that's what you mean -- only one file in common, CheckCLDR.java, and only a few lines
> 
> or do you mean a conflict in some other sense?

Just being cautious.  Somehow i thought this was merged but it wasn't.

This would be great to get in, brings a bit of a speedup.

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
```diff
@@ -455,450 +479,503 @@ public static void main(String[] args) throws IOException {
...
-            }
-            checkCldr.setCldrFileToCheck(file, new Options(options), result);
+                    List<CheckStatus> result = new ArrayList<>();
+                    Set<PathHeader> paths = new TreeSet<>(); // CLDRFile.ldmlComparator);
+                    Map<String, String> m = new TreeMap<>();
+                    Map<String, String> options = new HashMap<>();
+                    FlexibleDateFromCLDR fset = new FlexibleDateFromCLDR();
 
-            subtotalCount.clear();
+                    if (CLDRFile.isSupplementalName(localeID)) return;
```
macchiati, line 557: We should check to see if CLDRFile.isSupplementalName should be used anymore. It is a holdover from when the supplemental data files were in the same folder as the locale-based files (en.xml, etc).

srl295: => CLDR-17291

PR #3155, opened 2023-08-04
CLDR-15836 build: run check and build (tests) in parallel

srl295:
CLDR-15836

Note that this is parallelizing the GitHub actions, not parallizing within console check.

*Bottom Line:* looks like this runs in ~19m instead of ~32m+ total. So about 13m shorter.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -29,6 +29,8 @@ jobs:
         with:
           lfs: false
           fetch-depth: 0  # fetch all branches so that Spotless can resolve `origin/main`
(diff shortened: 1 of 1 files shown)


srl295:
@Manishearth any comments on the workflow side?

Our workflow is like this right now:

image


This changes it to run the 'check' in parallel: 

image

Manishearth:
Yeah, parallelizing is nice.

srl295:
should speed up all builds, because the maven caches are separated properly.

macchiati (approved):

PR #3162, opened 2023-08-09
CLDR-15269 Fix dalton, regenerate tests, and do a bit of code cleanup.

macchiati:
CLDR-15269

common/supplemental/units.xml
- Change the value and quantity of dalton (to mass). Use AMU constant to indicate approximation

In order to make the tests pass, I had to regenerate the units testData. 
That exposed that some changes to the charts could introduce problems for users of the test data. 
I also ended up doing a refactoring to make the enums clearer for printing different kinds of results.
And found that the approximate sign (~) was not always being included.

common/testData/units/unitsTest.txt
- Regenerate (haven't done in a while)
- Many of the value changes are just reformatting, others are from new units

- Refactored FormatStyle enum to make clearer names

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartUnitConversions.java
- Refactored FormatStyle enum to make clearer names
- Grouped units differently since we have more systems

- Cleaned up the formatting/parsing
- Make sure we can parse what we format
- Refactored FormatStyle enum to make clearer names
- Clean up the formatting so that we always emit ~ if the value is approximate
- Drop the thin-space around /, since that makes the format in the testdata less familiar

- Refactored FormatStyle enum to make clearer names

- Fix for testdata header to give right instructions
- Refactored FormatStyle enum to make clearer names
- Added TestRationalFormatting, since we didn't have a test.

ALLOW_MANY_COMMITS=true


srl295 (approved):
some interesting formatter work… maybe upstream to icu4j eventually?

PR #3176, opened 2023-08-11
CLDR-16982 Fix the likely test generation to be script-first

macchiati:
CLDR-16982

ALLOW_MANY_COMMITS=true


macchiati:
This is a draft PR so that Frank can check out the change to the file in ICU

docs/ldml/tr35.md
```diff
@@ -2235,19 +2231,21 @@ This allows for implementations that use those denormalized subtags to use the d
 The reverse operation removes fields that would be added by the first operation.
 
 1. First get max = AddLikelySubtags(inputLocale). If an error is signaled, return it.
```
FrankYFTang, line 2233: Could we also clarify what is the "it" in "return it" here? Is the "it" referring to "error" or referring to the "input string"?  For example, if the input is "qaa-CH-u-nu-latn", what should the algorithm return ?  
A. "qaa-CH-u-nu-latn"
B. "und"
C. an error

sffc:
ICU4X sample impl: https://github.com/unicode-org/icu4x/pull/3874

docs/ldml/tr35.md
```diff
@@ -2191,18 +2191,22 @@ This operation is performed in the following way.
...
-   3. _language<sub>s</sub>\_script<sub>s</sub>_
+   2. _language<sub>s</sub>\_script<sub>s</sub>_
+   3. _language<sub>s</sub>\_region<sub>s</sub>_
    4. _language<sub>s</sub>_
-   5. und\__script<sub>s</sub>_
 3. **Return**
-   1. If there is no match, either return
-      1.  an error value, or
-      2.  the match for "und" (in APIs where a valid language tag is required).
+   1. If there is no match, signal an error and stop.
```
FrankYFTang, line 2198: I think I find another bug in this algorithm. This algorithm does not return earlier if we already have LSR and do not need to maximize. Because of that , if there are no match following the lookup, we will still return error. For example, 

qaa-Cyrl-CH ;	FAIL ;	 ;	

This is what the algorithm requested (because the spec currently does not have an early return for LSR)
but in ICU and ICU4X implementation, we currently just reutrn. But why should we returan error for this case? How about if we have qaa-Cyrl-CH-u-nu-thai? we should just return qaa-Cyrl-CH-u-nu-thai right?

docs/ldml/tr35.md
```diff
@@ -2225,26 +2229,29 @@ For example, if und_AF ⇒ fa_Arab_AF, then:
...
 
-1. First get max = AddLikelySubtags(inputLocale). If an error is signaled, return it.
-2. Remove the variants from max.
+1. First get max = AddLikelySubtags(inputLocale). If an error is signaled AddLikelySubtags, signal that same error and stop.
+2. Remove the variants and extensions from max.
 3. Get the components of the max (_languagemax_, _scriptmax_, _regionmax_).
 4. Then for _trial_ in {_languagemax_, _languagemax_regionmax_, _languagemax_scriptmax_}
-   * If AddLikelySubtags(_trial_) = max, then return _trial_ + variants.
-5. If you do not get a match, return max + variants.
+   * If AddLikelySubtags(_trial_) = max, then return _trial_ + variants + extensions.
```
FrankYFTang, line 2242: Since any call to AddLikelySubtags(_trial_) may get an error to indicate no match, we need to define what the algorith should do in that case. Without specifying that may imply the error should be propagated up and terminate the lookup loop. We should explicitly mention the error from the calling of AddLikelySubtags(_trial_) should not be propogate up but ignored and consider no matching only and continue the next look up.

FrankYFTang: For example, consider the case of 
Remove Likely Subtags ("qaa-Cyrl-CH-u-nu-thai") here, what will happen?

macchiati:
I agree that we should just skip out if all three fields are filled. I'd
considered how to do this, but as you point out, it is really necessary.
Will add.

macchiati:
Fixed the problem that Frank found. So if the tests pass I think we are good to go.

macchiati:
Changes in https://github.com/unicode-org/cldr/pull/3176/commits/43e7e984b711ebac79c74be993cdf5a432dcb5c3

macchiati:
Done, please review.

macchiati:
I'd like to go ahead and merge this, then make any necessary changes in a follow-on PR.

FrankYFTang (approved):

PR #3183, opened 2023-08-15
CLDR-16831 fix comma to pipe in annotations

srl295:
CLDR-16831

- update DAIP
- exclude family group
- re-run `CLDRModify -fp` to update en and all others
 (note: contains other CLDRModify -fp changes)

ALLOW_MANY_COMMITS=true


common/annotations/af.xml
```diff
@@ -811,9 +811,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🤥" type="tts">gesig wat jok</annotation>
 		<annotation cp="🫨">aardbewing | gesig | gesig wat skud | skok | vibreer</annotation>
 		<annotation cp="🫨" type="tts">gesig wat skud</annotation>
-		<annotation cp="🙂‍↔">kop wat horisontaal skud | nee, skud kop</annotation>
```
AEApple, line 814: So the Afrikaans commas look intentional, but I also expect that since these are keywords they're probably fine.

srl295: hm. these may need re-review next time then.

This particular comma was present in the en, though, so it may just be an intentional translation.

AEApple: Yes, I think that one is probably intentional in English and many other languages

common/annotations/en.xml
```diff
@@ -821,9 +821,9 @@ annotations.
 		<annotation cp="🫨">earthquake | face | shaking | shock | vibrate</annotation> <!-- 1FAE8 -->
 		<annotation cp="🫨" type="tts">shaking face</annotation>
 		<!-- Generated lines from Emoji Data v15.1, using GenerateCldrData.java -->
-		<annotation cp="🙂‍↔">head shaking horizontally | no, shake</annotation> <!-- 1F642 200D 2194 -->
+		<annotation cp="🙂‍↔">head shaking horizontally | no | shake</annotation> <!-- 1F642 200D 2194 -->
 		<annotation cp="🙂‍↔" type="tts">head shaking horizontally</annotation>
-		<annotation cp="🙂‍↕">head shaking vertically | nod, yes</annotation> <!-- 1F642 200D 2195 -->
+		<annotation cp="🙂‍↕">head shaking vertically | nod | yes</annotation> <!-- 1F642 200D 2195 -->
```
srl295, line 826: @AEApple I'm not so sure.  There's "no, shake" and "nod, yes" but i think these should have been `no | shake` and `nod | yes`.

srl295:
- exclude name (`tts`) 
- use [cldrmodify](https://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-passes/cldrmodify-using-config-file) with https://docs.google.com/spreadsheets/d/1vJxttdRRCDP3PtPhNK_lECvJMhwIaW7GEdhHE_IxO-o/edit#gid=0

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -610,6 +610,10 @@ private String processLocaleSpecificInput(String path, String value, boolean isU
 
     private String processAnnotations(String path, String value) {
         if (AnnotationUtil.pathIsAnnotation(path)) {
+            if (!path.contains("🧑‍🧒")) {
+                // if not a 'family' sequence
+                value = value.replaceAll(",", "|");
```
macchiati, line 615: That is not what we talked about in the meeting. This is also affecting "tts" paths, and it shouldn't. Example:

		<annotation cp="😁" type="tts">шырока ўсміхаецца</annotation>

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
srl295: Mark, look at the history. This predates the meeting, which consensus is recorded here: https://github.com/unicode-org/cldr/pull/3183#issuecomment-1680896238

I've moved this back to draft to clarify.

macchiati:
Ah, that explains it.

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
srl295: PTAL

srl295:
@macchiati this is in the spreadsheet but will cause dup keywords

> `locale=nn; | action=add; | new_value=familie: vaksen, barn, barn; | new_path=//ldml/annotations/annotation[@cp="🧑‍🧒‍🧒"]`

macchiati:
There is a cldr modify pass that cleans up the annotations, and duplicates within the value for a single <locale,path> are just removed. Is that what you mean?

srl295:
> There is a cldr modify pass that cleans up the annotations, and duplicates within the value for a single <locale,path> are just removed. Is that what you mean?

No, I mean the commas were of this form in the English: `Family: adult, child, child` but changing these to vertical bars makes it `Family: adult | child | child` which will then become just `Family: adult | child` which doesn't seem intended.

My DAIP code excluded the family entries, but they were present in the spreadsheet.

Or is your spreadsheet only proposing changes where there is `Bad English` in the first column?

macchiati:
The non-tts paths are a list of search terms. Searching for [child child]
is the same as searching for [child], so cldr modify eliminates the
duplication. That being said, this should really be family | adult | child
in English.

srl295:
> The non-tts paths are a list of search terms. Searching for [child child] is the same as searching for [child], so cldr modify eliminates the duplication. That being said, this should really be family | adult | child in English.
> […](#)

@macchiati in the "comma problems" spreadsheet, was I supposed to only apply the rows marked "Bad English"? or all rows?

srl295:
@macchiati please re-review

macchiati (approved):

PR #3185, opened 2023-08-15
🚧 CLDR-16947 TestUniqueness passes when re-enabled

srl295:
CLDR-16947

ALLOW_MANY_COMMITS=true

The test passes, but is it valuable? It did not find any collisions that weren't found by CheckDisplayCollisions.

If it does something valuable, I think it should be moved into a console check

### DIFF
--- a/common/annotations/de_CH.xml
+++ b/common/annotations/de_CH.xml
@@ -24,7 +24,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="🏾" type="tts">↑↑↑</annotation>
 		<annotation cp="🏿">↑↑↑</annotation>
 		<annotation cp="🏿" type="tts">↑↑↑</annotation>
-		<annotation cp="‾">Oberstrich | Überstrich</annotation>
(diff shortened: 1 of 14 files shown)


srl295:
@macchiati thoughts? My evaluation is complete, passing the ticket back to you. (You can pass it back to me if you tell me what the recommendation is!)

macchiati:
I agree. The main reason I wanted this ticket was to verify that there wasn't anything that snuck through. So we should be good to just close this ticket.

srl295:
> I agree. The main reason I wanted this ticket was to verify that there wasn't anything that snuck through. So we should be good to just close this ticket.

That will leave the test disabled as a log known issue.  Let's call this option "A"

Option "B" would be to delete the test, since it seems to duplicate the collision tests

Option "C" would be to reinstate the test, without the warning (which is what this PR does, though I'm not proposing this). 

I was proposing B, which were you proposing? "A" seems suboptimal if we close the ticket also.

macchiati:
Ah, yes. I meant option B.

macchiati (changes requested):
Agreed on option B, remove the test in favor of CheckDisplayCollisions

PR #3187, opened 2023-08-16
CLDR-16986 Restore Persian calendar time formats for ckb_IR

DraganBesevic:
CLDR-16986

Restore Persian calendar time formats for ckb_IR, since Persian calendar is default for IR, which is 24 hour calendar cycle, unlike IQ which is default region for ckb.

ALLOW_MANY_COMMITS=true


pedberg-icu:
This is repeating work that was done in CLDR 43 per https://github.com/unicode-org/cldr/pull/2614

pedberg-icu (approved):

PR #3192, opened 2023-08-16
CLDR-16995 Add space between variable $IPA_CONSONANT and combining mark

pedberg-icu:
CLDR-16995

In the new fa-fa_FONIPA transform, one usage of the new variable $IPA_CONSONANT was immediately followed by a combining mark U+0651, so that combination was interpreted as a single undefined variable name. Just added a space to separate the U+0651

ALLOW_MANY_COMMITS=true


srl295:
Need a ticket for next time to catch this in CLDR

macchiati (approved):

macchiati:
What we need is to catch when a variable is undefined (in this case [$IPA_CONSONANT and combining mark](https://github.com/unicode-org/cldr/pull/3192/commits/e8694aead04dc0505ef500a4cf3ff685d76eacc5)).

PR #3193, opened 2023-08-16
CLDR-16831 update FileReaders (used by CLDRModify -fk) to take paths

srl295:
- CLDRModify -fk has a -k option which takes a path relative to a class
- update FileReaders so that if the path starts with a slash (/), it's interpreted as a normal absolute file path

CLDR-16831

ALLOW_MANY_COMMITS=true

This way the cldrmodify input doesn't need to be in the repo

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/FileReaders.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/FileReaders.java
@@ -23,7 +23,22 @@ public class FileReaders {
         return openFile(class1, file, StandardCharsets.UTF_8);
     }
 
+    /**
+     * @param file path of file relative to class1, or, absolute file if starts with `/`
+     * @return
+     */
     public static BufferedReader openFile(Class<?> class1, String file, Charset charset) {
+        if (charset == null) {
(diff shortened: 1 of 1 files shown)


srl295:
useful locally but breaks other things. I'll revisit.

tools/cldr-code/src/main/java/org/unicode/cldr/util/FileReaders.java
```diff
@@ -23,7 +23,22 @@ public static BufferedReader openFile(Class<?> class1, String file) {
...
     public static BufferedReader openFile(Class<?> class1, String file, Charset charset) {
+        if (charset == null) {
+            charset = StandardCharsets.UTF_8;
+        }
+        if (file.startsWith("/")) {
+            try {
+                return new BufferedReader(
+                        new InputStreamReader(new FileInputStream(file), charset), 1024 * 64);
+            } catch (FileNotFoundException e) {
+                throw new IllegalArgumentException("Could not open file " + file, e);
```
macchiati, line 39: Minor, but should probably be UncheckedIOException(...) // I think that's the name
Otherwise looks good.

PR #3199, opened 2023-08-17
CLDR-17003 fix the currency formats unit pattern

macchiati:
CLDR-17003

Uses CLDRModify -fk to make the changes. There is also a minor tool change to SearchXML, so that the search can be restricted to only locales inheriting directly from root.

ALLOW_MANY_COMMITS=true


macchiati:
Ping

srl295 (approved):

PR #3201, opened 2023-08-17
CLDR-16994 BRS: Algorithmic Locales

srl295:
CLDR-16994

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/sr_Latn.xml
+++ b/common/main/sr_Latn.xml
@@ -13,9 +13,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	</identity>
 	<localeDisplayNames>
 		<localeDisplayPattern>
-			<localePattern>{0} ({1})</localePattern>
-			<localeSeparator>{0}, {1}</localeSeparator>
-			<localeKeyTypePattern>{0}: {1}</localeKeyTypePattern>
+			<localePattern>↑↑↑</localePattern>
+			<localeSeparator>↑↑↑</localeSeparator>
+			<localeKeyTypePattern>↑↑↑</localeKeyTypePattern>
 		</localeDisplayPattern>
 		<languages>
 			<language type="aa">afarski</language>
@@ -63,6 +63,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="bem">bemba</language>
 			<language type="bez">bena</language>
 			<language type="bg">bugarski</language>
+			<language type="bgc">harijanski</language>
 			<language type="bgn">zapadni belučki</language>
 			<language type="bho">bodžpuri</language>
 			<language type="bi">bislama</language>
@@ -99,8 +100,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="chr">čeroki</language>
 			<language type="chy">čejenski</language>
 			<language type="ckb">centralni kurdski</language>
-			<language type="ckb" alt="menu">centralni kurdski</language>
-			<language type="ckb" alt="variant">centralni kurdski</language>
+			<language type="ckb" alt="menu">↑↑↑</language>
+			<language type="ckb" alt="variant">↑↑↑</language>
 			<language type="clc">čilkotin</language>
 			<language type="co">korzikanski</language>
 			<language type="cop">koptski</language>
@@ -532,7 +533,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="twq">tasavak</language>
 			<language type="ty">tahićanski</language>
 			<language type="tyv">tuvinski</language>
-			<language type="tzm">centralnoatlaski tamazigt</language>
+			<language type="tzm">centralnoatlaski tamašek</language>
 			<language type="udm">udmurtski</language>
 			<language type="ug">ujgurski</language>
 			<language type="uga">ugaritski</language>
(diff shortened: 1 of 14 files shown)


srl295:
@pedberg-icu @btangmu hant/hans puzzler:

yue(_Hant) has:

```xml
<annotation cp="≈" type="tts">約等於</annotation>
<annotation cp="≒" type="tts">大約等於</annotation>
```

yue_Hans turned into:
```xml
<annotation cp="≈" type="tts">大约等于</annotation>
<annotation cp="≒" type="tts">大约等于</annotation>
```

there's a 'retain' policy here, and so it looks like the 大 was retained in the Hans even though the Hant dropped it.

I'll fix this by changing yue_Hans to:
`<annotation cp="≈" type="tts">约等于</annotation>`

which looks like it matches the expected change in yue

macchiati (commented):
There is something funny going on. I'm seeing a lot of cases where the values are identical to last time, but draft="contributed" is added; effectively downgrading the items. Can you check where these are coming from?

common/main/sr_Latn.xml
```diff
@@ -2875,22 +2878,22 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
-					<relativeTimePattern count="one">pre {0} dana</relativeTimePattern>
-					<relativeTimePattern count="few">pre {0} dana</relativeTimePattern>
+					<relativeTimePattern count="one">↑↑↑</relativeTimePattern>
+					<relativeTimePattern count="few">↑↑↑</relativeTimePattern>
 					<relativeTimePattern count="other">pre {0} dana</relativeTimePattern>
 				</relativeTime>
 			</field>
 			<field type="day-short">
 				<displayName>d.</displayName>
-				<relative type="-2">prekjuče</relative>
```
macchiati, line 2889: This is very suspicious. sr_Latn inherits directly from root, so replacing by ↑↑↑ means throwing away prekjuče and replacing by a root value.

macchiati (changes requested):

btangmu:
@srl295 dropping 大 in yue_Hans 大约等于 for consistency with yue(_Hant) 約等於 looks OK to me. I'm not familiar with the "retain" policy or how CLDR derives yue_Hans from yue, and I don't know whether they're always supposed to match each other except for the orthography.

BTW, it seems sort of random that they dropped 大 for ≈ (U+2248 ALMOST EQUAL TO) but not for ≒ (U+2252 APPROXIMATELY EQUAL TO OR THE IMAGE OF), and it seems dubious that there's nothing corresponding to "IMAGE" (像?) in the annotation for ≒ ...

srl295:
> @srl295 dropping 大 in yue_Hans 大约等于 for consistency with yue(_Hant) 約等於 looks OK to me. I'm not familiar with the "retain" policy or how CLDR derives yue_Hans from yue, and I don't know whether they're always supposed to match each other except for the orthography.
> 
> BTW, it seems sort of random that they dropped 大 for ≈ (U+2248 ALMOST EQUAL TO) but not for ≒ (U+2252 APPROXIMATELY EQUAL TO OR THE IMAGE OF), and it seems dubious that there's nothing corresponding to "IMAGE" (像?) in the annotation for ≒ ...

retain means that the locale isn't regenerated each time, only _missing_ values are copied from Hant to Hans.

common/main/sr_Latn.xml
srl295: 🤔

srl295: @macchiati two things:

1. this conforms to what's in `sr.xml` (see below)
2. this is `day-short` - so it will inherit from `day` which still has these values.


`sr.xml`: 

```xml
			<field type="day">
				<displayName>дан</displayName>
				<relative type="-2">прекјуче</relative>
				<relative type="-1">јуче</relative>
				<relative type="0">данас</relative>
				<relative type="1">сутра</relative>
				<relative type="2">прекосутра</relative>
				<relativeTime type="future">
					<relativeTimePattern count="one">за {0} дан</relativeTimePattern>
					<relativeTimePattern count="few">↑↑↑</relativeTimePattern>
					<relativeTimePattern count="other">за {0} дана</relativeTimePattern>
				</relativeTime>
				<relativeTime type="past">
					<relativeTimePattern count="one">↑↑↑</relativeTimePattern>
					<relativeTimePattern count="few">↑↑↑</relativeTimePattern>
					<relativeTimePattern count="other">пре {0} дана</relativeTimePattern>
				</relativeTime>
			</field>
			<field type="day-short">
				<displayName>д.</displayName>
				<relative type="-2">↑↑↑</relative>
				<relative type="-1">↑↑↑</relative>
				<relative type="0">↑↑↑</relative>
				<relative type="1">↑↑↑</relative>
				<relative type="2">↑↑↑</relative>
```

`sr_Latn.xml`: 

```xml
			<field type="day">
				<displayName>dan</displayName>
				<relative type="-2">prekjuče</relative>
				<relative type="-1">juče</relative>
				<relative type="0">danas</relative>
				<relative type="1">sutra</relative>
				<relative type="2">prekosutra</relative>
```

srl295:
rebased

common/main/sr_Latn.xml
macchiati: Ok, thought it over and that will work. Thanks for the explanation.

macchiati (approved):

PR #3203, opened 2023-08-18
CLDR-16380 BRS Update likely subtags

macchiati:
CLDR-16380

Needing to do it manually this release, so incorporating data from SIL. Also regenerating the test data, which picks up the upgraded DDL locales. Removed deprecated languages, etc.

Also adds likely subtag data for new CLDR Basic locales.

Disabled GenerateMaximalLocales, because it needs fixes to not override necessary changes. Will file ticket for tht.

ALLOW_MANY_COMMITS=true


common/supplemental/likelySubtags.xml
```diff
@@ -10179,5 +8867,642 @@ not be patched by hand, as any changes made in that fashion may be lost.
...
 		<likelySubtag from="zzj" to="zzj_Hani_CN" origin="sil1"/>	<!-- Zuojiang Zhuang ➡︎ Zuojiang Zhuang (Han, China) -->
+
+		<!-- Replacements for locales maximizing to ..._ZZ -->
+
+		<likelySubtag from="aai" to="aai_Latn_PG" origin="sil1"/>	<!-- Arifama-Miniafia ➡︎ Arifama-Miniafia (Latin, Papua New Guinea) -->
+		<likelySubtag from="aak" to="aak_Latn_PG" origin="sil1"/>	<!-- Ankave ➡︎ Ankave (Latin, Papua New Guinea) -->
+		<likelySubtag from="aau" to="aau_Latn_PG" origin="sil1"/>	<!-- Abau ➡︎ Abau (Latin, Papua New Guinea) -->
+		<likelySubtag from="abi" to="abi_Latn_CI" origin="sil1"/>	<!-- Abidji ➡︎ Abidji (Latin, Côte d’Ivoire) -->
+		<likelySubtag from="abt" to="abt_Latn_PG" origin="sil1"/>	<!-- Ambulas ➡︎ Ambulas (Latin, Papua New Guinea) -->
+		<likelySubtag from="aby" to="aby_Latn_PG" origin="sil1"/>	<!-- Aneme Wake ➡︎ Aneme Wake (Latin, Papua New Guinea) -->
```
sffc, line 8878: Question: do these really all go to `Latn`? We shouldn't put `Latn` in the likely subtag unless they are written with the latin script. We could put `Zzzz` there if we don't know what the script is.

macchiati: These are all based on SIL data, and we have no reason to doubt it. (Latin is often the primary script in languages with small populations.)

macchiati:
Fixed test failures.

sffc (approved):

PR #3209, opened 2023-08-21
CLDR-17021 Fix one code, add test, regenerate likelySubtags.txt

macchiati:
CLDR-17021

In the course of fixing this, I resurrected a validity test for locales, and did a bit of cleanup.

common/supplemental/likelySubtags.xml
- Remove the remaining macroregion (missed, because it is a CLDR 2-letter code)

common/testData/localeIdentifiers/likelySubtags.txt
- Regenerate with correct data.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateLikelySubtagTests.java
- Disable this tool, since it can be chosen mistakenly.

tools/cldr-code/src/main/java/org/unicode/cldr/util/LanguageTagParser.java
- Rewire to use better validator

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocaleValidator.java
- Better validator

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
- Add enum value for extensions

tools/cldr-code/src/main/java/org/unicode/cldr/util/Validity.java
- Add useful constant

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestLocale.java
- Add test, covering selected items + likely subtags + language population + bcp47 extensions

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestValidity.java
- Accounting for new enum value for extensions

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

srl295 (approved):
I can think of a number of places where this could be useful

PR #3210, opened 2023-08-22
CLDR-16947 remove TestAnnotations.TestUniqueness

srl295:
- redundant with CheckDisplayCollisions
- per discussion in #3185

CLDR-16947

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
@@ -5,8 +5,6 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
-import com.google.common.collect.Multimap;
-import com.google.common.collect.TreeMultimap;
 import com.ibm.icu.dev.util.UnicodeMap;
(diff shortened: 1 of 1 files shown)


btangmu:
@srl295 it's not clear to me that CheckDisplayCollisions tests uniqueness of annotations. It has:

`else if (CHECK_FLAG_AND_EMOJI && myType == Type.ANNOTATIONS)`

but CHECK_FLAG_AND_EMOJI is false. What's the basis for thinking TestAnnotations.TestUniqueness is redundant?

srl295:
> @srl295 it's not clear to me that CheckDisplayCollisions tests uniqueness of annotations. It has:
> 
> `else if (CHECK_FLAG_AND_EMOJI && myType == Type.ANNOTATIONS)`
> 
> but CHECK_FLAG_AND_EMOJI is false. What's the basis for thinking TestAnnotations.TestUniqueness is redundant?

it gave the same collision errors as CheckDisplayCollisions. Mark's confirmation in #3185 .   

@macchiati  thoughts? I'm not familiar with the details of the test case

pedberg-icu (approved):

srl295:
@macchiati thoughts?

macchiati (changes requested):
I think it is ok to remove this *if* the consoleCheck for the build is also run over the derived annotations. I'm not sure it is, so that needs to be verified.

PR #3211, opened 2023-08-23
CLDR-17028 Add test for person name formatting and regenerate test files

macchiati:
CLDR-17028

common/testData/personNameTest/af.txt ...
- Changed the format of the 'Order' parameter to include explicit givenFirst or surnameFirst instead of 'none', since ICU now supports that, and 'none' had some problems. Filed ticket to restore that, but I don't want it to hold things up.
- Regenerated with current data
- Added the copyright notice to the generation tool
- This added some locales that were not Org=cldr.
- As a byproduct of bug fixes, the order of parameters changed. That should not affect any users.
- Added a comment for each of the sample names, with where it comes from, such as # nativeFull

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
- Changed the format of the 'Order' parameter to include explicit givenFirst or surnameFirst instead of 'none', since ICU now supports that, and 'none' had some problems. Filed ticket to restore that, but I don't want it to hold things up.
- Fixed the generation to avoid the issue with 'none'.
- Added the copyright notice to the generation tool

- Minor changes so that the FormatParameters.from(String) handled final semicolon

tools/cldr-code/src/main/java/org/unicode/cldr/util/personname/SimpleNameObject.java
- Minor changes so that we could choose the separator (needed because some name data includes ","). Rather than replace the separator it is parameterized, so that we don't have to touch the tests.

- Enabled the test so it is run as part of the normal suite.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDataTest.java
- New test file set up to (in the future) ensure that all test data is tested on the CLDR side. This will help prevent cases where we forget to regenerate the test data.
- It adds the files that are tested to a list, so that we can ensure that all test files are covered.

ALLOW_MANY_COMMITS=true


macchiati:
This is a preliminary review; There are 3 errors I need to investigate.

```
name ; given; ไอริณ
name ; surname; กล้าหาญ
name ; locale; th_AQ
...
expectedResult; ไก
...
parameters; surnameFirst; long; monogram; informal

yields:
    Error: (TestDataTest.java:1) : th (169) : expected "ไก", got "กไ"

The relevant th.xml file data is:

		<personName order="surnameFirst" length="long" usage="monogram" formality="informal">
			<namePattern>{surname-monogram-allCaps}{given-informal-monogram-allCaps}</namePattern>
		</personName>

so the expected result is not right in the test file. Need to track that down.

2 similar ones for Lao
    Error: (TestDataTest.java:1) : lo (168) : expected "ນເ", got "ເນ"
    Error: (TestDataTest.java:1) : lo (169) : expected "ນເ", got "ເນ"

```

macchiati:
Also tools/cldr-rdf/.settings/org.eclipse.jdt.core.prefs was mistakenly included. Need to remove that

macchiati:
Rich, because ICU now has the "force order" I am making use of givenFirst/surnameFirst. That would take a change to the ICU test code. If that is a problem, I can restore the old hack.

common/testData/personNameTest/af.txt
```diff
@@ -53,8 +53,8 @@
 # =====
 
 enum ; field ; title, given, given2, surname, surname2, generation, credentials
-enum ; modifiers ; informal, allCaps, initialCap, initial, monogram, prefix, core
-enum ; options ; none, sorting
+enum ; modifiers ; informal, allCaps, initialCap, initial, retain, monogram, prefix, core, vocative, genitive
+enum ; order ; givenFirst, surnameFirst, sorting
```
richgillam, line 57: If you have the ability, it might be better to call these `forceGiven` and `forceSur`, or something along those lines, for clarity and consistency.

common/testData/personNameTest/af.txt
```diff
@@ -64,33 +64,51 @@ name ; locale; af_AQ
...
+parameters; givenFirst; long; referring; formal
+parameters; givenFirst; long; referring; informal
+parameters; givenFirst; medium; addressing; formal
+parameters; givenFirst; medium; addressing; informal
+parameters; givenFirst; medium; referring; formal
+parameters; givenFirst; medium; referring; informal
+parameters; givenFirst; short; addressing; formal
+parameters; givenFirst; short; addressing; informal
+parameters; givenFirst; short; referring; formal
+parameters; givenFirst; short; referring; informal
```
richgillam, line 78: As you mention, this'll require a change on the ICU side, but I don't think it'll be a big deal.  My only real concern about that is that the exhaustive ICU tests will fail until I can get in there and fix it to deal with this (not to mention any other issues this uncovers), and I've already gotten no end of flak for that test failing over the past few weeks.  I'd like to find a way to coordinate our changes so that we don't have that period of failing tests.

Any thoughts on how best to accomplish that?  One option might be to leave the test data files on the ICU side alone until I have the corresponding code changes ready to go.  Then I can include the updated data files as part of my PR.  Is there a ay to make that happen, or is the whole process automated now?

common/testData/personNameTest/af.txt
```diff
@@ -64,33 +64,51 @@ name ; locale; af_AQ
...
+parameters; surnameFirst; short; referring; informal
 
 expectedResult; P
 
-parameters; none; long; monogram; formal
-parameters; none; long; monogram; informal
-parameters; none; medium; monogram; formal
-parameters; none; medium; monogram; informal
-parameters; none; short; monogram; formal
-parameters; none; short; monogram; informal
```
richgillam, line 93: Do we still want the "none" stuff around, as a way of testing automatic selection of the correct field order, or are you doing that some other way?

common/testData/personNameTest/mt.txt
```diff
@@ -0,0 +1,239 @@
+# Test data for unit conversions
+#  Copyright © 1991-2023 Unicode, Inc.
+#  For terms of use, see http://www.unicode.org/copyright.html
+#  SPDX-License-Identifier: Unicode-DFS-2016
+#  CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+#
+# CLDR person name formatting test data for: mt
```
richgillam, line 7: This looks like a whole new data file.  We didn't have person-name data for this locale before?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
```diff
@@ -168,24 +167,28 @@ public static void main(String[] args) {
...
+                        //                        Options options; // use this to get the 'none'
+                        // case
+                        //                        if (order == nameOrder) {
+                        //                            options = Options.none;
+                        //                        } else if (order == Order.sorting && nameOrder ==
+                        // myOrder) {
+                        //                            options = Options.sorting;
+                        //                        } else {
+                        //                            continue; // skip otherwise
+                        //                        }
```
richgillam, line 180: Looks like you have an automated code formatter that messed up the indentation here...

Beyond that, I guess I should repeat my question from earlier-- does this version of the test somehow test automatic determination of field order based on the formatter and name locales?  Because that was something we were (theoretically) getting out of the old "none" stuff.  Or am I wrong to think the older "none" test cases were actually doing that in a useful way?  If so, does that mean that this set of tests doesn't really exercise that functionality, and that my other unit tests are the only thing checking that, and is that how we want it?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
```diff
@@ -168,24 +167,28 @@ public static void main(String[] args) {
...
-                            options = Options.sorting;
-                        } else {
-                            continue; // skip otherwise
-                        }
-
-                        String formatted = formatter.format(entry.getValue(), parameters);
+                        String formatted =
+                                formatter
+                                        .format(entry.getValue(), parameters)
+                                        .replace("ᵛ", ""); // remove special CLDR ST hack
```
richgillam, line 185: Just for my own curiosity, what was the "CLDR ST hack"?  What did this character mean in that context?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDataTest.java
```diff
@@ -0,0 +1,236 @@
...
+            filepath.forEach(x -> checkPersonNameTests(x));
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        FILES_CHECKED.add(PERSON_NAMES_DIR);
+    }
+
+    private void checkPersonNameTests(Path filePath) {
+        String name = filePath.toFile().getName();
+        if (name.startsWith("_") || !name.endsWith(".txt")) {
```
richgillam, line 62: What are we using `_` at the beginning of the filename for?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDataTest.java
```diff
@@ -0,0 +1,236 @@
...
+                    }
+                }
+            } catch (Exception e) {
+                throw new IllegalArgumentException(localeId + " (" + count + ") " + line, e);
+            }
+        }
+    }
+
+    /** Make sure this is the last test */
+    public void testΩ() {
```
richgillam, line 200: A Greek letter in a function name!  Wow!  I keep forgetting you can do things like that in Java...

richgillam (commented):
I think this looks good, and I don't think the `givenFirst`/`surnameFirst` stuff will cause a lot of hardship on the ICU side, except for during the period between when your changes land and my changes land.

It looks like your stuff is going through a code formatter that messes up the indentation of commented-out code, which is annoying.  I wonder if there's a way to defeat that?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDataTest.java
macchiati: Those are for files like the readme.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
macchiati: In CLDR, we don't have a way to substitute the vocative form of sample names, so there is a hack using the ᵛ to indicate that it would (if the implementation supports it) need the vocative form.

common/testData/personNameTest/mt.txt
macchiati: The test generates data if there are two fields with explicit values in the file: 
//ldml/personNames/nameOrderLocales[@order=\"givenFirst\"]
//ldml/personNames/nameOrderLocales[@order=\"surnameFirst\"]

So if people added values in the SurveyTool we generate the test file for the locale in that release.

common/testData/personNameTest/af.txt
macchiati: I think we want to have a separate test for that, and for handling the name order locale, etc. Have to think more about that.

common/testData/personNameTest/af.txt
macchiati: Good question. This PR should fix the problems with the discontinuity between the locale PNF data and the PNF test file. We will not be able to check in data that affects the test without regenerating the testData. So that will prevent future integration problems with ICU.

common/testData/personNameTest/af.txt
macchiati: Good idea. That is a bit more than I want to do right now, but I'll write a ticket to handle in the future.

macchiati: https://unicode-org.atlassian.net/browse/CLDR-17032

common/testData/personNameTest/af.txt
macchiati: https://unicode-org.atlassian.net/browse/CLDR-17032 added to that ticke

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
macchiati: I think the 'none' values were doing that in an incomplete way. We do want to test it, but I was focusing here on getting the tests cleaned up.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDataTest.java
macchiati: And now many other programming languages: Swift, and even C++

macchiati:
Had to squash, because a commit message contained a comma (ugg)

srl295:
The details were squashed out, but this PR also has a workaround for https://github.com/actions/cache/issues/1225

common/testData/personNameTest/af.txt
richgillam: Thank you!

common/testData/personNameTest/af.txt
richgillam: I guess what I'm worried about here is if we still have any algorithmic differences between CLDR and ICU.  If we do, the next CLDR integration will cause a crisis on the ICU side when this test starts failing.  And come to think of it, we _know_ we're going to have an algorithmic difference because you're using new syntax in the test file that my unit test doesn't yet know about.

Maybe it's not a big crisis if we keep this test as part of the exhaustive tests, but Frank wants to move it to the regular test suite (and usually yells at me if I break the exhaustive tests anyway).  Maybe we just have to live with that, but I'm not looking forward to it.

The other choice would be for me to try to get ahead of things by fixing my code _before_ the new data files land in the ICU build.  Maybe we should figure out a way to do that?

common/testData/personNameTest/af.txt
richgillam: Thanks; I'm fine with this, although I definitely think this is something we're going to want to address.

common/testData/personNameTest/mt.txt
richgillam: I see.  Thanks for the explanation.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
richgillam: You said above you thought we should find a way to test automatic determination of the name order, but that this isn't the right way, and you've filed a ticket to add something for that.  That all sounds fine to me.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
richgillam: Ah!  Okay.  Thank you!

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDataTest.java
richgillam: I see.  Thanks for the explanation.

richgillam (approved):
Looking good.  Thanks for the answers to all my questions!

macchiati:
Thanks, Rich, Steven!

macchiati:
Merged, so this should unblock fixing the ICU test

PR #3215, opened 2023-08-23
CLDR-16393 assume cldr-archive is present for tests

srl295:
- UNLESS -DNO_CLDR_ARCHIVE=true is set

CLDR-16393

This will cause tests to FAIL unless:
- cldr-archive is setup OR
- `-DNO_CLDR_ARCHIVE=true` is set.

ALLOW_MANY_COMMITS=true

⚠️ if we merge this, users will start getting failures unless they have an up-to-date `cldr-archive` OR disable the check as described here.

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDtd.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDtd.java
@@ -21,6 +21,7 @@ public class GenerateDtd {
         final DTD2Markdown dtd2md = new DTD2Markdown();
         // System.setProperty("show_all", "true");
         for (DtdType type : DtdType.values()) {
+            if (type.getStatus() != DtdType.DtdStatus.active) continue;
             if (type == DtdType.ldmlICU) {
                 continue;
             }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDtd.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDtd.java
@@ -21,8 +21,9 @@ public class GenerateDtd {
         final DTD2Markdown dtd2md = new DTD2Markdown();
         // System.setProperty("show_all", "true");
         for (DtdType type : DtdType.values()) {
-            if (type.getStatus() != DtdType.DtdStatus.active) continue;
-            if (type == DtdType.ldmlICU) {
+            if (type.getStatus() != DtdType.DtdStatus.active) {
+                continue;
+            } else if (type == DtdType.ldmlICU) {
                 continue;
             }
             DtdData data = DtdData.getInstance(type);
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateItemCounts.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateItemCounts.java
@@ -197,6 +197,7 @@ public class GenerateItemCounts {
 
         SupplementalDataInfo supp = SUPPLEMENTAL_DATA_INFO;
         for (DtdType dtdType : DtdType.values()) {
+            if (dtdType.getStatus() != DtdType.DtdStatus.active) continue;
             if (dtdType == DtdType.ldmlICU) {
                 continue;
             }
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
@@ -7,13 +7,17 @@ import java.io.File;
 import org.junit.jupiter.api.Test;
 
 public class TestCLDRPaths {
+    public static String HAS_CLDR_ARCHIVE = "HAS_CLDR_ARCHIVE";
 
     @Test
     void TestCanUseArchiveDirectory() {
         if (!canUseArchiveDirectory()) {
-            // We print this warning as part of the unit tests.
+            // We print this warning once as part of the unit tests.
             System.err.println(
-                    "WARNING: skipping using cldr-archive. Ideally, -DNO_CLDR_ARCHIVE=false and see <https://cldr.unicode.org/development/creating-the-archive>");
+                    String.format(
+                            "WARNING: -D%s=false, so skipping tests which use cldr-archive\n" +
+                            "See <%s>",
+                            HAS_CLDR_ARCHIVE, CLDRURLS.CLDR_ARCHIVE));
         }
     }
 
@@ -21,8 +25,8 @@ public class TestCLDRPaths {
      * @return true if it's OK to read CLDRPaths.ARCHIVE_DIRECTORY, false to skip.
      */
     public static final boolean canUseArchiveDirectory() {
-        if (CLDRConfig.getInstance().getProperty("NO_CLDR_ARCHIVE", false)) {
-            return false; // skip, NO_CLDR_ARCHIVE is set.
+        if (!CLDRConfig.getInstance().getProperty("HAS_CLDR_ARCHIVE", true)) {
+            return false; // skip due to property
         }
 
         final File archiveDir = new File(CLDRPaths.ARCHIVE_DIRECTORY);
@@ -31,16 +35,18 @@ public class TestCLDRPaths {
                     String.format(
                             "Could not read archive directory %s. "
                                     + "Please: "
-                                    + "1) setup the archive, <https://cldr.unicode.org/development/creating-the-archive>, "
+                                    + "1) setup the archive, <%s>, "
                                     + "2) set the -DARCHIVE= property to the correct archive location, or "
-                                    + "3) inhibit reading of cldr-archive with -DNO_CLDR_ARCHIVE=true",
-                            archiveDir.getAbsolutePath()));
+                                    + "3) inhibit reading of cldr-archive with -D%s=false",
+                            archiveDir.getAbsolutePath(), CLDRURLS.CLDR_ARCHIVE, HAS_CLDR_ARCHIVE));
         }
         return true; // OK to use
     }
 
     @Test
     void TestReadPrevSDI() {
+        // This is also an example of a test that's skipped (by the next line) if
+        // cldr-archive isn't available.
         assumeTrue(canUseArchiveDirectory());
         SupplementalDataInfo SDI_LAST =
                 SupplementalDataInfo.getInstance(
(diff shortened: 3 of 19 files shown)


btangmu (approved):

macchiati (approved):

srl295:
old tests fail though, 

```
Error:  (TestBasic.java:1454)  Error: : ldml DTD - Trunk children of «zone» must be superset of v1.5.1, and must contain «usesMetazone»: expected != null

Error:  (TestBasic.java:1454)  Error: : ldml DTD - Trunk children of «zone» must be superset of v1.5.0.1, and must contain «usesMetazone»: expected != null

Error:  (TestBasic.java:1454)  Error: : ldmlICU DTD - Trunk children of «zone» must be superset of v1.5.1, and must contain «usesMetazone»: expected != null

Error:  (TestBasic.java:1454)  Error: : ldmlICU DTD - Trunk children of «zone» must be superset of v1.5.0.1, and must contain «usesMetazone»: expected != null
```

maybe a new logknown issue to skip the failing ones?

srl295:
> old tests fail though,
> 
> ```
> Error:  (TestBasic.java:1454)  Error: : ldml DTD - Trunk children of «zone» must be superset of v1.5.1, and must contain «usesMetazone»: expected != null
> 
> Error:  (TestBasic.java:1454)  Error: : ldml DTD - Trunk children of «zone» must be superset of v1.5.0.1, and must contain «usesMetazone»: expected != null
> 
> Error:  (TestBasic.java:1454)  Error: : ldmlICU DTD - Trunk children of «zone» must be superset of v1.5.1, and must contain «usesMetazone»: expected != null
> 
> Error:  (TestBasic.java:1454)  Error: : ldmlICU DTD - Trunk children of «zone» must be superset of v1.5.0.1, and must contain «usesMetazone»: expected != null
> ```
> 
> maybe a new logknown issue to skip the failing ones?

I'll probably do this, but later

macchiati:
Looks like roughly the right direction, but we are running short of time; I will check in later quickly, but basically I'm done through the 6th and you through the 15th.

We will need to regenerate charts and do other things that depend on platforms (which are far more timely than Subdivisiions) so I suggest again that you restore platforms with a dummy file so that we can proceed.

srl295:
> Looks like roughly the right direction, but we are running short of time; I will check in later quickly, but basically I'm done through the 6th and you through the 15th.
> 
> We will need to regenerate charts and do other things that depend on platforms (which are far more timely than Subdivisiions) so I suggest again that you restore platforms with a dummy file so that we can proceed.

There are other bugs in the unit tests besides platforms. It's not respecting the firstVersion parameter. 

This PR currently knows about but skips DtdType.platform in all loops, is that sufficient ?what process "needs" platforms?

macchiati:
No process uses it; most of the attributeValidity is no longer used.
I'm just leery of such a big change at a bad time.

srl295:
> No process uses it; most of the attributeValidity is no longer used.
> 
> I'm just leery of such a big change at a bad time.

Mark, I hear you; but there are other major breakages in the tests. FirstVersion is ignored in many cases. ElementInfo also ignores firstVersion. 

Can you comment on the unit comparator "jp-bu" I think it was ?

macchiati:
Enabling the tests to work across all versions is a worthy goal. But it is
*not* a show stopper for this release. You have very little time until you
are gone, so we want to make the most of it. The separate, immediate goal
is to get GenerateSubdivisions to work.

For that, I think the only thing I think we need is to restore the platform
enum value and dtd file.

macchiati:
> Can you comment on the unit comparator "jp-bu" I think it was ?

I don't have enough context to know what you mean by this. There is a unit
comparator, and there is a bu-jp unit that it will compare (below), but I
don't know what the problem is.

  <convertUnit source='bu-jp' baseUnit='square-meter'
factor='tsubo_to_m2*1' systems="jpsystem"/>

srl295:
image

@macchiati this is the next issue hit, in the code that GenerateSubdivisions (and others) will hit,  that the `bu-jp` units aren't present prior to v44.

macchiati:
Lot's of units come into new versions, so it must be something special about that one.

srl295:
But do you agree to put this on ice for the moment?

macchiati:
Yes, I agree. Since you are going to be out in car there are problems.

srl295:
just rebasing this PR in my queue.

srl295:
this was affected by CLDR-17078 rename ldmlKeyboard.dtd to ldmlKeyboard3.dtd #3274 because we are retaining the old DTDs. It doesn't attempt to have new code test the old DTDs.

srl295:
@macchiati when this merges, you will be able to create new tests that use cldr-archive with the following snippet:

```java
        if (!TestCLDRPaths.canUseArchiveDirectory()) {
            return; // skipped
        }

        // ... use cldr-archive ...
```

srl295:
> Can you comment on the unit comparator "jp-bu" I think it was ?
> I don't have enough context to know what you mean by this. There is a unit comparator, and there is a bu-jp unit that it will compare (below), but I don't know what the problem is. <convertUnit source='bu-jp' baseUnit='square-meter' factor='tsubo_to_m2*1' systems="jpsystem"/>

this issue didn't surface with current code/data.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDtd.java
```diff
@@ -21,6 +21,7 @@ public static void main(String[] args) throws IOException {
         final DTD2Markdown dtd2md = new DTD2Markdown();
         // System.setProperty("show_all", "true");
         for (DtdType type : DtdType.values()) {
+            if (type.getStatus() != DtdType.DtdStatus.active) continue;
```
macchiati, line 24: Missing {...}

macchiati: This happens multiple times below, so check for those also.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.jupiter.api.Test;
+
+public class TestCLDRPaths {
+
+    @Test
+    void TestCanUseArchiveDirectory() {
+        if (!canUseArchiveDirectory()) {
+            // We print this warning as part of the unit tests.
+            System.err.println(
+                    "WARNING: skipping using cldr-archive. Ideally, -DNO_CLDR_ARCHIVE=false and see <https://cldr.unicode.org/development/creating-the-archive>");
```
macchiati, line 16: Unclear message. Does it mean the following? 

Ideally you should set -DNO_CLDR_ARCHIVE=false **AND** create the archive as per <https://cldr.unicode.org/development/creating-the-archive>

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
```diff
@@ -0,0 +1,50 @@
...
+            return false; // skip, NO_CLDR_ARCHIVE is set.
+        }
+
+        final File archiveDir = new File(CLDRPaths.ARCHIVE_DIRECTORY);
+        if (!archiveDir.isDirectory()) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "Could not read archive directory %s. "
+                                    + "Please: "
+                                    + "1) setup the archive, <https://cldr.unicode.org/development/creating-the-archive>, "
```
macchiati, line 34: Ambiguous: does this mean #1 and (#2 or #3), or (#1 and #2) or #3?

macchiati (changes requested):
Nice work. Just a few tweaks needed.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
srl295: They are already opting out. The message is to suggest no longer opting out.

macchiati: double negatives are often confusing: HAS_CLDR_ARCHIVE=true would be better

srl295: Except the default is false. 

How about: SKIP_CLDR_ARCHIVE=false as the default. You set it to true if you don't want to setup CLDR-archive.

srl295:
false-by-default means that omitting any property (such as `HAS_CLDR_ARCHIVE`) means it has a false value, that's how other settings work. But i can make this one true by default.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
srl295: changes to HAS_CLDR_ARCHIVE=false per discussion

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCLDRPaths.java
srl295: i've changed this (see example run below).

I took out the option of changing `-DARCHIVE` by instead folding that into the informative. message (`-DARCHIVE=/a/b/c`) - so if /a/b/c is the wrong path the user can correct that property.  

Will need to update the docs page after this of course.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateItemCounts.java
```diff
@@ -197,6 +197,7 @@ private static void checkBadAttributes(Relation<String, String> path2value2, Str
 
         SupplementalDataInfo supp = SUPPLEMENTAL_DATA_INFO;
         for (DtdType dtdType : DtdType.values()) {
+            if (dtdType.getStatus() != DtdType.DtdStatus.active) continue;
```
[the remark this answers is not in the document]
srl295: Since were using spotless it would be better to set spotless to do this.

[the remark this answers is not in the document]
srl295: @macchiati there are 2,000+ cases of single-line if statements in the code. I couldn't find a spotless setting to fix this, though eclipse has one.

PR #3218, opened 2023-08-24
CLDR-16956 Updating the charts refreshed the coverage levels.

macchiati:
CLDR-16956

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/properties/coverageLevels.txt
+++ b/common/properties/coverageLevels.txt
@@ -20,7 +20,7 @@ bgc ;	basic ;	Haryanvi
 bho ;	basic ;	Bhojpuri
 blo ;	moderate ;	Anii
 bn ;	modern ;	Bangla
-br ;	moderate ;	Breton
+br ;	basic ;	Breton
 brx ;	basic ;	Bodo
 bs ;	modern ;	Bosnian
 bs_Cyrl ;	basic ;	Bosnian (Cyrillic)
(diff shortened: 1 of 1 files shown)


macchiati:
Review is mostly pro forma. The downgrade of Breton was a surprise, however, because I don't think we added anything at modern coverage.

macchiati:
I meant: added any paths at moderate coverage.

srl295 (approved):

srl295:
> Review is mostly pro forma. The downgrade of Breton was a surprise, however, because I don't think we added anything at modern coverage.

that is surprising, and curious why it changed now.

Oh— actually, thoughts:

- sr_Latn etc is due to algorithmic. should update BRS to regen this chart when algorithmic is turned.
- Sierra Leone currency?  
- Remember, that we've been having to add new languages (vmw, etc) to coverage80 once they "make basic".  perhaps that's causing some of these downgrades at this juncture.

srl295:
if it is the language stuff, then we should have some mechanism so that the coverage only changes "next time"

srl295:
@macchiati here's what's missing from breton at modern:

image

srl295:
SLE is going to bite a lot of locales because it was in _comprehensive_ until after close of surveytool.

macchiati:
We had talked about copying the values, but there wasn't a ticket to do so. I filed one at https://unicode-org.atlassian.net/browse/CLDR-17038

Because of shift in dates we should have time to get this in.

PR #3222, opened 2023-08-26
CLDR-16956 BRS v44 generate charts 2

macchiati:
CLDR-16956

- Adds attributes to missing-basic TSV chart

ALLOW_MANY_COMMITS=true


btangmu:
@macchiati there's a problem with the first commit message -- one solution is to squash (click Details to the right of "single-commit -- This PR has 2 commits"

macchiati:
A small change, but would like to get this in to unblock another PR

pedberg-icu (approved):

PR #3224, opened 2023-08-28
CLDR-16208 Fix inheritance for default number systems in Arabic locales

btangmu:
-If the special paths are not present, add them

-New set toRetainSpecial; unlike toRetain, add if not already present even if CONSTRAINED_RESTORATION

-New code/data encapsulates which locales/paths are special

-Unrelated refactoring: use LocaleNames.ROOT instead of hard-coded root

-Unrelated refactoring: remove unused directoryName

CLDR-16208

ALLOW_MANY_COMMITS=true


macchiati (approved):

macchiati (commented):
My only comment is that we should add a test of the generated Arabic files to verify that each has the numbering system.

PR #3225, opened 2023-08-28
CLDR-16980 CLDRModify on common/main; regen scriptToExemplars.txt, update TestUnicodeSetExamples

pedberg-icu:
CLDR-16980

Ran CLDRModify basic and -fP passes on common/main. The first rearranged availableFormats in en_001; the second rearranged the main exemplar set in kxv_Deva, including replacing the composed U+095C with U+0921 + U+093C (preferred form for NFC); note that nukta U+093C is not used productively in kxv_Deva, it is only used for this combination. That latter change cause a unit test failure LikelySubtagsTest.testGetResolvedScriptVsExemplars which required re-generating scriptToExemplars.txt; doing that also required updating the expected results in TestExampleGenerator.TestUnicodeSetExamples.

ALLOW_MANY_COMMITS=true


DraganBesevic (approved):
LGTM

macchiati (approved):

PR #3226, opened 2023-08-28
CLDR-16633 kbd: use NMTOKEN for ids

srl295:
CLDR-16633

thanks @eggrobin 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -696,7 +696,9 @@ This element defines a mapping between an abstract key and its output. This elem
 
 _Attribute:_ `id`
 
-> The `id` attribute uniquely identifies the key. NMTOKEN, restricted to `[a-zA-Z0-9_.-]`. It can (but needn't be) the Latin key name for a Latn script keyboard (a, b, c, A, B, C, …), or any other valid token (e-acute, alef, alif, alpha, …)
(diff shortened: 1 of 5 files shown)


mcdurdin (approved):

srl295:
@macchiati  @eggrobin  comments?  OK for this release?

macchiati (dismissed):
I think we need to discuss this. NMToken is very broad, and includes a huge amount of garbage. Would be much better to restrict to Unicode identifiers (UAX 31)

srl295:
@macchiati please discuss, see comments in the ticket.  I was also proposing that for *CLDR* repository data, we could verify that keyboards only use UAX31.  This adds a burden on implementations and ties it to a specific Unicode version.

The proposal, as noted in discussion on the ticket, is to expand from ASCII to NMTOKEN for this release, with a note in the spec to revisit next time.

miloush:
@macchiati just to note NMTOKEN was a compromise between making it more restricted but also more open. This is just a dictionary with a string key to be looked up. What is the concern of including arbitrary (or in this case NMTOKEN) characters in the key?

srl295:
I will add a test here that verifies that all keyboards actually conform to UAX 31

mcdurdin:
@eggrobin thank you for the detail on NMTOKEN -- really helpful background!

srl295:
@macchiati ping?

srl295:
keeping the old keyboard.dtd file plays havoc with git merges

PR #3230, opened 2023-08-29
CLDR-17042 adjust coverage

macchiati:
CLDR-17042

common/main/yi_UA.xml
- Change Yiddish to be UA (001 is only for constructed languages)

common/properties/coverageLevels.txt
- After these changes, some coverage increases

- Remove the new locales from those required for Basic.
- Note: used MinimizeRegex.java to reformat the regex as a flat list, for ease of editing. If you want to look at the before and after, you can use that with the previous value to see differences.

common/supplemental/likelySubtags.xml
- Yiddish was failing because only constructed languages are allowed to have 001 (currently)

- Add values for tok and vo
- Change default contents for Yiddish

common/supplemental/supplementalMetadata.xml
- Change Yiddish to be UA (001 is only for constructed languages)

- Added note

- The logic was wrong, causing no to fail. Need to avoid truncation parents

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
- Update Locales.txt to reflect locales with dedicated resources (thus setting to and qu to more realistic values)

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/LocaleMatcherTest.java
- Fix Yiddish

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverage.java
- Clean up TestLocales to actually fail when a locale is not Core, thus preventing future non-core locales

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
- Clean up test to exclude new basic
- Also reformatted for maintenance

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/XLocaleMatcherTest.java
- Fix Yiddish

ALLOW_MANY_COMMITS=true


macchiati:
@btangmu The extra code that the charts call to check supplemental items that the locale is missing is 

`CoreCoverageInfo.getCoreCoverageInfo(testFile, errors);
`
You can unpack the errors to find the path that failed. You can used this to have the ST match the charts. Let me know if you have questions

macchiati:
Draft for now; need to track down some errors

macchiati:
Now ready for review and approval

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
```diff
@@ -503,46 +503,36 @@ public void TestCoverageCompleteness() {
...
-                                + "dgr|mak|inh|lun|ts|fj|na|kpe|sr_ME|trv|bug|ban|xal|oc|alt|nia|myv|ain|krl|ay|"
-                                + "syr|kv|umb|cu|prg|vo|"
-                                + "atj|blt|clc|crg|crj|crk|crl|crm|crr|csw|cwd|hax|hdn|hnj|hur|ike|ikt|"
-                                + "kwk|lil|moe|ojb|ojc|ojg|ojs|ojw|oka|pqm|slh|str|tce|tgx|tht|trw|ttm)");
+                        "(" // start
+                                + "nds_NL|fa_AF|ro_MD|sr_ME|sw_CD"
+                                // Length 4
+                                + "|root"
+                                // Length 3
+                                + "|ace|ach|ada|ady|aeb|afh|agq|ain|akk|akz|ale|aln|alt|ang|ann|anp|apc|arc|arn|aro|arp|arq|ars|arw|ary|arz|asa|ase|atj|avk|awa"
```
srl295, line 515: what's… the point of this?

srl295 (approved):
LGTM. Need to revisit next time re: policies but LGTM for now

PR #3231, opened 2023-08-29
CLDR-16974 en_AU/GB/IN: for yMMMEd fmts, add yMMMEEEEd without comma; en_AU/GB no comma in MMMEd fmts

pedberg-icu:
CLDR-16974

ALLOW_MANY_COMMITS=true

- For en_AU, en_GB, en_IN: Remove comma after weekday if weekday is fully spelled out. (Note, many en_IN formats had a comma before year that was used pretty consistently throughout, did not remove that).
- Update standard full formats to match.
- This was done with reference to the fully resolved date formats in en_001 (parent of all of these), that is, the formats in en_001 with missing values there filled in from en. So some additions you see here replace the format that would have been inherited from en_001 (via ↑↑↑, and then possibly pattern expansion for E → EEEE etc.)
- I did not do this in en_001 itself because it has around 90 other descendants and we do not know whether this change is desirable for all of them, or even for a majority of them.

### DIFF
--- a/common/main/en_AU.xml
+++ b/common/main/en_AU.xml
@@ -1617,7 +1617,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<dateFormats>
 					<dateFormatLength type="full">
 						<dateFormat>
-							<pattern>↑↑↑</pattern>
+							<pattern>EEEE d MMMM y G</pattern>
 							<datetimeSkeleton>↑↑↑</datetimeSkeleton>
 						</dateFormat>
 					</dateFormatLength>
@@ -1691,6 +1691,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						<dateFormatItem id="GyMMM">↑↑↑</dateFormatItem>
 						<dateFormatItem id="GyMMMd">↑↑↑</dateFormatItem>
 						<dateFormatItem id="GyMMMEd">↑↑↑</dateFormatItem>
+						<dateFormatItem id="GyMMMEEEEd">EEEE d MMM y G</dateFormatItem>
 						<dateFormatItem id="h">↑↑↑</dateFormatItem>
 						<dateFormatItem id="H">↑↑↑</dateFormatItem>
 						<dateFormatItem id="hm">↑↑↑</dateFormatItem>
@@ -1703,7 +1704,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						<dateFormatItem id="MMM">↑↑↑</dateFormatItem>
 						<dateFormatItem id="MMMd">↑↑↑</dateFormatItem>
 						<dateFormatItem id="MMMEd">↑↑↑</dateFormatItem>
+						<dateFormatItem id="MMMEEEEd">EEEE d MMM</dateFormatItem>
 						<dateFormatItem id="MMMMd">↑↑↑</dateFormatItem>
+						<dateFormatItem id="MMMMEEEEd">EEEE d MMMM</dateFormatItem>
 						<dateFormatItem id="ms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="y">↑↑↑</dateFormatItem>
 						<dateFormatItem id="yMd">dd/MM/y</dateFormatItem>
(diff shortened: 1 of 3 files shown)


AEApple:
I thought commas would only be removed when the full weekday name is used? Am I misreading the date intervals?

common/main/en_AU.xml
```diff
@@ -1617,7 +1617,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<dateFormats>
 					<dateFormatLength type="full">
 						<dateFormat>
-							<pattern>↑↑↑</pattern>
+							<pattern>EEEE d MMMM y G</pattern>
```
stenshamn, line 1620: According to the information I have, and is included in the ticket, this format should retain a comma after the weekday. It is only when the weekday OR the year is omitted that the comma goes away.

pedberg-icu:
> I thought commas would only be removed when the full weekday name is used? Am I misreading the date intervals?

Hmm. I interpreted the example from your linguists as a further refinement from what Fredrik was proposing, i.e.:
- Apple linguists for en_GB, en_AU: Only use the comma in date formats including both weekday and year...
- Google linguists for en_GB, en_AU, en_IN: ... and only if the weekday is abbreviated.

(The Google example was only about a date with both weekday and year; the common behavior seems to depend heavily on whether those are both included).

common/main/en_AU.xml
pedberg-icu: > According to the information I have, and is included in the ticket, this format should retain a comma after the weekday. It is only when the weekday OR the year is omitted that the comma goes away.

@Fredrik Yes but per the Google linguists, the comma should also be omitted in that case if the weekday is fully spelled out. The examples from Apple only showed an abbreviated weekday.

AEApple:
> > I thought commas would only be removed when the full weekday name is used? Am I misreading the date intervals?
> 
> @AEApple Hmm. I interpreted the example from your linguists as a further refinement from what Fredrik was proposing, i.e.:
> 
> * Apple linguists for en_GB, en_AU: Only use the comma in date formats including both weekday and year...
> * Google linguists for en_GB, en_AU, en_IN: ... and only if the weekday is abbreviated.
> 
> (The Google example was only about a date with both weekday and year; the common behavior seems to depend heavily on whether those are both included).
> 
> So for en_GB and en_AU I removed the comma from abbreviated formats that do not have both weekday and year (thus restoring the old behavior). And for en_GB, en_AU, en_IN I removed the comma from formats with both weekday and year that use a full weekday.

Updated ticket to clarify that the feedback was on whether the full or abbreviated weekday name was used. There was no differentiation based on whether the year was included or not.

macchiati (approved):

common/main/en_AU.xml
stenshamn: The feedback from Apple's vetters were abbreviation-agnostic I guess you could say. The example values in the ticket were added by Annemarie I believe.

common/main/en_IN.xml
```diff
@@ -1618,9 +1633,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 							<greatestDifference id="y">↑↑↑</greatestDifference>
 						</intervalFormatItem>
 						<intervalFormatItem id="yMMMEd">
-							<greatestDifference id="d">E d – E d MMM y G</greatestDifference>
-							<greatestDifference id="M">E d MMM – E d MMM y G</greatestDifference>
-							<greatestDifference id="y">E d MMM y – E d MMM y G</greatestDifference>
+							<greatestDifference id="d">E d – E d MMM, y G</greatestDifference>
```
AEApple: I don't understand this change?

pedberg-icu: Yes, I have already reverted that, it was a mistake.

AEApple (approved):
LGTM

Two questions:
- Do we need to change anything in coverage for next cycle?
- Should we also update en_001 since en_CA inherits from en now and therefore won't be impacted if we update en_001 and all the other major English locales all seem aligned on this change?

pedberg-icu:
@AEApple 
> Two questions:
> * Do we need to change anything in coverage for next cycle?

Probably. I can file a ticket for that.

> * Should we also update en_001 since en_CA inherits from en now and therefore won't be impacted if we update en_001 and all the other major English locales all seem aligned on this change?

I wanted to be conservative about this since CLDR has around 90 other locales that inherit from en_001 (and Apple has a lot more than that), and although we do not have vetters for these, we get a lot of input from users of these other locales who feel strongly about formats.

PR #3236, opened 2023-08-29
CLDR-16857 SLE/SLL data migration a-j

srl295:
- copy SLL to SLE
- add date range to SLL

(second half coming)

CLDR-16857

ALLOW_MANY_COMMITS=true


stenshamn (approved):
Clever way to fix that, hat's off. :-)
Approving with the caveat that I only spot-checked the files and the GitHub web interface stops me from going further down the list than "jv.xml".

macchiati (approved):

srl295:
@stenshamn `jv` is the last one.  Will pick up others (and a couple missed) tomorrow

PR #3237, opened 2023-08-29
CLDR-16981 update the spec for -ms-

macchiati:
CLDR-16981

ALLOW_MANY_COMMITS=true


richgillam (approved):
Looks good.

PR #3244, opened 2023-08-30
CLDR-16358 default time cycle H→h for es_419 children in Latin Amer except BR,BZ

pedberg-icu:
CLDR-16358

ALLOW_MANY_COMMITS=true

Change default time cycle from h to H for most es_419 child locales (except BR).

1.  In supplemental `<timeData>`:
    - Move es_419 child regions that are actually in Latin America and have primary language Spanish from a section like `preferred="H" allowed="H h hB hb"` to the section that would be obtained by moving h to the beginning, e.g. in this case `preferred="h" allowed="h H hB hb"`.
    - Did not do this for BR, though we have es_BR under es_419; the primary language for BR is pt, and pt_BR uses a 24-hour cycle. Similarly, did not do this for es_BZ; the primary language for BZ is en, and en_BZ uses a 24-hour cycle.
    - Also added 419 as a region to supplemental `<timeData>`

2. Updated `<timeFormats>` in specific locales
    - In es_419, changed standard timeFormats to use h
    - Five es_419 child locales in Latin America were already overriding `<timeFormats>` to use h, so just removed those overrides: es_ locales for CO, DO, PA, PR, VE. Other es_419 child locales in Latin America were just inheriting `<timeFormats>` so they get updated automatically, except...
    - es_BR was just inheriting `<timeFormats>` from es_419, so now it need to explicitly add `<timeFormats>` using H. Same with es_BZ.

### DIFF
--- a/common/main/es_419.xml
+++ b/common/main/es_419.xml
@@ -1793,26 +1793,26 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<timeFormats>
 					<timeFormatLength type="full">
 						<timeFormat>
-							<pattern>HH:mm:ss zzzz</pattern>
-							<datetimeSkeleton>HHmmsszzzz</datetimeSkeleton>
+							<pattern>h:mm:ss a zzzz</pattern>
+							<datetimeSkeleton>ahmmsszzzz</datetimeSkeleton>
 						</timeFormat>
 					</timeFormatLength>
 					<timeFormatLength type="long">
 						<timeFormat>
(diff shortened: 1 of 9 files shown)


macchiati (approved):

pedberg-icu:
@macchiati Thanks for the review. I had to fix an issue for BZ (primary language en_BZ is 24 hour => es_BZ should be too) and similarly undo one timeData move that affected some other en locale. Can you re-stamp please?

macchiati (approved):
Passt bereits Pläne Lage

PR #3260, opened 2023-09-06
CLDR-17057 Set CUC as invalid after 2021-01-01

pedberg-icu:
CLDR-17057

ALLOW_MANY_COMMITS=true

Set CUC as invalid after 2021-01-01. 

@markusicu Thanks for the research on this!


markusicu (approved):
lgtm tnx

AEApple (approved):

PR #3259, opened 2023-09-05
CLDR-17062 BRS v44 CLDRModify passes before alpha2 integration

pedberg-icu:
CLDR-17062

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true
**Note: This PR must be merged using a Merge Commit (temporarily enable in github gear menu)**

Run CLDRModify passes before v44 alpha2 integration and after the CLDRModify changes for [CLDR-16972](https://unicode-org.atlassian.net/browse/CLDR-16972) were merged:

1. The no-options pass just sorted elements properly: availableFormats for en_GB, annotations entries for en_001, yo_BJ
2. The -fQ pass ensured that "tts" names were added to the keyword annotations, re-ordered those keyword annotations, and removed duplicates. In some cases there were apparent errors: the keyword annotations were promoted from unconfirmed to confirmed even though the tts names were also unconfirmed (bew, cv, kab, to); or keywords appears to be deleted (hy). The files with those issues were not included as part of the commit for -fQ changes, and I have filed [CLDR-17063](https://unicode-org.atlassian.net/jira/software/c/projects/CLDR/issues/CLDR-17063) for those issues.
3. The -fP pass fixed a doubled space for `bas`, and made some further annotations fixes (reordering keyword annotations and removing real duplicates) for en_001, hy, to. 

[CLDR-16972]: https://unicode-org.atlassian.net/browse/CLDR-16972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CLDR-17063]: https://unicode-org.atlassian.net/browse/CLDR-17063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3261, opened 2023-09-06
CLDR-17073 regen algorithmic locales before alpha2

pedberg-icu:
CLDR-17073

ALLOW_MANY_COMMITS=true

This is a replacement for [PR #3248](https://github.com/unicode-org/cldr/pull/3248) since I could not figure out how to undo the CLDRModify -fp (lowercase) in that one. But note: Unlike that one, this one was begun **after** [PR #3259](https://github.com/unicode-org/cldr/pull/3259) was merged (which ran the 3 standard CLDRModify passes), so there are some differences. And the CLDRModify step for this one (using -fP, uppercase P) produced no changes.
For this one:

1. CLDRileTransformer, no arguments, produced commit 1
2. CLDRModify -fP for the files modified in step 1, produced no changes
3. GenerateDerivedAnnotations, no arguments, produced commit 2


macchiati (approved):
I did a quick scan of the others, and they look good. 
As for Swiss German, the main change there is from ß to ss, and the repository is already clean for that.

PR #3262, opened 2023-09-09
:construction: CLDR-0000 show keyman keyboard

srl295:
Attempts to load a random keyboard matching the current locale.  See photo.

image

https://cldr-staging.unicode.org/cldr-apps/v#/csw/Languages_K_N/ experimental keyman web integration.  Go to any language.   Then, pop up the Search Box (top right corner).   Text input will show a web keyboard.

Lots of issues such as we might want to store/persist a preferred keyboard per language (as an example the Tamil keyboard chosen by default is arbitrary, it's not the best one for users).

- use language code

CLDR-0000

ALLOW_MANY_COMMITS=true


srl295:
A simple proposal for this would to add a button under the gear menu that enables the keyboard for the current session. It could be marked as a techpreview.

PR #3264, opened 2023-09-10
CLDR-17049 fix retain in fr PNF initials

macchiati:
CLDR-17049

ALLOW_MANY_COMMITS=true


pedberg-icu:
> Rich, Peter, Dragan, if this would cause problems for the integration, we can hold off on it.

I do not think it will be a problem

pedberg-icu (dismissed):

pedberg-icu (approved):

pedberg-icu:
I fixed the test data to match the pattern changes (tests were failing).

pedberg-icu:
I am going to go ahead and merge so we have for CLDR-ICU integration

macchiati:
Thanks!

PR #3267, opened 2023-09-11
CLDR-16358 Fix timeData for es_BO/EC/PE; qu_BO/EC/PE locales should use region pref cycle h

pedberg-icu:
CLDR-16358

ALLOW_MANY_COMMITS=true

In the original [PR 3244](https://github.com/unicode-org/cldr/pull/3244) for this I neglected two things:
- When I moved BO, EC, PE from preferred cycle H to h in timeData, I did not see that there were also separate entries for `es_BO`, `es_EC`, and `es_PE`. Those just need to be deleted.
- The `gu` locales (for BO, EC, PE) now need to have explicit 12-hour standard time formats, instead of inheriting the H format from root.

### DIFF
--- a/common/main/qu.xml
+++ b/common/main/qu.xml
@@ -1485,26 +1485,26 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<timeFormats>
 					<timeFormatLength type="full">
 						<timeFormat>
-							<pattern>↑↑↑</pattern>
-							<datetimeSkeleton>↑↑↑</datetimeSkeleton>
+							<pattern>h:mm:ss a zzzz</pattern>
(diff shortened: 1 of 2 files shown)


AEApple:
PR description says `gu` but file changed is `qu` is that expected?

DraganBesevic:
> PR description says `gu` but file changed is `qu` is that expected?

I think `qu` is correct one. Description needs to be updated

DraganBesevic (approved):

pedberg-icu:
> > PR description says `gu` but file changed is `qu` is that expected?
> 
> I think `qu` is correct one. Description needs to be updated

Yes, thanks, fixed description

PR #3268, opened 2023-09-12
CLDR-16405 Fix odd likelySubtags, and add test

macchiati:
CLDR-16405

It was hard to know exactly how to fix this issue without knowing the failure in ICU, but I took a shot.

The commented lines were in fact superfluous, and I removed them (or replaced them). For an example of replacement, see from="rom_Cyrl_BG" to="rom_Cyrl_BG".

I also wrote a test to find any other superfluous lines. It exposed those and some others in the sil1 data, plus a few in the non-sil1 data. I manually fixed 3 non-sil1 data cases also, and filed a ticket to fix the others (it takes a bit of thought for each one to see the right approach.)

So hopefully that works.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -510,8 +510,6 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<!--{ Hebrew; ?; ? } => { Hebrew; Hebrew; Israel }-->
 		<likelySubtag from="hi" to="hi_Deva_IN"/>
 		<!--{ Hindi; ?; ? } => { Hindi; Devanagari; India }-->
-		<likelySubtag from="hi_Latn" to="hi_Latn_IN"/>
-		<!--{ Hindi; Latin; ? } => { Hindi; Latin; India }-->
 		<likelySubtag from="hif" to="hif_Latn_FJ"/>
 		<!--{ Fiji Hindi; ?; ? } => { Fiji Hindi; Latin; Fiji }-->
 		<likelySubtag from="hil" to="hil_Latn_PH"/>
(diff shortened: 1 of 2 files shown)


macchiati:
Note: the likely subtags test data didn't need updating with GenerateTestData.java

pedberg-icu:
> It was hard to know exactly how to fix this issue without knowing the failure in ICU, but I took a shot.

The ticket does have the ICU errors...

common/supplemental/likelySubtags.xml
```diff
@@ -9122,9 +9109,9 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="jib" to="jib_Latn_NG" origin="sil1"/>	<!-- Jibu ➡︎ Jibu (Latin, Nigeria) -->
 		<likelySubtag from="jra" to="jra_Latn_VN" origin="sil1"/>	<!-- Jarai ➡︎ Jarai (Latin, Vietnam) -->
 		<likelySubtag from="jra_Khmr" to="jra_Khmr_KH" origin="sil1"/>	<!-- Jarai (Khmer) ➡︎ Jarai (Khmer, Cambodia) -->
-		<!-- CLDR-16405 the following two conflict with the metadata aliases treating aju as the primary child of jrb -->
-		<!-- <likelySubtag from="jrb" to="jrb_Hebr_IL" origin="sil1"/> -->	<!-- Judeo-Arabic ➡︎ Judeo-Arabic (Hebrew, Israel) -->
-		<!-- <likelySubtag from="jrb_Arab_MA" to="jrb_Arab_MA" origin="sil1"/>-->	<!-- Judeo-Arabic (Arabic, Morocco) ➡︎ Judeo-Arabic (Arabic, Morocco) -->
+		<likelySubtag from="jrb" to="jrb_Hebr_IL" origin="sil1"/> 	<!-- Judeo-Arabic ➡︎ Judeo-Arabic (Hebrew, Israel) -->
```
pedberg-icu, line 9112: This one I am unsure about. We have `aju` as the primary child of `jrb` and the default script for `aju` is I think supposed to be `Arab`. But we will see how this goes during the next ICU integration, at worst we might have to comment out this line again.

pedberg-icu (approved):

common/supplemental/likelySubtags.xml
macchiati: Hmmm. We've introduced direct inheritance from language A to language B (eg nb -> no), but I don't think we have a test that the maximized script doesn't change in doing so, which is pretty fragile. Should probably add such a test, which would force us to change, eg from aju -> jrb into aju -> jrb_Arab.

Note, according to Wikipedia, it appears that https://en.wikipedia.org/wiki/Judeo-Arabic_dialects were consistently written in Hebrew

PR #3270, opened 2023-09-12
CLDR-17081 Fix PNF test data for ICU

macchiati:
CLDR-17081

This is not yet complete; it fixes some of the "inclusion of draft data" plus the stray superscript

ALLOW_MANY_COMMITS=true


macchiati:
Still draft for now, so just fyi

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
```diff
@@ -69,7 +71,9 @@ public static void main(String[] args) {
             }
 
             try {
-                CLDRFile cldrFile = factory.make(locale, true);
+                CLDRFile cldrFile =
+                        factory.make(locale, true, DraftStatus.contributed); // don't include
+                // draft=unconfirmed/provisional
```
richgillam, line 76: I don't know what I'm looking at, but is `DraftStatus.contributed` a property of the whole file?  What I was seeing in some files was that _some_ person-name resources had their draft status set to "contributed" (or, actually, not mentioned at all), and _others_ had their draft status set to "provisional" or "unconfirmed".  Does this catch that?

richgillam (commented):
Changes look good so far...

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
macchiati: It sets the minimum bar for items to be included. So contributed means either contributed or approved (which is implicit if there is no draft value). That's the general bar we use for ICU, so this should get better results. cc @pedberg-icu

common/main/gl.xml
```diff
@@ -9797,27 +9797,26 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
 		</personName>
 		<personName order="sorting" length="long" usage="referring" formality="informal">
 			<namePattern>{surname-core} {surname2}, {given-informal} {surname-prefix}</namePattern>
-			<namePattern alt="1">↑↑↑</namePattern>
+			<namePattern alt="1">{surname-core}, {given-informal} {surname-prefix}</namePattern>
 		</personName>
 		<personName order="sorting" length="medium" usage="referring" formality="formal">
 			<namePattern>{surname-core} {surname2}, {given} {given2-initial} {surname-prefix}</namePattern>
-			<namePattern alt="1">{surname-core} {surname2}, {given} {given2} {surname-prefix}</namePattern>
+			<namePattern alt="1">{surname-core}, {given} {given2} {surname-prefix}</namePattern>
```
richgillam, line 9807: Did you want this to be `{given2}` when the first entry has `{given2-initial}`?

richgillam (dismissed):
All the changes look great.  Thank you!

- Is it worth looking through all the files to make sure we don't have any other `<namePattern alt="1">↑↑↑</namePattern>` stuff in there?
- Are there other changes coming, or is this all ready to go?

richgillam (approved):
Looks good!  But see question about `{given2-initial}` in my comments above.

common/main/gl.xml
macchiati: Missed this, will examine in 1/2 hour

macchiati: Got it (lined up all the others for comparison, too, so I wouldn't miss any others). Good catch.

richgillam (dismissed):
Cut and print!

macchiati:
Dang, forgot to rerun the GenerateTestData! Can you stamp again?

richgillam (approved):
Stamped again! 😀

PR #3272, opened 2023-09-18
CLDR-17063 CLDRModify -fQ debugging, real/fake keyword paths

btangmu:
-This illustrates cause of draft=unconfirmed bug

-The old, fake keyword path is derived from tts path by removeAttribute

-The new, real keyword path gotten from the CLDRFile

-In general, they are not the same, common difference is draft unconfirmed

CLDR-17063

ALLOW_MANY_COMMITS=true


macchiati:
good catch. I think we need a policy decision if the draft status of the two items are different, when adding the tts to the non-tts. Example:

tts=provisional   // name
non-tts=contributed.   // search keywords

Here is a strawman: 

- If there are no non-tts items, use the draft status of the tts for the non-tts when adding.
- if there are non-tts items, leave the draft status of the non-tts alone when adding.

btangmu:
I'll await further input about a policy decision. In the meantime I'm implementing a fix to keep the original draft status for each path...

PR #3274, opened 2023-09-18
CLDR-17078 rename ldmlKeyboard.dtd to ldmlKeyboard3.dtd

srl295:
- also rename root element to `<keyboard3>`, etc.
- also, restore prior ldmlKeyboard.dtd and ldmlPlatform.dtd but marked deprecated
- test/tool fixes as needed
- CLDR APIs won't attempt to read prior keyboard files
- remove prior-version Keyboard code

CLDR-17078

ALLOW_MANY_COMMITS=true


srl295:
kbd-check is known to fail here and unrelated.  It's actually showing that the change doesn't cause a problem for kmc

miloush:
I would think we do not need to introduce new xsd for depracted dtds

keyboards/dtd/ldmlKeyboard.dtd
```diff
@@ -4,263 +4,161 @@ For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-DFS-2016
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
 -->
-<!-- Important Note:
+<!-- This is the prior keyboard format, and is included for backwards compatibility. -->
```
miloush, line 7:
```suggestion
<!-- This is a descriptive keyboard format frozen in v43 and included for backwards compatibility. 
     All new applications should use the prescriptive format available in ldmlKeyboard2.dtd -->

File: keyboards/dtd/ldmlPlatform.dtd [hunk 2]
```diff
@@ -0,0 +1,22 @@
+<!--
+Copyright © 1991-2023 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+-->
+<!-- This is part of the prior keyboard format, and is included for backwards compatibility. -->
```
miloush, line 7:
```suggestion
<!-- This is a descriptive keyboard format frozen in v43 and included for backwards compatibility. 
     All new applications should use the prescriptive format available in ldmlKeyboard2.dtd -->
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdType.java
```diff
@@ -29,16 +29,22 @@ public enum DtdType {
             "transforms",
             "validity"),
     ldmlBCP47("common/dtd/ldmlBCP47.dtd", "1.7.2", null, "bcp47"),
-    // keyboard 3.0
-    keyboard("keyboards/dtd/ldmlKeyboard.dtd", "44.0", null, "../keyboards/3.0"),
-    keyboardTest("keyboards/dtd/ldmlKeyboardTest.dtd", "44.0", null, "../keyboards/test");
+    keyboard2("keyboards/dtd/ldmlKeyboard2.dtd", "44.0", null, "../keyboards/3.0"),
```
miloush, line 32: Yeah, shouldn't it be called ldmlKeyboard3.dtd? :-)

miloush:
I am not terribly opposed to have the root also be `keyboard2` or any other name distinct from `keyboard`. Outside of CLDR data it would significantly simplify detecting which kind of file you are dealing with.

macchiati:
I think it would also be preferable. Eg, enums tracking the root element
names would need exceptional handling. Simpler and easier to understand if
unique.

srl295:
> I would think we do not need to introduce new xsd for depracted dtds

I think I neglected to delete those xsds, I don't think they are actually generated anymore

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdType.java
srl295: 3? right…

srl295:
> I think it would also be preferable. Eg, enums tracking the root element names would need exceptional handling. Simpler and easier to understand if unique.

all of the exceptional handling is implemented in this PR.

srl295:
> > I would think we do not need to introduce new xsd for depracted dtds
> 
> I think I neglected to delete those xsds, I don't think they are actually generated anymore

fixed

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdType.java
miloush: Can be either way, but probably would make sense if it was unified, either ldmlKeboard2 and keyboards/2.0 or ldmlKeyboard3 and keyboards/3.0

srl295:
PTAL.  `<keyboard3>`

srl295:
tests pass locally.  Again, please ignore `kbd-check`

miloush (dismissed):

srl295:
My reading of the Adhoc notes is that the adhoc could approve of this.

srl295:
@macchiati how does this look?

mcdurdin (approved):

srl295:
PTAL.  Keeping the old keyboard.dtd file plays havoc with git merges, they end up being applied to the wrong place.

srl295:
please re-review, had to rebase due to another change

mcdurdin (approved):
The rebased changes look fine (reverting id restrictions from an alnum regex to 'any' in the dtd)

srl295:
The rebase was reviewed, will merge and we can iterate if need be…

PR #3275, opened 2023-09-20
CLDR-16372 Code to replace lateral inheritance markers in trunk

btangmu:
-New CLDRModify -fZ, Zero lateral, replace INHERITANCE_MARKER with bailey for all lateral inheritance

-For -fZ new boolean CONSTRUCTED_PSEUDO_PATH_NOT_LATERAL in VoteResolver#reviseInheritanceAsNeeded, no change in behavior as it stands

-Revise CLDRModify -fV, values that would inherit laterally, do not crash if parentValue is null

-Comments

CLDR-16372

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
@@ -79,6 +79,7 @@ import org.unicode.cldr.util.SupplementalDataInfo;
 // import org.unicode.cldr.util.Log;
 import org.unicode.cldr.util.SupplementalDataInfo.PluralInfo;
 import org.unicode.cldr.util.SupplementalDataInfo.PluralInfo.Count;
+import org.unicode.cldr.util.VoteResolver;
 import org.unicode.cldr.util.XMLSource;
 import org.unicode.cldr.util.XPathParts;
 import org.unicode.cldr.util.XPathParts.Comments;
@@ -2876,11 +2877,11 @@ public class CLDRModify {
 
                     @Override
                     public void handleStart() {
-                        // skip if the locale id's parent isn't root. That is, it must be at level-1
-                        // locale.
+                        // skip if the locale is root.
                         skip = getLocaleID().equals(XMLSource.ROOT_ID);
                         if (!skip) {
                             parentId = LocaleIDParser.getParent(getLocaleID());
+                            // This locale is "L1" (level one) if its parent is root.
                             isL1 = parentId.equals(XMLSource.ROOT_ID);
                             parentFile = null; // lazy evaluate
                         }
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2931,14 +2932,19 @@ public void handlePath(String xpath) {
                                     parentFile = factory.make(parentId, true);
                                 }
                                 String parentValue = parentFile.getStringValueWithBailey(xpath);
-                                if (!parentValue.equals(baileyValue)) {
```
btangmu, line 2934: there was NPE for parentValue null here

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -3052,6 +3058,26 @@ public void handlePath(String xpath) {
...
+                        }
+                        String newValue =
+                                VoteResolver.reviseInheritanceAsNeeded(xpath, value, getResolved());
+                        if (value.equals(newValue)) {
+                            return;
+                        }
+                        String fullXPath = cldrFileToFilter.getFullXPath(xpath);
+                        replace(fullXPath, fullXPath, newValue);
+                    }
+                });
```
btangmu, line 3080: This new "-fZ" is results in more "hardening" than "-fV", mainly, I think, because "-fZ" doesn't distinguish between level one locales and other locales -- and also, because "-fZ" treats constructed values as lateral inheritance, while "-fV doesn't

This PR re-enables -fV, and newly enables -fZ. One or the other may be useful for revising the production data, comparing the two is instructive.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -2261,8 +2261,18 @@ public static String reviseInheritanceAsNeeded(String path, String value, CLDRFi
...
+            // values as inherited, given that pathWhereFound doesn't equal path.
+            // This original behavior corresponds to CONSTRUCTED_PSEUDO_PATH_NOT_LATERAL = false.
+            // Reference: https://unicode-org.atlassian.net/browse/CLDR-16372
+            final boolean CONSTRUCTED_PSEUDO_PATH_NOT_LATERAL = false;
             value =
-                    pathWhereFound.value.equals(path)
+                    (pathWhereFound.value.equals(path)
+                                    || (CONSTRUCTED_PSEUDO_PATH_NOT_LATERAL
+                                            && GlossonymConstructor.PSEUDO_PATH.equals(
+                                                    pathWhereFound.value)))
```
btangmu, line 2275: This is a change without a difference (in behavior). It paves the way for making -fZ (and vote resolution, etc., in general) a bit more similar to -fV by changing false to true

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2876,11 +2877,11 @@ public void handleEnd() {
 
                     @Override
                     public void handleStart() {
-                        // skip if the locale id's parent isn't root. That is, it must be at level-1
-                        // locale.
+                        // skip if the locale is root.
```
btangmu, line 2880: The comment wasn't accurate -- this just changes the comment to match the code

btangmu:
@macchiati @srl295 this ticket is still "blocks-release" for v44, and the PR is a necessary step but doesn't finish the ticket. The finishing step will be to run either -fV or -fZ to "harden" about a quarter of a million values. So it will be great to have this PR reviewed for the code, and I'll follow up with one or more PRs for the data.

srl295 (approved):

PR #3276, opened 2023-09-20
CLDR-16604 kbd: complete scancode data

srl295:
- Also adds 'ks' as per CLDR-16873

CLDR-16604

ALLOW_MANY_COMMITS=true

### DIFF
--- a/keyboards/import/scanCodes-implied.xml
+++ b/keyboards/import/scanCodes-implied.xml
@@ -9,20 +9,45 @@
         Values are space separated hex bytes.
         Frame keys are not included.
     -->
+
     <form id="us">
-        <scanCodes codes="29 02 03 04 05 06 07 08 09 0A 0B 0C 0D" /> <!-- 13 keys, ` to = -->
+        <scanCodes codes="29 02 03 04 05 06 16607 08 09 0A 0B 0C 0D" /> <!-- 13 keys, ` to =-->
         <scanCodes codes="10 11 12 13 14 15 16 17 18 19 1A 1B 2B" /> <!-- 13 keys, Q to \ -->
         <scanCodes codes="1E 1F 20 21 22 23 24 25 26 27 28" />  <!-- 11 keys, A to ' -->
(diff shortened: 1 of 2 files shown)


keyboards/import/scanCodes-implied.xml
```diff
@@ -9,20 +9,45 @@
         Values are space separated hex bytes.
         Frame keys are not included.
     -->
+
     <form id="us">
-        <scanCodes codes="29 02 03 04 05 06 07 08 09 0A 0B 0C 0D" /> <!-- 13 keys, ` to = -->
+        <scanCodes codes="29 02 03 04 05 06 16607 08 09 0A 0B 0C 0D" /> <!-- 13 keys, ` to =-->
```
miloush, line 14: are you suuure

miloush:
```suggestion
        <scanCodes codes="29 02 03 04 05 06 07 08 09 0A 0B 0C 0D" /> <!-- 13 keys, ` to = -->
```

miloush (approved):

miloush:
Note that if you map keys from the left, `iso`=`abnt2`=`jis` and `us`=`ks`, with the only difference being `56` next to left shift. So currently we have two physical arrangements, either with or without that key.

miloush:
@srl295 what I mean is that they behave identical when the keys are not used, so technically there doesn't need to be that many forms defined. For example, you could have a form with 14 scancodes defined in the first row (13 on second, 12 on third and fourth), and `<row>` for the first row will have either 13 keys or 14 keys, it doesn't matter. (And to push it further, we could define a blank key and then have everyone use one prescribed form that includes 56 and not worry about forms, and/or think what a keyboard that has the key should do.)

It is important to have overridable/custom forms though, so nothing needs to change, it is just an observation. There is no harm in specifying them all individually explicitly.

srl295:
Jan, that makes sense. It could be worth noting, since keys are optional on
the right edges of any row.

PR #3277, opened 2023-09-20
CLDR-16918 kbd: respond to macchiati feedback

srl295:
CLDR-16918

Contains multiple items responding to @macchiati feedback in the above ticket.
Review note: can be reviewed commit-by-commit.

ALLOW_MANY_COMMITS=true

***NOTE*** Key modifiers will be covered in a separate ticket.

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -536,7 +536,7 @@ _Attribute:_ `indicator`
 
 Element used to store any names given to the layout.
 
-These names are not currently localized.
+These names are not localized but are informative names for the keyboard. Localization of these names would be done as separate data items elsewhere in CLDR.
 
 **Syntax**
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1389,8 +1389,8 @@ _Attribute:_ `modifier` (required for `hardware`)
 > To indicate that no modifiers apply, the reserved name of `none` is used.
 > The following modifier components can be used, separated by spaces.
 > Note that `L` or `R` indicates a left- or right- side modifier only (such as `altL`)
-> whereas `alt` indicates _either_ left or right alt key.
-> `shift` also indicates either shift key.
+> whereas `alt` indicates _either_ left or right alt key (that is, `altL` or `altR`). `ctrl` indicates either left or right ctrl key (that is, `ctrlL` or `ctrlR`).
+> `shift` also indicates either shift key. The left and right shift keys are not distinguishable in this specification.
 >
 > - `none` (no modifier, may not be combined with others)
 > - `alt`
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -520,11 +520,15 @@ _Attribute:_ `normalization`
 _Attribute:_ `layout`
 
 > The `layout` attribute describes the layout pattern, such as QWERTY, DVORAK, INSCRIPT, etc. typically used to distinguish various layouts for the same language.
+>
+> This attribute is not localized, but is an informative identifier for implementation use.
 
 _Attribute:_ `indicator`
 
-> The `indicator` attribute describes a short string to be used in currently selected layout indicator, such as US, SI9 etc.
+> The `indicator` attribute describes a short string to be used in currently selected layout indicator, such as `US`, `SI9` etc.
 > Typically, this is shown on a UI element that allows switching keyboard layouts and/or input languages.
+>
+> This attribute is not localized.
 
 * * *
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -355,7 +355,9 @@ For purposes of this current draft specification, the value should always be `te
 
 _Attribute:_ `locale` (required)
 
-This attribute represents the primary locale of the keyboard using BCP 47 [Unicode locale identifiers](tr35.md#Canonical_Unicode_Locale_Identifiers) - for example `"el"` for Greek. Sometimes, the locale may not specify the base language. For example, a Devanagari keyboard for many languages could be specified by BCP-47 code: `"mul-Deva"`. For further details, see [Keyboard IDs](#Keyboard_IDs).
+This attribute represents the primary locale of the keyboard using BCP 47 [Unicode locale identifiers](tr35.md#Canonical_Unicode_Locale_Identifiers) - for example `"el"` for Greek. Sometimes, the locale may not specify the base language. For example, a Devanagari keyboard for many languages could be specified by BCP-47 code: `"und-Deva"`. However, it is better to list out the languages explicitly using the [`locales`](#element-locales) element.
+
+For further details about the choice of locale ID, see [Keyboard IDs](#Keyboard_IDs).
 
 **Example** (for illustrative purposes only, not indicative of the real data)
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -988,6 +988,8 @@ _Attribute:_ `base`
 _Attribute:_ `path` (required)
 
 > If `base` is `cldr`, then the `path` must start with a CLDR version (such as `techpreview`) representing the CLDR version to pull imports from. The imports are located in the `keyboard/import` subdirectory of the CLDR source repository.
+> Implementations are not required to have all CLDR versions available to them, and an implementation may reject a keyboard file which references an unavailable CLDR version.
+>
 > If `base` is omitted, then `path` is an absolute or relative file path.
 
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -474,7 +474,7 @@ _Attribute:_ `cldrVersion` (fixed by DTD)
 ```xml
 <keyboard3 locale="tok">
     …
-    <version numb  er="1"/>
+    <version number="1"/>
     …
 </keyboard3>
 ```
@@ -988,7 +988,7 @@ _Attribute:_ `base`
 _Attribute:_ `path` (required)
 
 > If `base` is `cldr`, then the `path` must start with a CLDR version (such as `techpreview`) representing the CLDR version to pull imports from. The imports are located in the `keyboard/import` subdirectory of the CLDR source repository.
-> Implementations are not required to have any or all all CLDR versions available to them.
+> Implementations are not required to have all CLDR versions available to them.
 >
 > If `base` is omitted, then `path` is an absolute or relative file path.
 
@@ -1324,15 +1324,15 @@ _Attribute:_ `form` (required)
 > This attribute specifies the physical layout of a hardware keyboard,
 > or that the form is a `touch` layout.
 >
-> When using an on-screen keyboard, if there is not a `<layers form="touch">`
-> element, the hardware elements can be used for on-screen use. 
+> When using an on-screen touch keyboard, if the keyboard does not specify a `<layers form="touch">`
+> element, a `<layers form="{hardware}">` element can be used as an fallback alternative.
 > If there is no `hardware` form, the implementation may need
 > to choose a different keyboard file, or use some other fallback behavior when using a
 > hardware keyboard.
 >
 > Because a hardware keyboard facilitates non-trivial amounts of text input,
-> it is recommended to always have at least one hardware (non-touch) form, unless a layout
-> will only be used in touch form.
+> and many touch devices can also be connected to a hardware keyboard, it
+> is recommended to always have at least one hardware (non-touch) form.
 >
 > Multiple `<layers form="touch">` elements are allowed with distinct `minDeviceWidth` values.
 > At most one hardware (non-`touch`) `<layers>` element is allowed. If a different key arrangement is desired between, for example, `us` and `iso` formats, these should be separated into two different keyboards.
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -956,9 +956,9 @@ where a flick to the Northeast then South produces two code points.
 ### <a name="Element_import" href="#Element_import">Element: import</a>
 
 The `import` element is used to reference another xml file so that elements are imported from
-another file. The use case is to be able to import a standard set of vkeys, transforms, and similar
-from the CLDR repository.  `<import>` is not recommended as a way for keyboard authors to
-split up their keyboard into multiple files, as the intent is for each single XML file to contain all that is needed for a keyboard layout.
+another file. The use case is to be able to import a standard set of `vkey`s, `transform`s, and similar
+from the CLDR repository, especially to be able to share common information relevant to a particular script.
+The intent is for each single XML file to contain all that is needed for a keyboard layout, other than required standard import data from the CLDR repository.
 
 `<import>` can be used as a child of a number of elements (see the _Parents_ section immediately below). Multiple `<import>` elements may be used, however, `<import>` elements must come before any other sibling elements.
 If two identical elements are defined, the later element will take precedence, that is, override.
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -777,7 +777,7 @@ _Attribute:_ `to`
 
 _Attribute:_ `transform="no"` (optional)
 
-> The `transform` attribute is used to define a key that does not participate in a transform (until the next keystroke). This attribute value must be `no` if the attribute is present.
+> The `transform` attribute is used to define a key that does not participate in a transform (until the next keystroke). This attribute value must be `"no"` if the attribute is present.
 > This attribute is useful where it is desired to output where two different keys could output the same characters (with different key or modifier combinations) but only one of them is intended to participate in a transform.
 > When the next keystroke is pressed, the prior output may then combine using other transforms.
 >
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1406,7 +1406,7 @@ _Attribute:_ `modifier` (required for `hardware`)
 >
 > The set of modifiers must match `(none|([A-Za-z0-9]+)( [A-Za-z0-9]+)*)`
 >
-> To share a layer between to modifier sets, the layer data must be duplicated.
+> To share a layer between two modifier sets, the layer data must be duplicated.
 
 * * *
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -487,10 +487,10 @@ Element containing informative properties about the layout, for displaying in us
 **Syntax**
 
 ```xml
-<info [author="{author}"]
-      [normalization="{form}"]
-      [layout="{hint of the layout}"]
-      [indicator="{short identifier}"] />
+<info author="{author}"
+      normalization="{form}"
+      layout="{hint of the layout}"
+      indicator="{short identifier}" />
 ```
 
 > <small>
@@ -598,7 +598,7 @@ An element used to keep track of layout specific settings. This element may or m
 **Syntax**
 
 ```xml
-<settings [fallback="omit"] [transformFailure="omit"] [transformPartial="hide"] />
+<settings fallback="omit" />
 ```
 
 > <small>
@@ -616,7 +616,6 @@ _Attribute:_ `fallback="omit"`
 > The presence of this attribute means that when a modifier key combination goes unmatched, no output is produced. The default behavior (when this attribute is not present) is to fall back to the base map when the modifier key combination goes unmatched.
 
 If this attribute is present, it must have a value of omit.
-
 
 **Example**
 
@@ -678,17 +677,17 @@ This element defines a mapping between an abstract key and its output. This elem
 ```xml
 <key
  id="{key id}"
- [flicks="{flicks identifier}"]
- [gap="true"]
- [longPress="{long press keys}"]
- [longPressDefault="{default longpress target}"]
- [multiTap="{the output on subsequent taps}"]
- [stretch="true"]
- [switch="{layer id}"]
- [to="{the output}"]
- [transform="no"]
- [width="{key width}"]
- /><!-- {Comment to improve readability (if needed)} -->
+ flicks="{flicks identifier}"
+ gap="true"
+ longPress="{long press keys}"
+ longPressDefault="{default longpress target}"
+ multiTap="{the output on subsequent taps}"
+ stretch="true"
+ switch="{layer id}"
+ to="{the output}"
+ transform="no"
+ width="{key width}"
+ />
 ```
 
 > <small>
@@ -2153,11 +2152,11 @@ The relative ordering of `<reorder>` elements is not significant.
 <transformGroup>
     <!-- one or more <import/> elements are allowed at this point -->
     <reorder from="{combination of characters}"
-    [before="{look-behind required match}"]
-    [order="{list of weights}"]
-    [tertiary="{list of weights}"]
-    [tertiaryBase="{list of true/false}"]
-    [preBase="{list of true/false}"] />
+    before="{look-behind required match}"
+    order="{list of weights}"
+    tertiary="{list of weights}"
+    tertiaryBase="{list of true/false}"
+    preBase="{list of true/false}" />
     <!-- other <reorder/> elements... -->
 </transformGroup>
 ```
@@ -2402,7 +2401,7 @@ In text editing mode, different keyboard layouts may behave differently in the s
 ```xml
 <transforms type="backspace">
     <transformGroup>
-        <transform from="{combination of characters}" [to="{output}"] />
+        <transform from="{combination of characters}" to="{output}" />
     </transformGroup>
 </transforms>
 ```
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -191,14 +191,14 @@ Note that in parts of this document, the format `@x` is used to indicate the _at
 
 > 👉 Note: CLDR-TC has agreed that the changes required were too extensive to maintain compatibility. For this reason, the `ldmlKeyboard3.dtd` DTD used here is _not_ compatible with DTDs from prior versions of CLDR such as v43 and prior.
 >
-> To process earlier XML files, use the prior DTD and specification, such as v43.1 found at <https://www.unicode.org/reports/tr35/tr35-69/tr35.html>
+> To process earlier XML files, use the data and specification from v43.1, found at <https://www.unicode.org/reports/tr35/tr35-69/tr35.html>
 
 ### <a name="Accessibility" href="#Accessibility">Accessibility</a>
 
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-keyboards.md
```diff
@@ -591,12 +598,12 @@ _Attribute:_ `value` (required)
 
 ### <a name="Element_settings" href="#Element_settings">Element: settings</a>
 
-An element used to keep track of layout specific settings. This element may or may not show up on a layout. These settings reflect the normal practice by the implementation. However, an implementation using the data may customize the behavior.
+An element used to keep track of layout-specific settings. This element may or may not show up on a layout. These settings reflect the normal practice by the implementation. However, an implementation using the data may customize the behavior.
```
macchiati, line 601: By implementations.

docs/ldml/tr35-keyboards.md
```diff
@@ -770,7 +776,7 @@ _Attribute:_ `to`
 
 _Attribute:_ `transform="no"` (optional)
 
-> The `transform` attribute is used to define a key that does not participate in a transform (until the next keystroke). This attribute value must be `no` if the attribute is present.
+> The `transform` attribute is used to define a key that does not participate in a transform (until the next keystroke). This attribute value must be `"no"` if the attribute is present.
```
macchiati, line 779: The **value** is the part in quotes, and does not include the quotes. X='no' is the same as X="no"

docs/ldml/tr35-keyboards.md
```diff
@@ -981,6 +987,8 @@ _Attribute:_ `base`
 _Attribute:_ `path` (required)
 
 > If `base` is `cldr`, then the `path` must start with a CLDR version (such as `techpreview`) representing the CLDR version to pull imports from. The imports are located in the `keyboard/import` subdirectory of the CLDR source repository.
+> Implementations are not required to have all CLDR versions available to them, and an implementation may reject a keyboard file which references an unavailable CLDR version.
```
macchiati, line 990: Would not support a keyboard that

docs/ldml/tr35-keyboards.md
```diff
@@ -1314,11 +1322,18 @@ _Attribute:_ `form` (required)
 > This attribute specifies the physical layout of a hardware keyboard,
 > or that the form is a `touch` layout.
 >
-> It is recommended to always have at least one hardware (non-touch) form.
+> When using an on-screen keyboard, if there is not a `<layers form="touch">`
+> element, the hardware elements can be used for on-screen use. However, a `touch` keyboard
+> cannot be used with a hardware keyboard.
```
macchiati, line 1327: There are fancy hardware keyboards that have software controlled keycaps

docs/ldml/tr35-keyboards.md
```diff
@@ -1388,15 +1400,18 @@ _Attribute:_ `modifier` (required for `hardware`)
...
 >
-> Note that `alt` is sometimes referred to as _opt_ or _option_.
+> Note that `alt` in this specification is referred to on some platforms as "opt" or "option".
 >
 > Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
 >
-> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
-
+>
+> Matching is affected by the set of available layers. For example, if the `shift` and `altL` keys are held down, `<layer id="A" modifiers="altL"/>` will match if there are no other layers which include `altL`. However, if there is also a `<layer id="B" modifiers="altL shift"/>`, then layer `B` will be selected because it is a more exact match.
```
macchiati, line 1408: Specify the exact algorithm that determines closest match.

macchiati (dismissed):

docs/ldml/tr35-keyboards.md
mcdurdin: > There are fancy hardware keyboards that have software controlled keycaps

But generally they are still not touch keyboards with arbitrary layout -- they have a fixed hardware layout, and they generate known scan codes by position, no different to German vs French labelling on a traditional ISO hardware keyboard, where the electronics are identical, and it's just the text printed on the key caps that are changing.

miloush: "cannot be used" is not clear what it means, does it mean "implementations must reject" or "not guaranteed to work" or "undefined behavior"?

docs/ldml/tr35-keyboards.md
miloush: I don't recall us discussing this and I am not sure it is a good idea. I would prefer if it was no match, no output. That's easy to understand and easy to implement. Otherwise it will get messy, what if there is `shift` only layer as well as `altL` only layer? Would you match `caps+shift` to either of them?  (If needed, we can introduce an attribute refering to a base layer to simplify reusing layers for different modifiers.)

docs/ldml/tr35-keyboards.md
miloush: @macchiati is that to replace "may reject"? then that is a stronger statement, yet more of a claim rather than instruction. Currently it reads as "implementation can inform the user about unavailability of required data and let him decide to go ahead with version that is available". If your intention is taht implementation shall not allow this, than it needs to be stated more explicity.

I would also change "all CLDR versions" to "all or any CDLR versions".

srl295: I think it's still may, because it should be up to the implementation.

docs/ldml/tr35-keyboards.md
srl295: I would expect this to be "implementations must reject"

docs/ldml/tr35-keyboards.md
srl295: perhaps explicit match is a better policy then. We already said in committee that we weren't going to reuse layers at present.

docs/ldml/tr35-keyboards.md
srl295: If the hardware keyboard is really a big touch panel, then it's more of a touch keyboard.

docs/ldml/tr35-keyboards.md
srl295: I'm not sure what you are requesting then, could you use the suggest feature?

docs/ldml/tr35-keyboards.md
mcdurdin: Great catch. I'm with you @miloush. Disappointed that I missed this in my review!

We need to be careful with toggle keys (e.g. caps/num/scroll on US keyboards). Num Lock and Scroll Lock are currently excluded as available modifiers but conceptually the spec supports this, and if we ended up with another toggle key in future it could break keyboards if we are not careful.

miloush: Also note that the example @srl295 gave was easily achieved in the descriptive LDML using `altL altL+shift` (where space worked like _or_ and + like _and_). I am a bit concered that this is not allowed in the prescriptive one, where space means _and_. And since there is no fallback/base mechanism, people need to repeat the whole layer for each combination of modifiers. It's not a tragedy, just checking that it is intentional change.

docs/ldml/tr35-keyboards.md
```diff
@@ -591,12 +598,12 @@ _Attribute:_ `value` (required)
...
 ### <a name="Element_settings" href="#Element_settings">Element: settings</a>
 
-An element used to keep track of layout specific settings. This element may or may not show up on a layout. These settings reflect the normal practice by the implementation. However, an implementation using the data may customize the behavior.
+An element used to keep track of layout-specific settings by implementations. This element may or may not show up on a layout. These settings reflect the normal practice by the implementation. However, an implementation using the data may customize the behavior.
 
 **Syntax**
 
 ```xml
-<settings [fallback="omit"] [transformFailure="omit"] [transformPartial="hide"] />
+<settings fallback="omit" />
```
jahorton, line 606: Without context here... what is `fallback="omit"` supposed to indicate?  It's here as an example, but the meaning of the example is unclear.

docs/ldml/tr35-keyboards.md
srl295: It was an attempt to simplify. Could be too simple, though.

docs/ldml/tr35-keyboards.md
srl295: Just below there: 

> Attribute: fallback="omit"
>
> The presence of this attribute means that when a modifier key combination goes unmatched, no output is produced. The default behavior (when this attribute is not present) is to fall back to the base map when the modifier key combination goes unmatched.

jahorton: Ah, I must have searched the wrong side of the source for it.

docs/ldml/tr35-keyboards.md
macchiati: "cannot be used" and "reject" are both ill-defined (here and elsewhere). Is it non-conformant to use a touch keyboard layout with a hardware keyboard? Or is it just a ill-advised, because it cannot reasonably be interpreted in a way that would be useful with a hardware keyboard? 

And what would it mean for an implementation to "reject" a keyboard. I know what it would mean to have an application for a mortgage to be "rejected". But a software data structure? I could understand "must not use", but that also seems overloaded. If a touch keyboard layout simply couldn't apply to hardware keyboard, that's like saying you must not use a skateboard on a ski-slope.

miloush: > what would it mean for an implementation to "reject" a keyboard.

My interpretation of that is user tries to load a xml file into OS and OS says no, cannot use this file.

docs/ldml/tr35-keyboards.md
```diff
@@ -1388,15 +1402,17 @@ _Attribute:_ `modifier` (required for `hardware`)
...
 > - `shift`
 >
-> Note that `alt` is sometimes referred to as _opt_ or _option_.
+> Note that `alt` in this specification is referred to on some platforms as "opt" or "option".
 >
 > Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
 >
-> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
-
+> Matching is done exactly. For example, if the `shift` and `altL` keys are held down, `<layer id="A" modifiers="altL"/>` will not match, but `<layer id="B" modifiers="altL shift"/>`will match.
```
macchiati, line 1409: The spec has <layer id="layerId" modifier="{Set of Modifier Combinations}">
However, since it is a set, `modifiers` (with an S) would be the right attribute name.

There needs to be a precise, algorithmic definition of "match". That is, given an input set of modifiers IS, and a list of lines of the form:
<layer id="A" modifiers="{ModifierSet1}"/>
<layer id="A" modifiers="{ModifierSet2}"/>
...
It should be completely determinant which one of those lines matches. Some of the spec sounds like

Also, there need to be a precise list of exactly what the modifier key identifiers are. It appears to be the following, but it is not at all clear.

- none (no modifier, may not be combined with others)
- alt
- altL
- altR
- caps
- ctrl
- ctrlL
- ctrlR
- shift

If on the other hand, the IDs are purely up to the developer, then that opens the door to (massive) incompatibilities.

Phrases like "If there is a layer with a modifier alt, there may not be another layer with altL or altR. Similarly, if there is a layer with a modifier ctrl, there may not be a layer with ctrlL or ctrlR." are unclear. Does that mean such a keyboard is ill-formed? Must it be "rejected" — whatever that means?

```
Matching is done exactly. For example, if the shift and altL keys are held down, <layer id="A" modifiers="altL"/> will not match, but <layer id="B" modifiers="altL shift"/>will match.
``` 

This makes it sound like there must be an exact match, so if I have modifiers="shift", that cannot match an input set {shift, caps}, and so on. Is that really the case? If so, I can foresee a lot of duplicate layers in order to express current behavior of keyboards. If not, then the text of the specification is wrong.

Is the desire really to treat unmentioned modifiers as "a don't care", but pick the line with the most matches?

Anyway, this needs a precise specification.

docs/ldml/tr35-keyboards.md
```diff
@@ -981,6 +987,8 @@ _Attribute:_ `base`
 _Attribute:_ `path` (required)
 
 > If `base` is `cldr`, then the `path` must start with a CLDR version (such as `techpreview`) representing the CLDR version to pull imports from. The imports are located in the `keyboard/import` subdirectory of the CLDR source repository.
+> Implementations are not required to have any or all all CLDR versions available to them. An implementation may reject a keyboard file which references an unavailable CLDR version.
```
macchiati, line 990: "reject" is imprecise phrasing. Must not support? Must not allow the user to select? 
What does "unavailable" mean? Unavailable to the implementation? Non-existent?

docs/ldml/tr35-keyboards.md
```diff
@@ -187,17 +189,16 @@ Note that in parts of this document, the format `@x` is used to indicate the _at
 
 ### <a name="Compatibility_Notice" href="#Compatibility_Notice">Compatibility Notice</a>
 
-> 👉 Note: CLDR-TC has agreed that the changes required were too extensive to maintain compatibility. For this reason, the DTD used here is _not_ compatible with DTDs from prior versions of CLDR such as v41 and prior.
->
-> To process earlier XML files, use the prior DTD and specification, such as v41 found at <https://www.unicode.org/reports/tr35/tr35-66/tr35.html>
+> 👉 Note: CLDR-TC has agreed that the changes required were too extensive to maintain compatibility. For this reason, the `ldmlKeyboard3.dtd` DTD used here is _not_ compatible with DTDs from prior versions of CLDR such as v43 and prior.
 >
+> To process earlier XML files, use the prior DTD and specification, such as v43.1 found at <https://www.unicode.org/reports/tr35/tr35-69/tr35.html>
```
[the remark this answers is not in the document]
OkponkoXchange: **_[image]()_**

docs/ldml/tr35-keyboards.md
macchiati: If you mean that, you should say it. But is that a hard requirement on an OS, that it "must say 'no, you cannot use "this file"'. And does that mean an application (non-OS) that supports keyboards can use it?

miloush: That is what I was asking in the first place...

My preference would be we leave that to whoever is trying to load the file. Perhaps they decide they can deal with it, perahps they decide it's safer to not attempt.

docs/ldml/tr35-keyboards.md
miloush: > Implementations are not required to have any or all CLDR versions available to them.

We could just leave it at that.

srl295: > "reject" is imprecise phrasing. Must not support? Must not allow the user to select? 
> 
> What does "unavailable" mean? Unavailable to the implementation? Non-existent?
> 
> 
> 
> 
> 
> 
> 
> 

Jan's idea is to just drop the sentence.

srl295:
```suggestion
> Implementations are not required to have any or all all CLDR versions available to them.
```

docs/ldml/tr35-keyboards.md
srl295: @macchiati and all, I'd like to take the modifier logic into a separate PR. So I'd propose to merge this section as-is and have a followon PR (keeping this ticket open) to sort out the modifier(s).

docs/ldml/tr35-keyboards.md
srl295: i've already dropped the second clause.

docs/ldml/tr35-keyboards.md
srl295: I'd like to take this into a separate PR, merging the text as-is, leaving this ticket open to amend the modifier logic.

docs/ldml/tr35-keyboards.md
```diff
@@ -1314,11 +1322,18 @@ _Attribute:_ `form` (required)
 > This attribute specifies the physical layout of a hardware keyboard,
 > or that the form is a `touch` layout.
 >
-> It is recommended to always have at least one hardware (non-touch) form.
+> When using an on-screen keyboard, if there is not a `<layers form="touch">`
+> element, the hardware elements can be used for on-screen use. However, a `touch` keyboard
+> may not be used with a hardware keyboard.
+>
```
srl295, line 1328:
```suggestion
> When using an on-screen keyboard, if there is not a `<layers form="touch">`
> element, the hardware elements can be used for on-screen use. 
```

there's already text below that says that the implementation may need to choose a different file. Doesn't make a statement about conformance.

docs/ldml/tr35-keyboards.md
mcdurdin: I think it's a little stronger than that. We deliberately designed the touch layouts to be completely independent of hardware layouts -- from scan codes on up. Attempting to jam a touch layout into a hardware layout through some algorithmic process may kinda work, for some layouts. But I want to avoid the expectation that this will even be possible -- it just doesn't really make sense and we shouldn't be leaving that interpretation open on the data model. And if we do leave it open, we may end up with implementers pushing back on more novel touch layout designs because it breaks their touch-to-hardware algorithms.

That said, I'd be happy for this to be expressed as a implementation recommendation or guideline if that's preferable to others.

srl295:
Will update.  But *other than modifier keys* are there any outstanding issues with this PR?

srl295:
Now up to date.

srl295:
Rebased. please review, knowing that the matching mechanism will be evaluated separately.

docs/ldml/tr35-keyboards.md
srl295: please see what you think of the current recommendation

docs/ldml/tr35-keyboards.md
```diff
@@ -187,17 +189,16 @@ Note that in parts of this document, the format `@x` is used to indicate the _at
...
 
 Keyboard use can be challenging for individuals with various types of disabilities. For this revision, the committee is not evaluating features or architectural designs for the purpose of improving accessibility. Such consideration could be fruitful for future revisions. However, some points on this topic should be made:
 
 1. Having an industry-wide standard format for keyboards will enable accessibility software to make use of keyboard data with a reduced dependence on platform-specific knowledge.
-2. Some features, such as multiTap and flicks, have the potential to reduce accessibility and thus should be discouraged. For example, multiTap requires pressing keys at a certain speed, and flicks require a more complex movement (press-and-flick) beyond a simple tap. Alternatively, inclusion of accessible methods of generating the same outputs (for example, simple keys on an additional layer), should be considered.
+2. Features which require certain levels of mobility or speed of entry should be considered for their impact on accessibility. This impact could be mitigated by means of additional, accessible methods of generating the same output.
```
mcdurdin, line 201: I am wondering if even this statement is too much. Implementations should be able to synthesize alternate ways of accessing these features with accessibility controls, so IMO it may not be necessary or even desirable to have multiple ways of accessing keys specified in a keyboard layout.

docs/ldml/tr35-keyboards.md
```diff
@@ -471,7 +474,7 @@ _Attribute:_ `cldrVersion` (fixed by DTD)
 ```xml
 <keyboard3 locale="tok">
     …
-    <version number="1"/>
+    <version numb  er="1"/>
```
mcdurdin, line 477:
```suggestion
    <version number="1"/>
```

typo?

docs/ldml/tr35-keyboards.md
```diff
@@ -485,10 +488,10 @@ Element containing informative properties about the layout, for displaying in us
...
 
 ```xml
-<info [author="{author}"]
-      [normalization="{form}"]
-      [layout="{hint of the layout}"]
-      [indicator="{short identifier}"] />
+<info author="{author}"
+      normalization="{form}"
+      layout="{hint of the layout}"
+      indicator="{short identifier}" />
```
mcdurdin, line 494: Why did you drop the `[` `]` which indicated optionality? Having that visible was really helpful I thought.

docs/ldml/tr35-keyboards.md
```diff
@@ -950,9 +956,9 @@ where a flick to the Northeast then South produces two code points.
 ### <a name="Element_import" href="#Element_import">Element: import</a>
 
 The `import` element is used to reference another xml file so that elements are imported from
-another file. The use case is to be able to import a standard set of vkeys, transforms, and similar
-from the CLDR repository.  `<import>` is not recommended as a way for keyboard authors to
-split up their keyboard into multiple files, as the intent is for each single XML file to contain all that is needed for a keyboard layout.
+another file. The use case is to be able to import a standard set of `vkey`s, `transform`s, and similar
```
mcdurdin, line 959: Noting that `vkey` has been deleted in another recent PR.

docs/ldml/tr35-keyboards.md
```diff
@@ -982,6 +988,8 @@ _Attribute:_ `base`
 _Attribute:_ `path` (required)
 
 > If `base` is `cldr`, then the `path` must start with a CLDR version (such as `techpreview`) representing the CLDR version to pull imports from. The imports are located in the `keyboard/import` subdirectory of the CLDR source repository.
+> Implementations are not required to have any or all all CLDR versions available to them.
```
mcdurdin, line 991:
```suggestion
> Implementations are not required to have any or all CLDR versions available to them.
```

But given the statement above "other than required standard import data from the CLDR repository", this seems disingenuous?

docs/ldml/tr35-keyboards.md
```diff
@@ -1316,11 +1324,16 @@ _Attribute:_ `form` (required)
 > This attribute specifies the physical layout of a hardware keyboard,
 > or that the form is a `touch` layout.
 >
-> It is recommended to always have at least one hardware (non-touch) form.
+> When using an on-screen keyboard, if there is not a `<layers form="touch">`
+> element, the hardware elements can be used for on-screen use.
```
mcdurdin, line 1328: This is a little ambiguous. On-screen keyboard may refer to a visual representation of a hardware keyboard or a touch layout. Typically, the implementation would use the current hardware layout if it is intended to be a visual representation for user reference.

Something like:
```suggestion
> When using an on-screen touch keyboard, if the keyboard does not specify a `<layers form="touch">`
> element, a `<layers form="{hardware}">` element can be used as an fallback alternative. 
```

docs/ldml/tr35-keyboards.md
```diff
@@ -1316,11 +1324,16 @@ _Attribute:_ `form` (required)
...
-> It is recommended to always have at least one hardware (non-touch) form.
+> When using an on-screen keyboard, if there is not a `<layers form="touch">`
+> element, the hardware elements can be used for on-screen use. 
 > If there is no `hardware` form, the implementation may need
 > to choose a different keyboard file, or use some other fallback behavior when using a
 > hardware keyboard.
 >
+> Because a hardware keyboard facilitates non-trivial amounts of text input,
+> it is recommended to always have at least one hardware (non-touch) form, unless a layout
+> will only be used in touch form.
```
mcdurdin, line 1335: The "unless" here doesn't seem necessary. Might be helpful to note that Android and iOS allow pairing of a hardware keyboard.

```suggestion
> Because a hardware keyboard facilitates non-trivial amounts of text input,
> and many touch devices can also be paired with a hardware keyboard, it 
> is recommended to always have at least one hardware (non-touch) form.
```

docs/ldml/tr35-keyboards.md
```diff
@@ -1390,13 +1402,15 @@ _Attribute:_ `modifier` (required for `hardware`)
...
-> Note that `alt` is sometimes referred to as _opt_ or _option_.
+> Note that `alt` in this specification is referred to on some platforms as "opt" or "option".
 >
 > Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
 >
 > For hardware layouts, the use of `@modifier` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
 >
 > The set of modifiers must match `(none|([A-Za-z0-9]+)( [A-Za-z0-9]+)*)`
+>
+> To share a layer between to modifier sets, the layer data must be duplicated.
```
mcdurdin, line 1413:
```suggestion
> To share a layer between two modifier sets, the layer data must be duplicated.
```

mcdurdin (commented):
Here's some feedback; I've not really modified the sections on modifiers.

docs/ldml/tr35-keyboards.md
srl295: Makes sense. and this is certainly a common use case.

docs/ldml/tr35-keyboards.md
srl295: sounds good (going with 'connected' as a more general instead of 'paired' even though that is technically the common case)

docs/ldml/tr35-keyboards.md
srl295: The implied import data is required, so perhaps this should just be 'all'.

docs/ldml/tr35-keyboards.md
srl295: yes, already fixed and merged in that PR. i'll be merging that into this one shortly…

docs/ldml/tr35-keyboards.md
srl295: it was inconsistent with other parts of the spec, and also led to the XML being not quite XML.  

Perhaps something like: `author="{author (optional)}"`

docs/ldml/tr35-keyboards.md
srl295: This text no longer says "additional methods _such as adding keys to the base layer_." So an additional, accessible method could be synthesized, or it could be in the layout itself.  The point was just to note items tha tmight be of concern.

With that, does it still need revision?

docs/ldml/tr35-keyboards.md
mcdurdin: I guess. The square brackets were really clear I thought but consistency probably trumps that :grin:

docs/ldml/tr35-keyboards.md
mcdurdin: I think the mitigation advice should be clearly directed at implementers, not keyboard authors. That'd be enough.

docs/ldml/tr35-keyboards.md
miloush: I also prefer the square brackets and we possibly introduced them as I was fixing the keyboard LDML earlier. They should make the other parts of the spec match this one! :-)

srl295:
OK - ready to merge this one?

miloush (approved):

mcdurdin (approved):

docs/ldml/tr35-keyboards.md
```diff
@@ -515,19 +518,23 @@ _Attribute:_ `normalization`
...
+> This attribute is not localized.
 
 * * *
 
 ### <a name="Element_names" href="#Element_names">Element: names</a>
 
 Element used to store any names given to the layout.
 
-These names are not currently localized.
+These names are not localized but are informative names for the keyboard. Localization of these names would be done as separate data items elsewhere in CLDR.
```
macchiati, line 537: should be "unique identifiers for the keyboard". That is, they should be *keys* for localized names.

srl295: These keyboards are independently produced by their own authors and don't necessarily end up in the CLDR repo. How would you propose to allocate and enforce "unique identifiers"?  

I think this could be a good discussion, but it seems like a new item - could you file a new ticket on unique identifiers? 

The locale identifiers are not required to be unique.

PR #3282, opened 2023-09-25
CLDR-16604 kbd:  jis scancodes incorrect

srl295:
CLDR-16604

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true

found in review of https://github.com/keymanapp/keyman/pull/9615

### DIFF
--- a/keyboards/import/scanCodes-implied.xml
+++ b/keyboards/import/scanCodes-implied.xml
@@ -35,12 +35,12 @@
         <scanCodes codes="29 02 03 04 05 06 07 08 09 0A 0B 0C 0D 7D" /> <!-- 14 keys -->
         <scanCodes codes="10 11 12 13 14 15 16 17 18 19 1A 1B" />    <!-- 12 keys -->
         <scanCodes codes="1E 1F 20 21 22 23 24 25 26 27 28 2B" />    <!-- 12 keys -->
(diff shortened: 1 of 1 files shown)


mcdurdin (dismissed):

miloush:
The standard btw. is available here: https://www.shoshin.co.jp/c/cons/oadg/oadghwd.pdf. Physical layouts starting on p. 274.

For 109A and all the others:
image

None of them has the physical key that would be number 45. I confirm the Japanese hardware I have also does not have it. It is however, reserved to the expected scancode (p. 286):

image

miloush (dismissed):

miloush:
However, the Korean I belive is KS X 5003:2001 (confirmed 2021). That is available at https://e-ks.kr/streamdocs/view/sd;streamdocsId=72059288176667875. Key numbers p. 9, scan code mapping p. 15. As far as I can tell, the key next to backspace should not be 7D but 2B (with the second row having only 12 keys):

image

image

Sadly I don't have any Korean hardware myself at the moment, although I can see you can buy both one-row- and two-row-enter keyboards, and I suspect the former would have 7D and the latter 2B next to the backspace. Poor people.

If we want to have an explicit Korean form I would tend to follow the standard. The other option is already covered by the `jis` form.

srl295:
@mcdurdin backslash again ^

@miloush PTAL

miloush (approved):

mcdurdin (approved):

PR #3285, opened 2023-09-27
CLDR-16943 kbd: Normalization

srl295:
- remove <info normalization>, add <settings normalization>
- dtd and spec update

Co-authored-by: Marc Durdin <marc@durdin.net>

CLDR-16943

ALLOW_MANY_COMMITS=true


docs/ldml/tr35-keyboards.md
```diff
@@ -596,7 +585,7 @@ An element used to keep track of layout specific settings. This element may or m
 **Syntax**
 
 ```xml
-<settings [fallback="omit"] [transformFailure="omit"] [transformPartial="hide"] />
+<settings fallback="omit" normalization="mixed" />
```
miloush, line 588: The brackets were denoting an optional attribute. You kept them above in the `info` element, it should be unified.

docs/ldml/tr35-keyboards.md
```diff
@@ -632,6 +621,26 @@ Indicates that:
 2.  If a transform is escaped, output the contents of the buffer.
 3.  During a transform, hide the contents of the buffer as the user is typing.
 
+_Attribute:_ `normalization`
+
+> Normalization will not typically be the responsibility of the keyboard author, rather this will be managed by the implementation.
+> Output from the keyboard, following application of all transform rules, will be normalized to implementation or application-requested form.
```
miloush, line 627: This makes it sound like normalization only happens on output. We should be a bit more explicit about what normalization allowed means for when things need to be matched (transforms, display).

mcdurdin: Yes; the additional normalization details should be an implementation guideline -- that implementers should normalize 'context' inputs from the app and all strings within keyboard to NFD internally. (NFD because it allows consistent interleaving of markers.)

srl295: PTAL think i addressed this

docs/ldml/tr35-keyboards.md
```diff
@@ -629,6 +618,27 @@ Indicates that:
 2.  If a transform is escaped, output the contents of the buffer.
 3.  During a transform, hide the contents of the buffer as the user is typing.
 
+_Attribute:_ `normalization`
+
+> Normalization will not typically be the responsibility of the keyboard author, rather this will be managed by the implementation.
+> The implementation will apply normalization as appropriate when matching transform rules and `<display>` value matching.
+> Output from the keyboard, following application of all transform rules, will be normalized to implementation or application-requested form.
+>
+> However, it is recognized that there may be some keyboards which, for compatibility or legacy reasons, need to manage their own normalization. The implementation in that case will do no normalization at all. The keyboard author must make use of transforms in the keyboard to any required normalization.
```
mcdurdin, line 627: This needs to specify that `normalization="mixed"` is the attribute to enable this mode.

Perhaps it should be `normalization="disabled"` -- that seems clearer given the description

srl295: that makes some sense

mcdurdin (approved):

PR #3289, opened 2023-09-28
CLDR-17126 Update CLDR segment tailorings to reflect ICU 74 word break changes

pedberg-icu:
CLDR-17126

Update CLDR segmentation tailorings to reflect word break tailorings updates in ICU 74 (CLDR 44 did not make any tailoring updates that need to be synced to ICU).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/segments/en_US_POSIX.xml
+++ b/common/segments/en_US_POSIX.xml
@@ -16,7 +16,6 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<segmentation type="WordBreak">
 			<variables>
 				<variable id="$MidNumLet">[[$MidNumLet]-[.]]</variable>
-				<variable id="$MidLetter">[[$MidLetter]-[\:]]</variable>
 				<variable id="$MidNum">[[$MidNum] [.]]</variable>
 			</variables>
 		</segmentation>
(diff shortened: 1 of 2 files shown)


srl295:
Can we get a link to a ticket with the ICU side changes and rationale? (probably should go in this ticket)

pedberg-icu:
> Can we get a link to a ticket with the ICU side changes and rationale? (probably should go in this ticket)

The relevant ICU tickets are:
- First [ICU-22112](https://unicode-org.atlassian.net/browse/ICU-22112) word break updates for @,colon; colon tailorings for fi,sv
- Then [ICU-22360](https://unicode-org.atlassian.net/browse/ICU-22360) revert portions of ICU-22112 which included @ in ALetter for wordbreak, update tests

I will add this to the ticket as well.

markusicu (approved):
rslgtm

@eggrobin

eggrobin:
Scripsit @markusicu
> @eggrobin

Should we upstream these root tailorings to UAX29? (In which case a PAG issue needs to be filed; at a glance we would probably do that by changing the Word_Break property of the characters involved, though I would need to look at the context more closely to summarize it for a recommendation.)

markusicu:
> Should we upstream these root tailorings to UAX29? (In which case a PAG issue needs to be filed; at a glance we would probably do that by changing the Word_Break property of the characters involved, though I would need to look at the context more closely to summarize it for a recommendation.)

--> https://github.com/unicode-org/properties/issues/187

PR #3291, opened 2023-09-29
CLDR-17129 Fix inconsistency in fr dayPeriods for night1; always use [du] matin, not [de la] nuit

pedberg-icu:
CLDR-17129

Notice this when investigating a problem in production data with `fr_CA` abbreviated day periods longer than wide.

`fr` was inconsistent in naming of day periods. In most cases (including for different widths) is used forms of "matin" (morning) for both morning1 and night1 (as did `fr_CA`); however for format abbreviated it used "nuit" (night) for night1. I first thought the use of "matin" for night1 was an error until I read the Survey Tool forum comments; the vetters for fr (and fr_CA) all agreed that the period 00:00-04:00 (night1 for fr) should be referred to using morning ("matin"), along with the the period 04:00-12:00 (morning1). So the only necessary consistency change was to fix the one place that instead used "nuit" for night1, and change it to "matin".

Also (as a quick hack fix) allow display collisions between the names for night1 and either morning1 or am, if night1 starts at midnight. Filed https://unicode-org.atlassian.net/browse/CLDR-17132 for a better fix for this French night1 issue.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/fr.xml
+++ b/common/main/fr.xml
@@ -3026,7 +3026,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 							<dayPeriod type="morning1">matin</dayPeriod>
 							<dayPeriod type="afternoon1">après-midi</dayPeriod>
 							<dayPeriod type="evening1">soir</dayPeriod>
-							<dayPeriod type="night1">nuit</dayPeriod>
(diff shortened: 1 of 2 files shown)


srl295 (approved):
quick hack LGTM

pedberg-icu:
> quick hack LGTM

Thanks!

PR #3292, opened 2023-09-29
CLDR-17133 CLDRModify -fQ. update annotations from TTS name, some annotations got improperly merged

DraganBesevic:
CLDR-17133

CLDRModify no arguments:
- No changes

CLDRModify -fQ:
- Updated TTS names to annotations to remove the superfluous entries
- Had to manually restore draft="unconfirmed"  status that got removed for some regular keyword entries in 4 files: bew.xml, cv.xml, kab.xml, to.xml

CLDRModify -fP:
- No changes

ALLOW_MANY_COMMITS=true


pedberg-icu (changes requested):
This still has the problem of removing the draft="unconfirmed" status for the regular entry when both the regular and tts entries have draft="unconfirmed". That is non-starter. will check with Tom.

pedberg-icu (approved):

common/annotations/bew.xml
```diff
@@ -15,47 +15,47 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="😀" type="tts" draft="unconfirmed">muka girang</annotation>
 		<annotation cp="😃" draft="unconfirmed">lèbar | muka | nyengir | seneng | sumringah</annotation>
 		<annotation cp="😃" type="tts" draft="unconfirmed">muka nyengir lèbar</annotation>
-		<annotation cp="😄" draft="unconfirmed">girang | ketawa | mata ketutup | muka | seneng</annotation>
+		<annotation cp="😄">girang | ketawa | mata ketutup | muka | muka girang mata ketutup | seneng</annotation>
```
btangmu: @DraganBesevic @pedberg-icu I think I see why this happens: `CLDRModify.CLDRFilter.replace` is supposed to be called with `fullpath` not just the path gotten by looping through the CLDRFile

pedberg-icu:
> This still has the problem of removing the draft="unconfirmed" status for the regular entry when both the regular and tts entries have draft="unconfirmed". That is non-starter. will check with Tom.

Fixed that manually before merging.

common/annotations/bew.xml
btangmu: https://github.com/unicode-org/cldr/pull/3307

PR #3303, opened 2023-10-03
CLDR-16866 Add algorithm for first day of week.

macchiati:
CLDR-16866

Note: it is slightly different than the ticket algorithm. I realized that the sd can only conflict with the region subtag if there is an explicit region subtag, so reordering the explicit region subtag first makes the exposition a bit simpler.

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

sffc:
> Note: it is slightly different than the ticket algorithm. I realized that the sd can only conflict with the region subtag if there is an explicit region subtag, so reordering the explicit region subtag first makes the exposition a bit simpler.

This misses the point of putting subdivision first. Northern Island could have a different first day of week preference than United Kingdom.

PR #3304, opened 2023-10-03
CLDR-16403 Add v44 PersonName features

macchiati:
CLDR-16403

There is a lot of overlap between this ticket and https://unicode-org.atlassian.net/browse/CLDR-17124. I went ahead and included the extra text for nativeSpaceReplacement (replacing some old text), so I think we can mark https://unicode-org.atlassian.net/browse/CLDR-17124 as a duplicate.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-personNames.md
+++ b/docs/ldml/tr35-personNames.md
@@ -135,6 +135,13 @@ The following features are currently out of scope for Person Names formating:
 
 A draft API for formatting personal names is included in ICU4J 73. (“Draft” means that the full functionality is present, but the API might be refined before it is stabilized.) The implementation can be found at [PersonNameFormatter.java](https://github.com/unicode-org/icu/blob/main/icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java) and [SimplePersonName.java](https://github.com/unicode-org/icu/blob/main/icu4j/main/classes/core/src/com/ibm/icu/text/SimplePersonName.java).
 
+In addition to the settings in this document, it is recommended that implementations provide some additional features in their APIs to allow more control for clients, notably:
+
+1. forceGivenFirst — no matter what the values are in nameOrderLocales or in the NameObject, display the name as givenFirst.
+2. forceSurnameFirst — no matter what the values are in nameOrderLocales or in the NameObject, display the name as surnameFirst.
+3. forceNativeOrdering — no matter what the values are in nameOrderLocales or in the NameObject, display the name with the same ordering as the native locale.
+4. surnameFirstAllCaps — display the surname and surname2 fields in all caps **if** not using native order. Thus where the foreign name ordering is surnameFirst, the name {given=Shinzo, surname=Abe} would display as “ABE Shinzo”.
+
 ### Person Name Formatting Overview
 
 Logically, the model used for applying the CLDR data is the following:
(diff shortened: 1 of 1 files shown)


pedberg-icu (approved):

macchiati:
Ping?

pedberg-icu:
> Ping?

Looking at it now, was backed up behind some other also-urgent things

docs/ldml/tr35-personNames.md
```diff
@@ -135,6 +135,13 @@ The following features are currently out of scope for Person Names formating:
 
 A draft API for formatting personal names is included in ICU4J 73. (“Draft” means that the full functionality is present, but the API might be refined before it is stabilized.) The implementation can be found at [PersonNameFormatter.java](https://github.com/unicode-org/icu/blob/main/icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java) and [SimplePersonName.java](https://github.com/unicode-org/icu/blob/main/icu4j/main/classes/core/src/com/ibm/icu/text/SimplePersonName.java).
```
pedberg-icu, line 136: I will make a separate PR to update this line.

pedberg-icu (approved):

pedberg-icu:
I am going to go ahead and merge this. Any follow-on fixes can go into a separate PR.

docs/ldml/tr35-personNames.md
```diff
@@ -135,6 +135,13 @@ The following features are currently out of scope for Person Names formating:
...
 
+In addition to the settings in this document, it is recommended that implementations provide some additional features in their APIs to allow more control for clients, notably:
+
+1. forceGivenFirst — no matter what the values are in nameOrderLocales or in the NameObject, display the name as givenFirst.
+2. forceSurnameFirst — no matter what the values are in nameOrderLocales or in the NameObject, display the name as surnameFirst.
+3. forceNativeOrdering — no matter what the values are in nameOrderLocales or in the NameObject, display the name with the same ordering as the native locale.
```
richgillam, line 142: Not that it matters, but I didn't implement this in ICU.  I think that was partly because what we had at the time wasn't clear on just what this was supposed to do, but also because what I thought it was supposed to do was going to require a lot of extra work in the code that wouldn't normally be worth it.  Is this really worth doing?  If so, you might want to file an ICU ticket.

docs/ldml/tr35-personNames.md
```diff
@@ -135,6 +135,13 @@ The following features are currently out of scope for Person Names formating:
...
+4. surnameFirstAllCaps — display the surname and surname2 fields in all caps **if** not using native order. Thus where the foreign name ordering is surnameFirst, the name {given=Shinzo, surname=Abe} would display as “ABE Shinzo”.
```
richgillam, line 143: I don't think I implemented this this way-- I think my ICU code always capitalizes the surname, regardless of field order.  This is the first I've heard that we wanted to restrict that behavior to just names that are displayed in the opposite of their normal order.  Again, (without looking at the code right now) this seems like a significant amount of extra work at run time to get and check the order-- is it really worth it?  If so, please file an ICU ticket.

richgillam (commented):
Belated review.  Generally looks great, but there were a couple things in here that surprised me-- I've called those out below.

PR #3305, opened 2023-10-03
CLDR-16638 remove mention of "reformed" in spec

pedberg-icu:
CLDR-16638

However we should consider formally deprecating "reformed" as a collation type, I filed https://unicode-org.atlassian.net/browse/CLDR-17143 for that (and then we can get the Unicode FAQ updated, etc.). With this PR I have removed mention of it in the LDML spec. Note (as Markus pointed out) that clients who explicitly request the now-absent  "reformed" collation type in Swedish will fall back to getting the "standard" behavior, which is the desired result ("standard" has the data that used to be "reformed").

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/collation/sv.xml
+++ b/common/collation/sv.xml
@@ -37,5 +37,10 @@ For terms of use, see http://www.unicode.org/copyright.html
 				&[before 1]ǀ<å<<<Å<ä<<<Ä<<æ<<<Æ<<ę<<<Ę<ö<<<Ö<<ø<<<Ø<<ő<<<Ő<<œ<<<Œ<<ô<<<Ô
 			]]></cr>
 		</collation>
+		<collation type="reformed">
+			<cr><![CDATA[
+				[import sv-u-co-standard]
+			]]></cr>
+		</collation>
 	</collations>
 </ldml>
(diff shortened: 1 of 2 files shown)


common/collation/sv.xml
```diff
@@ -37,5 +37,10 @@ For terms of use, see http://www.unicode.org/copyright.html
 				&[before 1]ǀ<å<<<Å<ä<<<Ä<<æ<<<Æ<<ę<<<Ę<ö<<<Ö<<ø<<<Ø<<ő<<<Ő<<œ<<<Œ<<ô<<<Ô
 			]]></cr>
 		</collation>
+		<collation type="reformed">
+			<cr><![CDATA[
+				[import sv-u-co-standard]
+			]]></cr>
+		</collation>
```
markusicu, line 44: Why do we need this?
The regular fallback already goes to the default collation tailoring, which is normally "standard". (It explicitly says so in this file on line 14. That's also the default if that line was missing.)
https://www.unicode.org/reports/tr35/tr35-collation.html#Collation_Type_Fallback

Adding this tailoring here will duplicate the data in ICU, making everyone want to cut it back out.

markusicu (changes requested):

common/collation/sv.xml
pedberg-icu: @markusicu Ah thanks for pointing that out, I did not realize that. I have removed the restoration of "reformed" in Swedish collation.

markusicu (approved):
lgtm tnx

PR #3310, opened 2023-10-03
CLDR-17091 kbd: move names into info name="…"

srl295:
- drop the <names> element
- make <info> required
- add <info name="…"> as a required attribute
- update docs and dtd

CLDR-17091

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -74,8 +74,6 @@ The LDML specification is divided into the following parts:
   * [Element: locale](#Element_locale)
   * [Element: version](#Element_version)
   * [Element: info](#Element_info)
-  * [Element: names](#Element_names)
-  * [Element: name](#Element_name)
   * [Element: settings](#Element_settings)
   * [Element: keys](#Element_keys)
   * [Element: key](#Element_key)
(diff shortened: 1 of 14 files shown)


mcdurdin (dismissed):

docs/ldml/tr35-keyboards.md
```diff
@@ -496,10 +496,27 @@ Element containing informative properties about the layout, for displaying in us
...
-> Occurrence: optional, single
+> Occurrence: required, single
 >
 > </small>
 
+_Attribute:_ `name` (required)
+
+> This is the only required attribute.
+>
+> This name is not localized, but is an informative name for the keyboard.
```
miloush, line 507: When you say not localized, does that mean it is in the primary locale of the keyboard? Or English? Or arbitrary locale of choice of the author? 

Not sure why there is _but_ here.

srl295: I'll take out 'but', and 'not localized' doesn't add a lot here.  It was meant to say that there weren't multiple names per language for the keyboard

miloush: I think you raised a good point. If not giving guidance to authors, we should at least have agreed understanding of what do we expect people to put here in terms of language.

srl295: it could say 'this name may be in the primary language of the keyboard' or something?  `Bulgarian (Phonetic Traditional)` isn't a great example then.

srl295:
@miloush @mcdurdin thoughts?

miloush (dismissed):

srl295:
Sorry, need another rubber stamp.

- https://github.com/unicode-org/cldr/compare/c8f2fecd471ed8357ba1f123c782ee4ac166c809..29af7f29f53512584821850e230bb30934cbba66

You can see from the link that there's no real change. But somehow the merge from main didn't work properly.

srl295:
merge-from-main didn't seem to work well here.

miloush (approved):

mcdurdin (approved):

PR #3313, opened 2023-10-03
CLDR-17113 kbd: element/attr renaming

srl295:
- layers/@form > layers/@formId
- key/@flicks > key/@flickId
- key/@switch >  key/@layerId
- key/@to > key/@output
- display/@to > display/@output

- updates to dtd, samples
- drop some old data in tests

CLDR-17113

ALLOW_MANY_COMMITS=true


srl295:
needs spec yet

docs/ldml/tr35-keyboards.md
```diff
@@ -879,7 +879,7 @@ _Attribute:_ `id` (required)
 **Syntax**
 
 ```xml
-<flickSegment directions="{list of directions}" to="{the output}" />
+<flickSegment directions="{list of directions}" "{the output}" />
```
mcdurdin, line 882:
```suggestion
<flickSegment directions="{list of directions}" output="{the output}" />
```

docs/ldml/tr35-keyboards.md
```diff
@@ -1090,19 +1090,19 @@ This attribute may be escaped with `\u` notation, see [Escaping](#Escaping).
 ```xml
 <keyboard3>
     <keys>
-        <key id="a" to="a" longpress="\u{0301} \u{0300}" />
-        <key id="shift" switch="shift" />
+        <key id="a" output="a" longpress="\u{0301} \u{0300}" />
+        <key id="shift" layerId="shift" />
```
mcdurdin, line 1094: Note: a limitation of longpress is we can't do e.g. `longpress="ctrl"` to get to the ctrl layer

mcdurdin (dismissed):
Minor typo apart from that lgtm

docs/ldml/tr35-keyboards.md
miloush: One way to solve this _if needed_ would be to make the `longpress` refer to key ids

docs/ldml/tr35-keyboards.md
```diff
@@ -1065,23 +1065,23 @@ The `display` element describes how a character, that has come from a `keys/key`
...
 
 > Specifies the character or character sequence from the `keys/key` element that is to have a special display.
 > This attribute may be escaped with `\u` notation, see [Escaping](#Escaping).
-> The `to` attribute may also contain the `\m{…}` syntax to reference a marker. See [Markers](#markers). Implementations may highlight a displayed marker, such as with a lighter text color, or a yellow highlight.
+> The `output` attribute may also contain the `\m{…}` syntax to reference a marker. See [Markers](#markers). Implementations may highlight a displayed marker, such as with a lighter text color, or a yellow highlight.
 
 _Attribute:_ `id` (optional)
 
-> Specifies the `key` id. This is useful for keys which do not produce any output (no `to=` value), such as a shift key.
+> Specifies the `key` id. This is useful for keys which do not produce any output (no `output=` value), such as a shift key.
```
miloush, line 1078: Since we currently don't use any frame keys, perhaps we shouldn't use them as examples.

```suggestion
> Specifies the `key` id. This is useful for keys which do not produce any output (no `output=` value), such as layer switching keys.
```

docs/ldml/tr35-keyboards.md
mcdurdin: That would certainly be elegant structurally, but it would then require another lot of key elements to do the mapping to outputs, and _those_ keys would not be able to have longpress themselves, etc. Would require a significant reshuffle of the current design.

miloush: They could. Longpress keys could be switching layers, and - if the options are presented in the UI - have multi tap, flicks or more longpress keys (nobody says the UI must dismiss as soon as you touch it) or assigned a `display` element. Implementations can simply not support it. I am not sure you need _that_ much structural redesign, but also I am not sure how useful this would be and the cost of that is more verbosity in the xml, hence emphasis on "if needed".

Compare this with multi tap situation. In this case, if you wanted to allow layer switching keys or long press keys, then you need significant implementation redesign as things change under your hands, and allowing "nested" multi tap does not make sense in the first place.

mcdurdin: My feeling on this is: layering gestures on gestures would lead to significant implementation complexity, a big decrease in usability and discoverability, and very little additional benefit.

As far as I can tell, only the `layerId` attribute is important as an attribute for any gesture outputs. But if we did refactor in this way, then it would be nice to also specify the `longPressDefault` as an attribute of the longPress rather than the pattern we have right now `<longPress default='true'>`.

@DavidLRowe noted that there are extant examples of keyboards in the Keyman keyboards repository that have layerId-type actions assigned to longpress. My quick and somewhat rough search located at least 81 keyboards. There are two kinds of uses:

1. Keyboards that have a modifier key with additional layer switch keys in the longpress, e.g. [koreguaje](https://github.com/keymanapp/keyboards/blob/59c44e8384b736f5f8e51a0af39f91bc677f0cce/experimental/k/koreguaje/source/koreguaje.keyman-touch-layout#L248-L269). As far as I have checked, these are mostly translations from desktop layouts.

2. Keyboards that emit an output and then swap to another layer. This pattern is widely used, e.g. in [naijatype](https://github.com/keymanapp/keyboards/blob/1729d48b84c535d0f783667ee76c0dc9c50259a5/experimental/n/naijatype/source/naijatype.keyman-touch-layout#L10-L78). This is an important usability pattern -- and restricting that ability to base keys seems like a poor outcome.

Given the current restriction we have with CLDR's XML structure, supporting this would lead to significantly more verbose files.

Is it too late to be shuffling these deck chairs?

srl295: maybe for consistency, the flick output should _also_ be a key.

That would allow `<display>` to affect flick output. 

> Is it too late to be shuffling these deck chairs?

I think the too-late ship has already sailed in a sense, that we're already renaming elements.

I'll update.

srl295:
comments from review with @mcdurdin  and @rc-swag
- drop `keyLists` and move id lists into attributes
- add `<key longPressDefault=true` 
- disallow multiTapKeyIds to refer to itself

miloush:
Not sure I quite understand `longPressDefault=true`, can we see some example usage?

miloush:
Maybe this should be done in a separate PR, it's getting a bit crowded.

The PR now has
```xml
<keyList id="lower-a" defaultKeyId="a-grave" keyIds="a-grave a-caret a-acute a-umlaut a-tilde a-ring a-caron" />
```

You said it's moving from `key` to `longpress` but the latest comments from @srl295 says `key longPressDefault=` not `longpress default=` and also suggests to drop the `keyLists` altogether?

srl295:
> Not sure I quite understand `longPressDefault=true`, can we see some example usage?

OK I see. I just took notes from a review with @mcdurdin and @rc-swag , I didn't actually make the changes yet

miloush:
I see now what was meant. Why is that better than having `longPressDefaultKeyId`? This change would introduce a requirement that the default long press is in the list of long presses, but if there is only long press default and no long presses, implementation does not need to display any UI and just perform the long-press action.

srl295:
> I see now what was meant. Why is that better than having `longPressDefaultKeyId`? This change would introduce a requirement that the default long press is in the list of long presses, but if there is only long press default and no long presses, implementation does not need to display any UI and just perform the long-press action.

I can see that— that also means implementations don't need to enforce "only one default". 

I think the longPress list of a key should be able to include itself, right?

srl295:
> I see now what was meant. Why is that better than having `longPressDefaultKeyId`? This change would introduce a requirement that the default long press is in the list of long presses, but if there is only long press default and no long presses, implementation does not need to display any UI and just perform the long-press action.

@miloush PTAL

srl295:
also i should have done this before, but i disabled the `kbd-check` workflow. It won't make sense until I get to re-implement all of these changes.

miloush:
> I think the longPress list of a key should be able to include itself, right?

I don't expect keyman to support nested long presses but as spec goes that sounds fine.

mcdurdin:
> > > I think the longPress list of a key should be able to include itself, right?
> > 
> > 
> > I don't expect keyman to support nested long presses but as spec goes that sounds fine.
> 
> It's not that the long press is nested. It's that long pressing on a key `a` can have `a` as one of the long press choices.

I think the spec should note that implementations are not expected to support nested long press (or other gestures), otherwise authors will depend on this behaviour and be disappointed.

srl295:
> > > > I think the longPress list of a key should be able to include itself, right?
> > > 
> > > 
> > > I don't expect keyman to support nested long presses but as spec goes that sounds fine.
> > 
> > 
> > It's not that the long press is nested. It's that long pressing on a key `a` can have `a` as one of the long press choices.
> 
> I think the spec should note that implementations are not expected to support nested long press (or other gestures), otherwise authors will depend on this behaviour and be disappointed.

Noted PTAL

docs/ldml/tr35-keyboards.md
```diff
@@ -666,6 +666,8 @@ _Attribute:_ `longPressKeyIds="{list of key ids}"` (optional)
 >
 > In a list of keys specified by `longPressKeyIds`, the key matching `longPressDefaultKeyId` attribute (if present) specifies the default long-press target, which could be different than the first element. It is an error if the `longPressDefaultKeyId` key is not in the `longPressKeyIds` list.
 >
+> Gestures are not nested: other gestures (such as flick, multiTap, longPress) are ignored on the keys in the `longPressKeyIds` list.
```
miloush, line 669: Like before, do not use "are" and similar factual statements. It is unclear who they target and what they mean. Say "Implementations may ignore/not support" or shall not or whatever is appropriate.

srl295: Well put. I will revise.

srl295: updated

miloush:
I see you went with an implementation supporting these to be non-conforming... :-)

miloush (approved):

PR #3314, opened 2023-10-04
CLDR-17140 kbd: specify modifier matching

srl295:
CLDR-17140

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -90,6 +90,9 @@ The LDML specification is divided into the following parts:
   * [Element: scanCodes](#Element_scanCodes)
   * [Element: layers](#Element_layers)
   * [Element: layer](#Element_layer)
+    * [Layer Modifier Components](#layer-modifier-components)
+    * [Modifier Left- and Right- keys](#modifier-left--and-right--keys)
+    * [Layer Modifier Matching](#layer-modifier-matching)
   * [Element: row](#Element_row)
   * [Element: variables](#Element_variables)
   * [Element: string](#element-string)
@@ -1384,41 +1387,76 @@ _Attribute:_ `modifier` (required for `hardware`)
 
 > This has two roles. It acts as an identifier for the `layer` element for hardware keyboards (in the absence of the id= element) and also provides the linkage from the hardware modifiers into the correct `layer`.
 >
-> To indicate that no modifiers apply, the reserved name of `none` is used.
-> The following modifier components can be used, separated by spaces.
-> Note that `L` or `R` indicates a left- or right- side modifier only (such as `altL`)
-> whereas `alt` indicates _either_ left or right alt key (that is, `altL` or `altR`). `ctrl` indicates either left or right ctrl key (that is, `ctrlL` or `ctrlR`).
-> `shift` also indicates either shift key. The left and right shift keys are not distinguishable in this specification.
->
-> If there is a layer with a modifier `alt`, there may not be another layer with `altL` or `altR`.  Similarly, if there is a layer with a modifier `ctrl`, there may not be a layer with `ctrlL` or `ctrlR`.
->
-> - `none` (no modifier, may not be combined with others)
-> - `alt`
-> - `altL`
-> - `altR`
-> - `caps`
-> - `ctrl`
-> - `ctrlL`
-> - `ctrlR`
-> - `shift`
->
-> Note that `alt` in this specification is referred to on some platforms as "opt" or "option".
->
-> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
->
 > For hardware layouts, the use of `@modifier` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
 >
 > The set of modifiers must match `(none|([A-Za-z0-9]+)( [A-Za-z0-9]+)*)`
 >
-> To share a layer between two modifier sets, the layer data must be duplicated.
+> To indicate that no modifiers apply, the reserved name of `none` is used.
+
+**Syntax**
+
+```xml
+<layer id="base"        modifier="none">
+    <row keys="a" />
+</layer>
+
+<layer id="upper"       modifier="shift">
+    <row keys="A" />
+</layer>
+
+<layer id="altgr"       modifier="altR">
+    <row keys="a-umlaut" />
+</layer>
+
+<layer id="upper-altgr" modifier="altR shift">
+    <row keys="A-umlaut" />
+</layer>
+```
+
+#### Layer Modifier Components
+
+ The following modifier components can be used, separated by spaces.
+
+ - `none` (no modifier, may not be combined with others)
+ - `alt`
+ - `altL`
+ - `altR`
+ - `caps`
+ - `ctrl`
+ - `ctrlL`
+ - `ctrlR`
+ - `shift`
+
+(`alt` in this specification is referred to on some platforms as "opt" or "option".)
+
+#### Modifier Left- and Right- keys
+
+1. `L` or `R` indicates a left- or right- side modifier only (such as `altL`)
+ whereas `alt` indicates _either_ left or right alt key (that is, `altL` or `altR`). `ctrl` indicates either left or right ctrl key (that is, `ctrlL` or `ctrlR`).
+
+2.  If there are any layers with a modifier `alt`, there may not also be another layer with `altL` or `altR`.  Similarly, if there is a layer with a modifier `ctrl`, there may not be a layer with `ctrlL` or `ctrlR`.
+
+3. Left- and right- side modifiers may not be mixed together in a single `modifier` attribute value, so neither `altL ctrlR"` nor `altL altR` are allowed.
+
+4. `shift` indicates either shift key. The left and right shift keys are not distinguishable in this specification.
+
+#### Layer Modifier Matching
+
+Layers are matched exactly based on the modifier keys which are down. For example:
+
+- `none` as a modifier will only match if *all* of the keys `caps`, `alt`, `ctrl` and `shift` are up.
+
+- `alt` as a modifier will only match if either `alt` is down, *and* `caps`, `ctrl`, and `shift` are up.
+
+- `altL ctrl` as a modifier will only match if the left `alt` is down, either `ctrl` is down, *and* `shift` and `caps` are up.
+
+- To share a layer between two modifier sets, the layer data must be duplicated.
 
 * * *
 
 ### <a name="Element_row" href="#Element_row">Element: row</a>
 
 A `row` element describes the keys that are present in the row of a keyboard.
-
-
 
 **Syntax**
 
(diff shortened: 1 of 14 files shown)


srl295:
this is less powerful, but simpler to explain and understand.

mcdurdin (dismissed):
LGTM. 

But I would be okay with supporting `modifier="none, shift caps"` to match two different states. Another example would be `modifier="ctrlL altL, altR"` pattern that Windows uses as mentioned by @xadxura. That would go a long way towards reducing duplicated layers.

docs/ldml/tr35-keyboards.md
```diff
@@ -1341,42 +1344,77 @@ _Attribute:_ `modifier` (required for `hardware`)
...
+
+2.  If there are any layers with a modifier `alt`, there may not also be another layer with `altL` or `altR`.  Similarly, if there is a layer with a modifier `ctrl`, there may not be a layer with `ctrlL` or `ctrlR`.
+
+3. Left- and right- side modifiers may not be mixed together in a single `modifier` attribute value, so neither `altL ctrlR"` nor `altL altR` are allowed.
+
+4. `shift` indicates either shift key. The left and right shift keys are not distinguishable in this specification.
+
+#### Layer Modifier Matching
+
+Layers are matched exactly based on the modifier keys which are down. For example:
```
macchiati, line 1402: **Problems**

1. The attribute name `modifier` is confusing (as it was in the previous version of keyboards). Please change it to `modifiers`.

The other issues are more serious. I do not think this PR approaches a solution to the problems raised earlier. Without a solution, I don't think we can advance beyond tech preview.

1. The proposed algorithm is still under-specified. In particular, it does not indicate what to do in either of these cases:
    1. None of the layers match. **It is very user-unfriendly to have "gaps" where nothing happens.**
    2. More than one of the layers match.
2. The algorithm and structure are underpowered. For more complex cases (as in **actual keyboards** captured by the previous keyboard structure) one would have to repeat entire layers multiple times. For example, having the caps key down should not affect the result of a control layer; nor an alt layer. But that means duplicating layers in order to have the "crtl caps" and "ctrl", even through the same results obtain.

There are various options to solve these.

1.1 No match
  A. There could be a separate structure that indicates which layer is used if nothing matches.
  B. There could be a special 'modifier' like 'default' that indicates which layer is used.
  C. This relates to a solution for #3

1.2 Multiple matches
  A. Require reading the layers in order.
  B. Declare that such a keyboard is ill-formed, and cannot be interpreted.

2. Underpowered
  A. Revert to the previous keyboard modifier model: modifiers="ctrl caps?"
  B. Introduce an 'or'. That is modifiers="ctrl caps or ctrl" would match either "ctrl caps" or "ctrl". This isn't as concise as the old model, but is trivial to describe and implement.

macchiati (changes requested):

srl295:
@macchiati very briefly, thanks for reviewing, also note that SC is **not** requesting exit from tech preview in 44, per your suggestion.

srl295:
@macchiati  

> gaps where nothing happens

Is that really an issue?

In any event, perhaps could be a `default` layer that matches if nothing else does. 

> More than one layer matches

I don't think this is allowed by the proposal as given.  If there's an overlap between layers, there should be an error given.

> repeat entire layers multiple times

We had talked about allowing multiple modifier sets on one layer.  While the prior syntax may be well specified and powerful, the SC thought that it is more challenging to understand.

docs/ldml/tr35-keyboards.md
srl295: @macchiati the current PR allows multiple modifier sets, so would not require as much repetition.

> For more complex cases (as in actual keyboards captured by the previous keyboard structure) 

These can be helpful for comparison, but it should be noted that it's not a goal to be able to capture or support all actual keyboards.

srl295:
@macchiati comments on the latest?

srl295:
@miloush thoughts on this one?

docs/ldml/tr35-keyboards.md
```diff
@@ -1373,46 +1374,90 @@ _Attribute_ `id` (required for `touch`)
...
+3. Only one layer may have `none`, and only one layer may have `other` specified.
+
+#### Modifier Left- and Right- keys
+
+1. `L` or `R` indicates a left- or right- side modifier only (such as `altL`)
+ whereas `alt` indicates _either_ left or right alt key (that is, `altL` or `altR`). `ctrl` indicates either left or right ctrl key (that is, `ctrlL` or `ctrlR`).
+
+2.  If there are any layers with a modifier `alt`, there may not also be another layer with `altL` or `altR`.  Similarly, if there is a layer with a modifier `ctrl`, there may not be a layer with `ctrlL` or `ctrlR`.
+
+3. Left- and right- side modifiers may not be mixed together in a single `modifier` attribute value, so neither `altL ctrlR"` nor `altL altR` are allowed.
```
miloush, line 1435: This looks arbitrary, why? If implementation cannot distinguish left and right keys that's one thing, but why does the specification need such limitation?

docs/ldml/tr35-keyboards.md
miloush: I agree that it is not the goal to support all existing keyboards, but also that it would help adoption if vendors could switch to it for their data.

That said, Windows only needs repeated layers to cover altR = ctrl+alt cases. Does @macchiati have any specific example in mind which would need lots of repetition?

docs/ldml/tr35-keyboards.md
```diff
@@ -1373,46 +1374,90 @@ _Attribute_ `id` (required for `touch`)
...
+ - `ctrlL`
+ - `ctrlR`
+ - `shift`
+ - `other` (matches if no other layers match)
+
+1. `alt` in this specification is referred to on some platforms as "opt" or "option".
+
+2. `none` and `other` may not be combined with any other components.
+
+3. Only one layer may have `none`, and only one layer may have `other` specified.
```
miloush, line 1426: per form, presumably. Why would these be different than other modifier combinations?

srl295: good point. I'll drop this.

docs/ldml/tr35-keyboards.md
srl295: I lost my comment here, but basically, is this needed in real keyboards we expect to be developed? It can be done later very easily if there's a solid use case.

srl295: +1

mcdurdin: From my end, I think this was just a usability thing. No technical reason I am aware of (AFAIK, hardware keyboards don't struggle with those L/R modifier chording pairs).

srl295:
@miloush Is this an improvement? If so we can merge this and continue making editorial improvements?

tools/cldr-code/src/test/resources/org/unicode/cldr/tool/KeyboardFlatten/broken-import-missing.xml
```diff
@@ -45,31 +45,31 @@
...
+        <layer modifiers="altR">
             <row keys="grave gap gap pound gap gap gap gap gap gap gap gap gap" />
             <row keys="gap gap e-grave gap gap gap u-grave i-grave o-grave gap open-square close-square" />
             <row keys="a-grave gap gap gap gap gap gap gap gap gap gap gap" />
             <row keys="backslash gap gap gap gap gap gap gap gap gap gap" />
             <row keys="space" />
         </layer>
 
-        <layer modifier="altR-shift">
+        <layer modifiers="altR-shift">
```
miloush, line 72: Hyphen? I don't think that is the syntax described above, and the DTD pattern does not allow hyphens (though maybe it should, together with underscores).

tools/cldr-code/src/test/resources/org/unicode/cldr/tool/KeyboardFlatten/broken-import-unknownver.xml
```diff
@@ -47,31 +47,31 @@
...
+        <layer modifiers="altR">
             <row keys="grave gap gap pound gap gap gap gap gap gap gap gap gap" />
             <row keys="gap gap e-grave gap gap gap u-grave i-grave o-grave gap open-square close-square" />
             <row keys="a-grave gap gap gap gap gap gap gap gap gap gap gap" />
             <row keys="backslash gap gap gap gap gap gap gap gap gap gap" />
             <row keys="space" />
         </layer>
 
-        <layer modifier="altR-shift">
+        <layer modifiers="altR-shift">
```
miloush, line 74: hyphen

tools/cldr-code/src/test/resources/org/unicode/cldr/tool/KeyboardFlatten/broken-import-wrongparent.xml
```diff
@@ -45,31 +45,31 @@
...
+        <layer modifiers="altR">
             <row keys="grave gap gap pound gap gap gap gap gap gap gap gap gap" />
             <row keys="gap gap e-grave gap gap gap u-grave i-grave o-grave gap open-square close-square" />
             <row keys="a-grave gap gap gap gap gap gap gap gap gap gap gap" />
             <row keys="backslash gap gap gap gap gap gap gap gap gap gap" />
             <row keys="space" />
         </layer>
 
-        <layer modifier="altR-shift">
+        <layer modifiers="altR-shift">
```
miloush, line 72: hyphen

docs/ldml/tr35-keyboards.md
```diff
@@ -1373,46 +1374,88 @@ _Attribute_ `id` (required for `touch`)
...
-> - `ctrlL`
-> - `ctrlR`
-> - `shift`
->
-> Note that `alt` in this specification is referred to on some platforms as "opt" or "option".
->
-> Left- and right- side modifiers (such as `"altL ctrlR"` or `"altL altR"`) should not be used together in a single `modifier` attribute value.
->
-> For hardware layouts, the use of `@modifier` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
+> For hardware layouts, the use of `@modifiers` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in a keyboard.
```
miloush, line 1381: ...in a form (rather than keyboard). This will not fly if the layer needs to be refered by a "switch to" key, but that is now out of scope for this PR. I will have to look into the whole modifiers model, but let's do that separately.

tools/cldr-code/src/test/resources/org/unicode/cldr/tool/KeyboardFlatten/broken-import-wrongparent.xml
srl295: thanks. these are test files for import, so they fail before they get to check modifiers. But will fix.

docs/ldml/tr35-keyboards.md
srl295: PTAL

miloush (approved):

srl295:
Thanks. Two reviews, planning to merge this.

srl295:
Post script:  ticket https://unicode-org.atlassian.net/browse/CLDR-17177 is for revisiting modifier matching in v45

PR #3318, opened 2023-10-04
CLDR-17128 GenerateProductionData should not add defaultNumberingSystem items to ar_001

btangmu:
-The method localeIsSpecial should return false for the default content locale ar_001

CLDR-17128

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
@@ -575,7 +575,7 @@ public class GenerateProductionData {
     }
 
     private static boolean localeIsSpecial(String localeId) {
-        return localeId.equals("ar") || localeId.startsWith("ar_");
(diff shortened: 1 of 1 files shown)


DraganBesevic (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
```diff
@@ -575,7 +575,7 @@ private static boolean directoryIsSpecial(String directory) {
     }
 
     private static boolean localeIsSpecial(String localeId) {
-        return localeId.equals("ar") || localeId.startsWith("ar_");
+        return localeId.equals("ar") || (localeId.startsWith("ar_") && !"ar_001".equals(localeId));
```
pedberg-icu, line 578: This works for now. In the long run it might be better to check whether a locale is the default content locale (there is info in supplementalMetadata for that), since we could potentially change which locale is the default content locale for `ar`.

pedberg-icu (approved):

pedberg-icu:
I am going to merge so we can include in beta3 tag...

PR #3322, opened 2023-10-05
CLDR-16249 Describe EBNF syntax more clearly

macchiati:
CLDR-16249

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -246,7 +246,13 @@ External specifications may also reference particular components of Unicode loca
 
 > _Field X can contain any Unicode region subtag values as given in Unicode Technical Standard #35: Unicode Locale Data Markup Language (LDML), excluding grouping codes._
 
+### EBNF
+The BNF syntax used in LDML is a variant of the Extended Backus-Naur Form (EBNF) notation used in [W3C XML Notation](https://www.w3.org/TR/REC-xml/#sec-notation). The main differences are:
 
+1. Bounded repetition following Perl regex syntax is allowed, such as alphanum{3,8}
+2. Constraints (well-formedness or validity) use separate notes
+
+In the text, this is sometimes referred to as "EBNF (Perl-based)".
 
 ## <a name="Locale" href="#Locale">What is a Locale?</a>
 


pedberg-icu (approved):

pedberg-icu:
Going ahead and merging this...

docs/ldml/tr35.md
```diff
@@ -246,7 +246,13 @@ External specifications may also reference particular components of Unicode loca
 
 > _Field X can contain any Unicode region subtag values as given in Unicode Technical Standard #35: Unicode Locale Data Markup Language (LDML), excluding grouping codes._
 
+### EBNF
+The BNF syntax used in LDML is a variant of the Extended Backus-Naur Form (EBNF) notation used in [W3C XML Notation](https://www.w3.org/TR/REC-xml/#sec-notation). The main differences are:
 
+1. Bounded repetition following Perl regex syntax is allowed, such as alphanum{3,8}
+2. Constraints (well-formedness or validity) use separate notes
```
gibson042, line 253: @macchiati this list misses some differences that I see in the document.
```suggestion
The BNF syntax used in LDML is a variant of the Extended Backus-Naur Form (EBNF) notation used in [W3C XML Notation](https://www.w3.org/TR/REC-xml/#sec-notation). The main differences are:

1. Bounded repetition following Perl regex syntax is allowed, such as `alphanum{3,8}`
2. Whitespace inside bracketed enumerations and ranges is ignored (e.g., `[A-Z a-z]` is the same as `[A-Za-z]`)
3. A backslash may be used to escape a following "x"-prefixed hexadecimal code point (e.g., `\x20` is the same as `#x20`) or the immediately following character (e.g., `[\&\-]` is the same as `[#x26#x2D]`)
4. Constraints (well-formedness or validity) use separate notes
```

(backslash escaping appears in the [Unicode Sets](https://unicode.org/reports/tr35/#Unicode_Sets) grammar, which could and probably should be expressed without it)

srl295:
@gibson042  comment was post merge, i'll add a note on the ticket

docs/ldml/tr35.md
pedberg-icu: I filed https://unicode-org.atlassian.net/browse/CLDR-17210 to pick up the suggested additions

gibson042: Is there a process by which I can submit this kind of fix? Looking over [CONTRIBUTING.md](https://github.com/unicode-org/cldr/blob/main/CONTRIBUTING.md), `docs/` is not mentioned in [Areas where contributions are welcome](https://github.com/unicode-org/cldr/blob/main/CONTRIBUTING.md#areas-where-contributions-are-welcome).

macchiati: Yes, these can be submitted in the docs/ directory. And thanks for your reviews!

gibson042: https://unicode-org.atlassian.net/browse/CLDR-17498

gibson042: > (backslash escaping appears in the [Unicode Sets](https://unicode.org/reports/tr35/#Unicode_Sets) grammar, which could and probably should be expressed without it)

@macchiati To elaborate on this point: [UnicodeSet syntax](https://unicode.org/reports/tr35/#unicodeset-syntax) includes expressions like `s [A-Za-z0-9] [A-Za-z0-9_\x20]* s`, in which `\x20` is to be interpreted as matching U+0020 SPACE. That table also includes `[[\u0000-\U00010FFFF]-[uxUN]]`, in which `\u0000` is to be interpreted as matching U+0000, `\U00010FFFF` is to be interpreted as matching U+10FFFF, `[\u0000-\U00010FFFF]` is to be interpreted as matching any code point in the inclusive range between those two (i.e., any code point), and the whole expression is to be interpreted as matching any code point other than "u", "x", "U", or "N" (a clever use of UnicodeSets in defining UnicodeSets, but one that does not conform with the defined BNF syntax). The latter expression should be replaced with something like `[^uxUN]`, leaving only `\x` escapes as described in point 3 of my above suggestion (which I have just modified to disallow `\u`, `\U`, and `\N` in anticipation of this issue recurring).

P.S. There's also a typo in that table—`pValuePerl` should use `[^ \\ \}]` rather than `[^\}]`.

macchiati:
Could you capture this in the ticket?

PR #3324, opened 2023-10-05
CLDR-17153 Investigate discrepancy between growth and coverage charts

macchiati:
CLDR-17153

There is a file that I mistakenly included. It is, however, a useful tool for comparing emoji so I'd keep it.

ALLOW_MANY_COMMITS=true


macchiati:
We can talk tomorrow about whether to merge this or not.

PR #3327, opened 2023-10-08
CLDR-16617 Additional tweak to transforms spec

macchiati:
CLDR-16617

A remaining item that was missed.

ALLOW_MANY_COMMITS=true


srl295:
needs to be on maint-44 or retargetted at 45

pedberg-icu (approved):

PR #3330, opened 2023-10-10
CLDR-17141 kbd: drop transform=no

srl295:
fyi @jahorton 

- expand documentation about use of markers to inhibit transforms

CLDR-17141

- [X] This PR completes the ticket.

also see #3332  which is a sub-PR of this

ALLOW_MANY_COMMITS=true

PR #3332, opened 2023-10-10
CLDR-17141 kbd: keep transform=no but document it

srl295:
CLDR-17141

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -624,6 +624,7 @@ This element defines a mapping between an abstract key and its output. This elem
  stretch="true"
  layerId="{switch layer id}"
  output="{the output}"
+ transform="no"
  width="{key width}"
  />
 ```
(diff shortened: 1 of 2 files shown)


mcdurdin:
I'd prefer to drop transform=no and simplify the mental model for keyboard authors. transform=no may _seem_ simpler, but the caveats mean they have to have some grasp of transforms anyway, and then it just becomes an additional layer to remember.

srl295:
@Jahorton fyi

srl295:
@miloush  thoughts here? @mcdurdin requesting to drop (meaning i can just remove this PR).   

I agree that the complexity and limitations of `transform=no` make it very limited use, especially this:

> the caveats mean they have to have some grasp of transforms anyway

As "syntactic sugar" it may not add a lot. Probably better to work on documentation, and tooling (including debugging and test environments, test cases, etc) for transforms.

srl295:
@miloush Are you OK if we close this one (that is, drop `transform=no`) and merge #3330 ?

miloush (approved):

miloush:
Oh sorry I meant yes, feel free to drop.

srl295:
@miloush sorry, i merged the wrong PR 😆

PR #3335, opened 2023-10-11
CLDR-15034 cherry-pick CLDR-17141 kbd: drop transform=no

srl295:
- expand documentation about use of markers to inhibit transforms

(cherry picked from commit 1a52361654f5d9c58f084d1d919080423862975a) #3330 

CLDR-15034

- [X] This PR completes the ticket!!

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -625,7 +625,6 @@ This element defines a mapping between an abstract key and its output. This elem
  stretch="true"
  layerId="{switch layer id}"
  output="{the output}"
- transform="no"
  width="{key width}"
  />
 ```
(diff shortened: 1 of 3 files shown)


docs/ldml/tr35-keyboards.md
```diff
@@ -1764,6 +1704,58 @@ Consider the following abbreviated example:
...
+    <key id="e" output="e" />
+</keys>
+…
+<transforms …>
+    <!-- Won't match ^e without marker. -->
+    <transform from="^\m{transform}e" output="ê"/>
+</transforms>
+```
+
+In this way, only the `X`, `e` keys will produce `^e` with a _transform_ marker (again, any name could be used) which will cause the transform to be applied. One benefit is that clicking or using the arrow key to navigate existing text with `^e` will never be affected by the transform, because the marker is not or no longer present.
```
mcdurdin, line 1757:
```suggestion
In this way, only the `X`, `e` keys will produce `^e` with a _transform_ marker (again, any name could be used) which will cause the transform to be applied. One benefit is that navigating to an existing `^` in a document and adding an `e` will result in `^e`, and not be affected by the transform, because there will be no marker present there (as markers are not stored with the document but only recorded in memory temporarily during text input).
``` 

I think `is not or no` is a bit confusing :dizzy:

docs/ldml/tr35-keyboards.md
```diff
@@ -1764,6 +1704,58 @@ Consider the following abbreviated example:
...
+
+```xml
+<keys>
+    <key id="caret" output="^\m{no_transform}"/>
+    <key id="X" output="^" />
+    <key id="e" output="e" />
+</keys>
+…
+<transforms>
+    <!-- this wouldn't match the key X output because of the marker -->
```
mcdurdin, line 1735: !! This is backwards

```suggestion
    <!-- this wouldn't match the key caret output because of the marker -->
```

srl295:
this is the cherry pick though…

mcdurdin:
RSLGTM

mcdurdin (approved):

PR #3341, opened 2023-10-12
CLDR-16727 Remove short variant for Gulf Standard Time for en_GB

DraganBesevic:
CLDR-16727

Removing the GTS value for short variant of Gulf Standard Time for en_GB. This is a quick fix, while the real fix should be disabling this in Survey Tool

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_GB.xml
+++ b/common/main/en_GB.xml
@@ -5504,9 +5504,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 				<long>
 					<standard>↑↑↑</standard>
 				</long>
-				<short>
-					<standard>GTS</standard>
-				</short>
 			</metazone>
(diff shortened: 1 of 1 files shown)


AEApple (approved):

pedberg-icu:
We have entries for `<metazone type="Gulf">/<short>` in 3 other en_* locales:
en_AU: "Gulf ST"
en_CA: "N/A"
en_IN: "GST"

The en_CA value, at least, seems bogus and should also be deleted.

AEApple:
In that case should we scan en_CA to see if they put in any other bogus N/A values? If they did it in one place, they may have done it elsewhere...

srl295 (approved):

PR #3344, opened 2023-10-17
CLDR-16369 Move VoteStatus enum from VettingViewer to VoteResolver

btangmu:
-VettingViewer should depend on VoteResolver, not vice-versa

CLDR-16369

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -435,6 +434,40 @@ public boolean canDeleteUsers() {
...
+         * The value for the path is either contributed or approved, and the user's organization
+         * chose the winning value. (see class def for null user)
+         */
+        ok,
+
+        /**
+         * The user's organization chose the winning value for the path, but that value is neither
+         * contributed nor approved. (see class def for null user)
+         */
+        provisionalOrWorse,
```
btangmu, line 455: The comment is wrong -- there is no implication whatsoever about the user's organization

This will be fixed in the next PR

srl295 (approved):

PR #3361, opened 2023-10-24
CLDR-17080 Add Regional Variants label for the Info Panel menu

btangmu:
-New component InfoRegionalVariants.vue; related refactoring, modernization

-Fix bug, erase deferred/placeholder help when click Next/Previous

-Remove dead code

CLDR-17080

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -18,6 +18,7 @@ import * as cldrVue from "./cldrVue.mjs";
 
 import InfoPanel from "../views/InfoPanel.vue";
 import InfoSelectedItem from "../views/InfoSelectedItem.vue";
+import InfoRegionalVariants from "../views/InfoRegionalVariants.vue";
 
 let containerId = null;
 let neighborId = null;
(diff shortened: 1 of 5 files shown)


btangmu:
This PR essentially completes the ticket, however I'd like to follow up with some further improvements:

- Set SIDEWAYS_DEBUG to false
- Remove the 2-second delay if (and only if) the user has just chosen from the menu -- this may be difficult/awkward given current dependence on `cldrLoad.reloadV()`
- Further refactoring/clean-up; `setMenuFromData()` is over 100 lines long
- Ideally modernize the client-server code, use API, avoid deprecated `cldrLoad.myLoad()`

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -93,7 +103,10 @@ function insertLegacyElement(containerEl) {
   appendDiv(el, PLACEHOLDER_HELP_ID);
   appendDiv(el, INFO_MESSAGE_ID);
   appendDiv(el, SELECTED_ITEM_ID);
-  appendDiv(el, INFO_REMAINDER_ID);
+  appendDiv(el, INFO_VOTE_TICKET_ID);
+  appendDiv(el, INFO_REGIONAL_ID);
+  appendDiv(el, INFO_FORUM_ID);
+  appendDiv(el, INFO_XPATH_ID);
 }
```
srl295, line 110: ideally we should try to start moving these into InfoPanel.vue

srl295: but to add,  good to add this subcomponent into a vue component

srl295 (approved):

PR #3362, opened 2023-10-25
CLDR-17195 v44: update en with locales now at basic

srl295:
CLDR-17195

- also fixes a test case which ought to have failed on this case.  Confirmed that there are no other cases besides these (`blo`,`xnr`, `kxv`)

ALLOW_MANY_COMMITS=true


stenshamn:
I'm no good with Java so I'll pass on the review.

DraganBesevic (approved):

PR #3368, opened 2023-10-27
CLDR-17202 kbd: add Bengali (bn) keyboard

srl295:
CLDR-17202

ALLOW_MANY_COMMITS=true


miloush:
Where are the key ids from? They seem to be in some sort of transliteration scheme but then there is ch instead of c...

srl295:
> Where are the key ids from? They seem to be in some sort of transliteration scheme but then there is ch, sh, w etc.

wikipedia

miloush:
heh that page is tagged with this issue, also from the Talk page:

> this article needs serious cleanup, it appears to be using some sort of home-grown transliteration scheme at present. 

I would be quite uncomfortable having this scheme, could we use one of the established ones? (ISO15919 if it was up to me, or I could live with ALA-LC too)

srl295:
> heh that page is tagged with this issue, also from the Talk page:
> 
> > this article needs serious cleanup, it appears to be using some sort of home-grown transliteration scheme at present.
> 
> I would be quite uncomfortable having this scheme, could we use one of the established ones? (ISO15919 if it was up to me, or I could live with ALA-LC too)

ok. i might use ala-lc as i have worked with that one.

srl295:
I'll use the ALA-LC table at https://www.loc.gov/catdir/cpso/romanization/bengali.pdf

One of the test failures seems to be a CLDR bug, tracking at https://unicode-org.atlassian.net/browse/CLDR-17204

miloush:
I checked and the only substantial difference between ALA-LC and ISO15919:2001 is sha in ALA-LC and ṣa in ISO-15919.

srl295:
will need to borrow this one from the Assamese table 

`<key id="wa" output="ৱ" /> <!-- Assamese-->`

miloush:
ISO15919 has this one:
image

srl295:
image

Chart so far

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+
+    <keys>
+        <import base="cldr" path="techpreview/keys-Zyyy-punctuation.xml" />
+        <import base="cldr" path="techpreview/keys-Zyyy-currency.xml" />
+
+        <!-- key names based on https://www.loc.gov/catdir/cpso/romanization/bengali.pdf + Assamese for /wa/ -->
+
+        <!-- UNSHIFTED KEYS -->
+        <!-- E: (top) row -->
+        <key id="n̐" output="\u{0981}" /> <!-- candrabindu-->
```
miloush, line 24: could just be id="candrabindu" and avoid comment

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+        <key id="ai" output="\u{09C8}" />
+        <key id="ṛ" output="\u{09C3}" />
+        <key id="tha" output="থ" />
+        <key id="ẏa" output="য়" /> <!-- Missing in Keyman version of file-->
+        <key id="ū" output="\u{09C2}" />
+        <key id="ī" output="\u{09C0}" />
+        <key id="au" output="\u{09CC}" />
+        <key id="pha" output="ফ" />
+
+        <key id="au-length" output="\u{09D7}" /> <!-- TODO: better name? -->
```
miloush, line 88: just wanted to double-check we want encourage people to be entering E+AU LENGTH MARK rather than VOWEL SIGN AU (the spec says length mark exists for compatibility and does not have a meaning on its own)

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+            <row keys="au-length sha dha ṭha gha gap jha kha gap colon double-quote" />
+            <row keys="gap vis-hasant cha ḍha bha ṅa nga open-angle nukta question" />
+            <row keys="space" />
+        </layer>
+    </layers>
+
+    <transforms type="simple">
+        <transformGroup>
+            <!-- these suport the 'more' key on q -->
+            <transform from="\u{q}:" to="\u{0983}" />
```
miloush, line 132: aren't these supposed to be `\m{q}` ?

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+        <key id="jha" output="ঝ" />
+        <key id="kha" output="খ" />
+        <!-- gap -->
+
+        <!-- gap -->
+        <key id="vis-hasant" output="\m{X}" />
+        <key id="cha" output="ছ" />
+        <key id="ḍha" output="ঢ" />
+        <key id="bha" output="ভ" />
+        <key id="ṅa" output="ণ" />
```
miloush, line 103: both ALA-LC and ISO15959 say this is ṇa

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+        <key id="nga" output="ঙ" />
```
miloush, line 104: ... and this is ṅa

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+    <keys>
+        <import base="cldr" path="techpreview/keys-Zyyy-punctuation.xml" />
+        <import base="cldr" path="techpreview/keys-Zyyy-currency.xml" />
+
+        <!-- key names based on https://www.loc.gov/catdir/cpso/romanization/bengali.pdf + Assamese for /wa/ -->
+
+        <!-- UNSHIFTED KEYS -->
+        <!-- E: (top) row -->
+        <key id="n̐" output="\u{0981}" /> <!-- candrabindu-->
+        <key id="1" output="১" />
```
miloush, line 25: So this keyboard cannot output western 0-9 digits?

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+            <transform from="\u{q}ত" to="ৎ" />
+            <transform from="\u{q}য" to="য়" />
+            <transform from="\u{q}র" to="ৰ" />
+            <transform from="\u{q}ল" to="ঌ" />
+            <transform from="\u{q}\u{09BE}" to="অ" />
+            <transform from="\u{q}\u{09BF}" to="ই" />
+            <transform from="\u{q}\u{09C0}" to="ঈ" />
+            <transform from="\u{q}\u{09C1}" to="উ" />
+            <transform from="\u{q}\u{09C2}" to="ঊ" />
+            <transform from="\u{q}\u{09C3}" to="ৠ" />
```
miloush, line 146: This is a long one (RR), are you sure you don't want short one (98B R)?  there is no long vowel sign (9C4)

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,155 @@
...
+            <transform from="\u{q}\u{09C7}" to="এ" />
+            <transform from="\u{q}\u{09C8}" to="ঐ" />
+            <transform from="\u{q}\u{09CB}" to="ও" />
+            <transform from="\u{q}\u{09CC}" to="ঔ" />
+            <transform from="\u{q}\u{09D7}" to="আ" />
```
miloush, line 151: Just to make it clear, the situation is:
<kbd>A</kbd> -> SIGN AA
<kbd>SHIFT</kbd>+<kbd>A</kbd> -> AU MARK

<kbd>Q</kbd>, <kbd>A</kbd> -> LETTER A
<kbd>Q</kbd>, <kbd>SHIFT</kbd>+<kbd>A</kbd> -> LETTER AA

(notably Q and long sign produce short letter unlike Q in combination with everything else)

miloush:
OK I checked the assignments. It's a bit difficult to review though, why would you order the keys in bag in layout order rather than [Brahmic] alphabetical order?

This probably should have a rupee sign.

keyboards/3.0/bn.xml
srl295: This is per romanization table

srl295:
Still todo

- [ ] Tests
- [ ] Add reorder rules

srl295:
> OK I checked the assignments. It's a bit difficult to review though, why would you order the keys in bag in layout order rather than [Brahmic] alphabetical order?
> 

It was for ease of entry. I can sort them now though. 


> 
> 
> This probably should have a rupee sign.

So as with a number of other comments: this is a port of an existing keyboard file (see link in xml). I'm fine with improving it, just noting that step 1 is a port. 

Thanks for the comments and thorough review. I'll keep working on it.

keyboards/3.0/bn.xml
srl295: That's what it came with. Could be on an alt key though

keyboards/3.0/bn.xml
srl295: I wasn't able to verify this one yet.

keyboards/3.0/bn.xml
srl295: thanks good catch

keyboards/3.0/bn.xml
srl295: thanks, not sure where nga came from!

keyboards/3.0/bn.xml
srl295: You're right ,Not sure what that one was doing here. I'll remove it.

keyboards/3.0/bn.xml
srl295: This matches the original keyboard at https://help.keyman.com/keyboard/sil_bengali_phonetic/1.2.1/sil_bengali_phonetic - q e = U+098F

keyboards/3.0/bn.xml
srl295: This does match the [upstream](https://help.keyman.com/keyboard/sil_bengali_phonetic/1.2.1/sil_bengali_phonetic)  
image

srl295:
image

here's a screenshot of a Windows system typing using this keyboard.

srl295:
@miloush what do you think, will this work as sample data?

Manishearth (approved):

srl295:
rebased and should build clean. @miloush or anyone can I get an approval on it?

keyboards/3.0/bn.xml
```diff
@@ -0,0 +1,153 @@
...
+            <row keys="gap ña ai ṛ tha ẏa ū ī au pha open-curly close-curly pipe" />
+            <row keys="au-lenghtener sha dha ṭha gha gap jha kha gap colon double-quote" />
+            <row keys="gap vis-hasant cha ḍha bha ṇa ṅa open-angle nukta question" />
+            <row keys="space" />
+        </layer>
+    </layers>
+
+    <transforms type="simple">
+        <transformGroup>
+            <transform from="\u{09C7}\m{A}" to="\u{09CC}" /> <!-- E + au-lenghtener = AU -->
```
mcdurdin, line 113:
```suggestion
            <transform from="\u{09C7}\m{A}" to="\u{09CC}" /> <!-- E + au-lengthener = AU -->
```

srl295:
- [ ] comment transforms / reorder

srl295:
@miloush and @mhosken i've attempted to document the reorder rules.

srl295:
i'm tracking a build issue causing kbd-check to not work in https://github.com/keymanapp/keyman/issues/10111

miloush:
Did we say the processing turns everything into NFD? Are the reordering matches done before or after that? Or are the match expressions also normalized? Because 9CB O/9CC AU which are under right side vowels decompose into left+right side vowels, in the latter case the right one being 09D7 AU LENGTH MARK which is not part of the rules.

We don't care about characters not entered through the keyboard (like using Alt+Numpad), correct? (especially thinking about Vedic marks)

srl295:
I'm working on a document for review on normalization https://unicode-org.atlassian.net/browse/CLDR-17192

> Did we say the processing turns everything into NFD?

We _will_ say that the match acts as if it's in NFD.

>  Are the reordering matches done before or after that?

Matches will be in NFD

> Or are the match expressions also normalized?

and the expression will be normalized also.

>  Because 9CB O/9CC AU which are under right side vowels decompose into left+right side vowels, in the latter case the right one being 09D7 AU LENGTH MARK which is not part of the rules.

I think I, earlier, removed 09D7. But I should re-add it for this reason.

> 
> We don't care about characters not entered through the keyboard (like using Alt+Numpad), correct? (especially thinking about Vedic marks)

srl295:
@miloush PTAL, U+09D7 is now included in the reorders

srl295:
@miloush OK to merge?

miloush (approved):

srl295:
@miloush thanks!

PR #3370, opened 2023-10-27
CLDR-17205 kbd: fix charts problem

srl295:
- charts broke with recent renames (such as to= to output=)

CLDR-17205

Note: against v44 for now.

ALLOW_MANY_COMMITS=true


srl295:
see image on #3368 for example of chart generation

miloush (approved):

PR #3371, opened 2023-10-28
CLDR-13970 Update ExampleDependencies and related code

btangmu:
-Bug fix: when generating example dependencies, disable caching in ICUServiceBuilder to prevent some dependencies from being missed

-New cldr.util.GenerateExampleDependencies replaces old cldr.test.TestExampleDependencies

-Include DO NOT EDIT comments in ExampleDependencies.java, clarify it is generated

-Bug fix: prevent NullPointerException in ExampleGenerator.getOtherGender/getOtherCase

-Bug fix: OpenAPI validation CWWKO1650E missing description for 404 in UserAPI.java (unrelated)

-Remove dead code, fix warnings

-Comments

CLDR-13970

ALLOW_MANY_COMMITS=true


btangmu:
This fixes problems with generating the dependencies, where the example for one path depends on the values for other paths. Some dependencies were formerly missed due to caching. The resulting file ExampleDependencies.java is now complete (as far as I can tell), and it can be kept up to date by running GenerateExampleDependencies when necessary (probably just before each time Survey Tool is opened for submission), as follows:

`java -DCLDR_DIR=$(pwd) -jar tools/cldr-code/target/cldr-code.jar generate-example-dependencies`

Unfortunately, there are still failures to update examples in the running Survey Tool. I suspect this is still due to caching, so that even though examples are regenerated, they're regenerated using outdated cached values -- not cached examples, but cached values (for names of months, etc.) in ICUServiceBuilder and/or elsewhere. I've added a new boolean constant ISB_CAN_CLEAR_CACHE for testing this hypothesis. So far, it looks like even when testing with ISB_CAN_CLEAR_CACHE true (which probably goes too far and has a bad effect on performance), the examples are still not regenerated correctly.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -224,10 +224,7 @@ private static Date getDate(int year, int month, int date, int hour, int minute,
 
     public void setCachingEnabled(boolean enabled) {
         exCache.setCachingEnabled(enabled);
-    }
-
-    public void setCacheOnly(boolean cacheOnly) {
-        exCache.setCacheOnly(cacheOnly);
+        icuServiceBuilder.setCachingEnabled(enabled);
```
btangmu, line 227: This is the most essential change, used with enabled = false only by GenerateExampleDependencies, not while ST is running

btangmu:
GenerateExampleDependencies now takes about an hour, where it used to take about 20 minutes.

The resulting file is about twice as large, and it would about 4 times as large were it not for using putAll instead of put.

btangmu:
We should add a BRS task for GenerateExampleDependencies, at least once before each time ST is opened for submission.

srl295 (approved):

PR #3372, opened 2023-10-31
CLDR-17080 Regional Variants menu follow-up

btangmu:
-Set SIDEWAYS_DEBUG to false

-Remove the 2-second delay if (and only if) the user has just chosen from the menu

-Shorten excessively long functions by moving nested functions to top level and making more subroutines

-Remove dependency of the Vue component on non-GUi modules cldrLoad, cldrStatus

-Refactor for clarity

-Comments

CLDR-17080

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrSideways.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSideways.mjs
@@ -8,10 +8,10 @@ import * as cldrLoad from "./cldrLoad.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
 
-const SIDEWAYS_DEBUG = true;
+const SIDEWAYS_DEBUG = false;
 
-const escape = "\u00A0\u00A0\u00A0"; // non-breaking spaces
-const unequalSign = "\u2260\u00A0"; // U+2260 = "≠"
+const NON_BREAKING_SPACES = "\u00A0\u00A0\u00A0"; // non-breaking spaces
+const UNEQUALS_SIGN = "\u2260\u00A0"; // U+2260 = "≠"
 
 /**
  * Array storing all only-1 sublocales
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/js/src/esm/cldrSideways.mjs
```diff
@@ -121,115 +138,131 @@ function setMenuFromData(regionalVariantsWrapper, json, cacheKey) {
...
-          dataList.push([json.novalue[s], differentValue]);
-        }
-      }
-      mergeReadBase(dataList);
+  /*
+   * Force the use of unequalSign in the regional comparison pop-up for locales in
+   * json.novalue, by assigning a value that's different from curValue.
+   */
+  if (json.novalue) {
+    const differentValue = curValue === "A" ? "B" : "A"; // anything different from curValue
```
srl295, line 161:
```suggestion
    const differentValue = NaN;
```

It'll never be === any other string.   Or it could even be `const differentvalue = Symbol("no value")`;

tools/cldr-apps/js/src/esm/cldrSideways.mjs
```diff
@@ -121,115 +138,131 @@ function setMenuFromData(regionalVariantsWrapper, json, cacheKey) {
...
+    }
+  }
+  return readLocale;
+}
+
+function appendLocaleList(list, curValue, topLocale, readLocale) {
+  const popupSelect = {
+    items: [],
+    label:
+      "Regional Variants for " + locmap.getRegionAndOrVariantName(topLocale),
```
srl295, line 240: ideally `cldrText`

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrSideways.mjs
```diff
@@ -8,27 +8,48 @@ import * as cldrLoad from "./cldrLoad.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
 
-const SIDEWAYS_DEBUG = true;
+const SIDEWAYS_DEBUG = false;
```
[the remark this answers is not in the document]
btangmu: that's worth thinking about...

PR #3373, opened 2023-10-31
CLDR-17220 SBRS v45 merge selected maint/maint-44 fixes back to main

pedberg-icu:
CLDR-17220

This merges relevant fixes on the maint/maint-44 branch back to main. Notes:
- The maint/maint-44 branch was created on Oct 11 after commit e58b346f62643dd546a1cb5d263d3664d1c93514 of Oct 10
- This merge includes everything on the maint/maint-44 branch after the above commit, up through commit fa9259fb6da482c81fed466a1c6e6b31b3a81d28 of Oct 30, except for the specific items listed in the next sections.
- The following commits on maint/maint-44 were not included because they were cherry-picks to maint/maint-44 of fixes that had already been made on main:
     - Commit 32a184260b9979a1d1c4334ac5f6a16eeb57aa61 Oct 11, 2 keyboard spec fixes
     - Commit fc7cf41746e7b581a875b3a5b6bffb1657cce32d Oct 11, 1 keyboard spec fix
- The following commits on maint/maint-44 were not included because they were specifically about updating various items to final status. On the main branch, we want to keep non-final status, but will update the version numbers for 45 under a separate PR.
    - Commit c291c054c1647fa95ec940658e8e087ac8a7c130 Oct 25, remove -SNAPSHOT in poms, updated README to final
    - Commit 6e3f841e36fb70cf3bb8382698a27b7c1fce40f3 Oct 26, update spec status to final (but there are a few other specific fixes here that will be rolled in when I do the version number update)
    - Commit 19406a15004e3e39d37f3b46ad98bcbe657cfc18 Oct 27, use final ICU4J libs

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartLocaleGrowth.java
```diff
@@ -355,7 +456,7 @@ private static Map<String, FoundAndTotal> addGrowth(
             }
             // END HACK
 
-            if (showMissing) {
+            if (false && showMissing) {
```
btangmu, line 459: dead code like this generally shouldn't be kept unless accompanied by an explanatory comment, preferably a TODO comment with an open ticket number, otherwise it leads to unmaintainable code

btangmu (approved):
LGTM though I'm unfamiliar with most of this code

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartLocaleGrowth.java
pedberg-icu: Agreed, but fixing those kinds of things probably does not belong in this kind of PR which is just merging maint branch changes back to main as is.

PR #3385, opened 2023-11-07
CLDR-16838 Login bug; preliminary refactoring/debugging

btangmu:
-Add a temporary debugging boolean HARD_CODE_REDIRECT_PORT to enable testing locally with nginx on port 8888

-Delete dead code and fix compiler warnings in WebContext.java, SurveyMain.java, and related files

-Remove the Deprecated label for the phase BETA since we still use it

-Delete the other deprecated phases and dependent code

-Add warning comments for methods that appear unused but are still accessed by jsp

-Comments

CLDR-16838

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
@@ -72,13 +72,10 @@ import org.unicode.cldr.util.CLDRURLS;
 import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.CoverageInfo;
 import org.unicode.cldr.util.Factory;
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -145,30 +143,14 @@ public enum Phase {
...
+
         /**
-         * Do not use.
-         *
-         * @deprecated
+         * Survey Tool is in Beta mode. Votes, announcements, etc., will be stored in special
+         * database tables with the "_beta" suffix, and will be used only for testing.
          */
-        @Deprecated
         BETA("Beta", CheckCLDR.Phase.SUBMISSION);
```
btangmu, line 154: We actually used BETA at the beginning of v44

PR #3386, opened 2023-11-07
CLDR-16838 Login bug; preliminary refactoring, clean-up; phase BETA not deprecated

btangmu:
-Delete dead code and fix compiler warnings in WebContext.java, SurveyMain.java, and related files

-Remove the Deprecated label for the phase BETA since we still use it

-Delete the other deprecated phases and dependent code

-Add warning comments for methods that appear unused but are still accessed by jsp

-Comments

CLDR-16838

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
@@ -72,13 +72,10 @@ import org.unicode.cldr.util.CLDRURLS;
 import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.CoverageInfo;
 import org.unicode.cldr.util.Factory;
-import org.unicode.cldr.util.Factory.DirectoryType;
-import org.unicode.cldr.util.Factory.SourceTreeType;
 import org.unicode.cldr.util.LDMLUtilities;
 import org.unicode.cldr.util.LocaleNormalizer;
 import org.unicode.cldr.util.LocaleSet;
 import org.unicode.cldr.util.Organization;
-import org.unicode.cldr.util.Pair;
 import org.unicode.cldr.util.PathHeader;
 import org.unicode.cldr.util.PathHeader.PageId;
 import org.unicode.cldr.util.PathHeader.SurveyToolStatus;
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -145,30 +143,14 @@ public enum Phase {
...
+
         /**
-         * Do not use.
-         *
-         * @deprecated
+         * Survey Tool is in Beta mode. Votes, announcements, etc., will be stored in special
+         * database tables with the "_beta" suffix, and will be used only for testing.
          */
-        @Deprecated
         BETA("Beta", CheckCLDR.Phase.SUBMISSION);
```
btangmu, line 154: I started to delete BETA along with the other deprecated items, then noticed that we did use BETA at the start of v44, and there's a lot of dependent code on both the front and back ends. Unlike the other deprecated items, which have caused a lot of confusion due to their overlapping/redundant meanings, BETA seems like it may still be useful and isn't likely to be used accidentally. It shouldn't be deprecated unless we actually intend to get remove it.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
```diff
@@ -1538,43 +1405,43 @@ public void setSession() {
...
-                        "We are swamped with about "
-                                + CookieSession.getUserCount()
-                                + " people using the SurveyTool right now! Please try back in a little while.");
-                return;
-            }
-        } else if (session == null || (session.user == null)) { // guest user
-            if (CookieSession.tooManyObservers()) {
-                if (session != null) {
+        // allow in administrator or TC.
+        if (!UserRegistry.userIsTC(user)) {
```
btangmu, line 1409: this is an auto-refactor by IntelliJ, to fix a warning about empty "if" block -- it comes out looking more complicated than it really is

srl295: makes sense.   Though, It'd be easier to read if the refactor was a separate PR or at least a separate commit.

srl295: Oh, this IS the refactor PR.  I reread it.  It looks like it has some wording cleanup too, which is fine.  LGTM!

srl295 (approved):

PR #3387, opened 2023-11-08
Update en.xml to add apc (Levantine Arabic)

a455bcd9:
Adding apc Levantine Arabic, surprisingly not present even though it was included in the 43 release ( https://cldr.unicode.org/index/downloads/cldr-43 ).

CLDR-_____

ALLOW_MANY_COMMITS=true


srl295:
@a455bcd9 please file a ticket unicode-org.atlassian.net 

Process wise, https://unicode-org.atlassian.net/browse/CLDR-15663 should cause this to be sorted out at least later in the cycle.

PR #3388, opened 2023-11-14
CLDR-16838 Fix login bug: removeLoginCookies if query has email and password

btangmu:
-Undeprecate UserRegistry.internalGetPassword, accessed not only by JSP but also by Auth.java

-Fix some compiler warnings for Auth.java

CLDR-16838

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
@@ -1330,10 +1330,18 @@ public class WebContext implements Cloneable, Appendable {
             if (jwt != null && !jwt.isBlank()) {
                 final String jwtId = CookieSession.sm.klm.getSubject(jwt);
                 if (jwtId != null && !jwtId.isBlank()) {
-                    User jwtInfo = CookieSession.sm.reg.getInfo(Integer.parseInt(jwtId));
-                    if (jwtInfo != null) {
-                        user = jwtInfo;
-                        logger.fine("Logged in " + jwtInfo + " #" + jwtId + " using JWT");
+                    if (!email.isEmpty() && !password.isEmpty()) {
+                        // If the user was already logged in as Admin/TC/Manager, then used a URL
+                        // with explicit email/password to log in as a different user, the old
+                        // cookies (especially JWT) must be removed to prevent staying logged
+                        // in as the first user
+                        removeLoginCookies(request, response);
+                    } else {
+                        User jwtInfo = CookieSession.sm.reg.getInfo(Integer.parseInt(jwtId));
+                        if (jwtInfo != null) {
+                            user = jwtInfo;
+                            logger.fine("Logged in " + jwtInfo + " #" + jwtId + " using JWT");
+                        }
                     }
                 }
             }
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserRegistry.java
@@ -380,8 +380,7 @@ public class UserRegistry {
             return password;
         }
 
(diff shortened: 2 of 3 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
```diff
@@ -1328,12 +1328,17 @@ public void setSession() {
...
-                if (jwtId != null && !jwtId.isBlank()) {
-                    User jwtInfo = CookieSession.sm.reg.getInfo(Integer.parseInt(jwtId));
-                    if (jwtInfo != null) {
-                        user = jwtInfo;
-                        logger.fine("Logged in " + jwtInfo + " #" + jwtId + " using JWT");
+                KeepLoggedInManager klm = CookieSession.sm.klm;
+                final String jwtId = klm.getSubject(jwt);
+                if (jwtId != null && !jwtId.isBlank() && !klm.jwtIsInExcludedSet(jwtId)) {
+                    if (!email.isEmpty() && !password.isEmpty()) {
+                        klm.addToExcludedSet(jwtId);
```
srl295, line 1335: What's the reason ?

btangmu: Otherwise, if I'm logged in as Admin, then I click a link to log in as another user, it doesn't work -- I'm still logged in as Admin, not as the other user. That's the bug that needs fixing. This PR fixes the bug by excluding the Admin JWT so it no longer prevents switching to a different user.

btangmu: The exclusion only happens in the special case where email and password are included in URL, as in https://cldr-staging.unicode.org/cldr-apps/survey?email=...&pw=...

srl295: ok.  so it's not the excluded set, it's the loggedOutSet because you are wanting to ignore this JWT from now on. 

i'd suggest renaming it or at least commenting that.

Also, at this point, you might call `WebContext.removeLoginCookies()`  or even `.logout()` so that the JWT doesn't come back.

srl295: since the excluded set isn't persistent, the JWT will cause the user to be logged back in next time ST restarts. Not sure if that's desireable or not.   Clearing the login cookies might actually be better in that regard.  then there's no need for an exclude set then.

btangmu: "the JWT will cause the user to be logged back in next time ST restarts" -- that was already the case, though, right? And in the typical case where the excluded/logged-out JWT is for Admin/TC/Manager, maybe that's appropriate.

"WebContext.removeLoginCookies() or even .logout() so that the JWT doesn't come back" -- OK, I'll try that

btangmu: removeLoginCookies works!

srl295: > "the JWT will cause the user to be logged back in next time ST restarts" -- that was already the case, though, right? And in the typical case where the excluded/logged-out JWT is for Admin/TC/Manager, maybe that's appropriate.

The odd part is that the behavior will change _depending on whether ST restarted_.

If you do whatever you do after clicking the password link, and then close the window, and come back an hour later to st.unicode.org: 
- if ST has restarted, you'll be logged in as the Admin/TC/Manager
- if ST hasn't restarted, you'll be logged in as the user you clicked the password link for

removeLoginCookies makes it consistent. If you want to login, you need to login again.

srl295 (approved):
LGTM!

PR #3389, opened 2023-11-15
CLDR-16838 Follow-up, new optional CLDR_REDIRECT_PORT in cldr.properties

btangmu:
-Enable running with nginx on port other than default 80, flexibility for testing on local machine

-No change in behavior if CLDR_REDIRECT_PORT not specified

-Property is retrieved only once, then cached for performance

CLDR-16838

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
@@ -28,6 +28,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 import org.unicode.cldr.test.DisplayAndInputProcessor;
+import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.Level;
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
```diff
@@ -719,14 +720,39 @@ void print(Throwable t) {
      */
     void redirect(String where) {
         try {
-            response.sendRedirect(where);
+            String port = getRedirectPort();
+            if (port != null) {
+                String url =
+                        request.getScheme() + "://" + request.getServerName() + ":" + port + where;
```
srl295, line 726: how do you know what the scheme is? 

couldn't this all just pull from `CLDR_SURVEY_URL` in the settings?

srl295 (approved):
LGTM if needed for debugging

tools/cldr-apps/src/main/java/org/unicode/cldr/web/WebContext.java
btangmu: maybe -- I didn't try that; I figured, like the original code, use the scheme and server name from the existing request, and just fix the port since that's what was broken when using 8888 for nginx

PR #3394, opened 2023-11-27
CLDR-17233 Have DAIP coalesce multiple spaces to strictest; fix in fr.xml, reorder zones in root

pedberg-icu:
CLDR-17233

Fixed DAIP to coalesce multiple spaces of different types to the strictest type (NNBSP if present, else NBSP if present).
Fixed the specific problem in fr.xml short units that prompted the ticket.
Ran CLDRModify passes to test; this also rearranged some time zone names in root.xml.

This is on the maint/maint-44 branch; will need to be merged into the main branch (possibly along with other CLDR 44.1 fixes) under a separate ticket.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -2907,6 +2907,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="America/St_Johns">
 				<exemplarCity>St. John’s</exemplarCity>
 			</zone>
+			<zone type="America/Curacao">
+				<exemplarCity>Curaçao</exemplarCity>
+			</zone>
 			<zone type="Africa/Asmera">
 				<exemplarCity>Asmara</exemplarCity>
 			</zone>
@@ -2955,8 +2958,17 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="Asia/Katmandu">
 				<exemplarCity>Kathmandu</exemplarCity>
 			</zone>
+			<zone type="America/Asuncion">
+				<exemplarCity>Asunción</exemplarCity>
+			</zone>
+			<zone type="Indian/Reunion">
+				<exemplarCity>Réunion</exemplarCity>
+			</zone>
 			<zone type="Atlantic/St_Helena">
 				<exemplarCity>St. Helena</exemplarCity>
+			</zone>
+			<zone type="Africa/Sao_Tome">
+				<exemplarCity>São Tomé</exemplarCity>
 			</zone>
 			<zone type="America/Lower_Princes">
 				<exemplarCity>Lower Prince’s Quarter</exemplarCity>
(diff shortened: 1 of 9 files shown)


macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -1289,6 +1297,12 @@ private String normalizeWhitespace(String path, String value) {
             value = PLACEHOLDER_SPACE_BEFORE.matcher(value).replaceAll("\u00A0{");
         }
 
+        // Finally, replace remaining space combinations with most restrictive type CLDR-17233
+        // If we have NNBSP U+202F in combination with other spaces, keep just it
+        value = NNBSP_AND_OTHER_SPACES.matcher(value).replaceAll("\u202F");
+        // Else if we have NBSP U+00A0 in combination with other spaces, keep just it
+        value = NBSP_AND_SPACES.matcher(value).replaceAll("\u00A0");
```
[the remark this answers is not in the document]
macchiati: Thanks

common/main/root.xml
```diff
@@ -2955,9 +2958,18 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="Asia/Katmandu">
 				<exemplarCity>Kathmandu</exemplarCity>
 			</zone>
+			<zone type="America/Asuncion">
```
macchiati, line 2961: Seems quite odd that these moved. Nothing to do with this PR, however.

btangmu (dismissed):
LGTM, though ideally the unit tests should be updated -- TestWhitespaceAdjustments and/or TestWhitespaceNormalization in TestDisplayAndInputProcessor.java -- in such a way that a test would fail if the new code weren't there or if it ever gets accidentally disabled in the future

common/main/root.xml
pedberg-icu: It is because I ran CLDRModify as part of this PR. I guess it had not been done after these zones were added.

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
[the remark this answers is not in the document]
pedberg-icu: @macchiati This also now handles thin space among other spaces. And I made the handling of other spaces more generic (using \h instead of an explicit list). Also added a unit test.

pedberg-icu:
> LGTM, though ideally the unit tests should be updated -- TestWhitespaceAdjustments and/or TestWhitespaceNormalization in TestDisplayAndInputProcessor.java -- in such a way that a test would fail if the new code weren't there or if it ever gets accidentally disabled in the future

@btangmu Thanks, I have now added unit tests.

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -1289,6 +1299,14 @@ private String normalizeWhitespace(String path, String value) {
...
         }
 
+        // Finally, replace remaining space combinations with most restrictive type CLDR-17233
+        // If we have NNBSP U+202F in combination with other spaces, keep just it
+        value = NNBSP_AMONG_OTHER_SPACES.matcher(value).replaceAll("\u202F");
+        // Else if we have NBSP U+00A0 in combination with other spaces, keep just it
+        value = NBSP_AMONG_OTHER_SPACES.matcher(value).replaceAll("\u00A0");
+        // Else if we have THIN SPACE U+2009 in combination with other spaces, keep just it
+        value = THIN_SPACE_AMONG_OTHER_SPACES.matcher(value).replaceAll("\u2009");
+
```
macchiati, line 1309: Hmmm. One more suggestion: if we have all ASCII space, collapse to one. 

Or is that done elsewhere?

macchiati (approved):
Catching more cases now!

Approving now in case we already catch the all ASCII spaces.

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
pedberg-icu: That appears to be already handled elsewhere...

macchiati:
Thanks!

PR #3407, opened 2023-12-05
CLDR-17248 Measure vote timing; remove obsolete debugging code

btangmu:
-Delete CLDR_TABLE_DEBUG_ZOOM and the block of code that depended on it

-Do not duplicate display of xpath in info panel that depended on CLDR_TABLE_DEBUG

CLDR-17248

ALLOW_MANY_COMMITS=true


btangmu:
The duplicate xpath is shown in this screenshot:

image

If CLDR_TABLE_DEBUG was true, the xpath was shown twice in the info panel if the user clicked on the leftmost column ("code cell"). This feature doesn't seem very useful since the xpath is already shown at the bottom right corner. It seems better for CLDR_TABLE_DEBUG not to have visible effects except in the console. Also it depended on the complex legacy function cldrInfo.listen, whose "str" arg is otherwise usually null or empty, so this simplification may help pave the way to cleaning up and modernizing that code.

btangmu:
I just noticed my branch has the wrong ticket number in it: t13970_d should have been 17248_d. I won't waste time fixing that, if that's OK

srl295 (approved):

PR #3410, opened 2023-12-06
CLDR-17230 add right-facing derived names and annotations

macchiati:
CLDR-17230

Had to make some fixes to the tooling; edge cases caused problems.

Data files are from running:
- GenerateDerivedAnnotations
- CLDRModify -fQ -s../../common/annotationsDerived
- CLDRModify -fp -s../../common/annotationsDerived
- CLDRModify -r -s../../common/annotationsDerived
- and will run CLDRFileTransformer

ALLOW_MANY_COMMITS=true


macchiati:
Hmmm. I ran Maven spotless, but it didn't reformat the files as it should have. Sigh.

macchiati:
@srl295 After running Maven formatter from the Unicode tools, whenever I run Maven spotless it doesn't run on the CLDR java files anymore, it just does unicodetools. Is there a setting I can use to fix that?

btangmu:
@macchiati I generally run this in Terminal in the cldr folder:

`mvn --file=tools/pom.xml spotless:apply`

Is that what you've tried?

srl295:
@macchiati i renamed the PR to note v44

srl295:
@macchiati please review and merge https://github.com/macchiati/cldr/pull/84 and spotless will be fixed

docs/ldml/tr35.md
```diff
@@ -2,10 +2,10 @@
 
 # Unicode Locale Data Markup Language (LDML)
 
-|Version|44        |
+|Version|44.1        |
 |-------|----------|
 |Editors|Mark Davis (<a href="mailto:markdavis@google.com">markdavis@google.com</a>) and <a href="tr35.md#Acknowledgments">other CLDR committee members</a>|
-|Date|2023-10-25|
+|Date|2023-12-05|
 |This Version|<a href="https://www.unicode.org/reports/tr35/tr35-70/tr35.html">https://www.unicode.org/reports/tr35/tr35-70/tr35.html</a>|
```
pedberg-icu, line 9: Need to update this to 71

pedberg-icu (changes requested):
Looks like there are problems if there is no annotation for the base. Also need to updated spec numerical sequence version.

macchiati:
Thanks, Peter. Will work on those two issues, at 11

common/annotationsDerived/es_MX.xml
```diff
@@ -1156,11 +1156,29 @@ Derived short names and annotations, using GenerateDerivedAnnotations.java. See
...
+		<annotation cp="🧑🏾‍🦯‍➡">Persona con bastón blanco | Persona con bastón blanco hacia la derecha</annotation>
+		<annotation cp="🧑🏾‍🦯‍➡" type="tts">Persona con bastón blanco hacia la derecha</annotation>
+		<annotation cp="🧑🏿‍🦯‍➡">Persona con bastón blanco | Persona con bastón blanco hacia la derecha</annotation>
+		<annotation cp="🧑🏿‍🦯‍➡" type="tts">Persona con bastón blanco hacia la derecha</annotation>
 		<annotation cp="🧑🏻‍🦽">persona en silla de ruedas manual | tono de piel claro</annotation>
 		<annotation cp="🧑🏼‍🦽">persona en silla de ruedas manual | tono de piel claro medio</annotation>
 		<annotation cp="🧑🏽‍🦽">persona en silla de ruedas manual | tono de piel medio</annotation>
 		<annotation cp="🧑🏾‍🦽">persona en silla de ruedas manual | tono de piel oscuro medio</annotation>
 		<annotation cp="🧑🏿‍🦽">persona en silla de ruedas manual | tono de piel oscuro</annotation>
+		<annotation cp="🧑‍🦽‍➡">persona en silla de ruedas manual</annotation>
```
pedberg-icu, line 1176: Seem to be missing the "facing right" phrase here...

common/annotationsDerived/es_419.xml
```diff
@@ -662,11 +662,23 @@ Derived short names and annotations, using GenerateDerivedAnnotations.java. See
...
+		<annotation cp="🚶🏼‍➡">caminar | caminata | persona | persona caminando</annotation>
+		<annotation cp="🚶🏽‍➡">caminar | caminata | persona | persona caminando</annotation>
+		<annotation cp="🚶🏾‍➡">caminar | caminata | persona | persona caminando</annotation>
+		<annotation cp="🚶🏿‍➡">caminar | caminata | persona | persona caminando</annotation>
 		<annotation cp="🏃🏻">correr | deporte | deportista | maratón | persona | persona corriendo | tono de piel claro</annotation>
 		<annotation cp="🏃🏼">correr | deporte | deportista | maratón | persona | persona corriendo | tono de piel claro medio</annotation>
 		<annotation cp="🏃🏽">correr | deporte | deportista | maratón | persona | persona corriendo | tono de piel medio</annotation>
 		<annotation cp="🏃🏾">correr | deporte | deportista | maratón | persona | persona corriendo | tono de piel oscuro medio</annotation>
 		<annotation cp="🏃🏿">correr | deporte | deportista | maratón | persona | persona corriendo | tono de piel oscuro</annotation>
+		<annotation cp="🏃‍➡">correr | deporte | deportista | maratón | persona | persona corriendo</annotation>
```
pedberg-icu, line 676: Missing the "facing right phrase here"

common/annotationsDerived/en_GB.xml
```diff
@@ -59,5 +59,11 @@ Derived short names and annotations, using GenerateDerivedAnnotations.java. See
 		<annotation cp="🧑🏽‍🦼">accessibility | medium skin tone | person in motorised wheelchair | person in powered wheelchair | wheelchair</annotation>
 		<annotation cp="🧑🏾‍🦼">accessibility | medium-dark skin tone | person in motorised wheelchair | person in powered wheelchair | wheelchair</annotation>
 		<annotation cp="🧑🏿‍🦼">accessibility | dark skin tone | person in motorised wheelchair | person in powered wheelchair | wheelchair</annotation>
+		<annotation cp="🧑‍🦼‍➡">accessibility | person in motorised wheelchair | person in powered wheelchair | wheelchair</annotation>
```
pedberg-icu, line 62: Missing the "facing right" phrase

pedberg-icu:
The added annotations in some child locales are missing the "facing right" phrase.

macchiati:
I realized what the problem is. 

When we run CldrModify -fQ, it adds the name to the keywords if there is a name. However, CldrModify only looks at the annotations in that file — not resolved. In the case of en_GB and some others, only a small subset of items are actually in the annotations file explicitly, the ones that are needed to override the parent (eg for 'person in motorised wheelchair' to get British spelling).

That processing for CLDRModify normally works, but in non-top locales it fails.

So what we need to do is if we get either an explicit keywords or explicit name for an emoji in a non-top locale, we need to do something.

macchiati:
I think I've fixed the child locale problem in the latest commit.

pedberg-icu (approved):
Looks good, thanks, this seems to have fixed some other existing issues as well!

PR #3420, opened 2023-12-16
CLDR-17288 vagrant/ansible: fix for openliberty / mysql

srl295:
CLDR-17288

- this is just to get the build going again, CLDR-17279 will be for fixing docs etc.
- had to update the Ubuntu version (which we need in CLDR-16735) because of build issues, but should be forward/backward compatible.

Discussion: https://github.com/orgs/community/discussions/73013

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/XPathTable.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/XPathTable.java
@@ -140,7 +140,6 @@ public class XPathTable {
             sql = null;
             s.close();
             s = null;
-            ourConn.commit();
         } finally {
             DBUtils.close(s);
             if (sql != null) {
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -36,3 +36,10 @@ jobs:
       run: (cd tools/scripts/ansible && ansible-galaxy install -r requirements.yml && ln -sv test-local-vars local-vars && vagrant validate)
     - name: Try the provision
       run: (cd tools/scripts/ansible && vagrant up)
+      env:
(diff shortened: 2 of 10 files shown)


btangmu (dismissed):

srl295:
@btangmu I'm wondering if a better way to go about this would be to actually _provision and setup an azure machine_ as part of the testing. then we could run tests against it, and then shut it down.

btangmu:
"actually provision and setup an azure machine as part of the testing" -- do you mean to integrate github with azure?

srl295:
@btangmu 'testprovision' now succeeds and prints:

> `ERROR this usually fails but we are ignoring the failure. Try it manually`

Please review and then we can manually try `vagrant up` locally.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/XPathTable.java
```diff
@@ -140,7 +140,6 @@ private void setupDB(Connection ourConn) throws SQLException {
             sql = null;
             s.close();
             s = null;
-            ourConn.commit();
```
btangmu, line 143: when called from SurveyMain, ourConn = getAConnection

but when called from TestXPathTable (twice) and TestSTFactory, ourConn = getDBConnection

seems like a bug either way; the callers need to be consistent?

btangmu (changes requested):
LGTM except ourConn.commit() removal -- some callers use getDBConnection not getAConnection

tools/cldr-apps/src/main/java/org/unicode/cldr/web/XPathTable.java
srl295: fixed

btangmu (approved):

PR #3421, opened 2023-12-19
CLDR-17249 be perf

srl295:
- don't use user settings for voting/getrow queries
- on voting, reuse the test bundle instead of recreating it

CLDR-17249

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -1355,7 +1355,8 @@ public class DataPage {
             CookieSession session,
             CLDRLocale locale,
             String prefix,
-            XPathMatcher matcher) {
+            XPathMatcher matcher,
+            TestResultBundle checkCldr) {
 
         SurveyMain sm =
                 CookieSession
@@ -1379,8 +1380,9 @@ public class DataPage {
             throw new InternalError("?!! ourSrc hsa no supplemental dir!");
         }
         synchronized (session) {
-            TestResultBundle checkCldr =
-                    sm.getSTFactory().getTestResult(locale, getOptions(session, locale));
+            if (checkCldr == null) {
+                checkCldr = sm.getSTFactory().getTestResult(locale, getSimpleOptions(locale));
+            }
             if (checkCldr == null) {
                 throw new InternalError("checkCldr == null");
             }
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
```diff
@@ -1419,10 +1421,24 @@ public static CheckCLDR.Options getOptions(CookieSession session, CLDRLocale loc
...
+        return new Options(locale, SurveyMain.getTestPhase(), defaultLevel, org);
+    }
+
+    /** Get options, but don't try to check user preferences */
+    public static CheckCLDR.Options getSimpleOptions(CLDRLocale locale) {
+        return new Options(
+                locale,
+                SurveyMain.getTestPhase(),
+                Level.COMPREHENSIVE.name(), /* localeType is not used?! */
+                "NOT USED");
```
btangmu, line 1439: CoverageLevel_localeType has "set" but no "get", and "localeType" doesn't occur in any jsp files

cruft? delete?

btangmu: "don't use user settings for voting/getrow queries" -- how do we know the user settings aren't accessed?

for that matter, how do we know whether the user settings are accessed when getOptions is used instead of getSimpleOptions?

srl295: > cruft

seems like it.  How does CheckCLDR use the 'level' setting? Maybe @macchiati can comment on it.

> how do we know the user settings aren't accessed

This PR stops calling into user settings for GET/POST on a row, and uses `COMPREHENSIVE` instead of checking the level.

> for that matter, how do we know whether the user settings are accessed when getOptions is used instead of getSimpleOptions?

`getListSetting()` calls out to the DB to check the user's setting.

```
    public static CheckCLDR.Options getOptions(CookieSession session, CLDRLocale locale) {
        CheckCLDR.Options options;
        final String def =
                CookieSession.sm.getListSetting(
                        session.settings(), SurveyMain.PREF_COVLEV, WebContext.PREF_COVLEV_LIST);

        final String org = session.getEffectiveCoverageLevel(locale.toString());
```

srl295: > CoverageLevel_localeType has "set" but no "get", and "localeType" doesn't occur in any jsp files
> 
> cruft? delete?

Want to split that out to a separate PR from the performance stuff?

btangmu (approved):

PR #3422, opened 2023-12-19
CLDR-17289 make setCldrFileToCheck a lazy load

srl295:
CLDR-17289

Update for performance, especially when calling CLDR Checks on a single XPath.

- rename implementation functions `setCldrFileToCheck()` to `handleSetCldrFileToCheck()`  which is a deferred call.
- New internal subclass API, `accept(…)` which checks must call once they determined they accept the xpath
- New internal subclass override `reset()` which checks can use to reset their state.
- New internal subclass override `handleCheckPossibleErrors()` for checks that want to provide immediate possibleErrors.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
@@ -1421,7 +1421,7 @@ public abstract class CheckCLDR implements CheckAccessor {
                                     "Test setup time for " + item.getClass().toString() + ": {0}");
                 try {
                     item.setPhase(getPhase());
-                    item.handleSetCldrFileToCheck(cldrFileToCheck, options, possibleErrors);
+                    item.setCldrFileToCheck(cldrFileToCheck, options, possibleErrors);
                     if (SHOW_TIMES) {
                         if (item.isSkipTest()) {
                             System.out.println("Disabled : " + testTime);
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
@@ -650,25 +650,6 @@ public abstract class CheckCLDR implements CheckAccessor {
     }
 
     /**
-     * Don't override this, use the other setCldrFileToCheck which takes an Options instead of a
-     * Map<>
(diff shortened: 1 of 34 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
```diff
@@ -697,6 +678,11 @@ public CheckCLDR setCldrFileToCheck(
         return this;
     }
 
+    public CheckCLDR setCldrFileToCheck(
```
macchiati, line 681: I assume this is a work in progress...

srl295: yes, as a draft

macchiati:
I think the basic idea was something like the following. Is that right?

- CheckCLDR.setCldrFileToCheck will just store the CLDRFile and options, and add a flag for initialized=false;
- in the subclasses, setCldrFileToCheck in the subclasses will be renamed to init(), and instead of using direct parameters, will use the stored parameters. At the end of being called, it will set initialized=true;
- in each handleCheck, the following code will be added after the check for paths of interest, like:
if (!initialized) {
  init();
}

macchiati:
In some cases, we might want to add additional checks in handleCheck before the init call; that would be a further optimization.

srl295:
> In some cases, we might want to add additional checks in handleCheck before the init call; that would be a further optimization.

yeah, I think I will split out a handleCheckPossibleErrors() just for that use. Some tests such as those that check the locale ID really want to perform a check right away, non-deferred.

srl295:
> In some cases, we might want to add additional checks in handleCheck before the init call; that would be a further optimization.

i think you may mean additional _skip opportunities_, i noted some in comments

btangmu:
"Error: Value always overridden in children: [uk_UA]" -- CheckChildren -- that's a puzzle

srl295:
> "Error: Value always overridden in children: [uk_UA]" -- CheckChildren -- that's a puzzle

I think the handleSetCldrFileToCheck isn't getting called properly

btangmu (approved):

srl295:
I'm getting this one in. I'd like to perf test it against a single-XPath use case, but we can test it in the larger context until then.

PR #3424, opened 2023-12-23
CLDR-17248 Enable creation of ST accounts for webdriver simulated users

btangmu:
-Simulated users have pseudo email addresses like driver-123@cldr-apps-webdriver.org

-New class AuthSurveyDriver

-In Auth, if login fails, call AuthSurveyDriver to enable creation of account if appropriate

-Only create simulated users if SurveyMain.isUnofficial, never on production server

-Only create simulated users if CLDR_WEBDRIVER_PASSWORD is defined in cldr.properties

CLDR-17248

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Auth.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Auth.java
@@ -18,6 +18,7 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+import org.unicode.cldr.web.AuthSurveyDriver;
 import org.unicode.cldr.web.CookieSession;
 import org.unicode.cldr.web.SurveyLog;
 import org.unicode.cldr.web.SurveyMain;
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Auth.java
```diff
@@ -82,10 +83,17 @@ public Response login(
...
-                        CookieSession.sm.reg.get(request.password, request.email, userIP);
+                UserRegistry.User user;
+                try {
+                    user = CookieSession.sm.reg.get(request.password, request.email, userIP);
+                } catch (LogoutException e) {
+                    user = null;
+                }
+                if (user == null) {
+                    user = AuthSurveyDriver.createTestUser(request.password, request.email);
+                }
```
btangmu, line 94: this is a little awkward since UserRegistry.get sometimes throws LogoutException and sometimes just returns null without throwing an exception -- currently it throws LogoutException for a new webdriver user, but I didn't want to rely on that here, so this also handles the null return possibility

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Auth.java
```diff
@@ -82,10 +83,17 @@ public Response login(
...
                 if (user == null) {
-                    return Response.status(403, "Login failed").build();
+                    throw new LogoutException();
```
btangmu, line 96: there's already a catch(LogoutException) below which returns `Response.status(403, "Login failed").build()`

if we returned that response directly here, the catch below would trigger an "unused" warning, and removing it would result in a many lines of code changing their indentation, and this change would look much larger than it really is

btangmu:
This goes together with https://github.com/unicode-org/cldr-apps-webdriver/pull/31

macchiati (approved):

PR #3426, opened 2023-12-28
CLDR-17270 datadog integration

srl295:
CLDR-17270

- add datadog integration
- incidentally fix CLDR-14843 (though not deployed)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrNotify.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrNotify.mjs
@@ -16,6 +16,10 @@
 
 // Reference: https://www.antdv.com/components/notification
 import { notification } from "ant-design-vue";
+import { datadogLogs } from "@datadog/browser-logs";
+
+/** if falsy, let sleeping dogs lie */
+const hasDataDog = !!window.dataDogClientToken;
 
 /**
  * For warnings and general notifications, automatically close after this many seconds
@@ -48,6 +52,9 @@ function open(message, description) {
  * @param {String} description the more detailed description
  */
 function warning(message, description) {
+  if (hasDataDog) {
+    datadogLogs.logger.warn(message, { description });
+  }
   notification.warning({
     message: message,
     description: description,
@@ -62,6 +69,9 @@ function warning(message, description) {
  * @param {String} description the more detailed description
  */
 function error(message, description) {
+  if (hasDataDog) {
+    datadogLogs.logger.error(message, { description });
+  }
   notification.error({
     message: message,
     description: description,
@@ -73,6 +83,9 @@ function error(message, description) {
  * Display an error notification, and when the user closes it, call the callback function
  */
 function errorWithCallback(message, description, callback) {
+  if (hasDataDog) {
+    datadogLogs.logger.error(message, { description });
+  }
   notification.error({
     message: message,
     description: description,
@@ -93,6 +106,9 @@ function exception(e, context) {
       name: "",
       message: e,
     };
+  }
+  if (hasDataDog) {
+    datadogLogs.logger.error(context, {}, e);
   }
   notification.error({
     message: "Internal error: " + e.name + " " + context,
--- a/tools/cldr-apps/js/src/index.js
+++ b/tools/cldr-apps/js/src/index.js
@@ -12,24 +12,23 @@ import "ant-design-vue/dist/antd.min.css";
 import * as cldrGui from "./esm/cldrGui.mjs";
 import * as cldrVue from "./esm/cldrVue.mjs";
 
-import { datadogRum } from '@datadog/browser-rum';
-import { datadogLogs } from '@datadog/browser-logs';
+import { datadogRum } from "@datadog/browser-rum";
+import { datadogLogs } from "@datadog/browser-logs";
 
+if (window.dataDogClientToken) {
+  datadogLogs.init({
+    clientToken: window.dataDogClientToken,
+    site: "us5.datadoghq.com",
+    forwardErrorsToLogs: true,
+    sessionSampleRate: 100,
+  });
 
-datadogLogs.init({
-  clientToken: 'pub0596cc2553803e176da42558924f0318',
-  site: 'us5.datadoghq.com',
-  forwardErrorsToLogs: true,
-  sessionSampleRate: 100
-});
-
-
-datadogRum.init({
-    applicationId: 'c58441d9-21fd-4195-b958-9840483630fb',
-    clientToken: 'pub0596cc2553803e176da42558924f0318',
-    site: 'us5.datadoghq.com',
-    service: 'surveytool',
-    env: 'staging',
+  datadogRum.init({
+    applicationId: window.dataDogAppId,
+    clientToken: window.dataDogClientToken,
+    site: "us5.datadoghq.com",
+    service: "surveytool",
+    env: window.dataDogEnv,
     // Specify a version number to identify the deployed version of your application in Datadog
     // version: '1.0.0',
     sessionSampleRate: 100,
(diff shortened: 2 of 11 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -220,6 +220,28 @@ private void includeCss(HttpServletRequest request, PrintWriter out) {
...
+                                    + "window.dataDogClientToken='%s';\n"
+                                    + "window.dataDogAppId='%s';\n"
+                                    + "window.dataDogEnv='%s';\n"
+                                    + "window.dataDogSha='%s';\n"
+                                    + "</script>\n",
+                            DD_CLIENT_TOKEN, DD_CLIENT_APPID, DD_ENV, DD_GIT_COMMIT_SHA));
+        } else {
+            out.write("<script>window.dataDogClientToken='';</script>\n");
+        }
+    }
```
srl295, line 243: @btangmu this might look familiar. Here, I need to get *runtime* config information (from the actual staging server), so the webpack process won't work here.

tools/scripts/ansible/templates/56-surveytool.j2
```diff
@@ -3,3 +3,4 @@
 {{ cldr_surveytool_user }} ALL=(root) NOPASSWD: /usr/sbin/service {{ cldr_openliberty_service }} stop
 {{ cldr_surveytool_user }} ALL=(root) NOPASSWD: /usr/sbin/service {{ cldr_openliberty_service }} restart
 {{ cldr_surveytool_user }} ALL=(root) NOPASSWD: /usr/sbin/service {{ cldr_openliberty_service }} status
+{{ cldr_surveytool_user }} ALL=(root) NOPASSWD: /usr/bin/rm -rf /srv/st/config/.cache
```
srl295, line 6: this fixes CLDR-14843

tools/scripts/ansible/templates/deploy-sh.j2
```diff
@@ -73,7 +75,7 @@ sudo -u root /usr/sbin/service ${SERVICE} stop
 if [[ -d /srv/st/config/.cache ]];
 then
    echo "Deleting cache /srv/config/.cache"
-   rm -rf /srv/st/config/.cache
+   sudo /usr/bin/rm -rf /srv/st/config/.cache
```
srl295, line 78: CLDR-14843

tools/scripts/ansible/templates/deploy-sh.j2
```diff
@@ -95,4 +97,6 @@ fi
 # exclude these two files
 rsync -r --exclude server.env --exclude workarea --exclude bootstrap.properties -v \
       --delete ./wlp/usr/servers/cldr/* /var/lib/openliberty/usr/servers/cldr/ || exit 1
+# update datadog
+sed -i.old -e "s/DD_GIT_COMMIT_SHA.*/DD_GIT_COMMIT_SHA=${GITHUB_SHA}/g" -e "s%DD_GIT_REPOSITORY_URL.*%DD_GIT_REPOSITORY_URL=github.com/${GITHUB_REPO}%g" /var/lib/openliberty/usr/servers/{{ cldr_openliberty_server }}/server.env
```
srl295, line 101: updates these. ignored if not configured.

tools/cldr-apps/js/src/esm/cldrNotify.mjs
```diff
@@ -48,6 +52,9 @@ function open(message, description) {
  * @param {String} description the more detailed description
  */
 function warning(message, description) {
+  if (hasDataDog) {
+    datadogLogs.logger.warn(message, { description });
+  }
```
srl295, line 57: is it going to be too noisy to ship all of these?

srl295:
@btangmu draft but ready to merge soon…

srl295:
@btangmu i think this is ready to go. If the properties aren't set, it doesn't make a difference on the client or server.

tools/cldr-apps/js/src/esm/cldrNotify.mjs
btangmu: I guess we'll find out -- hasDataDog will be false on production until/unless we enable it after testing on cldr-staging, right?

tools/cldr-apps/js/src/index.js
```diff
@@ -12,6 +12,39 @@ import "ant-design-vue/dist/antd.min.css";
...
+    defaultPrivacyLevel: "allow",
+    allowedTracingUrls: [
+      {
+        match: /https:\/\/.*\.unicode\.org/,
+        propagatorTypes: ["tracecontext"],
+      },
+    ],
+  });
+}
+
```
btangmu, line 47: Ideally this code should go in its own file instead of index.js, maybe cldrMonitor.mjs, for encapsulation

index.js should stay short and high level

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
btangmu: right -- still it's a drag to have more js in html in java -- ideally the front end should get the data from the back end as json

could we add a request for that json in cldrRun.mjs instead of having this in the initial html?

tools/scripts/ansible/templates/deploy-sh.j2
```diff
@@ -4,9 +4,11 @@
 if [[ ${1} = "--repo" ]];
 then
     shift
-    GIT_PREFETCH=https://github.com/$(echo -n "${1}" | tr -c 'A-Za-z0-9_.-/' '-').git
+    GITHUB_REPO=$(echo -n "${1}" | tr -c 'A-Za-z0-9_.-/' '-')
+    GIT_PREFETCH=https://github.com/${GITHUB_REPO}.git
     shift
 else
+    GITHUB_REPO=unicode-org/cldr
```
btangmu, line 11: ideally avoid "github" in name, in case we switch to different provider -- e.g., SOURCE_REPO

tools/scripts/ansible/templates/server-env.j2
```diff
@@ -5,3 +5,7 @@ MYSQL_PASSWORD={{ mysql_users[0].password }}
 MYSQL_DB=cldrdb
 WLP_SKIP_MAXPERMSIZE=true
 LOG_DIR=/var/log/openliberty/cldr
+
+# DataDog params. Will be updated by the deploy script.
+DD_GIT_COMMIT_SHA=0000000
+DD_GIT_REPOSITORY_URL=github.com/unicode-org/cldr
```
btangmu, line 11: I don't quite understand the duplication here and in deploy-sh.j2

btangmu (approved):

tools/scripts/ansible/templates/server-env.j2
btangmu: ideally "MONITOR" instead of "DD" for encapsulation, we might eventually have different/multiple monitoring tools using the same config

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
srl295: That adds another round trip.

tools/scripts/ansible/templates/deploy-sh.j2
srl295: Except that this is actually a github repo, and this is called from github. If we changed we would need to redo this script. The parameter comes in as USER/REPO from github.

tools/scripts/ansible/templates/server-env.j2
srl295: This parameter is also consumed by the datadog agent in this form using these names. It's a "URL" but doesn't include the scheme (https://). 
The duplication is because this file needs an initial value, which the deploy script updates as the comment notes.

tools/cldr-apps/js/src/esm/cldrNotify.mjs
srl295: correct

srl295:
merged, but i'll make a followup for refactoring

PR #3442, opened 2024-01-08
CLDR-17295 tz2023d data updates and cleaned up deprecated zones/metazone

yumaoka:
CLDR-17295

ALLOW_MANY_COMMITS=true


yumaoka:
@macchiati I found we did not update tzdata files in CLDR tools recently. 2023 was a quiet year for TZ database, and I guess we probably did not need to touch metazone. In CLDR 44, we deprecated some CLDR canonical time zones to match IANA zone.tab. Because we use canonical zone IDs as keys for zone display names and exemplar city names, locale data for deprecated canonical zone ID are no longer used. These names (e.g. exemplar city name for America/Yellowknife) are now just redundant. I noticed this when I run one of unit test case after importing the latest zone.tab to CLDR code base. So this PR contains removal of locale data for obsolete canonical zone IDs.

yumaoka:
@sffc Related to previous comment, TZ database folks concluded `America/Santa_Isabel` was not really required. CLDR created metazone `Mexico_Northwest`, but it turned out we really don't need this, because `America/Santa_Isabel` is in same zone with `America/Tijuana`. `America/Santa_Isabel` is removed from zone.tab a while ago. So metazone `Mexico_Northwest` was just introduced by the misunderstanding and at this point, it should not be used at all.

I removed the metazone `Mexico_Northwest` from metaZone.xml including the short metazone ID mapping in this PR.
```
<metazoneId shortId="meno" longId="Mexico_Northwest" />
```
Could you please check the removal of metazone short ID has no bad impacts in ICU4X implementation?

macchiati (approved):

sffc:
I believe there is no harm in adding or removing these IDs. It may mean that there are display names that are not reachable with mixed data sources.

PR #3444, opened 2024-01-08
CLDR-17088 Flesh out en.xml; new test for null/blank/inheritance-marker

btangmu:
-New TestEnInheritance.java

CLDR-17088

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
@@ -1,0 +1,46 @@
+package org.unicode.cldr.test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.logging.Logger;
+import org.junit.jupiter.api.Test;
+import org.unicode.cldr.util.*;
+
+/**
+ * Check modern coverage inherited values in en that are marked with up arrows or are blank; all
+ * values should be explicit in en
+ */
+public class TestEnInheritance {
+    private static final Logger logger = Logger.getLogger(TestEnInheritance.class.getName());
+
+    private static final String LOCALE_ID = "en";
+
+    @Test
+    void testEn() {
+        final CLDRConfig config = CLDRConfig.getInstance();
+        final CLDRFile cldrFile = config.getCLDRFile(LOCALE_ID, true);
+        final SupplementalDataInfo sdi = SupplementalDataInfo.getInstance();
+        final CoverageLevel2 coverageLevel = CoverageLevel2.getInstance(sdi, LOCALE_ID);
+        int pathsWithNullValue = 0, pathsWithBlankValue = 0, pathsWithMarkerValue = 0;
+        for (final String path : cldrFile.fullIterable()) {
+            if (coverageLevel.getLevel(path).getLevel() <= Level.MODERN.getLevel()) {
+                String value = cldrFile.getStringValue(path);
+                if (value == null) {
+                    complain("null value", ++pathsWithNullValue, path);
+                } else if (value.isBlank()) {
+                    complain("blank value", ++pathsWithBlankValue, path);
+                } else if (CldrUtility.INHERITANCE_MARKER.equals(value)) {
+                    complain("inheritance marker", ++pathsWithMarkerValue, path);
+                }
+            }
+        }
+        assertEquals(0, pathsWithNullValue, "null values");
+        assertEquals(0, pathsWithBlankValue, "blank values");
+        assertEquals(0, pathsWithMarkerValue, "inheritance marker values");
+    }
+
+    private void complain(String description, int count, String path) {
+        logger.severe(
+                this.getClass().getSimpleName() + " -- " + description + " " + count + " " + path);
+    }
+}


tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class TestEnInheritance {
+    private static final Logger logger = Logger.getLogger(TestEnInheritance.class.getName());
+
+    private static final String LOCALE_ID = "en";
+
+    @Test
+    void testEn() {
+        final CLDRConfig config = CLDRConfig.getInstance();
+        final CLDRFile cldrFile = config.getCLDRFile(LOCALE_ID, true);
```
macchiati, line 21: This is getting the resolved value, so the results will not be accurate.

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
```diff
@@ -0,0 +1,46 @@
...
+        final SupplementalDataInfo sdi = SupplementalDataInfo.getInstance();
+        final CoverageLevel2 coverageLevel = CoverageLevel2.getInstance(sdi, LOCALE_ID);
+        int pathsWithNullValue = 0, pathsWithBlankValue = 0, pathsWithMarkerValue = 0;
+        for (final String path : cldrFile.fullIterable()) {
+            if (coverageLevel.getLevel(path).getLevel() <= Level.MODERN.getLevel()) {
+                String value = cldrFile.getStringValue(path);
+                if (value == null) {
+                    complain("null value", ++pathsWithNullValue, path);
+                } else if (value.isBlank()) {
```
macchiati, line 30: There are a few values that can be blank, so they would need to be excluded. I can't remember if these are marked in the DTD or not; if not, we might consider that so that they can be tested with DtdData.

macchiati (changes requested):

btangmu:
Locally this results in these errors, and I expect it to have the same errors here:

```
[INFO] Running org.unicode.cldr.test.TestEnInheritance
Jan 08, 2024 1:07:39 PM org.unicode.cldr.test.TestEnInheritance complain
SEVERE: TestEnInheritance -- blank value 1 //ldml/personNames/nativeSpaceReplacement
Jan 08, 2024 1:07:39 PM org.unicode.cldr.test.TestEnInheritance complain
SEVERE: TestEnInheritance -- blank value 2 //ldml/personNames/foreignSpaceReplacement
Jan 08, 2024 1:07:39 PM org.unicode.cldr.test.TestEnInheritance complain
SEVERE: TestEnInheritance -- null value 1 //ldml/dates/timeZoneNames/metazone[@type="Gulf"]/short/standard
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.085 s <<< FAILURE! - in org.unicode.cldr.test.TestEnInheritance
[ERROR] org.unicode.cldr.test.TestEnInheritance.testEn  Time elapsed: 0.084 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: null values ==> expected: <0> but was: <1>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
```

The ticket specifies "blank" values as problematic, but maybe that's not to be taken literally, especially for nativeSpaceReplacement and foreignSpaceReplacement?

Also there's a null value for one path, the Gulf time zone.

There are no occurrences of INHERITANCE_MARKER (in Modern-coverage paths).

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
btangmu: nativeSpaceReplacement and foreignSpaceReplacement and ...?

btangmu: The second commit adds exclusions for nativeSpaceReplacement and foreignSpaceReplacement based on their constants DisplayAndInputProcessor.FSR_START_PATH/NSR_START_PATH -- not sure how we'd use the DTD...?

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
btangmu: Second commit changes resolving from true to false

btangmu:
Second commit running locally gets no more "blank" errors, still gets the Gulf time zone "null" error, and gets 155 new "null" errors for currencies including `//ldml/numbers/currencies/currency[@type="BND"]/symbol`

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEnInheritance.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class TestEnInheritance {
+    private static final Logger logger = Logger.getLogger(TestEnInheritance.class.getName());
+
+    private static final String LOCALE_ID = "en";
+
+    @Test
+    void testEn() {
+        final CLDRConfig config = CLDRConfig.getInstance();
+        final CLDRFile cldrFile = config.getCLDRFile(LOCALE_ID, false);
```
macchiati, line 21: Also, you need to check on this. We need to hit not only the main directory, but also the annotations directory. I think Factory getCommonAndSeedAndMainAndAnnotationsFactory() is the safest.

macchiati:
Good. I'm a bit surprised we don't get more of the other items in CODE_FALLBACK besides currency, but that's good news.

btangmu:
See https://github.com/unicode-org/cldr/pull/4110

PR #3446, opened 2024-01-09
CLDR-17109 st: Actually remove the rest of Derby

srl295:
remove derbyNotSupported(), db_Derby, AND db_Mysql AND getDBKind()

CLDR-17109

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
@@ -78,30 +78,25 @@ public class DBUtils {
     }
 
     // === DB workarounds
-    public static String DB_SQL_IDENTITY = "GENERATED ALWAYS AS IDENTITY";
-    public static String DB_SQL_VARCHARXPATH = "varchar(1024)";
-    public static String DB_SQL_WITHDEFAULT = "WITH DEFAULT";
-    public static String DB_SQL_TIMESTAMP0 = "TIMESTAMP";
-    public static String DB_SQL_CURRENT_TIMESTAMP0 = "CURRENT_TIMESTAMP";
-    public static String DB_SQL_MIDTEXT = "VARCHAR(1024)";
-    public static String DB_SQL_BIGTEXT = "VARCHAR(16384)";
-    public static String DB_SQL_UNICODE = "VARCHAR(16384)"; // unicode type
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
```diff
@@ -79,49 +68,35 @@ public static final void derbyNotSupported() {
...
+    public static final String DB_SQL_ENGINE_INNO = "ENGINE=InnoDB";
+    public static final String DB_SQL_MB4 = " CHARACTER SET utf8mb4 COLLATE utf8mb4_bin";
     public static int db_number_open = 0;
     public static int db_number_used = 0;
-    private static int db_UnicodeType = java.sql.Types.VARCHAR; /*
-                                                                * for setNull -
-                                                                * see
-                                                                * java.sql.Types
-                                                                */
+    private static final int db_UnicodeType = java.sql.Types.BLOB;
```
btangmu, line 99: On first glance this seems to go beyond dropping derby -- does this potentially change the behavior when using mysql? Now I see setupSqlForServerType had override for mysql... so I'm assuming there's no actual change for mysql

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
```diff
@@ -378,38 +310,19 @@ public boolean tableExists(String table) {
...
-                                canonColumn)
-                        > 0;
-            }
-        } catch (SQLException se) {
-            SurveyMain.busted("While looking for column '" + table + "." + column + "': ", se);
-            return false; // NOTREACHED
-        }
+        return sqlCount(
+                        conn,
+                        "select count(*) from information_schema.COLUMNS where table_name=? and column_name=?",
```
btangmu, line 322: ideally all caps for the sql reserved words -- out of scope for this ticket though

btangmu (approved):
Hurray!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DBUtils.java
srl295: Yes it moves the overrides up. I'm trying to make it easier to see what the MySQL settings are.

PR #3447, opened 2024-01-10
CLDR-17304 capture internalError check stack traces into the system log

srl295:
CLDR-17304

- before shipping off to front end, call system logger on any Subtype.internalError messages
- confirmed that they show up in the liberty logfile. so, should be picked up by datadog

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3450, opened 2024-01-10
CLDR-7646 TestCache concurrency issues

srl295:
- fix the TestCacheBundle and TestCacheResult caches to be concurrent safe
- was using concurrent safe maps, but not computing the results safely.

CLDR-7646

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/TestCache.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/TestCache.java
@@ -2,7 +2,6 @@ package org.unicode.cldr.test;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


srl295:
@btangmu i have this same question:

```java
    /*
     * TODO: document whether CLDR_TESTCACHE_SIZE is set on production server, and if so to what, and why;
     * evaluate why the fallback 12 for CLDR_TESTCACHE_SIZE is appropriate or too small. Consider not
     * using maximumSize() at all, depending on softValues() instead to garbage collect only when needed.
     */
```

srl295:
This PR seems to cause the dashboard concurrency errors to no longer occur.

macchiati (commented):
Question: is there a reasonable test that we can add to detect problems such as these?

srl295:
> Question: is there a reasonable test that we can add to detect problems such as these?

- #3447 makes sure that these internal error stack traces (such as concurrency and other issues) would get logged, so DataDog / server console would pick it up.
- ConsoleCheck does call TestCache in a threaded way, perhaps not in the right order.
- SurveyTool has typically run in a single threaded environment, so calls into this would not be contested.
- That said, we could add a threaded TestCache stress test, that could be a good idea. This change now calls each CheckCLDR within a single thread though, but can read the cache from multiple threads.

macchiati (approved):

btangmu:
> @btangmu i have this same question:
> 
> ```java
>     /*
>      * TODO: document whether CLDR_TESTCACHE_SIZE is set on production server, and if so to what, and why;
>      * evaluate why the fallback 12 for CLDR_TESTCACHE_SIZE is appropriate or too small. Consider not
>      * using maximumSize() at all, depending on softValues() instead to garbage collect only when needed.
>      */
> ```
> 
> it's NOT set in production. So we are only caching 12 values at best.

@srl295 @macchiati Per git blame I added that comment 2019-10-23! In the meantime I've come to believe that TODO comments should always include ticket numbers, and tickets shouldn't be closed until their TODOs are all done or moved to other open tickets. Anyway, it looks like CLDR_TESTCACHE_SIZE limits the number of TestResultBundle objects that are cached, and each TestResultBundle is for a specific locale (and coverage level, etc). So, max 12 locales have cached test results at any point in time. It's plausible that there are rarely vetters at work on more than 12 locales simultaneously. However, performance may suffer whenever a multiple-locale operation such as Priority Items Summary is run. What should we do? Documentation says "in most circumstances it is better to set a per-cache maximum size instead of using soft references" -- but our code is already using BOTH max size and soft ref. It's tempting to get rid of the max since it's so arbitrary, and trust (hope? verify?) that soft refs will give good results on their own.

srl295:
@btangmu you put some important analysis that i missed - this is the number of _locales_ not the number of _xpaths_.  12 locales is definitely going to be beneficial, 12 xpaths might not be worth doing. 

I think I remember something about how soft refs (or was that weak refs?) might not get evicted in a cpu- and io- starved situation, (such as recalculating priority items summary). So there is some risk there.  I wouldn't object to removing the limit, if it was tested with priority items summary etc.

btangmu:
I've made a new ticket about CLDR_TESTCACHE_SIZE: https://unicode-org.atlassian.net/browse/CLDR-17308

PR #3452, opened 2024-01-15
CLDR-17282 Add currency attributes tz, to-tz and docs; use for currency changes from 2022-01-01

pedberg-icu:
CLDR-17282

1. For the `currency` element, add attributes `tz` and `to-tz`. Document these in the spec, along with note that implementations may want to convert the combination of date time and timezone to a UTC timestamp or equivalent for use as runtime data.
    * I made these `@VALUE` since they are not distinguishing attributes (but rather annotations for the `to` and `from` attributes); however they are not really like element values either. I also updated PathHeader accordingly.

2. Update the supplemental currency data to use these new attributes for all currency transitions from 2022-01-01 onward. There were only two, for Croatia using tz Europe/Zagreb, and for Sierra Leone, using tz Africa/Freetown. Both of those zones are currently aliases in the backward file, but are still valid in CLDR I believe.

3. I have filed [ICU-22631](ttps://unicode-org.atlassian.net/browse/ICU-22631) for the associated work in the cldr-icu converter.

ALLOW_MANY_COMMITS=true

[ICU-22631]: https://unicode-org.atlassian.net/browse/ICU-22631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -75,12 +75,18 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!-- use from and to instead. -->
     <!--@VALUE-->
     <!--@DEPRECATED-->
+<!ATTLIST currency iso4217 NMTOKEN #REQUIRED >
+    <!--@MATCH:validity/currency-->
 <!ATTLIST currency from NMTOKEN #IMPLIED >
     <!--@MATCH:time/yyyy-MM-dd-->
 <!ATTLIST currency to NMTOKEN #IMPLIED >
     <!--@MATCH:time/yyyy-MM-dd-->
-<!ATTLIST currency iso4217 NMTOKEN #REQUIRED >
-    <!--@MATCH:validity/currency-->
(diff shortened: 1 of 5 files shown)


srl295 (approved):
LGTM. Only comment is if 'IANA timezone' should actually be CLDR timezone, including aliases and such.

pedberg-icu:
> LGTM. Only comment is if 'IANA timezone' should actually be CLDR timezone, including aliases and such.

Thanks, I will investigate and file another PR against the same ticket if it needs changing.

pedberg-icu:
> > LGTM. Only comment is if 'IANA timezone' should actually be CLDR timezone, including aliases and such.
> 
> Thanks, I will investigate and file another PR against the same ticket if it needs changing.

@srl295 OK, PR [#3468](https://github.com/unicode-org/cldr/pull/3468) changes the timezone ID specification to specifically reference the CLDR canonical “long” time zone ID.

PR #3453, opened 2024-01-16
CLDR-2671 add Latgalian language

Mazurs:
CLDR-2671

Probably not ready to merge. Based on Latvian locale.

ALLOW_MANY_COMMITS=true


srl295:
@Mazurs thanks - I'm closing this because it has too much data- the data should be input via the surveytool.  We now have your form submission on the ticket so we can put this in for v46. Let's discuss on the ticket.

PR #3454, opened 2024-01-17
CLDR-17313 st: only build supplemental factory once

srl295:
the supplemental disk factory should be a singleton. also. use the Suppliers.memoize() idiom in more places.

Because it's a singleton, `getSupplementalDiskFactory()` is now cheap to call.

Note: ✅ Sanity tested locally.

CLDR-17313

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
@@ -6,6 +6,7 @@
  */
 package org.unicode.cldr.web;
 
+import com.google.common.base.Suppliers;
 import com.ibm.icu.dev.util.ElapsedTimer;
 import com.ibm.icu.lang.UCharacter;
 import com.ibm.icu.text.ListFormatter;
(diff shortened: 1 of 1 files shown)


macchiati (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -2314,41 +2316,43 @@ private void doMain(WebContext ctx) {
...
+    public final Factory getDiskFactory() {
+        return gFactory.get();
+    }
+
+    private Supplier<Factory> gFactory =
+            Suppliers.memoize(
+                    () -> {
+                        final File[] list = getFileBases();
+                        CLDRConfig config = CLDRConfig.getInstance();
+                        // may fail at server startup time- should do this through setup mode
```
btangmu, line 2341: "may fail at server startup time" -- if so, what will happen?

srl295: It'll actually explain to the user that the survey tool isn't configured yet and invite hem to Edit cldr.properties.

btangmu: so it would only be expected to fail if cldr.properties isn't properly configured yet? if so, the comment could be less alarming if it spelled that out...

PR #3458, opened 2024-01-21
CLDR-7646 fix for TestCache issue

srl295:
- was initializing the cache incorrectly

CLDR-7646 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/TestCache.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/TestCache.java
@@ -2,6 +2,8 @@ package org.unicode.cldr.test;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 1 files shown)


macchiati (approved):

srl295:
@macchiati sorry, missed a spotless run

tools/cldr-code/src/main/java/org/unicode/cldr/test/TestCache.java
```diff
@@ -110,7 +118,7 @@ public List<CheckStatus> getPossibleProblems() {
     public TestResultBundle getBundle(final CheckCLDR.Options options) {
         TestResultBundle b;
         try {
-            b = testResultCache.get(options, () -> new TestResultBundle(options));
```
srl295, line 113: this was the wrong `get()` function to use as it doesn't pass the key.

macchiati (approved):

PR #3460, opened 2024-01-22
CLDR-17304 quell some test errs

srl295:
- add a logKnownIssue on CLDR-17320
- add a return to CheckCLDR.setLoggerLevel() so we can do save/restore level

CLDR-17304

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
@@ -67,9 +67,17 @@ public abstract class CheckCLDR implements CheckAccessor {
     /** protected so subclasses can use it */
     protected static Logger logger = Logger.getLogger(CheckCLDR.class.getSimpleName());
 
-    /** set the internal logger level. For ConsoleCheck. */
-    public static void setLoggerLevel(java.util.logging.Level newLevel) {
+    /**
+     * set the internal logger level. For ConsoleCheck.
+     *
+     * @returns the previous level
+     */
+    public static java.util.logging.Level setLoggerLevel(java.util.logging.Level newLevel) {
+        // NB: we use the full package name here, to avoid conflict with other CLDR classes named
+        // Level
+        java.util.logging.Level oldLevel = logger.getLevel();
         logger.setLevel(newLevel);
+        return oldLevel;
     }
 
     /** serialize CheckCLDR as just its class name */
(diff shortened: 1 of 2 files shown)


btangmu (commented):
It's not yet clear to me how this fits in with the big picture. It seems like whenever an unexpected exception occurs during a unit test, the test should fail; otherwise a passing test is essentially meaningless and only gives a false sense of security. If tests have been passing in spite of throwing NullPointerException, etc., and we weren't aware of that until now due to the logging level masking what we see in the console, maybe one lesson to be learned is that we need more visibility, not less, in which case this PR seems to go in the wrong direction...

srl295:
> It's not yet clear to me how this fits in with the big picture. It seems like whenever an unexpected exception occurs during a unit test, the test should fail; otherwise a passing test is essentially meaningless and only gives a false sense of security. If tests have been passing in spite of throwing NullPointerException, etc., and we weren't aware of that until now due to the logging level masking what we see in the console, maybe one lesson to be learned is that we need more visibility, not less, in which case this PR seems to go in the wrong direction...

This particular test has been passing in spite of NullPointerException. I'm not sure entirely what the test is trying to do, it looks like it purposely includes empty values in some places.

I agree we need more visibility, not less.  This PR doesn't remove the logging, it makes it a known issue to be handled in CLDR-17320 where we should revisit why we're using NullPointerException instead of a better subtype such as "missing exemplar string".  And it leaves several sibling tests that throw about 11 stack traces, just not 1,800 which makes the build log unusable from GitHub.  

Prior to my earlier PR, we weren't aware of the stack trace because it was captured internally, and would be shown to an end user randomly, or shown to developers WHEN data changes caused the stack trace.

macchiati:
Yes, the test is specifically testing against null values; I'm not sure
why, perhaps just to test robustness. The problem is that until recently,
it didn't generate tons of spurious messages on the console.

btangmu (approved):
Per our discussion on slack, I now understand a little better (I think), that: (1) the purpose of TestAllLocales is not to catch real data errors, but to test the code that catches data errors; (2) except in that limited context, this PR doesn't reduce error logging; and (3) this change is only temporary until fixing https://unicode-org.atlassian.net/browse/CLDR-17320

macchiati (approved):

srl295:
> Yes, the test is specifically testing against null values; I'm not sure why, perhaps just to test robustness. The problem is that until recently, it didn't generate tons of spurious messages on the console.

Thanks for confirming. So NullPointerExceptions are (at least were) expected in this test, which was not my assumption going in.

PR #3461, opened 2024-01-23
CLDR-17314 intern across multiple locations

srl295:
- Unit test to verify that CLDRFile paths are interned
- interning in XPathTable and StringId

CLDR-17314

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -3156,8 +3156,8 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         if (extraPaths == null) {
             extraPaths =
                     ImmutableSet.<String>builder()
-                            .addAll(getRawExtraPathsPrivate())
-                            .addAll(CONST_EXTRA_PATHS)
+                            .addAll(CharUtilities.internAll(getRawExtraPathsPrivate()))
+                            .addAll(CharUtilities.internAll(CONST_EXTRA_PATHS))
                             .build();
             if (DEBUG) {
                 System.out.println(getLocaleID() + "\textras: " + extraPaths.size());
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -55,6 +55,7 @@ import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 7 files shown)


srl295:
corrected ticket

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -3156,8 +3156,8 @@ public Set<String> getRawExtraPaths() {
         if (extraPaths == null) {
             extraPaths =
                     ImmutableSet.<String>builder()
-                            .addAll(getRawExtraPathsPrivate())
-                            .addAll(CONST_EXTRA_PATHS)
+                            .addAll(CharUtilities.internAll(getRawExtraPathsPrivate()))
```
macchiati, line 3159: Could these be done with lambdas, eg getREPP().stream().map(x -> x.intern()).collect(...)

tools/cldr-code/src/main/java/org/unicode/cldr/util/StringId.java
```diff
@@ -34,7 +34,7 @@ public final class StringId {
      * @return a value from 0 to 0x7FFFFFFFFFFFFFFFL.
      */
     public static long getId(CharSequence charSequence) {
-        String string = charSequence.toString();
+        String string = charSequence.toString().intern();
```
macchiati, line 37: Is this only used with paths? the problem with over-interning is that you can fill up memory.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
srl295: It's an Iterable<> interface. Was trying to avoid having to allocate another copy of all of the strings. This way it's just pass through. But you could use a lambda on a temporary object

tools/cldr-code/src/main/java/org/unicode/cldr/util/StringId.java
srl295: Good point. Will remove, catching this at the xpath table is probably sufficient

srl295: removed

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
srl295: PTAL

macchiati:
Ah, good point.

macchiati (approved):

btangmu (approved):

btangmu:
"Commit a814c25 is for CLDR-17313, but the PR is for CLDR-17314" -- probably a squash would fix that

PR #3462, opened 2024-01-23
CLDR-17327 speedup VoteResolver's transcript

srl295:
- skip String.format when unneeded
- defer all transcript operations until needed

measured as between a 5 and 10x speedup

CLDR-17327

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DeferredTranscript.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DeferredTranscript.java
@@ -7,8 +7,15 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 /**
- * lazily-calculates the transcript. Thread safe to call get() multiple times once commit() is
- * called and before clear() is called.
+ * A deferred list of formatted strings, one per line. The formatting is deferred until when (and
+ * if) get() is called. It's thread safe to call get() multiple times, but add() should be called
+ * from a single thread.
+ *
+ * <p>Once get() is called, the contents are 'resolved' into a string, and subsequent get() calls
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/DeferredTranscript.java
```diff
@@ -0,0 +1,72 @@
...
+ * lazily-calculates the transcript. Thread safe to call get() multiple times once commit() is
+ * called and before clear() is called.
+ */
+public class DeferredTranscript implements Supplier<String> {
+
+    private final class FormatEntry {
+        public final CharSequence fmt;
+        public final Object[] args;
+
+        public FormatEntry(String fmt, Object[] args) {
```
macchiati, line 19: Object... is better, because then it takes both a list and an array.

macchiati: Or make this private

macchiati (approved):

btangmu (approved):
Kudos for including the tests!

tools/cldr-code/src/main/java/org/unicode/cldr/util/DeferredTranscript.java
srl295: It actually is a private object. this is the internal storage for the deferred format.

The public API is `DeferredTranscript.add()`` which does take `Object...`

srl295:
This class could potentially have other uses, perhaps detailed explanations of error checks or things like that.  Any time you don't know _a priori_ whether a person is going to want to see the details or not.

macchiati (approved):

PR #3467, opened 2024-01-28
CLDR-16329 Add convertUnit exponent attribute for beaufort; update UnitConverter and some related code

pedberg-icu:
CLDR-16329

- Add the `exponent` attribute for the `convertUnit` element to support the non-linear unit conversions needed for Beaufort.
- Document how `exponent` is intended to be used in combination with the existing `factor` and `offset` attributes for conversions between sourceUnit and baseUnit.
- Add the `convertUnit` data for Beaufort to units.xml, using the new `exponent` attribute
- Update UnitConverter to handle `exponent`. May not be complete (can finish in separate PR).
    - As part of this, needed to add to Rational a version of `pow` that handles non-integer exponents. This initial version may not be as accurate or efficient as desired.
- Updated some of the test code to at least compile and pass tests with the updated version of UnitCoverter. More s probably needed in this area, can be a separate PR.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
@@ -277,7 +277,7 @@ public final class Rational implements Comparable<Rational> {
         if (exponent.equals(Rational.ONE)) {
             return this;
         }
-        // Temporary quick implementation
+        // TODO Temporary implementation, improve per CLDR-16329 additional PR or follow-on ticket
         double resultDouble = Math.pow(this.doubleValue(), exponent.doubleValue());
         return Rational.of(new BigDecimal(resultDouble));
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
@@ -273,6 +273,15 @@ public final class Rational implements Comparable<Rational> {
         return new Rational(numerator.pow(i), denominator.pow(i));
     }
 
+    public Rational pow(Rational exponent) {
+        if (exponent.equals(Rational.ONE)) {
+            return this;
+        }
+        // Temporary quick implementation
+        double resultDouble = Math.pow(this.doubleValue(), exponent.doubleValue());
+        return Rational.of(new BigDecimal(resultDouble));
+    }
+
     public static Rational pow10(int i) {
         return i > 0 ? TEN.pow(i) : TENTH.pow(-i);
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
@@ -286,12 +286,21 @@ public class UnitConverter implements Freezable<UnitConverter> {
         }
 
         public String toString(String unit) {
+            String exponentStart = "";
+            String exponentEnd = "";
+            if (exponent.denominator.compareTo(BigInteger.ONE) != 0) {
+                exponentStart = " ( ";
+                exponentEnd = " ) ";
+            }
             return factor.toString(FormatStyle.formatted)
                     + " * "
                     + unit
                     + (exponent.equals(Rational.ONE)
                             ? ""
-                            : " ^ " + exponent.toString(FormatStyle.formatted))
+                            : " ^ "
+                                    + exponentStart
+                                    + exponent.toString(FormatStyle.formatted)
+                                    + exponentEnd)
                     + (offset.equals(Rational.ZERO)
                             ? ""
                             : (offset.compareTo(Rational.ZERO) < 0 ? " - " : " + ")
@@ -303,12 +312,21 @@ public class UnitConverter implements Freezable<UnitConverter> {
         }
 
         public String toDecimal(String unit) {
+            String exponentStart = "";
+            String exponentEnd = "";
+            if (exponent.denominator.compareTo(BigInteger.ONE) != 0) {
+                exponentStart = " ( ";
+                exponentEnd = " ) ";
+            }
             return factor.toBigDecimal(MathContext.DECIMAL64)
                     + " * "
                     + unit
                     + (exponent.equals(Rational.ONE)
                             ? ""
-                            : " ^ " + exponent.toBigDecimal(MathContext.DECIMAL64))
+                            : " ^ "
+                                    + exponentStart
+                                    + exponent.toBigDecimal(MathContext.DECIMAL64)
+                                    + exponentEnd)
                     + (offset.equals(Rational.ZERO)
                             ? ""
                             : (offset.compareTo(Rational.ZERO) < 0 ? " - " : " + ")
@@ -827,12 +845,13 @@ public class UnitConverter implements Freezable<UnitConverter> {
         return new ConversionInfo(
                 numerator.divide(denominator),
                 Rational.ONE,
-                offset); // TODO: handle exponent here - how?
+                offset); // TODO: handle exponent here - how? CLDR-16329 additional PR or follow-on
+        // ticket
     }
 
     /** Only for use for simple base unit comparison */
     private class UnitComparator implements Comparator<String> {
-        // TODO: handle exponent here - how?
+        // TODO: handle exponent here - how? CLDR-16329 additional PR or follow-on ticket
         // TODO, use order in units.xml
 
         @Override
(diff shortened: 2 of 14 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -274,9 +289,12 @@ public String toString(String unit) {
             return factor.toString(FormatStyle.formatted)
                     + " * "
                     + unit
+                    + (exponent.equals(Rational.ONE)
+                            ? ""
+                            : " ^ " + exponent.toString(FormatStyle.formatted))
                     + (offset.equals(Rational.ZERO)
                             ? ""
-                            : (offset.compareTo(Rational.ZERO) < 0 ? " - " : " - ")
+                            : (offset.compareTo(Rational.ZERO) < 0 ? " - " : " + ")
```
sffc, line 297: Hmm, curious fix

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -788,11 +824,15 @@ public ConversionInfo parseUnitId(
...
-        return new ConversionInfo(numerator.divide(denominator), offset);
+        return new ConversionInfo(
+                numerator.divide(denominator),
+                Rational.ONE,
+                offset); // TODO: handle exponent here - how?
     }
 
     /** Only for use for simple base unit comparison */
     private class UnitComparator implements Comparator<String> {
+        // TODO: handle exponent here - how?
```
sffc, line 835: Hmm, a comparator with an exponent factor? The only way to actually compare is to have the specific quantities; they're not comparable in the general sense. For the purposes of this impl, I think it's probably best to consider exponents as sorted before all other factors (so Ax^B sorts first by B and then by A). This gives a total ordering which is true for sufficiently large values (the limit as x approaches infinity).

sffc (commented):
Looks good, though there are some TODOs in the code. Would also like @younies to take a look since he is working on this code in icu4x.

macchiati (commented):
BTW, I'd drafted this before an email discussion about TODOs (or Shanes comments), but hadn't posted.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
```diff
@@ -273,6 +273,15 @@ public Rational pow(int i) {
         return new Rational(numerator.pow(i), denominator.pow(i));
     }
 
+    public Rational pow(Rational exponent) {
+        if (exponent.equals(Rational.ONE)) {
+            return this;
+        }
+        // Temporary quick implementation
+        double resultDouble = Math.pow(this.doubleValue(), exponent.doubleValue());
```
macchiati, line 281: For any fractional exponents (like Beaufort, both to and from) we can't remain in Rational space. We might be able to get slightly better (if we want to).

Add a Rational pow(int exponent), which creates a rational from numerator.pow(exponent) / denominator.pow(exponent).

Add a Rational root(int integer), that takes the nth root. That's where (other than for 1) we have to get into BigDecimal space.

The Rational pow(Rational exponent) becomes this.pow(exponent.numerator).root(exponent.denominator)

[I think this works; need to double-check]

BTW, if we ever did log units, we'd have to add another field... https://en.wikipedia.org/wiki/Logarithmic_scale#Logarithmic_units

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: @sffc But B might be < 1. Should we check that first and then decide priority of A vs B based on that?

sffc: If B < 1 then it should sort before everything where B = 1 (all existing units).

I think this is a fair ordering because if `x` is sufficiently large, B < 1 will cause it to be at the bottom.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
```diff
@@ -273,6 +273,15 @@ public Rational pow(int i) {
         return new Rational(numerator.pow(i), denominator.pow(i));
     }
 
+    public Rational pow(Rational exponent) {
+        if (exponent.equals(Rational.ONE)) {
+            return this;
+        }
+        // Temporary quick implementation
```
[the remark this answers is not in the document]
pedberg-icu: Fixed this and other TODOs (added CLDR ticket).

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -274,9 +289,12 @@ public String toString(String unit) {
             return factor.toString(FormatStyle.formatted)
                     + " * "
                     + unit
+                    + (exponent.equals(Rational.ONE)
+                            ? ""
+                            : " ^ " + exponent.toString(FormatStyle.formatted))
```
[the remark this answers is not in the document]
pedberg-icu: Good point, fixed this.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Rational.java
pedberg-icu: I can look at improving this in the second PR for this ticket.

pedberg-icu:
> Looks good, though there are some TODOs in the code. Would also like @younies to take a look since he is working on this code in icu4x.

Thanks, I did add CLDR ticket number to the TODOs.

robertbastian:
I was just made aware of this by @sffc in the context of implementing Beaufort in ICU4X, and will leave this comment without diving too deep:

I'm confused by what's happening here. Beaufort is a scale, not a unit:
* It relates to wind speed like the scale ["freezing", "cold", "warm", "hot"] relates to temperature, or the amount of fresh snow fall relates to the five step avalanche danger scale
* The windspeed can be Bft 3, but it cannot be Bft 3.4.
* While it is possible to _look up_ 4 m/s on the Beaufort scale, converting the other way doesn't really work. Bft 3 represents the range 3.4 - 5.4 m/s, combined with a certain effect on the environment, and a given uncertainty
* You cannot combine it with another units: Bft/s doesn't make any sense, while (m/s)/s does.
* It doesn't make sense to add two Bft values together
  * Yes you can apparently do it if you can have the right exponential, but what does the result mean?
* Bft is inherently inaccurate as a measure. It's used to classify wind conditions over a longer period, like in a weather forecast. If you ask a sailor or a meteorologist what the current wind speed is, you will get an answer in m/s or knots.
  * As Wikipedia quotes:
  * > Nowadays, meteorologists typically express wind speed in kilometres or miles per hour or, for maritime and aviation purposes, [knots](https://en.wikipedia.org/wiki/Knot_(unit)); but Beaufort scale terminology is still sometimes used in weather forecasts for shipping[[4]](https://en.wikipedia.org/wiki/Beaufort_scale#cite_note-4) and the [severe weather](https://en.wikipedia.org/wiki/Severe_weather) warnings given to the public.[[5]](https://en.wikipedia.org/wiki/Beaufort_scale#cite_note-5)

While it is desirable for ICU to be able to localize the term "Bft 3", I don't think the units framework is the right place for it. Maybe ICU, or CLDR, needs a new concept for this; there are other scales that would fit like the Richter scale, other weather scales, and maybe things like academic grades.

pedberg-icu:
> I was just made aware of this by @sffc in the context of implementing Beaufort in ICU4X, and will leave this comment without diving too deep...

@robertbastian Thanks for your comment. The motivation for this is that in some regions (e.g. Netherlands, China), Beaufort-scale units are the preferred way of reporting wind speed in general weather reports (not just for maritime use). As such we would like to be able to add Beaufort units to the unit preferences data for usage wind speed. And to add any unit to the unit framework, it must be convertible to other units (notably to the base unit for the category). It is certainly the case that other units are often rounded in conventional usage so you may not get exact round-tripping; Beaufort is not especially exceptional in that way (e.g. a person's age is often expressed in integer years). There is a mathematical relationship between Beaufort-scale units and meters/sec. When converting the other way (to Beaufort) it is understood that the results should be rounded to the nearest integer (separately we are considering ways to provide that information).

robertbastian:
I absolutely see the need to interpret a speed value as a Beaufort number, but I don't follow how this implies it has to be a unit. I guess if you have a units conversion framework, every nail looks like a unit?

> It is certainly the case that other units are often rounded in conventional usage so you may not get exact round-tripping; Beaufort is not especially exceptional in that way (e.g. a person's age is often expressed in integer years).

It is though, because unlike my age, which we can write down with infinite precision, Bft just does not exist unrounded, and that's not a convention, that's the definition. There is no such thing as 3.456Bft, and ICU suggesting that there is is not a correct usage of the scale. ICU is too influential to handwave like this.

I don't see much discussions on the tickets, and I was unfortunately not in the CLDR meeting last August. The exponential units functionality seems useful for decibels, so it's probably nice to have, but not for Beaufort.

I think the only operation we should expose is to convert a speed into an integer Beaufort value between 0 and 12 (well, maybe 13 as an overflow value). If we use the units framework in the background that's fine, but I think doing irrational math is probably less efficient than converting to a canonical unit and then running a binary search over 12 numbers.

pedberg-icu:
@robertbastian I think you have convinced me, will discuss in CLDR meeting today

pedberg-icu:
Will replace with another PR

PR #3472, opened 2024-01-30
CLDR-17167 Replace lateral inheritance markers in trunk (v45, data)

btangmu:
-This is the effect of running ... modify -fV -scommon/main

-Running ... modify -fV -scommon/annotations has no effect

-The change to CLDRModify.java has no effect with current data (println not executed)

CLDR-17167

ALLOW_MANY_COMMITS=true


btangmu:
This is a DRAFT PR for investigation -- I don't recommend merging it, at least not yet

macchiati (changes requested):
I'm confused. According to the design doc, the only items we should need to harden are in top level locales (whose parent is root), but there are a lot of other locales here.

btangmu:
Per discussion in yesterday's infrastructure meeting, it's not clear whether "modify -fV" or anything like it is needed anymore. Conversion between hard bailey and inheritance marker is currently mainly governed by the method VoteResolver.reviseInheritanceAsNeeded. I don't know whether that is sufficient to prevent the kind of problem that occurred when we first attempted to "drop hard inheritance".

PR #3473, opened 2024-01-30
CLDR-17143 Deprecate collation type reformed, remove it from coverage and locale data

pedberg-icu:
CLDR-17143

Mark collation type reformed as deprecated, remove it from coverage and locale data

ALLOW_MANY_COMMITS=true


pedberg-icu:
Working on unit test failures (some were masked by unrelated NPEs)

common/main/en.xml
```diff
@@ -1376,7 +1376,6 @@ annotations.
 			<type key="collation" type="phonebook">Phonebook Sort Order</type>
 			<type key="collation" type="phonetic">Phonetic Sort Order</type>
 			<type key="collation" type="pinyin">Pinyin Sort Order</type>
-			<type key="collation" type="reformed">Reformed Sort Order</type>
```
markusicu, line 1379: On one hand, it seems like we might want to keep display names until some time (maybe 1 year?) after their item starts being deprecated.
On the other hand, I suspect that this collation type and its display names are rarely used... so fine with me.
Is it normal for CLDR to remove display data immediately when deprecating something?

markusicu (commented):
changes lgtm
looking forward to test fixes

common/main/en.xml
pedberg-icu: CLDR stopped using the "reformed" type in CLDR 42. We decided in TC today that it was okay to remove the translations in this PR.

markusicu (approved):

PR #3474, opened 2024-01-30
CLDR-17326 cache Factory.getSourceDirectoriesForLocale()

srl295:
CLDR-17326 

- includes a disabled test - could be run manually.

- Note: I wasn't able to measure a very large improvement, but this code did show up as a hotspot in profiling.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleFactory.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleFactory.java
@@ -2,6 +2,8 @@ package org.unicode.cldr.util;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleFactory.java
```diff
@@ -775,33 +779,55 @@ public File[] getSourceDirectories() {
         return sourceDirectories;
     }
 
+    LoadingCache<String, List<File>> sourceDirCache =
```
btangmu, line 782: How about making this private?

Also, if there are cases where a SimpleFactory is created but getSourceDirectoriesForLocale doesn't get called on it, then maybe lazy initialization would save some time/memory?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/SimpleFactory.java
srl295: +1 to private

> where a SimpleFactory is created but getSourceDirectories doesn't get called

THat would be a factory that never has `make()` called on it, which doesn't seem like a common use case, because factories are singletons.

btangmu (approved):

PR #3476, opened 2024-01-30
CLDR-17313 split out immutable DiskData in STFactory

srl295:
- stuff we only want to calculate once, versus what can be expired from the cache.

CLDR-17313

ALLOW_MANY_COMMITS=true


btangmu (approved):
Awesome!

PR #3487, opened 2024-02-01
CLDR-16329 Add convertUnit special attribute for beaufort; update UnitConverter and some related code

pedberg-icu:
CLDR-16329

This is an incomplete first pass at adding conversion support for Beaufort so it can be used in unitPreferences. It compiles and passes tests (but tests have not been updated to test or handle Beaufort). It replaces the first attempt in [#3467 ](https://github.com/unicode-org/cldr/pull/3467) whose approach was deemed wrong.
- Add the `special` attribute for the `convertUnit` element to support the special table-based conversions for Beaufort. When converting from meters/sec to Beaufort, these always map to an integral value that corresponds to a range of speeds that includes the input value. When converting from Beaufort to meters/sec, the Beaufort value is rounded to the nearest integer, then converted to a value in the corresponding range of speeds (generally the value from m/s = 0.836*Bft^(3/2), except for Bft 0).
- Document the new attribute.
- Add the `convertUnit` data for Beaufort to units.xml, using the new special attribute
- Update UnitConverter and SupplementalDataInfo to handle special (UnitConverter may not be quite complete, can finish in separate PR). Updated at least some of the tools code.
- Have not yet updated any of the test code.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -458,7 +458,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@MATCH:regex/[-+*/\._ 0-9a-zA-Z]+-->
     <!--@VALUE-->
 <!ATTLIST convertUnit special NMTOKEN #IMPLIED >
-    <!--@MATCH:regex/(100-)?[A-Za-z][-A-Za-z0-9]*-->
+    <!--@MATCH:regex/[A-Za-z][-A-Za-z0-9]*-->
     <!--@VALUE-->
 <!ATTLIST convertUnit systems NMTOKENS #IMPLIED >
     <!--@MATCH:set/literal/si, si_acceptable, metric, metric_adjacent, ussystem, uksystem, jpsystem, astronomical, person_age, other, prefixable-->
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -457,6 +457,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST convertUnit offset CDATA #IMPLIED >
     <!--@MATCH:regex/[-+*/\._ 0-9a-zA-Z]+-->
     <!--@VALUE-->
+<!ATTLIST convertUnit special NMTOKEN #IMPLIED >
+    <!--@MATCH:regex/(100-)?[A-Za-z][-A-Za-z0-9]*-->
+    <!--@VALUE-->
 <!ATTLIST convertUnit systems NMTOKENS #IMPLIED >
     <!--@MATCH:set/literal/si, si_acceptable, metric, metric_adjacent, ussystem, uksystem, jpsystem, astronomical, person_age, other, prefixable-->
     <!--@VALUE-->
--- a/common/supplemental/units.xml
+++ b/common/supplemental/units.xml
@@ -342,6 +342,8 @@ For terms of use, see http://www.unicode.org/copyright.html
        
         <!-- speed -->
         <convertUnit source='knot' baseUnit='meter-per-second' factor='1852/3600' systems="ussystem uksystem si_acceptable"/>
+        <convertUnit source='beaufort' baseUnit='meter-per-second' special='beaufort' systems="metric_adjacent"/>
+
         
         <!-- magnetic-induction -->
         <convertUnit source='tesla' baseUnit='kilogram-per-square-second-ampere' systems='si metric prefixable'/>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
@@ -242,7 +242,7 @@ public class UnitConverter implements Freezable<UnitConverter> {
         public final Rational factor;
         public final Rational offset;
         public String special;
-        public boolean inverse; // only used with special
+        public boolean specialInverse; // only used with special
 
         static final ConversionInfo IDENTITY = new ConversionInfo(Rational.ONE, Rational.ZERO);
 
@@ -250,20 +250,22 @@ public class UnitConverter implements Freezable<UnitConverter> {
             this.factor = factor;
             this.offset = offset;
             this.special = null;
-            this.inverse = false;
+            this.specialInverse = false;
         }
 
         public ConversionInfo(String special, boolean inverse) {
             this.factor = Rational.ZERO; // if ONE it will be treated as a base unit
             this.offset = Rational.ZERO;
             this.special = special;
-            this.inverse = inverse;
+            this.specialInverse = inverse;
         }
 
         public Rational convert(Rational source) {
             if (special != null) {
                 if (special.equals("beaufort")) {
-                    return (inverse) ? baseToBeaufort(source) : beaufortToBase(source);
+                    return (specialInverse)
+                            ? baseToScale(source, minMetersPerSecForBeaufort)
+                            : scaleToBase(source, minMetersPerSecForBeaufort);
                 }
                 return source;
             }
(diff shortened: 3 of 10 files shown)


pedberg-icu:
@younies and @robertbastian, please have a look (Robert, I know this is not completely what you were hoping for, but I think it is much closer...).

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -241,23 +241,113 @@ public UnitConverter cloneAsThawed() {
...
+                beaufortRound = BEAUFORT_MAX;
+            }
+            int beaufortIndex = beaufortRound.intValue();
+            return beaufortToMetersPerSec[beaufortIndex];
+        }
+
+        private Rational baseToBeaufort(Rational base) {
+            base = base.abs();
+            int beaufortIndex = BEAUFORT_MAX.intValue();
+            while (base.compareTo(minMetersPerSecForBeaufort[beaufortIndex]) < 0) {
```
robertbastian, line 341: How expensive is the rational comparison? This could be a binary search, or, assuming realistic values, should probably search from 0 up.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -241,23 +241,113 @@ public UnitConverter cloneAsThawed() {
...
+                if (special.equals("beaufort")) {
+                    return (inverse) ? beaufortToBase(source) : baseToBeaufort(source);
+                }
+                return source;
+            }
             return source.subtract(offset).divide(factor);
         }
 
+        private static final BigInteger BEAUFORT_MAX = BigInteger.valueOf(17);
+        private static final Rational[] beaufortToMetersPerSec = {
```
robertbastian, line 284: These values are *very* close to the midpoint of each bucket, at 12 Bft there's only a difference of 0.1 which could be a rounding error. You could calculate these from the other table.

common/dtd/ldmlSupplemental.dtd
```diff
@@ -457,6 +457,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST convertUnit offset CDATA #IMPLIED >
     <!--@MATCH:regex/[-+*/\._ 0-9a-zA-Z]+-->
     <!--@VALUE-->
+<!ATTLIST convertUnit special NMTOKEN #IMPLIED >
+    <!--@MATCH:regex/(100-)?[A-Za-z][-A-Za-z0-9]*-->
```
macchiati, line 461: Minor.
since this is a regular ID, it can be a bit simpler.
```suggestion
    <!--@MATCH:regex/[A-Za-z][-A-Za-z0-9]*-->
```

common/supplemental/units.xml
```diff
@@ -342,6 +342,8 @@ For terms of use, see http://www.unicode.org/copyright.html
        
         <!-- speed -->
         <convertUnit source='knot' baseUnit='meter-per-second' factor='1852/3600' systems="ussystem uksystem si_acceptable"/>
+        <convertUnit source='beaufort' baseUnit='meter-per-second' special='beaufort' systems="metric_adjacent"/>
```
robertbastian, line 345: except for the hardcoded table, the code is general enough to be applied to any scale of this type. Maybe this could be something like

```
special='scale[0.0, 0.3, 3.4, ..., 55.8]'
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -241,23 +241,113 @@ public UnitConverter cloneAsThawed() {
     public static final class ConversionInfo implements Comparable<ConversionInfo> {
         public final Rational factor;
         public final Rational offset;
+        public String special;
+        public boolean inverse; // only used with special
```
macchiati, line 245: if this were specialInverse the meaning would always be clear.
```suggestion
        public boolean specialInverse; // only used with special
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -241,23 +241,113 @@ public UnitConverter cloneAsThawed() {
...
+            return beaufortToMetersPerSec[beaufortIndex];
+        }
+
+        private Rational baseToBeaufort(Rational base) {
+            base = base.abs();
+            int beaufortIndex = BEAUFORT_MAX.intValue();
+            while (base.compareTo(minMetersPerSecForBeaufort[beaufortIndex]) < 0) {
+                beaufortIndex--;
+            }
+            return Rational.of(beaufortIndex);
```
macchiati, line 344: This should work fine, but just to mention that Java has https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#binarySearch-java.lang.Object:A-java.lang.Object-

robertbastian (commented):
I'm happy that the data model allows us to special-case this in ICU4X. I'm not super happy that we're special-casing Beaufort, and not something a bit more generic, but I guess we can change that later?

macchiati (changes requested):
NOTE: if you are in a hurry, let me know, because fixes for my comments could be postponed.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: Well the binary search would require two comparisons per iteration instead of one in the linear search; since the size is relatively small,I thought it might be a wash. But yes, probably more efficient to search from 0 up.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: Thanks, I should have checked that. Of course for Bft 17 there is no midpoint... But we can special-case that.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: Yes, as I mentioned to Robert,  the binary search would require two comparisons per iteration (both ends of range) instead of one in the linear search; since the size is relatively small,I thought it might be a wash.

pedberg-icu:
>...I'm not super happy that we're special-casing Beaufort, and not something a bit more generic, but I guess we can change that later?

Yes, we can certainly change that later, perhaps using something like your scale[...] suggestion

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
robertbastian: No, it would be a single comparison. Binary search gives you the insertion index of the value, which is the Bft number

macchiati: A rational comparison in CLDR is 2 multiplications and a comparison, of BigIntegers. For the CLDR implementation, that is not a problem because it is a reference implementation. Production implementations can optimize in a variety of ways.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: I was thinking more of code simplicity. For the CLDR reference implementation, accuracy and simplicity are much more important than shaving a few cycles off.

common/supplemental/units.xml
pedberg-icu: So in the latest version I made the code more generic: The functions converting between scale values and base value take an array of the minima for each scale value (plus an extra at the end to produce a reasonable "midrange" for the highest scale value). The special attribute still takes a string to select which array gets passed into these functions, instead of taking the array itself.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: I did end up using the binary conversion here, thanks.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: I did end up using this suggestion, computing the midpoints from the ranges, thanks.

common/dtd/ldmlSupplemental.dtd
pedberg-icu: Done, thanks

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
pedberg-icu: Done, I did end up switching to use Arrays.binarySearch

pedberg-icu:
@macchiati @robertbastian Thanks for your comments, I think I addressed all of them. Again, this is just the "part 1" for this ticket, still need to update the test code in a separate PR against the same ticket (just wanted to get this part in for the data freeze).

macchiati (approved):

PR #3488, opened 2024-02-01
CLDR-17311 Drop meatazone Macquarie, remove golden zone mapping for A…

yumaoka:
…merica_Mountain in Mexico.

CLDR-17311

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ab.xml
+++ b/common/main/ab.xml
@@ -3540,11 +3540,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					<daylight draft="unconfirmed">Макао, аԥхынтәи аамҭа</daylight>
 				</long>
 			</metazone>
-			<metazone type="Macquarie">
-				<long>
-					<standard draft="unconfirmed">Маккуори</standard>
-				</long>
-			</metazone>
 			<metazone type="Magadan">
 				<long>
 					<generic draft="unconfirmed">Магадан</generic>
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -4363,11 +4363,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					<daylight>Lord Howe-dagligtyd</daylight>
 				</long>
 			</metazone>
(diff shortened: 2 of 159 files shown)


yumaoka:
@macchiati There are basically two small changes in metaZones.xml.

**Dropping metazone Macquarie**
In the past, we assigned a name dedicated for Macquarie Island with metazone Macquarie. This region belongs to Australia and using the same time zone with Tasmania. TZ DB assigns short zone ID `AEST/AEDT`. This is same as other Australian east coast zones, such as Australia/Sydney, Australia/Melbourne. There is no convincing reason to assign a special name for Macquarie. At some point, we mapped Australia/Macquarie to metazone `Australia_Eastern`, but I probably forgot to remove golden zone mapping. The issue CLDR-17311 reported the golden zone mapping is incompatible with metazone mapping data for `Australia/Macquarie`. The right action should be deleting the bad golden zone mapping. Also metazone `Macquarie` is no longer used by any zones, it should be dropped. We have translation for the metazone in locale files. This is why a lot of files are included in this PR.

**Dropping golden zone mapping for America_Mountain in Mexico**
America/Hermosillo is a zone with standard offset of UTC-7. In Mexico, UTC-7 zone is called as Pacific, instead of Mountain. We found this and created a metazone Mexico_Pacific if my memory is correct. Because the zone has been using UTC-7, we probably mapped this zone to metazone `America_Mountain` before. Then when we update the metazone mapping to `Mexico_Pacific`, we probably forgot to drop the golden zone mapping below.
```
<mapZone other="America_Mountain" territory="MX" type="America/Hermosillo"/>
```
At this point, no Mexican time zone uses metazone mapping to America_Mountain except a zone recently introduced - `America/Ciudad_Juarez`. This zone is used by the region near the US border. This zone used different offsets in the past until recently, it is not qualified as a golden zone. So the final solution is to drop the golden zone mapping for America_Mountain in Mexico.

I'd like to ask you to review coverageLevels.xml change for Macquarie is what we want. I'm not familiar with this area.

macchiati (approved):

PR #3491, opened 2024-02-01
CLDR-17350 Fix Indonesian RBNF rules

grhoten:
CLDR-17350

ALLOW_MANY_COMMITS=true


richgillam (approved):
I don't speak Indonesian, but other than that, LOKTM.

PR #3494, opened 2024-02-05
CLDR-17357 alpha0 integration fix 1: Need valid end dates for currency ANG

pedberg-icu:
CLDR-17357

The first of possibly several alpha0 integration fixes:
- The end date for currency ANG needs to be valid (June 30, not June 31), follow-on for CLDR-17274

ALLOW_MANY_COMMITS=true


AEApple (approved):

macchiati:
Good catch, sorry!

PR #3495, opened 2024-02-07
CLDR-17177 kbd: update per feedback

srl295:
- documented 'implementation'
- added a Notation section
- corrections to discussion of repository and file layout

CLDR-17177

ALLOW_MANY_COMMITS=true

PR #3497, opened 2024-02-07
CLDR-17177 kbd: layers / modifiers update

srl295:
CLDR-17177

- this PR is just for the modifiers section
- attempted to avoid overlap with #3495 as well

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1356,7 +1356,7 @@ For example, in the element `<layer … modifiers="ctrlL altL, altR" …` the at
 - `ctrlL altL` (two components)
 - `altR` (one component)
 
-The order of the sets and the order of the components within each set is not significant.
+The order of the sets and the order of the components within each set is not significant. However, for clarity in reading, the canonical order within a set is in the order listed in Layout Modifier Components; the canonical order for the sets should be first by the cardinality of the sets (least first), then alphabetical.
 
 #### Layer Modifier Components
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1324,7 +1324,7 @@ _Attribute:_ `modifiers` (required for `hardware`)
 >
 > This attribute is a list of lists. It is a comma-separated (`,`) list of modifier sets, and each modifier set is a space-separated list of modifier components.
 >
-> Each modifier component must mach `[A-Za-z0-9]+`. Extra whitespace is ignored.
+> Each modifier component must match `[A-Za-z0-9]+`. Extra whitespace is ignored.
 >
 > To indicate that no modifiers apply, the reserved name of `none` is used.
 
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -1322,7 +1322,7 @@ _Attribute:_ `modifiers` (required for `hardware`)
 >
 > For hardware layouts, the use of `@modifiers` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in each  `form`.
 >
-> This attribute is a list of lists. It is a comma-separated (`,`) list of modifier sets, and each modifier set is a space-separated list of modifier components.
+> This attribute value is a list of lists. It is a comma-separated (`,`) list of modifier sets, and each modifier set is a space-separated list of modifier components.
 >
 > Each modifier component must match `[A-Za-z0-9]+`. Extra whitespace is ignored.
 >
@@ -1351,7 +1351,7 @@ _Attribute:_ `modifiers` (required for `hardware`)
 #### Layer Modifier Sets
 
 The `@modifiers` attribute value contains one or more Layer Modifier Sets, separated by commas.
-For example, in the element `<layer … modifiers="ctrlL altL, altR" …` the attribute consists of two sets:
+For example, in the element `<layer … modifiers="ctrlL altL, altR" …` the attribute value consists of two sets:
 
 - `ctrlL altL` (two components)
 - `altR` (one component)
@@ -1463,7 +1463,7 @@ Here is an example of a `row` element:
 
 This is a container for variables to be used with [transform](#element-transform), [display](#element-display) and [key](#element-key) elements.
 
-Note that the `id=` attribute must be unique across all children of the `variables` element.
+Note that the `id=` attribute value must be unique across all children of the `variables` element.
 
 **Example**
 
@@ -1500,7 +1500,7 @@ _Attribute:_ `id` (required)
 _Attribute:_ `value` (required)
 
 > Strings may contain whitespaces. However, for clarity, it is recommended to escape spacing marks, even in strings.
-> This attribute may be escaped with `\u` notation, see [Escaping](#escaping).
+> This attribute value may be escaped with `\u` notation, see [Escaping](#escaping).
 > Variables may refer to other string variables if they have been previously defined, using `${string}` syntax.
 > [Markers](#markers) may be included with the `\m{…}` notation.
 
@@ -1556,9 +1556,9 @@ _Attribute:_ `id` (required)
 
 _Attribute:_ `value` (required)
 
-> The `value` attribute is always a set of strings separated by whitespace, even if there is only a single item in the set, such as `"A"`.
+> The `value` attribute value is always a set of strings separated by whitespace, even if there is only a single item in the set, such as `"A"`.
 > Leading and trailing whitespace is ignored.
-> This attribute may be escaped with `\u` notation, see [Escaping](#escaping).
+> This attribute value may be escaped with `\u` notation, see [Escaping](#escaping).
 > Sets may refer to other string variables if they have been previously defined, using `${string}` syntax, or to other previously-defined sets using `$[set]` syntax.
 > Set references must be separated by whitespace: `$[set1]$[set2]` is an error; instead use `$[set1] $[set2]`.
 > [Markers](#markers) may be included with the `\m{…}` notation.
@@ -1866,7 +1866,7 @@ All of the `transform` elements in a `transformGroup` are tested for a match, in
 
 _Attribute:_ `from` (required)
 
-> The `from` attribute consists of an input rule for matching the input context.
+> The `from` attribute value consists of an input rule for matching the input context.
 >
 > The `transform` rule and output pattern uses a modified, mostly subsetted, regular expression syntax, with EcmaScript syntax (with the `u` Unicode flag) as its baseline reference (see [MDN-REGEX](https://developer.mozilla.org/docs/Web/JavaScript/Guide/Regular_Expressions)). Differences from regex implementations will be noted.
 
@@ -2040,7 +2040,7 @@ The following are additions to standard Regex syntax.
 
 _Attribute:_ `to`
 
-> This attribute represents the characters that are output from the transform.
+> This attribute value represents the characters that are output from the transform.
 >
 > If this attribute is absent, it indicates that the no characters are output, such as with a backspace transform.
 >
@@ -2170,7 +2170,7 @@ _Attribute:_ `before`
 
 _Attribute:_ `order`
 
-> This attribute gives the primary order for the elements in the matched string in the `@from` attribute. The value is a simple integer between -128 and +127 inclusive, or a space separated list of such integers. For a single integer, it is applied to all the elements in the matched string. Details of such list type attributes are given after all the attributes are described. If missing, the order value of all the matched characters is 0. We consider the order value for a matched character in the string.
+> This attribute value gives the primary order for the elements in the matched string in the `@from` attribute. The value is a simple integer between -128 and +127 inclusive, or a space separated list of such integers. For a single integer, it is applied to all the elements in the matched string. Details of such list type attributes are given after all the attributes are described. If missing, the order value of all the matched characters is 0. We consider the order value for a matched character in the string.
 >
 > * If the value is 0 and its tertiary value is 0, then the character is the base of a new run.
 > * If the value is 0 and its tertiary value is non-zero, then it is a normal character in a run, with ordering semantics as described in the `@tertiary` attribute.
@@ -2186,7 +2186,7 @@ _Attribute:_ `order`
 
 _Attribute:_ `tertiary`
 
-> This attribute gives the tertiary order value to the characters matched. The value is a simple integer between -128 and +127 inclusive, or a space separated list of such integers. If missing, the value for all the characters matched is 0. We consider the tertiary value for a matched character in the string.
+> This attribute value gives the tertiary order value to the characters matched. The value is a simple integer between -128 and +127 inclusive, or a space separated list of such integers. If missing, the value for all the characters matched is 0. We consider the tertiary value for a matched character in the string.
 >
 > * If the value is 0 then the character is considered to have a primary order as specified in its order value and is a primary character.
 > * If the value is non zero, then the order value must be zero otherwise it is an error. The character is considered as a tertiary character for the purposes of ordering.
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-keyboards.md
```diff
@@ -1387,9 +1400,15 @@ Layers are matched exactly based on the modifier keys which are down. For exampl
 
 - `other` as a modifier will match if no other layers match.
 
-Multiple modifier sets may be separated by commas.  For example, `none, shift caps` will match either no modifiers *or* shift and caps.  `ctrlL altL, altR` will match either  left-control and left-alt, *or* right-alt.
+Multiple modifier sets are separated by commas.  For example, `none, shift caps` will match either no modifiers *or* shift and caps.  `ctrlL altL, altR` will match either  left-control and left-alt, *or* right-alt.
+
+Keystrokes processed where there isn’t an explicitly matching layer, and where there is no layer with `other` specified, are ignored.
```
miloush, line 1405: What does "keystrokes" mean here? Do you ignore the key pressed if it does not result in a layer? For example, if there is a layer `ctrl` and you press `ctrl` and then add `shift`, the shift is ignored and layer `ctrl` matches? Perhaps it's just wording, "key combinations"? It would be good to add some example. "For example, if only `none`, `caps` and `shift` layers exist, pressing `caps` and `shift` will result in no output.

srl295: Yes, it's then ignored if it doesn't result in a layer. I'll add an example.

mcdurdin (dismissed):

docs/ldml/tr35-keyboards.md
```diff
@@ -1369,7 +1382,7 @@ _Attribute:_ `modifiers` (required for `hardware`)
 1. `L` or `R` indicates a left- or right- side modifier only (such as `altL`)
  whereas `alt` indicates _either_ left or right alt key (that is, `altL` or `altR`). `ctrl` indicates either left or right ctrl key (that is, `ctrlL` or `ctrlR`).
 
-2.  If there are any layers (in the same `form=`) with a modifier `alt`, there may not also be another layer with `altL` or `altR`.  Similarly, if there is a layer with a modifier `ctrl`, there may not be a layer with `ctrlL` or `ctrlR`.
+2. Keyboard authors will be warned if a keyboard mixes `alt` with `altL`/`altR`, or `ctrl` with `ctrlL`/`ctrlR`.
```
macchiati, line 1385: That is incorrect wording for a spec. I assume you mean something like:
```suggestion
2. Keyboard implementations must warn keyboard authors if a keyboard mixes `alt` with `altL`/`altR`, or `ctrl` with `ctrlL`/`ctrlR`.
```

docs/ldml/tr35-keyboards.md
```diff
@@ -1387,9 +1400,15 @@ Layers are matched exactly based on the modifier keys which are down. For exampl
 
 - `other` as a modifier will match if no other layers match.
 
-Multiple modifier sets may be separated by commas.  For example, `none, shift caps` will match either no modifiers *or* shift and caps.  `ctrlL altL, altR` will match either  left-control and left-alt, *or* right-alt.
+Multiple modifier sets are separated by commas.  For example, `none, shift caps` will match either no modifiers *or* shift and caps.  `ctrlL altL, altR` will match either  left-control and left-alt, *or* right-alt.
+
+Keystrokes processed where there isn’t an explicitly matching layer, and where there is no layer with `other` specified, are ignored. Example: If there is a `ctrl` and `shift` layer, but no `ctrl shift` nor `other` layer, no output will result from `ctrl shift X`.
```
macchiati, line 1405: Awkward sentence
```suggestion
Keystrokes must be ignored where there isn’t a layer that explicitly matches nor a layer with `other`. Example: If there is a `ctrl` and `shift` layer, but no `ctrl shift` nor `other` layer, no output will result from `ctrl shift X`.
```

docs/ldml/tr35-keyboards.md
```diff
@@ -1387,9 +1400,15 @@ Layers are matched exactly based on the modifier keys which are down. For exampl
...
+
+Layers are not allowed to overlap in their matching.  For example, the keyboard author will receive an error if one layer specifies `alt shift` and another layer specifies `altR shift`.
+
+The order of `<layer>` elements is not significant.
```
macchiati, line 1409:
```suggestion
The order of `<layer>` elements is thus not significant.
```

macchiati (changes requested):

docs/ldml/tr35-keyboards.md
srl295: good update, thanks

srl295:
@macchiati what's outstanding before this can merge?

docs/ldml/tr35-keyboards.md
```diff
@@ -1293,8 +1292,10 @@ A `layer` element describes the configuration of keys on a particular layer of a
 **Syntax**
 
 ```xml
-<layer id="layerId" modifiers="{Set of Modifier Combinations}">
-    ...
+<layer id="layerId" modifiers="…modifier modifier, …modifier modifier, …">
```
mcdurdin, line 1295:
```suggestion
<layer id="…layerId" modifiers="…modifier modifier, …modifier modifier, …">
```

docs/ldml/tr35-keyboards.md
```diff
@@ -1321,7 +1322,9 @@ _Attribute:_ `modifiers` (required for `hardware`)
 >
 > For hardware layouts, the use of `@modifiers` as an identifier for a layer is sufficient since it is always unique among the set of `layer` elements in each  `form`.
 >
-> The set of modifiers must match `(none|([A-Za-z0-9]+)( [A-Za-z0-9]+)*)`
+> This attribute is a list of lists. It is a comma-separated (`,`) list of modifier sets, and each modifier set is a space-separated list of modifier components.
+>
+> The set of modifiers must match `(none|([A-Za-z0-9]+)( [A-Za-z0-9]+)*)(,[ ]?([A-Za-z0-9]+)( [A-Za-z0-9]+)*)*`
```
mcdurdin, line 1327: Thoughts:
* This regex prevents extra whitespace between modifiers. Is that intended?
* This regex is hard enough to read that I wonder if it is helpful to include in the spec?
* Apart from `_`, the character classes here are just `\w`. Could we go with:

`(none|\w+([ ]+\w+)*(,[ ]*\w+([ ]+\w+)*)*`

macchiati: I find it is simpler to say and easier to understand as:  

Each modifier in each set must match `[A-Za-z0-9]+`

I assume that extra whitespace is ok, eg ctrl, crtl  shift,shift

srl295: Also should be "This attribute is either `none` or a list of lists.. " and add that extra whitespace is not significant.

docs/ldml/tr35-keyboards.md
```diff
@@ -1324,7 +1324,7 @@ _Attribute:_ `modifiers` (required for `hardware`)
 >
 > This attribute is a list of lists. It is a comma-separated (`,`) list of modifier sets, and each modifier set is a space-separated list of modifier components.
 >
-> The set of modifiers must match `(none|([A-Za-z0-9]+)( [A-Za-z0-9]+)*)(,[ ]?([A-Za-z0-9]+)( [A-Za-z0-9]+)*)*`
+> Each modifier component must mach `[A-Za-z0-9]+`. Extra whitespace is ignored.
```
mcdurdin, line 1327:
```suggestion
> Each modifier component must match `[A-Za-z0-9]+`. Extra whitespace is ignored.
```

mcdurdin (dismissed):

docs/ldml/tr35-keyboards.md
```diff
@@ -1369,7 +1382,7 @@ _Attribute:_ `modifiers` (required for `hardware`)
 1. `L` or `R` indicates a left- or right- side modifier only (such as `altL`)
  whereas `alt` indicates _either_ left or right alt key (that is, `altL` or `altR`). `ctrl` indicates either left or right ctrl key (that is, `ctrlL` or `ctrlR`).
 
-2.  If there are any layers (in the same `form=`) with a modifier `alt`, there may not also be another layer with `altL` or `altR`.  Similarly, if there is a layer with a modifier `ctrl`, there may not be a layer with `ctrlL` or `ctrlR`.
+2. Keyboard implementations must warn keyboard authors if a keyboard mixes `alt` with `altL`/`altR`, or `ctrl` with `ctrlL`/`ctrlR`.
```
macchiati, line 1385:
```suggestion
2. Mixing `alt` with `altL`/`altR`, or `ctrl` with `ctrlL`/`ctrlR`, is strongly discouraged. Keyboard implementations detect and should warn keyboard authors if this occurs.
```

We can't say "must" if there might not be a UI for alerting authors.

docs/ldml/tr35-keyboards.md
```diff
@@ -1387,9 +1400,17 @@ Layers are matched exactly based on the modifier keys which are down. For exampl
...
 - `other` as a modifier will match if no other layers match.
 
-Multiple modifier sets may be separated by commas.  For example, `none, shift caps` will match either no modifiers *or* shift and caps.  `ctrlL altL, altR` will match either  left-control and left-alt, *or* right-alt.
+Multiple modifier sets are separated by commas.  For example, `none, shift caps` will match either no modifiers *or* shift and caps.  `ctrlL altL, altR` will match either  left-control and left-alt, *or* right-alt.
+
+Keystrokes must be ignored where there isn’t a layer that explicitly matches nor a layer with `other`. Example: If there is a `ctrl` and `shift` layer, but no `ctrl shift` nor `other` layer, no output will result from `ctrl shift X`.
+
+Layers are not allowed to overlap in their matching.  For example, the keyboard author will receive an error if one layer specifies `alt shift` and another layer specifies `altR shift`.
+
+There is one special case:  the `other` layer matches if and only if no other layer matches. Thus logically the `other` layer is matched after all other layers have been checked.
```
macchiati, line 1409:
```suggestion
Layers are not allowed to overlap in their matching.  For example, the keyboard author will receive an error if one layer specifies `alt shift` and another layer specifies `altR shift`.
There is one special case:  the `other` layer matches if and only if no other layer matches. Thus logically the `other` layer is matched after all other layers have been checked.
```

macchiati: Should be 1 paragraph

docs/ldml/tr35-keyboards.md
```diff
@@ -1387,9 +1400,17 @@ Layers are matched exactly based on the modifier keys which are down. For exampl
...
+
+Because there is no overlap allowed between layers, the order of `<layer>` elements is not significant.
 
-Keystrokes where there isn’t an explicitly matching layer, and where there is no layer with `other` specified, are ignored.
+> Note: The modifier syntax may be enhanced in the future. but will remain backwards compatible with the syntax described here.
```
macchiati, line 1413:
```suggestion
> Note: The modifier syntax may be enhanced in the future, but will remain backwards compatible with the syntax described here.
```

macchiati (changes requested):

docs/ldml/tr35-keyboards.md
```diff
@@ -1387,9 +1400,17 @@ Layers are matched exactly based on the modifier keys which are down. For exampl
 
 - `other` as a modifier will match if no other layers match.
```
miloush, line 1401: Just to clarify, do we support combining `other` with others? Like `shit others` to match both shift with anything else and shift alone.

srl295: No, it would be `shift, other`

docs/ldml/tr35-keyboards.md
srl295: As this is an interchange format, it's assumed that there could be a UI or similar (such as a build step which could fail).

docs/ldml/tr35-keyboards.md
```diff
@@ -1345,9 +1348,19 @@ _Attribute:_ `modifiers` (required for `hardware`)
 </layer>
 ```
 
+#### Layer Modifier Sets
+
+The `@modifiers` attribute contains one or more Layer Modifier Sets, separated by commas.
```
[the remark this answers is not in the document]
srl295: good point, thanks, done

mcdurdin (approved):

docs/ldml/tr35-keyboards.md
```diff
@@ -1345,9 +1348,19 @@ _Attribute:_ `modifiers` (required for `hardware`)
...
+For example, in the element `<layer … modifiers="ctrlL altL, altR" …` the attribute consists of two sets:
+
+- `ctrlL altL` (two components)
+- `altR` (one component)
+
+The order of the sets and the order of the components within each set is not significant.
```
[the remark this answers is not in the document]
macchiati: I'm not seeing a reason for not having a canonical representation; that allows comparison of two files to return whether the are equivalent. It is also the format that you'd want processing tools to emit.

There might be a good reason; I'd just like to know.

docs/ldml/tr35-keyboards.md
macchiati: The phrasing is just weird. I author a keyboard, send it to you, you run a program on it. How does the program know who I am to notify me? Maybe "must emit a detect and emit a warning"?

docs/ldml/tr35-keyboards.md
macchiati: Huh? It is correcting an obvious typo.

macchiati (changes requested):

docs/ldml/tr35-keyboards.md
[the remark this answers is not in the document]
srl295: I thought I accepted this change first.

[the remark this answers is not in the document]
srl295: @macchiati I went through and double checked your review comments.  I had accepted your suggestion of a canonical ordering in https://github.com/unicode-org/cldr/pull/3497/commits/0982276d53b3594e4b74848822e8f407a4db157f perhaps it didn't show up properly somehow?

srl295:
> correcting an obvious typo

I think the typo you mention was accepted (and credited) in https://github.com/unicode-org/cldr/pull/3497/commits/0982276d53b3594e4b74848822e8f407a4db157f

docs/ldml/tr35-keyboards.md
srl295: That change was accepted in https://github.com/unicode-org/cldr/pull/3497/commits/0982276d53b3594e4b74848822e8f407a4db157f

docs/ldml/tr35-keyboards.md
srl295: changed.

macchiati (changes requested):
missing "values" after "attributes". Please do a global search.

srl295:
> missing "values" after "attributes". Please do a global search.

Mark, I see it now thanks.

srl295:
@macchiati still mysterious why i made some changes twice.  Maybe a bad merge.  Anyway, please take a look again.

macchiati:
There is still some awkward language, such as:

> The indicator attribute describes a short string to be used in currently selected layout indicator, such as US

Again, value is missing (but there are so many in this section...). And aside from that, if the attribute value *described* "US" I'd expect it to be something like "a two letter acronym standing for the United States". So should be:

The indicator attribute value is a short string used in currently selected layout indicator, such as US.

Anyway, such copy-editing can be done later, so approving.

macchiati (approved):

srl295:
@macchiati i copied a link to that comment https://github.com/unicode-org/cldr/pull/3497#issuecomment-1965551948 into Jira so it gets addressed.

PR #3498, opened 2024-02-08
CLDR-17351 Updating CLDR time zone data with IANA TZDB 2024a

yumaoka:
CLDR-17351

ALLOW_MANY_COMMITS=true


yumaoka:
```
<mapZone other="Kazakhstan_Eastern" territory="001" type="Asia/Almaty"/>
```
No change for this golden zone mapping. The metazone `Kazakhstan_Eastern` is used by `Asia/Almaty` and `Asia/Qostanay` now. The metazone won't be used by any zones after March 1. Therefore, Asia/Almaty can be the golden zone for reverse mapping unless other zones use this metazone in future.

common/supplemental/windowsZones.xml
```diff
@@ -537,7 +537,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<!-- (UTC+05:00) Ashgabat, Tashkent -->
 			<mapZone other="West Asia Standard Time" territory="001" type="Asia/Tashkent"/>
 			<mapZone other="West Asia Standard Time" territory="AQ" type="Antarctica/Mawson"/>
-			<mapZone other="West Asia Standard Time" territory="KZ" type="Asia/Oral Asia/Aqtau Asia/Aqtobe Asia/Atyrau"/>
+			<!-- Microsoft may create a new zone dedicated for Almaty and Qostanay. -->
+			<mapZone other="West Asia Standard Time" territory="KZ" type="Asia/Oral Asia/Almaty Asia/Aqtau Asia/Aqtobe Asia/Atyrau Asia/Qostanay"/>
```
yumaoka, line 541: Almaty and Qostanay will use UTC+5 with no DST, same as other zones in Kazakhstan. Almaty is the most populous city in Kazakhstan, so it can be the primary zone for the region. However, I think Microsoft eventually need to create a new zone for Almaty/Qostanay to preserve historic UTC offset changes separate from this. So this is a temporary solution. This is the reason why I did not change the primary zone (Asia/Oral) in this PR. When Microsoft decide to add a new Windows zone for Almaty/Qostanay, then we will move them to the new zone.

common/supplemental/windowsZones.xml
```diff
@@ -569,13 +570,12 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<mapZone other="Nepal Standard Time" territory="001" type="Asia/Katmandu"/>
 			<mapZone other="Nepal Standard Time" territory="NP" type="Asia/Katmandu"/>
 
-			<!-- (UTC+06:00) Astana -->
-			<mapZone other="Central Asia Standard Time" territory="001" type="Asia/Almaty"/>
+			<!-- (UTC+06:00) Astana --> <!-- Microsoft probably keeps Central Asia Standard Time, but change Astana to something else. -->
+			<mapZone other="Central Asia Standard Time" territory="001" type="Asia/Bishkek"/>
```
yumaoka, line 574: Almaty will be no longer UTC+6. Microsoft uses display text `(UTC+06:00) Astana`. Astana is the capital of Kazakhstan, but TZ database uses most populous city in the region (in this case, Almaty). Because Almaty moved to UTC+5, it cannot be the default mapping of this windows zone. I'm sure Microsoft preserves this zone, but changing the display text.

I usually pick the most populous region as 001 mapping. If we use this rule, Urumqi is the one among these zones. However, official zone for Urmuqi is actually China Standard Time (UTC+8). UTC+6 is used as "defacto" offset there. For this reason, Urmuqi might not be the zone used for 001 mapping. I picked Bishkek here. This is the second populous region among these regions.

macchiati (approved):

PR #3500, opened 2024-02-08
CLDR-17370 Changed kk (Kazakh language) plural forms

socketpair:
Even though they have two forms of nouns,
the same form of words is used when talking
about a specific quantity of something.

CLDR-17370 Changed kk (Kazakh language) plural forms

ALLOW_MANY_COMMITS=true


AEApple:
Additional information was requested in the ticket in order to evaluate this proposal.

PR #3502, opened 2024-02-13
CLDR-17144 Disable all CLDRModify -f except -fQ/-fP/-fV

btangmu:
-Revise checkSuboptions to throw IllegalArgumentException for FIX other than Q, P, V

CLDR-17144

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
@@ -329,8 +329,8 @@ public class CLDRModify {
             System.out.println(HELP_TEXT1 + fixList.showHelp() + HELP_TEXT2);
             return;
         }
-        checkSuboptions(options[FIX], fixList.getOptions());
-        checkSuboptions(options[JOIN_ARGS], allMergeOptions);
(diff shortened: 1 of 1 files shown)


srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -648,6 +649,15 @@ private static void checkSuboptions(UOption givenOptions, UnicodeSet allowedOpti
                             + CldrUtility.LINE_SEPARATOR
                             + "Use -? for help.");
         }
+        if (i == FIX) {
+            final UnicodeSet allowedFilters = new UnicodeSet().add('P').add('Q').add('V');
+            for (char c : givenOptions.value.toCharArray()) {
```
btangmu, line 654: bug here -- should check for null before calling toCharArray, see https://unicode-org.atlassian.net/browse/CLDR-17405

btangmu: See https://github.com/unicode-org/cldr/pull/3521

PR #3503, opened 2024-02-14
CLDR-17356 In testLanguageTagParserIsValid, mark co-reformed as valid (though deprecated)

pedberg-icu:
CLDR-17356

Fix the earlier test failure in testLanguageTagParserIsValid() the right way, by marking co-reformed as valid (even though it is deprecated), as is done with co-direct. Then remove the logKnownIssue added earlier.

ALLOW_MANY_COMMITS=true


pedberg-icu:
Oops, wrong ticket number on the commit and title. Closing this, replacing with https://github.com/unicode-org/cldr/pull/3505 which has the right ticket.

PR #3504, opened 2024-02-14
CLDR-17336 Fix parentLocales

macchiati:
CLDR-17336

This turned out to be a bit more work than anticipated; not that many lines but spread over a few files. I'll mention the highlights below.

- modified to allow multiple components (and add plurals & grammaticalFeatures)
- allow nonlikelyScript as a single value for locale=

- Make the changes in the ticket, to use nonlikelyScript
- Remove the redundancies, add grammaticalFeatures, plurals

docs/ldml/tr35.md
- document the changes, removing the complexities of having the other components overlay the main one.

### The rest of the items are making the code work 

tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
- Tell JSON that the components have multiple elements. (The error message is very helpful, but could use some more guidance in the file for which place to fix. I put the change into SPLITTABLE_ATTRS instead of the ones mentioned.)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateProductionData.java
tools/cldr-code/src/main/java/org/unicode/cldr/util/Factory.java
tools/cldr-code/src/test/java/org/unicode/cldr/util/TestLocaleIDParser.java
- There was a hack to turn off 'main' inheritance, now uses 'collations' because that was the previous use case.
- BTW I don't understand why it was used for isAnnotationsDir??

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRLocale.java
- document that the getParent method returns the *main* parent. (That is the normally expected one.)

tools/cldr-code/src/main/java/org/unicode/cldr/util/LocaleIDParser.java
- replaced the boolean ignoreParentLocale by the enum component
- cleaned up the code a bit.

- when we added components before, for ICU4X, we didn't actually retrieve the data (so fixing this ticket will also fix CLDR-16361)
- makes parentLocales be a map of maps, so that the parentLocales are segmented by component
- a little code cleanup, but not too much
- fixed the api to access the new information

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestInheritance.java
tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestLocale.java
- fixed three tests to work properly
- The issue was that they were expecting the nonlikely parents to be explicitly in the data,
whereas now they are part of LocaleIDParser.getParent()
- For one of the tests I went to the extra effort to use CLDRLocale. 
It is sometimes clumsy because not all apis accept them (eg likely subtags)

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestSupplementalDataInfo.java
- added a test that the nonlikelyScript locales work property for main, 
and that plurals and collations have their defined behavior.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -5425,7 +5425,7 @@ XXX Code for transations where no currency is involved
     </codeMappings>
 
 	<parentLocales>
-		<parentLocale parent="root" locales="az_Arab az_Cyrl bal_Latn blt_Latn bm_Nkoo bs_Cyrl byn_Latn cu_Glag dje_Arab dyo_Arab en_Dsrt en_Shaw ff_Adlm ff_Arab ha_Arab iu_Latn kk_Arab ks_Deva ku_Arab kxv_Deva kxv_Orya kxv_Telu ky_Arab ky_Latn ml_Arab mn_Mong mni_Mtei ms_Arab pa_Arab sat_Deva sd_Deva sd_Khoj sd_Sind shi_Latn so_Arab sr_Latn sw_Arab tg_Arab ug_Cyrl uz_Arab uz_Cyrl vai_Latn wo_Arab yo_Arab yue_Hans zh_Hant"/>
+		<parentLocale parent="root" locales="nonlikelyScript"/>
 		<parentLocale parent="en_001" locales="en_150 en_AG en_AI en_AU en_BB en_BM en_BS en_BW en_BZ en_CC en_CK en_CM en_CX en_CY en_DG en_DM en_ER en_FJ en_FK en_FM en_GB en_GD en_GG en_GH en_GI en_GM en_GY en_HK en_ID en_IE en_IL en_IM en_IN en_IO en_JE en_JM en_KE en_KI en_KN en_KY en_LC en_LR en_LS en_MG en_MO en_MS en_MT en_MU en_MV en_MW en_MY en_NA en_NF en_NG en_NR en_NU en_NZ en_PG en_PK en_PN en_PW en_RW en_SB en_SC en_SD en_SG en_SH en_SL en_SS en_SX en_SZ en_TC en_TK en_TO en_TT en_TV en_TZ en_UG en_VC en_VG en_VU en_WS en_ZA en_ZM en_ZW"/>
 		<parentLocale parent="en_150" locales="en_AT en_BE en_CH en_DE en_DK en_FI en_NL en_SE en_SI"/>
 		<parentLocale parent="en_IN" locales="hi_Latn"/>
@@ -5436,15 +5436,11 @@ XXX Code for transations where no currency is involved
 		<parentLocale parent="zh_Hant_HK" locales="zh_Hant_MO"/>
 	</parentLocales>
 
-	<parentLocales component="segmentations">
-		<parentLocale parent="zh" locales="zh_Hant"/>
+	<parentLocales component="segmentations grammaticalFeatures plurals">
 	</parentLocales>
 
 	<parentLocales component="collations">
-		<parentLocale parent="bs" locales="bs_Cyrl"/>
-		<parentLocale parent="sr" locales="sr_Latn"/>
 		<parentLocale parent="sr_ME" locales="sr_Cyrl_ME"/>
-		<parentLocale parent="zh" locales="zh_Hant"/>
 		<parentLocale parent="zh_Hant" locales="yue yue_Hant"/>
 		<parentLocale parent="zh_Hans" locales="yue_CN yue_Hans yue_Hans_CN"/>
 	</parentLocales>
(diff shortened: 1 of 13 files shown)


macchiati:
@sffc adding you as a reviewer is mostly FYI; you might look at just the xml file change.

macchiati:
Hmmm. Still showing a problem with JSON, LdmlConvertRulesTest.testSplittableAttributes
Will have to futz with it tomorrow.

common/supplemental/supplementalData.xml
```diff
@@ -5436,15 +5436,11 @@ XXX Code for transations where no currency is involved
...
-	<parentLocales component="segmentations">
-		<parentLocale parent="zh" locales="zh_Hant"/>
+	<parentLocales component="segmentations grammaticalFeatures plurals">
 	</parentLocales>
 
 	<parentLocales component="collations">
-		<parentLocale parent="bs" locales="bs_Cyrl"/>
-		<parentLocale parent="sr" locales="sr_Latn"/>
 		<parentLocale parent="sr_ME" locales="sr_Cyrl_ME"/>
-		<parentLocale parent="zh" locales="zh_Hant"/>
```
sffc, line 5447: Do we need/want to add a `<parentLocale parent="root" locales="nonlikelyScript"/>` since the rules no longer inherit? If not, we are changing behavior; for example, `az-Arab` will now inherit from `az` instead of from root for the purpose of collation data.

macchiati: Yes, we are changing behavior, but it is to what it should have been. Collation, segmentation, grammar, etc don't want to skip over the base language — that was a mistake in the original addition of splitting parentLocales by component. We found that out when we delved deeper into what ICU was doing, and realized that what it was doing was the behavior that we really wanted.

(they may also have customizations of their own, of course).

common/supplemental/supplementalData.xml
```diff
@@ -5425,7 +5425,7 @@ XXX Code for transations where no currency is involved
     </codeMappings>
 
 	<parentLocales>
-		<parentLocale parent="root" locales="az_Arab az_Cyrl bal_Latn blt_Latn bm_Nkoo bs_Cyrl byn_Latn cu_Glag dje_Arab dyo_Arab en_Dsrt en_Shaw ff_Adlm ff_Arab ha_Arab iu_Latn kk_Arab ks_Deva ku_Arab kxv_Deva kxv_Orya kxv_Telu ky_Arab ky_Latn ml_Arab mn_Mong mni_Mtei ms_Arab pa_Arab sat_Deva sd_Deva sd_Khoj sd_Sind shi_Latn so_Arab sr_Latn sw_Arab tg_Arab ug_Cyrl uz_Arab uz_Cyrl vai_Latn wo_Arab yo_Arab yue_Hans zh_Hant"/>
+		<parentLocale parent="root" locales="nonlikelyScript"/>
```
srl295, line 5428: might have been better to use `x-nonlikelyScript` this way it's still bcp47

srl295 (approved):
seems ok. It will be a change for consumers who will have to look up the non-likely parents.  Perhaps that can be generated such as for JSON

common/supplemental/supplementalData.xml
```diff
@@ -5436,15 +5436,11 @@ XXX Code for transations where no currency is involved
 		<parentLocale parent="zh_Hant_HK" locales="zh_Hant_MO"/>
 	</parentLocales>
 
-	<parentLocales component="segmentations">
-		<parentLocale parent="zh" locales="zh_Hant"/>
+	<parentLocales component="segmentations grammaticalFeatures plurals">
```
srl295, line 5439: this is going to be a challenge for json because collations for example was part of the xpath, and now it's a list instead of a single item

I think I have an option that will just duplicate the data.  So it will generate a segmentations, grammaticalFeatures, plurals subkey with duplicate data (an empty set).  That will be least disruptive to consumers.

srl295: https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-core/supplemental/parentLocales.json#L190

srl295 (approved):

macchiati:
We deliberately chose an impossible locale id so it wouldn't silently fail.

docs/ldml/tr35.md
```diff
@@ -1768,19 +1768,43 @@ _Examples:_
 <!ATTLIST parentLocale locales NMTOKENS #REQUIRED >
 ```
 
-In some cases, the normal truncation inheritance does not function well. This happens when:
+When the component does not occur, that is referred to as the ‘main’ component.
+Otherwise the component value typically corresponds to an element and its children, such as ‘collations’ or ‘plurals’.
 
-1.  The child locale is of a different script. In this case, mixing elements from the parent into the child data results in a mishmash.
-2.  A large number of child locales behave similarly, and differently from the truncation parent.
+The basic inheritance model for locales of the form <lang>_<script>_<region>_<variant1>_…<variantN> is to truncate from the end. That is,
```
srl295, line 1774: Missed this before: should be `<script>` etc, otherwise breaks html output

docs/ldml/tr35.md
```diff
@@ -1768,19 +1768,43 @@ _Examples:_
...
+Thus there are two cases where the truncation inheritance needs to be overridden:
+
+1.  When the parent locale would have a different script, and text would be mixed.
+2.  In certain exceptional circumstances where the parent.
 
 The `parentLocale` element is used to override the normal inheritance when accessing CLDR data.
 
-For case 1, the children are script locales, and the parent is "root". For example:
+For case 1, there is a special value for the locales, `nonlikelyScript`,
+which includes all locales of the form <lang>_<script>, where the <script> is not the likely script for <lang>.
```
srl295, line 1798: should be

```
`<lang>_<script>`
```

..etc

PR #3505, opened 2024-02-14
CLDR-17345 In testLanguageTagParserIsValid, mark co-reformed as valid (though deprecated)

pedberg-icu:
CLDR-17345

Fix the earlier test failure in testLanguageTagParserIsValid() the right way, by marking co-reformed as valid (even though it is deprecated), as is done with co-direct. Then remove the logKnownIssue added earlier.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestLocale.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestLocale.java
@@ -953,6 +953,8 @@ public class TestLocale extends TestFmwkPlus {
                         LstrType.extension,
                         new AllowedMatch("co", "direct"),
                         LstrType.extension,
+                        new AllowedMatch("co", "reformed"),
+                        LstrType.extension,
                         new AllowedMatch("rg", "unknown"),
                         LstrType.extension,
                         new AllowedMatch(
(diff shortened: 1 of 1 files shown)


markusicu (approved):
Thanks, this looks good as far as it is handling this like other deprecated key-value pairs.

However, I am still baffled that deprecated values should fail the validity test at all. It seems wrong that we need special handling in the test code.

pedberg-icu:
> Thanks, this looks good as far as it is handling this like other deprecated key-value pairs.
> 
> However, I am still baffled that deprecated values should fail the validity test at all. It seems wrong that we need special handling in the test code.

I think the LTP Is treating them as valid, but the test code is comparing LTP results against the set provided by SupplementalDataInfo which may not include the deprecated ones, so the test code needs to add those in. I will check.

markusicu:
> > Thanks, this looks good as far as it is handling this like other deprecated key-value pairs.
> > However, I am still baffled that deprecated values should fail the validity test at all. It seems wrong that we need special handling in the test code.
> 
> I think the LTP Is treating them as valid, but the test code is comparing LTP results against the set provided by SupplementalDataInfo which may not include the deprecated ones, so the test code needs to add those in. I will check.

Ok, so it seems like the best way to fix this is to ask SupplementalDataInfo for both "valid" and "deprecated" values.
We promise stable identifiers, which means that we deprecate but do not remove. Implementation and tests should both follow that.

PR #3506, opened 2024-02-14
CLDR-11155 Pages too big; move some paths to new People2 and Symbols3

btangmu:
-New People2 (People & Body 2) added to emoji-test.txt starting with subgroup person-fantasy

-New Symbols3 hard-coded similar to Symbols2, temporarily only the last two rows (for ◊) are moved to the new page

-Temporary: throw an exception for Smileys & People, believed to be obsolete, in which case the code should be removed

CLDR-11155

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -381,4 +357,38 @@ private static ImmutableSet<String> buildPaths(String suffix) {
...
+            case People:
+                if (minorToOrder.get(minor) < minorToOrder.get("person-fantasy")) {
+                    return PageId.People;
+                } else {
+                    return PageId.People2;
+                }
+            case Symbols:
+                if (minorToOrder.get(minor) < minorToOrder.get("transport-sign")) {
+                    if (pageId.getSectionId() != SectionId.Characters) {
+                        throw new RuntimeException("oops");
```
btangmu, line 382: I'll delete this after testing locally and before making ready for review

btangmu: Tested OK locally with Symbols, Symbols2, and Symbols3: there's no need to check the section id here so I'll remove that check

btangmu: Removed that check

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -74,34 +75,28 @@ public class Emoji {
...
+         * Example from emoji-test.txt:
+         *   # group: Smileys & Emotion
+         *   # subgroup: face-smiling
+         *   1F600 ; fully-qualified # 😀 grinning face
+         */
         Splitter semi = Splitter.on(CharMatcher.anyOf(";#")).trimResults();
         String majorCategory = null;
         String minorCategory = null;
-        long majorOrder = 0;
-        long minorOrder = 0;
```
btangmu, line 85: majorOrder and minorOrder were assigned to but unused, resulting in compiler warnings; cleaned up

srl295 (approved):

PR #3508, opened 2024-02-15
CLDR-11155 Pages too big; move some paths to new Pages

btangmu:
-New Travel_Places2 (Travel & Places 2) starting with subgroup transport-ground

-New Objects2 starting with subgroup money

-New EmojiSymbols, Punctuation, MathSymbols, OtherSymbols, replacing Symbols2

-New methods PathHeader.createNonEmojiMap, PathHeader.getCharacterPageId

-New unit test TestAnnotations.testSymbols

-Fix some compiler warnings including dead code

-Comments, including TODO for Volume/Volume2

CLDR-11155

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
@@ -88,16 +88,10 @@ public class Emoji {
                 line = line.substring(1).trim();
                 if (line.startsWith("group:")) {
                     majorCategory = line.substring("group:".length()).trim();
-                    Long oldMajorOrder = majorToOrder.get(majorCategory);
-                    if (oldMajorOrder == null) {
-                        majorToOrder.put(majorCategory, (long) majorToOrder.size());
-                    }
+                    majorToOrder.computeIfAbsent(majorCategory, k -> (long) majorToOrder.size());
                 } else if (line.startsWith("subgroup:")) {
                     minorCategory = line.substring("subgroup:".length()).trim();
-                    Long oldMinorOrder = minorToOrder.get(minorCategory);
-                    if (oldMinorOrder == null) {
-                        minorToOrder.put(minorCategory, (long) minorToOrder.size());
-                    }
+                    minorToOrder.computeIfAbsent(minorCategory, k -> (long) minorToOrder.size());
                 }
                 continue;
             }
@@ -312,14 +306,6 @@ public class Emoji {
         return majorCat;
     }
 
-    public static Set<String> getMajorCategories() {
-        return emojiToMajorCategory.values();
-    }
-
-    public static Set<String> getMinorCategories() {
-        return emojiToMinorCategory.values();
-    }
-
     public static Set<String> getMinorCategoriesWithExtras() {
         Set<String> result = new LinkedHashSet<>(emojiToMinorCategory.values());
         result.addAll(EXTRA_SYMBOL_MINOR_CATEGORIES.getAvailableValues());
@@ -337,15 +323,10 @@ public class Emoji {
     }
 
     private static Set<String> NAME_PATHS = null;
-    private static Set<String> KEYWORD_PATHS = null;
     public static final String TYPE_TTS = "[@type=\"tts\"]";
 
     public static synchronized Set<String> getNamePaths() {
         return NAME_PATHS != null ? NAME_PATHS : (NAME_PATHS = buildPaths(TYPE_TTS));
-    }
-
-    public static synchronized Set<String> getKeywordPaths() {
-        return KEYWORD_PATHS != null ? KEYWORD_PATHS : (KEYWORD_PATHS = buildPaths(""));
     }
 
     private static ImmutableSet<String> buildPaths(String suffix) {
@@ -368,21 +349,22 @@ public class Emoji {
         final String major = getMajorCategory(emoji);
         final String minor = getMinorCategory(emoji);
         final PageId pageId = PageId.forString(major);
+        final Long minorOrder = minorToOrder.get(minor);
         switch (pageId) {
+            case Objects:
+                return (minorOrder < minorToOrder.get("money")) ? PageId.Objects : PageId.Objects2;
             case People:
-                if (minorToOrder.get(minor) < minorToOrder.get("person-fantasy")) {
-                    return PageId.People;
-                } else {
-                    return PageId.People2;
-                }
+                return (minorOrder < minorToOrder.get("person-fantasy"))
+                        ? PageId.People
+                        : PageId.People2;
             case Symbols:
-                if (minorToOrder.get(minor) < minorToOrder.get("transport-sign")) {
-                    return PageId.Symbols;
-                } else if (minorToOrder.get(minor) < minorToOrder.get("other-symbol")) {
-                    return PageId.Symbols2;
-                } else {
-                    return PageId.Symbols3;
-                }
+                return (minorOrder < minorToOrder.get("transport-sign"))
+                        ? PageId.Symbols
+                        : PageId.EmojiSymbols;
+            case Travel_Places:
+                return (minorOrder < minorToOrder.get("transport-ground"))
+                        ? PageId.Travel_Places
+                        : PageId.Travel_Places2;
             default:
                 return pageId;
         }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
@@ -1,11 +1,13 @@
 package org.unicode.cldr.util;
 
 import com.google.common.base.Splitter;
+import com.ibm.icu.dev.util.UnicodeMap;
 import com.ibm.icu.impl.Relation;
 import com.ibm.icu.impl.Row;
 import com.ibm.icu.lang.UCharacter;
 import com.ibm.icu.text.Collator;
 import com.ibm.icu.text.Transform;
+import com.ibm.icu.text.UnicodeSet;
 import com.ibm.icu.util.ICUException;
 import com.ibm.icu.util.Output;
 import com.ibm.icu.util.ULocale;
@@ -91,10 +93,10 @@ public class PathHeader implements Comparable<PathHeader> {
          * Changes are allowed as READ_WRITE, but field is always displayed as LTR, even in RTL
          * locales (used for patterns).
          */
-        LTR_ALWAYS;
+        LTR_ALWAYS
     }
 
-    private static EnumNames<SectionId> SectionIdNames = new EnumNames<>();
+    private static final EnumNames<SectionId> SectionIdNames = new EnumNames<>();
 
     /**
      * The Section for a path. Don't change these without committee buy-in. The 'name' may be
@@ -114,7 +116,7 @@ public class PathHeader implements Comparable<PathHeader> {
         Supplemental,
         Special;
 
-        private SectionId(String... alternateNames) {
+        SectionId(String... alternateNames) {
             SectionIdNames.add(this, alternateNames);
         }
 
@@ -128,9 +130,9 @@ public class PathHeader implements Comparable<PathHeader> {
         }
     }
 
-    private static EnumNames<PageId> PageIdNames = new EnumNames<>();
-    private static Relation<SectionId, PageId> SectionIdToPageIds =
-            Relation.of(new TreeMap<SectionId, Set<PageId>>(), TreeSet.class);
+    private static final EnumNames<PageId> PageIdNames = new EnumNames<>();
+    private static final Relation<SectionId, PageId> SectionIdToPageIds =
+            Relation.of(new TreeMap<>(), TreeSet.class);
 
     private static class SubstringOrder implements Comparable<SubstringOrder> {
         final String mainOrder;
@@ -230,6 +232,9 @@ public class PathHeader implements Comparable<PathHeader> {
         Length(SectionId.Units),
         Area(SectionId.Units),
         Volume(SectionId.Units),
+        // TODO: enable splitting Volume into Volume/Volume2
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-11155
+        // Volume2(SectionId.Units),
         SpeedAcceleration(SectionId.Units, "Speed and Acceleration"),
         MassWeight(SectionId.Units, "Mass and Weight"),
         EnergyPower(SectionId.Units, "Energy and Power"),
@@ -313,10 +318,14 @@ public class PathHeader implements Comparable<PathHeader> {
         Animals_Nature(SectionId.Characters, "Animals & Nature"),
         Food_Drink(SectionId.Characters, "Food & Drink"),
         Travel_Places(SectionId.Characters, "Travel & Places"),
+        Travel_Places2(SectionId.Characters, "Travel & Places 2"),
         Activities(SectionId.Characters),
         Objects(SectionId.Characters),
-        Symbols2(SectionId.Characters),
-        Symbols3(SectionId.Characters),
+        Objects2(SectionId.Characters),
+        EmojiSymbols(SectionId.Characters, "Emoji Symbols"),
+        Punctuation(SectionId.Characters),
+        MathSymbols(SectionId.Characters, "Math Symbols"),
+        OtherSymbols(SectionId.Characters, "Other Symbols"),
         Flags(SectionId.Characters),
         Component(SectionId.Characters),
         Typography(SectionId.Characters),
@@ -324,7 +333,7 @@ public class PathHeader implements Comparable<PathHeader> {
 
         private final SectionId sectionId;
 
-        private PageId(SectionId sectionId, String... alternateNames) {
+        PageId(SectionId sectionId, String... alternateNames) {
             this.sectionId = sectionId;
             SectionIdToPageIds.put(sectionId, this);
             PageIdNames.add(this, alternateNames);
@@ -382,7 +391,6 @@ public class PathHeader implements Comparable<PathHeader> {
     static final SupplementalDataInfo supplementalDataInfo = SupplementalDataInfo.getInstance();
     static final Map<String, String> metazoneToContinent =
             supplementalDataInfo.getMetazoneToContinentMap();
-    static final StandardCodes standardCode = StandardCodes.make();
     static final Map<String, String> metazoneToPageTerritory = new HashMap<>();
 
     static {
@@ -524,6 +532,15 @@ public class PathHeader implements Comparable<PathHeader> {
         ;
     }
 
+    /**
+     * Compare this PathHeader to another one
+     *
+     * @param other the object to be compared.
+     * @return 0 if equal, -1 if less, 1 if more
+     *     <p>Note: if we ever have to compare just the header or just the code, methods to do that
+     *     were in release 44 (compareHeader and compareCode), but they were unused and therefore
+     *     removed in CLDR-11155.
+     */
     @Override
     public int compareTo(PathHeader other) {
         // Within each section, order alphabetically if the integer orders are
@@ -544,7 +561,7 @@ public class PathHeader implements Comparable<PathHeader> {
             }
             long longResult;
             if (0 != (longResult = codeOrder - other.codeOrder)) {
-                return longResult < 0 ? -1 : longResult > 0 ? 1 : 0;
+                return longResult < 0 ? -1 : 1;
             }
             if (codeSuborder != null) { // do all three cases, for transitivity
                 if (other.codeSuborder != null) {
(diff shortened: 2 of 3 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -368,21 +349,31 @@ public static PageId getPageId(String emoji) {
...
+                return (minorOrder < minorToOrder.get("transport-ground"))
+                        ? PageId.Travel_Places
+                        : PageId.Travel_Places2;
+                // TODO: enable splitting Volume into Volume/Volume2
+                // Since they aren't emoji, the following doesn't work
+                // Reference: https://unicode-org.atlassian.net/browse/CLDR-11155
+                // case Volume:
+                //    return (minorOrder < minorToOrder.get("acre-foot"))
+                //            ? PageId.Volume
+                //            : PageId.Volume2;
```
btangmu, line 376: I realize now this whole method probably shouldn't be in the Emoji class. I need to study what corresponding mechanism can be used for non-emoji pages...

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -91,10 +91,10 @@ public enum SurveyToolStatus {
          * Changes are allowed as READ_WRITE, but field is always displayed as LTR, even in RTL
          * locales (used for patterns).
          */
-        LTR_ALWAYS;
+        LTR_ALWAYS
```
btangmu, line 94: changes in this file are all refactoring to fix compiler warnings, except for the addition of Travel_Places2 and  Objects2 and a TODO comment for adding Volume2

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -312,14 +306,6 @@ public static String getMajorCategory(String emoji) {
         return majorCat;
     }
 
-    public static Set<String> getMajorCategories() {
```
macchiati, line 315: Why delete these? Remember that the unicodetools uses the CLDR APIs, so you can't assume that public APIs are only used internal to CLDR. We don't want to accidentally break unicodetools

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -368,21 +349,31 @@ public static PageId getPageId(String emoji) {
...
             case Symbols:
-                if (minorToOrder.get(minor) < minorToOrder.get("transport-sign")) {
-                    return PageId.Symbols;
-                } else if (minorToOrder.get(minor) < minorToOrder.get("other-symbol")) {
-                    return PageId.Symbols2;
-                } else {
-                    return PageId.Symbols3;
-                }
+                return (minorOrder < minorToOrder.get("transport-sign"))
+                        ? PageId.Symbols
```
macchiati, line 362: I didn't catch this in the earlier PR. PageId.Symbols is *not* emoji, but rather for **numbers**:
```
    public enum PageId {
...
        Symbols(SectionId.Numbers),
```
eg https://cldr-staging.unicode.org/cldr-apps/v#/de/Symbols/

The symbol page IDs for emoji should be: PageId.Symbols2, Symbols3, Symbols4

(We might want to give these more meaningful names in a later PR.)

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -337,17 +323,12 @@ public static UnicodeSet getNonConstructed() {
...
 
     private static Set<String> NAME_PATHS = null;
-    private static Set<String> KEYWORD_PATHS = null;
     public static final String TYPE_TTS = "[@type=\"tts\"]";
 
     public static synchronized Set<String> getNamePaths() {
         return NAME_PATHS != null ? NAME_PATHS : (NAME_PATHS = buildPaths(TYPE_TTS));
     }
 
-    public static synchronized Set<String> getKeywordPaths() {
```
macchiati, line 347: ditto

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
macchiati: It could go here as well, just for convenience, but I agree that for the items that are not in emoji-test.txt, we need a different mechanism. We could try categorizing by unicode General category; need to look at that.

macchiati (changes requested):
I didn't

macchiati:
I did some analysis of the non-emoji characters.

What I suggest is to add the following test, and rework the code to match. Part of this is renaming the current Symbols2 to be "Emoji Symbols", for clarity. It also adds 3 new pages (not just Symbols3) so that there is some semantic differentiation (and smaller pages).

- Punctuation
- Math Symbol
- Other Symbol

Here is the current output of this test.

```
  testSymbols {
    Error: (TestAnnotations.java:674) Mismatch in 3 cases: expected=null actual=Symbols2 [²³¹]
    Error: (TestAnnotations.java:674) Mismatch in 1 cases: expected=null actual=Symbols3 [µ]
    Error: (TestAnnotations.java:674) Mismatch in 116 cases: expected=Emoji Symbols actual=Symbols2 [‼⁉↔-↙↩↪⏏⏩-⏯⏸-⏺▶◀☢☣☦☪☮☯☸♀♂♈-♓♾♿⚛⚠⚧⛎⛔✖✝✡❓-❕❗➕-➗➡⤴⤵⬅-⬇〰🎦🏧💱💲📳-📶🔀-🔆🔙-🔞🔯🔼🔽🕉🕎🚫🚭-🚱🚳🚷-🚼🚾🛂-🛅🛐🛜🟰🪯]
    Error: (TestAnnotations.java:674) Mismatch in 30 cases: expected=Math Symbol actual=Symbols2 [+<->|~¬±×÷⁻←-↓∆-∈−√∞∩∪≈≠≡≤≥⊂♯]
    Error: (TestAnnotations.java:674) Mismatch in 103 cases: expected=Math Symbol actual=Symbols3 [⁄↚↛↠↣↦↮⇎⇏⇒⇔⇵∀∂∃∅∉∋∎∏∑∓∕∗-∙∝∟∠∣∥∧∫∬∮∴-∷∼-∾≃≅≌≒≖≣≦≧≪-≬≮≯≳≺≻⊁⊃⊆⊇⊕-⊛⊞⊟⊥⊮⊰⊱⊶⊹⊿⋁-⋃⋅⋆⋈⋒⋘⋙⋭-⋱▷◁◿⨧⨯⨼⩣⩽⪍⪚⪺]
    Error: (TestAnnotations.java:674) Mismatch in 20 cases: expected=Other Symbol actual=Symbols2 [\$\^`¢£¥¨°´₥₩€₱₹₽₿⇅⇆♪♭]
    Error: (TestAnnotations.java:674) Mismatch in 140 cases: expected=Other Symbol actual=Symbols3 [₢-₤₨₰₳₶₷℗↜-↟↡↢↤↥↧↨↫-↭↯-⇄⇇-⇍⇐⇑⇓⇖-⇪■-▩▬-▮▰▲-▵▸-▿◂-◙◜-◦◯◳◷﷼]
    Error: (TestAnnotations.java:674) Mismatch in 68 cases: expected=Punctuation actual=Symbols2 [!#%-')*,-/\:;?@\[-\]_\{\}¡§«¶·»¿،؛؟‐–-―‘-‚“-„†-•…‧‰′-‴※、。〈-】〔-〗・]
    Error: (TestAnnotations.java:674) Mismatch in 4 cases: expected=Punctuation actual=Symbols3 [‸‽‾⁂]
  } (0.317s) FAILED (9 failure(s))
```

```
static final UnicodeSet allRgiNoES = Emoji.getAllRgiNoES();
static final UnicodeSet punctuation = new UnicodeSet("[:P:]").freeze();
static final UnicodeSet mathSymbol = new UnicodeSet("[:Sm:]").freeze();
static final UnicodeSet otherSymbol = new UnicodeSet("[:S:]").removeAll(mathSymbol).freeze();

public void testSymbols() {
    CLDRFile root = CLDRConfig.getInstance().getAnnotationsFactory().make("root", false);
    UnicodeMap<String> expectedMap =
            new UnicodeMap<String>()
                    .putAll(punctuation, "Punctuation")
                    .putAll(mathSymbol, "Math Symbol")
                    .putAll(otherSymbol, "Other Symbol")
                    .freeze();
    Set<String> nonEmojiPages = expectedMap.values();
    UnicodeMap<Pair<String, String>> failures = new UnicodeMap<>();
    PathHeader.Factory phf = PathHeader.getFactory();
    for (String path : root) {
        String cp = parts.getAttributeValue(-1, "cp");
        if (cp == null) {
            continue; // non-annotation line
        }
        PathHeader ph = phf.fromPath(path);
        SectionId sectionId = ph.getSectionId();
        assertEquals("Section for " + cp, SectionId.Characters, sectionId);
        PageId pageId = ph.getPageId();
        final String actual = pageId.toString();

        // collect all the failures rather than having a long list of errors

        if (allRgiNoES.contains(cp)) { // check emoji
            if (nonEmojiPages.contains(actual)) {
                failures.put(cp, Pair.of("«Emoji-Page»", actual));
            } else if (actual.equals("Symbols2")) {
                failures.put(cp, Pair.of("Emoji Symbols", actual));
            }
        } else {
            String expected = expectedMap.get(cp);
            if (!actual.equals(expected)) {
                failures.put(cp, Pair.of(expected, actual));
            }
        }
    }
    if (!failures.isEmpty()) {
        for (Pair<String, String> value : ImmutableSortedSet.copyOf(failures.values())) {
            UnicodeSet uset = failures.getSet(value);
            errln(
                    "Mismatch in "
                            + uset.size()
                            + " cases: expected="
                            + value.getFirst()
                            + " actual="
                            + value.getSecond()
                            + " "
                            + uset.toPattern(false));
        }
    }
}
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
btangmu: I'm glad you pointed that out! Is there nothing in the cldr project to indicate which APIs are needed for unicodetools? Just now I did a search in unicodetools for getMajorCategories and the result was nothing found; here's a screenshot:
image

btangmu: To answer the question, "Why delete these?": dead code is detrimental in many ways, especially for maintainability. The ability to distinguish dead code from live code is essential. Do we have no way to do that within the cldr project?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
btangmu: I also just now searched for getKeywordPaths in unicodetools and it wasn't found -- but we should have some way of encapsulating such dependencies!

btangmu: I had never built unicodetools before. Just now I downloaded it, ran `mkdir -p Generated/BIN`, and ran `mvn install package -DCLDR_DIR=/[...]/cldr`. All the tests passed. There were warnings but no indication of methods missing from cldr.

macchiati:
Ok, great

srl295:
@macchiati good points about API, we could have a build step that tries to at least build unicodetools against latest.

btangmu:
@macchiati I've added your unit test, and made a few changes attempting to make the test pass, but it still fails with
```
    testSymbols {
      Error: (TestAnnotations.java:672) Mismatch in 4 cases: expected=Math Symbols actual=Emoji Symbols
[←-↓]
      Error: (TestAnnotations.java:672) Mismatch in 103 cases: expected=Math Symbols actual=Other Symbols
[⁄↚↛↠↣↦↮⇎⇏⇒⇔⇵∀∂∃∅∉∋∎∏∑∓∕∗-∙∝∟∠∣∥∧∫∬∮∴-∷∼-∾≃≅≌≒≖≣≦≧≪-≬≮≯≳≺≻⊁⊃⊆⊇⊕-⊛⊞⊟⊥⊮⊰⊱⊶⊹⊿⋁-⋃⋅⋆⋈⋒⋘⋙⋭-⋱▷◁◿⨧⨯⨼⩣⩽⪍⪚⪺]
      Error: (TestAnnotations.java:672) Mismatch in 2 cases: expected=Other Symbols actual=Emoji Symbols
[⇅⇆]
      Error: (TestAnnotations.java:672) Mismatch in 21 cases: expected=Other Symbols actual=Math Symbols
[\$\^`¢£¥¨°²-´¹₥₩€₱₹₽₿♪♭]
      Error: (TestAnnotations.java:672) Mismatch in 68 cases: expected=Punctuation actual=Math Symbols
[!#%-')*,-/\:;?@\[-\]_\{\}¡§«¶·»¿،؛؟‐–-―‘-‚“-„†-•…‧‰′-‴※、。〈-】〔-〗・]
      Error: (TestAnnotations.java:672) Mismatch in 4 cases: expected=Punctuation actual=Other Symbols
[‸‽‾⁂]
      Error: (TestAnnotations.java:672) Mismatch in 15 cases: expected=«Emoji-Page» actual=Math Symbols
[‼⁉♾✖❓-❕❗➕-➗〰💱💲🟰]
      Error: (TestAnnotations.java:672) Mismatch in 94 cases: expected=«Emoji-Page» actual=Other Symbols
[©®™ℹⓂ▪▫◻-◾☑♻⚕⚜⚪⚫✅✔✳✴❇❌❎➰➿⬛⬜⭕〽㊗㊙🅰🅱🅾🅿🆎🆑-🆚🈁🈂🈚🈯🈲-🈺🉐🉑💠📛🔘🔠-🔤🔰-🔻🟠-🟫]
    } (0.067s) FAILED (8 failure(s))
```
I don't understand how the division of old Symbols2 into Emoji Symbols, Math Symbols, and Other Symbols is supposed to work. Does that require changing the order in which the paths are presented in Survey Tool? For example, the test expects ⇅ under "Other Symbols", but this row

https://st.unicode.org/cldr-apps/v#/sv/Symbols2/41f6e1fb98b0cd4e

is very near the top of the current Symbols2 page. I was under the impression that the partitioning would preserve the current order. If the order needs to change that seems a lot more far-reaching...

macchiati:
The order should be preserved *within* the given section. (In the default
Unicode collation sequence, the emoji symbols are interspersed with other
symbols, but we want to separate out the non-emoji for our purposes.)

btangmu:
@macchiati "The order should be preserved *within* the given section" -- by "section" here do you mean "page"? For example, "Emoji Symbols" is a page?

If I'm understanding right, the current minorToOrder map can't be used to determine which page a character belongs on. Instead the page should be determined by code similar to testSymbols -- using `UnicodeSet("[:P:]")` and so forth.

macchiati:
Right.

macchiati:
The alternative would be to build it into the major order, but I think that
might be more complicated.

btangmu:
@macchiati My last commit adds your proposed unit test, and fixes the code so that test passes (at least locally). However, when I run the server, it fails here:

```
    private static PageId getCharacterPageId(String cp) {
        if (emojiSet.contains(cp)) {
            return Emoji.getPageId(cp);
        }
        PageId pageId = nonEmojiMap.get(cp);
        if (pageId == null) {
            throw new InternalCldrException("Failure getting character page id");
```
That's for the character ₳ = U+20B3 AUSTRAL SIGN, in the path  //ldml/annotations/annotation[@cp="₳"]

Evidently neither emojiSet nor nonEmojiMap contains the character ₳; which of them should contain it?

Stack:

getCharacterPageId:2480, PathHeader (org.unicode.cldr.util)
transform:2144, PathHeader$Factory$43 (org.unicode.cldr.util)
transform:2141, PathHeader$Factory$43 (org.unicode.cldr.util)
fix:2238, PathHeader$Factory (org.unicode.cldr.util)
fromPath:721, PathHeader$Factory (org.unicode.cldr.util)
fromPath:659, PathHeader$Factory (org.unicode.cldr.util)
<init>:38, SurveyMenus (org.unicode.cldr.web)
lambda$new$2:1824, STFactory (org.unicode.cldr.web)
get:-1, 376447219 (org.unicode.cldr.web.STFactory$$Lambda$775)
get:181, Suppliers$NonSerializableMemoizingSupplier (com.google.common.base)
getSurveyMenus:1829, STFactory (org.unicode.cldr.web)
processRequest:565, SurveyAjax (org.unicode.cldr.web)

On smoketest I find the path here in Symbols3:

https://cldr-smoke.unicode.org/cldr-apps/v#/ab/Symbols3/72c50da9e766dcb8

On production (earlier code I find it here in Symbols2:

https://st.unicode.org/cldr-apps/v#/fr/Symbols2/72c50da9e766dcb8

I do find the character ₳ under "other-symbol" in Emoji.java. I don't find it in emoji-test.txt.

btangmu:
I've changed this to a draft PR due to the problem just mentioned

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
btangmu: As it turns out, I'm keeping that method in Emoji, but only calling it if the character is in Emoji.getAllRgiNoES()

A new method PathHeader.getCharacterPageId handles the non-emoji cases (but not ₳ which is currently in limbo)

btangmu:
@macchiati BTW otherSymbols.size() = 1,112,322

macchiati:
re ₳: that's odd; maybe I missed it. It is a currency sign, so you can add
it there.

otherSymbols.size() = 1,112,322
That is because it is a grab-bag; for anything that has an annotations path
that are neither emoji, nor punctuation, nor ...

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
```diff
@@ -88,16 +88,10 @@ public class Emoji {
                 line = line.substring(1).trim();
                 if (line.startsWith("group:")) {
                     majorCategory = line.substring("group:".length()).trim();
-                    Long oldMajorOrder = majorToOrder.get(majorCategory);
-                    if (oldMajorOrder == null) {
-                        majorToOrder.put(majorCategory, (long) majorToOrder.size());
-                    }
+                    majorToOrder.computeIfAbsent(majorCategory, k -> (long) majorToOrder.size());
```
macchiati, line 91: nice

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
macchiati: Not without verifying that unicodetools still works. That is the main project that uses some of the CLDR api.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
macchiati: ok, good

tools/cldr-code/src/main/java/org/unicode/cldr/util/Emoji.java
macchiati: Needs work. This needs to test against the UnicodeSets (as outlined) rather than the minor orders.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -574,42 +582,6 @@ public int compareTo(PathHeader other) {
         }
     }
 
-    public int compareHeader(PathHeader other) {
```
macchiati, line 577: Add a comment that if we ever have to compare just the header or just the code, the code to do that is in release 44. (These are a pain to get right.)

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
btangmu: I've just added such a comment in the method compareTo

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -2516,4 +2456,42 @@ public boolean canReadAndWrite() {
...
+            throw new InternalCldrException("Failure getting character page id");
+        }
+        // TODO: enable splitting Volume into Volume/Volume2, starting Volume2 at "acre-foot"
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-11155
+        return pageId;
+    }
+
+    /**
+     * Create the map from non-emoji characters to pages. Call with lazy initialization to avoid
+     * static initialization bugs, otherwise PageId.OtherSymbols could still be null.
```
btangmu, line 2486: When nonEmojiMap was static initialized, surprisingly PageId.OtherSymbols was still null, and that's why I was getting strange results earlier -- putAll was putting null for PageId.OtherSymbols!

btangmu:
Testing locally, the new pages appear to load fine, so I've made this ready for review

macchiati:
Can you go ahead and post to cldr-staging so I can see it in action?

btangmu:
@macchiati here's one of the new pages: https://cldr-staging.unicode.org/cldr-apps/v#/ab/Punctuation/

macchiati (approved):
excellent

PR #3510, opened 2024-02-21
CLDR-17006 v46 ddl: zh_Hans_MY

srl295:
CLDR-17006

- creating as empty per contribution

ALLOW_MANY_COMMITS=true


AEApple:
LGTM! but I think it would be good to check with the member orgs if there is data which should be upstreamed e.g. CLDR-17183

srl295:
> LGTM! but I think it would be good to check with the member orgs if there is data which should be upstreamed e.g. [CLDR-17183](https://unicode-org.atlassian.net/browse/CLDR-17183)

yes, they may want to participate also

DavidLRowe (approved):
LGTM

PR #3511, opened 2024-02-21
CLDR-17371 stream filter to replace xmlns with DOCTYPE

srl295:
- optimized
- assumes current version DTD
- supports spec change in #3495 

CLDR-17371

ALLOW_MANY_COMMITS=true

### DIFF
--- a/keyboards/3.0/mt-t-k0-47key.xml
+++ b/keyboards/3.0/mt-t-k0-47key.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE keyboard3 SYSTEM "../dtd/ldmlKeyboard3.dtd">
-<keyboard3 locale="mt-t-k0-47key" conformsTo="techpreview">
+<keyboard3 xmlns="https://schemas.unicode.org/cldr/45/keyboard3" locale="mt-t-k0-47key" conformsTo="techpreview">
     <locales>
         <!-- English is also an official language in Malta.-->
         <locale id="en" />
--- a/keyboards/dtd/ldmlKeyboard3.dtd
+++ b/keyboards/dtd/ldmlKeyboard3.dtd
@@ -17,6 +17,9 @@ Please view the subcommittee page for the most recent information.
 <!ATTLIST keyboard3 conformsTo (techpreview) #REQUIRED >
     <!--@MATCH:any-->
     <!--@METADATA-->
+<!ATTLIST keyboard3 xmlns CDATA #IMPLIED >
+    <!--@MATCH:any-->
+    <!--@METADATA-->
 
 <!ELEMENT import EMPTY >
     <!--@TECHPREVIEW-->
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLFileReader.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLFileReader.java
@@ -182,6 +182,8 @@ public class XMLFileReader {
             AllHandler allHandler) {
         try {
             XMLReader xmlReader = createXMLReader(handlers, validating, allHandler);
+            // wrap the reader to insert a character stream
+            DoctypeXmlStreamWrapper.wrap(is);
             is.setSystemId(systemID);
             try {
                 xmlReader.parse(is);
@@ -198,7 +200,7 @@ public class XMLFileReader {
         }
     }
 
-    private static final XMLReader createXMLReader(
+    public static final XMLReader createXMLReader(
             int handlers, boolean validating, AllHandler allHandler)
             throws SAXNotRecognizedException, SAXNotSupportedException {
         XMLReader xmlReader = createXMLReader(validating);
--- a/keyboards/3.0/bn.xml
+++ b/keyboards/3.0/bn.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE keyboard3 SYSTEM "../dtd/ldmlKeyboard3.dtd">
-<keyboard3 locale="bn" conformsTo="techpreview">
+<keyboard3 xmlns="https://schemas.unicode.org/cldr/45/keyboard3" locale="bn" conformsTo="techpreview">
     <!--
         History:
           Based on
--- a/keyboards/3.0/fr-t-k0-azerty.xml
+++ b/keyboards/3.0/fr-t-k0-azerty.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE keyboard3 SYSTEM "../dtd/ldmlKeyboard3.dtd">
 <!--
     This file is part of the CLDR Keyboard Technical Preview.
     This is a sample data file.
@@ -8,7 +7,7 @@
 
 	Also NOTE: this is really a test keyboard.  CLDR-12026 will be for the real new azerty keyboard
 -->
-<keyboard3 locale="fr-t-k0-azerty" conformsTo="techpreview">
+<keyboard3 xmlns="https://schemas.unicode.org/cldr/45/keyboard3" locale="fr-t-k0-azerty" conformsTo="techpreview">
 	<locales>
 		<locale id="br" /> <!-- example of including Breton -->
 	</locales>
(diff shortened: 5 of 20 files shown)


srl295:
kbd-check fails because of the xmlns, to be addressed on the keyman side in https://github.com/keymanapp/keyman/issues/10803

keyboards/3.0/mt-t-k0-47key.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE keyboard3 SYSTEM "../dtd/ldmlKeyboard3.dtd">
-<keyboard3 locale="mt-t-k0-47key" conformsTo="techpreview">
+<keyboard3 xmlns="https://schemas.unicode.org/cldr/45/keyboard3" locale="mt-t-k0-47key" conformsTo="techpreview">
```
srl295, line 2: I've only changed one file, but the others can be changed as well once there's consensus.

keyboards/dtd/ldmlKeyboard3.dtd
```diff
@@ -17,6 +17,8 @@ Please view the subcommittee page for the most recent information.
 <!ATTLIST keyboard3 conformsTo (techpreview) #REQUIRED >
     <!--@MATCH:any-->
     <!--@METADATA-->
+<!ATTLIST keyboard3 xmlns CDATA #IMPLIED >
+    <!--@METADATA-->
```
macchiati, line 21: Should add appropriate MATCH annotations. Should be separate ticket, though.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLFileReader.java
```diff
@@ -173,6 +173,8 @@ public static void read(
             AllHandler allHandler) {
         try {
             XMLReader xmlReader = createXMLReader(handlers, validating, allHandler);
+            // wrap the reader to insert a character stream
+            DTDInsertingReaderFactory.wrap(is);
```
macchiati, line 177: It seems like we should control this with an input parameter, rather than doing it all the time. We could change the `boolean validating` parameter from a boolean to something like:

`enum Validation {nonValidating, validating, xmlms}`

keyboards/dtd/ldmlKeyboard3.dtd
srl295: maybe `<!--@MATCH:any-->`, but we shouldn't specify it further.  The default namespace is specified in https://www.w3.org/TR/xml-names/#defaulting and should not need to be specified here.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLFileReader.java
srl295: Why does it need to be controlled?
- The purpose of the wrap function is to present a validating, `<DOCTYPE`'ed xml document to code that's expecting such
- it's only applied in cases where there's NO DOCTYPE and xmlns: is present.
- if DOCTYPE is not found, the buffer is pushed back into the stream, and reading proceeds.

keyboards/dtd/ldmlKeyboard3.dtd
srl295: done

srl295:
@macchiati I don't think this  PR needs to block the _spec_, this PR is to show that the spec is supportable in Java.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLFileReader.java
macchiati: You are saying that there is no significant performance or maintenance impact of using DTDInsertingReaderFactory.wrap(is);

macchiati: BTW, seems odd that you are tossing the return value from .wrap

srl295: > BTW, seems odd that you are tossing the return value from .wrap

It's a fluent operation, it modifies the return value.

srl295: > You are saying that there is no significant performance or maintenance impact of using DTDInsertingReaderFactory.wrap(is);

There may be a performance impact, that's why I noted that this PR could stay open until those are known.

However, in most cases, the caller will not know whether the DTD is needed or not until the file is read.

macchiati: good point, the caller might not know. We would know, however, based on the directory.

macchiati:
> @macchiati I don't think this PR needs to block the _spec_, this PR is to show that the spec is supportable in Java. (once tests pass!)
> 
> So, I'm OK to leave this PR open and make improvements if needed. API changes, optimization etc

Makes sense

tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLFileReader.java
srl295: seems to be a statistically insignificant performance impact.

srl295:
Measured 11s±1s for 10,000 iterations. Perhaps slight improvement (due to read-ahead or inefficiencies in the XML parser?) when using the wrap.  The quick-check is now at the byte or char array level.

macchiati:
I'm a bit confused. Are you saying that it takes 11s / 10000 iterations
with the wrap and approximately the same without?

srl295:
> I'm a bit confused. Are you saying that it takes 11s / 1000 iterations with the wrap and approximately the same without?

11s/1000, sorry. I'm re-running the tests.

srl295:
Raw results.
10,000 iterations (faster, different, nearly idle system)

```xml
  <testcase name="TestBytePerf(boolean)[1] wrapped=false" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="74.146"/>
  <testcase name="TestBytePerf(boolean)[2] wrapped=true" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="74.809"/>
  <testcase name="TestBytePerf(boolean)[3] wrapped=false" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="76.95"/>
  <testcase name="TestBytePerf(boolean)[4] wrapped=true" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="75.434"/>
  <testcase name="TestBytePerf(boolean)[5] wrapped=false" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="77.519"/>
  <testcase name="TestBytePerf(boolean)[6] wrapped=true" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="74.999"/>
  <testcase name="TestBytePerf(boolean)[7] wrapped=false" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="77.698"/>
  <testcase name="TestBytePerf(boolean)[8] wrapped=true" classname="org.unicode.cldr.util.TestDoctypeXmlStreamWrapper" time="77.906"/>
```

macchiati (dismissed):

srl295:
Thanks. @macchiati pk to move ticket to accepted ? (Had sent mail to TC)

macchiati:
Ok, you can. Assuming pk =ok

srl295:
> Ok, you can. Assuming pk =ok
> 

Failing build step is not CLDR tests, it's the Keyman compiler now out of date with these PRs. Once they merge I can fix it.

macchiati:
Looks like the tests fail

mcdurdin:
> Looks like the tests fail

as @srl295 wrote [previously](https://github.com/unicode-org/cldr/pull/3511#issuecomment-1962417699):

> Failing build step is not CLDR tests, it's the Keyman compiler now out of date with these PRs. Once they merge I can fix it.

He is working on a PR to the Keyman compiler to allow the `<keyboard3 xmlns` attribute in the validation phase in the compiler, once that merges then that will resolve the test here.

srl295:
> Looks like the tests fail

CLDR's unit tests pass.

srl295:
> > Looks like the tests fail
> 
> CLDR's unit tests pass. 

I've disabled the keyboard check workflow for now. I'll re enable it once these merge and I can update the compiler to the changed spec.

srl295:
ping, can i get a review on this?  All unit tests pass.

macchiati (approved):

PR #3519, opened 2024-02-24
CLDR-11155 Test for ST pages with too many rows

macchiati:
CLDR-11155

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java
@@ -1,5 +1,15 @@
 package org.unicode.cldr.unittest;
 
+import com.google.common.base.Joiner;
+import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.TreeMultimap;
+import com.ibm.icu.impl.Relation;
+import com.ibm.icu.impl.Row;
+import com.ibm.icu.impl.Row.R2;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -19,7 +29,6 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.stream.Collectors;
-
 import org.unicode.cldr.test.CoverageLevel2;
 import org.unicode.cldr.test.ExampleGenerator;
 import org.unicode.cldr.util.CLDRConfig;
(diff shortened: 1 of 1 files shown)


macchiati:
I took the code I made to find out the page sizes, and made it into a test. Once the emoji pages are handled, the number of failures will go down. (We can also raise the error threshold or make some exceptions if we want.)

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java
```diff
@@ -1677,4 +1679,70 @@ public void TestUnitOrder() {
             }
         }
     }
+
+    public void testPageSize() {
+        final long minError = 600; // above this, emit error
+        final long minLog = 500; // otherwise above this, emit warning
+        Factory factory = CLDRConfig.getInstance().getCommonAndSeedAndMainAndAnnotationsFactory();
+        // "en", "cs", "ar", "pl"
```
srl295, line 1687:
```suggestion
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java
```diff
@@ -1677,4 +1679,70 @@ public void TestUnitOrder() {
...
+                long count = c.getCount(entry);
+                if (count > minError) {
+                    errln(
+                            locale
+                                    + "\t"
+                                    + entry.getSectionId()
+                                    + "\t"
+                                    + entry
+                                    + "\thas too many entries:\t"
+                                    + count);
```
srl295, line 1714: - String.format will avoid taking up 6 lines
- include `minError` also so it's clear what the target value is

btangmu:
https://github.com/unicode-org/cldr/pull/3508 is merged, so I'll restart the tests on this one

btangmu:
The test still fails, including:

```
build: tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java#L1707
(TestPathHeader.java:1707) Error: am Characters Symbols3 has too many entries: 684
build: tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java#L1707
(TestPathHeader.java:1707) Error: ar Date & Time Fields has too many entries: 705
build: tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPathHeader.java#L1707
(TestPathHeader.java:1707) Error: ar Units Volume has too many entries: 740
```

@macchiati should I keep dividing pages until the test passes with the max being 600? Anyway I was about to start on Volume/Volume2

macchiati:
I think we identified the pages that we wanted to address. At this point we probably want to set the limit a bit higher so we just knock down the bad cases. 

For volume, I recommend that we split by the system. Volume Metric and Volume Other. 

Some notes:

From any unit you can get its system, and check it.

make a static final UnicodeSystems METRIC = Sets.of(UnitSystem.metric, UnitSystem.metric_adjacent);

extract the unit from the path
this is the long unit ID, which has an extra prefix (historic)

get a UnitConverter from the SupplementalDataInfo
use getShortId to convert to the short form
use getSourceToSystems to get a map, then use the short ID with it to get the system set
if the set of systems intersect with the constant metric, then put it in the first batch.
Collections.disjoint can be used to test intersection

btangmu:
> Volume Metric and Volume Other

OK! I was in the middle of another approach, trying to split about halfway down starting Volume2 with "Gallon" but what you suggest sounds more meaningful

macchiati:
@btangmu Please integrate this in, setting the error size limit to a value that works, and incorporating into your work. You can reset the error size limit down after further work.

btangmu:
@macchiati This was the basis for #3563 and #3573 -- now redundant; OK to close?

PR #3520, opened 2024-02-24
CLDR-17399 Migrate from deprecated boxed primitive constructors

mihnita:
CLDR-17399

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/json/JSONArray.java
+++ b/tools/cldr-apps/src/main/java/org/json/JSONArray.java
@@ -124,7 +124,7 @@ public class JSONArray {
                 case ']':
                 case ')':
                     if (q != c) {
-                        throw x.syntaxError("Expected a '" + new Character(q) + "'");
+                        throw x.syntaxError("Expected a '" + q + "'");
                     }
                     return;
                 default:
@@ -542,7 +542,7 @@ public class JSONArray {
      * @return this.
      */
     public JSONArray put(double value) throws JSONException {
-        Double d = new Double(value);
+        Double d = value;
         JSONObject.testValidity(d);
         put(d);
         return this;
@@ -555,7 +555,7 @@ public class JSONArray {
      * @return this.
      */
     public JSONArray put(int value) {
-        put(new Integer(value));
+        put(value);
         return this;
     }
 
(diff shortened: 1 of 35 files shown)


macchiati:
Were these changes generated by a refactoring tool or ide?

srl295 (dismissed):

srl295:
org.json is external but has already diverged

srl295:
needs mvn spotless:apply

mihnita:
> Were these changes generated by a refactoring tool or ide?

I wish  :-)

Detected by erroprone. I added a profile for maven so that we can run from command line.
(not a friendly report though)

I use the report as a guideline, but did the fixes in IntelliJ, with assistance.
Could have accepted the fixes "as is", but they were not always the best option.

For example:
```
Integer foo = new Integer(123);
```
Proposed fix:
```
Integer foo = Integer.valueOf(123);
```

But that only introduced another errorprone issue (not reported if you limit yourself to test for a one issue type only).
What I did in the end was:
```
Integer foo = 123;
```

TLDR: IDE assisted

mihnita:
> org.json is external but has already diverged
I will reconcile.

> needs mvn spotless:apply
Done. I didn't know about it, I don't touch CLDR too much (at all :-).

Thanks,
M

markusicu (approved):
lgtm
I don't know CLDR policy on org.json code. I am surprised that CLDR has "vendored" code in its repo, I thought such code was to be replaced by Maven dependencies.
Steven seems to indicate that it's ok to change this code anyway.

tools/cldr-apps/src/main/java/org/json/JSONArray.java
```diff
@@ -555,7 +555,7 @@ public JSONArray put(double value) throws JSONException {
      * @return this.
      */
     public JSONArray put(int value) {
-        put(new Integer(value));
+        put(value);
```
srl295, line 558: @mihnita BTW, postscript, this is a stack overflow (recursive)

tools/cldr-apps/src/main/java/org/json/JSONArray.java
```diff
@@ -566,7 +566,7 @@ public JSONArray put(int value) {
      * @return this.
      */
     public JSONArray put(long value) {
-        put(new Long(value));
+        put(value);
```
srl295, line 569: @mihnita BTW, postscript, this is a stack overflow (recursive)

PR #3522, opened 2024-02-26
CLDR-17407 check for anomalies and fix

macchiati:
CLDR-17407

Regenerated the coverage and charts while doing this.

common/properties/coverageLevels.txt
- Toki Pona dropped from Basic

- modified because the previous method for skipping beaufort was not working

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
- Made the CODE_FALLBACK not include deprecated timezones (eg Currie). Guts are in SupplementalDataInfo

- While doing this, added some internal structure to make it easier to get BCP47 data. The old mechanism, searching for pairs is awkward to use.
- Also added a backmapping from the aliases (eg long TZ names) to the subtype (eg the short TZ names)
- goodTimezones takes the available codes, and subtracts those that are deprecated, by getting their shortIds, then looking up in the BCP47 info.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
- Dropped Nigerian Pidgin and Maori targets to moderate for v45 (they were really bad). Will let TC know about this change, so that they can be bumped up again for v46 if they are really targets.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestStandardCodes.java
- added quick test

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/properties/coverageLevels.txt
+++ b/common/properties/coverageLevels.txt
@@ -144,7 +144,6 @@ th ;	modern ;	Thai
 ti ;	basic ;	Tigrinya
 tk ;	modern ;	Turkmen
 to ;	basic ;	Tongan
-tok ;	basic ;	Toki Pona
 tr ;	modern ;	Turkish
 tt ;	basic ;	Tatar
 ug ;	basic ;	Uyghur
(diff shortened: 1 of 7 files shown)


macchiati:
Adding @AEApple & @srl295 because of the target change and Toki Pona drop

macchiati:
Resetting to Draft until I figure out why the build isn't working.

srl295:
for `pcm` there was an issue with one of the vetters using a different orthography.

macchiati:
This still doesn't handle beaufort properly, but I'd like to merge this and then have a followup.

pedberg-icu (approved):

PR #3525, opened 2024-02-26
CLDR-17110 update LICENSE SPDX, SPDX in other files

srl295:
- does not regenerate dtds or xml - that will be done in future LDMLModify steps.
- the person names generator had copy and pasted static text instead of using the correct generator
- https://github.com/unicode-org/.github/issues/15

CLDR-17110

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/tool/GeneratePersonNameTestData.java
```diff
@@ -244,12 +245,8 @@ public static void main(String[] args) {
                 try (TempPrintWriter output2 =
                         TempPrintWriter.openUTF8Writer(dir.toString(), locale + ".txt"); ) {
                     output2.write(
-                            "# Test data for unit conversions"
```
srl295, line 247: oops

pedberg-icu (approved):

PR #3526, opened 2024-02-26
CLDR-11155 Separate Volume_Metric and Volume_Other

btangmu:
-New subroutines makePathHeader and getVolumePageId

-Change example in Dashboard.java from Volume to Area

-Comments

CLDR-11155

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
@@ -231,10 +231,8 @@ public class PathHeader implements Comparable<PathHeader> {
         Graphics(SectionId.Units),
         Length(SectionId.Units),
         Area(SectionId.Units),
-        Volume(SectionId.Units),
-        // TODO: enable splitting Volume into Volume/Volume2
-        // Reference: https://unicode-org.atlassian.net/browse/CLDR-11155
-        // Volume2(SectionId.Units),
+        Volume_Metric(SectionId.Units, "Volume Metric"),
+        Volume_Other(SectionId.Units, "Volume Other"),
         SpeedAcceleration(SectionId.Units, "Speed and Acceleration"),
         MassWeight(SectionId.Units, "Mass and Weight"),
         EnergyPower(SectionId.Units, "Energy and Power"),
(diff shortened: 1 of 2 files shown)


btangmu:
This works except that some paths that should be in Volume_Metric are in Volume_Other, such as:

//ldml/units/unitLength[@type="narrow"]/unit[@type="volume-cubic-kilometer"]/displayName

I don't know why, except that unitSystemCollection is empty in some such cases...

macchiati:
I directed you to the wrong call. It should be the following, which
supplies systems for compound units.

    *public* Set<UnitSystem> getSystemsEnum(String unit)

btangmu:
@macchiati with getSystemsEnum it appears to work perfectly

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -766,6 +751,56 @@ public PathHeader fromPath(final String path, List<String> failures) {
...
+                Set.of(UnitConverter.UnitSystem.metric, UnitConverter.UnitSystem.metric_adjacent);
+
+        private PageId getVolumePageId(String path) {
+            // Extract the unit from the path. For example, if path is
+            // //ldml/units/unitLength[@type="narrow"]/unit[@type="volume-cubic-kilometer"]/displayName
+            // then extract "volume-cubic-kilometer" which is the long unit id
+            final String longUnitId =
+                    XPathParts.getFrozenInstance(path).findAttributeValue("unit", "type");
+            if (longUnitId == null) {
+                throw new RuntimeException("Missing unit in path " + path);
```
macchiati, line 792: We shouldn't throw very generic exceptions, because it makes debugging trickier. If you touch this again, it would be better to have a more specific exception.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
```diff
@@ -766,6 +751,56 @@ public PathHeader fromPath(final String path, List<String> failures) {
             }
         }
 
+        private PathHeader makePathHeader(RawData data, String path, String alt) {
+            // Caution: each call to PathHeader.Factory.fix changes the value of
+            // PathHeader.Factory.order
+            SectionId newSectionId = SectionId.forString(fix(data.section, 0));
+            String pageIdName = fix(data.page, 0);
+            PageId newPageId;
+            if ("Volume".equals(pageIdName)) {
```
macchiati, line 760: Ideally this should be done inside of the fix.

macchiati (approved):
Approved (some minor improvements could be made, but not blockers)

PR #3233, opened 2023-08-29
CLDR-16531 BRS: GenerateValidityXml

srl295:
- Zanb moved to special

CLDR-16531

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3531, opened 2024-02-27
CLDR-17416 v45 BRS validity

srl295:
- also note we are starting to see spdx changes from CLDR-17110

CLDR-17416

ALLOW_MANY_COMMITS=true


macchiati:
Tests failing

common/validity/script.xml
```diff
@@ -35,12 +35,12 @@
 			Vaii Vith
 			Wara Wcho
 			Xpeo Xsux
-			Yezi Yiii Zanb
```
srl295, line 38: what's going on with `Zanb`? @markusicu @macchiati

srl295: We've been here before.. https://github.com/unicode-org/cldr/pull/3233#issuecomment-1701848571 

blocked by CLDR-17056  - I'll back out `Zanb` for now.

common/validity/script.xml
```diff
@@ -35,7 +35,7 @@
...
+			Yezi Yiii Zanb <!-- Note: for Zanb see CLDR-17056 -->
```
srl295, line 38: ^ so that we see this the next time the generator is run

macchiati (changes requested):
I decided it was time to fix the tooling. I figured that rather than have the tooling be a separate PR, it was better to do it all at once. So could you review https://github.com/unicode-org/cldr/pull/3537 and if that's ok, merge it and close this PR?

PR #3532, opened 2024-02-27
CLDR-17417 v45 BRS: algorithmic locales

srl295:
CLDR-17417

I normally run CLDRModify `-fp` but that is disabled per #3502 - anything else to do here?

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/annotations/sr_Latn.xml
+++ b/common/annotations/sr_Latn.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2023 Unicode, Inc.
+<!-- Copyright © 1991-2024 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
-SPDX-License-Identifier: Unicode-DFS-2016
+SPDX-License-Identifier: Unicode-3.0
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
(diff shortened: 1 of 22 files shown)


common/annotations/sr_Latn.xml
```diff
@@ -1426,7 +1426,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<annotation cp="🧑‍🍼" type="tts">osoba hrani bebu</annotation>
 		<annotation cp="👼">anđeo | bajka | beba | lice | mašta</annotation>
 		<annotation cp="👼" type="tts">beba anđeo</annotation>
-		<annotation cp="🎅">bajka | božić | božić bata | Deda Mraz | proslava</annotation>
+		<annotation cp="🎅">bajka | božić | božić bata | deda mraz | Deda Mraz | proslava</annotation>
```
macchiati, line 1429: I think you also want to run -fP, which should remove the case duplicates here.

macchiati (changes requested):

common/annotations/sr_Latn.xml
srl295: `-fP` did not remove the case duplicates here.

macchiati: Hmmm. I'll file a ticket for that.

macchiati (approved):

common/annotations/sr_Latn.xml
pedberg-icu: FYI, a later CLDRModify pass did remove the case dup here

PR #3534, opened 2024-02-27
CLDR-17336 fix parent locales for migration

macchiati:
CLDR-17336

As per discussion in infrastructure team, we need a migration path. This adds a new attribute with the rule instead of using the locale attribute, which reverts to only having locales. The locales from last release are restored, for this version.

ALLOW_MANY_COMMITS=true


macchiati:
@srl295 can you review this so that I can get it merged, unblocking integrations?

srl295 (approved):
LGTM. This data should be noted as going away in the future though?

PR #3536, opened 2024-02-28
CLDR-17339 Fix problem in ChartDeltaDtds that was skipping keyboard3

macchiati:
CLDR-17339

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartDtdDelta.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartDtdDelta.java
@@ -152,7 +152,9 @@ public class ChartDtdDelta extends Chart {
                 String firstVersion = type.firstVersion; // FIRST_VERSION.get(type);
                 if (firstVersion != null
                         && current != null
-                        && current.compareTo(firstVersion) < 0) {
+                        && VersionInfo.getInstance(current)
+                                        .compareTo(VersionInfo.getInstance(firstVersion))
+                                < 0) {
                     // skip if current is too old to have “type”
                     continue;
                 }
@@ -174,7 +176,11 @@ public class ChartDtdDelta extends Chart {
                     continue;
                 }
                 DtdData dtdLast = null;
-                if (last != null && (firstVersion == null || last.compareTo(firstVersion) >= 0)) {
+                if (last != null
+                        && (firstVersion == null
+                                || VersionInfo.getInstance(last)
+                                                .compareTo(VersionInfo.getInstance(firstVersion))
+                                        >= 0)) {
                     // only read if last isn’t too old to have “type”
                     dtdLast = DtdData.getInstance(type, last);
                 }


macchiati:
Question for you, @srl295 . There was a keyboard3 in v44. The change here fixes a bug, but also only goes back to v45 (eg no comparison to v44). That's ok if we want to forget about the v44 dtd.

If we want to retain the comparison we would change the following to have 44.
```
    keyboard3("keyboards/dtd/ldmlKeyboard3.dtd", "45.0", null, "../keyboards/3.0"),
    keyboardTest3("keyboards/dtd/ldmlKeyboardTest3.dtd", "45.0", null, "../keyboards/test");
```
Second question is: we are sometimes calling this Keyboard 2.0 (since it is a breaking change), but we have keyboard3. Should we publicly call it Keyboard 3.0?

srl295:
> Question for you, @srl295 . There was a keyboard3 in v44. The change here fixes a bug, but also only goes back to v45 (eg no comparison to v44). That's ok if we want to forget about the v44 dtd.

I don't think there's any value to comparing to v44 here.

> If we want to retain the comparison we would change the following to have 44.
> 
> ```
>     keyboard3("keyboards/dtd/ldmlKeyboard3.dtd", "45.0", null, "../keyboards/3.0"),
>     keyboardTest3("keyboards/dtd/ldmlKeyboardTest3.dtd", "45.0", null, "../keyboards/test");
> ```
> 
> Second question is: we are sometimes calling this Keyboard 2.0 (since it is a breaking change),

I don't think it's being called 2.0 now.

> but we have keyboard3. Should we publicly call it Keyboard 3.0?

It's somewhat of the third revision, with 2.0 being CLDR-10926 (2018).

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartDtdDelta.java
```diff
@@ -152,7 +152,9 @@ public void writeContents(FormattedFileWriter pw) throws IOException {
                 String firstVersion = type.firstVersion; // FIRST_VERSION.get(type);
                 if (firstVersion != null
                         && current != null
-                        && current.compareTo(firstVersion) < 0) {
+                        && VersionInfo.getInstance(current)
+                                        .compareTo(VersionInfo.getInstance(firstVersion))
+                                < 0) {
```
srl295, line 157: good catch, needed to be a version compare not a lexical compare

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartDtdDelta.java
```diff
@@ -174,7 +176,11 @@ public void writeContents(FormattedFileWriter pw) throws IOException {
                     continue;
                 }
                 DtdData dtdLast = null;
-                if (last != null && (firstVersion == null || last.compareTo(firstVersion) >= 0)) {
+                if (last != null
+                        && (firstVersion == null
+                                || VersionInfo.getInstance(last)
+                                                .compareTo(VersionInfo.getInstance(firstVersion))
+                                        >= 0)) {
```
srl295, line 183: same

srl295 (approved):

macchiati:
Ok then this should work

PR #3549, opened 2024-03-03
Create SECURITY.md

Yehiaelmatrwy2023:
CLDR-_____

ALLOW_MANY_COMMITS=true


srl295:
Thanks for signing the CLA, but this is a matter for the TC to discuss instead. It looks like a template file was uploaded.

PR #3552, opened 2024-03-04
CLDR-17322 Fix en_001 & en_CA narrow standalone day periods

AEApple:
CLDR-17322

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_CA.xml
+++ b/common/main/en_CA.xml
@@ -1946,7 +1946,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						</dayPeriodWidth>
 						<dayPeriodWidth type="narrow">
 							<dayPeriod type="midnight">mid</dayPeriod>
-							<dayPeriod type="am">a.m.</dayPeriod>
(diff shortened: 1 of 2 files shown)


common/main/en_CA.xml
```diff
@@ -1948,7 +1948,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 							<dayPeriod type="midnight">mid</dayPeriod>
 							<dayPeriod type="am">a.m.</dayPeriod>
 							<dayPeriod type="noon">↑↑↑</dayPeriod>
-							<dayPeriod type="pm">pm</dayPeriod>
+							<dayPeriod type="pm">p.m.</dayPeriod>
```
stenshamn, line 1951: Considering the narrow format is without periods, shouldn't we follow that here and change a.m. to am instead?

AEApple: Sure, I'll update it to be consistent that way.

stenshamn (approved):

macchiati (approved):

PR #3556, opened 2024-03-06
CLDR-17389 Revert Dutch time zone names

DraganBesevic:
CLDR-17389 

CLDR 44 introduced the changes in time zone names for Dutch, mainly in America's region. Those changes implemented English time zone name translations, instead of the Dutch ones.

Per the discussion in TC, it is agreed to revert the time zone names back to pre-44 values

ALLOW_MANY_COMMITS=true


stenshamn (approved):
Looked at the new file and didn't see any English references, except GMT which I think is expected.

PR #3471, opened 2024-01-30
CLDR-17274 XCG translations

AEApple:
CLDR-17274

XCG translations for Curaçao and Sint Maarten's new currency

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3561, opened 2024-03-11
CLDR-17435 add timezone attributes for the 2024 currency changes in CW,SX

pedberg-icu:
CLDR-17435

Add the timezone attributes for the 2024 currency changes in CA and SX. Per https://unicode-org.atlassian.net/browse/CLDR-17282 we decided to add the timezone attributes for all currency changes from 2022-01-01 and later. I should have mentioned that when reviewing https://github.com/unicode-org/cldr/pull/3471 but did not, so we add the attributes with this PR.

ALLOW_MANY_COMMITS=true


macchiati (approved):

srl295 (commented):
post merge LGTM

PR #3562, opened 2024-03-12
CLDR-17420 Fix UnitConverter.parseUnitId for special, add tests & regen unit test files

pedberg-icu:
CLDR-17420

This is a followup to PR [#3487](https://github.com/unicode-org/cldr/pull/3487) for CLDR-16329

The main thing I had to do here was fix UnitConverter.parseUnitId() to handle special mappings (which also eliminated 3 TODOs in UnitConverter.java). I also added some beaufort cases to TestUnits.TestSpecialCases() and regenerated the units test data files.

ALLOW_MANY_COMMITS=true


pedberg-icu:
@macchiati Hi Mark, would you be able to do a review of this? (enhancements to CLDR UnitConverter & unit tests for beaufort).

macchiati (approved):

PR #3564, opened 2024-03-12
CLDR-17202 kbd: updates to bengali

srl295:
- move kmc from alpha to beta
- use a later kmc version that support current api (this will change again)
- fix a typo in the sample data

CLDR-17202

ALLOW_MANY_COMMITS=true


srl295:
the keyboard check workflow will pass again with this commit, because it bumps to a version of keyman which supports the alpha level of ldml

pedberg-icu (approved):

PR #3567, opened 2024-03-13
CLDR-17399 Fix infinite loops due to removing boxed primitive constructors

btangmu:
-JSONObject had recursive calls in put double/int/long

-Use Double/Integer/Long.valueOf to call the version of put with Object parameter

CLDR-17399

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/json/JSONObject.java
+++ b/tools/cldr-apps/src/main/java/org/json/JSONObject.java
@@ -897,7 +897,7 @@ public class JSONObject {
      * @throws JSONException If the key is null or if the number is invalid.
      */
     public JSONObject put(String key, double value) throws JSONException {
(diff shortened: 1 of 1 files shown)


macchiati (approved):

markusicu:
@mihnita FYI

mihnita:
Sorry, the change was a bit big and I didn't see it.
Thank you for catching and fixing this,
Mihai

btangmu:
@mihnita Thank you for improving the code! There's always some risk in a big modernization like this, and modernization is necessary. The problem here wasn't obvious. The fact that we didn't detect it until deployment on the production server (fortunately when it wasn't open for vetting) reveals a weakness in the automated testing, and we're going to study how to improve that. Cheers!

PR #3570, opened 2024-03-15
CLDR-4152 show alternate locale in info panel

srl295:
CLDR-4152

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/InfoPanel.vue
+++ b/tools/cldr-apps/js/src/views/InfoPanel.vue
@@ -1,5 +1,14 @@
 <template>
   <div>
+    <div v-if="altText" class="secondary">
+      <i>{{ altLang }}: </i
+      ><cldr-value @lang="altLang"> {{ altText }} </cldr-value>
+    </div>
+    <div>
+      <a-input placeholder="fr" v-model:value="altLang">
+        <template #prefix> Other Lang: </template>
+      </a-input>
+    </div>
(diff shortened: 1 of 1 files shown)


srl295:
https://github.com/unicode-org/cldr/assets/855219/3b43773d-bf2d-40e7-8f9d-95752b74ce0f

srl295:
maybe we should have a switch, 'CLDR Labs' that lets you try out new experimental features.

DavidLRowe:
Just dreaming, but I think that I'd find it more useful to have a control (button or whatever) that swapped the values in the column from English to the selected reference language (French or Spanish or whatever). Then I could see (at the same time) how all the months were translated from English to the reference language, rather than one by one.

srl295:
> Just dreaming, but I think that I'd find it more useful to have a control (button or whatever) that swapped the values in the column from English to the selected reference language (French or Spanish or whatever). Then I could see (at the same time) how all the months were translated from English to the reference language, rather than one by one.

That's a good point and a good reason to add an additional column, thanks.  I added it this way because it was easier to prototype- the general edit view is still not 'modernized'. 

I added a comment on the issue https://unicode-org.atlassian.net/browse/CLDR-4152?focusedCommentId=173802 which notes that only the English is fully vetted.  So, an additional column, or another row in the 'English' column might make sense, but I don't think it would be a good idea to actually hide the English.

PR #3572, opened 2024-03-20
CLDR-17425 Fix bugs involving locale USER

btangmu:
-Include real locale ID as json.loc in server response for page request

-Use optional chaining (?.) operator for special_type which may be undefined for USER

-Do not call dispatchEvent or setRef from setCurrentLocale for locale USER

-New method UserRegistry.substituteUserWildcardLocale replacing inline code in SurveyAjax, now called also by VoteAPIHelper

CLDR-17425

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -1055,21 +1055,7 @@ public class SurveyAjax extends HttpServlet {
          * loc can have either - or _; normalize to _
          */
         loc = loc.replace('-', '_');
-        /*
-         * Support "USER" as a "wildcard" for locale name, replacing it with a locale suitable for the
-         * current user, or "fr" (French) as a fallback.
-         */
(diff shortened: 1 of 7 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1055,21 +1055,7 @@ public static CLDRLocale validateLocale(PrintWriter out, String loc, String sess
          * loc can have either - or _; normalize to _
          */
         loc = loc.replace('-', '_');
```
srl295, line 1057: do we need this line ? or maybe it should also go into a utility function?

srl295 (approved):
thanks.. nice !

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
btangmu: I think we have to assume it's needed... ideally it should be encapsulated in a module specific to locales, not SurveyAjax; and as much as possible we should pass around CLDRLocale objects instead of strings...

PR #3576, opened 2024-03-26
CLDR-15954 Update spec for unit preferences

macchiati:
CLDR-15954

Clarify the algorithm for unit preferences to catch edge cases.

1. What to do with unitpreferences if there is no quantity for a unit, or no preferences for a quantity. Emphasizes the use of likelysubtags. It more explicit about edge cases when comparing input values to geq values (negative numbers, infinity, NaN). Provides more of a step-by-step approach to handling the subtags that have an impact (no change in effect, but much clearer). Moves examples to better places.
2. Edge cases (but important ones) for mixed units: handling negatives and rounding. One change is to be less prescriptive about the way that precision is handled, giving us a chance to work that our more in the future (because the current advice really doesn't work well.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-info.md
+++ b/docs/ldml/tr35-info.md
@@ -1132,28 +1132,71 @@ Different locales have different preferences for which unit or combination of un
 
 ### <a name="Unit_Preferences_Overrides" href="#Unit_Preferences_Overrides">Unit Preferences Overrides</a>
 
-The determination of preferred units depends on the locale identifer: the keys mu, ms, rg, their values, the base locale (language, script, region) and the user preferences data.
+The determination of preferred units uses the user preference data together with **input unit**, the **input usage**, and the **input locale identifer**.
+Within the locale identifier, the subtags that can affect the result are:
+  * the value so the keys mu, ms, rg
+  * the region in the locale identifier (if there is one)
+  * and otherwise the likely region subtag for the locale identifier
 
-The strongest is the mu key, then the ms key, then the rg key. Beyond that the region of the locale identifer is used, and if not present, the likely-subtag region. For example:
+The strongest priority is the mu key, then the ms key, then the rg key.
+Beyond that the region of the locale identifer is used, and if not present, the likely-subtag region.
+For example:
 
 |   | Locale                                | Result     | Comment                                                            |
 |---|---------------------------------------|------------|--------------------------------------------------------------------|
 | 1 | en-u-rg-uszzzz-ms-ussystem-mu-celsius | Celsius    | despite the rg and ms settings for US, and the likely region of US |
 | 2 | en-u-rg-uszzzz-ms-metric              | Celsius    | despite the rg setting for US, and the likely region of US         |
-| 3 | en-u-rg-dezzzz.                      | Celsius    | despite the likely region of US                                    |
-| 4 | en                                    | Fahrenheit | because the likely region for en with no region is US              |
+| 3 | en-u-rg-dezzzz.                       | Celsius    | despite the likely region of US                                    |
+| 4 | en-DE                                 | Celsius    | because explicit region is DE                                      |
+| 5 | en                                    | Fahrenheit | because the likely region for en with no region is US              |
 
-The **ms** value is used in the following way.
+If any of the above are invalid, then they are ignored. Thus the following constructs are ignored: 
 
-1. Find the corresponding Key-Value row in the table below.
-2. Get the unit preferences for the **locale**, **category**, and **usage**.
-3. If any of the units in that set have a measurement system that doesn’t match the -u-ms- value, get unit preferences again, but using the fallback region instead of the locale's region.
+  * -me-smoot // invalid unit
+  * -ms-stanford // invalid unit system
+  * -rg-aazzzz // invalid region 'AA'. 
+      * Only the region portion is currently used, so in -rg-usabcdef the "abcdef" is ignored, whether or not it is valid.
+  * -AA // invalid region 'AA'
+
+The following algorithym is used to return the **output unit**.
+
+#### Computing override units
+If there is a valid -mu value then let the **output unit** be the that value, and return it.
+This terminates the algorithm.
+
+#### Compute the regions
+If there is no valid -mu value, the following steps are used to determine a region R from the **input locale identifer**.
+(and optionally a Unit Systems Match (USM)):
+
+1. If there is a valid -ms value then let USM  be the corresponding value in column 2 of the table below. Otherwise FR is not used. In either case continue with step 2.
+2. If there is a valid -rg region, let R be that region and go to Compute the category.
+3. If there is a valid region in the locale, let R be that region go to Compute the category.
+4. Otherwise, compute the likely subtags for the locale. 
+     1. If there is a likely region, then let R be that region and go to Compute the category.
+	 2. Otherwise, let R be 001 and go to Compute the category
 
 | Key-Value   | Unit Systems Match          | Fallback Region for Unit Preferences |
 |-------------|-----------------------------|--------------------------------------|
 | ms-metric   | metric OR metric_adjacent   | 001                                  |
 | ms-ussystem | ussystem                    | US                                   |
 | ms-uksystem | uksystem                    | UK                                   |
+
+#### Compute the category
+
+A **category** is determined as follows from the input unit:
+
+1. From the input unit, use the conversion data in [baseUnit](tr35-info.html#Unit_Conversion) and let the **input base unit** be the baseUnit attribute value.
+    * eg, for `pound-force` the baseUnit is `kilogram-meter-per-square-second`.
+2. If there is no such base unit (such as for a an unusual unit like `ampere-pound-per-foot-square-minute`), convert the input unit to a combination of base units, reduce to lowest terms, and normalize. Let the **input base unit** be that value.
+    * eg, `ampere-pound-per-foot-square-minute` -> kilogram-ampere-per-meter-square-second
+3. If the **input base unit**, has a unitQuantity element, then let the **category** be the quantity attribute value
+    * eg, `force` from `<unitQuantity baseUnit='kilogram-meter-per-square-second' quantity='force'/>`
+3. If the result does not have a unitQuantity, let the output unit be the input base unit or an equivalent metric/SI unit, and return. This terminates the algorithm.
+    * eg, for `ampere-pound-per-foot-square-minute` return `kilogram-ampere-per-meter-square-second` or `pascal-ampere`
+	* That is, an implementation can use shorter units as long as long as the combination is equivalent in value.
+
+
+#### Examples
 
 **Example A: xx-SE-u-ms-metric, length, road**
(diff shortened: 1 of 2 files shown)


macchiati:
Still in draft, but if you have any feedback...

docs/ldml/tr35-info.md
```diff
@@ -1116,87 +1116,110 @@ The examples in #4 are due to the following ordering of the `unitQuantity` eleme
...
-As an example, ICU only uses the unit preferences (with rg, ms, and/or mu and the likely region) in formatting units when a usage parameter is set.
+1. From the input unit, use the conversion data in [baseUnit](tr35-info.html#Unit_Conversion) and let the **input base unit** be the baseUnit attribute value.
+    * eg, for `pound-force` the baseUnit is `kilogram-meter-per-square-second`.
+2. If there is no such base unit (such as for a an unusual unit like `ampere-pound-per-foot-square-minute`), convert the input unit to a combination of base units, reduce to lowest terms, and normalize. Let the **input base unit** be that value.
+    * eg, `ampere-pound-per-foot-square-minute` -> `kilogram-ampere-per-meter-square-second`
+3. If the **input base unit** has a unitQuantity element, then let the **category** be the quantity attribute value.
+    * eg, `force` from `<unitQuantity baseUnit='kilogram-meter-per-square-second' quantity='force'/>`
+3. If the **input base unit** does not have a unitQuantity, let the output unit be the input base unit or an equivalent metric/SI unit, and return. This terminates the algorithm; there is no need to use the unit preferences information.
```
younies, line 1214: I didn't understand this part ("an equivalent metric/SI unit"). How can we find the equivalent metric/SI unit?

We have a method to convert any unit to the base unit, but is there no way to convert to metric/SI units?

docs/ldml/tr35-info.md
```diff
@@ -1116,87 +1116,110 @@ The examples in #4 are due to the following ordering of the `unitQuantity` eleme
...
 
-Mixed units are expected to be rendered in the order of the tokens in the identifier. For example, the value 1.25 with the identifier "foot-and-inch" should be rendered as "1 foot and 3 inches" and 1.25 inch-and-foot should be rendered as “3 inches and 1 foot". **NOTE:** the correct application of this may require adding locales to the regions attribute set.
+Mixed unit identifiers should be from largest to smallest (eg foot-and-inch instead of inch-and-foot). 
+If it turns out that some locales present certain mixed units in a different order, additional structure will be needed in CLDR.
+
+If a maximum/minimum number of decimals or signficant digits is desired in formatting, 
+those values should should be adapted to the mixed radix used, starting with the largest unit, 
+and discarding smaller units where they would be overly accurate. 
+For example, if the goal is 1 part in 100, then "6 miles 1 foot" is overly precise, 
+while "6 foot 1 inch" has approximately the right accuracy.
```
sffc, line 1133: I think it would be much clearer and easier to implement if the decimal formatting options were applied to the smallest unit, with larger units always displayed as an integer.

This appears to be what you write farther down in the doc:

> * If the unit is mixed (eg foot-and-inch) the skeleton applies to the final subunit; the higher subunits are formatted as integers.

Also consider the case "1.9999999... ft" formatted with foot-and-inch. This should result in "2 feet", not "1 foot 12 inches".

docs/ldml/tr35-info.md
```diff
@@ -1116,87 +1116,110 @@ The examples in #4 are due to the following ordering of the `unitQuantity` eleme
...
 ## <a name="Unit_Preferences" href="#Unit_Preferences">Unit Preferences</a>
 
 Different locales have different preferences for which unit or combination of units is used for a particular usage, such as measuring a person’s height. This is more fine-grained than merely a preference for metric versus US or UK measurement systems. For example, one locale may use meters alone, while another may use centimeters alone or a combination of meters and centimeters; a third may use inches alone, or (informally) a combination of feet and inches.
 
 ### <a name="Unit_Preferences_Overrides" href="#Unit_Preferences_Overrides">Unit Preferences Overrides</a>
 
-The determination of preferred units depends on the locale identifer: the keys mu, ms, rg, their values, the base locale (language, script, region) and the user preferences data.
+The determination of preferred units uses the user preference data together with **input unit**, the **input usage**, and the **input locale identifer**.
+Within the locale identifier, the subtags that can affect the result are:
+  * the value of the keys mu, ms, and rg
```
sffc, line 1152: sd?

docs/ldml/tr35-info.md
```diff
@@ -1257,75 +1283,100 @@ The above information says that for default usage, in the US people use mile, fo
...
+  * ≥ 30.48 meters ⇒ foot-and-inch
+
+1. Search for the first matching unitPreference for the absolute value of the input measure. If there is no match (eg < 100 feet in the above example), take the last unitPreference. That is, the last unitPreference is effectively geq="0". In the above example, `<unitPreference regions="GB">yard</unitPreference>` is equivalent to `<unitPreference geq="0" regions="GB">yard</unitPreference>`
 
 For completeness, when comparing doubles to the geq values:
-* Negative numbers are treated as if they were positive.
-* _infinity_ is treated as being the largest possible value.
-* NaN is treated as the smallest possible value.
+* Negative numbers are treated as if they were positive, so in the above example -804.672 meters will format as "-0.5 mile".
+* _infinity_, NaN, and -_infinity_ match the largest possible value. Thus -∞ meters will format as "-∞ miles", not "-∞ yards".
```
sffc, line 1323: How are negative mixed units formatted? Probably similar to how we decided to format negative durations?

macchiati: Good point. I think the first unit should carry the negative sign.

-3 miles 150 feet
-3 hours 10 minutes
...

docs/ldml/tr35-info.md
macchiati: First, when I wrote "while "6 foot 1 inch" has approximately the right accuracy." I had flubbed the example. So disregard that bit.

> I think it would be much clearer and easier to implement if the decimal formatting options were applied to the smallest unit, with larger units always displayed as an integer.

I agree it is simpler for us to implement. And we're in agreement on the higher units always showing as integers.
But it gives less satisfactory results. Take the following example.

The programmer supplies: source unit is meters, and amount is 1616, and desired precision is 3 significant digits. 

For meters, that turns into "1.62 kilometers". That reflects internal values 1.62±0.005 km, or within 5 meters.

In miles, that turns into 1.004135847 decimal. If we take the floor*, we get 1 mile, with 21.83727034 feet remaining. If we apply 3 sig. digits to the last field, then we get "1 mile 21.8 feet". That accuracy is excessive: 21.8±0.05 feet is within 0.015 meters. What we really want in that case is something that reflects the original accuracy: a rounded number that is within 5 meters of the original amount (=~16 feet), which is something like "1 mile 20 feet", not "1 mile 22 feet" or "1 mile 21.8 feet".

\* As you point out, if we get satisfactory precision out of a higher unit, then we don't want to include the lower unit at all. And if that is the case, we would want to use round(x) and not floor(x).

What I was trying to express with the "should" was more of a warning: (a) there is an issue with mixed units, and (b) we don't have a definitive algorithm (yet) for what to do.

docs/ldml/tr35-info.md
macchiati: What I didn't want to do is proscribe a particular method of producing a shorter metric unit.

docs/ldml/tr35-info.md
sffc: I understand the desired behavior, but a hand-wavy algorithm can't be implemented. I would rather offer a suggestion of a mechanism that can be revised later than try to hitchhike on an existing mechanism (maximum significant and fractional digits) which are already hard to implement correctly.

> Implementations may offer mechanisms to limit the precision of the formatted mixed unit. For example, "1.62143 meters" with three significant figures could be rendered in meter-and-millimeter as "1 meter, 620 millimeters". This mechanism should be distinct from decimal digit options, such as those provided in a number formatting skeleton, which should be applied to only the smallest unit.

macchiati: Yah, I don't like the handwaviness either. I'll try some wording, but because implementations and (and IMO should) have other mechanisms, we need to not **require** that the precision be applied to the final unit.

macchiati: I'm leery of having yet another precision type for mixed units. I do not think that peopel 

BTW, what I was thinking of having is something like the following (wording is currently clumsy). I'm not going to include this wording, because we don't have time to work out any kinks, but just to give you a sense of what I'm thinking.

1. The original source (say meters) is first converted to the topmost unit, eg miles.
2. The precision is applied to that topmost unit to find an "accuracy" value.
3. Eg, **12.345678 miles** with 3 sig. digits gives **12.3 miles**, so the accuracy is **±0.05 miles**. 
4. As you convert to successively smaller units, stop when the value could be rounded to be within that range.
5. The first unit as an integer is not within that accuracy, so we use the integer value **12 miles**, and the remainder to carry over is **0.345678 miles**
6. The carry-over to feet is **1825.17984 feet**, and accuracy is **±264**. Since the accuracy is > 1, we know we are going to stop at this value, rounding to some amount, and not go on to inches.
7. Rounding to 1 digit (2000 feet) would be within the accuracy, so we stop, and round to that value, producing "12 miles 2000 feet".

This is not by any means complete, just to give you a sense.

docs/ldml/tr35-info.md
macchiati: We don't currently have anything in the spec about using sd in order to get a hint about the region. Was that implemented in ICU or ICU4X?

docs/ldml/tr35-info.md
macchiati: Actually, there is an example right below this. But I'll change the wording to be clearer.

docs/ldml/tr35-info.md
sffc: I thought we started using `sd` in the `-u-fw` resolution algorithm.

In ICU4X, I wrote infrastructure to handle data keyed on `-u-sd`, but nothing in CLDR currently utilizes that code path. It seems like it might want to do so in the future; for example, it's easy to imagine Scottland using different units for something like road mileage than Wales.

macchiati: I think that might be the case. The spec provides for subdivision values in the unit preferences data, but as yet there is no such data. I'll file a ticket for adding that in v46 (we'll have to work out some of the details).

https://unicode-org.atlassian.net/browse/CLDR-17474

docs/ldml/tr35-info.md
macchiati: Just added this, in the Mixed Units section.

docs/ldml/tr35-info.md
macchiati: Please take a look at the current text now, @sffc and @younies .

macchiati:
We're going to branch soon, so I'd appreciate if we can get it approved today or tomorrow.

srl295 (approved):
Generally LGTM.  Seems to address comments.

PR #3581, opened 2024-03-28
CLDR-16998 Fix Formatting

macchiati:
CLDR-16998

Took the suggestion to use monospacing in two of the suggested sections. The https://unicode.org/reports/tr35/#Key_And_Type_Definitions_ was more complicated, and would benefit less so that wasn't done.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -3805,9 +3805,10 @@ _Example:_
 
 > source="ja-Latn-fonipa-hepburn-heploc"
 >
-> rule  =`\<languageAlias type="und-hepburn-heploc" replacement="und-alalc97">`
+> rule  =`<languageAlias type="und-hepburn-heploc" replacement="und-alalc97">`
 >
 > result="ja-Latn-alalc97-fonipa"
+>
 > (note that CLDR canonical order of variants is alphabetical)
 
 ##### Territory Exception
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -1661,9 +1661,9 @@ __Lateral Inheritance__ is where resources are inherited from within the same lo
 
 | Element @Attribute          | Source | Context |
 | ---------------- | ------ | ------- |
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35.md
```diff
@@ -3803,13 +3803,12 @@ A matching rule can be used to transform the source fields as follows
...
-> source=ja-Latn-fonipa-hepburn-heploc
+> source="ja-Latn-fonipa-hepburn-heploc"
 >
-> rule  ="\<languageAlias type="und-hepburn-heploc"
+> rule  =`\<languageAlias type="und-hepburn-heploc" replacement="und-alalc97">`
 >
-> replacement="und-alalc97">"
->
-> result="ja-Latn-alalc97-fonipa" // note that CLDR canonical order of variants is alphabetical
+> result="ja-Latn-alalc97-fonipa"
```
AEApple, line 3810: I thought source and result were also supposed to be enclosed?

macchiati: I just changed the items that are fragments of XML syntax; that is more consistent with the rest of the document. Adding a note to the ticket.

docs/ldml/tr35.md
```diff
@@ -3803,13 +3803,13 @@ A matching rule can be used to transform the source fields as follows
...
 _Example:_
 
-> source=ja-Latn-fonipa-hepburn-heploc
+> source="ja-Latn-fonipa-hepburn-heploc"
 >
-> rule  ="\<languageAlias type="und-hepburn-heploc"
+> rule  =`<languageAlias type="und-hepburn-heploc" replacement="und-alalc97">`
 >
-> replacement="und-alalc97">"
+> result="ja-Latn-alalc97-fonipa"
```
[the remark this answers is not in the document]
macchiati: I left the strings alone, and just changed the XML fragments (for more consistency with the rest of the document)

BTW, we definitely could use a scrub to fix the formatting overall, but that's a big-ticket item!

AEApple (approved):

PR #3583, opened 2024-03-28
CLDR-16900 Limit frequency of announce and completion requests

btangmu:
-This PR only changes the front end

-New module cldrSchedule with class FetchSchedule used by both cldrAnnounce and cldrProgress

-Old code in cldrProgress mistakenly had && for condition on time since last request/response

-New code in cldrSchedule has || instead of &&

CLDR-16900

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrProgress.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrProgress.mjs
@@ -7,12 +7,17 @@ import * as cldrAjax from "./cldrAjax.mjs";
 import * as cldrCoverage from "./cldrCoverage.mjs";
 import * as cldrGui from "./cldrGui.mjs";
 import * as cldrNotify from "./cldrNotify.mjs";
+import * as cldrSchedule from "./cldrSchedule.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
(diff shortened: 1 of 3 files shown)


btangmu:
A subsequent PR should cache the data on the back end, and also maybe use Etag

tools/cldr-apps/js/src/esm/cldrProgress.mjs
```diff
@@ -361,18 +367,10 @@ function fetchLocaleData(unlessLoaded) {
...
-    localeProgressStats &&
-    localeProgressStats.locale === locale
-  ) {
+  if (unlessLoaded && localeProgressStats?.locale === locale) {
     // LocaleMeter is already set
-    // Still refresh if it's been long enough since last request and last response
-    const now = Date.now();
-    if (
-      now < timeLastReceivedLocaleData + LOCALE_METER_REFRESH_MS &&
-      now < timeLastRequestedLocaleData + LOCALE_METER_REFRESH_MS
```
btangmu, line 374: It was my mistake to use && in this condition; should be || as it is in the new classe

srl295 (approved):

PR #3584, opened 2024-03-28
CLDR-17481 re-enable directly loading Test.html

srl295:
CLDR-17481

- run.sh not needed. Just npm run install build-test and then load Test.html

ALLOW_MANY_COMMITS=true


srl295 (commented):
there are other improvements that could be made, this is the easiest.

btangmu (approved):

PR #3596, opened 2024-04-02
CLDR-16900 Shorten server response if unchanged; bug fix, refactor

btangmu:
-Refactor (encapsulate) to calculate unread/total counts in only one place; new callbackSetCounts

-On front end, remember the last announcement ID fetched (alreadyGotId)

-On back end, respond with 304 Not Modified if alreadyGotId matches

-Do not report 304 Not Modified response as error; log it to console only if debugging enabled

-New front-end method resetSchedule prevents delay and prevents 304 response

-Format Test.html (not specific to this ticket; unrelated PR left it unformatted)

-Comments

CLDR-16900

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
@@ -25,7 +25,10 @@ const schedule = new cldrSchedule.FetchSchedule(
 let thePosts = null;
 
 let callbackSetData = null;
+let callbackSetCounts = null;
 let callbackSetUnread = null;
+
+let alreadyGotId = 0;
 
 /**
  * Get the number of unread announcements, to display in the main menu
@@ -36,28 +39,27 @@ async function getUnreadCount(setUnreadCount) {
   if (setUnreadCount) {
     callbackSetUnread = setUnreadCount;
   }
-  await refresh(callbackSetData);
+  await refresh(callbackSetData, callbackSetCounts);
 }
 
 /**
  * Refresh the Announcements page and/or unread count
  *
  * @param {Function} viewCallbackSetData the callback function for the Announcements page, or null
+ * @param {Function} viewCallbackSetCounts the callback function for the Announcements page, or null
  *
  * The callback function for setting the data may be null if the Announcements page isn't open and
  * we're only getting the number of unread announcements to display in the main header
  */
-async function refresh(viewCallbackSetData) {
+async function refresh(viewCallbackSetData, viewCallbackSetCounts) {
   if (DISABLE_ANNOUNCEMENTS) {
     return;
   }
   if (viewCallbackSetData) {
-    if (!callbackSetData) {
-      // The Announcements page was just opened, so re-fetch the data immediately regardless
-      // of how recently it was fetched previously (for unread count)
-      schedule.reset();
-    }
     callbackSetData = viewCallbackSetData;
+  }
+  if (viewCallbackSetCounts) {
+    callbackSetCounts = viewCallbackSetCounts;
   }
   if (!cldrStatus.getSurveyUser()) {
     if (viewCallbackSetData) {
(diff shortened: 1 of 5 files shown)


tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
```diff
@@ -119,6 +149,7 @@ async function compose(formState, viewCallbackComposeResult) {
 }
 
 async function saveCheckmark(checked, announcement) {
+  resetSchedule();
```
btangmu, line 152: this fixes a bug -- otherwise, the "unread" number by the balloon icon in the main menu took 60 seconds before it updated in response to adding a checkmark for "I have read this"

tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
```diff
@@ -36,29 +39,28 @@ async function getUnreadCount(setUnreadCount) {
...
-async function refresh(viewCallbackSetData) {
+async function refresh(viewCallbackSetData, viewCallbackSetCounts) {
   if (DISABLE_ANNOUNCEMENTS) {
     return;
   }
   if (viewCallbackSetData) {
-    if (!callbackSetData) {
-      // The Announcements page was just opened, so re-fetch the data immediately regardless
-      // of how recently it was fetched previously (for unread count)
-      schedule.reset();
```
btangmu, line 58: This schedule.reset isn't needed here anymore since now AnnouncePanel.vue "created" method calls cldrAnnounce.resetSchedule

tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
```diff
@@ -68,36 +70,64 @@ async function refresh(viewCallbackSetData) {
...
     .doFetch(url)
     .then(cldrAjax.handleFetchErrors)
     .then((r) => r.json())
     .then(setPosts)
-    .catch((e) => console.error(e));
+    .catch(handleErrorOrNotModified);
+}
+
+function handleErrorOrNotModified(e) {
+  if (e.message === "Not Modified") {
```
srl295, line 89:
```suggestion
  if (e.status === 304) {
```

should be able to check the status code directly - the message can be variable.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Announcements.java
```diff
@@ -71,6 +76,9 @@ public Response getAnnouncements(@HeaderParam(Auth.SESSION_HEADER) String sessio
         if (SurveyMain.isBusted() || !SurveyMain.wasInitCalled() || !SurveyMain.triedToStartUp()) {
             return STError.surveyNotQuiteReady();
         }
+        if (alreadyGotId != 0 && alreadyGotId == AnnouncementData.getMostRecentAnnouncementId()) {
+            return Response.notModified().build();
```
btangmu, line 80: I have some doubt whether this is a standard way to use notModified (304)...

Possible alternatives include (1) setting headers so that the browser delivers cached response for 304 (but that's maybe less efficient -- needless work processing the identical json); (2) return 200 and short json, e.g., just json.unchanged = true; (3) Etag...

srl295 (approved):
nice

tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
btangmu: That would require changes to cldrAjax -- currently cldrAjax.handleFetchErrors is as follows:

```
function handleFetchErrors(response) {
  if (!response.ok) {
    throw new Error(response.statusText);
  }
  return response;
}
```
The Error only contains response.statusText -- I'm not sure how to set e.status in a standard way

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Announcements.java
```diff
@@ -71,6 +76,9 @@ public Response getAnnouncements(@HeaderParam(Auth.SESSION_HEADER) String sessio
...
+        }
```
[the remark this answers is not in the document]
btangmu: good idea

if tests pass, maybe best to merge this as-is since it does include bug fix and performance improvement; but keep ticket open and consider protocol improvements under discussion

PR #3600, opened 2024-04-02
CLDR-17493 fix cldr-apps tests to skip if no database

srl295:
CLDR-17493

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):
Thanks!

PR #3604, opened 2024-04-02
CLDR-17251 Fixes to likely subtags spec

macchiati:
CLDR-17251

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

sffc (commented):
Ok

PR #3606, opened 2024-04-03
CLDR-16034 Check for pairing bidi markup chars (illegal); remove 2 (unpaired) RLO in ff_Adlm

pedberg-icu:
CLDR-16034

Add test to check for bidi markup chars that are intended to be used in pairs; these are illegal in CLDR values )only LRM, RLM, ALM are ok). Then run test to find any issues; remove two (unpaired) RLO in ff_Adlm.xml

ALLOW_MANY_COMMITS=true


fitzsim:
Thank you for fixing these.  An automated testing tool I am using (rpminspect) was flagging these characters in OpenJDK's import of ff_Adlm.xml, and I was just about to ask a Unicode CLDR maintainer about them.

While investigating this, I found two more RLO characters in common/annotationsDerived/ff_Adlm.xml, lines 93 and 94, which I suspect are also invalid:

https://github.com/unicode-org/cldr/blob/9de952836fa17ce717b840b8f02947fbfa6b95c9/common/annotationsDerived/ff_Adlm.xml#L93
https://github.com/unicode-org/cldr/blob/9de952836fa17ce717b840b8f02947fbfa6b95c9/common/annotationsDerived/ff_Adlm.xml#L94

PR #3616, opened 2024-04-05
CLDR-17519 Move English data of subdivisions to root

garywade:
Moved the contents of localeDisplayNames in en.xml of subdivisions to root.xml so all locales will have fallback data.

CLDR-17519

ALLOW_MANY_COMMITS=true


pedberg-icu:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/unicode-org/cldr?pullRequest=3616) before we can accept your contribution.

@garywade @DraganBesevic @richgillam Gary, Apple already has a CLA with Unicode that covers the contributions of all Apple employees, so do not sign an individual CLA. Your name just needs to be added to the list of Apple contributors. Dragan or Rich, I think you need to e-mail Steven or Markus to get Rich added to the list of people covered by the Apple CLA.

macchiati:
Moving the subdivision codes to root should be put on the Design group
agenda.

For other data we have tried to specifically make the root locale data has
neutral as possible so that translators don't simply use the English when
it's inappropriate.

macchiati:
I added to the design meeting agenda. We talked a bit about subdivisions at
this week's meeting, but with no conclusion; it would be good if you could
join a future meeting, Gary.

garywade:
> I added to the design meeting agenda. We talked a bit about subdivisions at this week's meeting, but with no conclusion; it would be good if you could join a future meeting, Gary.

@macchiati I didn't catch this early enough to join on Monday.  Would it help to join the Wednesday morning CLDR-TC meeting?

macchiati:
We won't discuss in the main meeting until we have a recommendation from
the design team. So the design meeting is the right one to attend. There's
no real urgency because the current release is closed for new additions.

macchiati:
I converted to Draft until the issue is resolved in the Design meeting. Let us know when you want to attend.

macchiati:
BTW, I reviewed this again, and I can support it in the design meeting. However, a corresponding change needs to be made in the code that updates from ISO and Wikidata; otherwise the data will drift away from reality.

macchiati:
Looks like this isn't going anywhere (would need to have the updated code for importing from ISO and Wikidata). If it isn't, we should close it.

macchiati:
Closing for now.

PR #1668, opened 2021-12-18
CLDR-15253 Unicode 15 new scripts & script metadata

markusicu:
CLDR-15253

- [ ] This PR completes the ticket.

Updated according to https://cldr.unicode.org/development/updating-codes/updating-script-metadata
For comparison, see PR #1291 for Unicode 14.

The public language-subtag-registry does not yet have the two new scripts.
I cherry-picked them from their [dec08 registration forms](https://mailarchive.ietf.org/arch/msg/ietf-languages/re5D4wIgAeHUiN8YNLZS61UbTJg/), and set dec22 (+2 weeks) as preliminary addition dates.

Note that some of the tools made some changes in generated files that are not related to the two new scripts.
Apparently these tools have not been run since their code or their input data changed.

Some of these unrelated changes caused test failures, and I manually reverted them.

PR #3624, opened 2024-04-09
CLDR-17226 Unicode 16 new scripts & script metadata

markusicu:
CLDR-17226

Updated according to https://cldr.unicode.org/development/updating-codes/updating-script-metadata
For comparison, see PR #1668 for Unicode 15.

I did not run GenerateMaximalLocales because:
```
    public static void main(String[] args) throws IOException {
        if (true) {
            throw new IllegalArgumentException("Don't run this tool until it is fixed");
        }
```

ALLOW_MANY_COMMITS=true


macchiati (approved):
Looks good. We'll be correcting the GenerateMaximal... for next release.

srl295 (approved):
need more investigation into likely

PR #3626, opened 2024-04-10
CLDR-17424 v45: mfwg2 spec pour

srl295:
- pulled from https://github.com/unicode-org/message-format-wg/blob/main/spec

CLDR-17424

ALLOW_MANY_COMMITS=true


macchiati (approved):
I think we might want to tweak the h5 CSS to make them a bit larger, but that's separate from this.

srl295:
> I think we might want to tweak the h5 CSS to make them a bit larger, but that's separate from this.

There's a separate css file for this now, starting out blank.

macchiati:
Not yet built on https://cldr-smoke.unicode.org/spec/main/ldml/tr35-messageFormat.html. How often does the rebuild happen?

srl295:
> Not yet built on https://cldr-smoke.unicode.org/spec/main/ldml/tr35-messageFormat.html. How often does the rebuild happen?

We're not in main anymore.. 

https://cldr-smoke.unicode.org/spec/maint-45/ldml/tr35-messageFormat.html

You can actually get there from https://unicode.org/reports/tr35/proposed.html also

macchiati:
Doooh

macchiati:
Steven, could you fix the release page links *and* add to the BRS
instructions for the branch that we have to change those?

(I got there through the release page, which is where users will get to it
also.)

PR #3628, opened 2024-04-10
CLDR-17490 generate validity for units 2

macchiati:
CLDR-17490

ALLOW_MANY_COMMITS=true


macchiati:
Previous PR was to main, instead of maint-45

PR #3631, opened 2024-04-12
CLDR-17504 fix recursive update in XPathParts

srl295:
Co-authored-by: David Beaumont <david.beaumont+github@gmail.com>

CLDR-17504

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3632, opened 2024-04-12
CLDR-17540 update SupplementalDataInfo reader to not use reentrant XPathParts

srl295:
note this is for v46 - see also #3631 

- clarify distinction between XPathParts and SimpleXPathParts
- push XPathParts.size() and other API up to XPathValue.size()
- use SimpleXPathParts instead of XPathParts in SupplementalDataInfo

CLDR-17540

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
@@ -37,7 +37,7 @@ import java.util.concurrent.ConcurrentHashMap;
  *
  * <p>Caches values for fast lookup.
  *
- * <p>For use at static init time, or anywhere a simpler API can be used, see {@link
+ * <p>(If caching isn't needed, such as with supplemental data at static init time, see {@link
  * SimpleXPathParts#getFrozenInstance(String)}
  */
 public final class XPathParts extends XPathParser
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
@@ -34,6 +34,11 @@ import java.util.concurrent.ConcurrentHashMap;
  * //ldml/characters/exemplarCharacters[@type="auxiliary"] and a list of Element objects that depend
  * on xPath. Each Element object has an "element" string such as "ldml", "characters", or
  * "exemplarCharacters", plus attributes such as a Map from key "type" to value "auxiliary".
+ *
+ * <p>Caches values for fast lookup.
+ *
+ * <p>For use at static init time, or anywhere a simpler API can be used, see {@link
+ * SimpleXPathParts#getFrozenInstance(String)}
  */
 public final class XPathParts extends XPathParser
         implements Freezable<XPathParts>, Comparable<XPathParts> {
@@ -58,12 +63,7 @@ public final class XPathParts extends XPathParser
 
     /** See if the xpath contains an element */
     public boolean containsElement(String element) {
-        for (int i = 0; i < elements.size(); ++i) {
-            if (elements.get(i).getElement().equals(element)) {
-                return true;
-            }
-        }
-        return false;
+        return findElement(element) >= 0;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -58,12 +63,7 @@ public XPathParts() {}
 
     /** See if the xpath contains an element */
     public boolean containsElement(String element) {
-        for (int i = 0; i < elements.size(); ++i) {
-            if (elements.get(i).getElement().equals(element)) {
-                return true;
-            }
-        }
-        return false;
```
srl295, line 66: duplicate implementation!

srl295:
this was more work than made sense for v45 so split out.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -34,6 +34,11 @@
  * //ldml/characters/exemplarCharacters[@type="auxiliary"] and a list of Element objects that depend
  * on xPath. Each Element object has an "element" string such as "ldml", "characters", or
  * "exemplarCharacters", plus attributes such as a Map from key "type" to value "auxiliary".
+ *
+ * <p>Caches values for fast lookup.
+ *
+ * <p>For use at static init time, or anywhere a simpler API can be used, see {@link
```
macchiati, line 40: A downside of SimpleXPathParts is that it doesn't cache, right? That should be mentioned here also, because it is a downside to using SimpleXPathParts for repeated usage.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathValue.java
```diff
@@ -1,14 +1,33 @@
...
 
     /** Get the nth element. Negative values are from end */
     String getElement(int i);
 
+    /** See if the xpath contains an element */
+    public boolean containsElement(String element);
+
+    /**
+     * Get the attributes for the nth element (negative index is from end). Returns null or an empty
+     * map if there's nothing. PROBLEM: exposes internal map
```
macchiati, line 18: It would be better to return an ImmutableMap; that way we are guaranteed that nothing can change it.

Can discuss tradeoffs offline.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathValue.java
macchiati: That is, something like the following.

- from public Element makeImmutable() {
- extract the following into a method in a common location        
```
            if (attributes != null && !(attributes instanceof ImmutableMap)) {
                attributes = ImmutableMap.copyOf(attributes);
           }
```
- From public XPathParts freeze() {
- extract the following into a method in a common location, changing to call the above method instead of Element.makeImmutable()
```
            // ensure that it can't be modified. Later we can fix all the call sites to check
            // frozen.
            List<Element> temp = new ArrayList<>(elements.size());
            for (Element element : elements) {
                temp.add(element.makeImmutable());
            }
            elements = ImmutableList.copyOf(temp);
```
- then change the following to call that method after the handleParse call
```
    SimpleXPathParts(String xpath) {
        handleParse(xpath, true);
    }
```

There are alternatives, eg copy the code into SimpleXPathParts, or have an Elements base class, etc.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
srl295: Yes. This should say, static init time _where XPaths that are going to be read only once are used_.  Especially supplemental data, where all XPaths are unique and caching doesn't help.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathValue.java
srl295: `PROBLEM: exposes internal map` comment and code is from the original XPathParts implementation.  Perhaps that's a good follow up item as a separate concern from this one?

macchiati: Makes sense. Can you clone a ticket for that?

srl295: https://unicode-org.atlassian.net/browse/CLDR-17565

srl295:
@macchiati PTAL, updated the comments.  The immutability is in another ticket

macchiati (approved):

PR #3640, opened 2024-04-17
CLDR-15649 Dashboard using filters

btangmu:
-Add checkbox next to each notification category in Dashboard header

-The category is hidden if the checkbox is unchecked

-Error and Missing notifications still cannot be hidden with pre-existing right-side path-specific checkboxes, but they can be hidden categorically using the new checkboxes

-Show each path with notifications only once in Dashboard

-If a path has more than one notification, combine them on one line

-Perform data conversion on the front end; http response still has multiple notifications per path

-Refactor to avoid http code in Vue; call cldrAjax.mjs from cldrDash.mjs not from DashboardWidget.vue

-New classes DashData and DashEntry in cldrDash.mjs

-Use new method dashData.addEntriesFromJson for both whole-page json and single-path json

-Make hover titles clearer for circled abbreviations like EC for English Changed

-As before, always hide Abstained notifications if user is TC, but implement on the back end instead of the front end (simpler and more efficient)

-Update and extend the unit test TestCldrDash.mjs, since DashData is now different format from json

-Turn off debug-logging in cldrTable (CLDR_TABLE_DEBUG) and cldrVote (CLDR_VOTE_DEBUG)

CLDR-15649

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDash.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDash.mjs
@@ -306,7 +306,9 @@ function getFetchError() {
 function setData(json) {
   cldrProgress.updateVoterCompletion(json);
   const newData = convertData(json);
-  viewSetDataCallback(newData);
+  if (viewSetDataCallback) {
+    viewSetDataCallback(newData);
+  }
   return newData;
 }
 
@@ -339,7 +341,7 @@ function updatePath(dashData, json) {
       if (!json.notifications?.length) {
         // The path no longer has any notifications, so remove the entry
         dashData.removeEntry(dashEntry);
-        return;
+        return dashData; // for unit test
       }
       // Clear attributes that might have changed or disappeared as a result of voting.
       // They will be updated/restored from the json.
--- a/tools/cldr-apps/js/src/esm/cldrDash.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDash.mjs
@@ -2,38 +2,287 @@
  * cldrDash: encapsulate dashboard data.
  */
 import * as cldrAjax from "./cldrAjax.mjs";
+import * as cldrCoverage from "./cldrCoverage.mjs";
 import * as cldrNotify from "./cldrNotify.mjs";
 import * as cldrProgress from "./cldrProgress.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
-import * as cldrXlsx from "./cldrXlsx.mjs";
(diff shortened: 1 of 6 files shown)


srl295 (approved):
LGTM

macchiati:
Just tried this out on Smoke.
https://cldr-smoke.unicode.org/cldr-apps/v#/ab//

Two things I ran into:

   1. *I turned off several items, but then couldn't turn them back on. *The
   problem is that the check mark doesn't change to ON until the dashboard has
   refreshed. So there is a long wait, then it flips. In the meantime, I would
   click again, not realizing it was "in process". That would start the ball
   rolling again, and so on. I see two possibilities.
      1. Flip the check mark to an image indicating 'in process', then to
      check (or empty) once completed.
      2. Gray out the items in the dashboard until the refresh is done.
   2. *The "buttons" are really labels now, and should work like labels and
   look like labels. *That is, the check should go on the right side (not
   left), and clicking on the label should be equivalent to clicking on the
   check box. I see no need for us to have the box around the label — if
   anything, because it is not normal iconography, it misleads people.

btangmu:
@macchiati "Two things" -- 1. The response to clicking a checkbox takes about a second when I test it, and you're right it ought to update immediately at least to show the current checkbox state even if updating all the rows might take a second. I'll investigate that.
2. The buttons serve the important function of scrolling to the first row matching the notification category, so I wouldn't say they're "really [just] labels" now. Unless you disagree that the ability to scroll to the first matching row is important, it seems like they need to remain buttons. About placement of category names in relation to checkboxes, regardless of whether the names are buttons or labels, the convention I'm familiar with is that a checkbox should be to the left of its descriptive text -- see https://uxpickle.com/should-checkbox-labels-be-to-the-right-or-the-left/ -- what would be a counter-example in common usage?

macchiati:
- It would be better to have some indication that it is in progress,
   since in the cases I saw it was not just a second, but several seconds
   (long enough to make me think it wasn't working). One option is to have the
   same visual as when the dashboard is first opened; a spinning image,
   where once the dashboard is populated it disappears. I'm guessing since you
   already have that it would be straightforward.
   - Checkbox on the left is clearly the default for HTML that people are
   used to, eg
   https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_checkbox
   - The button functionality is very confusing. Since we have filters now
   I don't think it is necessary. If we really wanted that, we would have an
   additional icon after the label, something like (fast forward) rotated to
   point down.

So something like (ignore the placement/style)

 ✅ Missing (374) [image: Screenshot 2024-04-30 at 18.24.18.png]

PR #3645, opened 2024-04-19
CLDR-17336 Fix spec ambiguity surrounding parentLocales

sffc:
CLDR-17336

I fixed an ambiguous part of the spec where it said component parent locales "inherit" from the root component, only to be followed by a `Note` that this behavior was changed. I have now fully removed the text about inheriting from the root component and only "overriding".

I also moved the paragraph about components overriding `localeRules` to the section where components are introduced, because that paragraph references the concept of components.

CC @robertbastian


macchiati:
Good catches.

That's a v45 ticket, and we don't allow tickets to span releases. Since it
is copyediting, you can clone the ticket, and mark yourself as assignee,
and fixVersion as 46.

I have a couple of minor tweaks, but will save them for the altered PR.

macchiati (approved):

PR #3646, opened 2024-04-19
CLDR-17561 Utility for visiting DtdData elements

macchiati:
CLDR-17561

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java
@@ -1,8 +1,7 @@
 package org.unicode.cldr.unittest;
 
-import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Multimap;
+import com.ibm.icu.util.MatchElementAttribute;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


macchiati:
Draft; will need some dtd fixes and/or exceptions.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java
```diff
@@ -689,4 +693,84 @@ private int removeNonDistinguishing(
...
+        DtdVisitor visitor =
+                new DtdVisitor() {
+                    final MatchElementAttribute skipAttributeNames =
+                            new MatchElementAttribute()
+                                    .add( // Add, once checking to make sure that these are safe.
+                                            // Pairs of element, attribute
+                                            "", "references", //
+                                            "", "cp", //
+                                            "version", "", //
+                                            // "ruleset", "type", //
```
btangmu, line 707: Commented-out code and empty comments are messy to begin with, and messier due to deep indentation and formatting that breaks up some but not all pairs. Ideally the parameters for add() might be moved to a named constant defined outside the method call, less indented, maybe as an array Map.Entry objects for element/attribute pairs...

The semantics of these elements and attributes are still mostly beyond my understanding. I sometimes wonder whether defining constants for such strings, maybe as members of enums, and referencing them as such rather than as string literals could be helpful for understanding the code and maybe for catching typos...

btangmu (approved):
LGTM though partly over my head

PR #3649, opened 2024-04-19
CLDR-16101 Change Error Subtype to read from Google Sheets

srl295:
CLDR-16101

- Reads from https://docs.google.com/spreadsheets/d/1n7H_yt2Sxea1_AAp6Ggi5zo1HxtC5-xwieBlyM2kAcs/edit#gid=1962856317 for subtype
- automatically updates a sheet with ALL subtype names

ALLOW_MANY_COMMITS=true


macchiati (approved):
Looks good. Question: would it be hard to move the "read a spreadsheet" utility into the .util directory? I could see that being very useful for making lots of stuff more dynamic.

srl295:
> Looks good. Question: would it be hard to move the "read a spreadsheet" utility into the .util directory?

This doesn't download, but it does access the data.  It reads it into a cache directory in a manner of speaking.

>  I could see that being very useful for making lots of stuff more dynamic.

yes, although the authentication bit is interesting.  This mode is suitable for a service system, if you are envisioning a command line utility it would be easier to set it up to have the user login locally (slightly different setup).

PR #3643, opened 2024-04-18
CLDR-17483 add lld Ladin locale

DavidLRowe:
CLDR-17483

- [ ] This PR completes the ticket.

This PR adds lld base locale; ticket asks for additional variants to be added.

ALLOW_MANY_COMMITS=true

PR #3656, opened 2024-04-26
CLDR-17483 add Ladin (lld) variants

DavidLRowe:
CLDR-17483

This PR and the previous PR (#3643) should complete the ticket. The previous PR added lld.xml and lld_IT.xml. This PR adds:
lld_ANPEZO.xml
lld_FASCIA.xml
lld_FODOM.xml
lld_GHERD.xml
lld_VALBADIA.xml
for the anpezo, fascia, fodom, gherd and valbadia variants. Note that the valbadia file has only the identity block, since this variant is used as the default for lld.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/lld_ANPEZO.xml
+++ b/common/main/lld_ANPEZO.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="lld"/>
+		<variant type="ANPEZO"/>
+	</identity>
+	<characters>
+		<exemplarCharacters>[a-z à á â è é ê ë ì í î ò ó ô ö ù ú ü ć ś]</exemplarCharacters>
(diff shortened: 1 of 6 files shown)


DavidLRowe:
@srl295 The build fails because the five variants aren't in common/validity/variant.xml, however the file says that the contents are generated and should not be manually updated. Reference is to GenerateValidityXML.java but what needs to change to cause that tool to add the variants to variant.xml?

srl295:
> @srl295 The build fails because the five variants aren't in common/validity/variant.xml, however the file says that the contents are generated and should not be manually updated. Reference is to GenerateValidityXML.java but what needs to change to cause that tool to add the variants to variant.xml?

oh, are they in an update language subtag registry? perhaps that file needs a refresh.

DavidLRowe:
@srl295 I went ahead and added the 5 variant names to the generated data file. Hopefully you can figure out what changes are needed to get the file updated in the correct manner.

srl295:
> > @srl295 The build fails because the five variants aren't in common/validity/variant.xml, however the file says that the contents are generated and should not be manually updated. Reference is to GenerateValidityXML.java but what needs to change to cause that tool to add the variants to variant.xml?
> 
> oh, are they in an update language subtag registry? perhaps that file needs a refresh.

indeed <https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry> has an update that needs to be brought in.

srl295 (approved):

srl295:
I'll update LST in https://unicode-org.atlassian.net/browse/CLDR-17585

srl295:
@DavidLRowe CLA kicked by visiting <https://cla-assistant.io/check/unicode-org/cldr?pullRequest=3656>

PR #3650, opened 2024-04-21
CLDR-17566 Create downloads/cldr-45.md

macchiati:
CLDR-_____

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3660, opened 2024-04-29
CLDR-17566 fix to GitHub Pages build

srl295:
CLDR-17566

Looking at #3650 I realized that some subdirectories were not getting build properly. This fixes the build to use `github-pages build` instead of raw `jekyll`.  FYI @lianghai

To review, see: (note srl URL)

- https://srl295.github.io/cldr/site/downloads/cldr-45.html
- https://srl295.github.io/cldr/ (LDML spec should be same as usual)
- https://srl295.github.io/cldr/dev/generate-emoji-paths.html

These are broken in current main, for example <https://unicode-org.github.io/cldr/dev/generate-emoji-paths.html> and <https://unicode-org.github.io/cldr/site/downloads/cldr-45.html> are dead links.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,4 +1,10 @@
+
 name: Publish to gh-pages
+
+permissions:
+  pages: write
+  deployments: write
+  id-token: write
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


macchiati (dismissed):

srl295:
@macchiati Please re-review. works now

macchiati (approved):

PR #3663, opened 2024-04-30
CLDR-16853 remove raw URLs from checks, handle via subtype map

srl295:
CLDR-16853

ALLOW_MANY_COMMITS=true


srl295:
I think I rooted out all of the "For what to do" and "See" type messages.  They need to be handled via subtype map.  Plus, it should be noted that 

oops, i found a couple of others.. Plus they look really bad in the dashboard,  waste of space saying the same thing 3x per row

image

macchiati (approved):

PR #3667, opened 2024-05-01
CLDR-17582 BRS Add any necessary emoji annotations and collation info 

macchiati:
for the next Unicode Version

CLDR-17582

Following instructions in https://github.com/unicode-org/cldr/blob/main/docs/dev/generate-emoji-paths.md

Notes:
- Needed an update for the emoji-test.txt parser.
- The tooling didn't work for generating the annotations for adding to en.xml and root.xml, 
so I had to generate those by hand (well, with the assistance of a spreadsheet) from candidateData.txt
- The emoji charts for v16 candidates are also not posted, so I couldn't cross check with those.

Will forward this PR to Ned also.

TODO: collation changes
Can come come in a later PR, because they don't block submission.

ALLOW_MANY_COMMITS=true


macchiati:
The change in format for emoji-test also caused a bit test failure; will have to fix that.

pedberg-icu (approved):

nedley:
I sent you the GenerateCldrData.java output, which should help avoid future spurious diffs.

PR #3666, opened 2024-05-01
CLDR-17600 BRS 7 Update language names

macchiati:
CLDR-17600

As per BRS. This hasn't been done for a while. I updated 3 lists of languages:

- go to [supplemental/locale_coverage.html](https://www.unicode.org/cldr/charts/latest/supplemental/locale_coverage.html) 
- Take all with computed level ≥ basic. Drop script/region and remove dups. 
- Put Target level= blank or starts with ‡ into attributeValueValidity id $languageExceptions. 
- Put others in $language.
- Put older languages that didn't fit the above into $oldLanguages

Each list was canonicalized: 
reorganized to remove overlaps and duplicates, 
and put into alphabetical order with each new initial letter starting a row.

Notes:
- This makes all the Basic+ languages be at Modern coverage
- The translated names for languages below Basic won't be required for other locales to reach Modern.
- It doesn't remove any translated names of languages that are already in CLDR
- It does reorganize some of the languages, and makes the variable names match their contents, semantically.

If you want to see the names of $oldLanguages, go to https://util.unicode.org/UnicodeJsps/languageid.jsp?a=aa+ace+ada+ady+ain+ale+alt+anp+arp+ars+atj+av+awa+ay+ban+bi+bin+bla+bug+cay+ch+chk+chm+chp+chy+clc+crg+crj+crk+crl+crm+crr+crs+dak+dar+dgr+dzg+efi+eka+fj+fon+frc+gan+gil+gor+gwi+hai+hak+hax+hil+hmn+hsn+ht+hup+hur+hz+iba+ibb+ikt+ilo+inh+kac+kbd+kfo+kha+kj+kmb+kr+krc+krl+kru+kum+kv+kwk+lad+lez+li+lil+lld+lou+loz+lsm+ltg+lua+lun+lus+mad+mag+mak+men+mh+min+moe+mos+mwl+na+nan+nap+nb+new+ng+nia+niu+nog+ojb+ojc+ojs+ojw+oka+pag+pam+pau+pqm+quc+rap+rar+rup+sad+sba+sco+slh+sm+sma+smj++sms+snk+srn+str+suk+swb+sy+tce+tem+tet+tgx+tht+tlh+tli+ttm+tum+tvl+ty+udm+umb+war+wuu+xal+ybb

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3680, opened 2024-05-02
CLDR-17600 BRS 7 No2 Update language names

macchiati:
CLDR-17600

Note: https://github.com/unicode-org/cldr/pull/3666 was approved, but was too difficult to fix merge conflicts, so recreated here. Only significant changes were some additional test fixes.

ALLOW_MANY_COMMITS=true


DavidLRowe (approved):
RSLGTM

PR #3681, opened 2024-05-02
CLDR-17106 add more fixed candidate items

srl295:
- restructure DataPage / DiskDataCache to hold the candidate providers
- add a FixedCandidateProvider structure, designed to allow quick additions of future fixed candidates.

CLDR-17106

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DiskDataCache.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DiskDataCache.java
@@ -1,24 +1,49 @@
 package org.unicode.cldr.web;
 
+import static org.unicode.cldr.web.FixedCandidateProvider.forEnumWithFixedXpath;
+
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.google.common.collect.ImmutableList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Logger;
+import javax.annotation.Nonnull;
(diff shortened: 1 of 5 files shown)


srl295:
image

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DiskDataCache.java
```diff
@@ -1,24 +1,49 @@
...
 public class DiskDataCache {
     static final Logger logger = Logger.getLogger(DiskDataCache.class.getSimpleName());
 
     private final Factory factory;
-    private final CLDRFile english;
+    private final CLDRFile english; // TODO: Unused?
     private final PathHeader.Factory phf;
+    final SupplementalDataInfo sdi;
+
+    final List<FixedCandidateProvider> personNameProviders =
```
srl295, line 39: these are not locale dependent, so kept in a separate static list.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DiskDataCache.java
```diff
@@ -27,18 +52,57 @@ class DiskDataEntry {
...
+            }
+
+            @Override
+            protected Collection<String> getCandidates() {
+                return grammarInfo.get(
+                        GrammaticalTarget.nominal,
+                        GrammaticalFeature.grammaticalGender,
+                        GrammaticalScope.units);
+            }
+        }
```
srl295, line 87: this one is locale dependent depends on the supplementalDataInfo in the enclosing class

macchiati:
what are "fixed candidates"?

srl295:
> what are "fixed candidates"?

Candidates are options during voting… Fix candidates means that you have a fixed list instead of being able to use the plus button

srl295:
In this pull request, I tried to write it so that it was easy to add more of these later

macchiati (approved):

PR #3682, opened 2024-05-03
CLDR-17459 Add units for grammar

macchiati:
CLDR-17459

Added grammar to some units that were skipped earlier.
The breakdown is at https://docs.google.com/spreadsheets/d/1soBtcb3SwNaCMg3maD0vvbWqqimco4ZKpdtSpB7NJ1E/edit#gid=0

Also
- added a comparator for units, to help in examining the result. (And are needed for another ticket.)
- set gasoline-per-mile to not be metric_adjacent

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
@@ -745,21 +745,19 @@ public class GrammarInfo implements Freezable<GrammarInfo> {
                     "month",
                     "year");
 
+    // To see a list of the results for double-checking, run TestUnits with TestUnitsToTranslate -v
     static final Set<String> EXCLUDE_GRAMMAR =
             Set.of(
-                    "point",
-                    "dunam",
-                    "dot",
-                    "astronomical-unit",
-                    "nautical-mile",
-                    "knot",
-                    "dalton",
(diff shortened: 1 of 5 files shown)


srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/GrammarInfo.java
```diff
@@ -745,21 +745,19 @@ public static Set<String> getGrammarLocales() {
...
+                    "knot", // US/UK specific
+                    "dalton", // specialized
+                    "electronvolt", // specialized
                     "kilocalorie",
-                    "electronvolt",
-                    // The following may be reinstated after 45.
-                    "dot-per-centimeter",
-                    "millimeter-ofhg",
-                    "milligram-ofglucose-per-deciliter");
+                    "point");
```
btangmu, line 760: if these were alphabetized this might be less confusing

btangmu (approved):
rubber stamp... this is a strange combination of changes :-)

macchiati:
The new text, except for point, all has comments, and ones with the same comments are (mostly) grouped together. I'm taking the 'point' back into the committee to see if that should be removed (where removal means we add grammar).

PR #3685, opened 2024-05-03
CLDR-17492 Temporarily point to a fixed snapshot

mihnita:
Use a specific SNAPSHOT so that the cldr-utilities refactoring does not break the functionality.

Once the changes land in ICU, we will publish a new version. Then back here I update the package, and go back to using 76.0.1-SNAPSHOT, in one single commit.

CLDR-17492

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/pom.xml
+++ b/tools/pom.xml
@@ -25,7 +25,9 @@
 			SNAPSHOT for the ICU version that we want, so this should only need updating
 			when the ICU version changes e.g. from 74.0.1, to 74.1, to 75.0.1, then to 75.1.
 			You can use github actions in icu to manually push the latest SNAPSHOT version. -->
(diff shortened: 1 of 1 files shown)


srl295:
@mihnita needs a `CLDR-` ticket instead of `ICU-` (pr title and commit)

@sffc this is a bug in the pr checker, it's allowing an ICU PR in the CLDR repo.. https://github.com/unicode-org/jira-github-pr-check/issues/66

srl295 (approved):

markusicu:
@srl295 I think that @mihnita does not have permission to merge. Do you want to merge on his behalf? Do you want me to?

mihnita:
Thank you all!
Have a nice week-end.
M.

PR #3687, opened 2024-05-03
CLDR-16845 improve webpack, reduce load time

srl295:
- was still not able to get rid of separate bootstrap loads due to jquery issues

CLDR-16845

ALLOW_MANY_COMMITS=true


srl295:
@btangmu this one provided better load times but has a different look. CSS problem. . So I may try to bring some of the loader structure improvements in.

PR #3690, opened 2024-05-06
CLDR-16941 Improve coverage for non-latn default/native number symbols/patterns

pedberg-icu:
CLDR-16941

1. Improve coverage for number system symbols and patterns belonging to non-`latn` default, native, and finance numbering systems in all locales. In some cases we select coverage by script instead of or in addition to language, to disambiguate cases where languages are written in multiple scripts.
2. Add a unit test to verify that various number symbols and patterns are at the desired coverage level for each of the two or three number systems (default, native, and possibly finance) specified in a locale, for all locales. This currently takes some time (25 sec on my machine), maybe could be sped up.
3. Whet this does not do is group collections of number symbols or patterns into sets. That would require coverage variables with alternate values for elements in paths, which is currently not supported. We can discuss for a future enhancement, this would have performance impact.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldmlSupplemental.dtd
+++ b/common/dtd/ldmlSupplemental.dtd
@@ -1278,7 +1278,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST coverageLevel inLanguage CDATA #IMPLIED >
     <!--@MATCH:any-->
 <!ATTLIST coverageLevel inScript CDATA #IMPLIED >
-    <!--@MATCH:validity/script-->
+    <!--@MATCH:any-->
 <!ATTLIST coverageLevel inTerritory CDATA #IMPLIED >
     <!--@MATCH:any-->
 <!ATTLIST coverageLevel value CDATA #REQUIRED >
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -180,24 +180,42 @@ For terms of use, see http://www.unicode.org/copyright.html
 
 		<coverageVariable key="%anyAttribute" value='([^\x{22}]++)'/>
 
-		<!-- Number system coverage by language; default number system items at various levels, native system items all at modern -->
-		<coverageVariable key="%arabDefaultLanguages" value="(ar|ckb)"/>
-		<coverageVariable key="%arabextDefaultLanguages" value="(fa|ks|lrc|pa|ps|ur|uz)"/>
-		<coverageVariable key="%arabextNativeLanguages" value="(mzn|ug)"/>
-		<coverageVariable key="%bengDefaultLanguages" value="(as|bn)"/>
-		<coverageVariable key="%devaDefaultLanguages" value="(mr|ne)"/>
-		<coverageVariable key="%devaNativeLanguages" value="(brx|hi|kok)"/>
-		<coverageVariable key="%gujrNativeLanguages" value="(gu)"/>
-		<coverageVariable key="%guruNativeLanguages" value="(pa)"/>
-		<coverageVariable key="%hanidecNativeLanguages" value="(yue|zh)"/>
+		<!-- Number system coverage by language and/or script; default number system items at various levels, native & finance system items all at modern -->
+		<coverageVariable key="%adlmDefaultScripts" value="(Adlm)"/>
+		<coverageVariable key="%arabDefaultLanguages" value="(ar|ckb|sd|sdh)"/> <!-- && inScript="Arab" to disambiguate languages in multiple scripts -->
+		<coverageVariable key="%arabNativeLanguages" value="(dv|ha)"/> <!-- && inScript="Arab" -->
+		<coverageVariable key="%arabextDefaultLanguages" value="(az|bgn|fa|ks|lrc|mzn|pa|ps|ur|uz)"/> <!-- && inScript="Arab" -->
+		<coverageVariable key="%arabextNativeLanguages" value="(trw|ug)"/> <!-- && inScript="Arab" -->
+		<coverageVariable key="%arabextAllLanguages" value="(az|bgn|fa|ks|lrc|mzn|pa|ps|trw|ug|ur|uz)"/> <!-- union of default  & native, for display name -->
+		<coverageVariable key="%bengDefaultScripts" value="(Beng)"/> <!-- e.g. mni can be written in Beng or Mtei -->
+		<coverageVariable key="%cakmDefaultScripts" value="(Cakm)"/>
+		<coverageVariable key="%devaDefaultLanguages" value="(bgc|bho|mr|ne|raj|sa|sat)"/> <!-- && inScript="Deva" to disambiguate languages in multiple scripts -->
+		<coverageVariable key="%devaNativeLanguages" value="(brx|doi|hi|kok|kxv|mai|xnr)"/> <!-- && inScript="Deva" -->
+		<coverageVariable key="%gujrNativeScripts" value="(Gujr)"/>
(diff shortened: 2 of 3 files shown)


AEApple:
Are we also tracking whether the equivalent number formats to those currently in basic where the locale uses native digits by default in a separate ticket for resolution in esub or a future release instead (since they are available at a higher coverage level if we do need to change them in v46)?

pedberg-icu:
> Are we also tracking whether the equivalent number formats to those currently in basic where the locale uses native digits by default in a separate ticket for resolution in esub or a future release instead (since they are available at a higher coverage level if we do need to change them in v46)?

@AEApple Not sure I understand the question. We had coverage in certain locales for the number symbols and patterns for default and native number systems, at either moderate or modern coverage depending on the item. What this ticket does is ensure that we do that across _all_ locales for all of the number systems that are default, native, or finance systems in those locales.

common/dtd/ldmlSupplemental.dtd
```diff
@@ -1278,7 +1278,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST coverageLevel inLanguage CDATA #IMPLIED >
     <!--@MATCH:any-->
 <!ATTLIST coverageLevel inScript CDATA #IMPLIED >
-    <!--@MATCH:validity/script-->
+    <!--@MATCH:any-->
```
pedberg-icu, line 1281: This is so we can use coverage variables for inScript, as we do for inLanguage for example.

macchiati (approved):
looks great. A lot of work, though! Longer term, I think we need to set aside some time to review the structure of the coverage levels, and some up with a more concise format to lower the maintenance burden. We already talked about expanding elements; we also might have more predefined variables (that we can construct at boot time based on what's in a particular locale). And other ideas.

common/dtd/ldmlSupplemental.dtd
srl295: ideally `@MATCH:validity/script-with-variables` or something

Larger question is whether coverage level belongs in "supplemental data" when, like casing, it's primarily inward tool-focused data.  Perhaps we should split some of these out -- and give them their own DTD.

common/supplemental/coverageLevels.xml
```diff
@@ -180,24 +180,42 @@ For terms of use, see http://www.unicode.org/copyright.html
...
-		<coverageVariable key="%arabextDefaultLanguages" value="(fa|ks|lrc|pa|ps|ur|uz)"/>
-		<coverageVariable key="%arabextNativeLanguages" value="(mzn|ug)"/>
-		<coverageVariable key="%bengDefaultLanguages" value="(as|bn)"/>
-		<coverageVariable key="%devaDefaultLanguages" value="(mr|ne)"/>
-		<coverageVariable key="%devaNativeLanguages" value="(brx|hi|kok)"/>
-		<coverageVariable key="%gujrNativeLanguages" value="(gu)"/>
-		<coverageVariable key="%guruNativeLanguages" value="(pa)"/>
-		<coverageVariable key="%hanidecNativeLanguages" value="(yue|zh)"/>
+		<!-- Number system coverage by language and/or script; default number system items at various levels, native & finance system items all at modern -->
+		<coverageVariable key="%adlmDefaultScripts" value="(Adlm)"/>
```
srl295, line 184: I still don't think any of this belongs in data at all.  It should be read from the locale files themselves and cached.  But.. it's an improvement.

common/supplemental/coverageLevels.xml
```diff
@@ -265,29 +283,34 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel	value="moderate"	inLanguage="%traditionalCollationLanguages"	match="localeDisplayNames/types/type[@key='collation'][@type='traditional']"/>
 		<coverageLevel	value="moderate"	inLanguage="%CJK_Languages"	match="localeDisplayNames/types/type[@key='collation'][@type='unihan']"/>
 		<coverageLevel	value="moderate"	 	match="localeDisplayNames/types/type[@key='numbers'][@type='latn']"/>
-		<coverageLevel	value="moderate"	inLanguage="zh"	match="localeDisplayNames/types/type[@key='numbers'][@type='(han(s|t)(fin)?|hanidec)']"/>
+		<coverageLevel	value="moderate"	inLanguage="%hanidecNativeLanguages"	match="localeDisplayNames/types/type[@key='numbers'][@type='(han(s|t)(fin)?|hanidec)']"/>
```
srl295, line 286:
```suggestion
		<coverageLevel	value="moderate"	inLanguage="%hanidecNativeLanguages"	match="localeDisplayNames/types/type[@key='numbers'][@type='(hans|hant)(fin)?|hanidec)']"/>
```

more readable.. or, spell all of them out.

common/supplemental/coverageLevels.xml
```diff
@@ -423,90 +446,147 @@ For terms of use, see http://www.unicode.org/copyright.html
...
+		<coverageLevel	value="moderate"	inScript="%hmnpDefaultScripts"	match="numbers/symbols[@numberSystem='hmnp']/group"/>
+		<coverageLevel	value="moderate"	inScript="%hmnpDefaultScripts"	match="numbers/symbols[@numberSystem='hmnp']/minusSign"/>
+		<coverageLevel	value="moderate"	inScript="%hmnpDefaultScripts"	match="numbers/symbols[@numberSystem='hmnp']/percentSign"/>
+		<coverageLevel	value="moderate"	inScript="%hmnpDefaultScripts"	match="numbers/symbols[@numberSystem='hmnp']/plusSign"/>
+		<coverageLevel	value="moderate"	inScript="%mteiDefaultScripts"	match="numbers/symbols[@numberSystem='mtei']/decimal"/>
+		<coverageLevel	value="moderate"	inScript="%mteiDefaultScripts"	match="numbers/symbols[@numberSystem='mtei']/group"/>
+		<coverageLevel	value="moderate"	inScript="%mteiDefaultScripts"	match="numbers/symbols[@numberSystem='mtei']/minusSign"/>
+		<coverageLevel	value="moderate"	inScript="%mteiDefaultScripts"	match="numbers/symbols[@numberSystem='mtei']/percentSign"/>
+		<coverageLevel	value="moderate"	inScript="%mteiDefaultScripts"	match="numbers/symbols[@numberSystem='mtei']/plusSign"/>
```
srl295, line 488: i'll say again, I appreciate this being done.

srl295 (approved):

srl295:
> looks great. A lot of work, though! Longer term, I think we need to set aside some time to review the structure of the coverage levels, and some up with a more concise format to lower the maintenance burden. We already talked about expanding elements; we also might have more predefined variables (that we can construct at boot time based on what's in a particular locale). And other ideas.

drop the data from coverage*.xml entirely and derive it from the locale files

common/dtd/ldmlSupplemental.dtd
pedberg-icu: Filed https://unicode-org.atlassian.net/browse/CLDR-17618

common/supplemental/coverageLevels.xml
pedberg-icu: Covered by https://unicode-org.atlassian.net/browse/CLDR-17618

pedberg-icu:
> > looks great. A lot of work, though! Longer term, I think we need to set aside some time to review the structure of the coverage levels, and some up with a more concise format to lower the maintenance burden. We already talked about expanding elements; we also might have more predefined variables (that we can construct at boot time based on what's in a particular locale). And other ideas.
> 
> drop the data from coverage*.xml entirely and derive it from the locale files

Covered by https://unicode-org.atlassian.net/browse/CLDR-17618

PR #3691, opened 2024-05-06
CLDR-17610 v46 BRS: keyboard chart rename

srl295:
- rename from keyboard to keyboards, matching prior releases

CLDR-17610

ALLOW_MANY_COMMITS=true


macchiati (approved):

srl295:
@macchiati thanks, can i get a review on the two at https://github.com/unicode-org/cldr-staging/pulls too?

PR #3694, opened 2024-05-07
CLDR-17461 Forum posts in vetting view, boost performance, change icons

btangmu:
-Add a cache localeForumStatusMap in SurveyForum.java

-New class SurveyForum.LocaleForumStatus

-Get all the status for a locale with a single sql call

-Clear the cache for a locale when a post is added/closed

-Revise icons in the Comparison (English) column: 👁️‍🗨️ and 💬

-👁️‍🗨️ means: This item has one or more forum posts, some of which are open

-💬 means: This item has one or more forum posts, all of which are closed

-Change let to const for two items in cldrSurvey.mjs; remove @property

-Call updateRowEnglishComparisonCell even if isSetup, and always removeAllChildNodes; old optimization assumed the cell contents never changed

CLDR-17461

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrForum.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrForum.mjs
@@ -432,6 +432,7 @@ function loadHandlerForSubmit(data) {
       cldrLoad.reloadV(); // main Forum page
     } else {
       cldrForumPanel.updatePosts(null); // Info Panel
+      cldrSurvey.expediteStatusUpdate(); // update forum icons (👁️‍🗨️, 💬) in the main table
     }
   } else {
     const post = $(".post").first();
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
@@ -80,10 +80,9 @@ const statusActionTable = {
 /**
  * How often to fetch updates. Default 15s.
  * Used only for delay in calling updateStatus.
- * @property timerSpeed
  */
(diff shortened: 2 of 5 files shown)


btangmu:
Running this locally with JProfiler, there's no longer a major hot spot for the sql query. PathForumStatus is 0.0% at the bottom of this screenshot:

image

btangmu:
It turns out that making/closing a forum post DOES update json.localeStamp, but the main table cell wasn't getting updated due to an old optimization based on an assumption that the Comparison (English) column never changes!

In cldrTable.mjs:

```
  /*
   * Set up the "comparison cell", a.k.a. the "English" column.
   */
  if (comparisonCell && !comparisonCell.isSetup) {
    updateRowEnglishComparisonCell(tr, theRow, comparisonCell);
  }

```

If I remove ` && !comparisonCell.isSetup` then the cell updates, but in a funny way, new data adding to old data instead of replacing it:
 
image

That can be fixed by adding this line with `removeAllChildNodes`:

```
function updateRowEnglishComparisonCell(tr, theRow, cell) {
  cldrDom.removeAllChildNodes(cell);
```

I'll follow up...

srl295 (approved):
looks great to me!

btangmu:
My second commit fixed the comparisonCell problem; I'll squash...

srl295 (approved):

macchiati (approved):

btangmu:
@srl295 "we squash on merge anyway" -- was that always true? anyway, good to know, I agree the changes in a PR are often easier to see without the squashing

PR #3698, opened 2024-05-08
CLDR 16668 Revise order of units

macchiati:
CLDR-16668

Still draft because I'd like to discuss with Tom.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartUnitConversions.java

- Use the new collator in charts

- Use the new collator in XML paths

- add information to UnitSystem enum to make comparision easier.
- can use int for sort key.
- add QuantityGroup for more natural ordering of quantities; also intended as utility for ST (which could subdivide further)

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitPathType.java

- add simple utility for displayName

- enhance one of the tests to also print information about the quantities, standard units, current ST Pages, ordering information, systems, etc.

ALLOW_MANY_COMMITS=true


macchiati:
Still draft, but feedback is useful

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -1851,28 +1833,41 @@ public enum UnitSystem { // TODO convert getSystems and SupplementalDataInfo to
         si_acceptable,
```
btangmu, line 1833: Is the order significant -- possibly affecting order of rows/pages in ST for example? If so, comment would be helpful, otherwise alphabetical order might be appropriate

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -2372,4 +2371,77 @@ public String format(
                 nf3.locale(uLocale).format(outputAmount.doubleValue()).toString();
         return com.ibm.icu.text.MessageFormat.format(pattern, cldrFormattedNumber);
     }
+
+    public enum QuantityGroup {
+        Duration("duration", "year-duration"),
+        Length("length"),
```
btangmu, line 2377: Here again a brief comment explaining the significance of the ordering could be helpful

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: It would affect the order of rows within the pages, grouping the quantities better.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: I also have a proposal for breaking apart the Units Other (which also ends up breaking apart a couple of others. It's just draft for now until we can discuss.

PR #3702, opened 2024-05-08
CLDR-16941 Further number symbols/formats coverage fixes per TC discussion

pedberg-icu:
CLDR-16941

Follow-on to PR [#3690](https://github.com/unicode-org/cldr/pull/3690) - Further number symbols/formats coverage fixes mostly per TC discussion 2024-05-08 in doc [Non-latn number systems coverage](https://docs.google.com/document/d/1cp01T7wBl-5oAjCNYsrkC8NtAOJQ-sNC9XTKnLOX7eQ/edit#heading=h.ecg8w5wx2swc):
- Base forms for compact currency and compact decimal formats in numbers='latn' should be at moderate, not modern
- Where coverage is at basic/moderate for number symbols/formats with numbers='latn', coverage should be at the same level for non-latn number systems that are the defaultNumberingSystem in a locale. This will make several locales whose basic coverage is near 100% but moderate coverage is low drop out of basic until these items are supplied.
    - In the TC meeting we mentioned these affected locales: ccp, dz, ff_Adlm, ks, nqo, sa
    - However the following are also affected; they were not in the table discussed in the meeting because they have (probably incorrectly) native number system = 'latn' though default is non-latn: bgc bho mni raj sat
- The comment in localeCoverage.xml about coverage for non-latn native numbering system elements being all at modern was incorrect, fixed it.
- Spelled out all of the number systems in '(han(s|t)(fin)?|hanidec)' per comment in first PR.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -179,7 +179,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 
 		<coverageVariable key="%anyAttribute" value='([^\x{22}]++)'/>
 
-		<!-- Number system coverage by language and/or script; default number system items at various levels, native & finance system items all at modern -->
+		<!-- Number system coverage by language and/or script; default number system items at various levels, native & finance system items at moderate (for content) or modern -->
 		<coverageVariable key="%adlmDefaultScripts" value="(Adlm)"/>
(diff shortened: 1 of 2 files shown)


pedberg-icu:
Anybody, please feel free to merge when approved and tests pass.

macchiati (approved):

PR #3704, opened 2024-05-08
CLDR-17122 add kok-Latn

DavidLRowe:
CLDR-17122

ALLOW_MANY_COMMITS=true


macchiati (changes requested):
The PR title needs to be kok_Latn instead.

macchiati (approved):

PR #3706, opened 2024-05-09
CLDR-17483 lld remove variants

DavidLRowe:
CLDR-17483

Previous PRs first added the Ladin (lld) locale, then attempted to add five locales (one of which, valbadia, will be used as the default for Ladin). There were problems with the variants showing up in the Survey Tool. 

This PR removes the variants, leaving just the Ladin locale, which can be used in the Survey Tool for the addition of data using the valbadia variant.

ALLOW_MANY_COMMITS=true


srl295 (approved):

srl295:
LGTM. can you make sure we have a ticket for tracking the ST issues?

DavidLRowe:
Would it make sense to clone the CLDR-17483 ticket and note the problems with getting variants to work? I've never cloned a Jira ticket, so I don't know the procedure (or the advantages).

srl295:
> Would it make sense to clone the [CLDR-17483](https://unicode-org.atlassian.net/browse/CLDR-17483) ticket and note the problems with getting variants to work? I've never cloned a Jira ticket, so I don't know the procedure (or the advantages).

that would be fine. probably mark it new ddl

PR #3708, opened 2024-05-09
CLDR-7633 ansible: move performance improvements into ansible

srl295:
- http/2
- gzip compression fixes

CLDR-7633

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/scripts/ansible/cldr-apps-playbook.yml
+++ b/tools/scripts/ansible/cldr-apps-playbook.yml
@@ -53,6 +53,7 @@
         force: no
         update: no
         version: main
+        remote: origin
         # this is a deep clone because we will need to keep updating
         # it with history. It does not include LFS as that
         # is not needed for the surveytool.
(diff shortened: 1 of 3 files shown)


srl295:
@btangmu test failures due to https://unicode-org.atlassian.net/browse/CLDR-17297

tools/scripts/ansible/cldr-apps-playbook.yml
```diff
@@ -53,6 +53,7 @@
         force: no
         update: no
         version: main
+        remote: origin
```
btangmu, line 56: as contrasted with "upstream"? Looks like "origin" is default so this is redundant (not necessarily a bad thing) per https://docs.ansible.com/ansible/latest/collections/ansible/builtin/git_module.html

btangmu (approved):

tools/scripts/ansible/cldr-apps-playbook.yml
srl295: it should be, but seemed to fail until this was set.

PR #3712, opened 2024-05-11
CLDR-17566 Converting development process to markdown

chpy04:
CLDR-17566

Part of the [CLDR --> GFM conversion project](https://docs.google.com/document/d/1NoQX0zqSYqU4CUuNijTWKQaphE4SCuHl6Bej2C4mb58/edit)

View text diffs [here](https://docs.google.com/document/d/1P4rx1X91QvvBhLaYlCMJTKZl66sq6UNsbwHRk7wtW7w/edit?usp=sharing)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/development-process.md
+++ b/docs/site/development/development-process.md
@@ -6,8 +6,7 @@ title: Handling Tickets (bugs/enhancements)
 
 ### Filing tickets
 
-  1.  You can add set other fields, but don't change the Owner. That way the committee sees the incoming tickets (those marked as Owner: somebody).
-        
+  1.  You can add set other fields, but don't change the Owner. That way the committee sees the incoming tickets (those marked as Owner: somebody).  
   2.  _**If you really have to fix a bug right away, file it and send a notice to cldr-dev explaining what you are doing and why. Otherwise, file the bug, wait for the next Wednesday meeting for discussion and approval.**_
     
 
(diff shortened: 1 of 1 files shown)


docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+
+1.  Ticket comes in, and the default is **Priority**\=assess, **Milestone**\=to-assess
+    
+    1.  If user sets Component (we always should!) automatically assigns the Owner based on the component owner for the pre-assessment. \[Add link to Components and Owner\]
+        
+    2.  _At this point the Owner function is really as an assessor._
+        
+    3.  If user does not select the Component, it remains for the group TC triage with Component=to-assess, Owner=anybody. These tickets will be assessed after all the pre-assessed tickets have been triaged.
+        
+2.  All TC members, look at [To-assess by Owner / Component](https://www.google.com/url?q=https%3A%2F%2Funicode.org%2Fcldr%2Ftrac%2Freport%2F100&sa=D&sntz=1&usg=AOvVaw3RRcS9UWwOlotoRAFsV3fJ) before each triage day.
```
macchiati, line 28: You need to resolve links that start with `https://www.google.com/url?q=` first.

Easiest way is to paste into a browser's address bar; the browser will resolve it — in this case, to:

https://cldr.unicode.org/index/bug-reports#TOC-Filing-a-Ticket

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
+---
+title: Handling Tickets (bugs/enhancements)
+---
+
+# Handling Tickets (bugs/enhancements)
+
+### Filing tickets
+
+  1.  You can add set other fields, but don't change the Owner. That way the committee sees the incoming tickets (those marked as Owner: somebody).
+
```
macchiati, line 10: There are extra blank lines between list items; these should be removed.

macchiati: Also, in the Summary, add a link to the comparison between the old (in plaintext) and the new (in plaintext).

macchiati (changes requested):

macchiati:
Adding @btangmu and @srl for any comments they have

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+   1.  All reviewers should be done well before the end of the release, so that any problems can be taken care of.
+        
+    2.  At release time, all the tickets with status "fixed" are changed to "closed".
+        
+
+### Start of Release
+
+   1.  The future folder tickets are moved to the discuss folder
+        
+    2.  Unscheduled tickets (with no release number) are re-evaluated.
```
srl295, line 290:
```suggestion
   2.  Unscheduled tickets (with no release number) are re-evaluated.
```

wrong indentation for list

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+    6.  Hit "Submit changes"
+        
+4.  If any issues come up or the topic needs general review, send a message to the cldr@unicode.org list.
+    
+
+### Releasers
+
+   1.  All reviewers should be done well before the end of the release, so that any problems can be taken care of.
+        
+    2.  At release time, all the tickets with status "fixed" are changed to "closed".
```
srl295, line 283:
```suggestion
   2.  At release time, all the tickets with status "fixed" are changed to "closed".
```

check indentation so it lines up

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+*   **Make sure to run the tests in** [**Running Tests**](/development/running-tests)**!**
+    
+
+Skipping test failures
+
+If there is a test failure that is due to a bug that cannot be fixed right now (eg it requires a data change, but it is past the data freeze), then use the logKnownIssue() method to disable the test. See [http://site.icu-project.org/setup/eclipse/time](http://www.google.com/url?q=http%3A%2F%2Fsite.icu-project.org%2Fsetup%2Feclipse%2Ftime&sa=D&sntz=1&usg=AOvVaw1SrxsiozTNKEAoaq7owsCr) for details, and look in the CLDR test code for some examples of usage.
+
+### Survey Tool in Production Phase!
+
+1.  **Before** you make a commit, add a line to the Issues spreadsheet below ([Spreadsheet View](https://docs.google.com/spreadsheet/ccc?key=0AqRLrRqNEKv-dF9DdVRPVmhHdzhheDliOEpLSUhObXc#gid=3)), with status OPEN, bug number, short desc.
```
srl295, line 168: @macchiati Higher level comment: this is actually an obsolete process.

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+| dvet | The ticket should be completed before vetting starts |
+| rc  | The ticket must be completed before **Alpha — Final Data candidate**. No change to data affecting ICU thereafter. Other dtd, data, spec, docs, tool changes allowed. |
+| spec-beta | The ticket must be completed before **Beta — Final Candidate**. No dtd or data changes allowed thereafter. Docs, charts, spec changes allowed (= ICU release candidate) |
+| final | The ticket must be completed before **Release** ( = ICU Release) |
+
+### Assignees
+
+   1.  Assignees periodically review and take care of their tickets.
+        
+        1.  Go to [http://unicode.org/cldr/trac/report/61](http://www.google.com/url?q=http%3A%2F%2Funicode.org%2Fcldr%2Ftrac%2Freport%2F61&sa=D&sntz=1&usg=AOvVaw0UYmG2ygYOMySNUAGnbPKu) for **My Tickets**
```
srl295, line 110: obsolete link

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+    
+
+[Survey Tool Issues](https://docs.google.com/spreadsheets/d/0AqRLrRqNEKv-dF9DdVRPVmhHdzhheDliOEpLSUhObXc/edit)
+
+### Checkin
+
+   1.  All commits must have as their first line: "CLDR-XXX <description>", where XXX is the ticket number.
+        
+
+2.  **Don't mark the bug as fixed! Instead,**
```
srl295, line 227: list indentation issue here

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
...
+                
+
+            *   The description doesn't work
+                
+            *   Another mechanism looks better
+                
+
+### DTD Changes
+
+*   **If you are making any DTD changes, please follow the instructions on** [**Updating DTDs**](/development/updating-dtds)**.**
```
srl295, line 152:
```suggestion
*   **If you are making any DTD changes, please follow the instructions on** [**Updating DTDs**](https://cldr.unicode.org/development/updating-dtds)**.**
```

site-relative link (and there are others) but perhaps it's better to just make them absolute to start and then we can fix up en masse later?

srl295 (changes requested):

srl295:
The conversion is generally good but a lot of the content is out of date.  Getting it moved over, though is a good step towards improvement.

docs/site/development/development-process.md
```diff
@@ -0,0 +1,292 @@
+---
+title: Handling Tickets (bugs/enhancements)
+---
```
srl295, line 3: If this shows up in the output (I don't think it will) I can fix the conversion process.

macchiati:
Just to be clear, Chris's goal is to make sure that the text and formatting
are the same as the original, correcting the formatting where sites has
messed it up.

So he should not be trying to make any fixes to the text. That is for us to
do later.

docs/site/development/development-process.md
macchiati: Yes, good idea.

docs/site/development/development-process.md
macchiati: Right. We just want to get the source into markdown before fixing.

chpy04:
I believe this resolves the changes you guys were talking about. @macchiati, I am a bit confused as to what I should put for the diffs between the plain texts. I tried your suggestion with Google Docs history but it doesn't ignore whitespace so I couldn't get it to work and was just using a browser tool. Let me know what I should link to.

macchiati:
Hmmm.

We'd like to give the reviewer confidence that the plaintext is preserved.
One idea would be to normalize the spaces in both eg:

[ \t]+ => space
[ \t]\n[ \t] => \n

Alternatively, we could make a side PR, checking in the original plaintext
in one commit, and the generated plaintext in a second. Then people can use
the github diff (which can ignore whitespace).

Any other ideas?

chpy04:
I looked into it some more and it seems like the history on google docs just looks at what the user did not the differences, so even if you copy in the same text the history looks like deleting and re-entering. Another option would be to include a screenshot of my diff tool in the comment. This might not be as convincing as a PR but would be faster.

macchiati:
A screen shot would work, good idea.

PR #3713, opened 2024-05-13
CLDR-13864 Add night unit

macchiati:
CLDR-13864

Adds unit `night`, as in "Your hotel reservation is for 3 nights". Because it is metric_adjacent, will get grammar

Adds to the normal places: en.xml, root.xml, units.xml, unit.xml

Because night is not convertible to either seconds or months, it has a new quantity. That is reflected in: units.xml, DtdData.java, UnitConverter.java, TestUnits.java

Also changes checkUnitLocalePreferencesTestIcu to not cause an error in testing the preferences test data file, with a CLDR-17581 LogKnownIssue.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -7021,6 +7021,12 @@ annotations.
 				<unitPattern count="other">{0} days</unitPattern>
 				<perUnitPattern>{0} per day</perUnitPattern>
 			</unit>
+			<unit type="duration-night">
+				<displayName>nights</displayName>
+				<unitPattern count="one">{0} night</unitPattern>
+				<unitPattern count="other">{0} nights</unitPattern>
+				<perUnitPattern>{0}/night</perUnitPattern>
+			</unit>
 			<unit type="duration-hour">
 				<displayName>hours</displayName>
 				<unitPattern count="one">{0} hour</unitPattern>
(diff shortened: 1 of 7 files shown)


macchiati:
After this is merged, file a ticket to move 'special cases' into XML data. 

Example: There was a problem in TestCLDRPaths.TestReadPrevSDI because introducing a new quantity causes reading an old XML file fails.

macchiati:
ping

pedberg-icu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -1869,7 +1869,8 @@ private static final class UnitOrderHolder {
                                         "volume-sai",
                                         "volume-to-jp",
                                         "volume-koku",
-                                        "mass-fun"))
+                                        "mass-fun",
+                                        "duration-night"))
```
btangmu, line 1873: sounds like a big party!

btangmu (commented):
LGTM

common/main/en.xml
```diff
@@ -7021,6 +7021,12 @@ annotations.
 				<unitPattern count="other">{0} days</unitPattern>
 				<perUnitPattern>{0} per day</perUnitPattern>
 			</unit>
+			<unit type="duration-night">
+				<displayName>nights</displayName>
+				<unitPattern count="one">{0} night</unitPattern>
+				<unitPattern count="other">{0} nights</unitPattern>
+				<perUnitPattern>{0}/night</perUnitPattern>
```
AEApple, line 7028: @macchiati - Why is night different than day here? Day has '{0} per day'

macchiati:
Good catch. I'll make a PR to fix that.

PR #3720, opened 2024-05-14
CLDR-17634 Verify that all prior-Basic locale names have English and Coverage

srl295:
Also misc improvements:
- ToolConstants.previousVersion()
- CalculatedCoverageLevels.forVersion() for loading prior versions
- Level.isAbove()

CLDR-17634

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ToolConstants.java
@@ -62,6 +62,11 @@ public class ToolConstants {
         return last;
     }
 
+    /** Version prior to the current DEV version */
+    public static VersionInfo previousVersion() {
+        return previousVersion(DEV_VERSION_VI);
+    }
+
     public static String previousVersion(String version, int minFields) {
         VersionInfo result = previousVersion(VersionInfo.getInstance(version));
         return result.getVersionString(minFields, 2);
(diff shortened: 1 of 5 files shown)


srl295:
- Since Coverage is locale based, I’m testing in all `SupplementalDataInfo.getLanguageTcOrBasic()` locales,  that is if they are TC or Basic.  
- I’m testing to fail if any of those locales have any of the  language XPaths at level higher than `MODERN` (so fail if  `COMPREHENSIVE`)… what should the goal be here? should it fail if above MODERATE?

Example error message  (fake data, with real data it passes)

```
        org.opentest4j.AssertionFailedError: en.xml is missing translations for these languages' names:[cs, csw, cv, cy, da, de] ==> expected: <true> but was: <false>
        org.opentest4j.AssertionFailedError: coverageLevels.xml has a coverage level >modern for these language's names:[bho, brx, bs, ca, ceb, cs, cv, cy, da, doi, el, et, eu, fa] ==> expected: <true> but was: <false>
```

pedberg-icu (approved):

AEApple:
Does this mean we can deprecate the BRS item?

srl295:
> Does this mean we can deprecate the BRS item?

Retire that BRS debt!

Yes it should mean the test will fail during SBRS as soon as the version number is flipped. Could maybe make an option to make it fail early ?

macchiati (approved):

PR #3723, opened 2024-05-15
CLDR-17461 Forum posts in vetting view, change icon style

btangmu:
-Add margins and orange/green padding around 👁️‍🗨️/💬 icons, to make them easier to notice and distinguish

CLDR-17461

ALLOW_MANY_COMMITS=true


btangmu:
image

macchiati (approved):
These would definitely increase the visibility. I think the open forum posts are more interesting and need more highlight, so we might play a bit with the colors, but that can happen in a later PR.

srl295 (approved):
Also I noticed that the open/closed forum posts symbols show up even when not logged in.  Not sure that's intentional, i can add a comment to the ticket if you want

btangmu:
> even when not logged in

@srl295 there are comments about that in the ticket. I'm awaiting a decision whether the icons need to be hidden depending on user status (logged in? permission to view forum for the given locale?)

PR #3724, opened 2024-05-15
CLDR-17638 SBRS v46 CLDRModify passes before general submission

pedberg-icu:
CLDR-17638

CLDRModfi passes before general submission. Each pass was handled as a separate commit whole should be preserved when merging.
1. CLDRModify with no options; this did the following:
    * Reordered new units (speed-light-speed, concentr-portion-per-1e9, duration-night) to the end of each length group, instead of keeping them with their respective categories (this is likely due to missing/incorrect ordering data in DtdData or something)
    * Reordered exemplar entries
    * Reordered new annotation entries in root
    * Added some missing newlines at the ends of files
    * Updated headers as necessary (copyright etc)
2. CLDRModify -fQ; this did the following:
    * Normalized annotations (reordered the keywords, eliminated case duplicates). However for locales in which such modified entries had status contributed or lower, it also removed the draft status (bad). So only the changes to en.xml were committed (since it has no draft entries).
3. CLDRModify -fP (DAIP); this did the following:
    * Reformatted exemplars
    * Normalized annotations (reordered the keywords, eliminated case duplicates), duplicating what fQ is supposed to do, without removing the draft status (so the changes for all files were committed here). So that probably means -fQ is unnecessary.
    * Normalized apostrophes and hyphens, which did affect the `en` annotations

ALLOW_MANY_COMMITS=true


macchiati (approved):

btangmu (approved):
Spot-checked, LGTM. It often seems like some of these normalizations could be handled at earlier stages than CLDRModify and we'd benefit from less churn then...

macchiati:
We deliberately added emoji at the end, to make diff'ing easier.

PR #3726, opened 2024-05-15
CLDR-13864 Fix per night, long form

macchiati:
CLDR-13864

ALLOW_MANY_COMMITS=true


AEApple (approved):
LGTM! I'm assuming the unit ordering doesn't matter right now.

macchiati:
Right, that doesn't matter for now.

PR #3729, opened 2024-05-16
CLDR-16279 fix random spacing in the Numbers report for compact currency

pedberg-icu:
CLDR-16279

The Numbers report in Survey Tool (and Charts) is generated by util/VerifyCompactNumbers.java, which calls test/BuildIcuCompactDecimalFormat.java to assemble the formats. BuildIcuCompactDecimalFormat.buildCustomData() was iterating over the paths for compact number formats without paying attention to which ones were alt='alphaNextToNumber' (with spacing) and non-alt (normally without spacing), leading to random use of space in compact currency formats in the report (see the Numbers report for en_CA, for example). This PR fixes BuildIcuCompactDecimalFormat.buildCustomData() to explicitly choose the alt='alphaNextToNumber' format when appropriate (currency symbol has letter on side adjacent to value), and to skip the alt format (and use the standard format) otherwise.

It is possible that a similar change will also be needed in util/ICUServiceBuilder.java

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/BuildIcuCompactDecimalFormat.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/BuildIcuCompactDecimalFormat.java
@@ -2,6 +2,7 @@ package org.unicode.cldr.test;
 
 import com.ibm.icu.impl.number.DecimalFormatProperties;
 import com.ibm.icu.impl.number.PatternStringParser;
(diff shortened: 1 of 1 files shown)


macchiati (approved):

pedberg-icu:
After merging, verified in smoketest that this fixes the spacing issue in the Compact Numbers report.

PR #3732, opened 2024-05-19
CLDR-17199 Add missing time formats with h for several new locales

pedberg-icu:
CLDR-17199

Add missing time formats using h in several new locales so ICU tests pass (locales that inherit from root and do not have time formats will get the root formats with H, which may not match the region time-cycle preference causing an ICU error).
- css (for CA): Use the formats for en_CA but these inherit from en, so use those.
- kxv_Deva, kxv_Orya, kxv_Telu (for IN): Use the formats for kxv (Latn).
- zh_Hans_MY, zh_Hant_MY: Use Chinese-style time formats using (a before h, etc). These are in availableFormats e.g. for zh; but the are used as the standard formats in zh_Hans_SG so just copy from there.

Somewhere we have a CLDR ticket to add a CLDR-side test for this so we fix it before we get to ICU integration.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/csw.xml
+++ b/common/main/csw.xml
@@ -122,6 +122,36 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 						</dayPeriodWidth>
 					</dayPeriodContext>
 				</dayPeriods>
+				<timeFormats>
+					<timeFormatLength type="full">
+						<timeFormat>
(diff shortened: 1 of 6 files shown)


srl295 (approved):
+1 for contributed

PR #3733, opened 2024-05-19
CLDR-15764 Add examples with RTL context to reports; fix other report issues

pedberg-icu:
CLDR-15764

Add examples with RTL context (I addition to neutral context) to the date time & numbers reports where relevant. Also fix some other report issues:
- dateTime intervals involving a difference of just minutes had a bad end date.
- dateTime formats combining a relative day and a time should use the "atTime" variant of the combining pattern.
- the numbers reports did not show formats for negative numbers or zero, added those.

ALLOW_MANY_COMMITS=true


srl295 (approved):

pedberg-icu:
@srl295 Thanks for review. All looks good in smoketest except still not using the "atTime" variant of the dateTime combining format for the relative items. I will handle that with another PR, and probably another ticket since that was not actually one of the things requested in ticket for this.

pedberg-icu:
Filed https://unicode-org.atlassian.net/browse/CLDR-17653 for the atTime issue.

PR #3734, opened 2024-05-21
CLDR-10772 ST examples for month names, in appropriate format

pedberg-icu:
CLDR-10772

Add ST examples for month names showing them in appropriate context: format months in a format with day number (currently winning format for skeleton yMMMMd), and stand-alone months in a format without (currently winning format for skeleton yMMMM)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -470,6 +470,8 @@ public class ExampleGenerator {
             result = handleCurrency(xpath, parts, value);
         } else if (parts.contains("dayPeriods")) {
             result = handleDayPeriod(parts, value);
+        } else if (parts.contains("monthContext")) {
+            result = handleDateSymbol(parts, value);
         } else if (parts.contains("pattern") || parts.contains("dateFormatItem")) {
             if (parts.contains("calendar")) {
                 result = handleDateFormatItem(xpath, value, showContexts);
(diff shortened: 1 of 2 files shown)


pedberg-icu:
Once this is approved and tests pass, any TC member should feel free to merge so we can also review in smoketest.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1094,6 +1096,71 @@ private String handleDayPeriod(XPathParts parts, String value) {
...
+        int skeletonIndex = (length.equals("wide")) ? 0 : 1;
+        if (!context.equals("format")) {
+            skeletonIndex += 2;
+        }
+        String checkPath =
+                "//ldml/dates/calendars/calendar[@type=\""
+                        + calendarId
+                        + "\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\""
+                        + skeletons[skeletonIndex]
+                        + "\"]";
```
srl295, line 1123: I wonder if we should have an XPathParts Builder designed for constructing XPaths in a loop with syntax checking (i.e. that the xpath remains valid).

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
pedberg-icu: How many places would we use something like that?

pedberg-icu:
Looks good in smoketest too.

PR #3736, opened 2024-05-21
CLDR-17111 Deprecate old style zone IDs (e.g. EST5EDT) and make them …

yumaoka:
…aliases

CLDR-17111

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/bcp47/timezone.xml
+++ b/common/bcp47/timezone.xml
@@ -141,7 +141,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="cnurc" description="Ürümqi, China" alias="Asia/Urumqi Asia/Kashgar"/>
             <type name="cobog" description="Bogotá, Colombia" alias="America/Bogota"/>
             <type name="crsjo" description="Costa Rica" alias="America/Costa_Rica"/>
-            <type name="cst6cdt" description="POSIX style time zone for US Central Time" alias="CST6CDT" since="1.8"/>
+            <type name="cst6cdt" description="POSIX style time zone for US Central Time" deprecated="true" preferred="uschi"/>
             <type name="cuhav" description="Havana, Cuba" alias="America/Havana Cuba"/>
             <type name="cvrai" description="Cape Verde" alias="Atlantic/Cape_Verde"/>
             <type name="cxxch" description="Christmas Island" alias="Indian/Christmas"/>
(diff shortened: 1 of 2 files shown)


common/bcp47/timezone.xml
```diff
@@ -403,28 +403,28 @@ For terms of use, see http://www.unicode.org/copyright.html
...
+            <type name="usden" description="Denver, United States" alias="America/Denver America/Shiprock Navajo US/Mountain MST7MDT"/>
             <type name="usdet" description="Detroit, United States" alias="America/Detroit US/Michigan"/>
             <type name="ushnl" description="Honolulu, United States" alias="Pacific/Honolulu US/Hawaii Pacific/Johnston"/>
             <type name="usind" description="Indianapolis, United States" alias="America/Indianapolis America/Fort_Wayne America/Indiana/Indianapolis US/East-Indiana" iana="America/Indiana/Indianapolis"/>
             <type name="usinvev" description="Vevay (Indiana), United States" alias="America/Indiana/Vevay"/>
             <type name="usjnu" description="Juneau (Alaska), United States" alias="America/Juneau"/>
             <type name="usknx" description="Knox (Indiana), United States" alias="America/Indiana/Knox America/Knox_IN US/Indiana-Starke"/>
-            <type name="uslax" description="Los Angeles, United States" alias="America/Los_Angeles US/Pacific US/Pacific-New"/>
+            <type name="uslax" description="Los Angeles, United States" alias="America/Los_Angeles US/Pacific US/Pacific-New  PST8PDT"/>
```
justingrant, line 419:
```suggestion
            <type name="uslax" description="Los Angeles, United States" alias="America/Los_Angeles US/Pacific US/Pacific-New PST8PDT"/>
```

Typo: extra space

common/bcp47/timezone.xml
```diff
@@ -403,28 +403,28 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="umawk" description="Wake Island, U.S. Minor Outlying Islands" alias="Pacific/Wake"/>
             <type name="umjon" description="Johnston Atoll, U.S. Minor Outlying Islands" deprecated="true" preferred="ushnl"/>
             <type name="ummdy" description="Midway Islands, U.S. Minor Outlying Islands" alias="Pacific/Midway"/>
-            <type name="unk" description="Unknown time zone" alias="Etc/Unknown"/>
+            <type name="unk" description="Unknown time zone" alias="Etc/Unknown Factory"/>
```
justingrant, line 406:
```suggestion
            <type name="unk" description="Unknown time zone" alias="Etc/Unknown Factory" iana="Factory"/>
```

I think adding the `iana` attribute here was what we agreed in the meeting today?

sffc: We didn't have time to finish discussing this point but I have some concerns with using `iana="Factory"`. Since ECMAScript already returns "Etc/Unknown", I think it would be a potential web compat concern if we started returning "Factory". It might be better to just change ECMA-402 to allow "Etc/Unknown" as a return value.

justingrant: If I remember correctly from the last time I investigated this, different engines handle the Etc/Unknown value returned from ICU differently. And when we discussed the prospect of adding Etc/Unknown to ECMA-402 at a TG2 meeting (I think it was June 2023), if I remember correctly Yusuke did not support adding it. Also, Anba expressed the same anti-Etc/Unknown opinion here: https://github.com/tc39/ecma402/issues/825#issuecomment-1678675723.

I'm just looking for any cross-engine consensus for what ID to return when ICU returns Etc/Unknown. I don't have a strong opinion either way.  If we can convince the other implementers that Etc/Unknown is OK, then I'd support that.

macchiati (approved):

yumaoka:
This PR was just created for reviewing one of possible solutions. @justingrant and I could convince TZ database folks to make these legacy IDs as aliases. So this PR is no longer needed. https://github.com/unicode-org/cldr/pull/3754 created by @justingrant is reflecting coming TZ database changes (2024b).

PR #3737, opened 2024-05-21
CLDR-17111 Adding old style zone IDs (CET/EET/WET) as canonical zones

yumaoka:
CLDR-17111

ALLOW_MANY_COMMITS=true


macchiati (approved):

yumaoka:
This PR was just created for reviewing one of possible solutions. @justingrant and I could convince TZ database folks to make these legacy IDs as aliases. So this PR is no longer needed. https://github.com/unicode-org/cldr/pull/3754 created by @justingrant is reflecting coming TZ database changes (2024b).

PR #3739, opened 2024-05-22
CLDR-13523 add test for valid short day length; fix resulting data errs

pedberg-icu:
CLDR-13523

CheckDates had tests that abbreviated date symbols were no longer than wide (error), and that narrow date symbols were no longer than abbreviated (warning). But there was no test for short date symbols (only present in weekday names). Added a test that short days are no longer than abbreviated, and no shorter than narrow (both errors). Checked the with width both with a glyph width estimate (as used for the other tests) and a string length test, since the test was too sensitive otherwise (e.g. it failed on letter case differences).

Ran the test and fixed the resulting new data errors:
- `fil`: replaced current standalone narrow (same as abbreviated) with format short (shorter)
- `mai`: replaced the one provided short value (which was too long) with ↑↑↑ like the other short values, to inherit from abbreviated.
- `mgo`: swapped the narrow (in standalone) and short (in format) forms to get the right length ordering.
- `sa`: replace short (English abbreviated names) with inheritance from abbreviated (in Sanskrit)
- `so`: replaced the one provided standalone short value (too long) with ↑↑↑ like the other standalone short values, to inherit from format short which inherits from format abbreviated.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/fil.xml
+++ b/common/main/fil.xml
@@ -2246,13 +2246,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 							<day type="sat">↑↑↑</day>
 						</dayWidth>
 						<dayWidth type="narrow">
-							<day type="sun">Lin</day>
-							<day type="mon">Lun</day>
-							<day type="tue">Mar</day>
-							<day type="wed">Miy</day>
-							<day type="thu">Huw</day>
-							<day type="fri">Biy</day>
-							<day type="sat">Sab</day>
+							<day type="sun">Li</day>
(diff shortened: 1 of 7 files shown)


macchiati:
Looks good to me, but can't approve til tomorrow

pedberg-icu:
Any TC member, please feel free to merge when approved, thanks.

pedberg-icu:
PEr TC discussion 2024-05-22, decided to remove draft="contribute" and make the changes all approved.

macchiati (approved):

PR #3740, opened 2024-05-22
CLDR-17652 Import Old Votes NullPointerException and other bugs

btangmu:
-Always filter by validPaths to fix NullPointerException that was caused by obsolete/invalid paths

-Filter votes sooner to avoid repeatedly looping through votes that should be filtered

-For consistency and efficiency, move filtering code from viewOldVotes and submitOldVotes to filterPathsForManualImport, which in turn is called by getOldVotesRows, called by both viewOldVotes and submitOldVotes

-The method filterPathsForManualImport is based on old filterDowngradePaths

-In the import db table, use the value processed by current DAIP, not the old DAIP when the vote was made (although neither is guaranteed to produce the correct results in general using SQL only; this way enables further filtering with importTableContains)

-Filter out votes that match the current winning value after applying current DAIP

-Filter out votes that match already-imported value (per import table) after applying current DAIP; new method importTableContains

-Remove superfluous filter on coverage level; no level is greater than comprehensive

-Remove unused REQ_SESS

-Use sm instead of CookieSession.sm if already assigned

-Comments

CLDR-17652

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -70,7 +70,6 @@ public class SurveyAjax extends HttpServlet {
 
     private static final long serialVersionUID = 1L;
     public static final String REQ_WHAT = "what";
-    public static final String REQ_SESS = "s";
     public static final String WHAT_STATUS = "status";
     public static final String WHAT_GETSIDEWAYS = "getsideways";
     public static final String WHAT_GETXPATH = "getxpath";
@@ -643,11 +642,17 @@ public class SurveyAjax extends HttpServlet {
                         send(r, out);
                     } else if (what.equals(WHAT_OLDVOTES)) {
                         mySession.userDidAction();
-                        CookieSession.sm.getSTFactory().setupDB();
+                        sm.getSTFactory().setupDB();
                         boolean isSubmit = (request.getParameter("doSubmit") != null);
                         String confirmList = request.getParameter("confirmList");
                         SurveyJSONWrapper r = newJSONStatus(request, sm);
-                        importOldVotes(r, mySession.user, sm, isSubmit, confirmList, loc);
+                        importOldVotes(
+                                r,
+                                mySession.user,
+                                sm,
+                                isSubmit,
+                                confirmList,
+                                loc /* maybe null! */);
                         send(r, out);
                     } else if (what.equals(WHAT_GETSIDEWAYS) && l != null) {
                         mySession.userDidAction();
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1938,34 +1934,89 @@ private static Map<String, Object>[] getOldVotesRows(
...
+            }
+            if (importTableContains(localeId, xp, processedValue)) {
+                continue;
+            }
+            al.add(m);
         }
         return al.toArray(new Map[0]);
     }
 
+    private static boolean importTableContains(String localeId, int xp, String processedValue) {
```
btangmu, line 1998: This method is a last-minute addition, and makes a huge difference -- now, at least for the Filipino vetter, the manual import succeeds after "select all" and "import", with message "No old votes to import. You're done with this locale!"

macchiati (commented):
Looks good to me. Steven should review as well, because I'm not as acquainted with the code here.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1938,34 +1934,89 @@ private static Map<String, Object>[] getOldVotesRows(
...
+             * Skip abstentions (null value) and votes for inheritance.
+             * For inheritance, a candidate item is provided anyway when appropriate.
+             */
             Object obj = m.get("value");
-            String value = (obj == null) ? null : obj.toString();
+            if (obj == null) {
+                continue;
+            }
+            String rawValue = obj.toString();
+            if (rawValue == null || rawValue.equals(CldrUtility.INHERITANCE_MARKER)) {
```
srl295, line 1959: +1 on using the constant

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
srl295: I can see this is only called after a number of conditions including that DAIP actually makes a change.  Otherwise it would be expensive to call this one by one for each value.

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
btangmu: @srl295 "only called after a number of conditions including that DAIP actually makes a change" -- actually, it doesn't check that. The preceding check involves the winning value (curValue), not the old losing value (rawValue).

It is expensive, so not ideal in the long run, though testing with 11,451 old votes to import it seems pretty fast.

Possibly the condition should be changed from

`if (importTableContains(localeId, xp, processedValue)) {`

to

`if (!processedValue.equals(rawValue) && importTableContains(localeId, xp, processedValue)) {`

That would be more efficient, and would have the same result -- but only if the sql query in getOldVotesRows is guaranteed to filter out any old votes whose values were already imported when `processedValue.equals(rawValue)`. Unfortunately, it's not clear to me whether that's guaranteed, since DAIP may be different for each version... Probably it would be OK but I'd rather do further study and testing rather than assume it.

PR #3741, opened 2024-05-22
CLDR-17665 remove beaufort from GenerateUnitTestData.NOT_CONVERTABLE; 17665

pedberg-icu:
CLDR-17665

Remove `beaufort` from the NOT_CONVERTABLE list in  GenerateUnitTestData since that is no longer correct, then regenerate the unit test data. This regen left alone (as expected) the `unitsTest.txt` entry for `beaufort` which had been manually added in PR [#3562](https://github.com/unicode-org/cldr/pull/3562). It also added entries for 3 new units:
- night
- portion-per-1e9
- light-speed 

ALLOW_MANY_COMMITS=true


macchiati (approved):

macchiati:
Thanks, looks good.

macchiati:
BTW, Peter, I was thinking of handling 'generic' in the same way as some other units. That is, create its own Quantity. It would be convertible, but only to itself. That would make the code simpler for us and clients, since everything would have a quantity.

pedberg-icu:
> BTW, Peter, I was thinking of handling 'generic' in the same way as some other units. That is, create its own Quantity. It would be convertible, but only to itself. That would make the code simpler for us and clients, since everything would have a quantity.

@macchiati That seems like a great idea!

PR #3743, opened 2024-05-22
CLDR-17560 Overall Errors

srl295:
CLDR-17560

 - add new CldrError and OverallErrors components
 - add new REST endpoint for whole-locale-errors
 - remove Report for supplemental data (whole locale errors) and references thereto
 - remove link from info panel to supplemental chart
 - add whole-locale widget to General Info

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/package.json
+++ b/tools/cldr-apps/js/package.json
@@ -46,7 +46,7 @@
     "ant-design-vue": "^3.2.16",
     "browser-fs-access": "^0.35.0",
     "marked": "^4.3.0",
-    "swagger-client": "^3.28.1",
+    "swagger-client": "^3.26.7",
     "vue": "^3.2.47",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.0/xlsx-0.20.0.tgz"
   }
--- a/tools/cldr-apps/js/package.json
+++ b/tools/cldr-apps/js/package.json
@@ -46,7 +46,7 @@
     "ant-design-vue": "^3.2.16",
     "browser-fs-access": "^0.35.0",
     "marked": "^4.3.0",
-    "swagger-client": "^3.26.7",
+    "swagger-client": "^3.28.1",
     "vue": "^3.2.47",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.0/xlsx-0.20.0.tgz"
   }
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
@@ -788,11 +788,6 @@ function testsToHtml(tests) {
 
     newHtml += "</p>\n";
   }
-  if (hadEntireLocale) {
-    newHtml += `<p class='trInfo tr_Warning alert alert-warning fix-popover-help'>See also <a href='#r_supplemental/${cldrStatus.getCurrentLocale()}//'>${cldrText.get(
-      "special_r_supplemental"
-    )}</a></p>\n`;
-  }
   return newHtml;
 }
 
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
@@ -751,14 +751,12 @@ function testsToHtml(tests) {
   if (!tests) {
     return newHtml;
   }
-  var hadEntireLocale = false;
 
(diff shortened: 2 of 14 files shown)


srl295:
No output if there's no issue. 

Otherwise: 

image

srl295:
this is up on staging as of this writing

tools/cldr-apps/js/src/views/CldrError.vue
```diff
@@ -0,0 +1,59 @@
...
+        .split(/(?=[A-Z])/)
+        .map((s) => s)
+        .join(" ");
+    },
+  },
+  props: {
+    status: {
+      type: Object,
+      // any additional classes
+      default: "",
```
btangmu, line 40: should message, subtype and subtypeUrl be declared here? It looks like they're filled in from json from the back end. If that fails for any reason, the front end will throw cryptic errors like undefined subtype... I know we already have a lot of js code that makes this kind of implicit assumption. Still, objects with clearly defined and documented properties would be more ideal. I think status corresponds to back-end CheckCLDR.CheckStatus, and subtype corresponds to CheckCLDR.CheckStatus.Subtype. Comments could help...

tools/cldr-apps/js/src/views/CldrError.vue
```diff
@@ -0,0 +1,59 @@
...
+    subtypeString() {
+      return this.status.subtype
+        .split(/(?=[A-Z])/)
+        .map((s) => s)
+        .join(" ");
+    },
+  },
+  props: {
+    status: {
+      type: Object,
```
btangmu, line 38: is type an object, or a string? On the back end it's an enum, but in the json it looks very much like a string such as "Warning" -- or am I just totally confused?

btangmu (approved):
looks good aside from some minor quibbles

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
```diff
@@ -788,11 +788,6 @@ function testsToHtml(tests) {
 
     newHtml += "</p>\n";
   }
-  if (hadEntireLocale) {
```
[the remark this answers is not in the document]
srl295: true.

tools/cldr-apps/js/src/views/CldrError.vue
```diff
@@ -0,0 +1,59 @@
...
+// import * as cldrStatus from "../esm/cldrStatus.mjs";
+export default {
+  computed: {
+    myClass() {
+      return `cldrError tr_${this.status.type} `;
+    },
+    subtypeString() {
+      return this.status.subtype
+        .split(/(?=[A-Z])/)
+        .map((s) => s)
```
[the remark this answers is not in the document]
srl295: this was originally a lowercasing function. There's no titlecase in JS.

tools/cldr-apps/js/src/views/CldrError.vue
srl295: I don't think it will type check here.

tools/cldr-apps/js/src/views/CldrError.vue
srl295: It's an object.

srl295:
do we want to merge changes to https://unicode-org.atlassian.net/browse/CLDR-17664 before merging this?

tools/cldr-apps/js/package.json
```diff
@@ -46,7 +46,7 @@
     "ant-design-vue": "^3.2.16",
     "browser-fs-access": "^0.35.0",
     "marked": "^4.3.0",
-    "swagger-client": "^3.26.7",
+    "swagger-client": "^3.28.1",
```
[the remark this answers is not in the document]
srl295: it was to fix an issue with Chrome

srl295:
@btangmu i'm going to make this one draft because it's waiting on some further UI review

tools/cldr-apps/js/src/views/CldrError.vue
btangmu: I don't think we're talking about the same thing... `status` itself is an object, yes (I wrongly guessed CheckCLDR.CheckStatus but you say CheckStatusSummary and I believe you).

I was referring to `status.type`. On the back end that's `CheckCLDR.CheckStatus.Type`, which is an enum, and on the front end it's not an Object, it's a String. So, I think `type: Object` is wrong, it should be `type: String`

macchiati (approved):

PR #3744, opened 2024-05-23
CLDR-11883 update hour prefs for some en_XX locales

pedberg-icu:
CLDR-11883

Update hour cycles for some en_XX locales (while keeping the basic h vs H matching the region), which may be closer to en_001 than to the preferred hour cycle types for the default language in region XX.  See ticket for more details.

ALLOW_MANY_COMMITS=true


richgillam (approved):
LGTM

AEApple (approved):

PR #3752, opened 2024-05-24
CLDR-17658 Improve Dashboard performance and UI; work in progress

btangmu:
-For now, change only the front end, except for enabling Abstain for TC

-Abstain for TC is hidden (unchecked) by default

-Combine notifications in certain categories so that there is not more than one per page

-For now, apply this to only the Abstained category (not Missing); easy to change which categories are affected

-Add unit test

-Try to fix delayed visibility of checkbox click; alternatives and debugging for catCheckmarkChanged

-Remove delay from a-spin pending solution to delayed visibility of checkbox click

-Add Vue version to the About page; bugs may depend on Vue version

CLDR-17658

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDash.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDash.mjs
@@ -85,7 +85,7 @@ class DashData {
         const xpstrid = this.pageCombinedEntry[cat][page]; // not necessarily e.xpstrid
         if (!xpstrid) {
           console.error(
-            "Existing xpstrid not found in addAbstainedEntry for cat = " +
+            "Existing xpstrid not found in addCombinedEntry for cat = " +
               cat +
               ", page = " +
               page
@@ -95,7 +95,7 @@ class DashData {
         const dashEntry = this.pathIndex[xpstrid];
         if (!dashEntry) {
           console.error(
-            "Existing entry not found in addAbstainedEntry for cat = " +
+            "Existing entry not found in addCombinedEntry for cat = " +
               cat +
               ", page = " +
               page
--- a/tools/cldr-apps/js/src/esm/cldrDash.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDash.mjs
@@ -223,9 +223,13 @@ class DashData {
   }
 
   removeEntry(dashEntry) {
+    const xpstrid = dashEntry.xpstrid;
     this.removeEntryCats(dashEntry);
-    const index = this.entries.indexOf(dashEntry);
-    this.entries.splice(index, 1);
+    // Changed xpstrid means the entry wasn't really removed but was kept as representative of combined category
+    if (xpstrid === dashEntry.xpstrid) {
+      const index = this.entries.indexOf(dashEntry);
+      this.entries.splice(index, 1);
+    }
   }
 
   removeEntryCats(dashEntry) {
@@ -236,14 +240,14 @@ class DashData {
         delete this.catSize[cat];
       }
       if (CATS_ONE_PER_PAGE.includes(cat)) {
-        this.removeCombinedEntryCats(dashEntry, cat);
+        this.removeCombinedEntryCat(dashEntry, cat);
       } else if (this.catFirst[cat] === dashEntry.xpstrid) {
         this.findCatFirst(cat);
       }
     });
   }
 
-  removeCombinedEntryCats(dashEntry, cat) {
+  removeCombinedEntryCat(dashEntry, cat) {
     const page = dashEntry.page;
     this.pageCombinedEntries[cat][page].shift();
     if (this.pageCombinedEntries[cat][page].length > 0) {
--- a/tools/cldr-apps/js/src/esm/cldrDash.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDash.mjs
@@ -30,8 +30,7 @@ class DashData {
     // An object whose keys are xpstrid (xpath hex IDs like "db7b4f2df0427e4"), and whose values are DashEntry objects
     this.pathIndex = {};
     this.hiddenObject = null;
-    this.pageCombinedCount = {}; // map page to number of notifications on that page
-    this.pageCombinedEntry = {}; // map page to representative notification entry xpstrid on that page
+    this.pageCombinedEntries = {}; // map page to array of notification xpstrid on that page
   }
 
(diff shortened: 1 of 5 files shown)


macchiati:
I pushed to staging to try it out. Good progress!

- MUCH more responsive when there are many abstained!
- Logged in as a TC, I still don't see Abstained label (minor, but we wanted to fix this now that TCs wouldn't be overwhelmed by Abstained items)
- When I voted on an item, the Abstained dashboard row disappeared. But there were still Abstaineds on that page, and the Dashboard row it wasn't replaced by another Abstained dashboard row on that page, even after closing and opening the dashboard. [I'm guessing that that is a WIP. We were thinking of having a Dashboard row with just Section-Page that always goes to the first Abstained on the Page, but it is simpler to always just replace the one voted on by the next in the page (keeping the details of header and code, that would be fine.]

btangmu:
As discussed in infra meeting, I'm making this ready for review in spite of remaining problems. The dashboard doesn't update properly after voting. A work-around may be to refresh the dashboard frequently using the "↻" button.

tools/cldr-apps/js/src/views/AboutPanel.vue
```diff
@@ -42,12 +42,19 @@
             </span>
           </td>
         </tr>
+        <tr>
+          <td class="aboutKey">Vue version</td>
+          <td>
+            <span class="aboutValue">{{ vueVersion }}</span>
```
srl295, line 48: nice.

srl295: i wonder if there's a way to get antdv versions

tools/cldr-apps/js/src/views/DashboardWidget.vue
```diff
@@ -75,7 +75,8 @@
       </header>
       <section id="DashboardScroller" class="sidebyside-scrollable">
         <template v-if="updatingVisibility">
-          <a-spin delay="1000" size="large" />
+          <!-- for unknown reason, the a-spin fails to appear on current Chrome/Firefox if any :delay is specified here -->
+          <a-spin size="large" />
```
srl295, line 79: still couldn't repro this. but, the logistics of this update are pretty odd..

tools/cldr-apps/js/src/views/DashboardWidget.vue
```diff
@@ -367,22 +375,47 @@ export default {
...
-      this.updatingVisibility = true;
-      setTimeout(
-        () => this.updateVisibility(event.target.checked, category),
-        0
+      // Also the booleans catCheckboxIsUnchecked and catIsHidden are distinct in order for
+      // the checkbox itself to update immediately even if the rows for the corresponding
+      // category may take a long time to update.
+      // Unfortunately, neither of these mechanisms seems guaranteed to prevent a very very
+      // long delay between the time the user clicks the checkbox and the time that the checkbox
+      // changes its state.
```
srl295, line 389: I wouldn't say it's a weakness in Vue itself but a weakness in our usage of the order of 10k DOM objects.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/Dashboard.java
```diff
@@ -239,9 +239,6 @@ public ReviewOutput get(
                         sm.getSupplementalDataInfo(), sourceFactory, new STUsersChoice(sm));
         EnumSet<NotificationCategory> choiceSet =
                 VettingViewer.getDashboardNotificationCategories(usersOrg);
-        if (UserRegistry.userIsTC(user)) {
-            choiceSet.remove(NotificationCategory.abstained);
-        }
```
srl295, line 244: :100:

srl295 (approved):

macchiati:
Just got back to this; merging since Steven reviewed

tools/cldr-apps/js/src/esm/cldrDash.mjs
```diff
@@ -43,32 +52,107 @@ class DashData {
...
+        // is temporarily removed then added back, and in this case it may belong
+        // at the start of the array, not the end.
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-17658
+        if (this.updatingPath) {
+          this.pageCombinedEntries[cat][page].unshift(e.xpstrid);
+        } else {
+          this.pageCombinedEntries[cat][page].push(e.xpstrid);
+        }
+        // Use the FIRST item in the array as the representative,
+        // with its comment indicating the size of the array
```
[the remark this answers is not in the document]
btangmu: Yes, there's a ticket already: https://unicode-org.atlassian.net/browse/CLDR-15648

In the meantime, the complete (all-path) dashboard json response is sorted, however, after voting on a path, there's a miniature dash response for that path alone, and then the front end doesn't always know where to insert it...

tools/cldr-apps/js/src/esm/cldrDash.mjs
```diff
@@ -43,32 +52,107 @@ class DashData {
...
+      if (!this.pageCombinedEntries[cat][page]?.length) {
+        this.pageCombinedEntries[cat][page] = new Array(e.xpstrid);
+        this.addNewEntry(cat, group, e);
+      } else {
+        // TODO: make this work. unshift instead of push may be appropriate
+        // during an update (following a vote), since the combined notification
+        // is temporarily removed then added back, and in this case it may belong
+        // at the start of the array, not the end.
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-17658
+        if (this.updatingPath) {
```
btangmu, line 88: Bug: this.updatingPath is always false here, because here "this" = dashData (as intended), but below when this.updatingPath is set to true, "this" = cldrDash

tools/cldr-apps/js/src/esm/cldrDash.mjs
```diff
@@ -334,6 +452,7 @@ function convertData(json) {
  *               containing notifications for a single path (old format)
  */
 function updatePath(dashData, json) {
+  this.updatingPath = true;
```
btangmu, line 455: Bug here: "this" = cldrDash NOT dashData

btangmu: Note: the "this" bug was fixed in https://github.com/unicode-org/cldr/pull/3777

PR #3753, opened 2024-05-26
CLDR-17673 update plural forms for Sicilian (scn)

dpaleino:
CLDR-17673

- [ X] This PR completes the ticket.

ALLOW_MANY_COMMITS=true


conradarcturus (changes requested):
Thanks for making the change! Out of curiosity what are some examples of the difference?

plurals.xml already has an entry for `scn` -- can you remove the one on line 30?

We will need to regenerate `pluralRanges.xml` -- to do that run:
`java -jar tools/cldr-code/target/cldr-code.jar GeneratePluralRanges`

conradarcturus:
Apologies for the delays in getting this review -- we ended up manually pushing it forward so we could get it in v46 as that closes, see https://github.com/unicode-org/cldr/pull/3942

PR #3757, opened 2024-05-28
CLDR-17993 Update spotless to 2.43 and google-java-style 1.22 to support Java 21+

echeran:
(see: unicode-org/unicodetools#828)

CLDR-17993 - catch-all ticket for v47 dependency updates

ALLOW_MANY_COMMITS=true


srl295 (changes requested):
I don't think we should do this now. maybe at release. I'm concerned about disruption to in-flight PRs

echeran:
That's fine. Feel free to merge whenever you think it will be least disruptive.

echeran:
@srl295 When is a good time to merge? To cause the least disruption, I think either: 1) the last commit before cutting a release, or 2) the first commit at the beginning of a new release.

I'll leave it to you. I can help keep this PR up to date until then. If not sometime like the above, then there won't be a better time to wait for in order to merge.

srl295:
Probably right after v46 is cut.

echeran:
> Probably right after v46 is cut.

@srl295 It's almost that time. I'm about to cut ICU v76. Merge when it is convenient.

srl295 (approved):

echeran:
Thanks for the approval. Given that v46 is merged, and the Jira ticket was for a v46 BRS task, I'm going to update the PR to refer to the equivalent v47 Jira ticket.

PR #3758, opened 2024-05-28
CLDR-17263 add zh-Latn

srl295:
CLDR-17263

add:

- [X] zh-Latn-CN
- [ ] ko-Hani-KR - in https://github.com/unicode-org/cldr/pull/3759
- [ ] vi-Hani-VN - in https://github.com/unicode-org/cldr/pull/3759

ALLOW_MANY_COMMITS=true


srl295:
Ready.

DavidLRowe:
It might be clearer in the main exemplar to use `èéê` rather than `è-ê`, but not a big deal. In fact, it probably gets normalized at some later point, so don't bother changing here.

The "numbers" exemplar generally includes decimal separator and grouping separator, and often plus, minus and percentage symbols. If these are not added to this exemplar, will the vetter be blocked from entering them in the Survey Tool?

For reference, zh.xml has:
`[\- ‑ , . % ‰ + 0 1 2 3 4 5 6 7 8 9 〇 一 七 三 九 二 五 八 六 四]`

Other than that, it seems fine.

srl295:
Yes, all of these get normalized… And no, it won’t prevent anyone from
entering contact, it will produce a warning that we can sort out later

DavidLRowe (approved):
LGTM

PR #3761, opened 2024-05-29
CLDR-15744 make sure settings.user.touch() is called

srl295:
CLDR-15744

Verified:
- voting and forum posts, etc, update the 'last seen' date
- login/logout through different methods affect last seen
- make sure the named session gets logged out, even if cookies were different

ALLOW_MANY_COMMITS=true


srl295:
needs more testing

macchiati (changes requested):
This appears to be the wrong ticket. Am I missing something?

srl295:
> This appears to be the wrong ticket. Am I missing something?

good catch

macchiati (approved):

PR #3762, opened 2024-05-29
CLDR-17662 update Vetting Participation spreadsheet

srl295:
CLDR-17662 - Vetting Participation

Does not address all items, but many improvements

- major speedup due to parallel fetch of endpoint. 
- delete unneeded columns. (Can be present but hidden when debugging)
- use Org coverage instead of Cldr coverage - this is cached, so should be OK to fetch
- add sheets for Organization and Coverage goals
- move all non-direct vote counts to 'non-direct votes' (summing imported, bulk, etc)
- fix progress percent in sheet to be an actual percent, not a string

Minor:
- improve robustness of ST, account, and vetting participation pages.
These are exceptions I hit along the development path. Reduce the number of places where the tool just hangs.

ALLOW_MANY_COMMITS=true


macchiati:
Will want to see the result before approving. Maybe a screen shot (privately, because data)

srl295:
> Will want to see the result before approving. Maybe a screen shot (privately, because data)

I'll send you an updated spreadsheet

macchiati:
Ok, let's go ahead and merge.

macchiati (approved):

PR #3763, opened 2024-05-29
CLDR-17462 update ST status icons for clarity

srl295:
CLDR-17462

ALLOW_MANY_COMMITS=true

image

image_360

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -682,7 +682,10 @@ function getStatusIcon(statusClass) {
       return cldrText.get(`status_${statusClass}`);
     case "inherited-provisional":
     case "inherited-unconfirmed":
-      return cldrText.get("inherited");
+      return (
+        cldrText.get(`status_inherited`) +
+        getStatusIcon(statusClass.split("-")[1])
+      );
     default:
       return "\ufffd";
   }
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -684,6 +684,7 @@ function getStatusIcon(statusClass) {
     case "inherited-unconfirmed":
       return (
         cldrText.get(`status_inherited`) +
+        "\u202F" +
         getStatusIcon(statusClass.split("-")[1])
       );
     default:
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -684,7 +684,7 @@ function getStatusIcon(statusClass) {
     case "inherited-unconfirmed":
       return cldrText.get("inherited");
     default:
-      return "?"; // U+003F
+      return "\ufffd";
   }
 }
 
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -676,14 +676,13 @@ function getStatusIcon(statusClass) {
   switch (statusClass) {
     case "approved":
     case "contributed":
-      return "✓"; // U+2713
     case "missing":
     case "provisional":
     case "unconfirmed":
-      return "✘"; // U+2718
+      return cldrText.get(`status_${statusClass}`);
     case "inherited-provisional":
     case "inherited-unconfirmed":
-      return "↑"; // U+2191
+      return cldrText.get("inherited");
     default:
       return "?"; // U+003F
   }
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -684,7 +684,7 @@ function getStatusIcon(statusClass) {
     case "inherited-unconfirmed":
       return (
         cldrText.get(`status_inherited`) +
-        "\u202F" +
+        "\u200B" +
         getStatusIcon(statusClass.split("-")[1])
       );
     default:
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
@@ -209,21 +209,12 @@ public class SurveyTool extends HttpServlet {
 
     private void includeCss(HttpServletRequest request, PrintWriter out) {
         final String contextPath = request.getContextPath();
-        final String cb = getCacheBustingExtension(request);
-        out.write(
-                "<link rel='stylesheet' href='" + contextPath + "/surveytool" + cb + ".css' />\n");
         /*
          * Note: cldrForum.css is loaded through webpack
          */
         // bootstrap.min.css -- cf. bootstrap.min.js elsewhere in this file
         out.write(
                 "<link rel='stylesheet' href='//stackpath.bootstrapcdn.com/bootswatch/3.1.1/spacelab/bootstrap.min.css' />\n");
-        out.write(
-                "<link rel='stylesheet' href='"
-                        + contextPath
-                        + "/css/redesign"
-                        + cb
-                        + ".css' />\n");
     }
 
     private static final String DD_CLIENT_TOKEN = System.getenv("DD_CLIENT_TOKEN");
--- a/tools/cldr-apps/js/src/esm/cldrText.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrText.mjs
@@ -472,10 +472,10 @@ const strings = {
   report_missing: "Missing",
 
   // for approval status, see VoteResolver.Status
-  status_approved: "✅",
+  status_approved: "✅\uFE0F",
   status_contributed: "☑️",
   status_provisional: "✖️",
-  status_unconfirmed: "❌",
+  status_unconfirmed: "❌\uFE0F",
   status_missing: "🕳️",
   status_inherited: "⬆️",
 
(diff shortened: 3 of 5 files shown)


macchiati:
Has this been pushed to staging for UI review?

srl295:
> Has this been pushed to staging for UI review?

Yes and you reviewed it, i'm preparing another push incorporating your feedback (such as the word 'inherited')

srl295:
how's this for inherited+provisional? 
image

srl295:
@macchiati i've now updated staging to e39d2b80a8aa128491e26ef6ea340535d6b4951c

macchiati:
> how's this for inherited+provisional? image

I think that is ok, if you have a narrow space between them. That will allow for narrower columns

srl295:
> > how's this for inherited+provisional? image
> 
> I think that is ok, if you have a narrow space between them. That will allow for narrower columns

maybe a NNBSP

srl295:
> I didn't mean NBSP. That will force all columns on the page to be wide if one of them is. I was thinking the reverse; adding a narrow space (aka thin space) to allow the two icons to be on different lines. Not a non-break one, a breaking one.

different lines could be hard to scan, it'll look like alternating content.  Here's the result with ZWSP:

image

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -209,21 +209,12 @@ private void serveRunnningNormallyPage(
 
     private void includeCss(HttpServletRequest request, PrintWriter out) {
         final String contextPath = request.getContextPath();
-        final String cb = getCacheBustingExtension(request);
-        out.write(
-                "<link rel='stylesheet' href='" + contextPath + "/surveytool" + cb + ".css' />\n");
```
srl295, line 214: This wasn't working properly, tended to cache the out of date file.  Webpack worked correctly.

macchiati:
Hmmm. Ok, let's try with them on the same line, and we'll ask the TC about it. Note: that only happens when the item being inherited is provisional or worse, right? If it is Contributes or Accepted, we see just a regular ⬆️, right?

stenshamn:
> > I didn't mean NBSP. That will force all columns on the page to be wide if one of them is. I was thinking the reverse; adding a narrow space (aka thin space) to allow the two icons to be on different lines. Not a non-break one, a breaking one.
> 
> different lines could be hard to scan, it'll look like alternating content. Here's the result with ZWSP:
> 
> image

This looks OK to me, FWIW. I agree it's better than causing change in column width.

srl295:
> Hmmm. Ok, let's try with them on the same line, and we'll ask the TC about it. Note: that only happens when the item being inherited is provisional or worse, right? If it is Contributes or Accepted, we see just a regular ⬆️, right?

If I'm reading the logic correctly… only unconfirmed or provisional adds the "inherited" marker.  Otherwise it will only show the approved, contributed, missing symbols.

I don't see a code path where it shows only the uparrow.

macchiati:
If that is the case, we don't need the ⬆️ at all, just the X.

People will be able to see in the winning cell (from the color) that it is
inherited, and the X tells them that it isn't approved/contributed

srl295:
> If that is the case, we don't need the ⬆️ at all, just the X. People will be able to see in the winning cell (from the color) that it is inherited, and the X tells them that it isn't approved/contributed

That's fine, we can drop the inheritance marker symbol then. (Reference to its discussion: CLDR-11103 )

srl295:
So, thoughts? drop the inheritance icon?

macchiati:
Let's leave it in for now, so we don't delay.

macchiati (approved):

srl295:
@macchiati just to be really clear, you are approving  

## A: NNBSP [e0a503e](https://github.com/unicode-org/cldr/pull/3763/commits/e0a503e3f67c051579f02a7a53912c483911bc91) - this is the current code in the PR

> image

## B: https://github.com/unicode-org/cldr/commit/e39d2b80a8aa128491e26ef6ea340535d6b4951c what is currently on cldr-staging

## C: (not committed) using a ZWSP instead of NNBSP

> image

srl295:
> Let's leave it in for now, so we don't delay.

let me know A B or C and i'll accept the ticket and get ready to merge

macchiati:
Ok, if C: (not committed) using a ZWSP instead of NNBSP is ready to go, let's do that.

srl295:
@macchiati it's option C now

AEApple:
Looks nice!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
```diff
@@ -209,21 +209,12 @@ private void serveRunnningNormallyPage(
...
          */
         // bootstrap.min.css -- cf. bootstrap.min.js elsewhere in this file
         out.write(
                 "<link rel='stylesheet' href='//stackpath.bootstrapcdn.com/bootswatch/3.1.1/spacelab/bootstrap.min.css' />\n");
-        out.write(
-                "<link rel='stylesheet' href='"
-                        + contextPath
-                        + "/css/redesign"
-                        + cb
-                        + ".css' />\n");
```
btangmu, line 226: are you sure this isn't still needed by any legacy code, such as jsp?

I searched for "redesign" with a shortcut I like to use:

```
$ cldrgrep redesign
tools/cldr-apps/js/src/esm/cldrEvent.mjs: * (this is mostly derived from old redesign.js and should be divided
tools/cldr-apps/js/src/esm/cldrInfo.mjs:const ITEM_INFO_ID = "itemInfo"; // must match redesign.css
tools/cldr-apps/js/src/esm/cldrInfo.mjs:const ITEM_INFO_CLASS = "sidebyside-scrollable"; // must match redesign.css, cldrGui.mjs, DashboardWidget.vue
tools/cldr-apps/src/main/webapp/css/redesign.css:/* redesign.css contains a lot of the 'bootstrap' type css,
tools/cldr-apps/src/main/webapp/css/redesign.css: The following is here in redesign.css temporarily!
tools/cldr-apps/src/main/webapp/css/redesign.css: Note: redesign.css is loaded by SurveyTool.java AFTER surveytool.css and bootstrap, so has higher priority;
```
That confirms it doesn't occur in any jsp files. It points to this comment in redesign.css itself:

```
/***
 The following is here in redesign.css temporarily!
 Note: redesign.css is loaded by SurveyTool.java AFTER surveytool.css and bootstrap, so has higher priority;
 style.css (for webpack, currently empty) is effectively loaded first, has the lowest priority
***/

body,
html {
	width: 100vw !important;
	height: 100vh !important;
	overflow: hidden !important;
	margin: 0 !important;
	padding: 0 !important;
}
```

I might have written part of that comment myself a long time ago. What I'm wondering is, does the relative priority between these css files (redesign.css, surveytool.css, bootstrap, style.css) matter, and if it does matter, does this change preserve it?

btangmu: BTW I have
```
$ cat ~/bin/cldrgrep
#!/bin/bash
egrep -H "$@" tools/cldr-apps/js/package.json
egrep -H "$@" tools/cldr-apps/js/src/* -d skip
egrep -H "$@" tools/cldr-apps/js/src/esm/*
egrep -H "$@" tools/cldr-apps/js/src/views/*
egrep -H "$@" tools/cldr-apps/src/main/webapp/*.css
egrep -H "$@" tools/cldr-apps/src/main/webapp/css/*.css
egrep -H "$@" tools/cldr-apps/src/main/webapp/*.jsp
egrep -H "$@" tools/cldr-apps/src/main/webapp/WEB-INF/tmpl/*.jsp
egrep -H "$@" tools/cldr-apps/src/main/webapp/WEB-INF/jspf/*.jspf
```

macchiati (approved):

btangmu (approved):
I have some doubts about getCacheBustingExtension and redesign.css, see my comments on those; otherwise looks fine

tools/cldr-apps/js/src/esm/cldrTable.mjs
```diff
@@ -676,16 +676,19 @@ function getStatusIcon(statusClass) {
...
     case "inherited-unconfirmed":
-      return "↑"; // U+2191
+      return (
+        cldrText.get(`status_inherited`) +
+        "\u200B" +
+        getStatusIcon(statusClass.split("-")[1])
+      );
     default:
-      return "?"; // U+003F
+      return "\ufffd";
```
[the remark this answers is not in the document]
srl295: that's true, but this is an internal error of some sort.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
srl295: It is used when loading the error and waiting pages still, since the bundle may not be loaded properly.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyTool.java
srl295: The JSPs don't go through the 'serve normally' route… they would be loaded from their own jsp path.

This change makes sure our css files are loaded last and so can have the final override.

tools/cldr-apps/js/src/esm/cldrTable.mjs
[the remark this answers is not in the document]
btangmu: not an encoding error though; I think it should be reserved for errors in encoding, such as illegal sequences of UTF-8 bytes, or characters from other character sets with no corresponding code in Unicode

PR #3767, opened 2024-05-30
CLDR-17680 fix spurious CheckConsistentCasing and CheckCoverage

srl295:
CLDR-17680

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckConsistentCasing.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckConsistentCasing.java
@@ -64,18 +64,18 @@ public class CheckConsistentCasing extends FactoryCheckCLDR {
             } catch (Exception e) {
                 types = Collections.emptyMap();
             }
+            if ((types == null || types.isEmpty()) && !SpecialLocales.isScratchLocale(locale)) {
+                possibleErrors.add(
+                        new CheckStatus()
+                                .setCause(this)
+                                .setMainType(CheckStatus.warningType)
+                                .setSubtype(Subtype.incorrectCasing)
+                                .setMessage("Could not load casing info for {0}", locale));
+            }
         } else {
             // no casing info - since the types Map is global, and null checks aren't done,
             // we are better off  with an empty map here
             types = Collections.emptyMap();
-        }
-        if ((types == null || types.isEmpty()) && !SpecialLocales.isScratchLocale(locale)) {
-            possibleErrors.add(
-                    new CheckStatus()
-                            .setCause(this)
-                            .setMainType(CheckStatus.warningType)
-                            .setSubtype(Subtype.incorrectCasing)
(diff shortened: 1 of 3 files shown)


macchiati (commented):
It is hard to tell whether this is removing the errors from the main page; it doesn't appear that it does.

macchiati (changes requested):
Based on your email, this doesn't remove the messages from the main pages.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckConsistentCasing.java
```diff
@@ -64,19 +64,19 @@ public CheckCLDR handleSetCldrFileToCheck(
...
             types = Collections.emptyMap();
         }
-        if ((types == null || types.isEmpty()) && !SpecialLocales.isScratchLocale(locale)) {
-            possibleErrors.add(
-                    new CheckStatus()
-                            .setCause(this)
-                            .setMainType(CheckStatus.warningType)
-                            .setSubtype(Subtype.incorrectCasing)
-                            .setMessage("Could not load casing info for {0}", locale));
-        }
```
srl295, line 79: Previously, this warning would fire for Arabic, etc. because the script is not cased, so control would reach line 70.  Then types is checked again on this line and the warning fires.  

The change moves the warning to line 67, so it fires only if casing is expected but not present.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCoverage.java
```diff
@@ -131,9 +131,9 @@ public CheckCLDR handleSetCldrFileToCheck(
             supplementalData =
                     SupplementalDataInfo.getInstance(cldrFileToCheck.getSupplementalDirectory());
             coverageLevel = CoverageLevel2.getInstance(supplementalData, localeID);
-            PluralInfo pluralInfo = supplementalData.getPlurals(PluralType.cardinal, localeID);
-            if (pluralInfo == supplementalData.getPlurals(PluralType.cardinal, LocaleNames.ROOT)
-                    && !SpecialLocales.isScratchLocale(localeID)) {
+            PluralInfo pluralInfo =
+                    supplementalData.getPlurals(PluralType.cardinal, localeID, false);
+            if (pluralInfo == null && !SpecialLocales.isScratchLocale(localeID)) {
```
srl295, line 136: Previously, the warning would always fire for Japanese, etc, because the plural rules matched root. This changes to only fire if plural rules were not found.

srl295:
> Based on your email, this doesn't remove the messages from the main pages.

It now does

srl295:
> It is hard to tell whether this is removing the errors from the main page; it doesn't appear that it does.

I thought issue was just where it was incorrect. It now should remove from all

macchiati (approved):

PR #3774, opened 2024-06-03
CLDR-17686 change enum.valueOf to enum.forString in a couple of places

srl295:
- these were causing odd compilation errors on one environment

CLDR-17686

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
@@ -185,9 +185,8 @@ public class VoteAPIHelper {
         String xp = null;
 
         if (args.xpstrid == null && args.page != null) {
-            try {
-                pageId = PageId.valueOf(args.page);
-            } catch (IllegalArgumentException iae) {
+            pageId = PageId.forString(args.page);
+            if (pageId == null) {
                 throw new SurveyException(ErrorCode.E_BAD_SECTION);
             }
             if (pageId.getSectionId() == org.unicode.cldr.util.PathHeader.SectionId.Special) {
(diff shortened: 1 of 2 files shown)


srl295:
Deleting maven cache seemed to make this error go away.
So, not sure of the cause.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
```diff
@@ -185,9 +185,8 @@ public static RowResponse getRowsResponse(
         String xp = null;
 
         if (args.xpstrid == null && args.page != null) {
-            try {
-                pageId = PageId.valueOf(args.page);
-            } catch (IllegalArgumentException iae) {
+            pageId = PageId.forString(args.page);
+            if (pageId == null) {
```
btangmu, line 189: I guess this returns null if it catches the exception?

```
        public static PageId forString(String name) {
            try {
                return PageIdNames.forString(name);
            } catch (Exception e) {
                throw new ICUException("No PageId for " + name, e);
            }
        }
```

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
btangmu: that's a feature of Java I wasn't familiar with

btangmu: Oh, right; I misread it. It catches IllegalArgumentException then throws ICUException.

So the removal of `catch (Throwable t)` means that searchPathheader may throw ICUException where before it wouldn't have. That goes beyond merely changing valueOf to forString. Here `q` comes from an http request, so it could be anything... Line 765 of SurveyAjax will catch it:

```
        } catch (Throwable e) {
            SurveyLog.logException(
                    logger, e, "Error in SurveyAjax?what=" + what + ": " + e.getMessage());
            sendError(out, e);
        }
```

I guess that's OK, though shouldn't searchPathheader then be declared to throw ICUException? It still silently catches other exceptions further below; it just seems a little arbitrary...

btangmu (approved):

macchiati:
Please take a pass through PRs to make sure that they are merged before we branch (if there are other merges on the ticket)

PR #3775, opened 2024-06-03
CLDR-17560 improve Dashboard robustness

srl295:
- use second param to callback() to return err instead of global
- popup err message if dashboard fails to load

CLDR-17560

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/DashboardWidget.vue
+++ b/tools/cldr-apps/js/src/views/DashboardWidget.vue
@@ -203,7 +203,7 @@ export default {
     };
   },
 
-  created() {
+  mounted() {
     if (cldrStatus.getPermissions()?.userIsTC) {
       this.catIsHidden["Abstained"] = this.catCheckboxIsUnchecked[
         "Abstained"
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/js/src/views/DashboardWidget.vue
```diff
@@ -255,22 +255,23 @@ export default {
...
-        this.fetchErr = "Please log in to see the Dashboard.";
+        this.fetchErr.value = "Please log in to see the Dashboard.";
         return;
       }
       this.locale = cldrStatus.getCurrentLocale();
       this.level = cldrCoverage.effectiveName(this.locale);
       if (!this.locale || !this.level) {
-        this.fetchErr = "Please choose a locale and a coverage level first.";
+        this.fetchErr.value =
+          "Please choose a locale and a coverage level first.";
```
btangmu, line 265: Does this work right in conjunction with `<p v-if="fetchErr" class="st-sad">{{ fetchErr }}</p>`?

In other words, will `{{ fetchErr }}` automatically show fetchErr.value?

btangmu (commented):
will `{{ fetchErr }}` automatically show `fetchErr.value`?

if so, then this all looks good

btangmu (approved):
Maybe make fetchErr.value explicit where fetchErr is referenced near top of DashboardWidget.vue? And/or add a comment. We should test carefully after merging since there have been changes to these files in the meantime for virtual scroller; still it looks to me like they shouldn't be in conflict.

The change from created() to mounted() looks potentially important, I wonder if it might affect the "blinky" appearance we were discussing...

srl295:
> Maybe make fetchErr.value explicit where fetchErr is referenced near top of DashboardWidget.vue? And/or add a comment. We should test carefully after merging since there have been changes to these files in the meantime for virtual scroller; still it looks to me like they shouldn't be in conflict.

Right.

> The change from created() to mounted() looks potentially important, I wonder if it might affect the "blinky" appearance we were discussing...

https://vuejs.org/guide/essentials/lifecycle.html actually, `created()` might be better so that it does NOT blink.

AEApple:
This PR should either be abandoned or a new ticket created targetting a later release.

tools/cldr-apps/js/src/views/DashboardWidget.vue
srl295: yes - `.value` is the contents of a Vue reference.

tools/cldr-apps/js/src/views/DashboardWidget.vue
```diff
@@ -226,7 +226,7 @@ export default {
     };
   },
 
-  created() {
+  mounted() {
```
btangmu, line 229: Curious if this affects appearance, any delay/blinking/etc

btangmu (approved):

tools/cldr-apps/js/src/views/DashboardWidget.vue
srl295: Mounted is after the dom is available

PR #3783, opened 2024-06-05
CLDR-17226 UCA 16 beta jun05

markusicu:
CLDR-17226

Root collation data from Unicode 16.0 beta as of 2024-jun-05. (Fresh run of Unicode Tools "UCA Main", not yet otherwise publicly posted. Same tools run as for the DUCET test data: https://github.com/unicode-org/unicodetools/pull/852)

Jump from 15.1 to 16.0, adding thousands of new characters and changing most of the primary collation weights, mostly unreviewable...
If needed, I can send modified files for order comparison.

This includes the changes corresponding to
- https://github.com/unicode-org/unicodetools/pull/762

which includes CLDR-17468 “sort U+20A8 RUPEE SIGN & U+FDFC RIAL SIGN like their letter sequences”

----
The second commit updates the Unihan collators (common/collation/zh.xml) and common/transforms/Han-Latin.xml. (https://github.com/unicode-org/unicodetools/blob/main/docs/unihan.md)

ALLOW_MANY_COMMITS=true


macchiati (approved):

srl295:
> @macchiati @srl295 I had intended to preserve the two separate commits here, in a rebase-and-merge. Is that not allowed in CLDR at all?

Normally no. 

I'm going to propose scrapping linear history and going to only merge commits. But that's a bigger project to document and explain.

markusicu:
> I'm going to propose scrapping linear history and going to only merge commits.

That's not what I am asking for. I am rather fond of linear history on the main branch.

PR #3791, opened 2024-06-07
CLDR-17689 Dashboard download has empty Category column

btangmu:
-Replace non-existent cat with comma-concatenated cats

-Example: Missing, Warning, Abstained

CLDR-17689

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDash.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDash.mjs
@@ -446,8 +446,9 @@ async function downloadXlsx(data, locale, cb) {
     }
     const xpath = await getXpath();
     const url = `https://st.unicode.org/cldr-apps/v#/${e.locale}/${e.page}/${e.xpstrid}`;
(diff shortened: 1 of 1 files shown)


srl295 (approved):

btangmu:
While testing this I noticed that this loop can take a very long time, to the extent that the browser seems frozen:
```
  // Fetch all XPaths in parallel since it'll take a while
  cb(`Loading rows…`);
  const allXpaths = [];
  for (let dashEntry of entries) {
    if (dashEntry.section === "Reports") {
      continue; // skip this
    }
    allXpaths.push(dashEntry.xpstrid);
    try {
      await xpathMap.get(dashEntry.xpstrid);
    } catch (e) {
      throw Error(
        `${e}:  Could not load XPath for ${JSON.stringify(dashEntry)}`
      );
    }
  }
```

Seems like that should go in a separate module and be cached. There must be other places where the front end could use a mapping between xpstrid and xpath, and there are probably json responses that could be made smaller by reducing redundant data.

PR #3795, opened 2024-06-11
CLDR-16835 Non-TC locales remain open longer

srl295:
CLDR-16835

- New API `SubmissionLocales.isTcLocale()` and tests:
    - `root` is a TC locale
    - Anything listed in `TC_ORG_LOCALES` is a TC locale, such as, say, `fr`.
    - Any sublocale of a TC locale is a TC locale, such as, say, `fr_CA` or `fr_CM`
    - Anything else is NOT a TC locale, such as `csw`.
- a new config parameter: `CLDR_DDL_PHASE`.  This may NOT be set in main phase READONLY.
    - This new DDL phase only affects locales that are not TC locales.
- Internal API to get this phase, such as SurveyMain.phase() and SurveyMain.getCPhase() taking a CLDRLocale.
- UI where the DDL phase is shown with a notice, ONLY if it is different from the main phase. No UI change for other users or locales.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/MainHeader.vue
+++ b/tools/cldr-apps/js/src/views/MainHeader.vue
@@ -1,7 +1,8 @@
 <template>
   <header id="st-header">
+    <a-spin v-if="!loaded" :delay="250" />
     <ul>
-      <li>{{ stVersionPhase }}</li>
+      <li>{{ stVersion }} {{ stPhase }}</li>
       <li>
         <a href="#menu///"><span class="main-menu-icon">☰</span></a>
       </li>
@@ -79,6 +80,7 @@
 <script>
 import * as cldrAnnounce from "../esm/cldrAnnounce.mjs";
 import * as cldrCoverage from "../esm/cldrCoverage.mjs";
+import * as cldrLoad from "../esm/cldrLoad.mjs";
 import * as cldrMenu from "../esm/cldrMenu.mjs";
 import * as cldrStatus from "../esm/cldrStatus.mjs";
 import * as cldrText from "../esm/cldrText.mjs";
@@ -87,6 +89,7 @@ import * as cldrVote from "../esm/cldrVote.mjs";
 export default {
   data() {
     return {
+      loaded: false,
       announcementsTitle: null,
       coverageLevel: null,
       coverageMenu: [],
@@ -96,7 +99,9 @@ export default {
       orgCoverage: null,
       sessionMessage: null,
       specialHeader: null,
-      stVersionPhase: null,
+      stPhase: null,
+      stVersion: null,
+      tcLocale: true,
       unreadAnnouncementCount: 0,
       userName: null,
       voteCountMenu: null,
@@ -105,7 +110,12 @@ export default {
   },
 
   mounted() {
-    this.updateData();
+    // load after the localemap is ready
+    cldrLoad.onLocaleMapReady(() => {
+      this.updateData();
+    });
+    // reload if locale changes
+    cldrStatus.on("locale", () => this.updateData());
   },
 
   methods: {
@@ -114,9 +124,9 @@ export default {
      * This function is called both locally, to initialize, and from other module(s), to update.
      */
     updateData() {
-      const orgCoverage = cldrCoverage.getSurveyOrgCov(
-        cldrStatus.getCurrentLocale()
-      );
+      this.loaded = true;
+      const loc = cldrStatus.getCurrentLocale();
+      const orgCoverage = cldrCoverage.getSurveyOrgCov(loc);
       if (orgCoverage != this.orgCoverage) {
         this.orgCoverage = orgCoverage;
       }
@@ -155,11 +165,13 @@ export default {
       }
       this.sessionMessage = cldrStatus.getSessionMessage();
       this.specialHeader = cldrStatus.getSpecialHeader();
-      this.stVersionPhase =
-        "Survey Tool " +
-        cldrStatus.getNewVersion() +
-        " " +
-        cldrStatus.getPhase();
+      this.stVersion = "Survey Tool " + cldrStatus.getNewVersion();
+      this.tcLocale = cldrLoad?.getLocaleInfo(loc)?.tc;
+      if (!loc || this.tcLocale) {
+        this.stPhase = cldrStatus.getPhase();
+      } else {
+        this.stPhase = cldrStatus.getDdlPhase();
+      }
       cldrAnnounce.getUnreadCount(this.setUnreadCount);
     },
 
--- a/tools/cldr-apps/js/src/views/MainHeader.vue
+++ b/tools/cldr-apps/js/src/views/MainHeader.vue
@@ -2,7 +2,11 @@
   <header id="st-header">
     <a-spin v-if="!loaded" :delay="250" />
     <ul>
-      <li>{{ stVersion }} {{ stPhase }}</li>
+      <li>{{ stVersion }} {{ stPhase }}
+        <span class="ddlException" v-if="ddlException" title="Note: As a non-TC DDL locale, this phase has been extended.">
+          (extended)
+        </span>
+      </li>
       <li>
         <a href="#menu///"><span class="main-menu-icon">☰</span></a>
       </li>
@@ -102,6 +106,7 @@ export default {
       stPhase: null,
       stVersion: null,
       tcLocale: true,
+      ddlException: false,
       unreadAnnouncementCount: 0,
       userName: null,
       voteCountMenu: null,
@@ -169,8 +174,14 @@ export default {
       this.tcLocale = cldrLoad?.getLocaleInfo(loc)?.tc;
       if (!loc || this.tcLocale) {
         this.stPhase = cldrStatus.getPhase();
+        this.ddlException = false;
       } else {
         this.stPhase = cldrStatus.getDdlPhase();
+        if (cldrStatus.getDdlPhase() != cldrStatus.getPhase()) {
+          this.ddlException = true;
+        } else {
+          this.ddlException = false;
+        }
       }
       cldrAnnounce.getUnreadCount(this.setUnreadCount);
     },
@@ -273,4 +284,9 @@ label {
 #coverageLevel {
   width: 16ch;
 }
+
+.ddlException {
+  background-color: yellow;
+}
+
 </style>
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
@@ -1119,7 +1119,7 @@ public class SurveyAjax extends HttpServlet {
             locale.put("highestParent", loc.getHighestNonrootParent());
             locale.put("dcParent", dcParent);
             locale.put("dcChild", dcChild);
-            locale.put("tc", SubmissionLocales.TC_ORG_LOCALES.contains(loc.getBaseName()));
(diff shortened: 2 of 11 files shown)


srl295:
In the examples below, the vetter has permissions for both `csw` (non-tc) and `fr` (tc)

### Example: CLDR_PHASE=VETTING but CLDR_DDL_PHASE=SUBMIT


- `csw`

image

The yellow highlight has a tooltip, "Note, as a non-TC (DDL) locale, the phase has been extended." Note the plusses and items being added.


- `fr`

image

Note the phase is here VETTING and the plus is not available.

### CLDR_PHASE=VETTING_CLOSED but CLDR_DDL_PHASE=SUBMIT

- locale list

image

Shows as 'Vetting Closed', does NOT show `fr` but lists `csw` as clickable.

- `csw`

Same as before.

image


- `fr`

image

User is no longer able to vote. Forum remains open.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -2823,6 +2825,24 @@ private void doStartup() {
                                     + phaseString);
                 }
                 currentPhase = newPhase;
+                String ddlPhaseString = survprops.getProperty("CLDR_DDL_PHASE", null);
+                Phase ddlPhase = null;
+                try {
+                    if (ddlPhaseString != null && !ddlPhaseString.isEmpty()) {
+                        if (currentPhase == Phase.READONLY) {
+                            busted(
+                                    "Error: Cannot have a CLDR_DDL_PHASE when CLDR_PHASE=READONLY. Remove the CLDR_DDL_PHASE.");
```
btangmu, line 2834: maybe allow CLDR_DDL_PHASE=READONLY when CLDR_PHASE=READONLY?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
```diff
@@ -3570,6 +3590,21 @@ public static Phase phase() {
...
 
+    public static Phase phase(CLDRLocale locale) {
+        return (SubmissionLocales.isTcLocale(locale) ? phase() : getDDLPhase());
+    }
+
+    /**
+     * @returns the current phase for the locale. This is the preferred API.
+     */
+    public static CheckCLDR.Phase getCPhase(CLDRLocale loc) {
+        return phase(loc).getCPhase();
```
btangmu, line 3601: "C" for "current" is a little cryptic; I didn't know what it stood for until I got here. How about getCurrentPhase? Or a name that clarifies this method is appropriate for both DDL and non-DDL locales?

It seems like a terminology is needed, distinguishing 3 categories:

1.  methods/data specifically for DDL
2. methods/data specifically for non-DDL
3. methods/data that apply to either DDL or non-DDL, depending on the specified locale

There's also the long-standing confusion between CheckCLDR.Phase and SurveyMain.Phase, which still makes my head spin. Of course, fixing that is out of the scope of this ticket, but as long as new methods are being named, it would SO helpful for the new methods/data to use a naming convention that distinguishes them -- maybe checkPhase vs stPhase?

Also, it seems error prone to have three public methods: phase(), phase(locale), and getCPhase(loc). If the no-argument phase() ignores the existence of DDL, what ensures that various modules are consistent in treating DDL differently from non-DDL?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
srl295: Could, I was trying to simply have that unset when the ST is in readonly. Otherwise DDL would remain READONLY when the rest of it  reopens.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
srl295: it's not current, it's CheckCLDR.  CheckCLDR added its own 'Phase' enum distinct from ST's, and they haven't been fully reconciled yet. It's the "long-standing confusion between CheckCLDR.Phase and SurveyMain.Phase".

srl295: I will work on making these more clearer and consistent.  Perhaps phase() could be renamed to overallPhase() or something.

srl295:
@btangmu does the basic structure seem OK given the API improvement comments?

macchiati:
I'd suggest changing the ST Phase to be SPhase, and all api's accessing it to also be SPhase.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
btangmu: good point

srl295:
> I'd suggest changing the ST Phase to be SPhase, and all api's accessing it to also be SPhase.

Ideally we'd merge the two together. The problem is that SurveyTool distinguishes between READONLY and VETTING_CLOSED, whereas CheckCLDR doesn't.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyMain.java
btangmu: And that would need to be overallPhase(locale), right? Without the locale, it can't be overall. So ideally, phase() without a locale parameter might be private? currently it has about 14 callers (before this PR) (not counting jsp)

btangmu:
I'll approve as a step in the right direction, assuming that DDL users will be better off with it even if it's not fully consistent yet... Have you investigated the callers of phase() without locale? I'm especially concerned that DDL vetters will get mixed results, such as not getting phase=SUBMIT behavior in all circumstances where they're supposed to

tools/cldr-apps/js/src/views/MainHeader.vue
```diff
@@ -1,7 +1,17 @@
...
   <header id="st-header">
+    <a-spin v-if="!loaded" :delay="250" />
     <ul>
-      <li>{{ stVersionPhase }}</li>
+      <li>
+        {{ stVersion }} {{ stPhase }}
+        <span
+          class="ddlException"
+          v-if="ddlException"
+          title="Note: As a non-TC DDL locale, this phase has been extended."
```
btangmu, line 10: ideally this title might include ddlPhase, since "extended" is vague -- extended to what?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
```diff
@@ -1092,6 +1092,7 @@ private static JSONObject createJSONLocMap(SurveyMain sm) throws JSONException {
         // locales will have info about each locale, including name
         JSONObject locales = new JSONObject();
         SupplementalDataInfo sdi = sm.getSupplementalDataInfo();
+        final StandardCodes sc = StandardCodes.make();
```
btangmu, line 1095: is this "sc" unused? I see no reference to it in the changes

btangmu (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyAjax.java
srl295: good catch. replaced by SurveyMain.getLocales()

btangmu:
> I'd suggest changing the ST Phase to be SPhase, and all api's accessing it to also be SPhase.

The fact that I couldn't tell whether CPhase meant "current phase" or "CLDRCheck phase" points to a general problem with single-letter abbreviations. How about STPhase and CheckPhase, pending the merger Steven suggested?

srl295:
> > I'd suggest changing the ST Phase to be SPhase, and all api's accessing it to also be SPhase.
> 
> The fact that I couldn't tell whether CPhase meant "current phase" or "CLDRCheck phase" points to a general problem with single-letter abbreviations. How about STPhase and CheckPhase, pending the merger Steven suggested?

we also have classes such as `Level` which are used in two different scopes. Anyway, I've updated for clarity.

btangmu (dismissed):
it's not quite clear what "overall" means or contrasts with...

srl295:
> it's not quite clear what "overall" means or contrasts with...

I'm updating this per today's meeting. But, it should be reviewable as a delta.

macchiati:
> How about STPhase and CheckPhase, pending the merger Steven suggested?
SGTM

macchiati:
> Ideally we'd merge the two together. The problem is that SurveyTool
distinguishes between READONLY and VETTING_CLOSED, whereas CheckCLDR
doesn't.

I think as far as CLDR is concerned, those would be treated like FINAL_TESTING.
If so, then we could introduce those values in CheckCLDR.Phase, and change
the call sites for CheckCLDR.Phase.FINAL_TESTING. It appears that there are
less than a dozen of those.

        *if* (Phase.*FINAL_TESTING* == getPhase() || Phase.*BUILD* ==
getPhase()) {

            setSkipTest(*false*); // ok

turns into


        *if* (Phase.*FINAL_TESTING* == getPhase() || Phase.*BUILD* ==
getPhase() || Phase.READONLY == getPhase() | Phase.VETTING_CLOSED ==
getPhase()) {



We could make it even cleaner if we add two methods:



   1. isPhase(Set<Phase> testAgainst)
   2. isPhase(Phase... testAgainst)



That way someone could write:


        *if* (isPhase(Phase.*FINAL_TESTING, *Phase.*BUILD, *Phase.READONLY,
Phase.VETTING_CLOSED)) {


and/or we add some static sets for the common cases.


It also appears that in many cases, the desired behavior is really to test
against whether the ST is open, so cleaner in that case to have:



        *if* (!isPhase(Phase.*SUBMISSION, *Phase.*VETTING*)) {

tools/cldr-apps/js/src/views/MainHeader.vue
```diff
@@ -5,9 +5,9 @@
       <li>
         {{ stVersion }} {{ stPhase }}
         <span
-          class="ddlException"
-          v-if="ddlException"
-          title="Note: As a non-TC DDL locale, this phase has been extended."
+          class="extendedException"
+          v-if="extendedException"
+          title="Note: This phase has been extended for this locale."
```
btangmu, line 10: If TC locales are in Vetting mode, and DDL locales are still in Submission mode, then if a DDL locale is selected, will this show "Vetting (extended)" or "Submission (extended)"? I'm guessing the latter but not quite sure

btangmu (dismissed):

tools/cldr-apps/js/src/views/MainHeader.vue
srl295: The latter. And it doesn't have to be a DDL locale, it could be one on the extended list.

btangmu (dismissed):

macchiati:
The ! is ok for now. (We can make it clearer later, not blocker for this.)

btangmu (approved):

srl295:
> The ! is ok for now. (We can make it clearer later, not blocker for this.)

I think you meant to comment on #3796

macchiati:
Yes, sorry for mixing the PR comments.

macchiati (approved):
LGTM

macchiati:
However, I can't speak to the js or vue changes.

srl295:
Will plan to merge after this current push.

srl295:
> However, I can't speak to the js or vue changes.

Thanks.  Tom has reviewd.

PR #3797, opened 2024-06-11
CLDR-17699 DataDog in Ansible

srl295:
CLDR-17699

- we were missing sourcemap uploads for production and for smoketest (though smoketest isn't used with datadog currently)
- add ansible rules to enable datadog

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -52,6 +52,11 @@ jobs:
           -DskipTests=true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: DataDog sourcemap upload
+        run: npx --package=@datadog/datadog-ci datadog-ci sourcemaps upload tools/cldr-apps/src/main/webapp/dist/ --minified-path-prefix=/cldr-apps/dist/ --release-version=r${{ github.event.inputs.git-ref }} --service=surveytool
+        env:
+          DATADOG_SITE: ${{ secrets.DATADOG_SITE }}
+          DATADOG_API_KEY: ${{ secrets.DATADOG_API_KEY }}
       - name: Build liberty server
         run: >
           mvn -s .github/workflows/mvn-settings.xml -B  -pl cldr-apps liberty:create liberty:deploy liberty:package -Dinclude=usr --file tools/pom.xml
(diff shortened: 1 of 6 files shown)


.github/workflows/production.yml
```diff
@@ -52,6 +52,11 @@ jobs:
           -DskipTests=true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: DataDog sourcemap upload
+        run: npx --package=@datadog/datadog-ci datadog-ci sourcemaps upload tools/cldr-apps/src/main/webapp/dist/ --minified-path-prefix=/cldr-apps/dist/ --release-version=r${{ github.event.inputs.git-ref }} --service=surveytool
+        env:
+          DATADOG_SITE: ${{ secrets.DATADOG_SITE }}
+          DATADOG_API_KEY: ${{ secrets.DATADOG_API_KEY }}
```
btangmu, line 59: I don't know yml well enough to know if/how the duplication of code in maven.yml and production.yml could be avoided...

For future reference I found these questions on stackoverflow, but it would take me a long time to figure out if they're helpful in this case:

https://stackoverflow.com/questions/60544181/how-to-use-snippets-in-github-action-workflow-file-to-avoid-duplicates

https://stackoverflow.com/questions/59757355/reuse-portion-of-github-action-across-jobs

tools/scripts/ansible/hosts
```diff
@@ -2,9 +2,15 @@
 cldr-smoke.unicode.org
 cldr-staging.unicode.org
 
+[staging:vars]
+ddenv=staging
```
btangmu, line 6: a comment that "dd" stands for DataDog would be nice; also underscore dd_env would make this consistent with dd_key and dd_site (though "Foolish consistency is the hobgoblin of little minds")

btangmu (dismissed):

.github/workflows/production.yml
srl295: You can reuse an external action, but it's not just an #include. When you use `uses: ./.github/actions/my-action` you can't configure as much about what's inside.

That said we could have a datadog source upload action I suppose.

srl295: I added your links to CLDR-15085

btangmu (approved):

PR #3798, opened 2024-06-11
CLDR-17560 skip more cases of ReportId.valueOf("supplemental")

srl295:
CLDR-17560

ALLOW_MANY_COMMITS=true


btangmu (approved):

macchiati:
A quick question: where are these report ids coming from?

srl295:
> A quick question: where are these report ids coming from?

Users who filled out the report before it was withdrawn.

macchiati:
Ah, makes sense

PR #3800, opened 2024-06-12
CLDR-17719 update v46 charts

macchiati:
CLDR-17719

I found that the coverage_goals.html chart was omitting Meta. I put in a fix so that it would be driven by the Organization.getTCOrgs().

I also generate a .tsv that reformats the Locales.txt data. The reformatted data sorts TC locales first, then others. It does not overwrite Locales.txt, but rather creates the .tsv with the staging charts.

The format looks like:

```
#Organization	;	Locale	;	Level	;	Locale Name

google	;	*	;	moderate	;	ALL
google	;	af	;	modern	;	Afrikaans
google	;	ak	;	moderate	;	Akan
google	;	am	;	modern	;	Amharic
google	;	ar	;	modern	;	Arabic
google	;	as	;	modern	;	Assamese
google	;	az	;	modern	;	Azeri
google	;	be	;	modern	;	Belarusian
google	;	bg	;	modern	;	Bulgarian
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
@@ -8,7 +8,6 @@ package org.unicode.cldr.tool;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.Sets;
@@ -1374,10 +1373,8 @@ public class ShowLanguages {
 
         static final Set<Organization> TC_Vendors =
                 Sets.union(
-                        // this is a slightly odd construction, to add airbnb into the sorted items,
-                        // but then end with cldr
-                        ImmutableSortedSet.copyOf(
-                                Sets.union(Organization.getTCOrgs(), Set.of(Organization.airbnb))),
+                        Organization.getTCOrgs(),
+                        // This adds the CLDR org at the end of the list
                         Set.of(Organization.cldr));
 
         private void showCoverageGoals(PrintWriter pw) throws IOException {
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
@@ -1399,7 +1399,9 @@ public class ShowLanguages {
                                 .addColumn(
                                         "Code",
                                         "class='source'",
-                                        "<a href=\"http://www.unicode.org/cldr/data/common/main/{0}.xml\">{0}</a>",
+                                        "<a href=\""
+                                                + CLDRURLS.CLDR_REPO_ROOT
+                                                + "/blob/main/common/main/{0}.xml\">{0}</a>",
                                         "class='source'",
                                         false)
                                 .addColumn(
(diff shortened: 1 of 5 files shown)


AEApple:
re: Airbnb they are focusing on specific data types. Will it lead to any confusion if we include them?

srl295 (dismissed):

macchiati:
> re: Airbnb they are focusing on specific data types. Will it lead to any
confusion if we include them?

I'll modify to leave them out until we confirm, since it isn't really full
'modern'.

srl295 (dismissed):
OK for now but i have a uRL comment

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowLanguages.java
```diff
@@ -1372,171 +1372,187 @@ private String getLanguageName(String languageCode) {
...
+                        new TablePrinter()
+                                // tablePrinter.setSortPriorities(0,4)
+                                .addColumn(
+                                        "Language", "class='source'", null, "class='source'", false)
+                                .setSortPriority(0)
+                                .setBreakSpans(false)
+                                .addColumn(
+                                        "Code",
+                                        "class='source'",
+                                        "<a href=\"http://www.unicode.org/cldr/data/common/main/{0}.xml\">{0}</a>",
```
srl295, line 1402:
```suggestion
                                        "<a href=\"https://www.unicode.org/cldr/data/common/main/{0}.xml\">{0}</a>",
```

1. http should be https
2. should use a CLDRURLS constant anyway
3. this URL looks wrong and doesn't seem to work. were you wanting something such as <https://github.com/unicode-org/cldr/blob/main/common/main/fr.xml> etc?

macchiati:
Ah, yes, this probably dates way back. Will fix.

macchiati:
After making the fix you recommended, I searched for https://github.com/unicode-org/cldr and got the following hits in our source:

ConsoleCheckCLDR.java
2,010: "<a href='https://github.com/unicode-org/cldr/tree/main/common/main/" 

CLDRModify.java
2,680: // https://github.com/unicode-org/cldr/pull/2659 

GenerateTransformCharts.java
370: + "For the latest snapshot of the data files, see <a href='https://github.com/unicode-org/cldr/tree/main/common/transforms'>Transform XML Data</a>. " 

ShowLanguages.java
1,404: "<a href=\"https://github.com/unicode-org/cldr/blob/main/common/main/{0}.xml\">{0}</a>", 

ShowLocaleCoverage.java
70: "https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/" 
CLDRURLS.java (2 matches)
13: public static final String CLDR_REPO_BASE = "https://github.com/unicode-org/cldr"; 
25: public static final String CLDR_REPO_ROOT = "https://github.com/unicode-org/cldr"; 

TestCheckCLDR.java
639: * https://github.com/unicode-org/cldr-staging/blob/main/births/41.0/fr.txt (for the current 

NOTE

1. CLDR_REPO_BASE is only used in DEFAULT_COMMIT_BASE = CLDR_REPO_BASE + "/commit/";
2. CLDR_REPO_ROOT is only used in Chart.java.GITHUB_ROOT = CLDRURLS.CLDR_REPO_ROOT + "/blob/main/";
3. It feels like it would be better to add a constant: 
    CLDRURLS.CLDR_REPO_MAIN = CLDR_REPO_BASE + "/blob/main";
4. Then fix the above to use one of the two CLDR_REPO_xxxx 

What do you think?

macchiati:
FYI CLDR_REPO_ROOT and CLDR_REPO_BASE have the same definition currently

srl295:
@macchiati 

> `CLDRURLS.CLDR_REPO_MAIN = CLDR_REPO_BASE + "/blob/main"`

this is probably the best way to do it. Maybe include a trailing slash (`/blob/main/`) so you can do  `CLDRURLS.CLDR_REPO_MAIN + "common/main/fr.xml"`

macchiati:
Ok, will do

macchiati:
This works now, after the change in constants that Steven noted. So it's ready to go.

macchiati:
Could someone approve this?

srl295 (approved):

PR #3802, opened 2024-06-12
CLDR-17532 Reset keyword limit to 20

macchiati:
CLDR-17532

ALLOW_MANY_COMMITS=true


macchiati:
I loaded this on staging, and the vetters have done nice work on the overlong emoji annotations, so looks good.

srl295 (approved):

PR #3807, opened 2024-06-16
CLDR-17728 Rename "Palestinian territories" to "Palestine"

Jules-Bertholet:
CLDR-17728

ALLOW_MANY_COMMITS=true


macchiati:
For the locale data, we do not intake data like this via PRs. If the ticket is approved, then the data would be handled by translators, via the survey tool.

Jules-Bertholet:
@macchiati the data is already present for all these languages, just in the wrong place (as an alternative `"short"` name instead of the primary name). Does the survey tool have a mechanism for performing this bulk update? Or would we need to manually cut and paste the existing name for hundreds of languages?

macchiati:
Ah, I looked again at the PR. Since the 'short' name is already supplied,
there isn't anything more to do. That is, the implementations using CLDR
can choose which they want.

Jules-Bertholet:
> That is, the implementations using CLDR can choose which they want.

But, unless they choose to do additional work, they will use the long name by default. Also, the choice of primary name determines the name of the flag emoji (🇵🇸).

PR #3822, opened 2024-06-24
CLDR-17758 DDL: expand VettingViewerSummary for unaffiliated TC to all non-TC locales

srl295:
CLDR-17758

- as an Unaffiliated TC, TC can request a Priority Items Summary that includes ONLY non-TC locales.
- also split out Time Calc to improve the remaining time calculation for priority items

Re-do of #3133 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
@@ -1,0 +1,75 @@
+package org.unicode.cldr.util;
+
+import com.ibm.icu.text.MeasureFormat;
+import com.ibm.icu.text.MeasureFormat.FormatWidth;
+import com.ibm.icu.text.RelativeDateTimeFormatter;
+import com.ibm.icu.util.Measure;
+import com.ibm.icu.util.MeasureUnit;
+import com.ibm.icu.util.ULocale;
+import java.util.Locale;
+
+public class TimeDiff {
+    public static String timeDiff(long a) {
+        return timeDiff(a, System.currentTimeMillis());
+    }
+
+    public static String durationDiff(long a) {
+        return timeDiff(System.currentTimeMillis() - a);
+    }
+
+    private static final long ONE_SECOND = 1 * 1000;
+    private static final long ONE_MINUTE = 60 * 1000;
+    private static final long ONE_HOUR = 3600 * 1000;
+    private static final long ONE_DAY = 86400 * 1000;
+    private static final double ONE_YEAR = ONE_DAY * 364.25;
+
+    /**
+     * @returns string representation of the difference between the two millisecond values
+     */
+    public static String timeDiff(long a, long b) {
+        double del = (b - a);
+        final RelativeDateTimeFormatter fmt = RelativeDateTimeFormatter.getInstance(Locale.ENGLISH);
+
+        if (del > ONE_YEAR) {
+            del /= ONE_YEAR; // approximate
+            int years = (int) del;
+            return fmt.format(
+                    years,
+                    RelativeDateTimeFormatter.Direction.LAST,
+                    RelativeDateTimeFormatter.RelativeUnit.YEARS);
+        } else if (del > (ONE_DAY)) {
+            // more than a day: reference in days
+            del /= ONE_DAY;
+            int days = (int) del;
+            return fmt.format(
+                    days,
+                    RelativeDateTimeFormatter.Direction.LAST,
+                    RelativeDateTimeFormatter.RelativeUnit.DAYS);
+        } else if (del > ONE_HOUR) {
+            final double hours = (b - a) / (ONE_HOUR);
+            return fmt.format(
+                    hours,
+                    RelativeDateTimeFormatter.Direction.LAST,
+                    RelativeDateTimeFormatter.RelativeUnit.HOURS);
+        } else if (del > ONE_MINUTE) {
+            final double minutes = (b - a) / (ONE_MINUTE);
+            return fmt.format(
+                    minutes,
+                    RelativeDateTimeFormatter.Direction.LAST,
+                    RelativeDateTimeFormatter.RelativeUnit.MINUTES);
+
+        } else if (del > ONE_SECOND) {
+
+            final double seconds = (b - a) / (ONE_SECOND);
+            return fmt.format(
+                    seconds,
+                    RelativeDateTimeFormatter.Direction.LAST,
+                    RelativeDateTimeFormatter.RelativeUnit.SECONDS);
+
+        } else {
+            // fall back to milliseconds
+            return MeasureFormat.getInstance(ULocale.ENGLISH, FormatWidth.NARROW)
+                    .format(new Measure(del, MeasureUnit.MILLISECOND));
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final long ONE_SECOND = 1 * 1000;
+    private static final long ONE_MINUTE = 60 * 1000;
+    private static final long ONE_HOUR = 3600 * 1000;
+    private static final long ONE_DAY = 86400 * 1000;
+    private static final double ONE_YEAR = ONE_DAY * 364.25;
+
+    /**
+     * @returns string representation of the difference between the two millisecond values
+     */
+    public static String timeDiff(long a, long b) {
```
btangmu, line 29: I'm surprised there's not a standard library for this

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1084,6 +1084,27 @@ private void writeSummaryTable(
 
     private Map<String, String> getSortedNames(Organization org, Level desiredLevel) {
         Map<String, String> sortedNames = new TreeMap<>(CLDRConfig.getInstance().getCollator());
+        LocalesWithExplicitLevel cldrLocale =
+                new LocalesWithExplicitLevel(Organization.cldr, desiredLevel);
+
+        // TODO: another hack, unaffiliated gets everything
```
btangmu, line 1090: "TODO" should have a ticket number and clarify what needs to be done, otherwise just a comment not a "TODO"

btangmu (approved):
"TODO" comments should be addressed/removed before finishing the ticket

tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
srl295: Icu has the formatting but not the field resolution.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: The todo is not needed here. I'll remove it.

srl295: removed

tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
btangmu: 86400 = Duration.ofDays(1).getSeconds()

I guess there would be a tiny performance penalty

btangmu (approved):

srl295:
@btangmu didn't realize i had already closed that PR

tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
macchiati: This makes it very hard to import data into spreadsheets. You really don't want to force people to write functions to convert various measures of time into real numbers that can be compared.

For now, it should be just a number of days, eg 3.2 days — and without the "days" string; that should be in the column header.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1084,6 +1084,27 @@ private void writeSummaryTable(
 
     private Map<String, String> getSortedNames(Organization org, Level desiredLevel) {
         Map<String, String> sortedNames = new TreeMap<>(CLDRConfig.getInstance().getCollator());
+        LocalesWithExplicitLevel cldrLocale =
+                new LocalesWithExplicitLevel(Organization.cldr, desiredLevel);
```
macchiati, line 1088: This is a **TC user** in the unaffiliated locales, right?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -1084,6 +1084,27 @@ private void writeSummaryTable(
...
+
+        // A user in the Unaffiliated organization can access a list with all non-TC locales.
+        if (org == Organization.unaffiliated && desiredLevel == Level.BASIC) {
+            for (String localeID : cldrFactory.getAvailable()) {
+                final CLDRLocale loc = CLDRLocale.getInstance(localeID);
+                if (SubmissionLocales.isTcLocale(loc)) continue; // skip TC locales
```
macchiati, line 1094: Once this is in place, 
1. the Priority Items should have all and only those locales that are in Organization.cldr - Organization.special.
2. This list should include all and only those locales that are not in the Priority items.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
macchiati: In particular, if a locale has any errors (that are not duplicates of a parent locales errors), they should be listed in one of the two lists.

tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
srl295: This is for ephemeral information such as the vetting viewer page load, not downloaded data. Last seen in the spreadsheet uses a UTC timestamp.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: It's a user already allowed to access priority items summary.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: I didn't modify the non DDL list. Is it already correct?

srl295: The "normal" code is here

https://github.com/unicode-org/cldr/blob/e899b786b8003d9f7b984bf710beeffd940a7682/tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java#L1091

srl295: > In particular, if a locale has any errors (that are not duplicates of a parent locales errors), they should be listed in one of the two lists.

Could we include this under the separate ticket about deduplicating errors? This particular PR has a more limited scope, that simply having a way to get DDL reports at all.

macchiati: Yes, the deduping can be separate.

tools/cldr-code/src/main/java/org/unicode/cldr/util/TimeDiff.java
macchiati: ok, good

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: I don't understand this code. The local variable cldrLocale seems to be assigned a value that is never used?

srl295: @btangmu Thanks! Actually this looks like a copy and paste error from my previous PR, i'll remove it.

PR #3827, opened 2024-06-26
CLDR-17759 Clicking cell should select even if Info Panel is hidden

btangmu:
-Call updateCurrentId and setLastShown directly from listen if Info Panel is hidden

-This is a work-around; the selection should be independent of Info Panel and cldrInfo as stated in new comment

-Remove unused export for openPanel, called only locally

CLDR-17759

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -189,6 +189,14 @@ function listen(str, tr, theObj, fn) {
   cldrDom.listenFor(theObj, "click", function (e) {
     if (panelShouldBeShown()) {
       show(str, tr, theObj /* hideIfLast */, fn);
+    } else if (tr?.sethash) {
+      // These methods, updateCurrentId and setLastShown may be called from show(), if
+      // panelShouldBeShown() returned true. If the Info Panel is hidden then they still
+      // need to be called. Since they don't involve the Info Panel, the implementation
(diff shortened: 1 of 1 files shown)


btangmu:
@srl295 this is still "draft" since I'd like to test a bit more for unwanted side-effects; testing locally it does fix the bug

srl295 (dismissed):
ideally a ticket number in the TODO, but seems plausible

btangmu:
> ideally a ticket number in the TODO, but seems plausible

I intentionally didn't write "TODO" since deeply restructuring this code seems outside the scope of this ticket, and might not happen until we move the whole table to Vue. The comment may be helpful if and when this code gets revised further. "TODO" comments can call too much attention to themselves (often the editor highlights them), to the extent of becoming distracting and counterproductive "wishful thinking", unless there's a specific plan to fix them. My thinking about this has changed a lot over the years; there's a lot of subjectivity and compromise involved...

btangmu:
This isn't a complete solution; for example, clicking a "click to vote" button for a candidate item should also highlight the cell even if the info panel is closed, but it doesn't; fixing that will involve cldrInfo.wireUpButton...

Nevertheless this does fix the main bug. I think negative side-effects are unlikely, and I haven't found any in testing.

btangmu:
BTW note that "selection" has two effects: (1) the URL bar changes to reflect the row clicked on; (2) the cell gets a thick border

srl295 (approved):

PR #3842, opened 2024-07-03
CLDR-17781 CheckChildren should skip null values

srl295:
- if a child's value is null, skip that child. It's clearly not overridden.

CLDR-17781

- in this case, yo_BJ (which is algorithmic anyway) did not have a value at `//ldml/dates/timeZoneNames/metazone[@type="Kazakhstan"]/long/standard`
- I don't think there's a need to skip algorithmic or other special locales here, it would be a minor speed optimization that wouldn't affect many cases.

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

srl295:
Mis ticketed ..

PR #3860, opened 2024-07-10
CLDR-17572 Clarify in ST when text is commentary instead of content

btangmu:
-Make the existing style lrmarker-container more distinctive per ticket description

CLDR-17572

ALLOW_MANY_COMMITS=true


btangmu:
Here's what it looks like:

image

srl295 (approved):

PR #3861, opened 2024-07-10
CLDR-8823 Don't use Jan, Feb, for non-gregorian calendar systems

srl295:
- update PathHeader with a %G, a list of calendars which use the Gregorian month names.

CLDR-8823

image

Includes a test. The test catches some issues.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
@@ -20,7 +20,11 @@
 %L = (long|short|narrow)
 %M = (Alaska_Hawaii|Bering|Dominican|Goose_Bay|Greenland_Central|Dutch_Guiana|Africa_FarWestern|Liberia|British|Irish|Kuybyshev|Sverdlovsk|Baku|Tbilisi|Turkey|Yerevan|Aktyubinsk|Ashkhabad|Dushanbe|Frunze|Kizilorda|Oral|Samarkand|Shevchenko|Tashkent|Uralsk|Urumqi|Dacca|Karachi|Borneo|Malaya|Kwajalein)
 %N = (gregorian|generic|buddhist|chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|japanese|persian|roc)
-%O = (gregorian|chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|persian)
+# calendar systems that should use Gregorian months (just Gregorian)
+# We don't list roc, etc here because their months are hidden.
+%G = (gregorian)
+# all others use M## form months.
+%O = (coptic|dangi|ethiopic|hebrew|indian|islamic|persian)
 %P = (future|past)
 %R = (gregorian|buddhist|coptic|ethiopic|ethiopic-amete-alem|hebrew|indian|islamic|japanese|persian|roc)
 %S = ([^/]*+)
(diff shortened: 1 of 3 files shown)


pedberg-icu:
Changes look good so far but there is a test failure...

AEApple:
Does this mean it will show the month names of the default calendar for that locale?

pedberg-icu:
> Does this mean it will show the month names of the default calendar for that locale?

This is about the key in the left column (not the English name, which will remain as is). Instead of the key showing Jan, Feb, Mar... for non-gregorian calendars, it will show M1, M2, M3....

AEApple:
> > Does this mean it will show the month names of the default calendar for that locale?
> 
> This is about the key in the left column (not the English name, which will remain as is). Instead of the key showing Jan, Feb, Mar... for non-gregorian calendars, it will show M1, M2, M3....

Okay in the ticket the "English" showed "Month 1", "Month 2" etc.

macchiati:
I presume that this does not change Gregorian -based calendars, right? And
by that I include Julian, Thai Buddhist, etc

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -20,7 +20,11 @@
 %L = (long|short|narrow)
 %M = (Alaska_Hawaii|Bering|Dominican|Goose_Bay|Greenland_Central|Dutch_Guiana|Africa_FarWestern|Liberia|British|Irish|Kuybyshev|Sverdlovsk|Baku|Tbilisi|Turkey|Yerevan|Aktyubinsk|Ashkhabad|Dushanbe|Frunze|Kizilorda|Oral|Samarkand|Shevchenko|Tashkent|Uralsk|Urumqi|Dacca|Karachi|Borneo|Malaya|Kwajalein)
 %N = (gregorian|generic|buddhist|chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|japanese|persian|roc)
-%O = (gregorian|chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|persian)
+# calendar systems that should use Gregorian months (just Gregorian)
+# We don't list roc, etc here because their months are hidden.
+%G = (gregorian)
```
macchiati, line 25: This should support not only gregorian, but also gregorian-based calendars. (Those that use the same 12 month solar calendar: Thai Buddhist, Julian, etc.

macchiati (changes requested):

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
macchiati: Here are the calendars. Unfortunately, we don't have data here as to which is gregorian-based.
```
            <type name="buddhist" description="Thai Buddhist calendar"/>
            <type name="chinese" description="Traditional Chinese calendar"/>
            <type name="coptic" description="Coptic calendar"/>
            <type name="dangi" description="Traditional Korean calendar" since="22.1"/>
            <type name="ethioaa" description="Ethiopic calendar, Amete Alem (epoch approx. 5493 B.C.E)" alias="ethiopic-amete-alem"/>
            <type name="ethiopic" description="Ethiopic calendar, Amete Mihret (epoch approx, 8 C.E.)"/>
            <type name="gregory" description="Gregorian calendar" alias="gregorian"/>
            <type name="hebrew" description="Traditional Hebrew calendar"/>
            <type name="indian" description="Indian calendar"/>
            <type name="islamic" description="Hijri calendar"/>
            <type name="islamic-umalqura" description="Hijri calendar, Umm al-Qura" since="24"/>
            <type name="islamic-tbla" description="Hijri calendar, tabular (intercalary years [2,5,7,10,13,16,18,21,24,26,29] - astronomical epoch)" since="24"/>
            <type name="islamic-civil" description="Hijri calendar, tabular (intercalary years [2,5,7,10,13,16,18,21,24,26,29] - civil epoch)" since="24"/>
            <type name="islamic-rgsa" description="Hijri calendar, Saudi Arabia sighting" since="24"/>
            <type name="iso8601" description="ISO calendar (Gregorian calendar using the ISO 8601 calendar week rules)" since="2.0"/>
            <type name="japanese" description="Japanese Imperial calendar"/>
            <type name="persian" description="Persian calendar"/>
            <type name="roc" description="Republic of China calendar"/>
            <type name="islamicc" description="Civil (algorithmic) Arabic calendar" deprecated="true" preferred="islamic-civil"  alias="islamic-civil"/>
```

macchiati: I ran a quick check, and it appears that the following are gregorian-based: 
[gregoriy, iso8601, buddhist, japanese, roc]

Also filed a ticket to add metadata to cldr, so that this can be unhardcoded

srl295: So actually it seems the "intent" was to hide those so that you don't have a Japanese month January with a different translation than Gregorian January. That mystified me too but I think that's why %G is just one and not the others

macchiati: @pedberg-icu should also weigh in on this.

- It clearly should not be limited to exactly 'gregory', since iso8601 uses the same months, etc.
- In the case of thai buddhist, japanese, roc, M2, M2 etc seem ok.

srl295: If it's here, then it's visible for editing.  Do we want users giving a different translation for January in 8601 than for gregorian?

srl295:
OK the test failure was because I dropped Chinese inadvertently. I was just trying to subdivide %O into %G + %O, with different text.

The err message was accurate but didn't quite give me enough context. I updated it.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
srl295: @macchiati Specifically, this PR does not make any changes to the visibility (HIDE) or textual content for these calendars: `iso8601, buddhist, japanese, roc`. 

All that it does is change the following calendars to no longer have month codes based on Gregorian:

`chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|persian`

srl295:
Maybe we should discuss in a future meeting.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -20,7 +20,11 @@
...
+# all others use M## form months.
+%O = (chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|persian)
```
macchiati, line 27: This isn't working, I think, because what we want is that every calendar falls into one of these groups.

- Certain calendars are hidden.
- Certain calendars are non-gregorian-based, and use M1
- Certain calendars are gregorian-based, and use Jan

Because we NEVER want to miss a calendar, exactly one of these groups should be "any", and that needs to come last. That should not be the Hidden one, because we want to be safe, and better to show a calendar that we shouldn't, than hide a calendar that should be shown.

So it should be:

So
# all the calendars that should be hidden

...%XXX...

# all the calendars that should use M1

...%O...

# all the calendars that should use Jan

...%A...

macchiati (changes requested):

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
srl295: Can you detail what "isn't working"?

srl295:
@macchiati  a concrete question.

Currently, (with OR without this PR), Survey Tool does not allow the user to translate "January" differently in the Japanese calendar than in the Gregorian.  

Are you saying that users should be able to change the translation of "January" for the Japanese calendar?

macchiati:
Peter would know better, but I believe the same months are used in the
Gregorian Japanese calendar and the Imperial Japanese calendar. The only
difference is the era and year number.

macchiati (approved):

PR #3863, opened 2024-07-12
CLDR-17402 json: drop _cldrVersion path

srl295:
- drop //ldml/identity/version entirely (was producing _cldrVersion in every file)
- also, annotate the code working around CLDR-17790 as a TODO
- As a reminder, this will greatly reduce the diffs every time json is updated.

CLDR-17402

Example output with this change:

```diff
--- a/cldr-json/cldr-units-modern/main/zh/measurementSystemNames.json
+++ b/cldr-json/cldr-units-modern/main/zh/measurementSystemNames.json
@@ -2,9 +2,6 @@
   "main": {
     "zh": {
       "identity": {
-        "version": {
-          "_cldrVersion": "46"
-        },
         "language": "zh"
       },
       "localeDisplayNames": {
```

ALLOW_MANY_COMMITS=true


btangmu (approved):
There's still a lot of code involving "//ldml/identity/version" and this PR even adds a little bit more such code. Am I right in guessing that such paths are still required in some circumstances, such as in XML, even though they're no longer required in JSON?

srl295:
> There's still a lot of code involving "//ldml/identity/version" and this PR even adds a little bit more such code. Am I right in guessing that such paths are still required in some circumstances, such as in XML, even though they're no longer required in JSON?

- The data is part of the DTD, so needs to be present for backwards compatibility in XML.
- It's used in JSON, but it's extracted into the package manifest by other code.
- The version is read in XML by other Java code.

PR #3866, opened 2024-07-12
CLDR-17492 deps: update js deps

srl295:
CLDR-17492

ALLOW_MANY_COMMITS=true


srl295:
There's a whole problem with how we do the FE tests that prevents this. I will ahve to come back to this.

btangmu (approved):

PR #3867, opened 2024-07-13
CLDR-17792 Don't strip ZWSP for value compare, allow currency symbol collision for it

pedberg-icu:
CLDR-17792

Currently Survey Tools is reporting a display collision for \u200B (ZWSP) used as currency symbol for PTE in pt_PT, even though this is the value that has been there for several releases, and is the correct value to use since for this particular currency in pt_PT the symbol is actually in the decimal separator element; see https://st.unicode.org/cldr-apps/v#/pt_PT/C_SEEurope/7f78787c0e4c5be3 (best to view this with coverage set to Comprehensive to see the associated fields).

The error does not appear in offline unit tests or Console Check, so the fix here is somewhat speculative:
- Update SimpleXMLSource.NON_ALPHANUM so \u200B does not get stripped when comparing values across paths
- Update CheckDisplayCollisions to specifically allow a collision for  \u200B as a currency symbol if there is an explicit decimal separator for the same currency.
- Add a test for the above, and also a test to verify that DAIP is not altering the \u200B as a currency symbol.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
@@ -491,6 +491,15 @@ public class CheckDisplayCollisions extends FactoryCheckCLDR {
             if (path.contains("/decimal") || path.contains("/group")) {
                 return this;
             }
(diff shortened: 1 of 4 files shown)


macchiati (approved):

srl295 (commented):
post merge LGTM

PR #3871, opened 2024-07-16
CLDR-14788 Hide TC and manager level menu items from vetters

btangmu:
-Front end: hide buttons for Show all users, unless new function canListMultipleUsers returns true, that is, userIsManager

-Front end: in main menu hide My Organization section header if the section would be empty

-Front end: canListUsers and canMonitorVetting only if userIsManager

-Front end: canSeeUnofficialEmail only if isUnofficial and userIsManager

-Back end: canDoList only if isManagerOrStronger

-Back end: SurveyVettingParticipation only if isManagerOrStronger

-Back end: fix TestUserLevel.csv,  userCanDoList false for vetter

CLDR-14788

ALLOW_MANY_COMMITS=true


tools/cldr-apps/js/src/esm/cldrAccount.mjs
```diff
@@ -354,6 +356,10 @@ function getTableEnd(json) {
   return html;
 }
 
+function canListMultipleUsers() {
+  return !!cldrStatus.getPermissions()?.userIsManager;
```
srl295, line 360: another way would be to expand the `cldrStatus.getPermissions()` struct ( a string -> boolean map i think) to include a new item from the server side.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -381,7 +381,7 @@ public boolean canImportOldVotes(CheckCLDR.Phase inPhase) {
         }
 
         public boolean canDoList() {
-            return isVetter();
+            return isManagerOrStronger();
```
srl295, line 384: unit tests are failing here.

srl295 (dismissed):
needs test fix but it's basically ok.

The "My Account" panel under <https://cldr-smoke.unicode.org/cldr-apps/v#account///> may be hidden as a result of this. (try it when logged in on smoketest)

  A good followon might be to have a simple popover when you mouse over the  user's name in the header that shows the user's name, organization, status, email address etc.  Currently the user's name has a tooltip showing their email address.

tools/cldr-apps/js/src/esm/cldrAccount.mjs
btangmu: Good idea. I'll follow up with a PR to get canDoList from the back end...

tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
btangmu: Right. Fixed by the 2nd commit

srl295 (approved):
good that we have a test

btangmu:
> may be hidden as a result

Testing locally, "My Account" is still visible for a vetter, it just lacks the "Show all users" button. I'll double check on smoketest

btangmu:
> I'll double check on smoketest

I did, and it's OK: "My Account" is still visible for a vetter

PR #3872, opened 2024-07-16
CLDR-16844 Fix bug in last PR: require parameter for cldrInfo.closePanel

btangmu:
-Add the boolean parameter for both callers

-Log error to console if parameter is undefined

CLDR-16844

ALLOW_MANY_COMMITS=true


btangmu:
Obviously I didn't test the previous PR thoroughly! I tested it for the desired new behavior, that the panel reappears when it had been closed automatically, but I didn't test for the old behavior that it stays closed if it had been closed manually

macchiati (approved):

PR #3873, opened 2024-07-17
CLDR-17803 site: fix frontmatter errs, add docker build

srl295:
CLDR-17803

- fix some yaml errs in the frontmatter
- add a Docker build for local testing (with docs)

ALLOW_MANY_COMMITS=true


srl295:
The docker build lets you test the github-pages build locally.

srl295:
NB  to expand on this—  this may change drastically when we actually dfeploy to cloudflare… the purpose of the docker is to try to reproduce exactly what is being done when we deploy to github pages today, so that we can get a precise comparison.

btangmu (approved):

PR #3879, opened 2024-07-17
CLDR-17812 Generate VXML

btangmu:
- The target branch is unicode-org:brs/v46vxml. There will be a second PR for merging brs/v46vxml into main

- This PR has two commits. The second is for deleting common/annotations/ff.xml and common/annotations/nb.xml

- If there were not two commits, git would think ff.xml was renamed to ca_ES.xml and nb.xml was renamed to vec.xml

CLDR-17812

ALLOW_MANY_COMMITS=true


macchiati (approved):
Merging into branch so we can work from there.

PR #3856, opened 2024-07-08
CLDR-17661 ii: add yiii numbering system

srl295:
- this is just a draft
- this is NOT how Yiii numbers work, but just to test the mechanism.

CLDR-17661

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3881, opened 2024-07-18
CLDR-17812 ii: drop yiii numbering system

srl295:
CLDR-17812

#3856 is work in progress on the numbering system.

ALLOW_MANY_COMMITS=true


DavidLRowe (approved):
LGTM

PR #3887, opened 2024-07-18
CLDR-17812 v46 vxml: Add {0} for de speed-light-speed one, fix error wording in TestCheckCLDR/TestPlaceholders

pedberg-icu:
CLDR-17812

merge into [#3880](https://github.com/unicode-org/cldr/pull/3880)

`de` added unit data for speed-light-speed in which the "one" case has no placeholder; per discussion that reflects a misunderstanding of how this data is intended to be used, so add a placeholder so tests pass (the data will be deleted soon anyway). Also fix the wording of an error message in TestCheckCLDR/TestPlaceholders.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/de.xml
+++ b/common/main/de.xml
@@ -10677,10 +10677,10 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<unit type="speed-light-speed">
 				<gender>feminine</gender>
 				<displayName>Lichtgeschwindigkeit</displayName>
-				<unitPattern count="one">Lichtgeschwindigkeit</unitPattern>
-				<unitPattern count="one" case="accusative">Lichtgeschwindigkeit</unitPattern>
-				<unitPattern count="one" case="dative">Lichtgeschwindigkeit</unitPattern>
-				<unitPattern count="one" case="genitive">Lichtgeschwindigkeit</unitPattern>
+				<unitPattern count="one">{0} Lichtgeschwindigkeit</unitPattern>
+				<unitPattern count="one" case="accusative">{0} Lichtgeschwindigkeit</unitPattern>
+				<unitPattern count="one" case="dative">{0} Lichtgeschwindigkeit</unitPattern>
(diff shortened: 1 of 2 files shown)


macchiati (changes requested):
Thanks for taking this. I think the solution is a bit different.

The LOCALE_DEPENDENT path means that there is an extra test on the locale's plural rules to make sure that it behaves like Arabic (and other languages); that is, where a particular plural form is unique and therefore does not need a placeholder. For example, "1 book" in Arabic is just "كتاب".

So for this case, the error message is misleading, and the German should be fixed to prefix the values with 
"{0} "

pedberg-icu:
> Thanks for taking this. I think the solution is a bit different.
> ...
> So for this case, the error message is misleading, and the German should be fixed to prefix the values with "{0} "

@macchiati Hmm. What German currently has is
- one: Lichtgeschwindigkeit = Speed of light
- other: {0}-fache Lichtgeschwindigkeit = {0} times the speed of light

That seems very sensible to me. Saying "1 times the speed of light" is certainly awkward in English and I would think in German too. The particle is traveling at lightspeeed, or the particle is traveling at 0.3 times the speed of light.

macchiati:
I think there are significant issues with the data for a number of locales (will outline in email), so we probably want to retract the data for light-speed across locales. However, I suggest that we have a quick fix to the data and merge it into main before retracting so that we have a copy we can return to as needed.

So the quick fix I'm proposing is to add {0} before Lichtgeschwindigkeit

pedberg-icu:
> So the quick fix I'm proposing is to add {0} before Lichtgeschwindigkeit

Literally just "{0}" before, or "{0}-fache" before as with the `other` case?

macchiati:
Just "{0} " before is fine (just sent you a note about the complexities).

pedberg-icu:
@macchiati OK I think I fixed as requested, PTAL. The error message wording was still backwards, needed fixing (see the previous error message by viewing more context above).

macchiati (approved):

srl295 (approved):

srl295 (approved):

PR #3888, opened 2024-07-19
CLDR-17812 v46 vxml: Update TestExampleGenerator/TestUnicodeSetExamples to match hi change in #3883

pedberg-icu:
CLDR-17812

merge into [#3880](https://github.com/unicode-org/cldr/pull/3880)

Updated the `hi` test in TestExampleGenerator/TestUnicodeSetExamples to match the `hi`i exemplar set change in [#3883](https://github.com/unicode-org/cldr/pull/3883)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
@@ -1638,7 +1638,7 @@ public class TestExampleGenerator extends TestFmwk {
                 "hi",
                 "//ldml/characters/exemplarCharacters[@type=\"auxiliary\"]",
                 "[ॄ‌‍]",
-                "〖‎🗝️ ॑ ॒ ॠ ॡ ॻ ॼ ॾ ॿ ऱ ॢ ॣ〗〖❰ZWNJ❱ ≡ cursive non-joiner〗〖❰ZWJ❱ ≡ cursive joiner〗〖❬internal: ❭[ॄ‌‍]〗"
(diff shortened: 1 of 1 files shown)


macchiati (approved):

srl295 (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -1638,7 +1638,7 @@ public void TestUnicodeSetExamples() {
                 "hi",
                 "//ldml/characters/exemplarCharacters[@type=\"auxiliary\"]",
                 "[ॄ‌‍]",
-                "〖‎🗝️ ॑ ॒ ॠ ॡ ॻ ॼ ॾ ॿ ऱ ॢ ॣ〗〖❰ZWNJ❱ ≡ cursive non-joiner〗〖❰ZWJ❱ ≡ cursive joiner〗〖❬internal: ❭[ॄ‌‍]〗"
+                "〖‎🗝️ ॑ ॒ ॠ ॡ ॻ ॼ ॾ ॿ ॢ ॣ〗〖❰ZWNJ❱ ≡ cursive non-joiner〗〖❰ZWJ❱ ≡ cursive joiner〗〖❬internal: ❭[ॄ‌‍]〗"
```
srl295, line 1641: this would be a good example (heh) of where the test would be better off to use canned input data - a static exemplar set content rather than using live input.

`TestCLDRFile.testSourceLocale()` is an example of where there's a whole separate cldr tree with stabilized data. Maybe we could come up with something similar for examples.

srl295 (approved):

PR #3890, opened 2024-07-19
CLDR-17812 move to -z FINAL_TESTING

srl295:
CLDR-17812

ALLOW_MANY_COMMITS=true


macchiati:
It should have BUILD from start of cycle thru submission, then FINAL_TESTING in vetting to end of cycle. Is there a way to drive it off the phase?

macchiati (approved):

srl295:
> It should have BUILD from start of cycle thru submission, then FINAL_TESTING in vetting to end of cycle. Is there a way to drive it off the phase?

let's discuss in email or on https://unicode-org.atlassian.net/browse/CLDR-9149 or elsewhere

PR #3895, opened 2024-07-23
CLDR-15698 Time examples < 10 and > 12

e-ikwut:
[output.txt](https://github.com/user-attachments/files/16352544/output.txt)

before:
image

after:
image

CLDR-15698

- [CLDR-15698 ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -48,7 +48,6 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
 import org.unicode.cldr.tool.LikelySubtags;
 import org.unicode.cldr.util.AnnotationUtil;
 import org.unicode.cldr.util.CLDRConfig;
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2585,14 +2589,14 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
...
-                examples.add(dtf.format(DATE_SAMPLE));
+                examples.add(dtf.format(DATE_SAMPLE));  // 15698: "short time" -- 2 examples?
             }
 
             return formatExampleList(examples.toArray(new String[0]));
         } else {
             String id = parts.findAttributeValue("dateFormatItem", "id");
             if ("NEW".equals(id) || value == null) {
-                return startItalicSymbol + "n/a" + endItalicSymbol;
+                return startItalicSymbol + "n/a" + endItalicSymbol;  // 15698: doesn't seem relevant--meaning??
```
btangmu, line 2599: It's a mystery to me as well...

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2556,7 +2560,7 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
                                                         setBackground("'" + timeRange + "'"),
                                                         setBackground("'" + dfResult + "'")
                                                     }));
-                    examples.add(dtf.format(DATE_SAMPLE));
+                    examples.add(dtf.format(DATE_SAMPLE));  // 15698: don't need two times for range
```
macchiati, line 2563: You appear to be leaving comments in the code where you think that something is superfluous. Better to use github comments for that, rather than cluttering up the code.

macchiati:
The output.txt is useful, but best to also have before and after screen shots also. And for the text output, also best to have  the before and after.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: The id being "NEW" seems odd to me as well; I just checked the DTD, and it isn't permitted:
```
<!ATTLIST dateFormatItem id CDATA #REQUIRED >
<!--@MATCH:regex/((E|EEEE)?(H|HH|h|hh|Bh)(m|ms|mm|mmss)?(Z|z|zzzz|v|vvvv)?)|(ms|mmss)|((G|GGGGG)?(y|yy|yyyy)((M{1,4}((E|EEEE|cccc)?(d|dd))?)|(w|Q|QQQ|QQQQ))?)|(U(M|MMM)d?)|(M{1,4}(((E|EEEE|cccc)?(d|dd))|W)?)|((E|EEEE)?d)|(E|EEEE)-->
```
So that looks obsolete, and we can probably remove just that check.

I wouldn't remove the value==null check, however, without verifying that it can't occur at this point in the code.

macchiati (changes requested):

macchiati:
There is a commit message that fails validation. The message must be of the form:

CLDR-15698 <some message>

eg

CLDR-15698 Time examples < 10 and > 12

The "Details" button on the "jira-ticket — Commit message for 41fe3cb fails validation" gives you a way to fix that, by squashing. Or you can make a new PR. Let us know on slack if you have trouble with that.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
e-ikwut: Sorry, those were notes to self while I worked (what I meant by that comment was that I don't think I need to add an extra example in that case). Understood.

macchiati:
The screenshots look good.

AEApple:
Will the same examples show in the Info Panel?

e-ikwut:
> Will the same examples show in the Info Panel?

Yes, the same set shows in the pop-up and the info panel.

macchiati (approved):

macchiati:
Once you are ready for review, change the status to "Ready for Review". I approved, but you'll have to fix the conflicts and ask for a re-review.

macchiati (approved):

srl295:
@e-ikwut needs `mvn spotless:apply` can you do that? make sure the commit is formatted properly

srl295 (approved):

sffc (approved):

PR #3897, opened 2024-07-23
CLDR-17830 Update coverageLevels.txt

macchiati:
CLDR-17830

ALLOW_MANY_COMMITS=true


AEApple:
You're not going to wait to see if DDL is able to improve any of the DDL coverage levels which dropped?

macchiati:
I will regenerate the data afterwards with each chart generation, so it
will be updated as needed. This gives us a baseline.

pedberg-icu (approved):

PR #3899, opened 2024-07-24
CLDR-17761 Update en.xml -- add variant Mvskoke to Muscogee (mus)

conradarcturus:
This starts the ball rolling in adding recognition for the name Mvskoke -- an endonym for Muscogee. Initially we just want to add it as a variant. In the future we may want to swap the official & variant options if the Mvskoke/Muscogee linguistic authority prefers it.

[CLDR-17761](https://unicode-org.atlassian.net/browse/CLDR-17761)

- [ ] ~~This PR completes the ticket.~~
- [x] This PR addresses short-term issues with the ticket, pending long-term followup.

ALLOW_MANY_COMMITS=true

[CLDR-17761]: https://unicode-org.atlassian.net/browse/CLDR-17761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -426,6 +426,7 @@ annotations.
 			<language type="mua">Mundang</language>
 			<language type="mul">Multiple languages</language>
 			<language type="mus">Muscogee</language>
+			<language type="mus" alt="variant">Mvskoke</language>
 			<language type="mwl">Mirandese</language>
 			<language type="mwr">Marwari</language>
 			<language type="mwv">Mentawai</language>
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -426,7 +426,8 @@ annotations.
 			<language type="mua">Mundang</language>
 			<language type="mul">Multiple languages</language>
 			<language type="mus">Muscogee</language>
-			<language type="mus" alt="variant">Mvskoke</language>
+			<language type="mus" alt="variant">Muscogee</language>
+			<language type="mus" alt="official">Mvskoke</language>
 			<language type="mwl">Mirandese</language>
 			<language type="mwr">Marwari</language>
 			<language type="mwv">Mentawai</language>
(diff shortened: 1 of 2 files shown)


common/main/en.xml
```diff
@@ -426,6 +426,7 @@ annotations.
 			<language type="mua">Mundang</language>
 			<language type="mul">Multiple languages</language>
 			<language type="mus">Muscogee</language>
+			<language type="mus" alt="variant">Mvskoke</language>
```
macchiati, line 429:
```suggestion
			<language type="mus" alt="variant">Muscogee</language>
			<language type="mus" alt="official">Mvskoke</language>
```

macchiati: See the ticket comment from the TC. It needs to have 2 alt variants.
https://unicode-org.atlassian.net/browse/CLDR-17761?focusedCommentId=176091

macchiati (changes requested):

common/main/en.xml
conradarcturus: That approach was discussed in the TC but there is currently no other language using the tag `alt="official"` -- only "menu" and "variant". Is there documentation mentioning "official" as a supported value for the tag `alt`?

conradarcturus: Or said differently:  I just didn't see any previous language referring to alt="official" although I can definitely add it if its a supported value.

conradarcturus: If we make that change,  I can update the official documentation for the field https://cldr.unicode.org/translation/displaynames/languagelocale-names

macchiati:
We approved another ticket for adding official.

macchiati (approved):

macchiati (approved):

PR #3900, opened 2024-07-24
CLDR-17832 BRSv46 CLDRModify passes before alpha0

pedberg-icu:
CLDR-17832

CLDRModify passes before alpha0:
1. With no arguments: In personName data, add xml:space="preserve" where necessary.
2. With -fQ: In annotations, copy tts name into keywords, reorder keywords as necessary.
3. With -fP: In main zh_Latn, reformat exemplars.
4. Pass 2 above (-fQ) made some annotation keyword lists too long, so revert those 20 cases (in cs, eu, no, ro)

ALLOW_MANY_COMMITS=true


pedberg-icu:
To merge this: Go into settings, check "Allow rebase merging", then in the PR use the popup next to the "Squash and merge" button to instead select "Rebase merge" and merge with that, then go back to Settings and uncheck "Allow rebase merging".

macchiati (changes requested):
I'm sorry, I should have raised this earlier. I don't think we should add the tts names to the keywords, unless the keywords are empty, because it works against the guidance that people should break up the search keywords by word where possible.

I suggest that we hold off adding the tts name, and just do the daip (which reorders).

pedberg-icu:
> ... I don't think we should add the tts names to the keywords... I suggest that we hold off adding the tts name, and just do the daip (which reorders).

OK, at this point then I think the easiest thing is just to abandon this PR and start over, skipping the -fQ pass.

macchiati:
In that case shall I go ahead and merge the pound currency sign PR?

pedberg-icu:
This is replaced with PR [#3903](https://github.com/unicode-org/cldr/pull/3903)

PR #3903, opened 2024-07-25
CLDR-17832 BRSv46 CLDRModify passes before alpha0

pedberg-icu:
CLDR-17832

The replaces the now-closed earlier PR [#3900](https://github.com/unicode-org/cldr/pull/3900).

CLDRModify passes before alpha0:

1. With no arguments: In personName data, add xml:space="preserve" where necessary.
2. With -fP: In main, reformat parseLenient sets, zh_Latn exemplars; in annotations, reorder en annotations.

Note that the -fQ pass was explicitly skipped per Mark comment on earlier PR: " I don't think we should add the tts names to the keywords, unless the keywords are empty, because it works against the guidance that people should break up the search keywords by word where possible...I suggest that we hold off adding the tts name, and just do the daip (which reorders)".

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/annotations/en.xml
+++ b/common/annotations/en.xml
@@ -720,17 +720,17 @@ annotations.
 		<annotation cp="♯" type="tts">sharp</annotation>
 		<annotation cp="🪉">cupid | harp | instrument | love | music | orchestra</annotation>
 		<annotation cp="🪉" type="tts">harp</annotation>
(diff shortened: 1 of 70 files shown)


pedberg-icu:
To merge this when approved: Go into settings, check "Allow rebase merging", then in the PR use the popup next to the "Squash and merge" button to instead select "Rebase merge" and merge with that (to keep commits distinct on main), then go back to Settings and uncheck "Allow rebase merging".

macchiati (approved):

PR #3157, opened 2023-08-06
CLDR-15878 fixes to Esperanto

verdulo:
CLDR-15878

- [X] This PR completes the ticket.

I corrected numerous wrongly confirmed items, which couldn't be corrected using the Survey Tool.
- word “mondo” minuscule
- names of countries according to the list of [Bertilo](https://bertilow.com/lanlin/index.html) (and [AdE](https://www.akademio-de-esperanto.org/decidoj/landnomoj/listo_de_rekomendataj_landnomoj/))
- names of languages (mostly) according to the list of [Bertilo](https://bertilow.com/lanlin/lingvoj.html)
- [capitalized months](https://bertilow.com/pmeg/gramatiko/propraj_nomoj/majuskloj.html#i-klc) and names of languages according to recommends in [PMEG](https://bertilow.com/pmeg/gramatiko/propraj_nomoj/landoj_popoloj_lingvoj.html#i-bzv)
- nonreligious terms (_de nia erao/antaŭ nia erao_) for eras and religious terms as variants (_post Kristo/antaŭ Kristo_)
- fixed full date pattern (added _la_)
- fixed uncommon full time pattern

Changes are marked as "unconfirmed", so any contributor could change them during the next submitting phase in ST.

ALLOW_MANY_COMMITS=true

PR #3907, opened 2024-07-25
CLDR-17744 fixes to Esperanto #2

verdulo:
CLDR-17744

This PR corrects data, which cannot be fixed using Survey Tool because I am the only contributor with only one vote:
1. Names of countries and language standardized with 10OA by «Akademio de Esperanto».
2. Timezone Display Patterns are now more coherent.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/eo.xml
+++ b/common/main/eo.xml
@@ -395,7 +395,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="tlh" draft="contributed">klingona</language>
 			<language type="tli" draft="unconfirmed">tlingita</language>
 			<language type="tn">cvana</language>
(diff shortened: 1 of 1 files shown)


srl295:
#3157  was a prior PR

verdulo:
> #3157 was a prior PR
The mentioned PR fixed another errors.

The problem is, that with only guest account it is not possible to correct wrong data. My contributions [don't count into coverage levels](https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/46/tsv/locale-coverage.tsv) higher than basic.

The possible solution would be probable find 4 other people and reach consensus using the Survey Tool.

verdulo:
Fixed using Survey Tool.

PR #3908, opened 2024-07-25
CLDR-17675 Revert the decimal/grouping change, but …

macchiati:
Revert the decimal/grouping change, but add alts to allow implementations to choose between official and us variant.

From the discussion in the Design meeting, we agreed on 2 variants, plus the default, as the normal pattern. That allows implementations to pick among the two choices by name — without having to know which was the default — or pick the default. People are also in favor of using semantic alt names, rather than just 'variant'.

Still draft because I want to pass this by the TC; there were a couple of things I wanted to get feedback on.

1. I think the decision was to go with the official syntax (12,34) for the default, but wanted feedback.
2. For the US style, I used alt="us", as probably the most recognizable term. But other suggestions are welcome. (Perhaps alt="period"?)
3. The ticket also mentioned R5,87 as the standard currency format, eg no space between currency alphabetics and the digits. That is in place also.

CLDR-17675

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -2042,7 +2042,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT decimal ( #PCDATA ) >
 <!ATTLIST decimal alt NMTOKENS #IMPLIED >
-    <!--@MATCH:literal/variant-->
+    <!--@MATCH:literal/variant, us, official-->
 <!ATTLIST decimal draft (approved | contributed | provisional | unconfirmed | true | false) #IMPLIED >
     <!--@METADATA-->
     <!--@DEPRECATED:true, false-->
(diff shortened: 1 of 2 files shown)


macchiati:
BTW, I'm getting some bizarre errors; will have to look into those.
```
TestDateFormatsWithContext
Error:  (TestExampleGenerator.java:830)  Error: : Currency format example faulty: expected "【٥‏/٩‏/١٩٩٩〗【⃪٥‏/٩‏/١٩٩٩〗", got "【5‏/9‏/1999〗【⃪5‏/9‏/1999〗"
    TestScriptOfSample
Error:  (TestScriptMetadata.java:57)  Error: : Zinh: The sample character must have a single, valid script, no ScriptExtensions: -9: expected true
...
```

macchiati:
Bitten again by github being sensitive to special characters in a branch name. Sigh. Will need to close this PR and do another.

PR #3910, opened 2024-07-26
CLDR-17839 Fix units in he

AEApple:
CLDR-17839

Update he narrow units to inherit from short

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/he.xml
+++ b/common/main/he.xml
@@ -9015,7 +9015,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<unit type="concentr-karat">
 				<gender>masculine</gender>
 				<displayName>קראט</displayName>
-				<unitPattern count="one">קראט אחד</unitPattern>
+				<unitPattern count="one">{0} קראט</unitPattern>
 				<unitPattern count="two">{0} קראט</unitPattern>
 				<unitPattern count="many">{0} קראט</unitPattern>
 				<unitPattern count="other">{0} קראט</unitPattern>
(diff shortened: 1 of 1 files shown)


AEApple:
@macchiati - Does short inherit from narrow instead of the other way around? I only changed 'narrow' to inherit, so I don't understand why we would require a value for two under narrow but not short?

AEApple:
nm, maybe it's just due to a typo I had.

macchiati:
You can get rid of some of the errors by rebasing onto main. I can help with the others on Monday.

macchiati:
When you get back, can you rebase this on main? We can then take a look at why the other errors are happening.

AEApple:
interestingly it only gave me the option to refresh with a merge commit and not rebase

AEApple:
@macchiati - I think you have to fix the plurals issue first and then I should make my change on top of it. Right now I'm seeing errors due to items such as: https://st.unicode.org/cldr-apps/v#/he/Graphics/665721dc2c8221b3

macchiati:
I'm not sure what is going on here. (Sorry I missed your message of 4 days ago.) **I'll have to investigate tomorrow.** 

**Check failure on line 12161 in common/main/he.xml**
inheritance marker not allowed
 Error: Inheritance marker ↑↑↑ not allowed in a data value.

Very odd. The check appears to be malfunctioning, because ↑↑↑ is allowed everywhere inheritance is. 

**Check failure on line 12487 in common/main/he.xml**
missing placeholders
 Error: Need at least 1 placeholder(s), but only have 0. Placeholders are: {{0}={NUMBER_OF_UNITS}, e.g. “3”}; see <a href='http://cldr.unicode.org/translation/error-codes#missingPlaceholders'  target='cldr_error_codes'>missing placeholders</a>.

It is very strange because in many other cases "<unitPattern count="other">‎↑↑↑</unitPattern>" works fine.

AEApple:
Okay, well I went through and updated everything which seemed like it matched. There were a few things I was less sure about (e.g. durations). I sometimes had a more abbreviated value inherit where it seemed safer.

macchiati (dismissed):

macchiati:
Whew, a lot of work there! Approved

srl295 (approved):

AEApple:
I'm having people review in staging so that I can make any change if needed. I will merge before Dragan starts the alpha1 integration though.

AEApple (commented):
CLDR-17839 Fix two in durations

AEApple (commented):
CLDR-17839 Fix two in durations

macchiati (approved):

PR #3912, opened 2024-07-28
CLDR-17844 Fix breakages

macchiati:
CLDR-17844

Still draft, just adding reviewers for early feedback.

These surfaced locally for me once I updated Maven, so that appears to be the source. 

1. ScriptMetadata is from the change in the UCD to make some accents be explicit scripts (in script extensions)
2. Arabic is from the change in numberSystem for ar.xml. See notes in ticket
3. TestTransforms.java problems are probably more script changes. They also appear to be from the script extensions changes; that the tests are checking that all characters from a given script are converted, and suddenly there's this new character in the script. Suppressing those for now:

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[\u0301\u0300\u0306\u0302\u030C\u030A\u0308\u0303\u0307\u0304\u0309\u0310\u0323-\u0325\u0330\u0331\u00B7\u02BC]&g=ccc&i=

ALLOW_MANY_COMMITS=true


macchiati:
This is to unblock pending PRs. More work on the ticket to follow.

btangmu (approved):

PR #3913, opened 2024-07-29
CLDR-17544 Adjust tool to suppress Mn Ms differences

macchiati:
CLDR-17544

Adjust tool to suppress Mn Ms differences, if those are the only differences (affects basically Slavics)

ALLOW_MANY_COMMITS=true


macchiati:
Pinging for review

btangmu (approved):

PR #3916, opened 2024-07-30
CLDR-17544 Fix just the modern paths for gregorian calendar

macchiati:
Drawing on similarity with closely related patterns. Only modifying ↑↑↑ values. 

Note: There are lots of inconsistencies in formatting and ordering, so I limited the changes to what appears to be clear-cut cases. I think the best we can do is add warnings in an upcoming submission cycle, so I'll clone a ticket for that.

Small changes in search tooling.

CLDR-17544

ALLOW_MANY_COMMITS=true


macchiati:
Ping, I need a review.

btangmu (approved):

PR #3917, opened 2024-07-30
CLDR-7401 Add examples for quarters and eras

haytenf:
it_quarters_BEFORE
it_quarters_AFTER
ja_eras_BEFORE
ja_eras_AFTER
am_eras_BEFORE
am_eras_AFTER

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -274,6 +273,32 @@ public void setCachingEnabled(boolean enabled) {
 
     HelpMessages helpMessages;
 
+    /* For each calendar type, maps the closest two eras to 2025
+     * defined in that calendar to their corresponding start/end date
+     * TODO: include methods for calendarData in supplementalDataInfo API 
+     * to extract this data directly from supplementaldata.xml
+    */
+    public static final Map<String, String> calendarEras = new HashMap<String, String>() {{
```
macchiati, line 281: This can be much simpler. By using this format, you have to parse in public int[] getEraDateParts(String eraInfo).

Instead just map from a calendar type (a string) to a list of dates to format.

eg Map<String, List<Date>> ...
put("gregorian", Lists.of(firstDate, secondDate))

You don't actually need to know the era for each. Also, safest to use dates on 2 days before and 2 days after the era; that way you don't run into funny timezone problems.

Then the examples you store are already in the right format to hand to a calendar.

haytenf: Great point, thank you for the suggestion! I had a feeling I was overcomplicating things--I'll revise that.

macchiati: Sounds good; I'll watch for the result. BTW, the Date constructor is a odd for both the year and month (for historical reasons), so watch for that. That is, you need to use new Date(year-1900, month-1, day) to get the correct value.

haytenf: Thanks for the heads up--I was originally planning on using the Date constructor for this, but it seemed like it has been deprecated (from what I've seen [here](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#Date-int-int-int-)). I found a workaround that's a little more clunky code-wise (e.g.`new GregorianCalendar(1989, 0, 8).getTime()`) to avoid the deprecated method, but what do you think?

haytenf: Hello! Just following up on the previous question about which of the two constructors I should proceed with.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2861,6 +2888,87 @@ private String handleDateRangePattern(String value) {
...
+        List<Date> eraDates = CALENDAR_ERAS.get(id);
+        Date sample = eraDates.get(typeIndex);
+        String skeleton = "Gy";
+        String checkPath =
+                "//ldml/dates/calendars/calendar[@type=\""
+                        + calendarId
+                        + "\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\""
+                        + skeleton
+                        + "\"]";
+        String dateFormat = cldrFile.getWinningValue(checkPath);
```
macchiati, line 2920: The parts value has the original path in it, so you should use that.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2861,6 +2888,87 @@ private String handleDateRangePattern(String value) {
...
+        String type = parts.getAttributeValue(-1, "type");  // 1-indexed
+        int quarterIndex = Integer.parseInt(type) - 1;  
+        String skeleton = (width.equals("wide")) ? "yQQQQ" : "yQQQ";
+        String checkPath =
+                "//ldml/dates/calendars/calendar[@type=\""
+                        + calendarId
+                        + "\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\""
+                        + skeleton
+                        + "\"]";
+        String dateFormat = cldrFile.getWinningValue(checkPath);
```
macchiati, line 2955: As above, you should be using the path gotten from the parts variable.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2861,6 +2888,87 @@ private String handleDateRangePattern(String value) {
...
 
+    /**
+     * Add examples for eras. First checks if there is info for this
+     *  calendar type and this era type in the CALENDAR_ERAS map, then
+     * generates a sample date based on this info and formats it
+     */
+    private String handleEras(XPathParts parts, String value) {
+        String calendarId = parts.getAttributeValue(3, "type");
+        String type = parts.getAttributeValue(-1, "type");
+        String id = (calendarId.contains("islamic")) ? "islamic" : calendarId; // islamic variations map to same sample
```
macchiati, line 2899: Slightly better to use .startsWith()

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -1645,4 +1647,81 @@ public void TestUnicodeSetExamples() {
...
+            put("gregory", "gregorian");
+            put("ethioaa", "ethiopic-amete-alem");
+            put("islamic-civil", "islamic");
+            put("islamic-rgsa", "islamic");
+            put("islamic-tbla", "islamic");
+            put("islamic-umalqura", "islamic");
+            put("islamicc", "islamic");
+        }};
+        for (String id : calendars) {
+            if (id.equals("iso8601")) { // generic
```
macchiati, line 1665: iso8601 should be behave like gregorian, so you should go ahead as map it above and remove this.

macchiati (changes requested):
Just a few small things

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
haytenf: Got it

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
haytenf: Will change, thanks!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
haytenf: I want to make sure I understand this correctly--the original path would be something like `"//ldml/dates/calendars/calendar[@type="gregorian"]/eras/eraNames/era[@type="0"]"`, so to get the path I want to use for the formatting I'd do a few replacements (ex. `origPath.replace("eras", "dateTimeFormats").replace("eraNames", "availableFormats").replace(...);`, and similarly for the comment below. I guess I wanted to clarify that the path I'd get from the parts variable isn't the path I'm using in these two cases (although it can be modified to be).

macchiati: You are right! I read this too quickly. Yes, you do want to use the patterns from availableFormats in the example, as well as the value from this path. Same for the other one.

macchiati:
Hi, this has conflicts that need resolution. Can you do that? (If you don't know how, ask on Slack.).

macchiati (dismissed):

macchiati:
That is, I approved, but will need to reapprove once the conflicts are resolved.

macchiati (approved):

macchiati:
A commit statement failed. You will need to hit Details on the jira-ticket line above, and squash

haytenf:
Thanks, I've just squashed. All the checks look good from my end now.

PR #3919, opened 2024-07-30
CLDR-17014 Use special brackets to distinguish fallback values based on codes

btangmu:
-New CONSTRUCTED_VALUE_PREFIX, CONSTRUCTED_VALUE_SUFFIX, makeConstructedValue

CLDR-17014

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java
@@ -1595,7 +1595,22 @@ public abstract class XMLSource implements Freezable<XMLSource>, Iterable<String
         };
 
         private static final boolean SKIP_SINGLEZONES = false;
+
+        /**
+         * A special source for codes used as fallback. For example, the path
+         * //ldml/localeDisplayNames/languages/language[@type="blo"] gets a constructed value based
+         * on the code "blo".
+         */
         private static XMLSource constructedItems = new SimpleXMLSource(CODE_FALLBACK_ID);
+
+        /**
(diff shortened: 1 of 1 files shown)


btangmu:
There are a few test failures like

```
Test4528
Error:  (TestExampleGenerator.java:957)  Error: : specifics: expected "〖❬123.456,79 ❭BMD〗", got "〖❬123.456,79 ❭⟦BMD⟧〗"
```

and

```
TestLocaleDisplay
Warning:  (TestLocale.java:514)  Warning: 
Use -v to get samples for tests

Error:  (TestLocale.java:587)  Error: : en; en-u-cu-chf: expected "English (Currency: CHF)", got "English (Currency: ⟦CHF⟧)"

Error:  (TestLocale.java:587)  Error: : en; en-u-tz-uslax: expected "English (Time Zone: Los Angeles Time)", got "English (Time Zone: ⟦Los Angeles⟧ Time)"

Error:  (TestLocale.java:587)  Error: : af; zh-Hans-fonipa: expected "Chinees (Vereenvoudig, FONIPA)", got "Chinees (Vereenvoudig, ⟦FONIPA⟧)"
```

Evidently the "fallback" values are treated as correct, rather than as placeholders for missing data, for some paths, so it might be necessary to distinguish two kinds of path that currently both rely on `constructedItems`...

srl295:
I'm not sure about this approach, and I added comments on the ticket https://unicode-org.atlassian.net/browse/CLDR-17014

btangmu:
Instead of special brackets, we're now looking at simply having no fallback values; that can be implemented using extra paths, see https://github.com/unicode-org/cldr/pull/4254

PR #3922, opened 2024-07-31
CLDR-17843 Update ug.xml Add missing values to keep basic coverage.

conradarcturus:
Arabic and Arabic Extended are missing number formats -- just inherit. Same for symbols. Separately the exemplar characters need to be updated too.

https://unicode-org.atlassian.net/browse/CLDR-17843

CLDR-17843 

- [x] This PR completes part of the ticket.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ug.xml
+++ b/common/main/ug.xml
@@ -1157,6 +1157,7 @@ Reviewed by Waris Abdukerim Janbaz <oyghan@gmail.com> of the Uyghur Computer Sci
 		<exemplarCharacters type="auxiliary">[\u200E\u200F ئ]</exemplarCharacters>
 		<exemplarCharacters type="index" draft="contributed">[{ئا} {ئه} {ئو} {ئۇ} {ئۆ} {ئۈ} {ئې} {ئى} ا ە ب پ ت ج چ خ د ر ز ژ س ش غ ف ق ك گ ڭ ل م ن ھ و ۇ ۆ ۈ ۋ ې ى ي]</exemplarCharacters>
 		<exemplarCharacters type="numbers">↑↑↑</exemplarCharacters>
+		<exemplarCharacters type="punctuation">[\- ‑ , ; : ! ? . ( ) \[ \] { } ‹ › « »]</exemplarCharacters>
 	</characters>
 	<delimiters>
 		<quotationStart draft="contributed">»</quotationStart>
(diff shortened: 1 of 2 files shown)


srl295 (dismissed):

srl295:
remember we'll need to regenerate coverageLevels.txt after this merges as in #3897

conradarcturus:
I'm not able to run the "build" and "check" jobs -- is it a permissions thing? Waiting on that, all of my pull requests are blocked right now.

srl295:
Probably permissions

common/main/ug.xml
```diff
@@ -1157,6 +1157,7 @@ Reviewed by Waris Abdukerim Janbaz <oyghan@gmail.com> of the Uyghur Computer Sci
 		<exemplarCharacters type="auxiliary">[\u200E\u200F ئ]</exemplarCharacters>
 		<exemplarCharacters type="index" draft="contributed">[{ئا} {ئه} {ئو} {ئۇ} {ئۆ} {ئۈ} {ئې} {ئى} ا ە ب پ ت ج چ خ د ر ز ژ س ش غ ف ق ك گ ڭ ل م ن ھ و ۇ ۆ ۈ ۋ ې ى ي]</exemplarCharacters>
 		<exemplarCharacters type="numbers">↑↑↑</exemplarCharacters>
+		<exemplarCharacters type="punctuation">[\- ‑ , ; : ! ? . ( ) \[ \] { } ‹ › « »]</exemplarCharacters>
```
srl295, line 1160: @conradarcturus i think `{` and `}` need to be escaped. Are they really used here?

conradarcturus: Great! Nice to see the error message. I'll escape it.

I included the curly braces because they were inherited from root. https://st.unicode.org/cldr-apps/v#/ug/Alphabetic_Information/6cf943e652b01478

common/main/ug.xml
```diff
@@ -1157,7 +1157,7 @@ Reviewed by Waris Abdukerim Janbaz <oyghan@gmail.com> of the Uyghur Computer Sci
 		<exemplarCharacters type="auxiliary">[\u200E\u200F ئ]</exemplarCharacters>
 		<exemplarCharacters type="index" draft="contributed">[{ئا} {ئه} {ئو} {ئۇ} {ئۆ} {ئۈ} {ئې} {ئى} ا ە ب پ ت ج چ خ د ر ز ژ س ش غ ف ق ك گ ڭ ل م ن ھ و ۇ ۆ ۈ ۋ ې ى ي]</exemplarCharacters>
```
macchiati, line 1158: Re {...} on this like: The {} are needed in the index IF they surround a string with 2+ characters, and that sequence (eg, {ئا}) is something that in used in indexing (eg {SCH} is used in German indexes).

macchiati (dismissed):

srl295 (approved):
Approving

common/properties/coverageLevels.txt
```diff
@@ -152,6 +152,7 @@ tn ;	basic ;	Tswana
 to ;	basic ;	Tongan
 tr ;	modern ;	Turkish
 tt ;	moderate ;	Tatar
+ug ;	basic ;	Uyghur
```
macchiati, line 155: This will be overridden as soon as common/properties/coverageLevels.txt is regenerated. But harmless, so approving.

macchiati (approved):

common/properties/coverageLevels.txt
conradarcturus: Does that happen automatically or does that happen if a contributor runs the script? After this commit it should be always covered.

I figured this is a good test to prove my commit works too ;)

macchiati:
It gets rebuilt when we rebuild the charts. We currently don't do that with pushes (it takes a while), but we've talked about doing something like every day. (Or we could rebuild just part of the charts).

srl295:
> It gets rebuilt when we rebuild the charts. We currently don't do that with pushes (it takes a while), but we've talked about doing something like every day. (Or we could rebuild just part of the charts).

I think an additional job that runs on main (maybe daily) that *signals if the rebuild makes a difference* could be a lower impact way to handle this. Then we'd get notification, the charts are out of date, click here to update (as it'd be already built).

PR #3926, opened 2024-08-01
CLDR-17852 fix characters in ak.xml

DavidLRowe:
CLDR-17852

As requested in CLDR TC meeting 2024-07-31, this PR changes epsilon and open-o characters to match main exemplar. Specifically:
- epsilon U+0511 (Cyrillic range) changed to U+025B
- open-o U+037B (Greek range) changed to U+0254

Additionally some accented vowels (which occur in the existing data) are added to the auxiliary exemplar.

ALLOW_MANY_COMMITS=true


DavidLRowe:
@srl295 Ticket is now accepted. Is there an easy way to rerun the jira-ticket test? I suppose I could squash the branch to trigger rerunning all the tests.

macchiati (approved):

PR #3934, opened 2024-08-03
CLDR-16844 Show the Dashboard by default, and remember user preference

btangmu:
-New boolean dashWanted in cldrDashContext.mjs indicates whether the user wants the Dashboard open when appropriate

-New boolean method cldrDashContext.shouldBeShown takes into account dashWanted, and whether a Special page (other than GENERAL_SPECIAL) is open

-Add code to open the Dashboard (if shouldBeShown) where needed

-Encapsulate the string, general, as cldrLoad.GENERAL_SPECIAL

-Move message text to dash_needs_locale_and_coverage in cldrText.mjs, referenced in two places

-Fix bug: effectiveName called without locale arg; console.error if it happens again

-Remove confusing checkmark (vote.png) from General Info page; move style into GeneralInfo.vue

-Fix call to getClient, await is required

-Mark getClient as async to prevent VS Code wrongly warning that await is not needed

-Remove dead code

-Comments

CLDR-16844

ALLOW_MANY_COMMITS=true


tools/cldr-apps/js/src/esm/cldrClient.mjs
```diff
@@ -15,9 +15,9 @@ const RESOLVED_ROOT = new URL(OAS3_ROOT, document.baseURI);
  *
  * <https://github.com/swagger-api/swagger-js/blob/master/docs/usage/tags-interface.md#openapi-v3x>
  *
- * @returns Promise<SwaggerClient>
+ * @returns {Promise<SwaggerClient>}
  */
-function getClient() {
+async function getClient() {
   return new SwaggerClient({
```
srl295, line 21: I thought I changed this code to cache the client. Maybe it's stuck in some PR.

Looking at https://github.com/swagger-api/swagger-js/blob/master/docs/migration/migration-2-x-to-3-x.md#promises 

I think the following is what we should be doing.

```suggestion
  return await SwaggerClient({
```

btangmu: I'll test that locally...

tools/cldr-apps/js/src/esm/cldrVueMap.mjs
```diff
@@ -24,7 +26,7 @@ const specialToComponentMap = {
   add_user: AddUser,
   auto_import: AutoImport,
   downgraded: DowngradedVotes,
-  general: GeneralInfo,
+  general: GeneralInfo, // see cldrLoad.GENERAL_SPECIAL
```
srl295, line 29: we could integrate this better, but this is sufficient for now.

tools/cldr-apps/js/src/views/GeneralInfo.vue
```diff
@@ -52,8 +53,23 @@ export default {
...
   },
+
   methods: {
+    showDashboardOrButton() {
+      // If Dashboard is already visible, hide the button for opening it.
+      if (cldrDashContext.isVisible()) {
+        const els = document.getElementsByClassName("general-open-dash");
+        for (let i = 0; i < els.length; i++) {
+          els[i].style.display = "none";
+        }
```
srl295, line 67: why this and not a reactive vue property?

srl295 (dismissed):
it looks OK, some suggestions for improvement which could be future.

tools/cldr-apps/js/src/esm/cldrClient.mjs
srl295: the `async` nor `await` is not needed in that case actually. just pass `return SwaggerClient` as a Promise back to the callers who await it.

srl295: I'd say if the PR as is is working let's address this in a future one?

btangmu: Oh, well... I just tested with changing new to await, and it worked, so I made a commit...

tools/cldr-apps/js/src/views/GeneralInfo.vue
btangmu: It was based on similar code in cldrDashContext.mjs. I can change it to a property...

tools/cldr-apps/js/src/esm/cldrClient.mjs
btangmu: I'll test `return SwaggerClient` and make that change if it works

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrVueMap.mjs
btangmu: Really all of these keys could benefit from encapsulation. I'm not sure what how/where, though... using `cldrLoad.GENERAL_SPECIAL` in place of `general` for the key isn't allowed in the initializer, but we could have initialization code like
```
specialToComponentMap[cldrLoad.GENERAL_SPECIAL] = GeneralInfo;
```

and likewise for the other keys...

srl295 (approved):

tools/cldr-apps/js/src/views/GeneralInfo.vue
```diff
@@ -7,12 +7,12 @@
...
       class="cldr-nav-btn btn-primary open-dash"
       type="button"
-      v-if="!dashButtonShouldBeVisible()"
+      v-if="dashButtonShouldBeVisible()"
     >
       Open Dashboard
     </button>
 
-    <!-- <cldr-overall-errors /> -->
+    <cldr-overall-errors />
```
srl295: wait, did you mean to re-enable this? Why? I think we need to discuss first!

tools/cldr-apps/js/src/views/GeneralInfo.vue
btangmu: Changed to a property

tools/cldr-apps/js/src/views/GeneralInfo.vue
btangmu: I saw it too just after the commit, sorry about the noise! It was temporary for testing the call to cldrClient that motivated the whole tangent

srl295 (approved):

PR #3938, opened 2024-08-06
CLDR-17843 Restore uz_Cyrl to basic coverage by fixing numerals

conradarcturus:
This adds values for the symbols and number formats for `numberSystem="arab"` -- copying over the values from `numberSystem="arabext"`. The problem was that arab was added as a number system for uz.xml (== uz_Latn.xml) so uz_Cyrl.xml needed declarations.

Ran `java -jar tools/cldr-code/target/cldr-code.jar ShowLocaleCoverage` to regenerate coverage level text.

CLDR-17843

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/properties/coverageLevels.txt
+++ b/common/properties/coverageLevels.txt
@@ -155,6 +155,7 @@ tt ;	moderate ;	Tatar
 uk ;	modern ;	Ukrainian
 ur ;	modern ;	Urdu
 uz ;	modern ;	Uzbek
+uz_Cyrl ;	basic ;	Uzbek (Cyrillic)
 vec ;	moderate ;	Venetian
 vi ;	modern ;	Vietnamese
 vmw ;	basic ;	Makhuwa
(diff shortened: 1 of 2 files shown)


srl295 (approved):

common/properties/coverageLevels.txt
```diff
@@ -155,6 +155,7 @@ tt ;	moderate ;	Tatar
 uk ;	modern ;	Ukrainian
 ur ;	modern ;	Urdu
 uz ;	modern ;	Uzbek
+uz_Cyrl ;	basic ;	Uzbek (Cyrillic)
```
macchiati, line 158: common/properties/coverageLevels.txt is automatically rebuilt, so shouldn't have manual changes.

conradarcturus: Er, it's in a manual commit but I ran the script to make the change. I'm used to re-generating artefacts in the commit that changes them rather than waiting for scheduled commit -- just to keep everything up to date. It's also a good validation check to make sure the change did the right thing.

PR #3941, opened 2024-08-07
CLDR-17868 v46 BRS: algorithmic locales

srl295:
CLDR-17868

- [X] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #3943, opened 2024-08-07
CLDR-17866 BRSv46 CLDRModify passes before alpha1

pedberg-icu:
CLDR-17866

CLDRModify passes before alpha1:

1. With no arguments:
    * en_ZA, remove trailing space after element
    * he, replace entity ref for LRM with actual char
    * others, reorder entries (in annotations, specifically reorder entries for "₨", "﷼" due to collation updates)
2. With -fP: reformat/reorder exemplar & parseLenient sets, adjust literal quoting.

Note that the -fQ pass was explicitly skipped per Mark comment on previous CLDRModify PR: " I don't think we should add the tts names to the keywords, unless the keywords are empty, because it works against the guidance that people should break up the search keywords by word where possible...I suggest that we hold off adding the tts name, and just do the daip (which reorders)".

ALLOW_MANY_COMMITS=true


srl295:
I might need to rerun these for #3941

AEApple:
Makes sense, but shouldn't we just approve and merge this one and then just do another CLDRModify pass if there are additional changes that need to be modified?

srl295 (approved):
RSLGTM

srl295:
> Makes sense, but shouldn't we just approve and merge this one and then just do another CLDRModify pass if there are additional changes that need to be modified?

sorry, i had meant to approve this.

PR #3945, opened 2024-08-07
CLDR-17871 Create yue_Hant_CN stub locale

AEApple:
CLDR-17871

- Create stub locale for yue_Hant_CN.xml
- Update population data

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -2784,6 +2784,7 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="zh" populationPercent="90" officialStatus="official"/>	<!--Chinese-->
 			<languagePopulation type="wuu" populationPercent="6"/>	<!--Wu Chinese-->
 			<languagePopulation type="yue_Hans" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
+			<languagePopulation type="yue_Hant" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
 			<languagePopulation type="hsn" populationPercent="2.9"/>	<!--Xiang Chinese-->
 			<languagePopulation type="hak" populationPercent="2.3"/>	<!--Hakka Chinese-->
 			<languagePopulation type="nan" populationPercent="1.9"/>	<!--Min Nan Chinese-->
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -2783,8 +2783,8 @@ XXX Code for transations where no currency is involved
 		<territory type="CN" gdp="25360000000000" literacyPercent="95.1" population="1394020000">	<!--China-->
 			<languagePopulation type="zh" populationPercent="90" officialStatus="official"/>	<!--Chinese-->
 			<languagePopulation type="wuu" populationPercent="6"/>	<!--Wu Chinese-->
-			<languagePopulation type="yue" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
-			<languagePopulation type="yue_Hant" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
+			<languagePopulation type="yue_Hans" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
+			<languagePopulation type="yue" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
 			<languagePopulation type="hsn" populationPercent="2.9"/>	<!--Xiang Chinese-->
 			<languagePopulation type="hak" populationPercent="2.3"/>	<!--Hakka Chinese-->
 			<languagePopulation type="nan" populationPercent="1.9"/>	<!--Min Nan Chinese-->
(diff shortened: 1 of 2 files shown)


common/supplemental/supplementalData.xml
```diff
@@ -2784,6 +2784,7 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="zh" populationPercent="90" officialStatus="official"/>	<!--Chinese-->
 			<languagePopulation type="wuu" populationPercent="6"/>	<!--Wu Chinese-->
 			<languagePopulation type="yue_Hans" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
+			<languagePopulation type="yue_Hant" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
```
macchiati, line 2787: Since Hant is the default for yue, this should just be:
```suggestion
			<languagePopulation type="yue" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
```

macchiati (dismissed):

common/supplemental/supplementalData.xml
AEApple: Okay, but we're not changing the default in this ticket since we're still waiting for input, so I'll change yue_Hans to yue for now.

AEApple:
Okay, I think what I don't understand here is where the default is defined and if I need to do something specific since before yue_CN only had a stub locale for one script. Where is it set which script a locale uses by default?

common/supplemental/supplementalData.xml
```diff
@@ -2783,7 +2783,8 @@ XXX Code for transations where no currency is involved
 		<territory type="CN" gdp="25360000000000" literacyPercent="95.1" population="1394020000">	<!--China-->
 			<languagePopulation type="zh" populationPercent="90" officialStatus="official"/>	<!--Chinese-->
 			<languagePopulation type="wuu" populationPercent="6"/>	<!--Wu Chinese-->
-			<languagePopulation type="yue_Hans" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
+			<languagePopulation type="yue" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
+			<languagePopulation type="yue_Hant" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
```
macchiati, line 2787: These should be the following, to correct one of the errors. 

```suggestion
			<languagePopulation type="yue_Hans" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
			<languagePopulation type="yue" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
```

That is, the current defaults (which we are not changing yet) are:

yue has as default script Hant // in other words, if in isolation, yue_Hant == yue
yue_CN has as default script Hans // in other words, yue_Hans_CN == yue_CN

So yue_Hant normalizes to yue in the supplemental data.

The defaults are set by defaultContents and by likely subtags in the supplemental data, but we shouldn't need to mess with those for this ticket. 

See if that fixes the error, otherwise I'll look into it.

macchiati (changes requested):

pedberg-icu:
> Where is it set which script a locale uses by default?

There are several places where that is indicated, not sure which has primacy:
- In likelySubtags, you can check the likelySubtags for the language code by itself (`yue`→`yue_Hant_HK`).
- In supplementalMetadata, you can check the defaultContent locales for the language code (`yue_Hant` is defaultContent for `yue`).
- In supplementalData, you can check the languageData, if there is only one non-secondary script for the language (does not apply to `yue`), that is presumably the primary.

AEApple:
> > Where is it set which script a locale uses by default?
> 
> There are several places where that is indicated, not sure which has primacy:
> 
> * In likelySubtags, you can check the likelySubtags for the language code by itself (`yue`→`yue_Hant_HK`).
> * In supplementalMetadata, you can check the defaultContent locales for the language code (`yue_Hant` is defaultContent for `yue`).
> * In supplementalData, you can check the languageData, if there is only one non-secondary script for the language (does not apply to `yue`), that is presumably the primary.

Will check them all next.

macchiati:
These should all be *consistent* with the default, but the original determination is by likelySubtags, and the tests ensure that default content etc. are consistent with that.

common/supplemental/supplementalData.xml
pedberg-icu: This is wrong. `Hant` is the default for `yue` so `yue` by itself is the same thing as `yue_Hant`. We already had `yue_Hans` and need to keep that (the non-default script), we just needed to add `yue` by itself.

macchiati: right, that's what I had in [In supplementalData, you can check the languageData, if there is only one non-secondary script for the language (does not apply to yue), that is presumably the primary.](https://github.com/unicode-org/cldr/pull/3945#pullrequestreview-2226108995) [just different order]

macchiati (approved):

common/supplemental/supplementalData.xml
AEApple: Okay, I will fix it. I wasn't sure how the script defaults are handled in this file vs elsewhere. Thank you for answering all my questions :)

common/supplemental/supplementalData.xml
```diff
@@ -2784,6 +2784,7 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="zh" populationPercent="90" officialStatus="official"/>	<!--Chinese-->
 			<languagePopulation type="wuu" populationPercent="6"/>	<!--Wu Chinese-->
 			<languagePopulation type="yue_Hans" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Simplified)-->
+			<languagePopulation type="yue" populationPercent="5.2" references="R1219"/>	<!--Cantonese (Traditional)-->
```
srl295, line 2787: the .tsv file wasn't changed, but it'll get picked up when that is rerun.

PR #3947, opened 2024-08-07
CLDR-17320 check fixes to avoid stack traces

srl295:
- update CheckExemplars parseLenient to error instead of crash on null
- update CheckPlaceHolders to skip aliases instead of crashing
- remove the logknownissue which was to suppress the crash stack traces

CLDR-17320

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
@@ -302,6 +302,16 @@ public class CheckExemplars extends FactoryCheckCLDR {
 
     private void checkParse(
             String path, String fullPath, String value, Options options, List<CheckStatus> result) {
+        if (value == null) {
+            CheckStatus message =
+                    new CheckStatus()
+                            .setCause(this)
+                            .setMainType(CheckStatus.errorType)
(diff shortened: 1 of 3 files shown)


srl295:
without this, log known issues was untenable to analyze, with thousands of stack traces.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
```diff
@@ -302,6 +302,16 @@ public CheckCLDR handleCheck(
 
     private void checkParse(
             String path, String fullPath, String value, Options options, List<CheckStatus> result) {
+        if (value == null) {
```
macchiati, line 305: It is very strange that a null value could occur here. Probably only via a manually modified XML file, since it would have to be a path that didn't occur in root.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
srl295: it's from the unit test. so, it's just a robustness issue, which is presumably what the test is trying to check?

PR #3948, opened 2024-08-08
CLDR-17587 Add Haitian Creole

conradarcturus:
This adds core data for Haitian Creole so that in Fall 2024 / v47 we can fill out the information for this locale.

CLDR-17587

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ht.xml
+++ b/common/main/ht.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2024 Unicode, Inc.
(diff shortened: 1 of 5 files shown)


srl295:
Scheduling - this should target v47 and be merged *after* main is branched for v46

srl295 (changes requested):
Generally LGTM though:
- [x] needs to wait for v47 - maybe better to keep this in draft until then?
- [x] the explicit parentage fr -> ht is allowed but the parent locale was not set in this PR yet
- [ ] need to run spotless to cleanup the code ( see the `tools/README.md#spotless` docs )

conradarcturus:
@srl295  are there still any changes that need to be made?

srl295 (approved):

PR #3949, opened 2024-08-08
CLDR-10347 show one example for each day period, excluding midnight

haytenf:
CLDR-10347 Show one example for each day period. Currently suppresses examples for the day period "midnight" (due to [ICU-12278](https://unicode-org.atlassian.net/browse/ICU-12278), lack of controlled use of midnight), but handles remaining day periods and adds relevant samples.

| Locale           | Before | After |
| -------------- | -------------- |  -------------- |
| he       | Screenshot 2024-08-09 at 1 04 25 PM Screenshot 2024-08-09 at 1 04 46 PM| Screenshot 2024-08-09 at 1 05 36 PM Screenshot 2024-08-09 at 1 05 52 PM |
| de       | Screenshot 2024-08-09 at 12 50 09 PM | Screenshot 2024-08-09 at 12 53 02 PM |
| qu*      | Screenshot 2024-08-09 at 12 56 29 PM | Screenshot 2024-08-09 at 12 57 45 PM |

*Only 2 day periods are defined in Quechua, original code produces 3 examples for every locale, so modification results in fewer examples

ALLOW_MANY_COMMITS=true

[ICU-12278]: https://unicode-org.atlassian.net/browse/ICU-12278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -168,7 +168,6 @@ public class ExampleGenerator {
 
     private static final Date DATE_SAMPLE2;
     private static final Date DATE_SAMPLE3;
-    private static final Date DATE_SAMPLE4;
 
     static {
         Calendar calendar = Calendar.getInstance(ZONE_SAMPLE, ULocale.ENGLISH);
@@ -180,8 +179,6 @@ public class ExampleGenerator {
 
         calendar.set(1999, 8, 5, 7, 0, 0); // 1999-09-05 07:00:00
         DATE_SAMPLE3 = calendar.getTime();
-        calendar.set(1999, 8, 5, 23, 0, 0); // 1999-09-05 23:00:00
-        DATE_SAMPLE4 = calendar.getTime();
     }
 
     static final List<DecimalQuantity> CURRENCY_SAMPLES =
@@ -2634,8 +2631,8 @@ public class ExampleGenerator {
                         }
                         R3<Integer, Integer, Boolean> info = dayPeriodInfo.getFirstDayPeriodInfo(dayPeriod);
                         if (info != null) {
-                            int time = ((info.get0() + info.get1()) / 2); // time midpoint 
-                            String formatted = sdf.format(time);                      
+                            int time = ((info.get0() + info.get1()) / 2); // dayPeriod endpoints overlap, midpoint to disambiguate
+                            String formatted = sdf.format(time);
                             examples.add(formatted);
                         }
                     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -2623,15 +2623,23 @@ public class ExampleGenerator {
                     }
                 } else {
                     List<String> examples = new ArrayList<>();
-                    DayPeriodInfo dayPeriodInfo = supplementalDataInfo.getDayPeriods(DayPeriodInfo.Type.format, cldrFile.getLocaleID());
-                    Set<DayPeriod> dayPeriods = new LinkedHashSet<DayPeriod>(dayPeriodInfo.getPeriods());
+                    DayPeriodInfo dayPeriodInfo =
+                            supplementalDataInfo.getDayPeriods(
+                                    DayPeriodInfo.Type.format, cldrFile.getLocaleID());
+                    Set<DayPeriod> dayPeriods =
+                            new LinkedHashSet<DayPeriod>(dayPeriodInfo.getPeriods());
                     for (DayPeriod dayPeriod : dayPeriods) {
-                        if (dayPeriod.equals(DayPeriod.midnight)) { // suppress midnight, see ICU-12278 bug
+                        if (dayPeriod.equals(
+                                DayPeriod.midnight)) { // suppress midnight, see ICU-12278 bug
                             continue;
                         }
-                        R3<Integer, Integer, Boolean> info = dayPeriodInfo.getFirstDayPeriodInfo(dayPeriod);
+                        R3<Integer, Integer, Boolean> info =
+                                dayPeriodInfo.getFirstDayPeriodInfo(dayPeriod);
                         if (info != null) {
-                            int time = ((info.get0() + info.get1()) / 2); // dayPeriod endpoints overlap, midpoint to disambiguate
+                            int time =
+                                    ((info.get0() + info.get1())
+                                            / 2); // dayPeriod endpoints overlap, midpoint to
+                            // disambiguate
                             String formatted = sdf.format(time);
                             examples.add(formatted);
                         }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -2597,7 +2597,6 @@ public class ExampleGenerator {
                 String numbersOverride = parts.findAttributeValue("pattern", "numbers");
                 SimpleDateFormat sdf =
                         icuServiceBuilder.getDateFormat(calendar, value, numbersOverride);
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -1123,6 +1123,56 @@ private void checkDayPeriod(
         checkPathValue(exampleGenerator, path, cldrFile.getStringValue(path), expected);
     }
 
+    public void TestAllDayPeriods() { 
+        // excludes midnight, see ICU-12278 
+        String[][] tests = {
```
conradarcturus, line 1128: Yay! I love this test. My advice is only on style -- I think this test gets the job done well. I know many other tests in this file follow this paradigm but I'm personally allergic to it :p so feel free to ignore my advice or change it.

String[][] types are very unsafe -- it's not clear what the strings in the middle represent and they have undetermined length.

Rather, a better idea would be to make a new method like to `checkValue` -- but specific to the kind of test you are running `checkDayPeriodsForLocale`
https://github.com/unicode-org/cldr/blob/cf294f071f911fc751b1af5fff7a0410a7ab2a70/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java#L454-L461

In that test you will include the code that you have now in your for loop down below.

And then you will use it individually for each example, and now things will have nice variable names
https://github.com/unicode-org/cldr/blob/cf294f071f911fc751b1af5fff7a0410a7ab2a70/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java#L447-L452

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2626,9 +2626,19 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
...
-                    examples.add(sdf.format(DATE_SAMPLE4));
+                    DayPeriodInfo dayPeriodInfo = supplementalDataInfo.getDayPeriods(DayPeriodInfo.Type.format, cldrFile.getLocaleID());
+                    Set<DayPeriod> dayPeriods = new LinkedHashSet<DayPeriod>(dayPeriodInfo.getPeriods());
+                    for (DayPeriod dayPeriod : dayPeriods) {
+                        if (dayPeriod.equals(DayPeriod.midnight)) { // suppress midnight, see ICU-12278 bug
+                            continue;
+                        }
+                        R3<Integer, Integer, Boolean> info = dayPeriodInfo.getFirstDayPeriodInfo(dayPeriod);
+                        if (info != null) {
+                            int time = ((info.get0() + info.get1()) / 2); // time midpoint
```
conradarcturus, line 2637: Why the midpoint? I'm cool with it, I'm just not sure why then midpoint was chosen.

conradarcturus (commented):
Thanks for adding these :) It looks great. My old feedback is based on style and I'm myself also pretty new to CLDR so any feedback is optional.

Thanks for including screenshots. A recommendation I have for those screenshots to make it more readable is to 1) zoom in on the specific areas that the difference is (the context is good for like 1 screenshot but then the others are just harder to quickly grok). and 2) use markdown tables like this:

|Context|Before|After|
|--|--|--|
|I'm showing X|Insert the before image|Insert the after image
|I'm now showing Y|Insert the before image|Insert the after image

Regardless none of these comments are blocking so I'm happy to accept if you think its ready -- but I'll leave it open ended in case one of the more seasoned contributors wants to take a look.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2626,9 +2626,19 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
                     }
                 } else {
                     List<String> examples = new ArrayList<>();
-                    examples.add(sdf.format(DATE_SAMPLE3));
-                    examples.add(sdf.format(DATE_SAMPLE));
-                    examples.add(sdf.format(DATE_SAMPLE4));
```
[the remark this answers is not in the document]
haytenf: Good catch, thank you!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
haytenf: You're right, the midpoint specifically is an arbitrary choice--it could be any value in between the start/end time. Setting the sample to either the start time (`info.get0()`) or the end time (`info.get1()`) doesn't predictably get the intended day period, since day periods have overlapping start/end times (ex. noon starts/ends at 12pm and afternoon starts at 12pm, so using the start time as the sample produces two examples formatted as 12:00 noon, vs two distinct examples). So, I thought the midpoint would be the simplest in-between value :).

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
haytenf: Thanks for the pointer! I didn't know that about String[][], but your explanation makes sense and I'll revise the tests adding a `checkValue` type method instead.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
haytenf: I'll modify the comment to include that the midpoint was chosen to disambiguate!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -180,8 +179,6 @@ public class ExampleGenerator {
 
         calendar.set(1999, 8, 5, 7, 0, 0); // 1999-09-05 07:00:00
```
srl295, line 180: Hm. These actually will vary by locale, see:

- https://www.unicode.org/cldr/charts/latest/supplemental/day_periods.html 
- https://github.com/unicode-org/cldr/blob/main/common/supplemental/dayPeriods.xml
- from the class `SupplementalDataInfo` there is `getDayPeriodLocales()` and `getDayPeriods()` that should give you the period information.

That said, I think this is an improvement and we could go forward with it- but perhaps a followon to try to generate the dates based on the rules.

srl295 (dismissed):
This looks great!

macchiati (dismissed):

macchiati:
The checks weren't run while in draft, so now running

srl295:
@haytenf the [Fixed formatting using spotless](https://github.com/unicode-org/cldr/pull/3949/commits/2d1b73d95762b4b3ddcaf34af3a404bfd32488b1)  commit message needed to be `CLDR-10347 Fixed formatting using spotless` - you can fix it by one of these two:

1. https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/cldr/3949 click squash (this will make it more complicated on your local system) OR
2. `git commit --amend` locally - this will let you update the commit message and then run `git push -f` to push the branch back

macchiati:
"jira-ticket — Commit message for 2d1b73d fails validation"
As Steven says.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -1123,6 +1123,42 @@ private void checkDayPeriod(
         checkPathValue(exampleGenerator, path, cldrFile.getStringValue(path), expected);
     }
 
+    public void TestAllDayPeriods() { // excludes midnight, see ICU-12278
+        checkDayPeriodsForLocale(
```
conradarcturus, line 1127: Perfetto!

conradarcturus (approved):
Awesome! Looks great.

thumbs up cat

PR #3950, opened 2024-08-09
CLDR-17856 nan and hak in Taiwan

DavidLRowe:
CLDR-17856

ALLOW_MANY_COMMITS=true


macchiati:
The data changes for language population per region need to be made in the supplemental XML file

macchiati (changes requested):
The changes need to be made in the supplemental XML file.

srl295:
> The changes need to be made in the supplemental XML file.

they are made in country_language_population.tsv first, and then `ShowLanguages` is run to update the xml.

srl295:
probably wait until #3952 is merged before this goes in - running ConvertLanguageData to include the xml update.

srl295 (approved):
@macchiati how about we change the spreadsheet, and i'll regenerate the xml as part of my BRS?

macchiati:
> @macchiati how about we change the spreadsheet, and i'll regenerate the xml as part of my BRS?

You mean, change the tsv? Ok if so.

srl295:
> > @macchiati how about we change the spreadsheet, and i'll regenerate the xml as part of my BRS?
> 
> You mean, change the tsv? Ok if so.

Yes, merge this PR, and then the XML will be regenerated as part of BRS.

DavidLRowe:
@macchiati It seems you need to modify your request for changes.

PR #3956, opened 2024-08-13
CLDR-17751 add ZWG, drop ZWL

srl295:
- update ZWG per ISO amendment
- retain USD as valid due to feedback and articles
- update English

CLDR-17751

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -6636,6 +6636,11 @@ annotations.
 				<displayName count="one">Zimbabwean dollar (1980–2008)</displayName>
 				<displayName count="other">Zimbabwean dollars (1980–2008)</displayName>
 			</currency>
+			<currency type="ZWG">
+				<displayName>Zimbabwean Gold</displayName>
+				<displayName count="one">Zimbabwean Gold</displayName>
+				<displayName count="other">Zimbabwean Gold</displayName>
+			</currency>
 			<currency type="ZWL">
 				<displayName>Zimbabwean Dollar (2009)</displayName>
 				<displayName count="one">Zimbabwean dollar (2009)</displayName>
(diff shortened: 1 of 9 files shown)


AEApple:
Where is ZWG?

macchiati:
My thought too!

macchiati (approved):

srl295:
> My thought too!

yea… missing something important!

AEApple (approved):

AEApple:
This actually looks wrong in staging? ZWL and ZWG don't appear properly split out. Also I expect ZWL to have a date range, but only one date is showing?

macchiati:
What are you seeing on staging?

(if there is no "to" value, that just means that there is no set 'termination' date. That's typically for the current currency valid for a country)

common/main/en.xml
```diff
@@ -6636,6 +6636,11 @@ annotations.
 				<displayName count="one">Zimbabwean dollar (1980–2008)</displayName>
 				<displayName count="other">Zimbabwean dollars (1980–2008)</displayName>
 			</currency>
+			<currency type="ZWG">
+				<displayName>Zimbabwean Gold</displayName>
+				<displayName count="one">Zimbabwean Gold</displayName>
+				<displayName count="other">Zimbabwean Gold</displayName>
```
roozbehp, line 6642: I think the cases for one and other should use a lowercase "gold".

roozbehp (changes requested):

roozbehp (approved):

roozbehp (approved):

srl295:
@roozbehp appreciate your re-review. Github was down when I made the other, so i forgot that it was still in progress.

srl295:
@macchiati got a null pointer in `en_Dsrt` - it was trying to compare it as a logical group

srl295:
I found the issue. CountItems silently completed because I passed the -Dmethod parameter incorrectly. Noted this on https://unicode-org.atlassian.net/browse/CLDR-17555

roozbehp (approved):

PR #3960, opened 2024-08-14
CLDR-17226 Unicode 16 final collation data

markusicu:
CLDR-17226

- pinyin order changes from Unihan kMandarin
- stroke order changes from Unihan kTotalStrokes
- FractionalUCA `[radical]`-stroke order change & unihan-index change:
  - Match the order in [UAX38 section 2.1.2 Sorting Algorithm Used by the Radical-Stroke Indexes](https://www.unicode.org/reports/tr38/#SortingAlgorithm),
  - generated from https://github.com/unicode-org/unicodetools/pull/909
  - for CLDR-17224 FractionalUCA.txt: review radical-stroke order / re: simplified radicals
  - = https://github.com/unicode-org/unicodetools/issues/620

- [ ] This PR completes the ticket. -- maybe, will double-check

ALLOW_MANY_COMMITS=true


markusicu:
Tests were passing, then I pushed an updated of the documentation of the `[radical ...]` lines.
Current docs see https://www.unicode.org/reports/tr35/tr35-collation.html#File_Format_FractionalUCA_txt

pedberg-icu (approved):

PR #3962, opened 2024-08-15
CLDR-17892 Added aliases from "iso8601" to "gregorian", since "iso860…

richgillam:
…1" is a valid calendar ID but doesn't have

any actual associated data in CLDR.

CLDR-17892

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -1936,6 +1936,32 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					<alias source="locale" path="../../calendar[@type='islamic']/dateTimeFormats"/>
 				</dateTimeFormats>
 			</calendar>
+			<calendar type="iso8601">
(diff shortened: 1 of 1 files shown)


macchiati:
I don't think we want to handle it this way. See my comments on the ticket. If we wanted a temporary fix, we could copy the gregorian data (not alias it) in root into a new calendar element with iso8601.

macchiati (changes requested):
See other comment

richgillam:
Based on your comments in the ticket and in email, I think the right answer is for me to close this PR and for CLDR TC to assign the ticket to somebody more qualified to come up with a good answer.  I can sit on the ICU ticket until we have that.  There's no urgency coming from me (I picked this ticket to work on because I thought it would be easy); if this bleeds into 47/77, that's okay with me.

PR #3968, opened 2024-08-18
CLDR-17799 Use Vue components for forum form and buttons

btangmu:
-New ForumButton.vue, ForumForm.vue; revise cldrForum.mjs; related refactoring

CLDR-17799

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrForum.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrForum.mjs
@@ -10,9 +10,13 @@ import * as cldrForumFilter from "./cldrForumFilter.mjs";
 import * as cldrForumPanel from "./cldrForumPanel.mjs";
 import * as cldrForumType from "./cldrForumType.mjs";
 import * as cldrLoad from "./cldrLoad.mjs";
+import * as cldrNotify from "./cldrNotify.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
 import * as cldrText from "./cldrText.mjs";
+import * as cldrVue from "./cldrVue.mjs";
+
+import ForumButton from "../views/ForumButton.vue";
 
 class PostInfo {
   constructor(postType) {
@@ -25,23 +29,25 @@ class PostInfo {
     this.willFlag = false;
     this.replyTo = -1;
     this.parentPost = null;
+    this.isOpen = true; // may change in setReplyTo
   }
 
   /**
    * Set the parent data for a reply.
    *
-   * @param {Object} post - the data (as received from the back end) for the post to which
+   * @param {Object} rootPost - the data (as received from the back end) for the post to which
    * this is a reply. It is NOT a PostInfo object!
    */
-  setReplyTo(post) {
-    this.replyTo = post.id;
-    this.parentPost = post;
+  setReplyTo(rootPost) {
+    this.replyTo = rootPost.id;
+    this.parentPost = rootPost;
+    this.isOpen = rootPost.open;
     // Copy these values from the parent.
     this.setLocalePathValueSubject(
-      post.locale,
-      post.xpath,
-      post.value,
-      post.subject
+      rootPost.locale,
+      rootPost.xpath,
+      rootPost.value,
+      rootPost.subject
     );
   }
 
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/js/src/esm/cldrForumPanel.mjs
```diff
@@ -12,6 +12,11 @@ import * as cldrSurvey from "./cldrSurvey.mjs";
 import * as cldrTable from "./cldrTable.mjs";
 import * as cldrText from "./cldrText.mjs";
 
+/**
+ * Encapsulate this class name -- caution: it's used literally in surveytool.css
+ */
+const FORUM_DIV_CLASS = "forumDiv";
```
btangmu, line 18: FORUM_DIV_CLASS is moved from cldrForum to cldrForumPanel since no longer needs export

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
```diff
@@ -103,20 +104,16 @@ function getXpathMap() {
...
- * user clicks an add ("+") button. Added "popover-content" per https://unicode.org/cldr/trac/ticket/11265.
- *
- * Called only from CldrSurveyVettingLoader.js
+ * @return true if busy
  */
 function isInputBusy() {
-  if (!window.getSelection) {
-    return false;
+  if (cldrForum.isFormVisible()) {
+    return true;
```
btangmu, line 111: Adding this condition to isInputBusy is not elegant. It reflects the fact that we have Vue components (buttons) embedded in legacy DOM elements in the Info Panel, and the whole form is attached to one of those buttons and would get obliterated by an auto-scheduled page update if we didn't prevent it

tools/cldr-apps/js/src/views/ForumButton.vue
```diff
@@ -0,0 +1,104 @@
+<script setup>
```
btangmu, line 1: "script setup" appears to be the latest recommended syntax, and so is putting script, template, style in that order

See:

https://vuejs.org/api/sfc-script-setup
https://vuejs.org/style-guide/rules-recommended#single-file-component-top-level-element-order

tools/cldr-apps/js/src/views/ForumForm.vue
```diff
@@ -0,0 +1,140 @@
...
+    .catch((error) => {
+      if (DEBUG_FORUM_FORM) {
+        console.log("onSubmit validation error", error);
+      }
+    });
+}
+
+function formatParent(parentPost) {
+  const div = cldrForum.parseContent([parentPost], "parent");
+  return new XMLSerializer().serializeToString(div);
```
btangmu, line 73: This method takes DOM elements generated by legacy code (cldrForum.parseContent) and turns them into html using v-html. In the short term this is easy and achieves consistency with the legacy formatting. Eventually of course there should be a Vue component for rendering completed forum posts.

tools/cldr-apps/js/src/views/ForumButton.vue
srl295: it's very handy, has some caveats.

tools/cldr-apps/js/src/esm/cldrForum.mjs
```diff
@@ -464,28 +287,23 @@ function sendPostRequest(form) {
...
       cldrSurvey.expediteStatusUpdate(); // update forum icons (👁️‍🗨️, 💬) in the main table
     }
   } else {
-    const post = $(".post").first();
-    post.before(
-      "<i>Your post was added, #" + data.postId + " but could not be shown.</i>"
-    );
+    const message =
+      "Your post was added, #" + data.postId + " but could not be shown.";
+    cldrNotify.error("Error posting to Forum", message);
```
srl295, line 301: much better

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -452,6 +455,19 @@ function addForumPanel(tr) {
   }
 }
 
+/**
+ * An empty paragraph at the bottom of the info panel enables scrolling
+ * to bring the bottom content fully into view without being overlapped
+ * by the xpath shown by addXpath
+ */
+function addBottom() {
```
srl295, line 463: should it be `padding-bottom: 2em;` or something in the CSS for the element instead? that would have a similar result without requiring mutation of the html

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrInfo.mjs
btangmu: This is quick and avoids going deep into the legacy css. The order of elements in the info panel seems likely to change. Having this new "bottom" element at the bottom keeps its style separate from the rest.

PR #3970, opened 2024-08-19
CLDR-17868 v46 BRS: derived annotations, cldrmodify

srl295:
CLDR-17868

I also fixed a tool issue that's in several of our tools:  where missing empty parents aren't added.  New class `CLDRTreeWriter` to support writing a tree full of data.  It tracks additions and deletions and fills in missing items.   Examples:

1. if `ca_ES_VALENCIA.xml` is  added, it will automatically write `ca_ES.xml`
2. if you request to remove `sr_Cyrl` it wont' allow it since it has children, it will instead just write an empty file.

`CLDRTreeWriter` is AutoCloseable ;and must be closed after use.

ALLOW_MANY_COMMITS=true


macchiati (dismissed):
Looks like the annotations/* ones were mostly because collation changed.

macchiati (changes requested):
> where missing empty parents aren't added

That is a good addition, *for the main directory*. Logically, all of the other directories are read in conjunction with main, so an empty file in main is sufficient.

So the tool (and files) needs to be changes to only create empty files, etc in the main directory.

srl295:
I'll file a ticket separately for the parent locale issue.

macchiati (approved):

srl295:
CLDR-17906 for further parent locale issues

PR #3972, opened 2024-08-19
CLDR-17535 Remove obsolete tools, fix UncheckedIOExceptions

macchiati:
CLDR-17535

Did some cleanup after https://github.com/unicode-org/cldr/pull/3966

- Removed obsolete tools
- Used ICUUncheckedException (the java equiv. was causing problems with eclipse, and there's no functional difference)

ALLOW_MANY_COMMITS=true


srl295 (approved):

macchiati:
Hmmm. When I ran Maven test locally, it didn't pick up on the missing symbols, even after a Maven clean. So I could only see these problems on the server after committing.

srl295:
> Hmmm. When I ran Maven test locally, it didn't pick up on the missing symbols, even after a Maven clean. So I could only see these problems on the server after committing.

could try mvn install and then test.

srl295 (approved):

PR #3973, opened 2024-08-20
CLDR-17830 Regenerate segmentation data

macchiati:
CLDR-17830

Nothing special, just regenerating the rules from unicodetools to match the soon-to-be-released Unicode.

ALLOW_MANY_COMMITS=true


srl295 (approved):
RSLGTM

PR #3976, opened 2024-08-20
CLDR-11535 Fixes incorrect spacing in currency examples

e-ikwut:
CLDR-11535

ALLOW_MANY_COMMITS=true

Before:
image

After:
image


e-ikwut:
I plan to add examples for local currency back (in addition to EUR/USD).

macchiati:
Looking good. You'll just need to add back the old examples.

NOTE: The background settings may be confusing; they are supposed to hide the stuff that isn't controlled by the pattern or entity. I will comment more on that in slack.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2681,35 +2683,50 @@ private String handleCurrencyFormat(XPathParts parts, String value, boolean show
...
+        DecimalFormat df_eur = icuServiceBuilder.getNumberFormat(pattern_eur, numberSystem);
+        DecimalFormat df_usd = icuServiceBuilder.getNumberFormat(pattern_usd, numberSystem);
+
+        double sampleAmount = 1295.00;
+        example = addExampleResult(formatNumber(df_eur, sampleAmount), example, showContexts);
+        example = addExampleResult(formatNumber(df_eur, -sampleAmount), example, showContexts);
+        example = addExampleResult(formatNumber(df_usd, sampleAmount), example, showContexts);
+        example = addExampleResult(formatNumber(df_usd, -sampleAmount), example, showContexts);
+
+        DecimalFormat df = icuServiceBuilder.getCurrencyFormat(currency, "¤", numberSystem);
```
e-ikwut, line 2718: Passing "¤" is ugly but seems to work since it is not one of the characters that triggers space insertion

srl295:
- can use the squash button to fix the commit messages. But, you will need to merge in from main to deal with conflicts. Write me on slack if you need help with these.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2681,35 +2683,50 @@ private String handleCurrencyFormat(XPathParts parts, String value, boolean show
...
+        example =
+                addExampleResult(
+                        formatNumber(df, sampleAmount).replace("¤", currencySymbol),
+                        example,
+                        showContexts);
+        example =
+                addExampleResult(
+                        formatNumber(df, -sampleAmount).replace("¤", currencySymbol),
+                        example,
+                        showContexts);
```
btangmu, line 2729: Where do things stand with this draft PR? it adds more examples, using df_eur and df_usd...

@srl295 I see you made a more recent commit (91e10790f6c) to this method: it returned String but is now void and ends with `examples.add(example)`

Emiyare also made a more recent commit (00e1fc23f1d) inserting a few lines into this method. It looks like the old changes could be wrapped into the current code fairly easy but not just by rebasing...

PR #3978, opened 2024-08-21
CLDR-15817 Update Example Generator

haytenf:
CLDR-15817

First draft

- [ ] This PR does not complete the ticket.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -442,6 +442,13 @@ public class ExampleGenerator {
         return result;
     }
 
+    // New method that supplies a list
+    private String constructExampleHtml(String xpath, String value, boolean nonTrivial) {
+        List<String> examples = new ArrayList<>();
+        constructExampleHtmlExtended(xpath, value, nonTrivial, examples); // would be void once all handlers are modified
+        return formatExampleList(examples);
+    }
+
     /**
      * Do the main work of getExampleHtml given that the result was not found in the cache.
      *
@@ -451,7 +458,7 @@ public class ExampleGenerator {
      *     markup)
      * @return the example HTML, or null
      */
-    private String constructExampleHtml(String xpath, String value, boolean nonTrivial) {
+    private String constructExampleHtmlExtended(String xpath, String value, boolean nonTrivial, List<String> examples) {
         String result = null;
         boolean showContexts =
                 isRTL || BIDI_MARKS.containsSome(value); // only used for certain example types
@@ -461,7 +468,7 @@ public class ExampleGenerator {
          */
         XPathParts parts = XPathParts.getFrozenInstance(xpath).cloneAsThawed();
         if (parts.contains("dateRangePattern")) { // {0} - {1}
-            result = handleDateRangePattern(value);
+            handleDateRangePattern(value, examples); // modified to add lines to examples, void
         } else if (parts.contains("timeZoneNames")) {
             result = handleTimeZoneName(parts, value);
         } else if (parts.contains("localeDisplayNames")) {
@@ -2850,15 +2857,12 @@ public class ExampleGenerator {
         return null;
     }
 
-    private String handleDateRangePattern(String value) {
-        String result;
+    private void handleDateRangePattern(String value, List<String> examples) {
         SimpleDateFormat dateFormat = icuServiceBuilder.getDateFormat("gregorian", 2, 0);
-        result =
-                format(
-                        value,
-                        setBackground(dateFormat.format(DATE_SAMPLE)),
-                        setBackground(dateFormat.format(DATE_SAMPLE2)));
-        return result;
+        examples.add(format(
+            value,
+            setBackground(dateFormat.format(DATE_SAMPLE)),
+            setBackground(dateFormat.format(DATE_SAMPLE2))));
     }
 
     /**


haytenf:
Incomplete draft that creates a new method that calls createExampleHtml passing a list as an additional argument and updates one of the handlers ( `handleDateRangePattern`) to use this list -- doesn't pass tests since only one handler has been changed to fit new approach. Just to make sure on the right track for updating the handlers. If anyone could take a look when they get a chance @macchiati @srl295

btangmu:
@haytenf to fix the formatting you can run this:
`mvn --file=tools/pom.xml spotless:apply && sh tools/cldr-apps/js/test/pretty.sh`

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -461,7 +468,7 @@ private String constructExampleHtml(String xpath, String value, boolean nonTrivi
          */
         XPathParts parts = XPathParts.getFrozenInstance(xpath).cloneAsThawed();
         if (parts.contains("dateRangePattern")) { // {0} - {1}
-            result = handleDateRangePattern(value);
+            handleDateRangePattern(value, examples); // modified to add lines to examples, void
```
btangmu, line 471: seems like removing "result = " leads to test failures like:
`  TestExampleGenerator {
    Test4528
Error:  (TestExampleGenerator.java:957)  Error: : specifics: expected "〖❬1,23 ❭dollari delle Bermuda〗〖❬0,00 ❭dollari delle Bermuda〗", got null
`

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2850,15 +2857,12 @@ private String handleCurrency(String xpath, XPathParts parts, String value) {
...
-        result =
-                format(
-                        value,
-                        setBackground(dateFormat.format(DATE_SAMPLE)),
-                        setBackground(dateFormat.format(DATE_SAMPLE2)));
-        return result;
+        examples.add(format(
+            value,
+            setBackground(dateFormat.format(DATE_SAMPLE)),
+            setBackground(dateFormat.format(DATE_SAMPLE2))));
```
btangmu, line 2865: instead of making this method void, how about making it call the single-arg version and returning that String value

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
haytenf: Yes, I've only modified one of the handlers so there are two incompatible implementations within the file so for the handlers I haven't modified tests will fail. I'll change the rest of the handlers and re-commit once I finish, and hopefully the tests will pass!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
srl295: Right, it's adding to the 'examples' parameter

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2910,18 +2921,19 @@ private String handleDisplayNames(String xpath, XPathParts parts, String value)
...
                                         locales.get(i),
                                         false,
                                         localeKeyTypePattern,
                                         localePattern,
-                                        localeSeparator));
+                                        localeSeparator)));
             }
-            result = formatExampleList(examples);
+            return;
+            // result = formatExampleList(examples);
```
btangmu, line 2936: Commented-out code

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
btangmu: I would delete the comment, "modified to add lines to examples, void" since that was about work in progress

btangmu (approved):
Looks great! The code looks cleaner with your changes, except a couple of commented-out bits of code that should be removed. Hurray, the tests all pass! If there were lots of time to do more work on this I would suggested adding another unit test or two, but that doesn't seem essential. A couple screenshots of examples would be nice to see in the PR or the ticket to confirm the examples look the same as they did before the refactoring

macchiati (commented):
Looking good so far. I suggest not changing the main dispatch to a switch in a different PR; that will make it easier to test.

Also, for refactorings like this, once you are statisfied with this, you can pull in the original file as ExampleGeneratorOld.java, and write a test case that checks that the results are identical for old and new, iterating across all locales, and iterating across all paths in that locale. That's a small amount of code that gives us a great deal of confidence!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
srl295:
```suggestion
            handleDateRangePattern(value, examples);
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -914,17 +906,17 @@ private void addSubdivisionFlag(
...
-            return null;
+            return;
         }
         String cp = parts.getAttributeValue(-1, "cp");
         if (cp == null || cp.isEmpty()) {
-            return null;
+            return;
         }
-        Set<String> examples = new LinkedHashSet<>();
+        // Set<String> examples = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
srl295:
```suggestion
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
srl295:
```suggestion
```

srl295 (approved):

haytenf:
@btangmu @srl295 @macchiati In the most recent commit, I've added a unit test comparing the old/new example generator, cleaned up the comments, and added before/after screenshots that show the example doesn't change. 

I also wanted to explain a change I made that isn't related to the purpose of this ticket. With the new unit test I added (`TestRefactoring`), I ended up running into a null pointer exception (in both the old example generator and the updated example generator) that I traced back to the `getOtherGender` method that `handleMinimalPairs` calls. The exception was an issue with calling `grammarInfo.get(GrammaticalTarget.nominal, GrammaticalFeature.grammaticalGender, GrammaticalScope.units);` in line 1221 since `grammarInfo` could be `null`, so that's why I made that change.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1215,7 +1214,7 @@ private void handleMinimalPairs(
     }
 
     private String getOtherGender(String gender) {
-        if (gender == null) {
+        if (gender == null || grammarInfo == null) {
```
btangmu, line 1217: good catch

btangmu:
> attempting with the resolved file led to an extremely long run time

that's a puzzle -- does anyone know why?

srl295:
> > attempting with the resolved file led to an extremely long run time
> 
> that's a puzzle -- does anyone know why?

locally or on the github build? if it's the github build could have just been random.

haytenf:
> > > attempting with the resolved file led to an extremely long run time
> > 
> > 
> > that's a puzzle -- does anyone know why?
> 
> locally or on the github build? if it's the github build could have just been random.

I'm still not sure what could be the issue. It was an issue locally--it ended up running for 12+ hours at which point it ran into a runtime exception from running out of heap space.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2466,7 +2463,8 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
                     cldrFile.getWinningPath(xpath.replaceAll("dateTimeFormat", "timeFormat"));
             int tfLengthOffset = timeFormatXPathForPrefix.indexOf("timeFormatLength");
             if (tfLengthOffset < 0) {
-                return "";
+                examples.add("");
+                return;
```
srl295, line 2467: Does it make sense to add a `""` example here? or just to exit without adding anything?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGeneratorOld.java
srl295, line 1: If this is for comparison, this file should be removed from the PR.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -2076,4 +2079,27 @@ private String getResult(String starredPath, String attr) {
     public String sampleAttrAndValue(PathStarrer ps, final String separator, String value) {
         return ps.getAttributesString(separator) + "➔«" + value + "»";
     }
+
+    public void TestRefactoring() {
```
srl295, line 2083: OK, i see it's for the temporary test.

srl295: Ouch. All locales X all resolved paths will be way too much. 

Actually what you probably want to do here is to limit it by coverage and pathheader. I'll get some sample code for it.

srl295 (approved):
looking good and I think it's an improvement

srl295 (changes requested):
i think ExampleGeneratorOld.java shouldn't be in the PR

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
srl295: this code (will need some setup) should show whether a path would even be shown to users or not.  If not, it should be skipped from the test:

```java
// adapted from STFactory.isValidSurveyToolVote

        PathHeader.Factory phf = PathHeader.getFactory(null); // do this once!


        private boolean isValidSurveyToolVote(String xpath) {
            PathHeader ph = phf.fromPath(xpath); // will need a 
 PathHeaderFactory
            if (ph == null) return false;
            if (ph.getSurveyToolStatus() == PathHeader.SurveyToolStatus.DEPRECATED) return false;
            if (ph.getSurveyToolStatus() == PathHeader.SurveyToolStatus.HIDE
                    || ph.getSurveyToolStatus() == PathHeader.SurveyToolStatus.READ_ONLY) {
                return false;
            }

            if (supplementalDataInfo.getCoverageValue(xpath, locale.getBaseName())
                    > org.unicode.cldr.util.Level.COMPREHENSIVE.getLevel()) {
                return false;
            }
            return true;
        }
```

haytenf: Got it. I'll revise to test over only the paths that'd be shown to users. Even over that reduced set of paths, and considering the extra check needed to determine if a path shown to users, I'm guessing the resolved files may still be too large to test over.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
haytenf: Good catch! Yes, i think it makes sense to just `return` in this case.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -2076,4 +2079,51 @@ private String getResult(String starredPath, String attr) {
...
+        if (SDI.getCoverageValue(xpath, locale.getBaseName())
+                > org.unicode.cldr.util.Level.COMPREHENSIVE.getLevel()) {
+            return false;
+        }
+        return true;
+    }
+
+    public void TestRefactoring() {
+        for (ULocale locale : ULocale.getAvailableLocales()) {
+            String loc = locale.toString();
```
srl295, line 2107:
```suggestion
        for (final String loc : info.getCldrFactory().getAvailable()) {
```

Sorry, missed this before - slight change needed to make it run against all locales instead of just those ULocale thinks is available.

srl295 (commented):
Anyway, I think if you remove teh ExampleGeneratorOld and TestRefactoring this is ready to go - test passes!!

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
srl295: but, not important since you're going to be removing this test

srl295 (approved):

srl295:
@haytenf Thanks! 🎉

haytenf:
> @haytenf Thanks! 🎉

Thank you!

PR #3980, opened 2024-08-22
CLDR-17535 BRS Build language containment

macchiati:
CLDR-17535

The output is impossible to diff, so the DiffLanguageGroups.java is needed to make sense of the differences.

I modified DiffLanguageGroups to make it easier to compare the changes. The current results are in https://docs.google.com/spreadsheets/d/1vx3qzX_gO_FJmNl9NgcC-_z2D4Rxp02WgylnEINQdXc/edit?gid=653556033 (the Containment sheet).

It is highlighting the changes in the TC locales and Other Basic+ locales, to focus the review on them, although all of the differences are listed. I addressed the most concerning ones, by adding overrides (also cleaned up the code around that.

The DiffLanguageGroups is also automatically run at the end of the generation.

ALLOW_MANY_COMMITS=true


macchiati:
Tests pass now, so I'd appreciate a review.

srl295:
a local run of `git diff --word-diff` gives a good approximation of the xml changes

tools/cldr-code/src/main/java/org/unicode/cldr/util/DiffLanguageGroups.java
```diff
@@ -1,14 +1,18 @@
 package org.unicode.cldr.util;
```
srl295, line 1:
```suggestion
package org.unicode.cldr.tool;
```

If this is really a tool, it should go under the tool package.

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateLanguageContainment.java
```diff
@@ -603,67 +711,67 @@ private static Set<Set<String>> getAncestors(String leaf, Set<String> skipping)
         }
         return itemsFixed;
         // TODO: delete this commented-out code?
```
srl295, line 713: delete it instead of fixing its indent?

srl295 (approved):
still looking a the spreadsheet but I'm going to say LGTM for now

srl295:
Chart is helpful. This is a little concerning:

> Missing	Mapuche ⁅arn⁆, Betawi ⁅bew⁆, Choctaw ⁅cho⁆, Chickasaw ⁅cic⁆, Muscogee ⁅mus⁆, Papiamento ⁅pap⁆

any reason why these should be missing?

tools/cldr-code/src/main/java/org/unicode/cldr/util/DiffLanguageGroups.java
macchiati: Agreed. I just didn't want to do that in a single PR.

macchiati:
> a local run of `git diff --word-diff` gives a good approximation of the xml changes

But that doesn't let you know what all the codes actually mean, and when they switch from one parent to another (or are just removed).

macchiati:
> Chart is helpful. This is a little concerning:
> 
> > Missing	Mapuche ⁅arn⁆, Betawi ⁅bew⁆, Choctaw ⁅cho⁆, Chickasaw ⁅cic⁆, Muscogee ⁅mus⁆, Papiamento ⁅pap⁆
> 
> any reason why these should be missing?

The diffs are in the spreadsheet. https://docs.google.com/spreadsheets/d/1vx3qzX_gO_FJmNl9NgcC-_z2D4Rxp02WgylnEINQdXc/edit?gid=653556033#gid=653556033

I didn't focus on the OC cases, because they haven't reached basic; once they do, then this is worth following up on; but by that time, Wikidata might have data.

Version | Key | Locale | Action | Information
-- | -- | -- | -- | --
v45 | OC | Mapuche ⁅arn⁆ | Removed | Ω | 
v45 | OC | Betawi ⁅bew⁆ | Removed | Austronesian ⁅map⁆ ➡︎ Ω | 
v45 | OC | Choctaw ⁅cho⁆ | Removed | Ω  | 
v45 | OC | Chickasaw ⁅cic⁆ | Removed | Ω  | 
v45 | OC | Muscogee ⁅mus⁆ | Removed | Ω  | 
v45 | OC | Papiamento ⁅pap⁆ | Removed | Creoles and pidgins, Portuguese-based ⁅cpp⁆ ➡︎ Romance ⁅roa⁆ ➡︎ Italic  | languages ⁅itc⁆ ➡︎ Indo-European ⁅ine⁆ ➡︎ Ω | 

FYI: the removal of Ω doesn't mean much; either way, there isn't a known parent.

srl295:
Ok I misread I thought it has a parent before.

PR #3981, opened 2024-08-23
CLDR-17572 Move LR marker div below star

btangmu:
-Move the LR marker div below the i-star icon, simply by moving the call to checkLRmarker a few lines lower

-Also remove a comment in the same file about appendExample which is no longer true, and remove the export for appendExample which is no longer needed

CLDR-17572

ALLOW_MANY_COMMITS=true


btangmu:
Screenshot with this change, showing the star on the line above the added LR marker div:
image

macchiati (approved):

PR #3988, opened 2024-08-27
CLDR-17830 BRS Alpha release page 2

macchiati:
Small updates from TC meeting.

CLDR-17830

ALLOW_MANY_COMMITS=true


AEApple (dismissed):

macchiati:
Hi, I'd like to get these changes merged in, and then put it out to the TC for further refinements. Could you take a look?

srl295 (approved):
LGTM (will fixup syntax errs later)

AEApple (approved):
LGTM

I think we should still do another PR later and add the locale codes for all the locales which have increased coverage. Happy to put that in after this one merges if you'd like?

srl295:
Site issues cleared up in #4020

macchiati:
> LGTM
> 
> I think we should still do another PR later and add the locale codes for all the locales which have increased coverage. Happy to put that in after this one merges if you'd like?

Yes, that's be great!

PR #3989, opened 2024-08-27
CLDR-17803 docs/site: publish to CloudFlare, config updates

srl295:
CLDR-17803

## **Note** This is a major _deployment_ update - there's still a lot of work todo on the content! 

 - docs/site/* deploys to cloudflare
 - PR comment tells you where to go try it [example](https://github.com/unicode-org/cldr/pull/3989#issuecomment-2313705995)
 - Docker available for local test and build
 
 Split out docs/site* from other stuff in docs/
 - top level config for google sites (such as it is)
 - docs/site just for the site.

ALLOW_MANY_COMMITS=true


srl295:
@lianghai FYI

DavidLRowe:
From the https://6c2617a4.cldr.pages.dev/ page, if I click on the "Adding new locales" link, I end up at 
`https://github.com/unicode-org/cldr/blob/main/docs/requesting_changes.md#adding-new-locales`
which I was not expecting. Is this intended?

srl295:
> From the https://6c2617a4.cldr.pages.dev/ page, if I click on the "Adding new locales" link, I end up at `https://github.com/unicode-org/cldr/blob/main/docs/requesting_changes.md#adding-new-locales` which I was not expecting. Is this intended?

The content has all kinds of style and navigation problems. The point is that it's _deployed somewhere_.  Then we can improve the content.

macchiati:
It's hard to tell what is going on, because I only see one page. Are there supposed to be multiple pages hosted?

srl295:
> It's hard to tell what is going on, because I only see one page. Are there supposed to be multiple pages hosted?

You have to navigate to them directly - so, go to https://69a9a0e3.cldr.pages.dev/downloads/cldr-46

As I said, this is around the deployment, not the content.  It's no different from teh issues faced by https://unicode-org.github.io/cldr/site/

Please consider approving this incremental improvement.

srl295:
Could I get a review to merge this?

btangmu (approved):

PR #3990, opened 2024-08-28
CLDR-17830 BRS update subdivision translations

macchiati:
CLDR-17830

Adds names from wikidata. Note that some items disappear, because the codes were deprecated.

ALLOW_MANY_COMMITS=true


AEApple:
What leads some of these items to be provisional status, while the newer items have status contributed?

macchiati:
I'll have to look into that.

macchiati:
The documentation is basically to run the tool. It is run after we have
updated the subdivision codes (a separate step). Countries change their
codes all the time, and we don't really try to track that beyond keeping
the old codes in a deprecated list.

macchiati:
Updated tooling to change the new values to provisional, for consistency.

macchiati:
Ping, would like to merge this by EOD if possible

AEApple:
any idea why so many island territories are being removed? Is that expected?

btangmu:
> The documentation is basically to run the tool

Meaning WikiSubdivisionLanguages.java, I assume -- that wasn't obvious in the earlier version of this PR. I'll approve with a rubber stamp

btangmu (approved):

macchiati:
Over the years, I've looked into some of the big revamps for various regions. Sometimes it is because of big revamps in governmental structures, sometimes just to make the codes match the actual governmental structures more clearly.

macchiati:
Right, WikiSubdivisionLanguages.java. Sorry for being to terse.

PR #3993, opened 2024-08-28
CLDR-17382 languagematch Ukrainian should not fall back to Russian

stenshamn:
CLDR-17382

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/languageInfo.xml
+++ b/common/supplemental/languageInfo.xml
@@ -145,8 +145,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<languageMatch desired="tum"	supported="en"	distance="30"	oneway="true"/>	<!-- Tumbuka: tum ⇒ en -->
 			<languageMatch desired="ug"	supported="zh"	distance="20"	oneway="true"/>	<!-- Uighur: ug ⇒ zh -->
 			
-			<languageMatch desired="uk"	supported="en"	distance="30"	oneway="true"/>	<!-- Ukrainian ⇒ English -->
-			<!-- CLDR-17382: languageMatch: Ukrainian should not fall back to Russian -->
+			<!-- CLDR-17382: Ukrainian should not fall back to Russian -->
+			<!-- languageMatch desired="uk" supported="ru" distance="20" oneway="true"/--> <!-- Ukrainian ⇒ Russian -->
 
 			<languageMatch desired="ur"	supported="en"	distance="30"	oneway="true"/>	<!-- Urdu: ur ⇒ en -->
 			<languageMatch desired="uz"	supported="ru"	distance="30"	oneway="true"/>	<!-- Uzbek: uz ⇒ ru -->
--- a/common/supplemental/languageInfo.xml
+++ b/common/supplemental/languageInfo.xml
@@ -144,7 +144,10 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<languageMatch desired="tt"	supported="ru"	distance="30"	oneway="true"/>	<!-- Tatar: tt ⇒ ru -->
 			<languageMatch desired="tum"	supported="en"	distance="30"	oneway="true"/>	<!-- Tumbuka: tum ⇒ en -->
 			<languageMatch desired="ug"	supported="zh"	distance="20"	oneway="true"/>	<!-- Uighur: ug ⇒ zh -->
-			<languageMatch desired="uk"	supported="ru"	distance="20"	oneway="true"/>	<!-- Ukrainian ⇒ Russian -->
+			
+			<languageMatch desired="uk"	supported="en"	distance="30"	oneway="true"/>	<!-- Ukrainian ⇒ English -->
+			<!-- CLDR-17382: languageMatch: Ukrainian should not fall back to Russian -->
+
 			<languageMatch desired="ur"	supported="en"	distance="30"	oneway="true"/>	<!-- Urdu: ur ⇒ en -->
 			<languageMatch desired="uz"	supported="ru"	distance="30"	oneway="true"/>	<!-- Uzbek: uz ⇒ ru -->
 			<languageMatch desired="wo"	supported="fr"	distance="30"	oneway="true"/>	<!-- Wolof: wo ⇒ fr -->


AEApple:
I don't think we're changing it to English, I think it should match what was done for Macedonian.

stenshamn:
> I didn't think we were changing the match to English, I think it should match what was done for Macedonian. I thought the goal was to remove the language match and not to explicitly cause it to match to English.

Before making the change, I looked at some of the other language matches and previous changes, and it seemed that a change, not deletion, was the right thing to do in this case. 
For Estonian they commented it out because (according to ticket/PR comments) they couldn't seem to decide between Russian and English and decided not to have a fallback at all. Many other languages have explicit fallback to English. I didn't want to leave the fallback to be random or unpredictable so I went with explicit.

common/supplemental/languageInfo.xml
```diff
@@ -144,7 +144,10 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<languageMatch desired="tt"	supported="ru"	distance="30"	oneway="true"/>	<!-- Tatar: tt ⇒ ru -->
 			<languageMatch desired="tum"	supported="en"	distance="30"	oneway="true"/>	<!-- Tumbuka: tum ⇒ en -->
 			<languageMatch desired="ug"	supported="zh"	distance="20"	oneway="true"/>	<!-- Uighur: ug ⇒ zh -->
-			<languageMatch desired="uk"	supported="ru"	distance="20"	oneway="true"/>	<!-- Ukrainian ⇒ Russian -->
+			
+			<languageMatch desired="uk"	supported="en"	distance="30"	oneway="true"/>	<!-- Ukrainian ⇒ English -->
+			<!-- CLDR-17382: languageMatch: Ukrainian should not fall back to Russian -->
```
AEApple, line 149: In that case then can we have the comment say that Ukrainian should match with English if we are adding an explicit match?

macchiati: I think it is ok to have the explicit fallback to en.

macchiati (approved):

common/supplemental/languageInfo.xml
markusicu: I disagree very much. That's not how this is designed, nor advertised.

markusicu (dismissed):

common/supplemental/languageInfo.xml
markusicu, line 149:
```suggestion
			<!-- CLDR-17382: Ukrainian should not fall back to Russian -->
			<!-- languageMatch desired="uk" supported="ru" distance="20" oneway="true"/--> <!-- Ukrainian ⇒ Russian -->
```

stenshamn: @markusicu , do you mean that you basically don't want us to do this change?

stenshamn: Sorry, I missed "comment out" piece at first so I was really confused.
But can you elaborate on why we wouldn't explicitly falling back on English? We do it for several other languages.

macchiati: @markusicu 
We have an explicit fallback to 'en' in many, many cases. For example:
```
38: <languageMatch desired="ach" supported="en" distance="30" oneway="true"/> <!-- Acoli (Southern Luo dialect in Uganda): ach ⇒ en --> 
40: <languageMatch desired="ak" supported="en" distance="30" oneway="true"/> <!-- Akan: ak ⇒ en --> 
41: <languageMatch desired="am" supported="en" distance="30" oneway="true"/> <!-- Amharic ⇒ English --> 
46: <languageMatch desired="bem" supported="en" distance="30" oneway="true"/> <!-- Bemba (Zambia): bem ⇒ en --> 
48: <languageMatch desired="bn" supported="en" distance="30" oneway="true"/> <!-- Bangla: bn ⇒ en --> 
53: <languageMatch desired="chr" supported="en" distance="20" oneway="true"/> <!-- Cherokee: chr ⇒ en --> 
58: <languageMatch desired="cy" supported="en" distance="20" oneway="true"/> <!-- Welsh: cy ⇒ en --> 
59: <languageMatch desired="ee" supported="en" distance="30" oneway="true"/> <!-- Ewe: ee ⇒ en --> 
60: <languageMatch desired="eo" supported="en" distance="30" oneway="true"/> <!-- Esperanto: eo ⇒ en --> 
68: <languageMatch desired="ga" supported="en" distance="20" oneway="true"/> <!-- Irish: ga ⇒ en --> 
69: <languageMatch desired="gaa" supported="en" distance="30" oneway="true"/> <!-- Ga: gaa ⇒ en --> 
70: <languageMatch desired="gd" supported="en" distance="20" oneway="true"/> <!-- Scottish Gaelic: gd ⇒ en --> 
74: <languageMatch desired="ha" supported="en" distance="30" oneway="true"/> <!-- Hausa: ha ⇒ en --> 
75: <languageMatch desired="haw" supported="en" distance="20" oneway="true"/> <!-- Hawaiian: haw ⇒ en --> 
78: <languageMatch desired="ia" supported="en" distance="30" oneway="true"/> <!-- Interlingua: ia ⇒ en --> 
79: <languageMatch desired="ig" supported="en" distance="30" oneway="true"/> <!-- Igbo: ig ⇒ en --> 
80: <languageMatch desired="is" supported="en" distance="20" oneway="true"/> <!-- Icelandic: is ⇒ en --> 
82: <languageMatch desired="ka" supported="en" distance="30" oneway="true"/> <!-- Georgian: ka ⇒ en --> 
85: <languageMatch desired="km" supported="en" distance="30" oneway="true"/> <!-- Khmer: km ⇒ en --> 
86: <languageMatch desired="kn" supported="en" distance="30" oneway="true"/> <!-- Kannada: kn ⇒ en --> 
87: <languageMatch desired="kri" supported="en" distance="30" oneway="true"/> <!-- Krio: kri ⇒ en --> 
91: <languageMatch desired="lg" supported="en" distance="30" oneway="true"/> <!-- Luganda: lg ⇒ en --> 
93: <languageMatch desired="lo" supported="en" distance="30" oneway="true"/> <!-- Lao: lo ⇒ en --> 
94: <languageMatch desired="loz" supported="en" distance="30" oneway="true"/> <!-- Lozi: loz ⇒ en --> 
97: <languageMatch desired="mfe" supported="en" distance="30" oneway="true"/> <!-- Morisyen: mfe ⇒ en --> 
99: <languageMatch desired="mi" supported="en" distance="20" oneway="true"/> <!-- Māori: mi ⇒ en --> 
104: <languageMatch desired="ml" supported="en" distance="30" oneway="true"/> <!-- Malayalam: ml ⇒ en --> 
108: <languageMatch desired="mt" supported="en" distance="30" oneway="true"/> <!-- Maltese: mt ⇒ en --> 
109: <languageMatch desired="my" supported="en" distance="30" oneway="true"/> <!-- Myanmar: my ⇒ en --> 
110: <languageMatch desired="ne" supported="en" distance="30" oneway="true"/> <!-- Nepali: ne ⇒ en --> 
113: <languageMatch desired="nso" supported="en" distance="30" oneway="true"/> <!-- Northern Sotho: nso ⇒ en --> 
114: <languageMatch desired="ny" supported="en" distance="30" oneway="true"/> <!-- Nyanja: ny ⇒ en --> 
115: <languageMatch desired="nyn" supported="en" distance="30" oneway="true"/> <!-- Nyankole: nyn ⇒ en --> 
117: <languageMatch desired="om" supported="en" distance="30" oneway="true"/> <!-- Oromo: om ⇒ en --> 
118: <languageMatch desired="or" supported="en" distance="30" oneway="true"/> <!-- Odia: or ⇒ en --> 
119: <languageMatch desired="pa" supported="en" distance="30" oneway="true"/> <!-- Punjabi: pa ⇒ en --> 
120: <languageMatch desired="pcm" supported="en" distance="20" oneway="true"/> <!-- Nigerian Pidgin: pcm ⇒ en --> 
121: <languageMatch desired="ps" supported="en" distance="30" oneway="true"/> <!-- Pashto: ps ⇒ en --> 
124: <languageMatch desired="rn" supported="en" distance="30" oneway="true"/> <!-- Rundi: rn ⇒ en --> 
128: <languageMatch desired="sd" supported="en" distance="30" oneway="true"/> <!-- Sindhi: sd ⇒ en --> 
129: <languageMatch desired="si" supported="en" distance="30" oneway="true"/> <!-- Sinhalese: si ⇒ en --> 
130: <languageMatch desired="sn" supported="en" distance="30" oneway="true"/> <!-- Shona: sn ⇒ en --> 
131: <languageMatch desired="so" supported="en" distance="30" oneway="true"/> <!-- Somali: so ⇒ en --> 
132: <languageMatch desired="sq" supported="en" distance="30" oneway="true"/> <!-- Albanian: sq ⇒ en --> 
133: <languageMatch desired="st" supported="en" distance="30" oneway="true"/> <!-- Southern Sotho: st ⇒ en --> 
135: <languageMatch desired="sw" supported="en" distance="30" oneway="true"/> <!-- Swahili: sw ⇒ en --> 
136: <languageMatch desired="ta" supported="en" distance="30" oneway="true"/> <!-- Tamil: ta ⇒ en --> 
137: <languageMatch desired="te" supported="en" distance="30" oneway="true"/> <!-- Telugu: te ⇒ en --> 
139: <languageMatch desired="ti" supported="en" distance="30" oneway="true"/> <!-- Tigrinya: ti ⇒ en --> 
141: <languageMatch desired="tlh" supported="en" distance="30" oneway="true"/> <!-- Klingon: tlh ⇒ en --> 
142: <languageMatch desired="tn" supported="en" distance="30" oneway="true"/> <!-- Tswana: tn ⇒ en --> 
143: <languageMatch desired="to" supported="en" distance="30" oneway="true"/> <!-- Tonga: to ⇒ en --> 
145: <languageMatch desired="tum" supported="en" distance="30" oneway="true"/> <!-- Tumbuka: tum ⇒ en --> 
148: <languageMatch desired="ur" supported="en" distance="30" oneway="true"/> <!-- Urdu: ur ⇒ en --> 
151: <languageMatch desired="xh" supported="en" distance="30" oneway="true"/> <!-- Xhosa: xh ⇒ en --> 
152: <languageMatch desired="yi" supported="en" distance="30" oneway="true"/> <!-- Yiddish: yi ⇒ en --> 
153: <languageMatch desired="yo" supported="en" distance="30" oneway="true"/> <!-- Yoruba: yo ⇒ en --> 
155: <languageMatch desired="zu" supported="en" distance="30" oneway="true"/> <!-- Zulu: zu ⇒ en --> 
```
In view of the expressed preference in Ukraine for English over Russian, I don't see why using supported="en" is bothering you, or inconsistent with current practice. Could you explain more?

macchiati:
I think we can wait til tomorrow to see what Markus says, rather than Resolve the Conversation now.

markusicu:
The information we got is "Ukrainian-language users don't want to be matched with Russian-language contents". The corresponding data change is to remove (comment out) the languageMatch for this pair.

This is _language matcher_ data which feeds into an implementation (e.g., ICU LocaleMatcher) which a caller sets up with a list of supported languages plus an _optional_ default language. When there is no match, then the matcher returns the default language, when it's set, otherwise with a "no match" result.

The default language is chosen by the caller. It need not be English. And not setting one at all is a valid, important choice. Some callers have special strategies for what to do next.

When we overcorrect and force a "fallback" to English, then we short-circuit the functioning of the algorithm and defeat the caller's intent.

We should handle this like the other geopolitical cases in the past, like Macedonian.

----
I haven't reviewed other "fallbacks to English" in detail. I assumed that they were generally matches based on some information, like populations are actually somewhat likely to understand English because it's one of the local government and entertainment languages, or remaining influence in colonies, etc. I would expect similar data for one-way matches to French (e.g., Breton --> French), Portuguese, Chinese, and Arabic.

Some of these might not make sense. At a glance, I see a one-way match from Esperanto to English; that looks bogus.

A review of existing data deserves a separate ticket.

macchiati:
**Upshot**

I don't think it matters much whether we include an explicit fallback or not, and we don't for many languages. So to be conservative, we could omit the fallback mapping to English, then revisit this in the next cycle.

**Background**

I looked at this a bit. If someone uses the default settings with the proposed change, here is what happens.

- the default direction includes ONE_WAY
- the default max threshhold distance is 50
- the default locale is the first supported language (the caller really has to supply the supported locales)

If the user's desired languages are <Ukrainian, French> and the default language is set to German, then the priority order among the app's supported languages would be:

<Ukrainian, French, English, German>

So on systems that allow for secondary desired languages, such as iOS, Android, and MacOS, it is easy for users get the the desired result, if their favored fallback language is French (or Russian) rather than German. (Of course, that doesn't necessarily mean that users take advantage of this ability.)

The difference would be that English would come before what the system has set as a default. 

Now, if the system doesn't set a specific default, or doesn't order the supported languages to put a reasonable default as the first one, the results would be a bit random. On the other hand, that is the case for many of our current languages, since we don't always have a fallback for all major languages. I suspect that a very large number of users of LocaleMatcher will use a system default of English; although that might be different in Ukraine.

I think that's why we made an effort to have fallback locales for most locales that are not "top tier" (ie, those supported by most applications), so that the user would get some reasonable result for systems that don't set the default based on the likelyhood that users in that country would understand the language.

stenshamn:
@macchiati , I'm not sure if I can discern a tie-breaking call in your reply. Should we default to English or leave it commented out like Markus suggests?

markusicu:
> I don't think it matters much whether we include an explicit fallback or not, and we don't for many languages. So to be conservative, we could omit the fallback mapping to English, then revisit this in the next cycle.

Yes please let's just comment out the offensive mapping as requested in the ticket.

> I think that's why we made an effort to have fallback locales for most locales that are not "top tier" (ie, those supported by most applications), so that the user would get some reasonable result for systems that don't set the default based on the likelyhood that users in that country would understand the language.

That last part is important. *If it's reasonable to assume* that people who understand language x might also understand English or French or... then we should have a medium-high-distance one-way match for that. If not, then we shouldn't have a languageMatch entry. I would argue that Esperanto-->English is the latter. (And I am not asking for changing that in this PR nor under this ticket!)

macchiati:
> @macchiati , I'm not sure if I can discern a tie-breaking call in your reply. Should we default to English or leave it commented out like Markus suggests?

Let's comment it out for now, and revisit next cycle.

macchiati:
>  If it's reasonable to assume that people who understand language x might also understand English or French or... then we should have a medium-high-distance one-way match for that.

I don't see it that way. The goal for the fallbacks should be: based on the best information we have, in the absence of any other information, what are people most likely to understand if the language in question is not available. [Caveat geopolitical]

So when when people can't supply secondary languages, is it better to:
- pick that fallback
    - note that when that fallback is not supported, it will fall further back to the system default.
- or go with a system default.

AEApple (approved):
LGTM

macchiati (approved):

macchiati:
I'll go ahead and merge

PR #3994, opened 2024-08-28
CLDR-15697 Update AE first day of the week to mon

AEApple:
CLDR-15697

Update AE first day of the week to mon

ALLOW_MANY_COMMITS=true


AydaTolaymat (approved):
looks good.

PR #3996, opened 2024-08-28
CLDR-17803 docs/site: fix merge to main

srl295:
- only update PR if there's a PR - correct an error on merge to main previously.

CLDR-17803

ALLOW_MANY_COMMITS=true


macchiati:
BTW, is there a way to skip the tests when the only change is to markdown files? That would speed up turnaround.

srl295:
> BTW, is there a way to skip the tests when the only change is to markdown files? That would speed up turnaround.

i think i have another issue for that, but yes.

PR #3997, opened 2024-08-28
CLDR-17850 Show when patterns allow different widths or not

e-ikwut:
CLDR-17850

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -169,6 +171,7 @@ public class ExampleGenerator {
     private static final Date DATE_SAMPLE2;
     private static final Date DATE_SAMPLE3;
     private static final Date DATE_SAMPLE4;
+    private static final Date DATE_SAMPLE5;
```
e-ikwut, line 174: This is redundant with CLDR-15698 #3895, which is ready for merge

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2411,6 +2418,36 @@ private String handleTimeZoneName(XPathParts parts, String value) {
...
+                            break;
+                        case "HH":
+                            altPatternBuilder.append("H");
+                            break;
+                        default:
+                            altPatternBuilder.append(field.toString());
+                            break;
+                    }
+                } else {
+                    altPatternBuilder.append(item);
```
e-ikwut, line 2445: If I understand correctly, if item is not a VariableField, it's a literal string. The ticket description mentions using fp.toString(i, i + 1). Is this acceptable in its place?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2608,7 +2645,39 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
...
+                    String altId = generateAltPattern(id);
+                    if (!altId.equals(id)) {
+                        String altXpath = "//ldml/dates/calendars/calendar[@type=\""
+                            + calendar
+                            + "\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\""
+                            + altId
+                            + "\"]";
+                        String forcedAltValue = cldrFile.getWinningValue(altXpath);
+                        String unforcedAltValue = generateAltPattern(value);
+                        String forcePrefix = forcedAltValue != null ? "" : EXAMPLE_OF_INCORRECT;
```
e-ikwut, line 2659: Originally, I had the condition as `forcedAltValue.equals(unforcedAltValue)`
That didn't work (forcedAltValue was sometimes `null`), so I tried this instead. Does this make sense?

e-ikwut: This condition is supposed to determine whether or not a value is forced

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2608,7 +2645,39 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
...
+                        defaultNumberingSystem =
+                                cldrFile.getWinningValue("//ldml/numbers/defaultNumberingSystem");
+                        timeSeparator =
+                                cldrFile.getWinningValue(
+                                        "//ldml/numbers/symbols[@numberSystem='"
+                                                + defaultNumberingSystem
+                                                + "']/timeSeparator");
+                        dfs = sdf.getDateFormatSymbols();
+                        dfs.setTimeSeparatorString(timeSeparator);
+                        sdf.setDateFormatSymbols(dfs);
```
e-ikwut, line 2674: I copy-pasted this chunk from above. The only change is `icuServiceBuilder.getDateFormat(calendar, unforcedAltValue, numbersOverride);` instead of `icuServiceBuilder.getDateFormat(calendar, value, numbersOverride);` Is there a way to just change the pattern on a SimpleDateFormat instead of creating a new one?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2608,7 +2645,39 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
...
+
+                        example = addExampleResult(forcePrefix + sdf.format(DATE_SAMPLE) + ", " + sdf.format(DATE_SAMPLE5), example, showContexts);
+                    }    
+                    return example;
```
e-ikwut, line 2678: Should I account for the details checked below, e.g., `if (value.contains("MMM") || value.contains("LLL"))`?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2608,7 +2645,39 @@ private String handleDateFormatItem(String xpath, String value, boolean showCont
                 DateFormatSymbols dfs = sdf.getDateFormatSymbols();
                 dfs.setTimeSeparatorString(timeSeparator);
                 sdf.setDateFormatSymbols(dfs);
-                if (id == null || id.indexOf('B') < 0) {
+                if (parts.contains("availableFormats")) {
+                    String example = addExampleResult(sdf.format(DATE_SAMPLE) + ", " + sdf.format(DATE_SAMPLE5), "", showContexts);
```
e-ikwut, line 2649: I should only do this when `sdf.format(DATE_SAMPLE) != sdf.format(DATE_SAMPLE5)`

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2411,6 +2418,36 @@ private String handleTimeZoneName(XPathParts parts, String value) {
         return result;
     }
 
+    private String generateAltPattern(String dateTimeValue) {
```
e-ikwut, line 2421: Rereading the ticket, I see "the caller can specify the width of _various fields_." Right now, I'm only accounting for different widths of hours (I thought, at first, that only hours were relevant). Should I be checking every variable field? In that case, couldn't the number of possible examples be very high (e.g. if all of h,m,s can have their width changed, would 8 examples be desired)?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
e-ikwut: If it does make sense, I realize I've flipped the condition (forcing would be when the value is not null, not when the value is null)

However, I'm very dubious of this, since it's not what we discussed. As written, it seems like forcedAltValue is always null.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: For time, we only allow the first to be padded or not, eg:

9:08:07 or 09:08:07, but not 9:8:7

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: You want to leave the literals completely alone. So the above looks good.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: You could pull that chunk of code out into a function, and call with slightly different parameters. Better that code copying (which is harder to maintain).

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Add comment that this function could be generalizes for fields other than h/H. (If you had more time, you could do it, but the hourglass has run out....)

PR #3998, opened 2024-08-29
CLDR-17897 Fix unstable scripts when running GenerateLikelySubtags and ConvertLanguageData

conradarcturus:
CLDR-17897

While we are improving the population data and likely subtags we are generating side-effects from partial data. This adds new scripts so we can avoid these side-effects in future changes. Ultimately we will want to remove how many overrides are here but it's good to fix this.

See the data updated in this diagram:
Screenshot 2024-08-29 at 15 29 41

- [ ] This PR completes the ticket. -- I'm submitting this request first to separate the changes
- [x] This PR stabilizes the data so its easier to follow up and fix other overrides.

Run this command to regenerate data: `mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags`

ALLOW_MANY_COMMITS=true


common/supplemental/likelySubtags.xml
```diff
@@ -1036,6 +1039,7 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="und_Ahom" to="aho_Ahom_IN"/>		<!--?‧Ahom‧?	➡ Ahom‧Ahom‧India-->
 		<likelySubtag from="und_Arab" to="ar_Arab_EG"/>		<!--?‧Arabic‧?	➡ Arabic‧Arabic‧Egypt-->
 		<likelySubtag from="und_Arab_AF" to="fa_Arab_AF"/>		<!--?‧Arabic‧Afghanistan	➡ Persian‧Arabic‧Afghanistan-->
+		<likelySubtag from="und_Arab_AZ" to="tly_Arab_AZ"/>		<!--?‧Arabic‧Azerbaijan	➡ Talysh‧Arabic‧Azerbaijan-->
```
roozbehp, line 1042: If you have something unknown in Arabic script in Azerbaijan, it's probably not Talysh (which has a pretty small community compared to Azerbaijani, where they write in Latin). It's very probably Azerbaijani in the old orthography.

roozbehp (changes requested):

common/supplemental/supplementalData.xml
```diff
@@ -1890,7 +1890,7 @@ XXX Code for transations where no currency is involved
 		<language type="lv" scripts="Latn" territories="LV"/>
 		<language type="lwl" scripts="Thai"/>
 		<language type="lzh" scripts="Hans" alt="secondary"/>
-		<language type="lzz" scripts="Latn Geor"/>
+		<language type="lzz" scripts="Geor Latn"/>
```
srl295, line 1893: Is the ordering significant?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv
```diff
@@ -633,16 +635,17 @@ pdt	Plautdietsch	primary	Latn	Latin
...
 pi	Pali	secondary	Thai	Thai
 pis	Pijin	primary	Latn	Latin
 pko	Pökoot	primary	Latn	Latin
 pl	Polish	primary	Latn	Latin
 pms	Piedmontese	primary	Latn	Latin
-pnt	Pontic	primary	Cyrl	Cyrillic
 pnt	Pontic	primary	Grek	Greek
-pnt	Pontic	primary	Latn	Latin
+pnt	Pontic	secondary	Cyrl	Cyrillic
+pnt	Pontic	secondary	Latn	Latin
```
srl295, line 648: What's the basis of making these secondary?

common/supplemental/likelySubtags.xml
```diff
@@ -703,10 +706,10 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="tiv" to="tiv_Latn_NG"/>		<!--Tiv‧?‧?	➡ Tiv‧Latin‧Nigeria-->
 		<likelySubtag from="tk" to="tk_Latn_TM"/>		<!--Turkmen‧?‧?	➡ Turkmen‧Latin‧Turkmenistan-->
 		<likelySubtag from="tkl" to="tkl_Latn_TK"/>		<!--Tokelau‧?‧?	➡ Tokelau‧Latin‧Tokelau-->
-		<likelySubtag from="tkr" to="tkr_Latn_AZ"/>		<!--Tsakhur‧?‧?	➡ Tsakhur‧Latin‧Azerbaijan-->
+		<likelySubtag from="tkr" to="tkr_Cyrl_AZ"/>		<!--Tsakhur‧?‧?	➡ Tsakhur‧Cyrillic‧Azerbaijan-->
```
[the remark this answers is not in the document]
conradarcturus: Thanks for the close examination -- I'll re-introduce the overrides for these languages. I'm having a problem fighting the different sources of truth :p Definitely Latn should be considered the primary script in Azerbaijan.

I think the source problem is that "Cyrl" comes before "Latn" alphabetically and when the script is re-run now it takes the first alphabetical item.

common/supplemental/supplementalData.xml
conradarcturus: The code is writing it alphabetically, so when I re-generate the script it force-alphabetizes it. There is an argument it should be ordered by usage --- however the XML is just not a good way to capture this because the labelling is unclear.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv
conradarcturus: Promoting Grek to be the primary script for Pontic.

Really for all current Pontic speakers its Grek in Greece, Latn in Turkey, and Cyrl in Russia/Ukraine. Pontic is only spoken by very marginal populations in Turkey and Russia, but it's a large recognized community in Greece.

What's the basis for primary v secondary?

macchiati: The primary vs secondary should be based on the literate population sizes. I forget what the cutoff is, but clearly if >50% of the usage of the language is in a particular script, that would be primary, not secondary. (But again, there might be bug in the code.)

common/supplemental/supplementalData.xml
macchiati: The scripts (and regions) should be in ranked order, not sorted. If the code is sorting them, that's a bug.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateLikelySubtags.java
```diff
@@ -392,6 +393,9 @@ public static void main(String[] args) throws IOException {
                                 {"mro", "mro_Mroo_BD"},
                                 {"mro_BD", "mro_Mroo_BD"},
                                 {"ms_Arab", "ms_Arab_MY"},
+                                {"nan", "nan_Hans_CN"},
+                                {"nan_Hans", "nan_Hans_CN"},
```
macchiati, line 397: This won't hurt anything, but nan_Hans is redundant, because the algorithm will find {"nan", "nan_Hans_CN"}, and fill in.

There is a ticket open for dropping overrides that have no effect, so it is ok to keep this line for now.

common/supplemental/supplementalData.xml
macchiati: It looks like if Roozbehs' items are taken care of, then this would be ready to merge into 47.

conradarcturus: I'm glad Roozbeh took a look ;) I resolved this by changing the non-Latn script for these languages to be considered "secondary" in language_script.tsv

Merging this changes ended up getting really messy so I'll post a new pull request.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateLikelySubtags.java
conradarcturus: Interestingly, I need to keep this like otherwise in the produced `likelySubtags.xml` file, it will show `nan_Hans -> nan_Hans_TW` even though as we know the `Hant` script would be preferred in Taiwan. The problem is that we don't have population estimates on Simplified v Traditional Chinese script usage.

conradarcturus:
Thanks everyone for the comments! It helped me make a better version of this PR in https://github.com/unicode-org/cldr/pull/4015.

Apologies for making a separate one -- rebasing it to the ddl/v47 branch introduced weird merge artifacts so I just made a new PR.

PR #3653, opened 2024-04-23
CLDR-16499 Prepare to Require a CLA for contributions

srl295:
CLDR-16499

For v46, we do NOT require a CLA, but leave the code in place for the future

- property `REQUIRE_CLA` off by default, if on it will enable the flow
- no popups or menu items visible to users
- enables loading .md files to the front end and as assets
- improve cldrClient.getClient() to cache, speedup repeated calls
- add notice for synchronizing webpack configuration
- APIs for sign/revoke/read CLA
- web flow for signing CLA and handling various cases

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4010, opened 2024-09-03
CLDR-14913 Add Download XML button to Survey Tool

btangmu:
-Front end: new GenerateVxml.vue, cldrGenerateVxml.mjs

-Back end: new GenerateVxml.java, VxmlQueue.java, VxmlGenerator.java

-Run the task in its own thread, with multiple http requests fetching status until done

-Partly imitate implementation of Priority Items Summary: GenerateVxml.java corresponds to Summary.java; VxmlGenerator.java corresponds to VettingViewer.java; VxmlQueue.java corresponds to VettingViewerQueue.java

-Revise OutputFileManager.java as needed and also to fix some (not all) compiler warnings

-Disable Announce requests while VXML generation is in progress (only on front end; only affects the browser of the Admin user, not other users if running on remote server)

-Fix bug in XPathTable.loadXPaths, DB connection without commit, change getDBConnection to getAConnection for auto-commit (pre-existing bug was especially problematic throwing exception when generating VXML)

-Improve debugging feature in DBUtils.java slightly and add comment about remaining imperfection

-Use auto-commit in STFactory.setupDB and make the related boolean dbIsSetup private

-Remove dead code

-Comments

CLDR-14913

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
@@ -6,12 +6,6 @@ import * as cldrAjax from "./cldrAjax.mjs";
 import * as cldrSchedule from "./cldrSchedule.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 
-/**
- * This should be false for production. It can be made true during debugging, which
- * may be useful for performance testing.
- */
-const DISABLE_ANNOUNCEMENTS = false;
-
 const CLDR_ANNOUNCE_DEBUG = false;
 
 const ANNOUNCE_REFRESH_SECONDS = 60; // one minute
@@ -39,6 +33,16 @@ const MOST_RECENT_ID_UNKNOWN = -1; // must be less than zero
 let alreadyGotId = MOST_RECENT_ID_UNKNOWN;
 
 /**
+ * Ordinarily announcements are enabled. They may be temporarily disabled during
+ * critical operations such as VXML generation, or for debugging.
+ */
+let announcementsEnabled = true;
+
+function enableAnnouncements(enable) {
+  announcementsEnabled = Boolean(enable);
+}
+
+/**
  * Get the number of unread announcements, to display in the main menu
  *
  * @param {Function} setUnreadCount the callback function for setting the unread count
@@ -60,7 +64,7 @@ async function getUnreadCount(setUnreadCount) {
  * we're only getting the number of unread announcements to display in the main header
  */
 async function refresh(viewCallbackSetData, viewCallbackSetCounts) {
-  if (DISABLE_ANNOUNCEMENTS) {
+  if (!announcementsEnabled) {
     return;
   }
   if (viewCallbackSetData) {
@@ -185,6 +189,7 @@ export {
   canAnnounce,
   canChooseAllOrgs,
   compose,
+  enableAnnouncements,
   getUnreadCount,
   refresh,
   resetSchedule,
--- a/tools/cldr-apps/js/src/esm/cldrGenerateVxml.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrGenerateVxml.mjs
@@ -1,0 +1,88 @@
+/*
+ * cldrGenerateVxml: for Survey Tool feature "Generate VXML". The display logic is in GenerateVxml.vue.
+ */
+import * as cldrAjax from "./cldrAjax.mjs";
+import * as cldrAnnounce from "./cldrAnnounce.mjs";
+import * as cldrNotify from "./cldrNotify.mjs";
+import * as cldrStatus from "./cldrStatus.mjs";
+
+const SECONDS_IN_MS = 1000;
+
+const NORMAL_RETRY = 10 * SECONDS_IN_MS; // "Normal" retry: starting or about to start
+
+const VXML_URL = "api/vxml";
+
+// These must match the back end; used in requests
+class LoadingPolicy {
+  static START = "START"; // start generating vxml
+  static CONTINUE = "CONTINUE"; // continue generating vxml
+  static STOP = "STOP"; // stop (cancel) generating vxml
+}
+
+// These must match the back end; used in responses
+class Status {
+  static INIT = "INIT"; // before making a request (back end does not have INIT)
+  static WAITING = "WAITING"; // waiting on other users/tasks
+  static PROCESSING = "PROCESSING"; // in progress
+  static READY = "READY"; // finished successfully
+  static STOPPED = "STOPPED"; // due to error or cancellation (LoadingPolicy.STOP)
+}
+
+let canGenerate = false;
+
+let callbackToSetData = null;
+
+function canGenerateVxml() {
+  return canGenerate;
+}
+
+function viewMounted(setData) {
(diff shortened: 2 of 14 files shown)


btangmu:
Screenshot while running locally, generating VXML:
image

btangmu:
Screenshot while running locally, when VXML generation is complete, before scrolling down:
image

btangmu:
Screenshot while running locally, when VXML generation is complete, after scrolling down:
image

Note: the display of each locale name followed by "vxml xml rxml pxml" is like the old behavior with jsp, and is far from ideal. This PR should be followed by additional PRs to do the following:

- Remove obsolete admin-OutputAllFiles.jsp
- Improve the display, such as by displaying the list of locales more compactly (no need to scroll, no need for "vxml xml rxml pxml", ...)
- Improve the interface. For example, could add a button to copy the destination folder name to the clipboard; could automatically create a zip file; after Success or Stop, the button could change to Reset, to restore the original setup (same as if reloaded the page); generation of pxml, often superfluous, could be contingent on a checkbox
- Clean up the code

btangmu:
Note: To obtain those screenshots, I removed common/annotations/ff.xml locally. It is essentially empty and if present it leads to verification failure: "present in baseline but not in vxml". However, if absent, the result is “TestFileIds Error: (TestCLDRFile.java:814) Error: Missing simple parent (ff) for ff_Adlm in common/annotations; likely=ff_Adlm”

macchiati:
Note: the green on red is almost impossible to read. Black on red?

tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
```diff
@@ -38,6 +32,16 @@ const MOST_RECENT_ID_UNKNOWN = -1; // must be less than zero
  */
 let alreadyGotId = MOST_RECENT_ID_UNKNOWN;
 
+/**
+ * Ordinarily announcements are enabled. They may be temporarily disabled during
+ * critical operations such as VXML generation, or for debugging.
+ */
+let announcementsEnabled = true;
```
srl295, line 39: why does it make a difference?

tools/cldr-apps/js/src/esm/cldrGenerateVxml.mjs
```diff
@@ -0,0 +1,88 @@
...
+  const args = { loadingPolicy: loadingPolicy };
+  const init = cldrAjax.makePostData(args);
+  cldrAjax
+    .doFetch(VXML_URL, init)
+    .then(cldrAjax.handleFetchErrors)
+    .then((r) => r.json())
+    .then(setVxmlData)
+    .catch((e) => {
+      cldrNotify.exception(e, "generating VXML");
+    });
```
srl295, line 73: could consider using cldrClient  - provides error checking of input parameters. at least for future

tools/cldr-apps/src/main/java/org/unicode/cldr/web/OutputFileManager.java
```diff
@@ -320,20 +323,26 @@ private boolean outputAllFiles(Writer out, File vetDataDir) {
         }
     }
 
+    public static Set<CLDRLocale> createVxmlLocaleSet() {
+        Set<CLDRLocale> set = new TreeSet<>(SurveyMain.getLocalesSet());
+        // skip "en" and "root", since they should never be changed by the Survey Tool
+        set.remove(CLDRLocale.getInstance("en"));
+        set.remove(CLDRLocale.getInstance(LocaleNames.ROOT));
+        // Remove "mul", "mul_ZZ", etc.
+        set.removeIf(loc -> loc.getBaseName().startsWith(LocaleNames.MUL));
```
srl295, line 332: use SpecialLocales.java instead - in fact, if SpecialLocales.getType() returns non-null then skip. (sandbox (mul). algorithmic (sr_Latn) and other readonly.)

srl295 (commented):
Looks good so far.

tools/cldr-apps/js/src/esm/cldrAnnounce.mjs
btangmu: Thrown exceptions due to uncommitted non-auto-commit db connections were preventing VXML generation from running to completion. While debugging that, the Announce requests were running in parallel with the VXML requests, making the problem harder to track down. In the past there have also been concurrency bugs when generation was implemented as an auto-scheduled background task, and it's not clear whether those were ever solved or just avoided by not auto-scheduling. Even when all is going well, I like to keep an eye on the various things that are happening during VXML generation. Temporarily disabling announcements reduces complexity and server load and makes for one less thing to worry about.

tools/cldr-apps/js/src/esm/cldrGenerateVxml.mjs
btangmu: That's a good idea for future enhancement. I'll study it. I think there may still be problem with cldrClient when running locally with nginx, confusion between ports 8888 and 9080. It can be bypassed by skipping nginx and going directly to 9080, but I'd like to get it working right.

btangmu:
> green on red is almost impossible to read

Yes. As I mentioned in a comment with one of the screenshots, that text is carried over from the old interface and I don't see any benefit to it:

>Improve the display, such as by displaying the list of locales more compactly (no need to scroll, no need for "vxml xml rxml pxml", ...)

I'd like to do that in a follow-up PR since this PR is already far-reaching and accomplishes the essential goals of modernizing the client-server interface and using a separate thread on the server rather than taking ten minutes for the server to respond to the original request.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/OutputFileManager.java
btangmu: I tried that and got different results. These locales get removed, so VXML isn't generated: ha_NE, sr_Latn, sr_Latn_BA, sr_Latn_ME, sr_Latn_XK, yo_BJ. They are all "algorithmic". Their files are present and not "empty" in baseline. If they don't get VXML generated, we'll get "Verification failure, file(s) present in baseline but not in vxml".

Would it be OK to merge this PR as-is, and then follow up with improvements?

srl295: such files should be copied from baseline and not (re)generated.

tools/cldr-apps/js/src/esm/cldrGenerateVxml.mjs
srl295: it works via nginx in production. And anyway, I think the baseline url issue may have been fixed when I fixed a similar issue with Chrome. 

Or, maybe it's not fixed in main, it's waiting in the "CLA PR" which is awaiting review.

srl295 (approved):

tools/cldr-apps/js/src/esm/cldrGenerateVxml.mjs
btangmu: "CLA PR"?

srl295: > "CLA PR"?

#3653

PR #4011, opened 2024-09-03
CLDR-17938 Add test data generator for date time formatting

echeran:
CLDR-17938

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -0,0 +1,104 @@
+package org.unicode.cldr.tool;
+
+import java.io.IOException;
+import java.util.Arrays;
+import org.pcollections.HashPMap;
```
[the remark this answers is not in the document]
echeran: I support that concern generally. The reason I thought to introduce a persistent collection library is that there are multiple dimensions that need to be iterated over (generate combinations for all calendar systems, for all selected dates, for all selected times, for all ways of specifying style/length (skeleton vs. style options), etc.). In particular, the fields for skeleton and the style options in a test case should be mutually exclusive -- if one has data, then the other should be empty.

Using immutable maps across the nested loops eliminates chances for unintended consequences of stateful mutation. I strongly prefer immutable maps for this reason, and thus do not want to use regular collections. But that also means I need to append to immutable collections.

I didn't immediately see it in the Guava APIs, but [I guess it is possible](https://stackoverflow.com/questions/29828829/copy-an-immutablemap-with-additional-or-new-values), although doing so is clunky. Going this route would compel me to write helper methods to reduce that boilerplate, which would feel like I'm partially recreating the persistent collections library that point.

On the other hand, persistent collections are designed for supporting immutability + append/remove/etc., and have the added [benefit of code readability](https://www.baeldung.com/java-pcollections) b/c of concision when you do perform those operations. The PCollections library has [no dependencies and 75 KB of jar size](https://mvnrepository.com/artifact/org.pcollections/pcollections/4.0.2), FWIW.

I'm happy to do what you suggest, here. I would prefer that it entail immutable data structures -- either via Guava or PCollections.

srl295:
PR title needs to start with `CLDR-… ` as should the commit

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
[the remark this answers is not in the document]
macchiati: We don't really have time to do an analysis of a new package to see if it introduces any new security / performance issue If you are not modifying these structures once built, it is pretty easy to just build them and convert to immutable structures.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -0,0 +1,104 @@
...
+  private static final PVector<String> TIME_ZONES =
+      TreePVector.from(Arrays.asList(
+          "America/Los_Angeles",
+          "Africa/Luanda",
+          "Asia/Tehran",
+          "Europe/Kiev",
+          "Australia/Brisbane",
+          "Pacific/Palau",
+          "America/Montevideo"
+      ));
```
srl295, line 91: other code uses ImmutableSet.of("…", "…", "…") - does this provide value beyond that?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private static final PVector<String> LOCALES =
+      TreePVector.from(Arrays.asList("en-US", "en-GB",
+          "zh-TW", "vi", "ar", "mt-MT",
+          "bn", "zu",
+          "und"));
+
+  private static final HashPMap<String,Object> emptyMapStringToObject = HashTreePMap.empty();
+
+  private static final PVector<PMap<String, Object>> SPEC_OPTIONS =
```
srl295, line 30: Couldn't this just be read from the DTD?

srl295 (commented):
- i don't see the argument for the pcollections
- as to iterating - that data should be pulled from the DTD anyway instead of hard coded.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
echeran: I switched over to using Guava according to Mark's guidance, so it's a moot point now.

(There's a bit of cloning during runtime and verbosity of code that could be avoided if one were to use persistent data structures.)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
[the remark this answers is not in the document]
echeran: Okay, removed PCollections as a dep, using the existing Guava dep's collections instead.

echeran:
@macchiati @srl295 Do you know what the proper way to format a date within CLDR code is? @sffc and I found a few separate instances of date time formatting, such as in `cldr-code` -> [`org.unicode.cldr.draft.Misc#getFormatted`](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/java/org/unicode/cldr/draft/Misc.java#L289-L306) that uses ICU4J's `DateTimePatternGenerator` and `SimpleDateFormat`. Is that the way to do it?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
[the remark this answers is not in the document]
macchiati: Thanks!

macchiati:
There is a class called ICUServiceBuilder, that patches ICU calls with CLDR data. (If you are doing something new we might need to add APIs to that.)

srl295:
a good source of example code would be the ExampleGenerator.java

echeran:
Thanks @macchiati and @srl295 .  @sffc and I took a look at `ExampleGenerator.java` and found in `handleDateFormatItem` that finds the right xpaths and the values at that those xpaths for the date/time patterns, and then calls `icuServiceBuilder.getDateFormat(...)` using those values in the call site.

There's a lot of important and necessary code in that method that would be better off refactored into a reusable place rather than taking a copy-and-paste shortcut. The method itself has a lot of copy and-pasted code 3x over for full/long/short date format patterns. We think it would be good if we also cleaned up & refactored the duplicated code as a part of creating the test data generator. WDYT?

macchiati:
Sounds good; pulling useful methods out of ExampleGenerator (etc). There is
a pending PR (from one of the interns) that refactors ExampleGenerator, so
you should wait until that is merged to avoid conflicts.

srl295:
@echeran dont' forget to open a ticket and update the PR title and commit messages.

sffc:
> Sounds good; pulling useful methods out of ExampleGenerator (etc). There is a pending PR (from one of the interns) that refactors ExampleGenerator, so you should wait until that is merged to avoid conflicts.

Is it PR https://github.com/unicode-org/cldr/pull/3895?

It looks like it is just stuck on a linter. @echeran pushed the lint fix to the branch.

macchiati:
Still failing spotless, build — and is draft. Elango, do you want to try to fix this for v46, or punt to 47?

srl295:
yes thanks for merging #3895

echeran:
> Still failing spotless, build — and is draft. Elango, do you want to try to fix this for v46, or punt to 47?

Still trying to get this done ASAP in case it can fit for v46. When is the last call for v46?

macchiati:
Any test data that is targeted at ICU should be in before their freeze date
(Thursday??).

echeran:
> Any test data that is targeted at ICU should be in before their freeze date (Thursday??).

Things are looking good and complete now. The only thing is the Jira ticket isn't accepted yet. The key files are `GenerateDateTimeTestData.java` and its output, `datetime.json`.

sffc:
The data currently looks like

```
  {
    "dateStyle": "short",
    "timeStyle": "short",
    "calendar": "gregorian",
    "formatType": "standard",
    "locale": "en-US",
    "dateLength": "short",
    "timeLength": "short",
    "timeZone": "Pacific/Palau",
    "input_string": "1969-07-16T16:00+09:00[Pacific/Palau]",
    "verify": "7/16/69, 4:00 PM"
  },
```

A few suggestions for improvement:

1. Remove `formatType` since we aren't using it and it is a bad name (unclear what it means)
2. Keep dateStyle/timeStyle or dateLength/timeLength but not both
3. Remove `timeZone` since it is already in the input string
4. Rename the field "input_string" to "input" or "datetime"
5. Rename the field "verify" to "output" or "expected"

echeran:
> A few suggestions for improvement:

Done. All suggestions were good. PTAL.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -513,11 +513,11 @@ private static ImmutableMap<Object, Object> getTestCaseForZonedDateTime(
                         dateTimeGluePatternFormatType,
                         icuServiceBuilder);
 
-        // "input_string" = the ISO 18601 UTC time zone formatted string of the zoned date time
-        optionsBuilder.put("inputString", zdt.toString());
+        // "input" = the ISO 18601 UTC time zone formatted string of the zoned date time
```
macchiati, line 516: ISO - International Organization for Standardization

ISO 18601:2013 - Packaging and the environment — General requirements for the use of ...

macchiati:
I can't review the Jason file right now, because it's too big for the browser.

echeran:
> I can't review the Jason file right now, because it's too big for the browser.

For the purposes of just quick reviewing, you can view the PR files including the large JSON file by replacing the `.com` with `.dev` in the PR URL, and it spins up VS Code in the browser to make viewing the files manageable:

https://github.dev/unicode-org/cldr/pull/4011

(If you had to do it at the command line, you would have to check out the branch via `gh pr checkout 4011` and navigate yourself to the file, which takes a few more steps.)

macchiati (approved):
Thanks for the tip, Elango. Data looks ok to me. 

(I was on a tablet, and so constained in how I could review.)

(longer term, for each testdata file I think we want to have a unittest in CLDR; and in the testdata file point to the generator and the unit test. That will make it easier to manage.)

echeran:
Blocked on the Jira ticket being accepted. Everything else is ready to go.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -0,0 +1,656 @@
...
+    private static final CLDRConfig CLDR_CONFIG = CLDRConfig.getInstance();
+
+    private static final Factory CLDR_FACTORY = CLDR_CONFIG.getCldrFactory();
+
+    private static final Gson GSON = new GsonBuilder().setPrettyPrinting().create();
+
+    private static final ImmutableSet<String> NUMBERING_SYSTEMS =
+            ImmutableSet.of("latn", "arab", "beng");
+
+    // Use underscores for locale id b/c of CLDR historical reasons, even though dash is preferable
```
srl295, line 56: could be in bcp47 and then convert them to CLDR using ULocale.forLanguageTag

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
srl295:
```suggestion
        // "input" = the ISO 8601 UTC time zone formatted string of the zoned date time
```

srl295 (approved):
OK to move forward (minor comments)

PR #4028, opened 2024-09-09
CLDR-17935 site: add crawler sitemap

srl295:
CLDR-17935

- add sitemap.xml for crawlers

This is unrelated to site organization or navigation. It's so that the crawlers can index the site properly.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/.gitignore
+++ b/docs/site/.gitignore
@@ -1,3 +1,5 @@
 /node_modules
 /assets/json
 /assets/vendor
+/sitemap.xml
+/sitemap.md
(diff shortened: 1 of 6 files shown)


macchiati (changes requested):
Needs some work; let's discuss Tuesday

srl295:
> Needs some work; let's discuss Tuesday

I took out the human-readable `sitemap.md` - a higher priority may be having sitemap.xml in so that when the new site is live, that it is crawlable, because crawlers may or may not be able to navigate the popup menu.

https://094f5448.cldr.pages.dev/sitemap.xml

So, at this point the PR has no visible change, but the sitemap (for crawlers) might be worth getting in as or soon after going live.

conradarcturus (approved):
Let the robots in. Looks solid.

srl295:
@macchiati any objection?

macchiati (approved):

srl295:
Great… The spiders should pick this up automatically, but I'll update them manually just in case

PR #4033, opened 2024-09-10
CLDR-17934 site: drive from sitemap.tsv

srl295:
CLDR-17934

- `sitemap.tsv` now controls the order and hierarchy
- errors will be given from build.mjs (and at PR CI time) if a page is missing or other errs
- this PR DOES NOT reorg the site at all, but it paves the way for reorg
- also de-yellow the title line

image

ALLOW_MANY_COMMITS=true


macchiati (commented):
TSV looks good

AEApple:
Will we just re-order the TSV in a different ticket?

macchiati (approved):

macchiati:
> Will we just re-order the TSV in a different ticket?

We will be able to make PRs to change the sitemap, so we should have a lot of flexibility.
For example, I think `stable-links-info` needs to go into internal development (shouldn't be so prominent) soon

PR #4034, opened 2024-09-10
CLDR-17071 Do not create unused db table sf_loc2forum; improve localeToForum

btangmu:
-Remove code that creates unused db table sf_loc2forum

-Revise the implementation of localeToForum by calling getLanguage instead of toULocale().getLanguage

CLDR-17071

ALLOW_MANY_COMMITS=true


btangmu:
Note: after this is merged, we should remove the actual table sf_loc2forum from the production db:

`drop table sf_loc2forum;`

srl295 (approved):

PR #4036, opened 2024-09-10
CLDR-16720 json: add transforms

srl295:
- new package cldr-transforms
- add manifest file transforms.json at the top level
- each transform has a metadata file (transforms/ID.json) and a raw text file (transforms/ID.txt).
- metadata has all of the keys from the transform rule
- the _rulesFile key formally indicates the textfile's name (in case we need to massage the id for some reason in the future).

Sample data available at this branch: https://github.com/unicode-org/cldr-json/tree/cldr-16720/transforms/cldr-json/cldr-transforms

CLDR-16720

ALLOW_MANY_COMMITS=true


srl295:
Deploy will fail because I used my personal fork, so no preview URL. 

Please review the Markdown change carefully.

macchiati (approved):

sffc:
The mix of JSON and TXT files in the same directory might make it hard to parse. The file extension is the only way to tell the difference, which usually isn't ideal. Can you split them into separate directories?

Also, since the JSON files are basically metadata on the TXT files, I have a mild expectation that there would be a single index.json with all the transform metadata in one place.

robertbastian:
> Also, since the JSON files are basically metadata on the TXT files, I have a mild expectation that there would be a single index.json with all the transform metadata in one place.

This is per-locale data. Nowhere in CLDR-JSON multiple locales are merged in an index file, they always use the file system structure.

srl295:
> The mix of JSON and TXT files in the same directory might make it hard to parse. The file extension is the only way to tell the difference, which usually isn't ideal. Can you split them into separate directories?

It's designed so clients don't need to crawl the directory:

1. Look at the `transforms.json` file. It has a list of ids, with no extension
2. For each id, there is `transforms/`_id_`.json` with the metadata.
3. One of the metadata items is `_rulesFile` which has the path to the .txt file.

> Also, since the JSON files are basically metadata on the TXT files, I have a mild expectation that there would be a single index.json with all the transform metadata in one place.

There is, it's `transforms.json`.

srl295:
I see a couple of bugs:
- `_source` and `_target` need to be bcp47, not old IDs.
- There's a bug in `_alias` that has some corruption.
- the 2nd level key in the metadata .json has a problem (because that ID might have slashes in it)

sffc:
> > Also, since the JSON files are basically metadata on the TXT files, I have a mild expectation that there would be a single index.json with all the transform metadata in one place.
> 
> This is per-locale data. Nowhere in CLDR-JSON multiple locales are merged in an index file, they always use the file system structure.

https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-core/supplemental/plurals.json

> Look at the transforms.json file. It has a list of ids, with no extension

OK, that looks cool, thanks! I didn't see it the first time.

I still think I mildly favor not putting the JSON and TXT in the same directory, but I'll leave that to @robertbastian to weigh in on.

srl295:
I don't see why being in the same directory would be a problem. 

As Mark suggested, we need docs on the formats. That will be a separate effort.

macchiati (approved):

macchiati (dismissed):
I had approved this, but Steven mentioned 3 outstanding bugs:
-  _source and _target need to be bcp47, not old IDs.
-  There's a bug in _alias that has some corruption.
-  the 2nd level key in the metadata .json has a problem (because that ID might have slashes in it)

macchiati:
Shane, are there any other blockers?

srl295 (commented):
'source' and 'target' aren't strictly locale IDs, so we cant just convert them to bcp47. 

The aliases field includes bcp47 aliases, so perhaps nothing more needs to be done.

srl295:
Please review sample data in https://github.com/unicode-org/cldr-json/tree/cldr-16720/transforms/cldr-json/cldr-transforms

I now think source and target should not be bcp47 as they aren't always locale IDs.  The `alias` field contains a bcp47 alias.

macchiati:
Right, as long as the alias data is preserved. Eg,

<transform source="Any" target="Accents" direction="both" alias="
*und-t-d0-accents*" backwardAlias="*und-t-s0-accents*">

However it would probably be better to pull out the bcp47 items into
separate fields. Could be done later, though.

sffc:
What is the name of the second level key? In Amharic-Latin-BGN.json it is BGN

```
{
  "transforms": {
    "BGN": {
      "_value": "Amharic-Latin-BGN.txt",
      "_visibility": "external",
      "_alias": "Amharic-Latin/BGN am-Latn-t-am-m0-bgn",
      "_source": "am",
      "_target": "am_Latn",
      "_direction": "forward"
    }
  }
}
```

But in most other files it is just "transform".

I agree that it would be nice to name the files consistently with either the alias name or the BCP-47 name but not a mix as is currently in the branch.

robertbastian:
I'm not a fan of the nesting in the JSON files. 

```json
{
  "transforms": {
    "BGN": {
      "_value": "Arabic-Latin-BGN.txt",
      "_visibility": "external",
      "_alias": "Arabic-Latin/BGN ar-Latn-t-ar-m0-bgn",
      "_source": "ar",
      "_target": "ar_Latn",
      "_direction": "forward"
    }
  }
}
```

could be represented as

```json
{
  "_value": "Arabic-Latin-BGN.txt",
  "_visibility": "external",
  "_alias": "Arabic-Latin/BGN ar-Latn-t-ar-m0-bgn",
  "_source": "ar",
  "_target": "ar_Latn",
  "_variant": "BGN",
  "_direction": "forward"
}
```

As far as I can tell no JSON file will have multiple values in the `transforms` map.

macchiati:
Sounds reasonable. The _value should be more meaningful, like: rules.
It doesn't need a _ since it doesn't correspond to an attribute in XML.

srl295:
> What is the name of the second level key? In Amharic-Latin-BGN.json it is BGN

That's a bug. Can we fix it in a future PR?

> ```
> {
>   "transforms": {
>     "BGN": {
>       "_value": "Amharic-Latin-BGN.txt",
>       "_visibility": "external",
>       "_alias": "Amharic-Latin/BGN am-Latn-t-am-m0-bgn",
>       "_source": "am",
>       "_target": "am_Latn",
>       "_direction": "forward"
>     }
>   }
> }
> ```
> 
> But in most other files it is just "transform".
> 
> I agree that it would be nice to name the files consistently with either the alias name or the BCP-47 name but not a mix as is currently in the branch.

The *files* are by the id name, which is neither the bcp47 nor the alias name. As explained.

srl295:
Agh. another bug. _value is supposed to be _ruleFile.

srl295:
I'll change it to:

```json
{
  "transform":  { 
       ...
  }
}
```

all of the json files have duck typed content similarly.

srl295:
please recheck sample data in https://github.com/unicode-org/cldr-json/tree/cldr-16720/transforms/cldr-json/cldr-transforms

sffc:
I still don't understand why half of these are identified by BCP-47 and half are identified by their alias.

```
Armenian-Latin-BGN.json
{
  "_visibility": "external",
  "_alias": "Armenian-Latin/BGN hy-Latn-t-hy-m0-bgn",
  "_source": "hy",
  "_target": "hy_Latn",
  "_direction": "forward",
  "_rulesFile": "Armenian-Latin-BGN.txt"
}

am-Ethi-t-am-brai.json
{
  "_backwardAlias": "Braille-Ethiopic/Amharic am-Ethi-t-am-brai",
  "_visibility": "external",
  "_alias": "Ethiopic-Braille/Amharic am-Brai-t-am-ethi",
  "_source": "am_Ethi",
  "_target": "am_Brai",
  "_direction": "both",
  "_rulesFile": "am-Ethi-t-am-brai.txt"
}
```

srl295:
> I still don't understand why half of these are identified by BCP-47 and half are identified by their alias.
> 
> ```
> Armenian-Latin-BGN.json
> {
>   "_visibility": "external",
>   "_alias": "Armenian-Latin/BGN hy-Latn-t-hy-m0-bgn",
>   "_source": "hy",
>   "_target": "hy_Latn",
>   "_direction": "forward",
>   "_rulesFile": "Armenian-Latin-BGN.txt"
> }
> 
> am-Ethi-t-am-brai.json
> {
>   "_backwardAlias": "Braille-Ethiopic/Amharic am-Ethi-t-am-brai",
>   "_visibility": "external",
>   "_alias": "Ethiopic-Braille/Amharic am-Brai-t-am-ethi",
>   "_source": "am_Ethi",
>   "_target": "am_Brai",
>   "_direction": "both",
>   "_rulesFile": "am-Ethi-t-am-brai.txt"
> }
> ```

that's how they are identified in the source data.

macchiati:
In CLDR, the BCP47 version of the ID is in the _alias list (resp _backwardAlias)
Those alias lists also contain the non-BCP47 alias.

In JSON, we could filter these to break:
```
  "_backwardAlias": "Braille-Ethiopic/Amharic am-Ethi-t-am-brai",

  "_alias": "Ethiopic-Braille/Amharic am-Brai-t-am-ethi",
```
into
```
  "_backwardAlias": "Braille-Ethiopic/Amharic",
  "_backwardAliasBcp47": "am-Ethi-t-am-brai",

  "_alias": "Ethiopic-Braille/Amharic",
  "_aliasBcp47": "am-Brai-t-am-ethi",
```

Even better would be for us to do this in the XML source, but that's not something we could do in v46

srl295:
@macchiati filter how? How do I know which alias is which?

macchiati:
I think it is sufficient to parse with a strict BCP47 parser. If it
succeeds without error, it is BCP47, otherwise legacy.

srl295:
> I think it is sufficient to parse with a strict BCP47 parser. If it succeeds without error, it is BCP47, otherwise legacy.

seems a bit imprecise/mixed but OK

macchiati:
It will work, but as I wrote earlier we should add better structure

srl295:
Can you recommend a parser and options ?

srl295:
OK, try the latest, same review link.

```json
{
  "_backwardAlias": "Latin-Ethiopic/Tekie_Alibekit",
  "_visibility": "external",
  "_backwardAliasBcp47": "byn-Ethi-t-byn-latn-m0-tekieali",
  "_alias": "Ethiopic-Latin/Tekie_Alibekit",
  "_aliasBcp47": "byn-Latn-t-byn-ethi-m0-tekieali",
  "_source": "byn_Ethi",
  "_direction": "both",
  "_target": "byn_Latn",
  "_rulesFile": "byn-Ethi-t-byn-latn-m0-tekie-alibekit.txt"
}
```

macchiati (approved):
Respot-checked data, looks good to me.

PR #4037, opened 2024-09-10
CLDR-17892 ISO 8601 calendar

macchiati:
CLDR-17892

Most of the time was in writing something that would show the differences clearly. See the results on:

https://docs.google.com/spreadsheets/d/1fIwEYN8AxoJruBuHuMCoer5WIkEBgFD8i1M2quacJiw/edit

Changes from TC review: now retain the G and 'h...a'.

- ISO order doesn't allow for DOW, so I put that after the day, thus before the time.
- The test now makes a few sanity checks, like order of variables
- Had to modify some other files (PathHeader.*, TestDisplayAndInputProcessor.java)
- There were a number of failures due to the wrong space, eg around –. (We should check to make sure that DAIP runs on root too!)

ALLOW_MANY_COMMITS=true


AEApple:
Interesting it looks like a lot of the differences in the date ranges may be in spacing. Do you want me to propose edits or wait for tomorrow since I wasn't sure how your are generating this list.

macchiati:
I put in a column for comments, you can add either a recommended change there or a comment.

macchiati:
I am generating the spreadsheet contents from the root file. That file has the new calendar iso8601, which aliases the gregorian for names of stuff (months, weekdays, etc), but otherwise uses the explicit patterns that were based on root's version of gregorian.

AEApple:
> I am generating the spreadsheet contents from the root file. That file has the new calendar iso8601, which aliases the gregorian for names of stuff (months, weekdays, etc), but otherwise uses the explicit patterns that were based on root's version of gregorian.

ah okay! We could use en-CA as a reasonable proxy, for the most part en-CA is intentionally following ISO with a few exceptions.

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<dateFormatItem id="Ehm">E HH:mm</dateFormatItem>
+						<dateFormatItem id="EHm">E HH:mm</dateFormatItem>
+						<dateFormatItem id="Ehms">E HH:mm:ss</dateFormatItem>
+						<dateFormatItem id="EHms">E HH:mm:ss</dateFormatItem>
+						<dateFormatItem id="Gy">y</dateFormatItem>
+						<dateFormatItem id="GyMd">y-MM-dd</dateFormatItem>
+						<dateFormatItem id="GyMMM">y MMM</dateFormatItem>
+						<dateFormatItem id="GyMMMd">y MMM d</dateFormatItem>
+						<dateFormatItem id="GyMMMEd">y MMM d, E</dateFormatItem>
+						<dateFormatItem id="h">HH</dateFormatItem>
```
pedberg-icu, line 2058: No, a skeleton of h should always map to a pattern with h or K, we say that explicitly in the spec: https://www.unicode.org/reports/tr35/tr35-dates.html#availableFormats_appendItems

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<dateFormatItem id="H">HH</dateFormatItem>
+						<dateFormatItem id="hm">HH:mm</dateFormatItem>
```
pedberg-icu, line 2060: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<dateFormatItem id="Hm">HH:mm</dateFormatItem>
+						<dateFormatItem id="hms">HH:mm:ss</dateFormatItem>
```
pedberg-icu, line 2062: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<dateFormatItem id="Hms">HH:mm:ss</dateFormatItem>
+						<dateFormatItem id="hmsv">HH:mm:ss v</dateFormatItem>
```
pedberg-icu, line 2064: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<dateFormatItem id="Hmsv">HH:mm:ss v</dateFormatItem>
+						<dateFormatItem id="hmv">HH:mm v</dateFormatItem>
```
pedberg-icu, line 2066: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="y">y MMM d – y MMM d</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="GyMMMEd">
+							<greatestDifference id="d">y MMM d, E – MMM d, E</greatestDifference>
+							<greatestDifference id="G">y MMM d, E – y MMM d, E</greatestDifference>
+							<greatestDifference id="M">y MMM d, E – MMM d, E</greatestDifference>
+							<greatestDifference id="y">y MMM d, E – y MMM d, E</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="h">
+							<greatestDifference id="a">HH–HH</greatestDifference>
```
pedberg-icu, line 2155: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="h">HH–HH</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="H">
+							<greatestDifference id="H">HH–HH</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="hm">
+							<greatestDifference id="a">HH:mm–HH:mm</greatestDifference>
```
pedberg-icu, line 2162: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="h">HH:mm–HH:mm</greatestDifference>
+							<greatestDifference id="m">HH:mm–HH:mm</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="Hm">
+							<greatestDifference id="H">HH:mm–HH:mm</greatestDifference>
+							<greatestDifference id="m">HH:mm–HH:mm</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="hmv">
+							<greatestDifference id="a">HH:mm–HH:mm v</greatestDifference>
```
pedberg-icu, line 2171: No, map to pattern with h or K per previous comment

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="h">HH:mm–HH:mm v</greatestDifference>
+							<greatestDifference id="m">HH:mm–HH:mm v</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="Hmv">
+							<greatestDifference id="H">HH:mm–HH:mm v</greatestDifference>
+							<greatestDifference id="m">HH:mm–HH:mm v</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="hv">
+							<greatestDifference id="a">HH–HH v</greatestDifference>
```
pedberg-icu, line 2180: No, map to pattern with h or K per previous comment

pedberg-icu (dismissed):
Skeletons with h are only allowed to map to patterns with h or K per the spec, see inline comments.

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<intervalFormatItem id="Bhm">
+							<greatestDifference id="B">h:mm B – h:mm B</greatestDifference>
+							<greatestDifference id="h">h:mm–h:mm B</greatestDifference>
+							<greatestDifference id="m">h:mm–h:mm B</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="d">
+							<greatestDifference id="d">d–d</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="Gy">
+							<greatestDifference id="G">y – y</greatestDifference>
```
pedberg-icu, line 2117: In this case the first year is BCE, consider adding G?

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="y">y–y</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="GyM">
+							<greatestDifference id="G">y-MM – y-MM</greatestDifference>
```
pedberg-icu, line 2121: In this case the first year is BCE, consider adding G?

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="M">y-MM – y-MM</greatestDifference>
+							<greatestDifference id="y">y-MM – y-MM</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="GyMd">
+							<greatestDifference id="d">y-MM-dd – y-MM-dd</greatestDifference>
```
pedberg-icu, line 2126: In this case the first year is BCE, consider adding G?

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+							<greatestDifference id="G">y-MM-dd – y-MM-dd</greatestDifference>
+							<greatestDifference id="M">y-MM-dd – y-MM-dd</greatestDifference>
+							<greatestDifference id="y">y-MM-dd – y-MM-dd</greatestDifference>
+						</intervalFormatItem>
+						<intervalFormatItem id="GyMEd">
+							<greatestDifference id="d">y-MM-dd, E – y-MM-dd, E</greatestDifference>
+							<greatestDifference id="G">y-MM-dd, E – GGGGG y-MM-dd, E</greatestDifference>
```
pedberg-icu, line 2133: In this case the second data is CE so remove GGGGG; but consider adding GGGGG for the first date which is BCE.

macchiati:
The tests might not succeed yet, but I'd appreciate review of the output (now with a column for en_CA).

https://docs.google.com/spreadsheets/d/1fIwEYN8AxoJruBuHuMCoer5WIkEBgFD8i1M2quacJiw/edit

AEApple:
I added a few questions, I wasn't sure the answers though.

macchiati:
> Interesting it looks like a lot of the differences in the date ranges may be in spacing. Do you want me to propose edits or wait for tomorrow since I wasn't sure how your are generating this list.

I fixed a batch of cases where the space character was caught by tests (normal vs narrow-nobreak).

AEApple:
I'm happy to approve since it seems like we just need to make a judgement call for any of the date patterns which are not covered in the ISO spec.

macchiati:
> ah okay! We could use en-CA as a reasonable proxy, for the most part en-CA is intentionally following ISO with a few exceptions.

I have en-CA in the spreadsheet for comparison. They match pretty well for the pure numeric stuff.

AEApple:
Okay, naive question, are any other locales inheriting from root, and if so will this mess up their dates?

macchiati:
What will happen is that the formats 

> Okay, naive question, are any other locales inheriting from root, and if so will this mess up their dates? It does look (at least from the surveytool that 'en' inherits a bunch of values from root.

Not naive at all. I had added the following to the release page:

- Added iso8601 patterns to root. These will use localized months, days of the week, day periods, and timezones. In this first version, the separators are not localized, and will use "-" within numeric dates, ":" within times, and " " or ", " between major elements. Full localization will await the next submission phase for CLDR.

common/main/root.xml
```diff
@@ -1936,6 +1936,313 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
+						<dateFormat>
+							<pattern>y-MM-dd</pattern>
+							<datetimeSkeleton>yMMdd</datetimeSkeleton>
+						</dateFormat>
+					</dateFormatLength>
+				</dateFormats>
+				<timeFormats>
+					<timeFormatLength type="full">
+						<timeFormat>
+							<pattern>h:mm:ss a zzzz</pattern>
```
pedberg-icu, line 1984: The standard patters should use HH per ISO format (this would also make them match the corresponding dateTimeSkeleton...)

macchiati: Fixed in https://github.com/unicode-org/cldr/pull/4037/commits/d954cf9c5e574f1aeeb2ef0fc9773cae39808666 (and spreadsheet).

pedberg-icu (commented):
This looks good to me now, thanks!

pedberg-icu (approved):

AEApple:
I figured out why the AM times confused me for the standard time format examples. I am used to 24 hr times always being zero padded, so when the morning time examples were not zero-padded I automatically read them as 12hr time even though they weren't.

macchiati:
All merged

PR #4044, opened 2024-09-11
CLDR-16526 spec: link check spec/rfc

srl295:
- run lychee action
- adds a summary output to the bottom of the workflow run, example https://github.com/unicode-org/cldr/actions/runs/10816554786#summary-30007966306

image

CLDR-16526

ALLOW_MANY_COMMITS=true


srl295:
https://github.com/unicode-org/cldr/actions/runs/10815398470/job/30004183744#step:11:90

Doesn't warn or error in the PR yet, but does do the run.

srl295:
Great!  

The errors now show up in the summary page for the run: https://github.com/unicode-org/cldr/actions/runs/10816554786

@AEApple @macchiati FYI, try the link above ^

macchiati:
Looks great. Is it everything in docs/ ?

srl295:
> Looks great. Is it everything in docs/ ?

This one is only docs/ldml and docs/rfc but I can set up similar for site.

macchiati:
great!

AEApple:
LGTM! So are we just going to file another ticket to clean up the initial batch of problematic links?

macchiati:
Yes, we'll need that

macchiati:
To confirm, this does not *block* PRs from being merged (we don't want that until we have time to clean them up).
Also, does this work for both the site and for the spec?

srl295:
> To confirm, this does not _block_ PRs from being merged (we don't want that until we have time to clean them up).

Correct. Otherwise this PR would show as blocked.

>  Also, does this work for both the site and for the spec?

No, the site is a separate build step.

macchiati (approved):

AEApple:
So does this just mean that there will be a list of errors after every time someone makes a spec change and it builds? Or is it every time there is a PR? (I wasn't sure if the spec is included in the build step every time a PR is created or only if the PR includes a spec change.)

srl295:
> So does this just mean that there will be a list of errors after every time someone makes a spec change and it builds? Or is it every time there is a PR? (I wasn't sure if the spec is included in the build step every time a PR is created or only if the PR includes a spec change.)

When a PR includes spec changes, you will see a list of errors on the PR.

srl295:
Pr needs update. But the link check part worked.

macchiati:
It'd be great to get this in soon, since it would catch many of our link problems. 

I just ran across one: in https://unicode.org/reports/tr35/tr35-general.html#pu_single_unit there is a link on [Private-Use Units](https://github.com/unicode-org/cldr/edit/main/docs/ldml/tr35-general.html#Private_Use_Units) that fails; should be omitting `https://github.com/unicode-org/cldr/edit/main/docs/ldml/`. Will this catch that?

srl295:
> It'd be great to get this in soon, since it would catch many of our link problems. 
> 
> 
> 
> I just ran across one: in https://unicode.org/reports/tr35/tr35-general.html#pu_single_unit there is a link on [Private-Use Units](https://github.com/unicode-org/cldr/edit/main/docs/ldml/tr35-general.html#Private_Use_Units) that fails; should be omitting `https://github.com/unicode-org/cldr/edit/main/docs/ldml/`. Will this catch that?

It should.

PR #4045, opened 2024-09-12
CLDR-17934 site: add a 404 page

srl295:
CLDR-17934

ALLOW_MANY_COMMITS=true


macchiati (approved):

macchiati:
And it's live: https://cldr.unicode.org/translation2 works (eg goes to 404
landing page)

PR #4051, opened 2024-09-13
CLDR-17930 Fix malformed hex escapes in common/transforms

lionel-rowe:
https://github.com/unicode-org/icu/pull/3143#issuecomment-2346804758

CLDR-17930

ALLOW_MANY_COMMITS=true


srl295 (dismissed):

srl295:
@lionel-rowe we'll need to accept the ticket, but also the test data will need an update

srl295:
@lionel-rowe Are you able to update the test data? if so we can put this in 46

PR #4053, opened 2024-09-13
CLDR-17948 kbd: bump version to v46

srl295:
CLDR-17948

- bump schema to support v45 or v46
- bump test files to v46 - note, these won't compile in keyman until keyman is updated
  - the test mechanism doesn't currently support multiple "old" keyboard versions. In the future, cldr-archive could be leveraged for this.
- change the ElementAttributeInfo logic to use a new 'special' keyboard XML file, with a DTD, so it can be parsed. This results in a 'no sample file' warning when testing against v45 since that file didn't exist in v45.  (But v45 and v46 schemas are identical save for the version number)

This will quell most of the warnings noted in CLDR-17574, CLDR-17745.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/keyboard.yml
+++ b/.github/workflows/keyboard.yml
@@ -34,7 +34,7 @@ jobs:
             ${{ runner.os }}-nodekbd-
             nodekbd-
       - name: Install kmc
-        run: npm install -g @keymanapp/kmc@alpha
+        run: npm install -g @keymanapp/kmc@beta
       - name: Compile Keyboards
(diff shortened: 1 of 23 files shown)


srl295:
- keyboard-compile will fail because the compiler has never seen CLDR v46 until this merges.

macchiati:
Steven, can you close this or shift to another ticket?

srl295:
> Steven, can you close this or shift to another ticket?

Obviated by #4118

PR #4054, opened 2024-09-16
CLDR-17947 site: pin image width, win fix

srl295:
CLDR-17947

- css: pin image width to 100% of the parent (the page), and also pin height to 90% of the viewable area.

also:
- fix builder for windows pathnames

## Testing
- before: https://cldr.unicode.org/index/language-support-levels
- after: https://b1085791.cldr.pages.dev/index/language-support-levels

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/assets/css/page.css
+++ b/docs/site/assets/css/page.css
@@ -180,3 +180,8 @@ header#atViewHeader {
 div > header {
   display: none !important;
 }
+
+img {
+  max-width: 100%;
+  max-height: 90dvh;
+}
--- a/docs/site/assets/js/build.mjs
+++ b/docs/site/assets/js/build.mjs
@@ -31,6 +31,7 @@ const coll = new Intl.Collator(["und"]);
 async function processFile(d, fullPath, out) {
   const f = await fs.readFile(fullPath, "utf-8");
   const m = matter(f);
+  fullPath = fullPath.replace(/\\/g, '/'); // backslash with slash, for win
   if (m && m.data) {
     const { data } = m;
     out.all.push({ ...data, fullPath });


docs/site/assets/css/page.css
```diff
@@ -180,3 +180,8 @@ header#atViewHeader {
 div > header {
   display: none !important;
 }
+
+img {
+  max-width: 100%;
```
macchiati, line 185: Sounds like we should use svh & svw

macchiati: https://dev.to/frehner/css-vh-dvh-lvh-svh-and-vw-units-27k4

macchiati (changes requested):

macchiati:
That is handy:
https://b1085791.cldr.pages.dev/index/language-support-levels

srl295:
> Small enough change and I know the context so I'm happy to accept.
> 
> For future PRs -- when you post PRs can you include screenshots and links to the before/after? It will help future CLDR developers have context for the fix.
> 
> From slack here's an example page that is helped by this fix: https://cldr.unicode.org/index/language-support-levels

added to the top

docs/site/assets/css/page.css
srl295: @macchiati what are you seeing instead? When I used `dvh/dvw` from your original proposal, the scale didn't work properly- it was far too wide still.

srl295:
```css
img {
  max-width: 100svw;
  max-height: 90svh;
}
```

gives this which looks wrong to me (still a huge image, twice as wide as the page):

image

-----

compare to the current PR version: https://b1085791.cldr.pages.dev/index/language-support-levels#language-selection-ui

image


Do you prefer the former?

macchiati:
That is very strange. Let's go with what you have; we can check out alternatives later.

macchiati (approved):

macchiati:
Just tried it out; looks great.

PR #4057, opened 2024-09-16
CLDR-17947 Add stub pages to site

macchiati:
CLDR-17947

This is a set of mostly stub pages, meant for cleaning up the site map as per [CLDR site organization](https://docs.google.com/spreadsheets/d/1MCOP6LYLmmn6ny_tt-gWJ6-XPgJGyNTkztlwjGpb91c/edit).

We can clean up and flesh out these pages over time.

Note: I added a few missing pages with just links to the old site, until we get them converted.

ALLOW_MANY_COMMITS=true


macchiati:
I realize that this is under the site improvement ticket for CSS. I realized afterwards that there is a somewhat better ticket number, but I'd prefer not to have to redo the PR.

AEApple (approved):
LGTM, I have minor edits that I would propose to some of these, but I think that is probably better done as we update pages.

macchiati:
Sorry, I was adding some pages during the time you were reviewing. Done now, if you could stamp.

These are the missing pages with just links to the old site.

AEApple (approved):

AEApple:
LGTM! Let's clean up the pages in a separate ticket since we should probably update some of these a bit.

Do you think we can add a freshness date going forward so it's easy to know the last time a page was fully reviewed vs. just minorly edited?

macchiati:
I agree that it would be handy to have that info. Let's discuss the best way to do that in the infra wg

PR #4064, opened 2024-09-19
CLDR-11567 Add population counts for El Salvadoran Languages

conradarcturus:
See https://unicode-org.atlassian.net/browse/CLDR-11567 and https://translatorswithoutborders.org/language-data-for-el-salvador

Re-generated the xml with `mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags`

CLDR-11567

ALLOW_MANY_COMMITS=true


conradarcturus:
I'm not sure what is causing the problem with `TestShim`, do you know what's causing it?

conradarcturus:
I checked with Steven, the error is coming from a timezone update introduced in an earlier PR unrelated to population census work, Etc/unknown. It's not significant for this PR and can be safely ignored.

macchiati (approved):

PR #4067, opened 2024-09-20
CLDR-17947 site: css: fix breakage

srl295:
merge conflict broke css for img

compare the images https://cldr.unicode.org/index/keyboard-workgroup to https://c6075b6b.cldr.pages.dev/index/keyboard-workgroup

CLDR-17947

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/assets/css/page.css
+++ b/docs/site/assets/css/page.css
@@ -201,6 +201,7 @@ section.body h5 {
 section.body h6 {
   font-size: 1em;
   font-style: italic;
+}
 img {
   max-width: 100%;
   max-height: 90dvh;


mcdurdin (approved):

macchiati (approved):

macchiati:
If the image width was the problem, it now appears solved:
https://cldr.unicode.org/index/language-support-levels#language-selection-ui

macchiati:
Note that the h1, h2, ... styles are still incorrect.

Eg, https://cldr.unicode.org/downloads/cldr-46#current-levels

This is an h4 element, but italic. All of them down to h5 should be bold
and non-italic. It appears that our h4 is not overriding the font-style
correctly; all of *our* site styles should have

font-weight: bold;
font-style: normal;

except for h6, which will have

font-weight: bold;
font-style: italic;

srl295:
> Note that the h1, h2, ... styles are still incorrect. Eg, https://cldr.unicode.org/downloads/cldr-46#current-levels This is an h4 element, but italic. All of them down to h5 should be bold and non-italic. It appears that our h4 is not overriding the font-style correctly; all of *our* site styles should have font-weight: bold; font-style: normal; except for h6, which will have font-weight: bold; font-style: italic;
 
this is just the PR for the broken curly brace. I wasn't aware something was incorrect.

macchiati:
Moved my comment to https://unicode-org.atlassian.net/browse/CLDR-17947?focusedCommentId=177549

PR #4073, opened 2024-09-24
CLDR-17770 Clarify what the alias element path value is relative to

pedberg-icu:
CLDR-17770

Clarify what the alias element path value is relative to (the path to the parent of the alias element, not the path of the alias element itself).

ALLOW_MANY_COMMITS=true


richgillam (approved):
Looks reasonable to me.

PR #4077, opened 2024-09-25
CLDR-13651 Add spec text involving compact long currency format

sffc:
CLDR-13651

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-numbers.md
+++ b/docs/ldml/tr35-numbers.md
@@ -407,6 +407,8 @@ Formatting 1200 in USD would result in “1.2 K $”, while 990 implicitly maps 
 
 The short format is designed for UI environments where space is at a premium, and should ideally result in a formatted string no more than about 6 em wide (with no fractional digits).
 
+When `currencyFormatLength` is not included for a desired format length (short or long), the currency should be formatted without compact notation and then substituted as the numeric value into the corresponding `decimalFormatLength`, adjusting for the number of 0s. For example, using the above data, 12000 USD in compact long format would result in “12 $ mille”.
+
 #### <a name="Currency_Formats" href="#Currency_Formats">Currency Formats</a>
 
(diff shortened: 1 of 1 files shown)


sffc:
@macchiati The proposed algorithm doesn't seem to work correctly for the example data in the spec.

docs/ldml/tr35-numbers.md
```diff
@@ -407,6 +407,8 @@ Formatting 1200 in USD would result in “1.2 K $”, while 990 implicitly maps
 
 The short format is designed for UI environments where space is at a premium, and should ideally result in a formatted string no more than about 6 em wide (with no fractional digits).
 
+When `currencyFormatLength` is not included for a desired format length (short or long), the currency should be formatted without compact notation and then substituted as the numeric value into the corresponding `decimalFormatLength`, adjusting for the number of 0s. For example, using the above data, 12000 USD in compact long format would result in “12 $ mille”.
```
sffc, line 410: I assume that the correct localized string should be "12 mille $" not "12 $ mille"

macchiati (changes requested):
I think we need to take this back to Design

docs/ldml/tr35-numbers.md
sffc: No, that's not the algorithm I understood from the ticket and our CLDR-WG discussion. The algorithm I understood was:

- otherwise:
    - Adjust the numeric value according to compact currency rules. For example, 20034.56 EUR becomes 20.03456c3, and after rounding becomes 20c3
    - Format the mantissa using `<currencyFormat type="standard">`, as in "20€"
    - Substitute that value into the correct compact currency form (the one corresponding to "c3"), as in "20€ mille"

However, maybe your algorithm is better; at least it works in French, and my algorithm doesn't :)

PR #4080, opened 2024-09-25
CLDR-17982 Add permanent redirect from original location

AEApple:
CLDR-17982

Add back page as redirect for anyone who had the original page linked.

ALLOW_MANY_COMMITS=true


srl295 (approved):
Impossible to redirect from here since it's outside of the site. So this can be a permanent page.

srl295:
Brs ticket should be accept-able

PR #4084, opened 2024-09-26
CLDR-17982 Add 46 beta news item

AEApple:
CLDR-17982 Add 46 beta news item

ALLOW_MANY_COMMITS=true


macchiati (approved):
We usually replace the alpha but it's ok

AEApple:
Yeah, I thought about that, but then I thought people might want to know. I was planning to remove both alpha and beta when we get to the final release.

PR #4091, opened 2024-09-30
CLDR-17977 Update _redirects

srl295:
CLDR-17977

> It appears that [https://cldr.unicode.org/index/cldr-spec/unicode-transliteration-guidelines](https://cldr.unicode.org/index/cldr-spec/unicode-transliteration-guidelines#h.wh84w0csexl6) is another link that used to work but doesn’t any more.

## Testing:

- https://6a4424c6.cldr.pages.dev/index/cldr-spec/unicode-transliteration-guidelines#h.wh84w0csexl6
- 
ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/_redirects
+++ b/docs/site/_redirects
@@ -7,3 +7,6 @@
 
 # redirect old download pages [CLDR-17977]
 /index/downloads/* /downloads/:splat
+
+# pages that were renamed [CLDR-17977]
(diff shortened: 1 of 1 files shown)


srl295:
looks like the `unicode-` was dropped in the conversion process

macchiati (approved):

PR #4104, opened 2024-10-02
CLDR-18002 Update population and likely subtags for MU, TK, ZM and SL

conradarcturus:
There are 4 manual overrides in GenerateLikelySubtags.java that conflict with other data: for MU, SL, TK, and ZM.

For each country, the local language (mfe, kri, tkl, and bem) is spoken by far more than English, even if English is the main language of instruction. Education and literacy in each country is low enough that the local languages should be considered the dominant ones.

I was able to find censuses listing language characteristics for MU, TK and ZM. SL I wasn't able to find data, but I removed the override.

To regenerate data use this command ` mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateTestData`

CLDR-18002

ALLOW_MANY_COMMITS=true


macchiati:
I'm generally ok with this, but for the following:

> Education and literacy in each country is low enough that the local languages should be considered the dominant ones.

I think that would clearly be the case for voice UIs. Not so clear for text UIs. We should discuss more about how to cleanly segment those. For example, we might have a separate set of likely subtags / locale matching for voice than for text.

conradarcturus:
> we might have a separate set of likely subtags / locale matching for voice than for text.

@macchiati  I think that's a great idea to differentiate likely subtags for voice & text content. We can also perhaps make a policy using with macrolanguages, eg. Arabic, Chinese, Fulah dialects. For the most part, most text would be best classified as just zh/ar/ff. However spoken content will have significant differences. Do we have any initiatives getting CLDR/Unicode to work better for spoken content?

macchiati:
Inflections and RBNF play a role, but no organized initiatives yet. We have made room for separate tagging, eg
```		<languageMatches type="written_new">
```
(intending to allow for a a 'voice' in the future.)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateLikelySubtags.java
```diff
@@ -433,15 +433,11 @@ public static void main(String[] args) throws IOException {
                                 {"und_Latn_PH", "fil_Latn_PH"},
                                 {"und_ML", "bm_Latn_ML"},
                                 {"und_Latn_ML", "bm_Latn_ML"},
-                                {"und_MU", "mfe_Latn_MU"},
```
macchiati, line 436: The more we can delete from this list, the better!

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateLikelySubtags.java
conradarcturus: There are 4 manual overrides in GenerateLikelySubtags.java that conflict with other data: for MU, SL, TK, and ZM.

In each of these cases, while the spoken language in these countries is a very local language, English or French is used the most in information technology devices so that should always be the default.

PR #4105, opened 2024-10-02
CLDR-18002 Fix likely subtag inconsistency

macchiati:
CLDR-18002

- Add internal test to the generation to check for additive consistency
- Add 4 override items (Conrad will remove those and the related overrides in v47, after fixing the population figures)
- Regenerate likely subtags with the fixes
- Add one unrelated fix (for tools/cldr-code/src/main/java/org/unicode/cldr/tool/ChartLocaleGrowth.java), just a version number update to a BRS tool.

ALLOW_MANY_COMMITS=true


macchiati:
PS. regenerating test data now.

conradarcturus (approved):
Thanks for adding the fixes. I'll follow up with more data

PR #4108, opened 2024-10-04
CLDR-18006 site: fix h-levels

srl295:
- also, run prettier on html and css
- merge reports-v2.css into page.css
- add prettier to package.json so it can be launched with 'npm run pretty'

originally filed at CLDR-17947

CLDR-18006

## Testing notes:

I don't have a test page checked in but see for example the CLDR 46 release page

Example test:

image

ALLOW_MANY_COMMITS=true


macchiati (approved):
The change in file name and reformatting make this hard to review, but the results look good.

srl295:
yes, but the formatting should keep it stable in the future.

PR #4111, opened 2024-10-07
CLDR-18008 Define currency digits as being derived from ISO 4217

sffc:
CLDR-18008

Alternative A. Compare to Alternative B: #4112


macchiati:
See ticket comments.

PR #4112, opened 2024-10-07
CLDR-18008 Specify that currency digits might NOT be derived from ISO 4217

sffc:
CLDR-18008

Alternative B. Compare to Alternative A: #4111

### DIFF
--- a/docs/ldml/tr35-numbers.md
+++ b/docs/ldml/tr35-numbers.md
@@ -946,7 +946,10 @@ Each `currencyData` element contains one `fractions` element followed by one or 
 The `fractions` element contains any number of `info` elements, with the following attributes:
 
 * **iso4217:** the ISO 4217 code for the currency in question. If a particular currency does not occur in the fractions list, then it is given the defaults listed for the next two attributes.
-* **digits:** the minimum and maximum number of decimal digits normally formatted. The default is 2. For example, in the en_US locale with the default value of 2 digits, the value 1 USD would format as "$1.00", and the value 1.123 USD would format as → "$1.12".
+* **digits:** the minimum and maximum number of decimal digits normally formatted. 
+The default is 2. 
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-numbers.md
```diff
@@ -946,7 +946,7 @@ Each `currencyData` element contains one `fractions` element followed by one or
 The `fractions` element contains any number of `info` elements, with the following attributes:
 
 * **iso4217:** the ISO 4217 code for the currency in question. If a particular currency does not occur in the fractions list, then it is given the defaults listed for the next two attributes.
-* **digits:** the minimum and maximum number of decimal digits normally formatted. The default is 2. For example, in the en_US locale with the default value of 2 digits, the value 1 USD would format as "$1.00", and the value 1.123 USD would format as → "$1.12".
+* **digits:** the minimum and maximum number of decimal digits normally formatted. The default is 2. For example, in the en_US locale with the default value of 2 digits, the value 1 USD would format as "$1.00", and the value 1.123 USD would format as → "$1.12". This value of this field may or may not equal the "minor unit" value from ISO 4217.
```
macchiati, line 949:
```suggestion
* **digits:** the minimum and maximum number of decimal digits normally formatted. 
The default is 2. 
For example, in the en_US locale with the default value of 2 digits, the value 1 USD would format as "$1.00", and the value 1.123 USD would format as → "$1.12". 
This value of this field is based on the "minor unit" value from ISO 4217, but may deviate from ISO 4217 where there is compelling evidence for different customary practice.
```

macchiati: Reviewed in the design meeting. Note that the **usage** of currencies in countries may also deviate from ISO, based on customary practice.

macchiati (changes requested):

macchiati (approved):

sffc:
I re-opened the issue because the PR hadn't been merged yet.

PR #4114, opened 2024-10-07
CLDR-8038 Fix link to Who Uses CLDR

btangmu:
-New link ID is #who-uses-cldr

CLDR-8038

ALLOW_MANY_COMMITS=true


docs/site/index.md
```diff
@@ -25,7 +25,7 @@ The Unicode Common Locale Data Repository (CLDR) provides key building blocks fo
 
 CLDR uses the XML format provided by [UTS #35: Unicode Locale Data Markup Language (LDML)](http://www.unicode.org/reports/tr35/). LDML is a format used not only for CLDR, but also for general interchange of locale data, such as in Microsoft's .NET.
 
-## Who uses CLDR?
+## Who uses CLDR? {#who-uses-cldr}
```
srl295, line 28:
```suggestion
## Who uses CLDR?
```

the anchor fragment will be `#who-uses-cldr` automatically

As written here, the anchor fragment is `#who-uses-cldr-who-uses-cldr`

srl295: it seems to work in production as written (learn something new!) but it's still not needed.

Github parses it as <https://github.com/unicode-org/cldr/blob/e50a74bb88d2e0ef33378591e5acb0cca988e8e9/docs/site/index.md#who-uses-cldr-who-uses-cldr>

btangmu: It seems to work correctly, so I'm suspicious whether the anchor fragment is really `#who-uses-cldr-who-uses-cldr`. Anyway, though, if `## Who uses CLDR?` automatically acts as if it were followed by `{#who-uses-cldr}` and if it's considered preferable to use the automatic link instead of an explicit one, I'll make your suggested change. Strange that the documentation gives the example `[Heading IDs](#heading-ids)` and doesn't seem to mention automatic links.

srl295: @btangmu it works correctly on the site, but not in Github.

>  I'm suspicious whether the anchor fragment is really #who-uses-cldr-who-uses-cldr.

it is when viewed via github, see the link my my above comment. 

> Strange that the documentation gives the example [Heading IDs](#heading-ids) and doesn't seem to mention automatic/implicit anchors

Yeah that's very strange.

I would go with https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links instead, we follow Github Flavored Markdown more closely.

srl295 (approved):

macchiati:
BTW, Tom, I find it much easier to user the Web view to make spec and site
PRs

PR #4116, opened 2024-10-07
CLDR-17807 minor edits

DavidLRowe:
CLDR-17807

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -2393,7 +2393,7 @@ Used in the `to=`
 
     - The `from=` and `to=` sides of the pattern must both be using `set` variables. There is no way to insert a set literal on either side and avoid using a variable.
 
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-keyboards.md
```diff
@@ -1825,7 +1825,7 @@ These may be then used in multiple contexts:
 <key id="acute_key" output="${acute}" />
 …
 <!-- Display ´ instead of the non-displayable marker -->
-<display output="${acute}" display="${backquote}" />
+<display output="${grave}" display="${backquote}" />
```
srl295, line 1828: this will fail because there's no `<string id="grave" …` defined. Would it work to change 'acute' everywhere else in this example to grave?

DavidLRowe: Okay. Changed other 'acute' occurences to 'grave'.

srl295 (approved):
looks good now thanks

PR #4122, opened 2024-10-09
CLDR-17893 Fix rg validation example

macchiati:
CLDR-17893

Fix for Frank's comment in https://unicode-org.atlassian.net/browse/CLDR-17893?focusedCommentId=177844

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):
Looks good, and looks like it addresses Frank's comment.

PR #4123, opened 2024-10-09
CLDR-18001 site: improve subpage and ToC 

srl295:
CLDR-18001

- move arrow into `&nbsp;»` and off right edge
- indent ToC entries according to their level

ALLOW_MANY_COMMITS=true

## Review Notes

observe moved arrow and indents

image

### DIFF
--- a/docs/site/assets/css/page.css
+++ b/docs/site/assets/css/page.css
@@ -772,6 +772,10 @@ body.page header .subpages a:hover {
   text-decoration: underline;
 }
 
+body.page .subpages a.hasChildren:after {
+  content: " »";
+}
+
 .navHeader {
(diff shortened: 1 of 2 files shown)


macchiati (approved):
Looks great; I think the sidebar is good for 46 release!

I have one other thing, but lower priority, so your other tickets should take priority (especially cleanup of the top bar & removal of "This navigation UI is temporary..." 
Should I open another ticket, where we can accumulate other lower priority site sidebar feedback?
- We can skip the H1 level; that is the title and it is right there in the page. That also saves some horizontal space.

srl295:
Maybe a new subtask of the umbrella

PR #4124, opened 2024-10-09
CLDR-18006 site: updates to top bar

srl295:
- drop Site Map from header
- drop "temporary navigation" notice

CLDR-18006

ALLOW_MANY_COMMITS=true

## Review Notes

- check the main and site map pages' top nav bar
- still doesn't line up 100%

image

### DIFF
--- a/docs/site/_layouts/page.html
+++ b/docs/site/_layouts/page.html
@@ -14,11 +14,6 @@
           <!-- Vue mount here -->
         </div>
       </div>
-      <div class="message">
-        <i
-          >This navigation UI is temporary, just to give access to the pages.</i
-        >
-      </div>
     </header>
 
     <section class="body">
--- a/docs/site/_layouts/sitemap.html
+++ b/docs/site/_layouts/sitemap.html
@@ -14,12 +14,6 @@
           <!-- Vue mount here for normal navigation -->
         </div>
       </div>
-      <div class="message">
-        <i
-          >This navigation UI is temporary, just to give access to the pages.</i
-        >
-      </div>
-      <!-- <div class="bar"><a href="/sitemap" class="bar">Site Map</a></div> -->
     </header>
 
     <section class="body">{{ content }}</section>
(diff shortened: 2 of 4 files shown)


srl295:
@macchiati this means you can't get to the site map unless you're on the home page. Maybe that's OK?

macchiati:
They still aren't vertically centered. Look at the following:

```
<!DOCTYPE html>
<html>
<head>
<style>
.verticalCenter {
  display: flex;
  align-items: center;
}

</style>
</head>
<body>

<div class='verticalCenter'>
  <img src="https://cldr.unicode.org/assets/img/logo60s2.gif" style='margin-right:1em'><span>Unicode CLDR Project ❱ CLDR Technical Committee ❱ CLDR Keyboard Subcommittee</span>

</div>
</body>
</html>
```

In an online test, say, https://www.w3schools.com/cssref/tryit.php?filename=trycss_text-indent

Not complete, but you can see the idea. Narrow the window, and the alignment still works.

macchiati:
As to the sitemap, for now we can have it just in the first subpages. We can discuss later whether we need it on each page, and where to put it. (People generally look for it in a footer.)

srl295:
Yes, I specifically said that they were not vertically centered.

macchiati:
Right, I was just pointing out a simple method to vertically center them.

srl295:
image

give this a try

macchiati:
It is not as good as the flex approach when the breadcrumb exceeds the window width.
Screenshot 2024-10-10 at 09 17 02

However, it is good enough for v46 release, if you want to spin off a future ticket to take care of that.

macchiati (approved):
But open a ticket to refine (or add to existing ticket)

srl295:
> It is not as good as the flex approach when the breadcrumb exceeds the window width. Screenshot 2024-10-10 at 09 17 02
> 
> However, it is good enough for v46 release, if you want to spin off a future ticket to take care of that.

it's using flex, at least partly…

macchiati:
I opened a new subtask https://unicode-org.atlassian.net/browse/CLDR-18017 as you suggested.

PR #4126, opened 2024-10-10
CLDR-18001 site: fix link in ToC to entire page, fix some titles

srl295:
CLDR-18001

- suppress link to title in ToC, hide ToC if empty
- suppress title h1 entirely, add centered title
- shift ToC over to the left one

ALLOW_MANY_COMMITS=true

## Review Notes

- check links in ToC on index and subpages
- check the /development page as an example of a page that will not have a 'Content' sidebar in this change

image

image

### DIFF
--- a/docs/site/assets/css/page.css
+++ b/docs/site/assets/css/page.css
@@ -808,23 +808,21 @@ body.page .subpages a.hasChildren:after {
 }
 
 /* these are the ToC entries in the sidebar, according to what element they came from originally. */
-.pagecontents .headingH1 {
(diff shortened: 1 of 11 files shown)


macchiati (approved):
Looks good.

BTW
- We definitely need a page explaining the title requirements. Not part of GFM and not documented that quotes are required. (What happens if they aren't there??)
- I think we need to drop the first line of the contents; I'd added that to https://unicode-org.atlassian.net/browse/CLDR-18017. But that can be later.

macchiati (approved):

PR #4127, opened 2024-10-14
CLDR-18022 Fix site and spec links

macchiati:
CLDR-18022

Simple tooling to look for bad links in the source files of site and spec, where the format of the links is incorrect.

ALLOW_MANY_COMMITS=true


macchiati:
Still draft, but if you have any comments, that'd be welcome. The tooling is targeted at v47, though fixing the the spec would be for v46.

btangmu:
LGTM... is there a way to see the test output linked to this PR? I couldn't find it under "All checks have passed"...

macchiati:
The output is lengthy; I copied it into the spreadsheet linked in the meeting notes (infra). The assert statements just have the number of failures; if they fail, the failures just print to console. That can be cleaned up later.

macchiati:
This can definitely use some refinement/tweaking; since it is not hooked up in TestAll, however, it will not cause failures.

So the question is whether to get the first version in, and then refine, or to wait.

PR #4129, opened 2024-10-14
CLDR-18022 fix spec links

macchiati:
CLDR-18022

- Fixed links as per ticket, except for message-format.md (will be done separately)
- Also added some missing links to the EBNF for units.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -1477,7 +1477,7 @@ nostr "no:n"
 The generated yesexpr and noexpr would be:
 
 ```
-yesexpr "^([yY]​([eE][sS])?)"
+yesexpr "^([yY]([eE][sS])?)"
 ```
 
 This would match y,Y,yes,yeS,yEs,yES,Yes,YeS,YEs,YES.
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -592,11 +592,11 @@ The ordering of the characters in the set is irrelevant, but for readability in 
 
 ### ~~<a name="Character_Mapping" href="#Character_Mapping">Mapping</a>~~
 
-**This element has been deprecated.** For information on its structure and how it was intended to specify locale-specific preferred encodings for various purposes (e-mail, web), see the [Mapping](https://www.unicode.org/reports/tr35/tr35-39/tr35-general.html#Character_Mapping) section from the CLDR 27 version of the LDML Specification.
+**This element has been deprecated.** For information on its structure and how it was intended to specify locale-specific preferred encodings for various purposes (e-mail, web), see the [Mapping](tr35-general.html#Character_Mapping) section from the CLDR 27 version of the LDML Specification.
 
 ### ~~<a name="IndexLabels" href="#IndexLabels">Index Labels</a>~~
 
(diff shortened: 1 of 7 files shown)


docs/ldml/tr35-general.md
```diff
@@ -1477,7 +1477,7 @@ nostr "no:n"
 The generated yesexpr and noexpr would be:
 
 ```
-yesexpr "^([yY]([eE][sS])?)"
+yesexpr "^([yY] ([eE][sS])?)"
```
pedberg-icu, line 1480: Is the added space intentional?

docs/ldml/tr35-keyboards.md
```diff
@@ -2229,7 +2229,7 @@ _Attribute:_ `from` (required)
 
 - **Nested capturing groups**
 
-    `(?:[abc]([def]))|(?:[ghi])`
+    `(?:[abc] ([def]))|(?:[ghi])`
```
pedberg-icu, line 2232: Is the added space intentional?

pedberg-icu (approved):
I had a couple of questions about space in regexes, but approving to unblock things in case those are not issues.

docs/ldml/tr35-general.md
macchiati: Yes. We can't otherwise easily tell that this isn't a link in markdown, eg

[anchor](relative-link)

Could use a ZWSP so the appearance is better for people used to regex with significant spaces. What do you think?

macchiati:
We have a couple of days before this needs to go in, so we have time to discuss the space issue.

macchiati:
Went ahead and fixed those cases, so that we don't get false positives, and doesn't mislead people.

docs/ldml/tr35-keyboards.md
srl295: it looks like it was interpreted as a markdown link!

srl295 (changes requested):
no, please don't put a U+200B in the middle of the regex, that doesn't make sense… if the link checker is giving a false positive, please fix the tool…

docs/ldml/tr35-general.md
srl295: it's inside a code fence - triple backquotes.  A markdown parser can tell the difference.

docs/ldml/tr35-keyboards.md
```diff
@@ -2229,7 +2229,7 @@ _Attribute:_ `from` (required)
 
 - **Nested capturing groups**
 
-    `(?:[abc]([def]))|(?:[ghi])`
+    `(?:[abc]​([def]))|(?:[ghi])`
```
srl295, line 2232:
```suggestion
    `(?:[abc]([def]))|(?:[ghi])`
```

This is a sample, people will want to copy and paste it.

Another approach you can take for the link checker is to check the output HTML - then check the resolved `href=`

docs/ldml/tr35-general.md
```diff
@@ -1477,7 +1477,7 @@ nostr "no:n"
 The generated yesexpr and noexpr would be:
 
 ```
-yesexpr "^([yY]([eE][sS])?)"
+yesexpr "^([yY]​([eE][sS])?)"
```
srl295, line 1480:
```suggestion
yesexpr "^([yY]([eE][sS])?)"
```

don't change this one either.

srl295 (changes requested):
rest looks fine, there's no reason to damage regexes to workaround a tool problem.

macchiati:
Ok, retracting the 200B's; will handle the exceptions in tooling.

srl295 (approved):

srl295:
> Ok, retracting the 200B's; will handle the exceptions in tooling.

thanks.

Should be able to use a multiline regex to look for ```

PR #4130, opened 2024-10-15
CLDR-18023 site: Update keyboard-workgroup.md

srl295:
- update links to keyboard spec and current status

CLDR-18023

ALLOW_MANY_COMMITS=true

## Review notes
- Preview the site under https://2f163433.cldr.pages.dev/index/keyboard-workgroup or later

### DIFF
--- a/docs/site/index/keyboard-workgroup.md
+++ b/docs/site/index/keyboard-workgroup.md
@@ -61,13 +61,17 @@ The CLDR Keyboard Subcommittee was formed and has been meeting since mid-2020. I
 
 ### What is the current status?
 
+Discussion
+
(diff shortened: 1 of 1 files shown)


miloush (approved):

mcdurdin (approved):

macchiati:
Please make the images much smaller. Both can be a quarter of the height/width. (And there is no strong reason for the <row keys one to be an image instead of text.)

srl295:
> Please make the images much smaller. Both can be a quarter of the height/width. (And there is no strong reason for the <row keys one to be an image instead of text.)

I copied your feedback, thanks, to the ticket

PR #4132, opened 2024-10-16
CLDR-18026 Update commit-checker.yml

srl295:
CLDR-18026

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/commit-checker.yml
+++ b/.github/workflows/commit-checker.yml
@@ -25,7 +25,7 @@ on:
 jobs:
   build:
     name: Run Checker
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


srl295:
@AEApple it seems to work with this workaround at https://cldr-smoke.unicode.org/cldrcc/CLDR-Report-2024-10-16-v46-9d1845dee3.html

For now if you want to run the CC you need to set the 'workflow branch' to `srl295/cldr-18026/fix-commit-checker` - once this is merged the workflow branch will need to be `main` - there's no reaason to merge this to maint-46 because we can just use the `main` branch for the workflow in the future.

AEApple (approved):

PR #4136, opened 2024-10-16
CLDR-18032 Tooling for bad urls

macchiati:
CLDR-18032

Separated off part of a ticket that needed to go to main instead of maint-46.

While this needs further refinement, I'd like to get the basic version into main.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestURLs.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestURLs.java
@@ -1,0 +1,419 @@
+package org.unicode.cldr.unittest;
+
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
+import com.google.common.collect.TreeMultimap;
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestURLs.java
```diff
@@ -0,0 +1,419 @@
...
+        // For debugging
+        //        if (count == 0) {
+        //            int start = line.indexOf("href");
+        //            System.out.println(RegexUtilities.showMismatch(url, line.subSequence(start,
+        // line.length())));
+        //            int d=0;
+        //        }
+    }
+
+    public static boolean doesURLExist(URL url) {
```
srl295, line 396: use the `HttpStatusCache` instead, persists.

but, i'd really rather not rewrite an external link checker here, it would be better to use lychee which already deals with this situation.   See https://github.com/unicode-org/cldr/pull/4044 which link checks the spec/rfc - it's only blocked because it would fail until the links were cleaned up.

srl295 (approved):
OK for now- I don't think it's the best approach (partially parsing markdown with regex, awkward to use within a unit test)  long-term, but ok for now.

PR #4138, opened 2024-10-16
CLDR-15821 v47 add Analytics to spec

srl295:
- insert gtag.html after <HEAD>

I'll have to cherry pick this to the maint branch

CLDR-15821

ALLOW_MANY_COMMITS=true


macchiati (approved):

srl295:
@macchiati when the deploy goes through to smoke, you should see the analytics tags up on https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html

Do you have a preferred ticket for cherry-picking this back to v46 maint?

macchiati:
https://unicode-org.atlassian.net/browse/CLDR-18031 would be fine

PR #4146, opened 2024-10-21
CLDR-18031 v46 BRS: fix spec ToC / endings

srl295:
- part of running the ToC updater
- the ToCs were all fine, but it did catch these two issues.

CLDR-18031 

ALLOW_MANY_COMMITS=true


srl295:
sorry for the spam, definitely hit the wrong button there at first.

pedberg-icu (approved):

PR #4148, opened 2024-10-21
CLDR-17948 v46 dtd: conformsTo needs to include current CLDR version

srl295:
- also fix compiler version to pickup latest stable

CLDR-17948

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

srl295:
@macchiati already merged but thanks

PR #4152, opened 2024-10-21
CLDR-18043 Fix some transform issues

robertbastian:
CLDR-18043

ALLOW_MANY_COMMITS=true


common/transforms/Han-Spacedhan.xml
```diff
@@ -37,7 +37,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 〜→ '~';
 
 $terminalPunct = [\.\,\:\;\?\!．，：？！｡、；[:Pe:][:Pf:]];
-$initialPunct = [:Ps:][:Pi:];
+$initialPunct = [[:Ps:][:Pi:]];
```
macchiati, line 40: Items like this are great to fix. However, why are you titlecasing the properties; the matching is case-insensitive/

robertbastian: It's nice to have a consistent style, and ICU4X currently doesn't support case insensitive property names (it does support case-insensitive *property values*).

robertbastian: There's only a [strong recommendation](https://www.unicode.org/reports/tr18/#Categories:~:text=It%20is%20strongly%20recommended%20that%20both%20names%20be%20recognized%2C%20and%20that%20loose%20matching%20of%20property%20names%20and%20values%20be%20implemented%20following%20the%20guidelines%20in%20Section%205.9%20Matching%20Rules%20in%20%5BUAX44%5D.) to be able to loosely parse property names, so ideally we wouldn't uses unnormalized names in CLDR.

srl295: This doesn't make as much sense as a one-time update, if it's CLDR policy then there should be a tool that checks/corrects. 

Fixable in ICU4X? or update in icu4x during data intake?

srl295: maybe it would make sense to split this work into two - fixes vs. normalization issue

robertbastian: I agree that this should be enforced with tests, but I don't think that's blocking for 46, while this is something I'd like to get in.

srl295: @robertbastian you've targetted `main` instead of `maint-46`, so this PR will land in 47 not 46.

robertbastian: I plan to cherry-pick this and #4153. I find it confusing to land changes in maintenance branches and cherry-pick to main.

srl295: Generally we don't cherry pick from main.  if you can make the case this *must* be in 46 (I'm working on some of the final release tasks right now!) the ticket should target 46.

Is there a reason icu4x can't be fixed?

robertbastian: > Generally we don't cherry pick from main.

What's the process for backporting a fix then?

ICU4X can be fixed, CLDR can also be fixed.

srl295: > > Generally we don't cherry pick from main.
> What's the process for backporting a fix then?

What I mean is that if the issue is planned to be fixed in 46, the issue (the two tickets) should start with targeting 46.  

> ICU4X can be fixed, CLDR can also be fixed.

Not clear CLDR is broken (as to the casing).  I think that should be revisited post release.

robertbastian: The first two commits are bug fixes, and if we're changing transforms already, we might as well clean up the casing.

macchiati (approved):

common/transforms/Han-Spacedhan.xml
markusicu: @srl295 -- @macchiati approved these changes. You had comments that don't seem to be resolved, but you didn't block this PR. Would you mind if I merged this PR as is?

srl295: No objections

macchiati:
None here. Longer term, I am concerned that having ICU for ex only
recognize the normalized property names and values could cause
compatibility issues, that does not präventive these changes.

macchiati:
Robert, do you want us to merge for you (if you don't have permissions?)

robertbastian:
Yes please, I don't have permissions.

PR #4069, opened 2024-09-21
CLDR-9791 South Georgian and Bouvet Island languages

conradarcturus:
CLDR-9791

These two island communities don't have a permanent population but they have a top-level internet domain and they usually are inhabited by people speaking English or Norwegian.

- [x] This PR completes the ticket.

Scripts ran: `mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags`

ALLOW_MANY_COMMITS=true

PR #4166, opened 2024-10-29
CLDR-9791 Add language for South Georgia and Bouvet Islands

conradarcturus:
https://github.com/unicode-org/cldr/pull/4069 got erased when merging v47 back to main so I'm creating a new pull request.

These two island communities don't have a permanent population but they have a top-level internet domain and they usually are inhabited by people speaking English or Norwegian.

Need to also Adding en_GS.xml that inherits from en_001 since en_GS is an official language technically

CLDR-9791

ALLOW_MANY_COMMITS=true


common/testData/localeIdentifiers/localeDisplayName.txt
```diff
@@ -1310,6 +1310,32 @@ nl-Latn-BE; flamšćina (łaćonsce)
 zh-Hans-fonipa; chinšćina [zjednorjena] (FONIPA)
 
 
+@locale=ht
```
conradarcturus, line 1313: The Haitian Creole data here is added because it wasn't added when the locale xml was added. It was created by the automated script that I run to rebuild files that inherit from the population data.

`mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateTestData`

srl295: It's fine, of course test data is generated as part of the BRS cycle, it does not strictly need to be added when the locale was added.

srl295 (approved):

PR #4179, opened 2024-11-02
CLDR-15948 Clean up well-formedness and/or validity constraints

macchiati:
CLDR-15948

docs/ldml/tr35-general.md
- Added a clearer anchor (the old anchor remains valid)
- Having the unit_identifier ABNF in a table is too painful to edit so I converted to markdown.
- The constraints were turned into either well-formedness constraints or validity constraints.
- The substantive change was to add the following to the definition of `base_component`
    - [ vc:  Must be an attribute value of the `source` in: \<convertUnit source='…' …\> or the `type` in \<unitAlias type="…" replacement="…" …\> ]
- Fixed typos from ticket

docs/ldml/tr35-info.md
- moved "The determination of preferred units..." up a section, and fleshed it out to provide a full explanation (with sources) for unit identifiers and unit usage identifiers of their well-formedness and validity.

docs/ldml/tr35.md
- Reword the text for currency identifiers to clarify what the well-formedness and validity are.
- Added notes to the modifications section.
   - There isn't one for currency identifiers; when clarifications or copy-edits are very simple, we typically don't note them. 

We should also provide validity files for these, but this should be sufficient for the immediate future.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -68,7 +68,7 @@ The LDML specification is divided into the following parts:
   * [Unit Preference and Conversion Data](#Unit_Preference_and_Conversion)
   * [Unit Identifiers](#Unit_Identifiers)
     * [Nomenclature](#nomenclature)
-    * [Syntax](#syntax)
+    * [Unit Syntax](#unit-syntax)
   * [Unit Identifier Uniqueness](#Unit_Identifier_Uniqueness)
   * [Example Units](#Example_Units)
   * [Compound Units](#compound-units)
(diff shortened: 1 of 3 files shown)


macchiati:
It is still draft, but early comments are welcome.

docs/ldml/tr35-general.md
```diff
@@ -902,157 +902,160 @@ As with other identifiers in CLDR, the American English spelling is used for uni
...
-        </ul>
-	</td></tr>
-
-</tbody></table>
+<a name="syntax"></a>
+#### Unit Syntax
+
+The formal syntax for identifiers is provided below, in [EBNF](tr35.md#ebnf). 
+Some of the constraints reference data from various elements in the unit conversion data [units.xml](https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml). 
+These may be either element values or element attribute values.
```
aphillips, line 910: Do you try to avoid "fleeting mentions" of 2119 keywords (`may` in this case)?

docs/ldml/tr35-general.md
```diff
@@ -902,157 +902,160 @@ As with other identifiers in CLDR, the American English spelling is used for uni
...
+    * See also [Prefixes for binary multiples](https://physics.nist.gov/cuu/Units/binary.html)
+
+<a name='prefix_component' href='#prefix_component'>prefix_component</a> 
+<br/>:= [a-z]{3,∞}
+* [ vc:  must be value in: \<unitIdComponent type="prefix"\>. ]
+* *Notes:*
+    * The set of prefix components often expands in new releases, so the requirement to be one of these attribute values is a validity constraint, not a well-formedness constraint. *
+
+<a name='base_component' href='#base_component'>base_component</a> 
+<br/>:= [a-z]{3,∞}
```
aphillips, line 1020: infinity symbol? Usually this is `*`? I'm not au courant on your EBNF syntax... (there are instances before/after this one)

docs/ldml/tr35.md
```diff
@@ -1030,7 +1030,7 @@ The BCP 47 form for keys and types is the canonical form, and recommended. Other
 <tr><td>"cu"<br>(currency)</td>
     <td>Currency type</td>
     <td><i>ISO 4217 code,</i><p><i>plus others in common use</i></p></td>
-    <td><p>Codes consisting of 3 ASCII letters that are or have been valid in ISO 4217, plus certain additional codes that are or have been in common use. The list of countries and time periods associated with each currency value is available in <a href="tr35-numbers.md#Supplemental_Currency_Data">Supplemental Currency Data</a>, plus the default number of decimals.</p><p>The XXX code is given a broader interpretation as <i>Unknown or Invalid Currency</i>.</p></td></tr>
+    <td><p>Well-formed codes are of the form <code>[A-Za-z]{3}</code>, with the canonical format being <code>[A-Z]{3}</code>. The valid codes are or have been valid in ISO 4217, plus certain additional codes that are or have been in common use. The list of countries and time periods associated with each currency value is available in <a href="tr35-numbers.md#Supplemental_Currency_Data">Supplemental Currency Data</a>, plus the default number of decimals.</p><p>The XXX code is given a broader interpretation than in ISO 4217, as <i>Unknown or Invalid Currency</i>.</p></td></tr>
```
aphillips, line 1033: Line breaks make this easier to comment on.

The valid code are or have been => "The valid codes are ones that are or have been"

The "plus the default number of decimals" feels oddly tacked on. I would reverse this sentence to be something like:

> [Supplemental Currency Data](tr35-numbers.md#Supplemental_Currency_Data) provides the list of 
> countries/regions and time periods associated with each currency code.
> It also provides the default number of decimal places.

Note "currency/region", to avoid sensitivity over calling e.g. `HK` (or similar) a country (not to mention some currencies are region codes, such as `EUR` or some of the African basket ones)

docs/ldml/tr35.md
```diff
@@ -4308,7 +4308,12 @@ Other contributors to CLDR are listed on the [CLDR Project Page](https://www.uni
 
 **Differences from LDML Version 46**
 
-TBD
+### Unit Modifications
+- Updated the ABNF in [Unit Syntax](https://cldr-smoke.unicode.org/spec/main/ldml/tr35-general.html#syntax) to:
```
aphillips, line 4312: ABNF or EBNF?

docs/ldml/tr35.md
```diff
@@ -4308,7 +4308,12 @@ Other contributors to CLDR are listed on the [CLDR Project Page](https://www.uni
...
+    - Change the constraints into either well-formedness constraints or validity constraints.
+    - Add validity constraints for base-component.
+    - Reformat the ABNF to avoid using HTML tables.
```
aphillips, line 4315: ibid

aphillips (commented):
Some minor comments/questions

docs/ldml/tr35-general.md
macchiati: We haven't adopted use of the uppercase 2119 keywords in most Unicode standards. That's something we could consider..

docs/ldml/tr35.md
macchiati: Whoops, fixing

docs/ldml/tr35-general.md
macchiati: fixed to follow standard PRE — like {3,}, including other instances.

docs/ldml/tr35.md
macchiati: Fixed these, thanks.

As to linebreaks. One of our tasks is to finish the conversion to markdown, which is mostly fixing the tables that didn't translate, and then move towards Semantic linebreaks.

macchiati:
I think this is now ready to go: Addison proposed some additional wording, etc., which I made changes for.

aphillips (approved):
LGTM

AEApple:
Is there an easier way to preview the spec in markdown? the table doesn't look right in the rich diff preview in GtHub

macchiati:
It's no longer a table (markdown isn't strong enough)

sffc:
I'm not sure how to review this; I don't have a personal opinion on whether the grammars are better in the old format or the new format. I assume you are changing the format because it is better for MF2. I can do a sanity check that the grammar behavior is consistent across the refactor?

AEApple (approved):

macchiati:
I'm changing it more because it is far easier to maintain and review if it is in plain-ish markdown (including simple items like <br>). We didn't change the tables when we did the initial markdown, but basically been changing them as we have to touch various areas. It is now a bit easier given that we can paste a table into google docs, then Copy as markdown and fix up. But tables with merged cells or interior structure (like a list in a cell) still need reworking. 

The biggest one to fix will be https://www.unicode.org/reports/tr35/tr35.html#Key_Type_Definitions.

docs/ldml/tr35-general.md
```diff
@@ -902,157 +902,160 @@ As with other identifiers in CLDR, the American English spelling is used for uni
...
-			<li><em>Note:</em> CLDR does not provide conversions for currencies; this is only intended for formatting.
-				The locale data for currencies is supplied in the <code>currencies</code> element, not in the <code>units</code> element.</li>
-        </ul>
-	</td></tr>
-
-</tbody></table>
+<a name="syntax"></a>
+#### Unit Syntax
+
+The formal syntax for identifiers is provided below, in [EBNF](tr35.md#ebnf).
```
pedberg-icu, line 908: "provided below" does not make sense if the link is to a different document.

macchiati: I looked at this. The link on EBNF is a pointer to the documentation on our EBNF, which is applicable to what is below. But the wording is clumsy.

I'll make a PR to change it to 

The formal [EBNF](https://github.com/unicode-org/cldr/blob/main/docs/ldml/tr35.md#ebnf) syntax for identifiers is provided below:

PR #4183, opened 2024-11-07
CLDR-18074 Add specific-non-location -> specific location fallback

robertbastian:
CLDR-18074

Also CLDR-18084

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -1696,6 +1696,8 @@ Note: A generic location format is constructed by a part of time zone ID represe
 * "Pacific Daylight Time" (long)
 * "PDT" (short)
 
+**Specific location format:** This format does not have a symbol, but is used in the fallback chain for the _specific non-location format_. Like the _generic location format_ it uses time zone locations, but formats these in a zone-variant aware way, e.g. "France Summer Time".
+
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-dates.md
```diff
@@ -1877,15 +1879,19 @@ Some of the examples are drawn from real data, while others are for illustration
...
       * Examples:
-        * Europe/Rome → IT → "Italy Time" // for English
-        * Asia/Shanghai → CN → "China Time" // Asia/Shanghai is the _primaryZone_ for China
-        * Africa/Monrovia → LR → "Hora de Liberja"
-        * America/Havana → CU → "Hora de CU" // if CU is not localized
-   2. Otherwise format the exemplar city with the _regionFormat_, and return it.
-      1. America/Buenos_Aires → "Buenos Aires Time"
+        * America/Buenos_Aires → "Buenos Aires Time" // multiple zones in AR
+        * Asia/Shanghai → "China Time" // Asia/Shanghai is the _primaryZone_ for CN
+        * Asia/Shanghai → "China Summer Time" // daylight
```
macchiati, line 1891: This all looks good, with one caveat. In the examples, it looks like Asia/Shanghai can arbitrarily produce two different values. I think there is an implicit additional parameter, which we can add to make it clear what is happening. (Might also reformat as a table in that case, for clarity, and to have the comment in a separate column.)

robertbastian: done

macchiati (approved):

PR #4184, opened 2024-11-07
CLDR-18075 timezone names

robertbastian:
CLDR-18075

Also CLDR-18076, CLDR-18077, CLDR-18078. 

* Add a short name for territory `CC`, so that `Cocos (Keeling) Islands Time` (location format) becomes `Cocos Islands Time`, matching the non-location format
* Add a short name for territory `PN`, so that `Pitcairn Islands Time` (location format) becomes `Pitcairn Time`, matching the non-location format
* Add exemplar cities for some locations, so that the location format matches the non-location format
  * `Pacific/Easter`, so that `Easter Time` -> `Easter Island Time` (`en.xml`)
  * `Pacific/Wake`, so that `Wake Time` -> `Wake Island Time` (`en.xml`)
  * `America/Noronha`, so that `Noronha Time` -> `Fernando de Noronha Time` (`root.xml`)
* Change some non-location names that use cities when the location names use regions, as they are the only zone in a region (also to align with location format)
  * `Pyongyang Time` -> `North Korea Time`
  * `Taipei Time` -> `Taiwan Time` 
* Change the non-location name `Petropavlovsk-Kamchatski Time` to `Kamchatka Time`, as `Petropavlovsk-Kamchatski` is a city, but the location format uses `Kamchatka`
* Change some metazones names that only barely differ from the location names to match
  * `Brunei Darussalam Time` -> `Brunei Time`
  * `East Timor Time` -> `Timor-Leste Time`
  * `North Mariana Islands Time` -> `Northern Mariana Islands Time`
  * `Ponape Time` -> `Pohnpei Time`

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -996,6 +996,7 @@ annotations.
 			<territory type="BZ">Belize</territory>
 			<territory type="CA">Canada</territory>
 			<territory type="CC">Cocos (Keeling) Islands</territory>
+			<territory type="CC" alt="short">Cocos Islands</territory>
 			<territory type="CD">Congo - Kinshasa</territory>
 			<territory type="CD" alt="variant">Congo (DRC)</territory>
 			<territory type="CF">Central African Republic</territory>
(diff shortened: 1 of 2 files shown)


conradarcturus (approved):
Thanks for making the timezone updates

conradarcturus (approved):

PR #4185, opened 2024-11-07
CLDR-17730 Indonesian RBNF is missing a semicolon for spellout-ordinal

grhoten:
CLDR-17730

ALLOW_MANY_COMMITS=true


conradarcturus (approved):
Change is straightforward, lgtm code-wise. However we need to wait until next Wednesday for this to be reviewed by the CLDR TC so the Jira ticket can be accepted

PR #4187, opened 2024-11-07
CLDR-17025 zz-default numbering system rule is less precise than the other numbering systems

grhoten:
CLDR-17025

I confirmed that this change resolves the error that appears on the [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp?type=NumberingSystemRules).

ALLOW_MANY_COMMITS=true


conradarcturus (approved):
Great :D

PR #4188, opened 2024-11-07
CLDR-17386 Fix Russian fractional number spellout and fix grammatical case names

grhoten:
CLDR-17386

This changes 2 related issues.

1. The fractional cardinal number is always feminine.
2. The grammatical case names in the rules were chosen incorrectly. Fix them to the correct names, and alias the old names to the correct names to maintain backwards compatibility. The instrumental and prepositional forms are already correctly listed in [grammaticalFeatures.xml](https://github.com/unicode-org/cldr/blob/main/common/supplemental/grammaticalFeatures.xml)

ALLOW_MANY_COMMITS=true


conradarcturus (approved):
Thanks for making these updates

PR #4189, opened 2024-11-09
CLDR-18092 Clarify Venetian language in Mexico

conradarcturus:
There's a community of Mexicans that speak Venetian - but it is definitely not an offical language for the country, region, or city. This updates the data to remove this error.

CLDR-18092

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -2349,7 +2349,7 @@ XXX Code for transations where no currency is involved
 		<language type="ve" scripts="Latn"/>
 		<language type="ve" territories="ZA" alt="secondary"/>
 		<language type="vec" scripts="Latn"/>
-		<language type="vec" territories="BR HR IT MX SI" alt="secondary"/>
+		<language type="vec" territories="BR HR IT SI" alt="secondary"/>
 		<language type="vep" scripts="Latn"/>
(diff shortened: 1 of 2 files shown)


btangmu (approved):
LGTM; I wonder where the inaccurate data came from to begin with

srl295 (changes requested):
hold on, let me check. 
I put this in from very specific sources

srl295:
OK, i missed the comment on  https://unicode-org.atlassian.net/browse/CLDR-15943?focusedCommentId=166689 

still - should it be unofficial regional instead of removed?

conradarcturus:
> still - should it be unofficial regional instead of removed?

There's no unofficial_regional status.

There is:
* de_facto_official
* official
* official_regional
* recognized

I plan to add "recognized_regional". The difference between official and recognized is if it is used in the proceedings of government (official) and languages with government language support, eg. publication of laws, minority status (recognized).

Right now Venetian is NOT government recognized in any form in Mexico. While some people want it to be -- its not accurate today. Furthermore there are 68 indigenous languages recognized by the government so it is a bit silly to elevate Venetian here but ignore those. There's a separate ticket to add those recognized languages. https://unicode-org.atlassian.net/browse/CLDR-18094

PR #4191, opened 2024-11-11
CLDR-17818 Fix Bulgarian cardinal and ordinal spellings in the 100-199 range

grhoten:
CLDR-17818

This uses the following references as the right way to spell the cardinals and ordinals in Bulgarian. Some of the missing rule variants (e.g. other gendered forms) were extrapolated to use the same structure.

https://bg.wikipedia.org/wiki/21_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)
https://bg.wikipedia.org/wiki/150_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)
https://bg.wikipedia.org/wiki/151_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)
https://bg.wikipedia.org/wiki/200_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)
https://bg.wikipedia.org/wiki/201_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)
https://bg.wikipedia.org/wiki/221_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)
https://bg.wikipedia.org/wiki/400_(%D1%87%D0%B8%D1%81%D0%BB%D0%BE)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/rbnf/bg.xml
+++ b/common/rbnf/bg.xml
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
 <!--
-Copyright © 1991-2022 Unicode, Inc.
+Copyright © 1991-2024 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
 -->
 <ldml>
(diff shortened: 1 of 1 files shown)


conradarcturus (approved):
I trust you.

Is there a tool to view all of the generated numbers from this? It would be great to have a non-machine output that can be verified by actual Bulgarians.

But that's a wish-list idea for the time none of us have. I'll love it if it happens, but no need to slow down progress here.

grhoten:
The only way to see the results are to use the [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp) without the XML markup.

PR #4200, opened 2024-11-15
CLDR-18110 Improve Lithuanian RBNF

grhoten:
CLDR-18110

These changes don't complete the Lithuanian rules, but they're better than before. It takes a lot of effort to write the set of rules for this language. These rules add all of the grammatical cases, genders and grammatical numbers for cardinals and ordinals.

The ordinals don't include the pronomial forms, and only the positive degree is included. There is room to improve the parsing of the large ordinals beyond 100, since some of them don't roundtrip. At least the cardinal rules parse.  The rules are tedious to write without some way to have a rule that is a combination of both the prefix and suffix.

Note to future self or other reviewers, please review the following links for further helpful information on writing cardinal and ordinal numbers in Lithuanian.

https://en.wikipedia.org/wiki/Lithuanian_declension#Numbers
https://en.wiktionary.org/wiki/vienas#Lithuanian
https://lt.wiktionary.org/wiki/vienas
https://en.wiktionary.org/wiki/pirmas#Lithuanian
https://lt.wiktionary.org/wiki/pirmas
https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#lt

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/rbnf/lt.xml
+++ b/common/rbnf/lt.xml
@@ -1,9 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
 <!--
-Copyright © 1991-2013 Unicode, Inc.
-CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+Copyright © 1991-2024 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-DFS-2016
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
 -->
 <ldml>
     <identity>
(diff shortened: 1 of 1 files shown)


grhoten:
Please note that the grammatical case names is like the other RBNF names.

conradarcturus (approved):
Accepting to unblock.

Are there tests that run automatically that make sure this continues to function? Are there charts that show the current & proposed behavior?

grhoten:
To see the current behavior, I recommend using the Number Format Tester.  As you can see, the [current Lithuanian behavior](https://st.unicode.org/cldr-apps/numbers.jsp?locale=lt) is lacking a lot. If you copy and paste the rules attached to CLDR-18110 into the Number Format Tester, you get to see the new behavior.  I have to convert the ICU rules to XML with the RBNFWriter class in CLDR.

I generally haven't been writing tests in CLDR for RBNF, since the testing has been primarily done with Number Format Tester.  Some existing tests are in ICU too.  The current situation around testing is rather hard because the current XML based rules are hard to test as is. If CLDR-10884 was implemented, then adding tests would be much easier. This was briefly mentioned on the CLDR mailing list recently. It's easy to make the change, the concern remains about what might break by making the change, like the conversion to JSON, and who is really using the RBNF rules outside of ICU.

PR #4201, opened 2024-11-15
CLDR-18111 Add Gujarati RBNF

grhoten:
CLDR-18111

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/rbnf/gu.xml
+++ b/common/rbnf/gu.xml
@@ -1,0 +1,180 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!--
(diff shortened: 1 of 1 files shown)


conradarcturus (approved):
Thanks for mentioning [the Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp) in PR #4200 . Can you help out future reviewers by copy-pasting the before and after screenshots of that tool with/without the changed rules?

Nonetheless, I don't want to slow you down so I'm accepting this change so we can add it to the code. Thanks for the thorough changes! Bother me if you need other reviews.

grhoten:
FYI the ticket has the original rules to use in the Number Format Tester. You can copy and paste those rules in to see how they look. That's how I test them, but I'll add screenshots in the future too.

PR #4202, opened 2024-11-17
CLDR-15948 Fix wording in tr35-general.md

macchiati:
CLDR-15948

Fix wording as noted in https://github.com/unicode-org/cldr/pull/4179#pullrequestreview-2437621281

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):
Ah, this is much more clear, thanks

PR #4203, opened 2024-11-18
CLDR-18017 Update site map ordering

AEApple:
CLDR-18017

Update site map to re-order the releases/downloads page and requesting changes page higher in the nav bar on the CLDR home page

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/sitemap.tsv
+++ b/docs/site/sitemap.tsv
@@ -7,7 +7,6 @@
 #	If an item has 'sub items' it becomes a directory parent.		
 #	There must be a single 'index' entry here at the root.		
 index			
-	requesting_changes		
 	index/downloads		
 		downloads/cldr-47	
 		downloads/cldr-46	
@@ -40,6 +39,8 @@ index
 		index/cldr-spec/definitions	
 		index/bcp47-extension	
 	index/charts		
+	requesting_changes		
+	index/acknowledgments		
 	general-information		
 		index/language-support-levels	
 		index/locale-coverage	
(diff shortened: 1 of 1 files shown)


docs/site/sitemap.tsv
```diff
@@ -7,15 +7,7 @@
...
-		index/cldr-spec/core-data-for-new-locales	
-		index/cldr-spec/coverage-levels	
-		index/cldr-spec/collation-guidelines	
-		index/cldr-spec/picking-the-right-language-code	
-		index/cldr-spec/plural-rules	
-		index/cldr-spec/transliteration-guidelines	
-		index/cldr-spec/definitions	
-		index/bcp47-extension	
+	requesting_changes		
 	index/downloads
```
macchiati, line 11: I think it would be better to have index/downloads before requesting changes

AEApple: I kind of expect most people to follow the news item to the latest release instead of the nav bar but I could be wrong.

Okay, so this ordering?
- CLDR Releases
- CLDR Spec
- CLDR Charts
- Requesting changes
- Acknowledgements
- General Information
- Information Hub for Linguists
- CLDR Technical Committee
- Schedule
- Internal Development (I'm not sure why we have this from the front page instead of a sub section of the TC)
- Archived Pages (I'm not sure why we have this from the front page, but I'm not sure where it would fit)
- CLDR Site Map

macchiati: I think that works.

On schedule, we might put that right below General info

- For internal development, it is handy for us to have it at the top level
- For archived pages, it's hard to know where else to put it. They are mostly (but I think not all) internal development.

Note: we could also have Steven extend the site map tooling to allow for the following, if we want more flexibility.
- Having pages in multiple locations
- Hiding pages (we want to make that explicit, so we don't inadvertantly skip a page).

AEApple: SG! Let me put together an updated commit for review.

macchiati (approved):

AEApple:
Merging the change. If anyone has any follow up feedback we can always adjust the order if needed.

PR #4204, opened 2024-11-18
CLDR-18065 Make link clickable

AEApple:
CLDR-18065

Make link to project in Jira clickable

ALLOW_MANY_COMMITS=true


conradarcturus (approved):
Yea... it feels silly to do this but I don't know if we can enable the markdown parser to automatically do this for us.

PR #4205, opened 2024-11-19
CLDR-15948 Add additional well-formedness clauses

macchiati:
CLDR-15948

We need better well-formedness clauses for identifiers, for references. This doesn't make material changes, since (with some work) someone can derived that the various identifiers have those restrictions. Also added links to some of the EBNF for locale ids, for reference, and reversed the aliasing for keyword, key, type to be the more understandable (and parallel to -t-) terms.

Removed the quotes around the identifiers in the large table, since that was confusing to readers, replacing with `code` style.

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):
Lots of good cleanup, thanks. Nice to distinguish well-formed from valid.

macchiati:
I just realized that we should change the version to 46.1 (draft) until
46.1 releases, so that
https://unicode.org/reports/tr35/dev/tr35.html#Contents works correctly.

PR #4208, opened 2024-11-20
CLDR-15391 Update Canadian census and Fix Cans Script Match

conradarcturus:
The original purpose of this change is to update the default language for Canadian Aboriginal syllabics [Cans] from Inuktitut [iu] to Cree [cr] since Cree has a larger population. Understandably, both of these languages are macrolanguages with many variations -- so its funny to include both the Cree macro-language along with its constituents. But I think it's better to cover both groupings because depending on the consumer they may want [cr] data or constituent data.

While I was doing this I updated all of the Canadian locale data to the 2021 Census. I also added a few missing aboriginal Canadian languages: Woods Cree [cwd] and Western Ojibway [ojw].

See the 2021 Census table here: https://www150.statcan.gc.ca/t1/tbl1/en/tv.action?pid=9810021601

CLDR-15391

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv
```diff
@@ -177,65 +177,67 @@ Cameroon	CM	"25,640,965"	71%	"89,540,000,000"		Ngomba	jgo	"87,500"	30%		http://w
...
+Canada	CA	"36,328,480"	99%	"1,774,000,000,000"		Mi'kmaq	mic	0.0254%			https://www150.statcan.gc.ca/t1/tbl1/en/tv.action?pid=9810021601 Canada 2021 Census language "Knowledge of Language"; official status from Wikipedia Languages_of_Canada
+Canada	CA	"36,328,480"	99%	"1,774,000,000,000"	recognized	Atikamekw	atj	0.0187%			https://www150.statcan.gc.ca/t1/tbl1/en/tv.action?pid=9810021601 Canada 2021 Census language "Knowledge of Language"; official status from Wikipedia Languages_of_Canada
+Canada	CA	"36,328,480"	99%	"1,774,000,000,000"		Siksika	bla	0.0183%			https://www150.statcan.gc.ca/t1/tbl1/en/tv.action?pid=9810021601 Canada 2021 Census language "Knowledge of Language"; official status from Wikipedia Languages_of_Canada
+Canada	CA	"36,328,480"	99%	"1,774,000,000,000"		Woods Cree	cwd	0.0140%			https://www150.statcan.gc.ca/t1/tbl1/en/tv.action?pid=9810021601 Canada 2021 Census language "Knowledge of Language"; official status from Wikipedia Languages_of_Canada
```
macchiati, line 223: The unit tests should prevent aliased language codes (like cwd). We need to fix that, but in the meantime you should use cr instead of cwd.

common/supplemental/supplementalData.xml
```diff
@@ -1546,7 +1546,7 @@ XXX Code for transations where no currency is involved
 		<language type="enm" scripts="Latn" alt="secondary"/>
 		<language type="eo" scripts="Latn"/>
 		<language type="es" scripts="Latn" territories="AR BO CL CO CR CU DO EA EC ES GQ GT HN IC MX NI PA PE PR PY SV UY VE"/>
-		<language type="es" territories="AD BZ DE FR GB GI PH PT RO US" alt="secondary"/>
+		<language type="es" territories="AD BZ CA DE FR GB GI PH PT RO US" alt="secondary"/>
```
macchiati, line 1549: As we discussed, we need to deprecate the languageData element and restructure the tools to deprecate (and stop updating!) the data there. Remind me, do we have a ticket for doing that?

macchiati (changes requested):

conradarcturus:
@srl295  ah this was a larger can of worms than I anticipated.

So it looks like currently `cr` defaults to Woods Cree/`cwd`. But, really, `cr` is the Macrolanguage for all Cree variations https://iso639-3.sil.org/code/cwd

The original reported bug is that "und_Cans" is matching to "iu_Cans" not "cr_Cans" even though the Cree community is bigger than the Inuktitut community. However I need to figure out the macrolanguage matching to see hwo to move forward.


Happy to punt this and get back to this once we are all back in late November.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv
conradarcturus: Indeed, this change needs work.

I'd like to understand the macrolanguage better. As I've learned, Woods Cree `cwd` is being aliased to Cree `cr`. However, Woods Cree (5,110 speakers according to CA 2021 Census) isn't even the biggest Cree dialect -- Plains Creek `crk` is (12,005 speakers) -- however the vast majority of Cree speakers aren't grouped (there are 87,875 Cree speakers in total. 61,000 are not matched to a constituent dialect). So perhaps Woods Cree is actually the biggest.

So back to aliasing -- personally I'm not a big fan of that because they are grouping oranges with apples -- yes they are all fruits but they are different groups. Do you have any design docs about aliasing so I can understand the background?

For another example of aliasing losing data precision, when I worked on [CLDR-10478](https://unicode-org.atlassian.net/browse/CLDR-10478), the Macao census has different estimates for Traditional Chinese (written, any spoken dialect `zh_Hant`, at 98%), Simplified Chinese (written, any spoken dialect `zh_Hans`, at 5%) as well as Spoken Cantonese `yue` @ 86.2% and Spoken Mandarin `cmn` @ ~40%. I ended up just ignoring the Mandarin category because was being aliased back to `zh`.

Nonetheless, I have faith this has been thought-through a lot already, so maybe I just need to catch up to the design choices.

macchiati:
The handling of the 'macrolanguage' concept was introduced in BCP47 for backwards compatibility, but causes its own — more severe — compatibility problems. If 'zh' truly means 'any Chinese language', then it would be perfectly fine for an implementation to request 'zh' and for us to serve up 'yue' content in LDML. So we have a longstanding policy for macro/encompassed languages that I outlined. See also https://cldr.unicode.org/index/cldr-spec/picking-the-right-language-code.

There are times where we will adjust the aliasing, where there is a strong shift one way or another. But for small changes we favor stability. The key is, just treat 'cr' exactly as if it were 'cwd', and don't use 'cwd'.

conradarcturus:
Ooof that makes sense and I don't want this ticket to generate any more work than it already has so I am inclined to punt on the macrolanguage conversation. To return to the narrow origin of the ticket of 1) making `und_Cans` match to `cr_Cans` instead of `iu_Cans`. I'll still keep the official language + population updates though.

common/supplemental/supplementalData.xml
conradarcturus: Yes, https://unicode-org.atlassian.net/browse/CLDR-18087

Until that ticket is out of design I'll still regenerate <languageInfo> to keep it current.

common/supplemental/supplementalData.xml
```diff
@@ -1475,12 +1475,12 @@ XXX Code for transations where no currency is involved
 		<language type="cop" scripts="Arab Copt Grek" territories="EG" alt="secondary"/>
 		<language type="cps" scripts="Latn"/>
 		<language type="cr" scripts="Cans Latn"/>
-		<language type="cr" territories="CA" alt="secondary"/>
 		<language type="crg" scripts="Latn"/>
 		<language type="crh" scripts="Cyrl"/>
 		<language type="crj" scripts="Cans"/>
 		<language type="crj" scripts="Latn" alt="secondary"/>
 		<language type="crk" scripts="Cans"/>
+		<language type="crk" territories="CA" alt="secondary"/>
```
conradarcturus, line 1483: This change happened because Woods Cree is not an official language of a region of Canada.

Rather, Plains Cree is. Cree is only official in the Northern Territories (NT). Unfortunately, [the NT law](https://web.archive.org/web/20090324202430/http://www.justice.gov.nt.ca/PDF/ACTS/Official_Languages.pdf) does not specify which Cree variation. We can deduce the variation because [the only Cree language present in NT is Plains Cree [crk]](https://en.wikipedia.org/wiki/Cree_language#Dialect_criteria) so I infer that is the correct match.

macchiati: Sounds good

conradarcturus:
Per feedback from the CLDR DDL Subcommittee: For the purposes of CLDR the `cr` language code is Woods Cree (ISO cwd). Thereby `cr_Cans_CA` should be considered Woods Cree specifically. Broadly, for the community, the Cree languages are not mutually intelligible so they don't want Cree to be lumped into 2 language anyway.

So I'm not going to complete the original ticket's purpose (sorry!) -- but I would still like this change to update the Canadian Population data to the latest census.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv
```diff
@@ -172,6 +172,8 @@ cs	Czech	primary	Latn	Latin
 csb	Kashubian	secondary	Latn	Latin
 csw	Swampy Cree	primary	Cans	Unified Canadian Aboriginal Syllabics
```
srl295, line 173: csw and probably others may be secondary latin (they call it SRO - standard roman orthography)

srl295 (approved):

PR #4210, opened 2024-11-20
CLDR-18108 Give all languages a primary script: trivial cases

conradarcturus:
This change adds "primary" scripts to many languages in language_script.tsv.

This won't change likely subtags, rather this just future-proofs our data by recognizing a singular primary script, avoiding issues where ambiguities served customers the wrong script.

I also added scripts for languages in country_language_population.tsv that were missing.

CLDR-18108

There are many more tasks, especially more complicated ones, in the parent ticket https://unicode-org.atlassian.net/browse/CLDR-18102

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -1315,7 +1315,9 @@ XXX Code for transations where no currency is involved
 		<language type="ann" scripts="Latn"/>
 		<language type="anp" scripts="Deva"/>
 		<language type="aoz" scripts="Latn"/>
+		<language type="apc" scripts="Arab"/>
 		<language type="apc" territories="IL JO LB PS SY TR" alt="secondary"/>
+		<language type="apd" scripts="Arab"/>
 		<language type="apd" territories="SD" alt="secondary"/>
 		<language type="ar" scripts="Arab" territories="AE BH DJ DZ EG EH ER IL IQ JO KM KW LB LY MA MR OM PS QA SA SD SO SY TD TN YE"/>
 		<language type="ar" scripts="Syrc" territories="IR SS" alt="secondary"/>
@@ -1397,6 +1399,7 @@ XXX Code for transations where no currency is involved
 		<language type="bjj" territories="IN" alt="secondary"/>
 		<language type="bjn" scripts="Latn"/>
 		<language type="bjn" territories="ID" alt="secondary"/>
+		<language type="bjt" scripts="Latn"/>
 		<language type="bjt" territories="SN" alt="secondary"/>
 		<language type="bkm" scripts="Latn"/>
 		<language type="bku" scripts="Latn"/>
@@ -1422,6 +1425,7 @@ XXX Code for transations where no currency is involved
 		<language type="brx" scripts="Deva"/>
 		<language type="brx" territories="IN" alt="secondary"/>
 		<language type="bs" scripts="Cyrl Latn" territories="BA"/>
+		<language type="bsc" scripts="Latn"/>
 		<language type="bsc" territories="SN" alt="secondary"/>
 		<language type="bss" scripts="Latn"/>
 		<language type="bto" scripts="Latn"/>
@@ -1445,6 +1449,7 @@ XXX Code for transations where no currency is involved
 		<language type="cay" scripts="Latn"/>
 		<language type="cch" scripts="Latn"/>
 		<language type="ccp" scripts="Beng Cakm"/>
+		<language type="ccr" scripts="Latn" alt="secondary"/>
 		<language type="ce" scripts="Cyrl"/>
 		<language type="ce" territories="RU" alt="secondary"/>
 		<language type="ceb" scripts="Latn"/>
@@ -1712,6 +1717,7 @@ XXX Code for transations where no currency is involved
 		<language type="ilo" territories="PH" alt="secondary"/>
 		<language type="inh" scripts="Cyrl"/>
 		<language type="inh" scripts="Arab Latn" territories="RU" alt="secondary"/>
+		<language type="io" scripts="Latn" alt="secondary"/>
 		<language type="is" scripts="Latn" territories="IS"/>
 		<language type="it" scripts="Latn" territories="CH IT SM VA"/>
 		<language type="it" territories="DE FR HR MT US" alt="secondary"/>
@@ -1721,6 +1727,7 @@ XXX Code for transations where no currency is involved
 		<language type="ja" scripts="Jpan" territories="JP"/>
 		<language type="jam" scripts="Latn"/>
 		<language type="jam" territories="JM" alt="secondary"/>
+		<language type="jbo" scripts="Latn" alt="secondary"/>
 		<language type="jgo" scripts="Latn"/>
 		<language type="jmc" scripts="Latn"/>
 		<language type="jml" scripts="Deva"/>
@@ -1749,6 +1756,7 @@ XXX Code for transations where no currency is involved
 		<language type="kdt" scripts="Thai"/>
 		<language type="kea" scripts="Latn"/>
 		<language type="kea" territories="CV" alt="secondary"/>
+		<language type="ken" scripts="Latn"/>
 		<language type="kfo" scripts="Latn"/>
 		<language type="kfr" scripts="Deva"/>
 		<language type="kfr" territories="IN" alt="secondary"/>
@@ -1786,6 +1794,7 @@ XXX Code for transations where no currency is involved
 		<language type="kmb" territories="AO" alt="secondary"/>
 		<language type="kn" scripts="Knda"/>
 		<language type="kn" territories="IN" alt="secondary"/>
+		<language type="knf" scripts="Latn"/>
 		<language type="knf" territories="SN" alt="secondary"/>
 		<language type="knn" scripts="Deva"/>
 		<language type="knn" territories="IN" alt="secondary"/>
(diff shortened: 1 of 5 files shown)


docs/site/development/updating-codes/update-language-script-info/language-script-description.md
```diff
@@ -4,19 +4,16 @@ title: Language Script Description
...
 
-1. If the language and script are both modern, and the script is a major way to write the language in some country, then we should see that line marked as **primary**.
-2. Otherwise it should be marked **secondary**.
-
-Every language that is in official use in any country according to country\_language\_population  should have at least one primary script in the language\_script spreadsheet.
+1. Every language needs at least 1 script considered the **primary** script.
+    1. This data is used to determine [the most Likely language and region](likelysubtags-and-default-content) so there needs to be at least 1 primary value.
+    2. [Changed in v47] Include a primary script for historical languages (eg. Ancient Greek, Coptic). The primary script should reflect where the majority of the written corpus originates from.
+2. Languages written by significant populations with different scritps in different countries can have multiple **primary** scripts. The [likely subtags](https://www.unicode.org/cldr/charts/latest/supplemental/likely_subtags.html) patterns will use population counts to disambiguate the default script for each locale.
```
macchiati, line 12: I disagree with this; I think it complicates our work. It is simplest if there is exactly 1 primary script, period, that will match the maximized value for the language. All others are not needed, because they will be derived from language population data. Eg if Kurdish has Latin as the primary script, but ku_Arab is the largest population (of ku*) speakers, then we will generate ku_IQ => ku_Arab_IQ in likely subtags.

Is there any other purpose to multiple primaries or having secondaries?

common/supplemental/supplementalData.xml
```diff
@@ -1712,6 +1717,7 @@ XXX Code for transations where no currency is involved
 		<language type="ilo" territories="PH" alt="secondary"/>
 		<language type="inh" scripts="Cyrl"/>
 		<language type="inh" scripts="Arab Latn" territories="RU" alt="secondary"/>
+		<language type="io" scripts="Latn" alt="secondary"/>
```
macchiati, line 1720: secondary with no primary, see comment below.

macchiati: Same comment for other new secondaries.

docs/site/development/updating-codes/update-language-script-info/language-script-description.md
conradarcturus: I'm fine with that to (only ever 1 primary). Do you think the CLDR TC agrees with that? I don't think I had much involvement.

I was trying to strike a compromise with the concerns people had about picking a primary language (that was mentioned in the DDL group). I'll bring it up again in the next DDL meeting.

common/supplemental/supplementalData.xml
conradarcturus: This is auto-generated. TBH I don't understand why the ConvertLanguageData script did this for this entry but not others -- I suspect because its missing territory data. I'll have to edit ConvertLanguageData.

macchiati:
Thanks for all your work on this... We are getting some long-needed
improvements!

common/supplemental/supplementalData.xml
conradarcturus: I updated the script so it will no longer demote historical scripts/historical languages.

macchiati:
I think we should put this on hold for now, and focus on https://unicode-org.atlassian.net/browse/CLDR-18087.

docs/site/development/updating-codes/update-language-script-info/language-script-description.md
srl295: I agree that we should simply pick a primary, _just as_ we need to pick a single likely subtag (i.e. `zh-Hans`).  I know there are concerns, but structurally we just need to pick something.

Always having exactly 1 primary makes the most sense structurally.

docs/site/development/updating-codes/update-language-script-info/language-script-description.md
```diff
@@ -4,19 +4,13 @@ title: Language Script Description
...
-
-This is also reflected in CLDR main: languages with multiple scripts will have that reflected in their structure (eg sr\-Cyrl\-RS), with aliases for the language\-region combinations.
-
-Files in https://github.com/unicode-org/cldr/tree/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data
-
-1. country\_language\_population.tsv
-2. language\_script.tsv
+Languages with multiple ambiguous scripts should have that reflected in their CLDR structure (eg. `sr_Cyrl_RS`), with aliases for the language\-region combinations.
 
+In order to re-generate the XML data use ConvertLanguageData as written about in [the article about updating the language scripts](.../update-language-script-info.md).
```
srl295, line 16:
```suggestion
In order to re-generate the XML data use ConvertLanguageData as written about in [the article about updating the language scripts](/development/updating-codes/update-language-script-info.md).
```

use site relative links (for now)

srl295 (commented):
LGTM but i see there's other discussion

conradarcturus:
> I think we should put this on hold for now, and focus on https://unicode-org.atlassian.net/browse/CLDR-18087.

I can hold off on this change -- but I want to close out easy tickets rather in parallel with working on the more complex ones. `<languageData>` is taking a bit of time to fix the problems when deprecating it so I don't know if I can fix that before the new year.

conradarcturus:
As discussed in the Unicode CLDR Design meeting, let's move ahead with the straightforward updates -- and work on the long-term changes in parallel. Many of the long-term changes require larger refactors so I think we should still merge this change so the data quality improves incrementally, rather than large / hard to review changes.

docs/site/development/updating-codes/update-language-script-info/language-script-description.md
```diff
@@ -4,19 +4,13 @@ title: Language Script Description
...
-
-This is also reflected in CLDR main: languages with multiple scripts will have that reflected in their structure (eg sr\-Cyrl\-RS), with aliases for the language\-region combinations.
-
-Files in https://github.com/unicode-org/cldr/tree/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data
-
-1. country\_language\_population.tsv
-2. language\_script.tsv
+Languages with multiple ambiguous scripts should have that reflected in their CLDR structure (eg. `sr_Cyrl_RS`), with aliases for the language\-region combinations.
 
+In order to re-generate the XML data use ConvertLanguageData as written about in [the article about updating the language scripts](/development/update-language-script-info.md).
```
srl295, line 16:
```suggestion
In order to re-generate the XML data use ConvertLanguageData as written about in [the article about updating the language scripts](/development/updating-codes/update-language-script-info).
```

docs/site/development/updating-codes/update-language-script-info/language-script-description.md
```diff
@@ -4,19 +4,13 @@ title: Language Script Description
...
 
-The language\_script spreadsheet should list all of the language / script combinations that are in common modern use. The countries are not important, since their function has been overtaken by the country\_language\_population spreadsheet.
+The [`language\_script.tsv`](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv) data file should list all of the language / script combinations that are in common use. Usage by country is indicated in the [`country\_language\_population.tsv`](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv) spreadsheet.
 
-1. If the language and script are both modern, and the script is a major way to write the language in some country, then we should see that line marked as **primary**.
-2. Otherwise it should be marked **secondary**.
+1. Every language should have 1 script considered the **primary** script.
+    1. This data is used to determine [the most Likely language and region](likelysubtags-and-default-content) so there needs to be at least 1 primary value.
```
srl295, line 10:
```suggestion
    1. This data is used to determine [the most Likely language and region](/development/updating-codes/likelysubtags-and-default-content) so there needs to be at least 1 primary value.
```

srl295 (changes requested):
broken links otherwise OK

We should make sure this goes in the release notes under migration

docs/site/development/updating-codes/update-language-script-info/language-script-description.md
```diff
@@ -7,7 +7,7 @@ title: Language Script Description
 The [`language\_script.tsv`](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv) data file should list all of the language / script combinations that are in common use. Usage by country is indicated in the [`country\_language\_population.tsv`](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv) spreadsheet.
 
 1. Every language should have 1 script considered the **primary** script.
-    1. This data is used to determine [the most Likely language and region](likelysubtags-and-default-content) so there needs to be at least 1 primary value.
+    1. This data is used to determine [the most Likely language and region](/development/updating-codes/likelysubtags-and-default-content) so there needs to be at least 1 primary value.
```
macchiati, line 10:
```suggestion
    1. This data is currently used to determine [the likely script for a language](/development/updating-codes/likelysubtags-and-default-content) so there needs to be at least 1 primary value. Because it is the default, it determines the script of locales without language codes in the `<territoryInfo>`. 
```

See also https://unicode-org.atlassian.net/browse/CLDR-18155?focusedCommentId=178748

macchiati (changes requested):

macchiati (approved):

PR #4212, opened 2024-11-21
CLDR-11562 Update spelling of October

AEApple:
CLDR-11562

Update spelling of October both wide and abbreviated

ALLOW_MANY_COMMITS=true


roozbehp (approved):

srl295 (approved):

mihnita:
Shouldn't this be in `cldr-staging`?
Otherwise this will never make it in ICU (which generates the .txt files from `cldr-staging`)
Or is there some integration happening (once in a while) from `cldr` to `cldr-staging`?

PR #4215, opened 2024-11-27
CLDR-18129 Investigate and fix (where necessary) invalid codes

macchiati:
CLDR-18129

common/dtd/ldml.dtd
- for @MATCH language@type, use nlocale (locales for names)

- all @MATCH changes:
- languageAlias@type, use bc47-wellformed (locales for names), and list allowed ill-formed values
- likelySubtag@from&to, use xlocale (which allows for compatibility cases like "iw"
- pluralRules@locales, use xlocale but also allow "sh"

common/main/en.xml
common/main/fi.xml
common/main/la.xml
common/main/nl.xml
common/supplemental/supplementalMetadata.xml
- Tighten down the test to avoid deprecated / predominant-encompassed languages.
- Add XLocaleMatchValue, NLocaleMatchValue to allow for special cases

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
- tighten up validity/locale to disallow deprecated
- allow for setting the status for each field separately
- provide for subclassing to extend the language check

tools/cldr-code/src/main/java/org/unicode/cldr/util/Validity.java
- Allow for testing validity against past versions

- clean up some code that didn't allow for old versions

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAttributeValues.java
- Enhance testing to print (warnln) details of which items caused failures.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBasic.java
- Add compatibility test for bcp47 to ensure that identifiers never disappear

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -61,7 +61,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT language ( #PCDATA ) >
 <!ATTLIST language type NMTOKEN #REQUIRED >
-    <!--@MATCH:validity/locale-->
+    <!--@MATCH:validity/nlocale-->
 <!ATTLIST language alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/long, secondary, short, variant, menu, official-->
 <!ATTLIST language draft (approved | contributed | provisional | unconfirmed | true | false) #IMPLIED >
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -152,7 +152,7 @@ annotations.
 			<language type="csw">Swampy Cree</language>
 			<language type="cu">Church Slavic</language>
 			<language type="cv">Chuvash</language>
-			<language type="cwd">Woods Cree</language>
+			<language type="cr">Cree</language>
 			<language type="cy">Welsh</language>
 			<language type="da">Danish</language>
 			<language type="dak">Dakota</language>
@@ -256,7 +256,7 @@ annotations.
 			<language type="hak">Hakka Chinese</language>
 			<language type="haw">Hawaiian</language>
 			<language type="hax">Southern Haida</language>
-			<language type="hdn">Northern Haida</language>
+			<language type="hai">Haida</language>
 			<language type="he">Hebrew</language>
 			<language type="hi">Hindi</language>
 			<language type="hi_Latn">Hindi (Latin)</language>
@@ -284,7 +284,7 @@ annotations.
 			<language type="ig">Igbo</language>
 			<language type="ii">Sichuan Yi</language>
 			<language type="ik">Inupiaq</language>
-			<language type="ike">Eastern Canadian Inuktitut</language>
+			<language type="iu">Inuktitut</language>
 			<language type="ikt">Western Canadian Inuktitut</language>
 			<language type="ilo">Iloko</language>
 			<language type="inh">Ingush</language>
@@ -474,7 +474,7 @@ annotations.
 			<language type="oj">Ojibwa</language>
 			<language type="ojb">Northwestern Ojibwa</language>
 			<language type="ojc">Central Ojibwa</language>
-			<language type="ojg">Eastern Ojibwa</language>
+			<language type="oj">Ojibwa</language>
 			<language type="ojs">Oji-Cree</language>
 			<language type="ojw">Western Ojibwa</language>
 			<language type="oka">Okanagan</language>
--- a/common/main/fi.xml
+++ b/common/main/fi.xml
@@ -31,7 +31,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="afh">afrihili</language>
 			<language type="agq">aghem</language>
 			<language type="ain">ainu</language>
-			<language type="ajp">urduni</language>
+			<language type="apc">urduni</language>
 			<language type="ak">akan</language>
 			<language type="akk">akkadi</language>
 			<language type="akz">alabama</language>
--- a/common/main/la.xml
+++ b/common/main/la.xml
@@ -24,7 +24,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="az" draft="provisional">Atropatenica</language>
 			<language type="be" draft="provisional">Ruthenica Alba</language>
 			<language type="bg" draft="provisional">Bulgarica</language>
-			<language type="bh" draft="provisional">Bihari</language>
 			<language type="bn" draft="provisional">Bengalica</language>
 			<language type="bo" draft="provisional">Tibetana</language>
 			<language type="br" draft="provisional">Britonica</language>
@@ -66,12 +65,12 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="ia" draft="provisional">Interlingua</language>
 			<language type="ie" draft="provisional">Interlingue</language>
 			<language type="ig" draft="provisional">Igbonica</language>
-			<language type="in" draft="provisional">Indonesia</language>
+			<language type="id" draft="provisional">Indonesia</language>
 			<language type="is" draft="provisional">Islandica</language>
 			<language type="it" draft="provisional">Italiana</language>
-			<language type="iw" draft="provisional">Hebraica</language>
+			<language type="he" draft="provisional">Hebraica</language>
 			<language type="ja" draft="provisional">Iaponica</language>
-			<language type="ji" draft="provisional">Iudaeogermanica</language>
+			<language type="yi" draft="provisional">Iudaeogermanica</language>
 			<language type="jv" draft="provisional">Iavensis</language>
 			<language type="ka" draft="provisional">Georgiana</language>
 			<language type="kk" draft="provisional">Cazachica</language>
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -4928,7 +4928,7 @@ XXX Code for transations where no currency is involved
 		<weekOfPreference ordering="weekOfYear weekOfInterval" locales="zu"/>
 		<weekOfPreference ordering="weekOfDate" locales="ca es fr gl"/>
 		<weekOfPreference ordering="weekOfDate weekOfMonth" locales="en bn ja ka"/>
-		<weekOfPreference ordering="weekOfDate weekOfMonth weekOfInterval" locales="bg de iw pt ur zh"/>
+		<weekOfPreference ordering="weekOfDate weekOfMonth weekOfInterval" locales="bg de he pt ur zh"/>
 		<weekOfPreference ordering="weekOfDate weekOfYear weekOfMonth" locales="nl"/>
 		<weekOfPreference ordering="weekOfDate weekOfInterval weekOfMonth" locales="af"/>
 		<weekOfPreference ordering="weekOfMonth" locales="ar fil gu hu hy id kk ko"/>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
@@ -157,14 +157,22 @@ public abstract class MatchValue implements Predicate<String> {
         private final Predicate<String> variant;
 
         public LocaleMatchValue() {
-            this(null);
+            this(Set.of(Status.regular, Status.special, Status.unknown, Status.macroregion));
         }
 
         public LocaleMatchValue(Set<Status> statuses) {
-            lang = new ValidityMatchValue(LstrType.language, statuses, false);
-            script = new ValidityMatchValue(LstrType.script, statuses, false);
-            region = new ValidityMatchValue(LstrType.region, statuses, false);
-            variant = new ValidityMatchValue(LstrType.variant, statuses, false);
+            this(statuses, statuses, statuses, statuses);
+        }
+
+        public LocaleMatchValue(
+                Set<Status> langStatus,
+                Set<Status> scriptStatus,
+                Set<Status> regionStatus,
+                Set<Status> variantStatus) {
+            lang = new ValidityMatchValue(LstrType.language, langStatus, false);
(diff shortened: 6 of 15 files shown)


common/main/en.xml
```diff
@@ -152,7 +152,6 @@ annotations.
 			<language type="csw">Swampy Cree</language>
 			<language type="cu">Church Slavic</language>
 			<language type="cv">Chuvash</language>
-			<language type="cwd">Woods Cree</language>
```
conradarcturus, line 155: Do we need to remove these? I imagine they are considered invalid because`cr` & `hai` are macrolanguages -- since macrolanguages cannot have locale data (it's poorly defined) `cwd` is aliased to `cr` and `hdn` is aliased to `hai`.

However it depends on how this data is used. If its for translations of ISO names people will want these constituent languages. Additionally `cr` above is `<language type="cr">Cree</language>` -- contributing to the further confusion about what our `cr` data actually is.

imho we should keep `cwd` here. Additionally (I'm happy to file a ticket if you like the idea) we should move the current `cr.xml` to `cwd.xml` and leave `cr.xml` as a stub that points to `cwd.xml` -- that will make the macrolanguage/alias hierarchy much clearer to external users of the CLDR data.

cc @srl295 @DavidLRowe who also had a lot of context on the first nations macrolanguages.

common/main/la.xml
```diff
@@ -66,12 +65,12 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<language type="ia" draft="provisional">Interlingua</language>
 			<language type="ie" draft="provisional">Interlingue</language>
 			<language type="ig" draft="provisional">Igbonica</language>
-			<language type="in" draft="provisional">Indonesia</language>
```
conradarcturus, line 69: uffa, how did these sneak in! Thanks for fixing them.

common/dtd/ldml.dtd
```diff
@@ -61,7 +61,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT language ( #PCDATA ) >
 <!ATTLIST language type NMTOKEN #REQUIRED >
-    <!--@MATCH:validity/locale-->
+    <!--@MATCH:validity/nlocale-->
```
conradarcturus, line 64: Instead of adding this unclear jargon -- can we add an or here?
```suggestion
    <!--@MATCH:validity/locale||validity/deprecatedlocale-->
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -317,28 +418,39 @@ private ValidityMatchValue(LstrType type, Set<Status> statuses, boolean shortId)
...
-                    statuses == null ? EnumSet.allOf(Status.class) : ImmutableSet.copyOf(statuses);
+                    ImmutableSet.copyOf(statuses == null ? DEFAULT_STATUS.get(type) : statuses);
         }
 
         public static MatchValue of(String typeName) {
             if (typeName.equals("locale")) {
                 return new LocaleMatchValue();
             }
+            if (typeName.equals("xlocale")) {
+                return new XLocaleMatchValue();
```
conradarcturus, line 431: Instead of confusing variable names `xlocale` and `nlocale` can we spell it out? And since we can use logical ors we can just make specific sets eg. `deprecatedlocale` and `overlonglocale` and people use like`validity/locale||validity/overlonglocale`.

conradarcturus (changes requested):
Great! Thanks for finding invalid codes and improving the checker.

I left some comments that I think will improve CLDR readability for external developers/users. Happy to go back on forth on these ideas if you like.

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
macchiati: What I intended was having common semantics nlocale doesn't mean 'or deprecated'), but nlocale and xlocale are obscure. I'll change to more meaningful names.

common/main/en.xml
macchiati: Normally we use the common name, without terms to distinguish from other variants. So we say "Arabic" and not "Modern Standard Arabic". We do have the facility for the long form, however; you see that for zh.

So we could have 
<language type="cr">Cree</language>
<language type="cr" alt="long">Woods Cree</language>

That is consistent with other cases, so I can add that.

I disagree with the stub approach. We do that for 'no' and 'en-001' and it is a pain.

common/main/fi.xml
```diff
@@ -31,7 +31,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="afh">afrihili</language>
 			<language type="agq">aghem</language>
 			<language type="ain">ainu</language>
-			<language type="ajp">urduni</language>
```
srl295, line 34: this was a deprecated code so seems fine

common/main/en.xml
srl295: Yeah, I don't think removing encompassed languages makes sense here.  cwd is a valid display name and should be retained.

common/main/en.xml
```diff
@@ -256,7 +255,6 @@ annotations.
 			<language type="hak">Hakka Chinese</language>
 			<language type="haw">Hawaiian</language>
 			<language type="hax">Southern Haida</language>
-			<language type="hdn">Northern Haida</language>
```
srl295, line 259: why?

common/supplemental/supplementalData.xml
```diff
@@ -4928,7 +4928,7 @@ XXX Code for transations where no currency is involved
 		<weekOfPreference ordering="weekOfYear weekOfInterval" locales="zu"/>
 		<weekOfPreference ordering="weekOfDate" locales="ca es fr gl"/>
 		<weekOfPreference ordering="weekOfDate weekOfMonth" locales="en bn ja ka"/>
-		<weekOfPreference ordering="weekOfDate weekOfMonth weekOfInterval" locales="bg de iw pt ur zh"/>
```
srl295, line 4931: there was a bug on this somewhere.

exemplars/main/rna.xml
```diff
@@ -1,23 +0,0 @@
...
-		<language type="rna"/>
-	</identity>
-	<layout>
-		<orientation>
-			<characterOrder>left-to-right</characterOrder>
-			<lineOrder>top-to-bottom</lineOrder>
-		</orientation>
-	</layout>
-	<characters>
-		<exemplarCharacters draft="unconfirmed">[a b c d e f g h i j k m n o p r s t u v w y z]</exemplarCharacters>
```
srl295, line 20: Why was this deleted?

srl295: OK, I can't find what this was supposed to be, so fine to remove.

srl295:
java code LGTM (with q's though)

common/main/en.xml
macchiati: There was a hole whole where some deprecated language codes crept in (probably via raw XML editing or bulk upload).

common/supplemental/supplementalData.xml
macchiati: Yes, there are some odd locations for 'iw'. The @MATCH cases now have these explicit variant locale tests, so we can review them.

exemplars/main/rna.xml
macchiati: The file itself has a deprecated language code, 'rna', without any replacement alias provided by ISO.

common/main/en.xml
macchiati: In CLDR, the aliased encompassed codes have always been classed as deprecated.

common/main/en.xml
macchiati: See below

common/main/en.xml
```diff
@@ -139,6 +139,7 @@ annotations.
 			<language type="cop">Coptic</language>
 			<language type="cps">Capiznon</language>
 			<language type="cr">Cree</language>
+			<language type="cr" alt="long">Woods Cree</language>
```
srl295, line 142: I don't think this is a great situation. Woods Cree isn't a "longer" form of Cree. 

Content (whether a document, an audio recording, etc.) isn't going to be just in Cree, it's going to be in Woods Cree, or Plains Cree, or Swampy Cree, etc.  Perhaps the macrolanguage isn't helpful here and should be removed. (The pronunciation and orthography is distinct between these three, in both Latn and Cans scripts.)

For now, I kind of think the best and most consistent change would be `<language type="cr">Woods Cree</language>`

srl295:
```suggestion
			<language type="cr">Woods Cree</language>
```

macchiati: That would overturn from our long-standing policy for dealing with the compatibility mess that macrolangauges created.*

We **don't** say 
- Tosk Albanian, but rather Albanian
- Standard Arabic, but rather Arabic
- Standard Estonian, but rather Estonian
- Eastern Yiddish, but rather Yiddish
- Northern Uzbek, but rather Uzbek
- Iranian Persian, but rather Persian (or Farsi)

and so on. 

That is, we use the code and name of the macrolanguage for the predominant encompassed code. The one exception off the top of my head is for "zh", where we have:

```
<language type="zh">Chinese</language>
<language type="zh" alt="long">Mandarin Chinese</language>
```

So that's why I _was_ "consistent" in following that model for Cree. 

* If the macrolanguage policy had been _consistently_ applied by ISO, they'd have made German (de) as a macrolanguage, and given a new code for Standard German (eg, dex?) as an encompassed language (and made the 30ish other Germans like Colonia Tovar German, Swiss German, Low German, and so on also be encompassed languages). And the same for many other languages (French, Hindi, ...).

srl295: Would there be some option to have CLDR decide that cr isn't (Or isn't considered) a macrolanguage? And just allow `cwd` to stand on its own? 

I hear you about consistency but it does not work with the identity of the languages encompassed under cr.

The difference with German is that people do refer to `de` as Deutsch whereas even the name of the language differs between cwd/crk/csw - nehenawewin, neheyawewin and nehethawewin (not respectively).

srl295: I'm fine to land and then discuss. And I know this is long standing policy and not new.

srl295: > Those appear to be just the same word with slightly different pronunciations in different dialects, according to https://en.wikipedia.org/wiki/Cree_language#Dialect_criteria, which is exposed because written Cree is more phonetic than English.
> 
> 
> 
> It's as if we said that there were 3 different names for English: /ˈɪŋɡlɪʃ/ (UK), /ˈɪŋlɪʃ/ (US), /ˈɪŋɡləʃ/ (AU).
> 
> 
> 
> (They also appear to be really different dialects, not different languages, at a least according to https://en.wikipedia.org/wiki/East_Cree)
> 
> 
> 
> We should take this up in further in our next meeting: I filed https://unicode-org.atlassian.net/browse/CLDR-18142 about this.
> 
> 
> 
> In the meantime, I'll put out a call for taking the ticket for 47. We can always reverse the change for cr should TC decide it.

Right but one is not more general or more representative than the other. There are vocabulary differences as well.

macchiati: Generality, representativity, vocabulary differences, are not relevant to choosing the predominant encompassed language.

As I said, let's discuss in the meeting.

conradarcturus: As you mentioned, this change isn't about Macrolanguages -- it's about validity. So I'll accept to unblock. We probably don't have enough time to properly discuss the issue before the new year anyway, so let's not get ahead of ourselves.

We should follow up in a separate ticket -- but decide the timeline for this.

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -174,29 +187,74 @@ public String getName() {
...
         public String getSample() {
             return "de";
         }
     }
 
+    /**
+     * Check for the language OR certain backwards-compatible exceptions for data to support
+     * retaining variants, namely likelySubtags: "in","iw","ji","jw","mo","tl"
+     */
+    public static class XLocaleMatchValue extends LocaleMatchValue {
```
conradarcturus, line 224: personal style nit (you can choose your own style): have the explanation in the name
```suggestion
    public static class LocaleOrDeprecatedLocaleMatchValue extends LocaleMatchValue {
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
```diff
@@ -174,29 +187,74 @@ public String getName() {
...
+
+        @Override
+        public boolean checkLang(String language) {
+            return super.checkLang(language) // first check normal
+                    || exceptions.contains(language);
+        }
+
+        @Override
+        public String getName() {
+            return "validity/locale-for-likely";
```
conradarcturus, line 235: personal style nit (you can choose your own style): don't label "why", rather label "what". Why is good in comments where it is used, but this function doesn't itself matter for likely subtags -- and it COULD matter for other usages.
```suggestion
            return "validity/locale-or-deprecated-locale";
```

conradarcturus (approved):
Accepting to unblock. Any comments are stylistic or part of a bigger conversation that should not block this particular change.

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
macchiati: The problem is that _or-deprecated_ is not true. It is only _certain_ deprecated codes. Spelling that out is best left for documentation, since it would be quite long.

tools/cldr-code/src/main/java/org/unicode/cldr/util/MatchValue.java
macchiati: Thanks!

macchiati:
I also looked at updating the @MATCH documentation. I see a couple of different paths — thoughts welcome:

https://unicode-org.atlassian.net/browse/CLDR-18146

PR #4216, opened 2024-11-27
CLDR-10397 Fix languageInfo.xml conversion to json

arjanmels:
CLDR-10397

Adds the "desired" field to the languageMatch data

ALLOW_MANY_COMMITS=true

### Review Notes

See comment below for sample files.


srl295 (dismissed):

srl295:
scriptVariants #4396 were also broken in JSON, so will fix here too

srl295:
Sample files for review: 

- [languageMatching.json](https://github.com/user-attachments/files/19091830/languageMatching.json)

- [scriptData.json](https://github.com/user-attachments/files/19091832/scriptData.json)

btangmu (approved):
LGTM, though I'm not familiar with this code. Unit tests, that would fail before this and pass after, would increase confidence

srl295:
> LGTM, though I'm not familiar with this code. Unit tests, that would fail before this and pass after, would increase confidence

The only test we have is running the json converter. 

Perhaps we could have a test that builds a subset of locales to see that it at least builds.

srl295:
> Yes. I'm also thinking of smaller unit tests, like testing a single method that uses ARRAY_ITEM_PATTERN -- it should fail without the change to languageMatching and pass with that change

right, we could unit test the patterns, etc.

srl295:
Thanks @arjanmels

PR #4217, opened 2024-11-28
CLDR-18136 Fix ja, ko, zh, yue datetimeSkeleton

sffc:
CLDR-18136

ALLOW_MANY_COMMITS=true

I made a basic test for this in ICU4X and used it to identify what to fix in the first pass. https://github.com/unicode-org/icu4x/pull/5826


pedberg-icu (approved):

pedberg-icu:
This seems reasonable as a super-minimal change. In 47 we should change the standard patters to use MMM etc. so we can round-trip between the standard patterns and the corresponding skeletons.

macchiati (approved):

macchiati:
Shane, I'm taking the liberty of setting the ticket to Reviewing, with myself as Reviewer.

PR #4225, opened 2024-12-04
CLDR-18050 spec: use small caps for h5

srl295:
CLDR-18050

ALLOW_MANY_COMMITS=true

## Review Notes

* Before

- <https://www.unicode.org/reports/tr35/dev/tr35-dates.html#Semantic_Time_Field_Sets>

image

* After

image

### DIFF
--- a/tools/scripts/tr-archive/tr35.css
+++ b/tools/scripts/tr-archive/tr35.css
@@ -29,3 +29,7 @@
 .markdown-alert-important .markdown-alert-title {
     color: blueviolet;
 }
+
+.body h5 {
+    font-variant: small-caps;
+}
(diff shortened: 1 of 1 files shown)


tools/scripts/tr-archive/tr35.css
```diff
@@ -29,3 +29,7 @@
 .markdown-alert-important .markdown-alert-title {
     color: blueviolet;
 }
+
+.body h5 {
+    font-variant: small-caps;
```
macchiati, line 34: Needs to be the following, otherwise the size is too small (and we want to guarantee it is bold)
```suggestion
    font-variant: small-caps;
    font-size: medium;
    font-weight: bold;
```

macchiati (changes requested):

tools/scripts/tr-archive/tr35.css
srl295: done, now: 

image

macchiati (approved):

PR #4226, opened 2024-12-04
CLDR-15830 Refactor getName: new class NameGetter

btangmu:
-Each CLDRFile has a NameGetter; all getName methods moved to NameGetter

-Give GetName methods distinct names for clarify and to avoid excessive overloading

-Make methods private where possible

-Move some calls to nameGetter() out of loops, re-use

-Remove some dead code

-Comments

CLDR-15830

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
@@ -22,16 +22,16 @@ public class NameGetter {
      */
     public NameGetter(CLDRFile cldrFile) {
         if (cldrFile == null) {
-            throw new RuntimeException("NameGetter must have non-null CLDRFile");
+            throw new IllegalArgumentException("NameGetter must have non-null CLDRFile");
         }
         this.cldrFile = cldrFile;
     }
 
-    private static final String GETNAME_LOCALE_SEPARATOR =
+    private static final String GETNAME_LOCALE_SEPARATOR_PATH =
             "//ldml/localeDisplayNames/localeDisplayPattern/localeSeparator";
-    private static final String GETNAME_LOCALE_PATTERN =
+    private static final String GETNAME_LOCALE_PATTERN_PATH =
             "//ldml/localeDisplayNames/localeDisplayPattern/localePattern";
-    private static final String GETNAME_LOCALE_KEY_TYPE_PATTERN =
+    private static final String GETNAME_LOCALE_KEY_TYPE_PATTERN_PATH =
             "//ldml/localeDisplayNames/localeDisplayPattern/localeKeyTypePattern";
 
     private static final Joiner JOIN_HYPHEN = Joiner.on('-');
@@ -102,9 +102,9 @@ public class NameGetter {
                 lparser,
                 onlyConstructCompound,
                 null,
-                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_KEY_TYPE_PATTERN),
-                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_PATTERN),
-                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_SEPARATOR),
+                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_KEY_TYPE_PATTERN_PATH),
+                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_PATTERN_PATH),
+                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_SEPARATOR_PATH),
                 null);
     }
 
@@ -161,9 +161,9 @@ public class NameGetter {
         return getNameFromManyThings(
                 localeOrTZID,
                 onlyConstructCompound,
-                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_KEY_TYPE_PATTERN),
-                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_PATTERN),
-                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_SEPARATOR),
+                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_KEY_TYPE_PATTERN_PATH),
+                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_PATTERN_PATH),
+                cldrFile.getWinningValueWithBailey(GETNAME_LOCALE_SEPARATOR_PATH),
                 altPicker,
                 paths);
     }
@@ -361,12 +361,12 @@ public class NameGetter {
                 }
                 value = MessageFormat.format(localeKeyTypePattern, kname, oldFormatType);
                 if (paths != null) {
-                    paths.add(GETNAME_LOCALE_KEY_TYPE_PATTERN);
+                    paths.add(GETNAME_LOCALE_KEY_TYPE_PATTERN_PATH);
                 }
             }
             value = replaceBracketsForName(value);
             if (paths != null && !extras.isEmpty()) {
-                paths.add(GETNAME_LOCALE_SEPARATOR);
+                paths.add(GETNAME_LOCALE_SEPARATOR_PATH);
             }
             extras =
                     extras.isEmpty() ? value : MessageFormat.format(localeSeparator, extras, value);
@@ -380,7 +380,7 @@ public class NameGetter {
                             extension.getKey(),
                             JOIN_HYPHEN.join(extension.getValue()));
             if (paths != null) {
-                paths.add(GETNAME_LOCALE_KEY_TYPE_PATTERN);
+                paths.add(GETNAME_LOCALE_KEY_TYPE_PATTERN_PATH);
             }
             extras =
                     extras.isEmpty() ? value : MessageFormat.format(localeSeparator, extras, value);
@@ -390,7 +390,7 @@ public class NameGetter {
             return name;
         }
         if (paths != null) {
-            paths.add(GETNAME_LOCALE_PATTERN);
+            paths.add(GETNAME_LOCALE_PATTERN_PATH);
         }
         return MessageFormat.format(localePattern, name, extras);
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
@@ -299,15 +299,15 @@ public class NameGetter {
 
         // Look for key-type pairs.
         for (Map.Entry<String, List<String>> extension :
-            lparser.getLocaleExtensionsDetailed().entrySet()) {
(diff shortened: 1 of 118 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+    private final CLDRFile cldrFile;
+
+    /**
+     * Construct a new NameGetter.
+     *
+     * @param cldrFile must not be null
+     */
+    public NameGetter(CLDRFile cldrFile) {
+        if (cldrFile == null) {
+            throw new RuntimeException("NameGetter must have non-null CLDRFile");
```
macchiati, line 25: IllegalArgumentException is more specific.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+        }
+        this.cldrFile = cldrFile;
+    }
+
+    private static final String GETNAME_LOCALE_SEPARATOR =
```
macchiati, line 30: Might put PATH in the name, also for similar ones.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+     * <p>Ideally getNameFromTypestrCode and getNameFromTypenumCode should be replaced by a method
+     * that takes an enum rather than String or int. Compare StandardCodes.CodeType, and the
+     * integers defined in the vicinity of CLDRFile.TERRITORY_NAME
+     *
+     * @param type a string such as "currency", "language", "region", "script", "territory", ...
+     * @param code a code such as "JP"
+     * @return the name
+     */
+    public String getNameFromTypestrCode(String type, String code) {
+        return getNameFromTypenumCode(CLDRFile.typeNameToCode(type), code);
```
macchiati, line 52: Once you get this merged, a second refactoring is to inline this and similar ones. 'Typenum' is hard to read; TypeEnum is better.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    /**
+     * Get a name, given a type (as a int) and a code.
+     *
+     * @param type the type such as CLDRFile.TERRITORY_NAME
+     * @param code a code such as "JP"
+     * @return the name
+     */
+    public String getNameFromTypenumCode(int type, String code) {
```
macchiati, line 62: We should also get rid of the CLDR fake enums like CLDRFile.TERRITORY_NAME in favor of enums. This is probably easier in a second PR, later.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+
+    /**
+     * Returns the name of the given bcp47 identifier. Note that extensions must be specified using
+     * the old "\@key=type" syntax.
+     *
+     * @param localeOrTZID the bcp47 identifier (locale or timezone ID)
+     * @return the name of the given bcp47 identifier
+     */
+    public synchronized String getNameFromBCP47(String localeOrTZID) {
+        return getNameFromBCP47Bool(localeOrTZID, false);
```
macchiati, line 83: As long as we are writing new code, best to avoid having a boolean in any API, and rather use a meaningful enum. Again, could be done in later PR.

macchiati (commented):
Looks good, just had a few comments.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
btangmu: Fixed in latest commit

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
btangmu: Fixed in latest commit

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
btangmu: Yes, I'll address that in a subsequent PR. By "inline", do you mean, for example, to change `getNameFromTypestrCode("territory", territory)` to `getNameFromTypenumCode(CLDRFile.TERRITORY_NAME, territory)`? And then subsequently (or in the same PR) change CLDRFile.TERRITORY_NAME, etc., to an enum?

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
btangmu: Right; see above comment

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
btangmu: Good idea. I'll include that in a subsequent PR

btangmu:
I've made this ready for review. So far this is fairly trivial refactoring but affects many files, so better not to keep changing on the same PR

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+     * @param localeOrTZID the bcp47 identifier (locale or timezone ID)
+     * @param onlyConstructCompound if true, returns "English (United Kingdom)" instead of "British
+     *     English"
+     * @param localeKeyTypePattern the pattern used to format key-type pairs
+     * @param localePattern the pattern used to format primary/secondary subtags
+     * @param localeSeparator the list separator for secondary subtags
+     * @param paths if non-null, fillin with contributory paths
+     * @return the name of the given bcp47 identifier
+     */
+    private synchronized String getNameFromManyThings(
```
srl295, line 184: again, a builder model that sets up options might be better than a kitchen sink set of parameters.

`new NameFormatter(locale).onlyConstructCompound(true).withLocalePatter(pattern)…`

btangmu:
> I think it's OK but the naming is awkward. Many of the functions could have the same name, just with different overloaded parameters...

Steven, one reason for this refactor is that it was difficult to see how any particular version of getName() was actually used. With current main, IntelliJ reports getName "has 347 callers". Probably there's a way to find callers without including overloaded versions, but anyway 14 different methods in the same class with the same name is just too much for my little brain to work with. Refactoring is often easier if you can use global search-and-replace; methods with identical names get in the way of that.

Another reason for the refactor is that these methods were tangled up in the middle of CLDRFile which is just too large for maintainable code.

I agree that the names are somewhat awkward or arbitrary, but for me at least they give me a fighting chance to make sense what's calling what. After the code is reasonably clean, a nice finishing touch might be to restore some overloading, within reason.

I complete agree with you that some of these methods have a "kitchen sink of parameters" -- best practice is that no method should have more than about 4 parameters. A possible solution is to make them members of a class, and I suspect that's what you mean by a Builder model.

I can still hardly understand half of what's going on in this new class, so I can only revise so much at a time :-)

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+import com.ibm.icu.text.MessageFormat;
+import com.ibm.icu.text.Transform;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class NameGetter {
```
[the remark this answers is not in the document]
btangmu: > isn't it a display name formatter?

Good question! Is that what it is?

> BCP47 does not have @ signs in it, so a different name should be used.

That sounds important. Some of the methods already had comments saying parameters were bcp47 identifiers, and I carried that over to the related methods (callers/callees). If that's inaccurate, what would be a more accurate description? Part of the problem with this whole module is that there are strings and integers coming in without a clear explanation of what strings and integers are actually expected/allowed, and that's why Mark and I have been discussing an Enum instead of strings and integers, wherever that's possible...

srl295:
> > I think it's OK but the naming is awkward. Many of the functions could have the same name, just with different overloaded parameters...
> 
> Steven, one reason for this refactor is that it was difficult to see how any particular version of getName() was actually used. With current main, IntelliJ reports getName "has 347 callers". Probably there's a way to find callers without including overloaded versions, but anyway 14 different methods in the same class with the same name is just too much for my little brain to work with. Refactoring is often easier if you can use global search-and-replace; methods with identical names get in the way of that.

That's a good point for that analysis.  Since it's a library, I think it's harder to get this kind of analysis otherwise - SurveyTool might call it deep inside something for example (cross library), so "function is unused" warnings don't help.

> Another reason for the refactor is that these methods were tangled up in the middle of CLDRFile which is just too large for maintainable code.

💯 

> I agree that the names are somewhat awkward or arbitrary, but for me at least they give me a fighting chance to make sense what's calling what. After the code is reasonably clean, a nice finishing touch might be to restore some overloading, within reason.

sounds good… 

> I complete agree with you that some of these methods have a "kitchen sink of parameters" -- best practice is that no method should have more than about 4 parameters. A possible solution is to make them members of a class, and I suspect that's what you mean by a Builder model.

At leisure take a look at https://www.baeldung.com/java-fluent-interface-vs-builder-pattern I think it's the Builder model (not fluent, though they have similarities).  Yes,  you'd have basically a Settings-type object that gets modified by calling functions on it (which you could chain) and then it ends up with a toString() call when you're ready to output it as text.

> I can still hardly understand half of what's going on in this new class, so I can only revise so much at a time :-)

I'm with you! But it's a good refactor—— and Bottom line is that's why I hit **Approve**.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -0,0 +1,547 @@
...
+     *
+     * <p>Ideally getNameFromTypestrCode and getNameFromTypenumCode should be replaced by a method
+     * that takes an enum rather than String or int. Compare StandardCodes.CodeType, and the
+     * integers defined in the vicinity of CLDRFile.TERRITORY_NAME
+     *
+     * @param type a string such as "currency", "language", "region", "script", "territory", ...
+     * @param code a code such as "JP"
+     * @return the name
+     */
+    public String getNameFromTypestrCode(String type, String code) {
```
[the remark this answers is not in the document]
btangmu: yes, eventually we might not need/want to keep "getName" in the method names

PR #4227, opened 2024-12-05
CLDR-18150 Fix private-use units to avoid validity constraints

macchiati:
CLDR-18150

This is to fix a regression introduced by [CLDR-15948: Clarify the well-formedness and validity of certain Unicode CLDR identifiers for MF2](https://unicode-org.atlassian.net/browse/CLDR-15948).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -943,10 +943,9 @@ per
     * 100
 
 <a name='pu_single_unit' href='#pu_single_unit'>pu_single_unit</a> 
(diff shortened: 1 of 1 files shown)


macchiati:
Please review this spec fix for a regression introduced by a v46.1 ticket.

btangmu (approved):
LGTM, however I'm still not very familiar with LDML; in particular I don't know where the `[a-z0-9]{3,8}` pattern comes from

srl295 (approved):

PR #4236, opened 2024-12-09
CLDR-18170 46.1 release page

macchiati:
CLDR-18170

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-46.md
+++ b/docs/site/downloads/cldr-46.md
@@ -6,8 +6,8 @@ title: CLDR 46 Release Note
 
 | No. |    Date    | Rel. Note |  Data  |  Charts  | Spec |   Delta  | GitHub Tag | Delta DTD | CLDR JSON |
 |:---:|:----------:|:---------:|:------:|:--------:|:------------:|:---:|:----------:|:---------:|:---------:|
+|  46.1 | 2024-12-18 | [v46.1](/index/downloads/cldr-46.1) | [CLDR46.1](https://unicode.org/Public/cldr/46.1/) | [Charts46.1](https://unicode.org/cldr/charts/46.1/) | [LDML46](https://www.unicode.org/reports/tr35/tr35-73/tr35.html) | [Δ46.1](https://unicode-org.atlassian.net/issues/?jql=project+%3D+CLDR+AND+status+%3D+Done+AND+resolution+%3D+Fixed+AND+fixVersion+%3D+%2246.1%22+ORDER+BY+priority+DESC) | [release-46](https://github.com/unicode-org/cldr/releases/tag/release-46.1) | _n/a_ | [46.0.0](https://github.com/unicode-org/cldr-json/releases/tag/46.1.0) |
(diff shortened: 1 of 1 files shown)


AEApple:
Well the markdown table isn't rendering correctly, and shouldn't we say it's in Beta until we're ready to release it?

macchiati:
That was just a rough draft to get us started; I had a bunch of notes in a
different thread.

macchiati:
1. Text is a rough draft
2. I put the 46.1 section right after the overview rather than at the bottom, to highlight better. 
    3. Maybe make the header "Release 46.1"?
4. Needs the most-significant MF2.0 items list from Addison
5. In the header, there is a new row for 46.1. 
    6. The links etc need fixing
    7. The release note link should be changed to the new section
    8. For the charts, we could either say n/a, or build 46.1 charts in the https://unicode.org/cldr/charts/46/ directory.
    9. The spec needs to be LDML46.1
    10. The delta is ok.
    11. The release needs to say/link 46.1
    12. The CLDR JSON needs to say/link to 46.1
    13. Once it is all done, the new row needs to be copied to https://cldr.unicode.org/index/downloads
    14. In https://e6182513.cldr.pages.dev/, the table at the top is not HTMLified. Not sure why

docs/site/downloads/cldr-46.md
```diff
@@ -6,8 +6,8 @@ title: CLDR 46 Release Note
 
 | No. |    Date    | Rel. Note |  Data  |  Charts  | Spec |   Delta  | GitHub Tag | Delta DTD | CLDR JSON |
 |:---:|:----------:|:---------:|:------:|:--------:|:------------:|:---:|:----------:|:---------:|:---------:|
+|  46.1 | 2024-12-18 | [v46.1](/index/downloads/cldr-46.1) | [CLDR46.1](https://unicode.org/Public/cldr/46.1/) | [Charts46.1](https://unicode.org/cldr/charts/46.1/) | [LDML46](https://www.unicode.org/reports/tr35/tr35-73/tr35.html) | [Δ46.1](https://unicode-org.atlassian.net/issues/?jql=project+%3D+CLDR+AND+status+%3D+Done+AND+resolution+%3D+Fixed+AND+fixVersion+%3D+%2246.1%22+ORDER+BY+priority+DESC) | [release-46](https://github.com/unicode-org/cldr/releases/tag/release-46.1) | _n/a_ | [46.0.0](https://github.com/unicode-org/cldr-json/releases/tag/46.1.0) |
```
pedberg-icu, line 9: index/downloads/cldr-46.1 is not the right link for a release note, we are just adding stuff into index/downloads/cldr-46. And are we planning to generate new charts for 46.1? And the spec will be version 74, not 73. And the release tag will be "release-46-1" (like the other dot releases), not 46.1

pedberg-icu (changes requested):
See previous note about fixes

docs/site/downloads/cldr-46.md
pedberg-icu: And the entry for the json tag should have [46.1.0]

AEApple:
@pedberg-icu - should we just approve this and then make follow on changes since it would be good to have it updated for the beta until we finalize the release?

docs/site/downloads/cldr-46.md
pedberg-icu: Well, I will unblock this so we can merge and then make a separate PR to fix the issues noted above.

pedberg-icu:
> @pedberg-icu - should we just approve this and then make follow on changes since it would be good to have it updated for the beta until we finalize the release?

Yes

pedberg-icu (changes requested):
Actually I will re-block. We cannot merge this as is, it will break the existing public release page for 46.

PR #4237, opened 2024-12-10
CLDR-18155 Order languageData's scripts by number of users

conradarcturus:
CLDR-18155

This sorts the scripts in the SupplementalData <languageData><language> tags such that the first script has the highest population. This helps resolve some of the ambiguities interpreting the data.

ALLOW_MANY_COMMITS=true


common/supplemental/supplementalData.xml
```diff
@@ -1346,7 +1346,7 @@ XXX Code for transations where no currency is involved
 		<language type="awa" scripts="Deva"/>
 		<language type="awa" territories="IN" alt="secondary"/>
 		<language type="ay" scripts="Latn" territories="BO"/>
-		<language type="az" scripts="Arab Cyrl Latn" territories="AZ"/>
+		<language type="az" scripts="Arab Latn Cyrl" territories="AZ"/>
```
conradarcturus, line 1349: likely subtags for az is az_Latn_AZ but I will note there are more Azerbaijani speakers in Iran (az_Arab_IR). It's just that they haven't transitioned to the internet like Azerbaijan has.

common/supplemental/supplementalData.xml
```diff
@@ -1918,7 +1918,7 @@ XXX Code for transations where no currency is involved
 		<language type="mai" scripts="Tirh" territories="IN NP" alt="secondary"/>
 		<language type="mak" scripts="Latn"/>
 		<language type="mak" scripts="Bugi" territories="ID" alt="secondary"/>
-		<language type="man" scripts="Latn Nkoo"/>
+		<language type="man" scripts="Nkoo Latn"/>
```
conradarcturus, line 1921: This is driven by the Ghana language entry saying there is a large Manding population using Nkoo in Ghana. Really `man` is a macrolanguage code so we shouldn't give that much weight to it. https://en.wikipedia.org/wiki/Manding_languages

common/supplemental/supplementalData.xml
```diff
@@ -2435,7 +2435,7 @@ XXX Code for transations where no currency is involved
 		<language type="yrk" scripts="Cyrl"/>
 		<language type="yrl" scripts="Latn"/>
 		<language type="yua" scripts="Latn"/>
-		<language type="yue" scripts="Hans Hant" territories="MO"/>
+		<language type="yue" scripts="Hant Hans" territories="MO"/>
```
conradarcturus, line 2438: yue's likely subtag is yue_Hant_HK

conradarcturus (commented):
Added in-line comments for some languages with interesting changes. Otherwise they all match likely subtags and public records.

common/supplemental/supplementalData.xml
macchiati: Can you check that again? Nkoo figures can be skewed by advocacy.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -344,11 +346,23 @@ public BasicLanguageData setType(Type type) {
...
-            return setScripts(
-                    scriptTokens == null
-                            ? null
-                            : Arrays.asList(WHITESPACE_PATTERN.split(scriptTokens)));
+        // Adding scripts but leaving 0 as a placeholder when there is no population data
+        // input: a whitespace-separated list of scripts
+        public BasicLanguageData setScriptsWithoutPopulation(String scriptTokens) {
+            List<String> scripts = new ArrayList<>();
+            if (scriptTokens != null) {
+                scripts = Arrays.asList(WHITESPACE_PATTERN.split(scriptTokens));
```
macchiati, line 354: Splitter ha a split to list, fyi

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -344,11 +346,23 @@ public BasicLanguageData setType(Type type) {
...
+            return setScriptsWithoutPopulation(scripts);
+        }
+
+        // Adding scripts but leaving 0 as a placeholder when there is no population data
+        public BasicLanguageData setScriptsWithoutPopulation(Collection<String> scripts) {
+            Map<String, Integer> scriptsByPopulation = new TreeMap<>();
+            for (String script : scripts) {
+                scriptsByPopulation.put(script, 0);
+            }
+            return setScripts(scriptsByPopulation);
```
macchiati, line 365: Better to build once, make immutable.

macchiati (dismissed):

common/supplemental/supplementalData.xml
conradarcturus: I tried to find better figures but I could not verify Nko usage by Manding -- so I'll edit the population data so Latn takes precedence.

common/supplemental/supplementalData.xml
macchiati: While more speakers, it appears that there is more written in Latin (AZ) that Arabic (IR). So probably better Latn Arab Cyrl.

macchiati (approved):
Ok to merge. Can do fixups later if necessary

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
conradarcturus: I know generally what you mean but I'm not sure how you want me to do this. Do you want me to avoid calling setScripts and include the code here as well? Personally I prefer DRY code over WET.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
conradarcturus: Done :)

common/supplemental/supplementalData.xml
roozbehp: Strong agreement with Mark said: despite spoken Azerbaijani being quite common in Iran with more speakers than in Azerbaijan, literacy in the writing system is pretty low in Iran, due to it not being formally taught in schools. There are books and magazines published in az-Arab in Iran, but most Iranian Azerbaijani speakers have difficulty reading them, even if they are literate in Persian (which is is theoretically a close writing system). So if anyone runs into written Azerbaijani in the wild these days, there's a much higher chance it's in Latin than in Arabic.

I also believe there's much more spoken "content" (read "media") produced in the language in Azerbaijan than in Iran (there's limited Azerbaijani radio and TV stations in Iran, and I believe less podcasts and songs produced in Iran). But either way, I don't think we want to tag such content as az-Arab or az-Latn. That's just az (or azb/azj if you insist on using ISO 639-3), and not az-Latn or az-Arab.

common/supplemental/supplementalData.xml
```diff
@@ -2198,9 +2198,9 @@ XXX Code for transations where no currency is involved
 		<language type="sel" scripts="Cyrl"/>
 		<language type="ses" scripts="Latn"/>
 		<language type="sg" scripts="Latn" territories="CF"/>
-		<language type="sga" scripts="Latn Ogam" alt="secondary"/>
+		<language type="sga" scripts="Ogam Latn" alt="secondary"/>
```
macchiati, line 2201: Does this take into account the volume of Old Irish text that is in Latn? (It may, just asking.) 
The primary goal is to reflect current usage.

macchiati (changes requested):

common/supplemental/supplementalData.xml
conradarcturus: Thanks for taking a look. I believe you are correct, the likely subtag for Old Irish [sga] is wrong, it should be `Latn` not `Ogam` -- do you want me to fix that now or as part of the task to audit primary scripts (and thereby likely subtags) https://unicode-org.atlassian.net/browse/CLDR-18121

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
macchiati: I looked this over more, sorry for being so terse.

All the data from SupplementalDataInfo needs to be immutable; that is, you must not be able to change an object you fetch from it and have that modify what SupplementalDataInfo does. 

Now, the BasicLanguageData data is collected and accessed via  languageToBasicLanguageData, which is protected via

1,206: CldrUtility.protectCollection(languageToBasicLanguageData);

So as long as BasicLanguageData is Freezable, and Freeze is correctly implemented, we are in good shape. 

However, you add a new field scriptsByPopulation, which is not protected in the freeze() method. If any process can get access to scriptsByPopulation, then it needs to be.

That is, if scriptsByPopulation can be accessed by any public call to a BasicLanguageData call OR it can be modified by any public call, then it needs to be protected in the freeze call. 

It appears that it can be, by at least one call, addScript.

Cf. @srl295 @btangmu

macchiati:
Note: I filed a related ticket https://unicode-org.atlassian.net/browse/CLDR-18224

macchiati (approved):

PR #4241, opened 2024-12-11
CLDR-18170 BRS v46.1 first pass at release note update

pedberg-icu:
CLDR-18170

This takes the body text from closed PR https://github.com/unicode-org/cldr/pull/4236, but fixes the heading line.

ALLOW_MANY_COMMITS=true


AEApple (approved):
LGTM, but should we have the date as YYYY-MM-DD and just change it when we're ready to release?

pedberg-icu:
> LGTM, but should we have the date as YYYY-MM-DD and just change it when we're ready to release?

Dec 18 was the date Mark had in his draft PR, so I kept that. We will be updating this soon anyway to flesh out the MessageFormat stuff and can update then if necessary.

PR #4233, opened 2024-12-09
CLDR-17612 cla: update text, record version

srl295:
- update text of CLA
- improve format of CLA form
- also, fix date serialization issue in UserSettings (Gson setting issue) for the signed date

CLDR-17612

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

## Screenshot
image

PR #4244, opened 2024-12-13
CLDR-18165 cla: add GitHub login

srl295:
CLDR-18165

UI:
- CLA flow first offers to login with GitHub to validate signature. See [CLA flow](https://docs.google.com/drawings/d/1ucq8SsMLMlYZekrz_9a4rAtTjDXY97zFKYUPhrqoW8E/edit?usp=sharing) and screenshots below

Servlet:
- new servlet `/cldr-apps/github-login` to receive GitHub OAuth2 flow.

Under the hood:
- add LoginManager to manage login flow, in the future this could be leveraged to enable SSO (login to SurveyTool with Github, other providers).
- add STRestClient to add a simple JSON-based REST client. Just a refactor of some common code.
- we read from `signatures.json` which is a **download** from cla-assistant.io.  We re-read this file on every successful GitHub login currently, so that this file can be updated by some process (this is part of the workflow challenge on our end).

- TODO in a future PR: document! especially GitHub app creation and setup. [this article](https://docs.github.com/en/apps/creating-github-apps/writing-code-for-a-github-app/building-a-login-with-github-button-with-a-github-app) was used as a basis, but also need to document maintenance

This PR does NOT have an admin UI to let the administrator get users 'unstuck', that is a simple matter of SQL (SMOS) currently, with an admin UI for a future PR.

ALLOW_MANY_COMMITS=true

### Review Screenshots

> [!NOTE]
> _Some screenshots may be out of date, see comments below._

#### New user readonly
- New user (not a corp signatory) when they haven't signed the CLA yet. They get a readonly experience.
image
image

#### Main menu
- Main menu item - this way we can ask people to click it for support.
image

#### Login with GitHub
- Clicking the "CLA must be signed" banner or visiting via the main menu presents this screen:
image
- Clicking Sign Manually goes to the Manual Sign page.
- Clicking Login with Github gives them the normal login flow (_the app picture of me is temporary, tells you it's not the official app, and we would say "to continue to Unicode CLDR SurveyTool"_)
image

#### Successful sign with GitHub
- If the user has signed the CLA with GitHub, GitHub redirects to the `github-login` servlet, **records the CLA signing in the database**, and then shows this page:

image

The details (name, email, employer) are pulled from the GitHub signing data but are not editable.   The "to view" link is the usual cla-assistant.io link which gives you the full text and more details.

#### What if I haven't signed with GitHub?
- Logging in with GitHub redirects you to this page if you have NOT signed the CLA:
image
- Clicking Change GitHub account gives you GitHub's account picker page, you can change accounts (would list multiple)
image
- Depending on the signatory status, you either get the Success page or loop back to the "CLA Signature Not Found" page"
- Clicking Sign Manually goes to the Manual Sign page, below
- Contact Us writes the same contact email recently established for CLDR cla.

#### Manual Sign
- See #4233 for the updates to this page
image

- One new button that takes you BACK to the Login with Github page 
image

#### Corp users
- User of an org (`ibm`) that has already signed the Corp CLA (we don't show the CLA text, because it's not this CLA that was signed.). User would only see this if they click the CLA item in the main menu.

image


srl295:
needs major rebase on #4233

srl295:
I think this is ready for review (fyi @annebright )

tools/cldr-apps/js/src/esm/cldrAuth.mjs
```diff
@@ -0,0 +1,46 @@
...
+export async function getLoginUrl(o) {
+  const { service, intent, relogin } = o;
+  if (service !== "github")
+    throw Error(`only support service='github' but got ${service}`);
+  if (intent !== "cla")
+    throw Error(`only support intent='cla' but got ${intent}`);
+  const client = await cldrClient.getClient();
+  const { url } = (await client.apis.auth.oauthUrl()).body;
+  const u = new URL(url);
+  // TODO: may move this into the GithubLoginFactory#getLoginUrl()
```
btangmu, line 20: TODO comment ideally should include ticket number

tools/cldr-apps/js/src/views/SignCla.vue
```diff
@@ -154,28 +235,73 @@ if (!user) window.location.replace("#///"); // need a user
...
+    console.error("could not getGithubIdFromSession", e);
+  }
+
   loading.value = false;
   if (unauthorized || !signed) return;
   readonlyCla.value = readonly;
   userName.value = name;
   userEmail.value = email;
   userEmployer.value = employer;
   userSign.value = corporate ? 1 : 2;
```
btangmu, line 303: Are these numbers for userSign explained somewhere? Seems like named constants might help

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ClaGithubList.java
```diff
@@ -0,0 +1,158 @@
...
+                            break; // ignore
+                    }
+                }
+                jr.endObject();
+
+                // the list is in REVERSE order (latest first). So we only track the first
+                // appearance.
+                if (!allSigners.containsKey(e.user_name)) {
+                    allSigners.put(e.user_name, e);
+                    // TODO: we could break here.
```
btangmu, line 136: again, TODO ideally has ticket number -- but in this case it's not clear whether anything needs to be done; what happens if we don't break here?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginFactory.java
```diff
@@ -0,0 +1,201 @@
...
+    // private String getClientJwt() {
+    //     final long now = System.currentTimeMillis();
+    //     return Jwts.builder()
+    //             .setSubject("CLDR SurveyTool")
+    //             .signWith(privateKey)
+    //             .setIssuedAt(new Date(now - (60 * 1000)))
+    //             .setExpiration(new Date(now + (10 * 60 * 1000)))
+    //             .setIssuer(GITHUB_CLIENT_ID)
+    //             .compact();
+    // }
```
btangmu, line 140: commented-out code, better to omit?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginFactory.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Override
+    public String getLoginUrl(LoginIntent intent) {
+        if (intent != LoginIntent.cla) {
+            throw new IllegalArgumentException("LoginIntent must be cla but got " + intent);
+        }
+        // Note: does not set redirect URI, that can be done on the front end.
+        // TODO: add 'state=…' to a temporary token.
```
btangmu, line 148: this TODO looks more serious

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginFactory.java
```diff
@@ -0,0 +1,201 @@
...
+
+        public GithubAccessTokenRequest(String code) {
+            this.code = code;
+        }
+    }
+
+    public static final class GithubUserResponse {
+        public String login;
+        public String name;
+        // others…
```
btangmu, line 173: others...?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginServlet.java
```diff
@@ -0,0 +1,105 @@
...
+public class GithubLoginServlet extends HttpServlet {
+
+    private static final long serialVersionUID = 1L;
+    private static final Logger logger = SurveyLog.forClass(GithubLoginServlet.class);
+
+    @Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+        final String code = request.getParameter("code");
+        // TODO: getParameter("state")
```
btangmu, line 22: TODO

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/LoginFactory.java
```diff
@@ -0,0 +1,23 @@
...
+        // sso,
+        /** for CLA assertion */
+        cla,
+    };
+
+    /** return true if this factory is all ready for use */
+    public abstract boolean valid();
+
+    /**
+     * Get the URL for the login link. TODO: Probably need more context here eventually.
```
btangmu, line 18: Just leave out "TODO:"; the rest of the comment is OK

btangmu (approved):
Looks great, except some TODO comments and commented-out code worry me about what might need to be done before this is ready to use in production. This PR doesn't complete the ticket and it may help to have the work in progress committed for testing, so I'm approving, with the suggestion that each TODO either gets fleshed out soon with the ticket number and clarification when/why it needs doing, or gets removed or changed into a non-TODO comment

srl295:
> Looks great, except some TODO comments and commented-out code worry me about what might need to be done before this is ready to use in production. This PR doesn't complete the ticket and it may help to have the work in progress committed for testing, so I'm approving, with the suggestion that each TODO either gets fleshed out soon with the ticket number and clarification when/why it needs doing, or gets removed or changed into a non-TODO comment

I will either flesh out or remove the TODOs.

tools/cldr-apps/js/src/views/SignCla.vue
srl295: They are only used in the radio buttons, I'm not sure if it takes a constant

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ClaGithubList.java
srl295: It means that it's already found the data item that the only current caller of this function cares about, so if it doesn't break here, it will perform unnecessary processing.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginFactory.java
srl295: It's a correct calculation, but it turned out to be not needed for our use case

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ClaGithubList.java
btangmu: so, `break`?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginFactory.java
srl295: It is a security improvement, but I think it could be in a follow up PR before going live as it does not affect the core functionality or UI

srl295: Our application doesn't give the user any options or special privileges to be performed… The only thing they can do is temporarily associate their survey to account with GitHub and then the CLA is checked, so the GitHub login is not an attack Vector for elevating privileges

I can make another subtask to track.

annebright:
Hi Steven. 
1.  I think the first screen users see when they click the CLA banner is confusing in a couple of ways. I did not realize you planned screens before a user landed on the text of the SurveyTool-only mini-CLA. I think users need to encounter the introductory language ("Hello. We need your permission...") that I put in the mini-CLA right at the start of the process, and then be given options for affirming an existing CLA or agreeing to the mini-CLA. Can we add this intro language to the first screen users encounter?
2.  I also would not use the term "sign manually" anywhere, as this suggests that the full CLA is being signed in PDF or something. Rather, what you want to do is distinguish between a "full" Unicode CLA that allows contribution to all Unicode projects and the "mini" CLA that is just for Survey Tool and does not allow contribution to other projects. For consistency, I would use the terms "Unicode CLA" and "Survey Tool only CLA" to distinguish between these two. I would make that change throughout the flow.
3.  On the screen the user gets if no signature is found in Github, I would change the radio buttons to (i) Change Github Account, (ii) Sign the full Unicode CLA, (iii) sign the Survey Tool Only CLA, (iv) Contact Us.

Hope this makes sense - would obviously be a little easier to do in a call, but not possible at the moment....

srl295:
@annebright thanks! I am going to work on the low level comments first (as those are in progress) and then get a new revision incorporating your comments.

tools/cldr-apps/js/src/esm/cldrAuth.mjs
srl295: I will change this to a see also.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/ClaGithubList.java
srl295: I'd actually rather not break here because it makes the API harder to follow. (it would be `if user.name.equals(id)` anyway).

tools/cldr-apps/src/main/java/org/unicode/cldr/web/auth/GithubLoginFactory.java
srl295: i'll just keep this issue open

srl295:
@btangmu take a look, addressed comments. Won't merge w/o approval of the flow.

srl295:
Also… it actually looks like we _might_ be able to let the SurveyTool update its CLA list itself.  The token situation might be different than I thought.

If so, then we can have the SurveyTool automatically fetch the status… and even have a "recheck" button if the user has just signed. I will investigate - as a separate PR.

srl295:
Happy new year!
This had been updated per discussion. I'm working on hot reload of the list from cla assistant as a follow on.

annebright:
Hi Steven - Is this ready for my review yet? In looking at the updates you have made to the language, I still think we need further disambiguation of the Unicode CLA and the Survey Tool Only CLA - especially on the first screen. But it might be easier to discuss in a call, so let me know when you can do one.

srl295:
> Hi Steven - Is this ready for my review yet? In looking at the updates you have made to the language, I still think we need further disambiguation of the Unicode CLA and the Survey Tool Only CLA - especially on the first screen. But it might be easier to discuss in a call, so let me know when you can do one.

(For the record) I think it is ready for review, will send you more review materials and a call offline.

macchiati (dismissed):
It looks ok to me, caveat I'm out of my depth on all of this. If you want a more thorough review, maybe Tom?

btangmu (dismissed):
Looks good based on a quick look. It's a lot of code with a variety of code paths, so there should be some testing, including penetration testing, to try to make it fail. There are still some TODO comments.

srl295:
Thanks.
Minor updates to wording:

### On the main screen

image


### On the 'github OK but signature not found' case

image

macchiati:
As Anne is ok with the wording, LGTM

macchiati (approved):

PR #4249, opened 2024-12-18
CLDR-18189 BRS 46.1 Update cldr & spec versions & status for final

pedberg-icu:
CLDR-18189

- Update readme to reflect correct version and status.
- Remove -SNAPSHOT in pom.xml files 
- Update spec status

I generated production data after this and there were no updates, so the production data tagged beta3 is still the latest (and can presumably be tagged as final once the cldr maint/maint-46 branch is tagged as final).

ALLOW_MANY_COMMITS=true


AEApple:
Should Message format in the spec say it is 'Final Candidate'? Did I miss where it says that?

srl295:
The only thing I see is that there are no references to tech preview in the document. If it said it was a tech preview before, it doesn’t now.

--
Steven R. Loomis
Code Hive Tx, LLC
https://codehivetx.us

AEApple:
Well it isn't tech preview it is now 'Final candidate'. Is there standard wording we could use similar to what is done for tech preview to say that it is 'Final candidate'?

srl295:
@pedberg-icu I pushed table of contents updates to the PR that were missing (the ToC update tool needed to be run)

pedberg-icu:
> @pedberg-icu I pushed table of contents updates to the PR that were missing (the ToC update tool needed to be run)

Ah, thanks! I had missed that.

pedberg-icu:
As for the Message Format status, in regular 46 the [spec](https://www.unicode.org/reports/tr35/tr35-messageFormat.html) says nothing about tech preview of final candidate so we should not *add* that in 46.1.

srl295 (approved):

AEApple (approved):
Okay, SGTM in that case

PR #4250, opened 2024-12-19
Update lrc.xml

hosseinabaspanah:
Hello, some characters in Northern Luri did not match with the current characters and now the necessary changes have been made...

After these changes are confirmed, I will soon translate the file with the new characters...

CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/lrc.xml
+++ b/common/main/lrc.xml
@@ -344,9 +344,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		</orientation>
 	</layout>
 	<characters>
-		<exemplarCharacters>[ٙ ٛ آ أ ؤ ئ ا ب پ ت ث ج چ ح خ د ذ ر ز ژ س ش ص ض ط ظ ع غ ف ڤ ق ک گ ل م ن ھ ە و ۉ ۊ ی ؽ]</exemplarCharacters>
+		<exemplarCharacters>[آ ٱ اٛ اْ ا ب پ ت ج چ خ د ر ز ژ س ش ع غ ف ڤ ق ک گ ڳ ل لٛ م ن ە و ؤ ۉ ۏ ۊ ۋ ی ؽ ئ ء]</exemplarCharacters>
 		<exemplarCharacters type="auxiliary">[\u200B\u200C\u200D\u200E\u200F ً ٌ ٍ َ ُ ِ ّ ْ ٔ إ ة ك ه ى ي]</exemplarCharacters>
-		<exemplarCharacters type="index">[آ ا ب پ ت ث ج چ ح خ د ذ ر ز ژ س ش ص ض ط ظ ع غ ف ق ک گ ل م ن ھ و ی]</exemplarCharacters>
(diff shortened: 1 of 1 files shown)


roozbehp:
This shows a very detailed orthography which is vastly different from the existing one we have. What is the source of the information? Please refer to dictionaries, orthography rules published by a group of Luri experts, or other such authoritative sources.

roozbehp:
> Are the projects that I have translated to Luri useful???
> 
> If they are useful, tell me and I will send you the GitHub link and their name...

Unfortunately, no. You basically need to show that your pull request represents the most widely used orthography among Northern Luri users.

As I said, we would like to see authoritative sources like dictionaries or a consensus orthography document published by a group of experts in the language. Consider providing some scans of published books, magazines, children's primers, and dictionaries in the language.

roozbehp:
> How many magazines do you need to make this request???

It's better if you provide as many different kinds of sources as you have. For example, two different magazines from different publishers, as well as the other types of references I mentioned.

roozbehp:
> Thanks for your explanation... For example, is one, two, or three books or magazines from several sources enough for you???

I can't say for sure, but that's a good start. We can examine those and tell if more information would be needed.

roozbehp:
> Have you ever been sent a document for the Luri language??? Or am I the first person to do this???

Yes, the Unicode Consortium has previously received some documents regarding the Luri language from Mohammad Mogoei and Lateef Shaikh.

roozbehp:
> my question is why are documents still needed

Because CLDR only uses either the most common or the most official orthography for a language. Since Luri does not have a formal academy we can contact, if you want to change the existing orthography, you need to show us that our existing values are wrong and the orthography you are proposing is more common.

The current orthography used by CLDR appears to be similar to the public information available about the language, such as in https://www.omniglot.com/charts/luri.pdf.

The orthography you are proposing does not seem to be as well documented on the internet.

> and more importantly, how much and for how long should these documents be made available to you?

We don't need the actual books or magazines. Just assemble photos or scans of relevant pages discussing the orthography​, provide the bibliographic information (name of author, date and place of publication, name of publisher, ISBN, etc), and send us a PDF.

roozbehp:
> Is it still necessary to submit these documents to add a new language code to the cldr collection???

It depends. If you are following the most common orthography for the language known to us, it is fine. You can create the core data specified at https://cldr.unicode.org/index/cldr-spec/core-data-for-new-locales and then create a CLDR ticket and attach the data to that.

If you have a different orthography for the language, you may need to show that your orthography is the most common orthography.

conradarcturus:
Fantastic! Thanks Kore for working to help support Bakhtiari and thanks Roozbeh for the clarifying comments. In this case we want to create a new locale, see https://cldr.unicode.org/requesting_changes#adding-new-locales .

We are in luck, another individual has recently reached out to add Bakhtiari (bqi_IR) to CLDR so this is well-timed. I just created the JIRA task to track the new locale -- but give us a little bit more time to get organized since most of us are on holiday. You can follow the task here: https://unicode-org.atlassian.net/browse/CLDR-18190

Let me know if you have any comments on that Jira ticket. Come late Q1 2025 we will open the Survey Tool and you will be able to add more Bakhtiari translations. If you don't already have one, you can request a survey tool account here:  https://corp.unicode.org/reporting/cldr-id.html .

michelleperham:
@korekasnadar We appreciate your contributions. Would you like to have your name included on the CLDR acknowledgements page? https://cldr.unicode.org/index/acknowledgments

If you'd like to be added, please provide me with your full name. You can add it here or email it to docsubmit@unicode.org

PR #4254, opened 2024-12-21
CLDR-17014 No code fallbacks for language paths

btangmu:
-Use the extra-paths mechanism instead of code-fallback for language paths

CLDR-17014

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -3535,7 +3535,7 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
     public boolean isNotRoot(String distinguishedPath) {
         String source = getSourceLocaleID(distinguishedPath, null);
         return source != null
-                && !source.equals("root")
+                && !source.equals(LocaleNames.ROOT)
                 && !source.equals(XMLSource.CODE_FALLBACK_ID);
     }
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -3148,6 +3148,10 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
     private List<String> getRawExtraPathsPrivate() {
         Set<String> toAddTo = new HashSet<>();
         SupplementalDataInfo supplementalData = CLDRConfig.getInstance().getSupplementalDataInfo();
+
+        // languages
+        getLanguageExtraPaths(toAddTo);
+
         // units
         PluralInfo plurals = supplementalData.getPlurals(PluralType.cardinal, getLocaleID());
         if (plurals == null && DEBUG) {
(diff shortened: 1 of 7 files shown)


macchiati (changes requested):
Because these are all constant paths, they can be statically initiated once, then just added whenever needed.

 I'd suggest a separate class, ExtraPaths, with one method .append to(Collection<String>)

btangmu:
> a separate class, ExtraPaths

Yes, that sounds good, in fact I was thinking to follow this up with a refactoring PR, to move the extra-path code (not just for languages, but for all extra paths) out of CLDRFile into its own top-level class. It's still not clear to me what part, if any, of extra-path code is locale-specific; the current methods aren't static. It does seem wasteful to re-run code for each locale (each CLDRFile) if it's not locale-specific. Stepping back, though, does the general solution seem right, to use extra paths instead of code fallback for language paths?

There are some test failures, like this, that I still need to study:
```
  TestCoverageLevel {
    ...
    TestCoverageCompleteness
Error:  (TestCoverageLevel.java:728)  Error: Comprehensive & no exception for path =>	//ldml/localeDisplayNames/languages/language[@type="fa_AF"]
```

macchiati:
Everything in code-fallback is locale **independent** and immutable. So that's why you can move it all to the new class, and fetch it with no local parameter. We might be able to just call that when building the root cldrfile.

The other stuff in get extra paths **is** locale dependent. That could also be moved to the new class, but needs a locale parameter when fetched.

macchiati:
> Stepping back, though, does the general solution seem right, to use extra paths instead of code fallback for language paths?

Yes, with the changes I mention.

btangmu:
The second commit fixes some but not all test failures.

The remaining test failures are TestCoverageCompleteness and testLSR, both in TestCoverageLevel. These need more investigation.

After the tests are all passing, I'll address refactoring/optimizing with a new class ExtraPaths.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -3339,6 +3343,42 @@ private List<String> getRawExtraPathsPrivate() {
         return toAddTo.stream().map(String::intern).collect(Collectors.toList());
     }
 
+    private void getLanguageExtraPaths(Set<String> toAddTo) {
+        Set<String> codes =
+                StandardCodes.make().getSurveyToolDisplayCodes(NameType.LANGUAGE.getNameName());
+        codes.remove(XMLSource.ROOT_ID);
```
btangmu, line 3349: this is wrong, my mistake -- it actually modifies the set in StandardCodes! Need to make a copy of the set. We might want to prevent this kind of bug by making getSurveyToolDisplayCodes return an unmodifiable set, maybe using [Set.copyOf](https://docs.oracle.com/javase/10/docs/api/java/util/Set.html#copyOf(java.util.Collection)).

We currently have this, which doesn't work: `CLDRLanguageCodes = CldrUtility.protectCollection(CLDRLanguageCodes);`

btangmu: Fixed by my last commit -- we still should look into Set.copyOf as an alternative to protectCollection

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
```diff
@@ -1073,7 +1075,7 @@ public void testLSR() {
 
         // Get root LSR codes
 
-        for (String path : root) {
+        for (String path : root.fullIterable()) {
```
btangmu, line 1078: without fullIterable, the loop skipped the extra paths

this may be a concern in general: how often do we loop through a CLDRFile the simple way, which skips extra paths, and what are the consequences and rationale for that?

btangmu:
The test failures all seem fixed now except for TestVettingDataDriven

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
macchiati: Important: In general, use ImmutableSet.copyOf instead of Set.copyOf. The latter changes the order in the set, which may be important (TreeSet or LinkedHashSets get messed up, and it doesn't hurt for HashSet).

macchiati: Also, we need to fix CldrUtility.protectCollection(CLDRLanguageCodes); — if it does't work then a lot of other things could go wrong.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
macchiati: Good point. 

I think in most cases we want to iterate through the extra paths. So we might want make the iteration choice explicit, so that we can check to make sure. Maybe refactor (in separate ticket/PR) as follows.

```
for (String path : cldrFile) { // make this fail

for (String path : cldrFile.withoutExtras()) { // make this do what the line above did

for (String path : cldrFile.withExtras()) { // new name for fullIterable
```

Then we can search for the withoutExtras() calls and make sure that they are right.

btangmu: The TestVettingDataDriven failure may be related to this. Part of that test involves writing the CLDRFile to a temporary XML file and then reading it back to verify the newly voted-on value, and that's where it's failing. It looks like the language paths, which are now extra paths, don't get written to the file. To test that, I'm starting a new branch from main that will just have a few lines added to TestSTFactory.xml...

btangmu: See https://github.com/unicode-org/cldr/pull/4256

If the same failure happens there that I see locally, it seems to indicate a long-existing bug where CldrXmlWriter doesn't write extra paths

macchiati:
Well, it mustn't write the extra paths, because the have null values, which
cannot be serialized.

With the changes to code-fallback to not have values, there is a difference
in behavior. No surprise that that will require changes.

btangmu:
> it mustn't write the extra paths, because the have null values, which cannot be serialized

I wouldn't expect extra paths with null values to be written to disk, but I did expect extra paths with non-null values to be written to disk. The test failures here and in https://github.com/unicode-org/cldr/pull/4256 show that an empty path, originally having a null value (or no value), can get a value through voting, and the CLDRFile is updated successfully in memory, but it fails to get written to disk. The failure is happening in https://github.com/unicode-org/cldr/pull/4256 even though there is no change in that PR except for the test itself -- it's for a metazone path and doesn't involve any change to language paths, code fallback, etc.

The path `//ldml/dates/timeZoneNames/metazone[@type="Alaska"]/long/generic` does have a value in common/main/fr.xml for example. Mayb it's no longer called an "extra path" at that point; maybe part of my confusion is about terminology. Anyway I'm wondering how it got into fr.xml if not by voting followed by `CldrXmlWriter.write`.

macchiati:
Hmmm. Yes, the extra paths are no longer 'extra' once they have values. Not
sure why they are failing

>  to get written to disk.

There are plenty of other cases in the existing code that worked fine in
the past.

Guess the only recourse is to walk through what's happening in a debugger.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
btangmu: @macchiati following up, I made a new ticket about fullIterable: https://unicode-org.atlassian.net/browse/CLDR-18217

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
btangmu: New ticket https://unicode-org.atlassian.net/browse/CLDR-18218

btangmu:
I'm leaving this PR here for reference for now, but a newer PR is better, handling script as well as language: https://github.com/unicode-org/cldr/pull/4290

PR #4255, opened 2024-12-22
CLDR-11874 Add conversion of subdivisions to json

arjanmels:
CLDR-11874

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
@@ -102,7 +102,8 @@ public class Ldml2JsonConverter {
         annotations,
         annotationsDerived,
         bcp47(false, false),
-        transforms(false, false);
+        transforms(false, false),
+        subdivisions(false, true);
 
         private final boolean isTiered;
         private final boolean hasLocales;
@@ -748,6 +749,8 @@ public class Ldml2JsonConverter {
                 outFilename = filename + ".json";
             } else if (type == RunType.transforms) {
                 outFilename = filename + ".json";
+            } else if (type == RunType.subdivisions) {
+                outFilename = filename + ".json";
             } else if (js.section.equals("other")) {
                 // If you see other-___.json, it means items that were missing from
                 // JSON_config_*.txt
@@ -786,7 +789,8 @@ public class Ldml2JsonConverter {
                         }
                     } else if (type == RunType.rbnf
                             || type == RunType.bcp47
-                            || type == RunType.transforms) {
+                            || type == RunType.transforms
+                            || type == RunType.subdivisions) {
                         // untiered, just use the name
                         js.packageName = type.name();
                         tier = "";
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
@@ -18,8 +18,7 @@ class LdmlConvertRules {
 
     /** File sets that will not be processed in JSON transformation. */
     public static final ImmutableSet<String> IGNORE_FILE_SET =
-            ImmutableSet.of(
-                    "attributeValueValidity", "coverageLevels", "postalCodeData", "subdivisions");
+            ImmutableSet.of("attributeValueValidity", "coverageLevels", "postalCodeData");
 
     /**
      * The attribute list that should become part of the name in form of name-(attribute)-(value).
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/json/JSON_config_subdivisions.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/json/JSON_config_subdivisions.txt
@@ -1,0 +1,10 @@
+# Processed by LdmlConfigFileReader
+
+# These are earlier in the file to keep the packageDesc in one place.
+package=subdivisions ; packageDesc=Translated versions of locale display name elements: subdivisions.
+
+# mapping paths to sections
+section=subdivisions ; path=//cldr/subdivisions/[^/]++/localeDisplayNames/subdivisions/.* ; package=subdivisions
+
+# Depedencies
+dependency=core ; package=subdivisions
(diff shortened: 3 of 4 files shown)


arjanmels:
Also added containment for subdivisions.

srl295 (commented):
Code looks great need to test it

srl295:
```json
        "autas": "Tasmania",
        "auvic": "Victoria",
        "auwa": "Western Australia",
        "AW": "Aruba",
        "AX": "Åland Islands",
        "azabs": "Absheron",
        "azaga": "Agstafa",
        "azagc": "Aghjabadi",
        "azagm": "Agdam",
```

looks like it includes the *territories* also, not just subdivisions.

srl295:
## Files generated

### `cldr-core/supplemental/subdivisionContainment.json`

```json
{
  "supplemental": {
    "version": {
      "_unicodeVersion": "16.0.0",
      "_cldrVersion": "46"
    },
    "subdivisionContainment": {
      "AD": {
        "_contains": [
          "ad02",
          "ad03",
          "ad04",
          "ad05",
          "ad06",
          "ad07",
          "ad08"
        ]
      },
…
```

### cldr-subdivisions/subdivisions/en.json

```json
{
  "subdivisions": {
    "localeDisplayNames": {
      "subdivisions": {
        "ad02": "Canillo",
        "ad03": "Encamp",
        "ad04": "La Massana",
        "ad05": "Ordino",
        "ad06": "Sant Julià de Lòria",
        "ad07": "Andorra la Vella",
        "ad08": "Escaldes-Engordany",
        "aeaj": "Ajman",
        "aeaz": "Abu Dhabi",
        "aedu": "Dubai",
        "aefu": "Fujairah",
        "aerk": "Ras al-Khaimah",
        "aesh": "Sharjah",
        "aeuq": "Umm al-Quwain",
        "afbal": "Balkh",
…
```


Oddly it only generated en.json and ha.json - do you see something different?

I think those are the only two that aren't provisional.

But also it seems like the common/main and common/subdivision trees aren't being merged together properly, because we should see this data from French for example

```xml
                <subdivisions>
                        <subdivision type="gbeng">Angleterre</subdivision>
                        <subdivision type="gbsct">Écosse</subdivision>
                        <subdivision type="gbwls">Pays de Galles</subdivision>
                </subdivisions>
```

arjanmels:
> Oddly it only generated en.json and ha.json - do you see something different?
> I think those are the only two that aren't provisional.

You're right for my local builds I set: ```DRAFTSTATUS=provisional``` in my local-config.sh

arjanmels:
> looks like it includes the _territories_ also, not just subdivisions.

This is actually already in the source .xml. Not all territories are included (e.g. "NL", "FR", "GB), it seems onlt territories, which are NOT countries. Anyway the json is a direct reproduction of the xml.

srl295:
@macchiati indeed it does seem like this is in the source - is this a bug?

https://github.com/unicode-org/cldr/blob/main/common/subdivisions/en.xml#L211-L212

macchiati:
> ```json
>         "autas": "Tasmania",
>         "auvic": "Victoria",
>         "auwa": "Western Australia",
>         "AW": "Aruba",
>         "AX": "Åland Islands",
>         "azabs": "Absheron",
>         "azaga": "Agstafa",
>         "azagc": "Aghjabadi",
>         "azagm": "Agdam",
> ```
> 
> looks like it includes the _territories_ also, not just subdivisions.

I think what is happening is that some subdivisions are aliases of country codes. We're treating the country codes as predominant (we'll have better data for them). For the JSON version, I think we can skip the country codes; implementers are on notice that they should use the country codes.

srl295:
> I think what is happening is that some subdivisions are aliases of country codes. We're treating the country codes as predominant (we'll have better data for them). For the JSON version, I think we can skip the country codes; implementers are on notice that they should use the country codes.

The XML has

```xml
			<subdivision type="AW">Aruba</subdivision>
			<subdivision type="AX">Åland Islands</subdivision>
			<!-- AW : Aruba : NO SUBDIVISIONS
				 AX : Åland Islands : NO SUBDIVISIONS
				 AZ : Azerbaijan
			-->
```

The spec doesn't address it.  It looks like it's just included incorrectly.

The spec by the way at <https://unicode.org/reports/tr35/dev/tr35-general.html#Contents> has dashes in the subdivision codes, and doesn't mention the special subdivisions in common/main.

macchiati:
> The spec doesn't address it. It looks like it's just included incorrectly.

That's a bug. We'll want to exclude the country codes from the XML data (and via that, the JSON), and also modify the spec to make it clear that implementations need to look at the subdivision aliases, such as
```
<subdivisionAlias type="usgu" replacement="GU" reason="overlong"/> <!-- Guam => Guam -->
```

> The spec by the way at https://unicode.org/reports/tr35/dev/tr35-general.html#Contents has dashes in the subdivision codes, and doesn't mention the special subdivisions in common/main.

I was confused at first, because the TOC doesn't contain any subdivision codes. But I think you mean https://unicode.org/reports/tr35/dev/tr35-general.html#locale-display-name-fields, notably 
```
<subdivision type="AL-04">Fier County</subdivision>
<subdivision type="AL-FR">Fier</subdivision> <!-- in AL-04 : Fier County -->
<subdivision type="AL-LU">Lushnjë</subdivision> <!-- in AL-04 : Fier County -->
<subdivision type="AL-MK">Mallakastër</subdivision> <!-- in AL-04 : Fier County -->
```
Is that what you meant?

Anyway, can you file a ticket for these two items? (They are related since we'll want to change the spec to document the relation between the subdivisions and codes.

srl295:
@arjanmels Can you look into:
- supporting subdivisionAlias in the supplemental data (see above) from supplementalMetaData.xml - see https://unicode.org/reports/tr35/dev/tr35-info.html#supplemental-alias-information
- I think the Arbua issue is a data bug as mentioned, so does not need to be suppressed in this PR.

srl295:
For the data issue: https://unicode-org.atlassian.net/browse/CLDR-18219 - looks like this was fixed before but is a regression

srl295:
@arjanmels OK, sorry i missed it.

I'm going to approve this and look at the coverage issue separately

srl295 (approved):

arjanmels:
@srl295 great, thank you!

PR #4261, opened 2025-01-01
CLDR-16836 kbd: add EBNF to spec for transforms

srl295:
CLDR-16836

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -2428,7 +2428,6 @@ The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
 ```ebnf
 [ wfc: No more than 9 capture groups may be present. ]
 [ vc: all variables referenced must be defined in the <variables> element ]
-[ vc: The CLDR repository may define additional constraints on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters. ]
 
 from-match
          ::= '^'? atoms
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -2427,7 +2427,7 @@ The `from=` attribute MUST match the `from-match` rule in this grammar. Not all 
 * The CLDR repository may define additional requirements on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters.
 * It is an error to reference variables that are not defined.
 
-The following is the W3C EBNF format for the grammar:
+The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
 
 ```ebnf
 
@@ -2587,7 +2587,7 @@ This is the grammar for the `<transform to="…"/>` attribute.  The `to=` attrib
 * The CLDR repository may define additional requirements on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters.
 * It is an error to reference variables that are not defined.
 
-The following is the W3C EBNF format for the grammar:
+The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
 
 ```ebnf
 to-replacement
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -2420,16 +2420,15 @@ Used in the `to=`
 
 ##### Transform From Grammar
 
-The `from=` attribute MUST match the `from-match` rule in this grammar. Not all strings which match this grammar are valid:
-
-* It is an error if there are more than 9 capture groups
-* Unicode codepoints and escaped characters may not be invalid or unpaired surrogates
-* The CLDR repository may define additional requirements on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters.
-* It is an error to reference variables that are not defined.
+The `from=` attribute MUST match the `from-match` rule in this grammar. Not all strings which match this grammar are valid, specifically
 
 The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
 
+
 ```ebnf
+[ wfc: No more than 9 capture groups may be present. ]
+[ vc: all variables referenced must be defined in the <variables> element ]
+[ vc: The CLDR repository may define additional constraints on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters. ]
 
 from-match
          ::= '^'? atoms
@@ -2581,15 +2580,14 @@ NMTOKEN  ::= NAMECHAR+
 
 This is the grammar for the `<transform to="…"/>` attribute.  The `to=` attribute MUST match the `to-replacement` rule in this grammar. Not all strings which match this grammar are valid:
 
-* It is an error if a capture group is referenced that is not present in the match string.
-* It is an error if the to= string has the `$[1:…]` set format but there is not exactly one capture group with a set variable on the from= side. See [Replacement syntax](#replacement-syntax).
-* Unicode codepoints and escaped characters may not be invalid or unpaired surrogates
-* The CLDR repository may define additional requirements on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters.
-* It is an error to reference variables that are not defined.
-
 The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
 
 ```ebnf
+[ vc: A referenced capture group must be present in the from= match string. ]
+[ vc: The `$[1:…]` set format may only be used where there is exactly one capture group with a set variable on the from= match string. ]
+[ vc: all variables referenced must be defined in the <variables> element ]
+[ vc: The CLDR repository may define additional constraints on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters. ]
+
 to-replacement
          ::= atoms
 atoms    ::= atom*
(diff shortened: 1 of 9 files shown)


srl295:
fyi @stasm and @aphillips

srl295:
TODOs:
 - NMTOKEN is a little incomplete - had a tool problem with high values.
 - Need to do the to= side

macchiati (changes requested):
Looks good!
1. The EBNF links should point to https://unicode.org/reports/tr35/#ebnf. (Since the LDML EBNF is a superset of the vanilla EBNF that should work.
2. The validity / well-formedness constraints (eg no more than 9 capture groups), should use the w3c syntax.

srl295:
> Looks good!
> 
> 1. The EBNF links should point to https://unicode.org/reports/tr35/#ebnf. (Since the LDML EBNF is a superset of the vanilla EBNF that should work.

I will add this:

```markdown
The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
```

> 2. The validity / well-formedness constraints (eg no more than 9 capture groups), should use the w3c syntax.

The 9 capture groups is 9 *inner* capture groups.  So for example, valid:

```js
/(first)(second)(third)(fourth)(fifth)(sixth)(seventh)(eighth)(?:And possibly, (ninth))?/
```

but invalid:

```js
/(first)(second)(third)(fourth)(fifth)(sixth)(seventh)(eighth)(?:And possibly, (ninth))?(?:But not, (tenth)!)?/
```

because it's a nested group it may be challenging to express in the grammar, it's more like a resource (slot) limit.

srl295:
I'd like to get some Java based tooling, but that will be in a separate PR.

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,287 @@ Used in the `to=`
...
     Emits the named mark. Also see [Markers](#markers).
 
+#### Transform Grammar
+
+##### Transform From Grammar
+
+The `from=` attribute MUST match the `from-match` rule in this grammar. Not all strings which match this grammar are valid:
+
+* It is an error if there are more than 9 capture groups
+* Unicode codepoints and escaped characters may not be invalid or unpaired surrogates
```
miloush, line 2426: too many negatives, did you mean "may be invalid"?

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,287 @@ Used in the `to=`
...
+string-variable
+         ::= '${' var-id '}'
+set-variable
+         ::= '$[' var-id ']'
+group    ::= capturing-group
+           | non-capturing-group
+quantifier
+         ::= bounded-quantifier
+           | '?'
+codepointseq
```
miloush, line 2455: how is `codepointseq` different from `codepoint`?

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,287 @@ Used in the `to=`
...
+         ::= simple-matcher
+           | codepointseq
+           | variable
+variable ::= string-variable
+           | set-variable
+string-variable
+         ::= '${' var-id '}'
+set-variable
+         ::= '$[' var-id ']'
+group    ::= capturing-group
```
miloush, line 2450: Is there a supposed ordering to the grammar? It feels like the idea is to use first, define next, on depth-first basis, but then why is this not followed by `var-id` and then `codepointseq`? I don't mind whether its define-then-use or use-then-define or alphabetical, but it would help if it was systematic.

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,287 @@ Used in the `to=`
...
+         ::= '\' ( '\' | '{' | '|' | '}' )
+ws       ::= [ #x3000]
+           | HTAB
+           | CR
+           | LF
+IDCHAR   ::= ALPHA
+           | DIGIT
+           | '_'
+ASCII-CTRLS
+         ::= [#x1-#x8#xB-#xC#xE-#x1F]
```
miloush, line 2547: these are not allowed in XML 1.0

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,287 @@ Used in the `to=`
...
+           | '.'
+           | match-marker
+match-marker
+         ::= '\m{.}'
+           | match-named-marker
+match-named-marker
+         ::= '\m{' marker-id '}'
+marker-id
+         ::= NMTOKEN
+var-id   ::= IDCHAR+
```
miloush, line 2482: why are variables limited to ASCII?

docs/ldml/tr35-keyboards.md
macchiati: That's why I mentioned the w3c well-formedness / validity constraints. We should not talk about errors, but rather constraints:
```
[ wfc: No more than 9 capture groups are allowed. ]
[ wfc: Surrogate code points are not allowed. ]
```
Note: the 2nd one can also be expressed in the BNF.

docs/ldml/tr35-keyboards.md
srl295: I think that's in the spec

srl295:
Yes, my comment about surrogates is redundant because they are already
covered in the BNF. As to errors, what I was trying to express is that an
implementation would reject a keyboard layout with an error… But I can
refer to that as a constraint for consistency

docs/ldml/tr35-keyboards.md
srl295: The spec allows for a space separated list of code points… However, where specifying a character class range start or end, only a single code point must be allowed, that's why there are two different rules. 

I'll clarify this.

docs/ldml/tr35-keyboards.md
srl295: There isn't a specific order. I can put them in some better order.

macchiati (changes requested):
I requested changes for the constraints.

docs/ldml/tr35-keyboards.md
srl295: I have moved all of this  into wfc / vc

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,282 @@ Used in the `to=`
...
+
+The `from=` attribute MUST match the `from-match` rule in this grammar. Not all strings which match this grammar are valid, specifically
+
+The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
+
+
+```ebnf
+[ wfc: No more than 9 capture groups may be present. ]
+[ vc: all variables referenced must be defined in the <variables> element ]
+[ vc: The CLDR repository may define additional constraints on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters. ]
```
macchiati, line 2431:
```suggestion
[ vc: If a keyboard definitions is submitted to the the CLDR repository, it must satisfy additional constraints on the character repertoire. For more information, see [CLDR keyboard repertoire constraints](#repertoire-constraints). ]
```
We have to point to where those constraints are documented. So add a little section header for that and point to it. Also make a change to a similar line below.

I suggest that the contents of that section be:

- No characters can have any of the following General_Category values in the latest version of the Unicode Standard: 
    - Private Use (Co)
    - Surrogate (Cs)
    - Unassigned (Cn)

macchiati (changes requested):

docs/ldml/tr35-keyboards.md
srl295: Not sure we can link from ebnf to spec. 

Also the repository requirements are a separate activity - I don't see a reason to specify them here in detail (though I'll take the list back to kbd-wg). Perhaps it's better to just say- see a future version of the spec.

srl295: What about this

```ebnf
[ vc: If a keyboard definitions is submitted to the the CLDR repository, it must satisfy additional constraints on the character repertoire. ]
```

macchiati: The problem with that is there is no way for the reader to find out what those requirements are — we can't have undefined vc requirements.

srl295: Then I'd propose to drop it from vc and just note.

srl295, line 2431:
```suggestion
```

docs/ldml/tr35-keyboards.md
```diff
@@ -2412,6 +2416,282 @@ Used in the `to=`
...
+
+This is the grammar for the `<transform to="…"/>` attribute.  The `to=` attribute MUST match the `to-replacement` rule in this grammar. Not all strings which match this grammar are valid:
+
+The following is the [LDML EBNF](./tr35.md#ebnf) format for the grammar:
+
+```ebnf
+[ vc: A referenced capture group must be present in the from= match string. ]
+[ vc: The `$[1:…]` set format may only be used where there is exactly one capture group with a set variable on the from= match string. ]
+[ vc: all variables referenced must be defined in the <variables> element ]
+[ vc: The CLDR repository may define additional constraints on the repertoire, such as requiring all characters to be in a published Unicode version and disallowing private-use characters. ]
```
srl295, line 2588:
```suggestion
```

macchiati: Ok.

docs/ldml/tr35-keyboards.md
macchiati: You can go ahead and commit these two suggestions for deleting.

macchiati (approved):

srl295:
@miloush how does it look?

PR #4270, opened 2025-01-09
CLDR-16836 kbd: ebnf:  move tests to js out of shell, check XML files

srl295:
CLDR-16836

- also, test XML files for validity of from/to

Followon to #4261 this moves the test case out of shell script into its own JS project. No changes to EBNF or test cases.

ALLOW_MANY_COMMITS=true


tools/scripts/keyboard-abnf-tests/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `npm t`
+
+## To update
+
+Note there are four files. There's a `.d` directory for each ABNF file in keyboards/abnf/.  The "pass" files are expected to pass the ABNF and the "fail" to fail it.  Lines beginning with # are comments and skipped.
+
+- transform-from-required.d/from-match.pass.txt
+- transform-from-required.d/from-match.fail.txt
+- transform-to-required.d/to-replacement.pass.txt
+- transform-to-required.d/to-replacement.fail.txt
```
btangmu, line 17: where are these files?

btangmu (commented):
this looks great but I don't see the 4 files referred to, 'from-match.pass.txt' etc.; probably I'm not understanding something

tools/scripts/keyboard-abnf-tests/README.md
srl295: They are in that same directory (keyboard-abnf-tests) and were added in #4261

srl295: that previous PR had a shell script (now deleted) to test the files, this changes it to JS

btangmu (approved):
Hurray for js replacing sh!

PR #4271, opened 2025-01-09
CLDR-18212 Update copyright to 2025

conradarcturus:
Ran the following scripts to re-generate the year in copyright stamps or update

```
mvn package -DskipTests=true
java -jar tools/cldr-code/target/cldr-code.jar GenerateDtd
java -jar tools/cldr-code/target/cldr-code.jar GenerateTestData
java -jar tools/cldr-code/target/cldr-code.jar GenerateValidityXML
java -jar tools/cldr-code/target/cldr-code.jar CLDRModify -d common/main/
```

Unfortunately `GenerateDerivedAnnotations` did not work on my computer so I did not run that.

All other files updated with  `sed 's/© 1991-2024/© 1991-2025/g' *`

CLDR-18212

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

srl295 (dismissed):
we do not need to bump all copyrights

srl295:
> All other files updated with  sed 's/© 1991-2024/© 1991-2025/g' *

We don't need to update ALL files!! Only files that actually change.

srl295:
we don't need to update the © on generated files unless they are actually regenerated.

We do need to update the top README copyright. but not every single copyright.

srl295 (approved):
I'm RSLGTM for now.

I think in the future it would be better to fix all tools to only bump copyright if something actually changes, and otherwise revert.   That would help us see more accurately if a file really changes

conradarcturus:
@srl295 Yea, that makes sense. It would be good to add "only update year if file changed" to the generating output is different. IMHO it also would be good to add checksums to generated files to help protect against manual changes (the changes should be overrides during generation).

The reason I preferred doing this (bulk copyright update) is that these scripts today will update copyrights all over the place if we run them, polluting other PRs. This isolates the unrelated changes in 1 PR. However I understand that we don't need to update the copyright if we have not updated the code inside...

PR #4273, opened 2025-01-14
CLDR-18220 Dashboard Other category for all other paths

srl295:
CLDR-18220

- Adds a new notification category 'other'. It's not included in the default choice sets.
- The query param includeOther=true adds this category.
- UI in the dashboard makes the Other category look like similar categories, except that
  internally it reloads with the query param set/unset. (This reduces load in the normal case.)

The benefit here is especially in being able to download a spreadsheet of all values in a coverage
level.

### Unchecked

image

### Checked

image

### Downloaded

image

[Dash_csw_default(2).xlsx](https://github.com/user-attachments/files/18414370/Dash_csw_default.2.xlsx)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/DashboardWidget.vue
+++ b/tools/cldr-apps/js/src/views/DashboardWidget.vue
@@ -55,6 +55,13 @@
           </template>
         </span>
         <span class="right-control">
+          <span
+            v-if="!includeOther"
+            class="cldr-nav-btn"
+            @click="reloadIncludeOther"
+          >
+            <input type="checkbox" />&nbsp;Other
+          </span>
           <a-spin v-if="downloadMessage">
             <i>{{ downloadMessage }}</i>
           </a-spin>
@@ -214,6 +221,7 @@ export default {
       catCheckboxIsUnchecked: {}, // default unchecked = false, checked = true
       catIsHidden: {}, // default hidden = false, visible = true
       updatingVisibility: false,
+      includeOther: false,
     };
   },
 
@@ -287,6 +295,16 @@ export default {
       this.fetchData();
     },
 
+    reloadIncludeOther() {
+      this.catIsHidden["Other"] = false;
+      this.catCheckboxIsUnchecked["Other"] = false;
+      this.includeOther = true;
+      this.reloadDashboard();
+    },
+    reloadWithoutOther() {
+      this.includeOther = false;
+      this.reloadDashboard();
+    },
     fetchData() {
       this.locale = cldrStatus.getCurrentLocale();
       this.level = cldrCoverage.effectiveName(this.locale);
@@ -298,7 +316,7 @@ export default {
       }
       this.localeName = cldrLoad.getLocaleName(this.locale);
       this.loadingMessage = `Loading ${this.localeName} dashboard at ${this.level} level`;
-      cldrDashData.doFetch(this.setData);
+      cldrDashData.doFetch(this.setData, { includeOther: this.includeOther });
       this.fetchErr = cldrDashData.getFetchError();
     },
 
@@ -434,6 +452,14 @@ export default {
       // long delay between the time the user clicks the checkbox and the time that the checkbox
       // changes its state.
       // NOTE: this complication may be unnecessary now that DashboardScroller is in use.
+      if (category === "Other") {
+        // special case: Unchecking Other reloads without Other.
+        nextTick().then(() => {
+          this.reloadWithoutOther();
+        });
+        return;
+      }
+
       this.catCheckboxIsUnchecked[category] = !event.target.checked; // redundant?
       const USE_NEXT_TICK = true;
       this.updatingVisibility = true;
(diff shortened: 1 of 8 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/NotificationCategory.java
```diff
@@ -57,7 +57,8 @@ public enum NotificationCategory {
 
     /** You have abstained, or not yet voted for any value */
     abstained('A', "Abstained", "You have abstained, or not yet voted for any value."),
-    ;
+
+    other('O', "Other", "All other values");
```
conradarcturus, line 61: Since it's without errors, warnings and it isn't abstained, that this group should be thought of as "voted, no issues"? To keep the name minimal we can call it other, but I wonder if we should make it clearer.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Summary.java
```diff
@@ -551,7 +561,7 @@ public Response getParticipationFor(
             coverageLevel = org.unicode.cldr.util.Level.fromString(level);
         }
         ReviewOutput reviewOutput =
-                new Dashboard().get(loc, target, coverageLevel, null /* xpath */);
+                new Dashboard().get(loc, target, coverageLevel, null /* xpath */, false);
```
conradarcturus, line 564: nit: never add a boolean without a comment what it is 
```suggestion
                new Dashboard().get(loc, target, coverageLevel, null /* xpath */, false /* includeOther */);
```

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
```diff
@@ -402,7 +402,7 @@ private static Dashboard.ReviewNotification[] getOnePathDash(
             String baseXp, CLDRLocale locale, CookieSession mySession) {
         Level coverageLevel = Level.get(mySession.getEffectiveCoverageLevel(locale.toString()));
         Dashboard.ReviewOutput ro =
-                new Dashboard().get(locale, mySession.user, coverageLevel, baseXp);
+                new Dashboard().get(locale, mySession.user, coverageLevel, baseXp, false);
```
conradarcturus, line 405: Same nit
```suggestion
                new Dashboard().get(locale, mySession.user, coverageLevel, baseXp, false /* includeOther */);
```

tools/cldr-apps/js/src/views/DashboardWidget.vue
```diff
@@ -55,6 +55,13 @@
           </template>
         </span>
         <span class="right-control">
+          <span
+            v-if="!includeOther"
+            class="cldr-nav-btn"
+            @click="reloadIncludeOther"
+          >
+            <input type="checkbox" />&nbsp;Other
```
btangmu, line 63: suggestion for future improvement: hover text for this control with brief explanation what it does, e.g., "Include notifications for ALL paths within the chosen coverage level that do not belong to another notification category. Note: the number of 'other' notifications may be very large"

conradarcturus (approved):
Fantastic! I am glad this was straightforward.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Summary.java
srl295: I'm used to getting that filled in by the IDE but sounds good

image

btangmu (approved):
This seems like a very useful feature for those who need it, when they need it. On the other hand it has a bit of a "creeping featurism" aspect to it, complicating the interface with something that most (?) users might not use. Is the benefit mainly for downloading, rather than scrolling in the Dashboard itself? If so, it might make sense to apply it only to downloading, and make it an option in the download interface only. Performance may also be a concern, especially if it takes a long time to update the dashboard after a user clicks on the checkbox when they're merely curious to see what happens

tools/cldr-code/src/main/java/org/unicode/cldr/util/NotificationCategory.java
srl295: - It's not 'voted' if the user's logged out.
- How about just "No Issues"?

srl295:
> This seems like a very useful feature for those who need it, when they need it. On the other hand it has a bit of a "creeping featurism" aspect to it, complicating the interface with something that most (?) users might not use. Is the benefit mainly for downloading, rather than scrolling in the Dashboard itself? If so, it might make sense to apply it only to downloading, and make it an option in the download interface only. Performance may also be a concern, especially if it takes a long time to update the dashboard after a user clicks on the checkbox when they're merely curious to see what happens

That's an idea.  
Yeah, the reason for the UI distinction is that this category could have 20k items in it that would otherwise not need to be visible.

The Download button currently operates off the data you've already downloaded, BUT, what we could do is have it always do a fresh download (wouldn't be too bad…) 


Suggestion is to merge this for today (applying suggestions… for this week) and review if need be.  I think it stays out of the way of the normal case

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/Summary.java
conradarcturus: Ah! If you have a moment during the Stanford visit I'd love for your help setting up my IDE to for things like this. Do you use Eclipse or VSCode?

srl295: VS code with the red hat, Java package

PR #4279, opened 2025-01-15
CLDR-18226 Fix und-Ethi-t-und-latn-m0-beta_metsehaf-geminate

robertbastian:
CLDR-18226

The reference is missing an underscore.

ALLOW_MANY_COMMITS=true


common/transforms/und-Ethi-t-und-latn-m0-beta_metsehaf-geminate.xml
```diff
@@ -19,7 +19,7 @@
 ########################################################################
 #
 :: Amharic-Amharic/Geminate ; 
-:: Ethiopic-Latin/BetaMetsehaf ;
+:: Ethiopic-Latin/Beta_Metsehaf ;
```
btangmu, line 22: This is needed for consistency with the related file und-Ethi-t-und-latn-m0-beta_metsehaf.xml

btangmu (approved):

PR #4280, opened 2025-01-15
CLDR-18217 Refactor by removing Iterable interface from CLDRFile

btangmu:
-Instead of Iterable interface, use new methods iterableDefault, iterableWithoutExtras

-Private boolean CLDRFile.DEFAULT_ITERABLE_INCLUDES_EXTRAS determines whether iterableDefault calls fullIterable or iterableWithoutExtras

-Call iterableWithoutExtras directly where tests would otherwise fail with DEFAULT_ITERABLE_INCLUDES_EXTRAS true

-Rename original iterator method to iteratorWithoutExtras

-Iterable and iterator are distinct: iterableWithoutExtras references iteratorWithoutExtras

-There is not yet any iteratorDefault (only iterableDefault)

CLDR-18217

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/CompareCLDRFile.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/CompareCLDRFile.java
@@ -22,7 +22,7 @@ public class CompareCLDRFile {
             this.file = f;
             this.title = t;
 
-            for (String s : f) {
+            for (String s : f.iterableDefault()) {
                 xpaths.add(s);
             }
 
--- a/tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSTFactory.java
+++ b/tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSTFactory.java
@@ -73,7 +73,7 @@ public class TestSTFactory {
         STFactory fac = getFactory();
         CLDRFile mt = fac.make(locale, false);
         BallotBox<User> box = fac.ballotBoxForLocale(locale);
-        mt.iterator();
+        mt.iteratorWithoutExtras();
         final String somePath = "//ldml/localeDisplayNames/keys/key[@type=\"collation\"]";
         box.getValues(somePath);
     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRFilePseudolocalizer.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRFilePseudolocalizer.java
@@ -403,7 +403,7 @@ public class CLDRFilePseudolocalizer {
         // Create input CLDRFile object resolving inherited data.
         CLDRFile input = factory.make(ORIGINAL_LOCALE, false);
         XMLSource outputSource = new SimpleXMLSource(outputLocale);
-        for (String xpath : input) {
+        for (String xpath : input.iterableWithoutExtras()) {
             String fullPath = input.getFullXPath(xpath);
             String value = input.getStringValue(xpath);
             if (!value.isEmpty()) {
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -98,7 +98,7 @@ import org.xml.sax.helpers.XMLReaderFactory;
  * http://java.sun.com/j2se/1.4.2/docs/api/org/xml/sax/DTDHandler.html
  */
 
-public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleStringProvider {
+public class CLDRFile implements Freezable<CLDRFile>, LocaleStringProvider {
 
     private static final ImmutableSet<String> casesNominativeOnly =
             ImmutableSet.of(GrammaticalFeature.grammaticalCase.getDefault(null));
@@ -548,7 +548,7 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
 
     /** Prints the contents of the file (the xpaths/values) to the console. */
     public CLDRFile show() {
-        for (Iterator<String> it2 = iterator(); it2.hasNext(); ) {
+        for (Iterator<String> it2 = iteratorWithoutExtras(); it2.hasNext(); ) {
             String xpath = it2.next();
             System.out.println(getFullXPath(xpath) + " =>\t" + getStringValue(xpath));
         }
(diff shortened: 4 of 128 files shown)


btangmu:
This draft PR is intended for investigation, not (at least not yet) for merging.

Note that it has DEFAULT_ITERABLE_INCLUDES_EXTRAS true. If DEFAULT_ITERABLE_INCLUDES_EXTRAS were false it would merely be a refactoring of the current code, and the tests would pass trivially (assuming the refactoring is only that).

With DEFAULT_ITERABLE_INCLUDES_EXTRAS, it was necessary to change some iterableDefault into iterableWithoutExtras in order to make the tests pass.

It's an open question, when extra paths should be included in a CLDRFile. See https://unicode-org.atlassian.net/browse/CLDR-17014 for motivation for believing the default should be that extra paths are included. The first PR for that ticket fixed a pre-existing bug involving metazone paths (which were already implemented using extra paths). There are probably more such bugs, not yet discovered, which could be fixed by making DEFAULT_ITERABLE_INCLUDES_EXTRAS true.

btangmu:
By the way, the versions of `iteratorWithoutExtras` that have one or more parameters do not yet have any corresponding `iteratorDefault` methods. So far (first commit), this PR only investigates the methods without parameters.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/CompareCLDRFile.java
```diff
@@ -22,7 +22,7 @@ private Entry(CLDRFile f, String t) {
             this.file = f;
             this.title = t;
 
-            for (String s : f) {
+            for (String s : f.iterableDefault()) {
```
btangmu, line 25: This is the typical situation. Since CLDRFile no longer implements Iterable, it's necessary to call a method on it to get an iterable.

tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSTFactory.java
```diff
@@ -73,7 +73,7 @@ public void TestBasicFactory() throws SQLException {
         STFactory fac = getFactory();
         CLDRFile mt = fac.make(locale, false);
         BallotBox<User> box = fac.ballotBoxForLocale(locale);
-        mt.iterator();
+        mt.iteratorWithoutExtras();
```
btangmu, line 76: This is "iterator" not "iterable"; the method has merely been renamed.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRFilePseudolocalizer.java
```diff
@@ -403,7 +403,7 @@ public CLDRFile generate() {
         // Create input CLDRFile object resolving inherited data.
         CLDRFile input = factory.make(ORIGINAL_LOCALE, false);
         XMLSource outputSource = new SimpleXMLSource(outputLocale);
-        for (String xpath : input) {
+        for (String xpath : input.iterableWithoutExtras()) {
```
btangmu, line 406: Here, there would be a test failure if it were `iterableDefault`; likewise everywhere that `iterableWithoutExtras` is called outside of CLDRFile

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -1486,6 +1485,20 @@ public Iterator<String> iterator(String prefix, Comparator<String> comparator) {
         return orderedSet.iterator();
     }
 
+    private final boolean DEFAULT_ITERABLE_INCLUDES_EXTRAS = true;
```
btangmu, line 1488: Here's the key private boolean

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -3081,7 +3094,7 @@ public String getWinningPath(String path) {
     public Collection<String> getExtraPaths() {
         Set<String> toAddTo = new HashSet<>();
         toAddTo.addAll(getRawExtraPaths());
-        for (String path : this) {
+        for (String path : this.iterableWithoutExtras()) {
```
btangmu, line 3097: java.lang.StackOverflowError would occur here due to infinite loop if extras were included

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCldrFactory.java
```diff
@@ -92,7 +92,7 @@ enum Status {
 
     private Status checkAnnotations(CLDRFile cldrFile) {
         Status status = Status.none;
-        for (String xpath : cldrFile) {
+        for (String xpath : cldrFile.iterableWithoutExtras()) {
```
btangmu, line 95: test would fail if with extras

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCldrFactory.java
```diff
@@ -125,7 +125,7 @@ private Status checkAnnotations(CLDRFile cldrFile) {
      */
     private String getUncontainedPath(CLDRFile subset, CLDRFile superset) {
         int debugCount = 0;
-        for (String xpath : subset) {
+        for (String xpath : subset.iterableWithoutExtras()) {
```
btangmu, line 128: test would fail if with extras

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTransforms.java
```diff
@@ -211,7 +211,7 @@ public void TestUzbek() {
 
             Set<String> latinFromCyrillicSucceeds = new TreeSet<>();
             Set<String> latinFromCyrillicFails = new TreeSet<>();
-            for (String path : uzCyrl) {
+            for (String path : uzCyrl.iterableDefault()) {
```
btangmu, line 214: test would fail if with extras

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTransforms.java
```diff
@@ -885,7 +885,7 @@ public void TestLocales() {
             }
             badPlusSample.clear();
             CLDRFile file = factory.make(locale, false);
-            for (String path : file) {
+            for (String path : file.iterableWithoutExtras()) {
```
btangmu, line 888: test would fail if with extras

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUnits.java
```diff
@@ -3018,7 +3018,7 @@ public void TestGenderOfCompounds() {
                             new TreeMap<String, Object>(),
                             Boolean.class);
 
-            for (String path : cldrFile) {
+            for (String path : cldrFile.iterableWithoutExtras()) {
```
btangmu, line 3021: test would fail if with extras

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -3165,7 +3178,8 @@ private List<String> getRawExtraPathsPrivate() {
             Set<Count> pluralCountsRaw = plurals.getCounts();
             if (pluralCountsRaw.size() != 1) {
                 // we get all the root paths with count
-                addPluralCounts(toAddTo, pluralCounts, pluralCountsRaw, this);
+                addPluralCounts(
+                        toAddTo, pluralCounts, pluralCountsRaw, this.iterableWithoutExtras());
```
btangmu, line 3182: StackOverflowError again if with extras

PR #4286, opened 2025-01-16
CLDR-16530 Fix links in survey-tool-accounts.md

btangmu:
-It was unclear where [you must list all of the locale(s) that you would like to provide data for, using the locale codes on...] should point; picking-the-right-language-code was the best I could find

CLDR-16530

ALLOW_MANY_COMMITS=true


btangmu:
I noticed something. If a user new to Survey Tool wants to know the language identifier for, say, French, Survey Tool doesn't make it very easy to find it. I made a new ticket for that: https://unicode-org.atlassian.net/browse/CLDR-18230

conradarcturus (approved):

PR #4287, opened 2025-01-17
CLDR-15830 Refactor/simplify NameType

btangmu:
-Remove getNameName, getNameTypeName, String[][] NameTable, INDEX_LANGUAGE, ..., nameTableIndex, fromNameTableIndex

-Associate the enum values with the corresponding path fragments by means of constructors

-Move FIX_KEY_NAME, fixKeyName from CLDRFile to NameType

-New method NameType.toCodeType converts NameType to CodeType and obviates conversion to/from String as intermediate step

-Comments

CLDR-15830

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
@@ -84,26 +84,6 @@ public class StandardCodes {
                     return NameType.NONE;
             }
         }
-
-        public static CodeType fromNameType(NameType nameType) {
-            switch (nameType) {
-                case LANGUAGE:
-                    return language;
-                case SCRIPT:
-                    return script;
-                case TERRITORY:
-                    return territory;
-                case VARIANT:
-                    return variant;
-                case CURRENCY:
-                case CURRENCY_SYMBOL:
-                    return currency;
-                case TZ_EXEMPLAR:
-                    return tzid;
-                default:
-                    throw new IllegalArgumentException("Unsupported name type");
-            }
-        }
     }
 
     private static final Set<CodeType> TypeSet =
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
@@ -84,6 +84,26 @@ public class StandardCodes {
                     return NameType.NONE;
             }
         }
+
+        public static CodeType fromNameType(NameType nameType) {
+            switch (nameType) {
+                case LANGUAGE:
+                    return language;
+                case SCRIPT:
+                    return script;
+                case TERRITORY:
+                    return territory;
+                case VARIANT:
+                    return variant;
+                case CURRENCY:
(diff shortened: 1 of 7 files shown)


srl295 (dismissed):
great refactor

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
```diff
@@ -84,6 +84,26 @@ public NameType toNameType() {
                     return NameType.NONE;
             }
         }
+
```
macchiati, line 87: Better to make the related enum values be elements of the other one.

That is, CodeType would have a field set from its private constructor:

```
final NameType nameType;
```

Then toNameType becomes:

```
public NameType toNameType() {
  return nameType;
}
```
Similarly, NameType would have a field set from its private constructor:

```
final CodeType codeType;
```
Then fromNameType moves to be a method NameType.toCodeType.

```
public CodeType toNameType() {
  if (codeType == null) {
    throw new IllegalArgumentException("Unsupported name type");
  }
  return codeType;
}
```

macchiati (changes requested):
Just one item, otherwise looks great.

tools/cldr-code/src/main/java/org/unicode/cldr/util/StandardCodes.java
btangmu: good idea, I'll make that change

btangmu: @macchiati I thought it was a great idea, but it failed due to the sequence of initialization. If NameType constructor references CodeType, and CodeType constructor references NameType, one or the other will get null values.

See https://stackoverflow.com/questions/1506410/java-enums-two-enum-types-each-containing-references-to-each-other which suggestions various workarounds, but they're all kind of ugly.

Instead I just made it a little more symmetrical by implementing NameType.toCodeType instead of CodeType.fromNameType.

Maybe there's still a more elegant or efficient way; I suggest leaving that for a follow-up PR if and when it seems worthwhile to pursue.

macchiati:
I've done it with something like the following, to avoid switches. However, I think your solution is good enough.

```	enum Foo {a, b, c}
	
	enum Bar {D(Foo.a), E(Foo.b), F(Foo.c);
		
		private Foo foo;
		
		private Bar(Foo foo) {
			this.foo = foo;
		}
		
		public Foo toFoo() {
			return foo;
		}
		
		private static Map<Foo, Bar> BACKMAP = Arrays.stream(Bar.values())
                          .collect(Collectors.toUnmodifiableMap(bar -> bar.foo, bar -> bar));
		
		public static Bar fromFoo(Foo foo) {
			return BACKMAP.get(foo);
		}
	}

	private static void checkFooBar() {
		Arrays.stream(Foo.values()).forEach(foo -> {
			Bar bar = Bar.fromFoo(foo);
			Foo foo2 = bar.toFoo();
			System.out.println(foo + "\t" + bar + "\t" + foo2);
			});
	}
```

macchiati (approved):

PR #4288, opened 2025-01-17
CLDR-18105 Remove parent locale for Haitian Creole

conradarcturus:
Haitian Creole `ht` is erroneously being reported as "modern coverage" because its parent locale `fr_HT` is technically fully populated (since that, in turn, inherits from `fr`)

See the incorrect characterization in the v47 charts https://www.unicode.org/cldr/charts/47/supplemental/locale_coverage.html#ht

The quickest way to register the language as not submitted is to just remove the parent locale.

CLDR-18105

ALLOW_MANY_COMMITS=true


macchiati:
We need to discuss this, because there are conflicting goals.

macchiati:
Just for the paper trail; we sometimes use cross-language inheritance where the users of the inheriting language are used to seeing intermixed data from the inherited language, and it would be better for them to see that than than nothing (or English). However, that can mask the true coverage of the inheriting language both in the ST, and in the charts and coverage text file (causing implementations to misread how fleshed out it is). So we need some way to fix that. Cf @AEApple

conradarcturus:
Oh thanks -- yes you are completely right. It was better to fix the way inheritance computed coverage level and not extra difficult.

PR #4290, opened 2025-01-17
CLDR-17014 No code fallbacks for language/script paths

btangmu:
-Use the extra-paths mechanism instead of code-fallback; new ExtraPaths.java

-For each type (language/script), create the set of extra paths only once

-Change es-419 to es (419) in localeDisplayName.txt so TestLocaleDisplay passes

-Revise testExtraPaths, testGetPaths, so they pass

-Comments

CLDR-17014

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
@@ -1,0 +1,100 @@
+package org.unicode.cldr.util;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+
+public class ExtraPaths {
+
+    private static final Map<NameType, ExtraPaths> instances = new HashMap<>();
+
+    public static ExtraPaths getInstance(NameType nameType) {
+        return instances.computeIfAbsent(nameType, ExtraPaths::new);
+    }
+
+    private final NameType nameType;
+    private final Collection<String> paths;
+
+    private ExtraPaths(NameType nameType) {
+        this.nameType = nameType;
+        paths = new HashSet<>();
+    }
+
+    void append(Collection<String> toAddTo) {
+        if (paths.isEmpty()) {
+            populatePaths();
+        }
+        toAddTo.addAll(paths);
+    }
+
+    private void populatePaths() {
+        // TODO: https://unicode-org.atlassian.net/browse/CLDR-17014
+        // StandardCodes.CodeType codeType = StandardCodes.CodeType.fromNameType(nameType);
+        // See https://github.com/unicode-org/cldr/pull/4287
+        StandardCodes.CodeType codeType;
+        switch (nameType) {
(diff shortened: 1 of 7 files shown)


btangmu:
This PR is very similar to https://github.com/unicode-org/cldr/pull/4254, butin addition to language paths it also applies to script paths, and the new class ExtraPaths is refactored to facilitate adding more types in addition to language/script

btangmu:
@macchiati please let me know when it's OK to deploy this to cldr-staging for testing

screenshot running locally:

http://localhost:8888/cldr-apps/v#/aa/Scripts/44d4d43894cfbd45

image

compare production https://st.unicode.org/cldr-apps/v#/root/Scripts/44d4d43894cfbd45

image

macchiati:
Ok to deploy now (or when you are back at work); and everyone should try it
out, since there will be differences in UI. Best to try with a locale at
(say) Basic but set the Coverage up to Moderate to see how it behaves.

We should also make sure to include a notice about this change in the
release notes. (Maybe we should have a label releaseNote for that in
tickets? @***@***.***)

btangmu:
Per today's infra meeting, this is now ready for review

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
```diff
@@ -0,0 +1,100 @@
...
+                addAltPath("yue", "menu");
+                addAltPath("zh", "menu");
+                addAltPath("zh_Hans", "long");
+                addAltPath("zh_Hant", "long");
+                break;
+            case SCRIPT:
+                addAltPath("Hans", "stand-alone");
+                addAltPath("Hant", "stand-alone");
+        }
+    }
```
srl295, line 88: why are these two tables in code instead of in data?

srl295 (approved):
LGTM.   Need to get those lists out of code though (which isn't this PR)

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
macchiati: Could be, but best to be in a different ticket.

btangmu: Moving to a data file is an interesting idea. This ticket moves the code from XMLSource.java to ExtraPaths.java, and so far only for language/script paths. There's still a lot of code for other kinds of extra paths in CLDRFile.java, and for other kinds of "constructed" paths (like currencies) in XMLSource.java. A plan to move such code to a data file should probably take into account not only the code here but also what remains in CLDRFile and XMLSource. Not all of those paths in CLDRFile currently have corresponding NameType values; dayPeriods paths for example don't. All those in XMLSource do have corresponding NameType values. Even if we keep the fallback values for currencies, etc., we might consider taking them out of XMLSource.java and treating them more like extra paths. There could be more of a comprehensive framework rather than scattered code/data.

srl295: it seems like perhaps it's a coverage level issue or something. Not sure.

PR #4292, opened 2025-01-17
CLDR-17760 syr, delete obsolete decimalFormats/currencyFormats without numberSystem

pedberg-icu:
CLDR-17760

In `syr`, delete the obsolete `<decimalFormats>` and ` <currencyFormats>` elements that do **_not_** have a `numberSystem` attribute at all (these probably date from when the `syr` data was first added in xml to the seed directory years ago). These do not show up in Survey Tool and (I think) are ignored by ICU code. The current version of these elements does have a `numberSystem` attribute, and `syr` does also have `<decimalFormats>` and ` <currencyFormats>` elements (and also `<percentFormats>`) **_with_** `numberSystem="latn"`; these are the ones that appear in Survey Tool and get used by the code.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/syr.xml
+++ b/common/main/syr.xml
@@ -4402,13 +4402,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<symbols numberSystem="vaii">
 			<timeSeparator draft="unconfirmed">↑↑↑</timeSeparator>
 		</symbols>
-		<decimalFormats>
-			<decimalFormatLength>
-				<decimalFormat>
-					<pattern draft="unconfirmed">#,##0.###;#,##0.###-</pattern>
-				</decimalFormat>
-			</decimalFormatLength>
-		</decimalFormats>
 		<decimalFormats numberSystem="latn">
 			<decimalFormatLength>
 				<decimalFormat>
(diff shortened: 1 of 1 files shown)


AEApple:
How does deprecation of structure work in other languages? I'm assuming that at some point in the past there was only one script possible per locale and there was no option to have a number format per language.

Either way if the data is unconfirmed that in theory no one is likely to be using it, but I'd like to understand the larger context because I've seen unconfirmed data elsewhere in locales that aren't vetted by any organization and are unlikely to be vetted by guest vetters either.

conradarcturus (approved):

pedberg-icu:
@AEApple We are not deprecating any structure here. We are only removing unconfirmed data is that is in an obsolete form and does not appear in survey tool (since there co-exists with it confirmed data that is in the current form and does appear in survey tool). I assume at some point in the distant past it was that case the number elements did not specify a number system because "latn" was assumed, but I think for at least the past 15 years number elements have specified the number system that they applied to (so the entries being removed are from before that time).

macchiati:
I think we might want to generalize this action, and remove all of the old 'scriptless' data.

PR #4293, opened 2025-01-18
CLDR-18193 add Piemontese

conradarcturus:
CLDR-18193

Putting this up to merge into main -- but won't actually merge until the v47 cutoff, targetting v48. There are a lot of data updates on the main branch not in the v48 branch that I'd like to keep.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/pms.xml
+++ b/common/main/pms.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2025 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-3.0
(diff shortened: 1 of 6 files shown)


srl295:
We can merge from main to 48 too.

srl295:
nothing had been merged to ddl/v48 yet but ddl/v48 is now up to date with main. So i moved the branch target to v48

common/testData/localeIdentifiers/localeDisplayName.txt
srl295, line 1: please don't regenerate this file in such a PR, it's really not needed.  It's a late stage BRS task.

srl295 (dismissed):
OK except for the `testData` file being changed, just remove that

common/testData/localeIdentifiers/localeDisplayName.txt
conradarcturus: Yea something doesn't look right there. I'll change that file back.

By practice I try to regenerate everything that would change in each PR -- I think it's good practice to continually regenerate files, otherwise consequences of a well-intentioned change may not be realized.

srl295 (approved):

PR #4294, opened 2025-01-19
CLDR-18233 Remove ISO 639-5 language families from Likely Subtag

conradarcturus:
CLDR-18233

ALLOW_MANY_COMMITS=true


conradarcturus:
Resolved in a different PR https://github.com/unicode-org/cldr/pull/4598

PR #4295, opened 2025-01-19
CLDR-18080 add Hant-Latn transform, add Hans-Latn alias for Hani-Latn

pedberg-icu:
CLDR-18080

This ticket adds a `Hant-Latn` transform which uses 101 readings that are different for Hant/TW, then calls the main `Hani-Latn` transform for everything else. It also adds `Hans-Latn` as an alias for the current `Hani-Latn` transform, and adds test data for both. Also, as noted in the ticket: 
- The current Han-Latin transform has the following context rule:
`沈 } \u0020? 阳 ->shěn;# 沈 is shěn (not chén) if followed by 阳 yáng: 沈阳 city Shěnyáng`; this dates from before Unicode 14, at which point the kMandarin value for U+6C88 沈 was changed from “chén” to “shěn chén”. With that change the rule became obsolete for `Hans`.  However it is still slightly relevant for `Hant`; if the _simplified_ form 沈阳 of the name for the city Shenyang is encountered in a Hant/Taiwan context, the first character should be transliterated as shěn rather than as chén. So this rule should be moved to the `Hant-Latn` transform. 

ALLOW_MANY_COMMITS=true


common/transforms/Hant-Latin.xml
```diff
@@ -0,0 +1,116 @@
...
+<supplementalData>
+	<version number='$Revision$'/>
+	<transforms>
+		<transform source="Hant" target="Latn" direction="forward" alias="und-Latn-t-und-hant">
+			<tRule>
+# Warning: does not do round-trip mapping!!
+# Convert compounds; these are added individually, not derived from Unihan kMandarin.
+# Here Han-Spacedhan() has not yet been applied.
+# The following was moved from Hans-Latn; in a Hant/Taiwan context, the simplified-form city name 沈阳 should still transform to shěnyáng.
+沈 } 阳 →shěn;# 沈 is shěn (not chén) if followed by 阳 yáng: 沈阳 city Shěnyáng
```
robertbastian, line 17: `Hans-Latn` also matched `沈 阳`, but this only matches `沈阳` without a space; is this intended?

common/transforms/Hant-Latin.xml
```diff
@@ -0,0 +1,116 @@
...
+[崾]→yǎo;    # U+5D3E
+[畬]→yú;   # U+756C
+[薁]→yù;   # U+8581
+[嶦]→zhān;   # U+5DA6
+[著]→zhe;   # U+8457
+[徵]→zhēng;   # U+5FB5
+[苧]→zhù;   # U+82E7
+# END From Unicode 17, the above should be autogenerated:
+# Then run the normal Hani-Latn transform for the rest
+::Hani-Latn();
```
srl295, line 112: nice solution!

srl295 (approved):
Looks good besides @robertbastian comment about spaces

common/transforms/Hant-Latin.xml
pedberg-icu: I think the behavior for Hant is more correct, we should fix Hans-Latn to do the Han-SpacedHan after handling the compounds (under a different PR).

PR #4296, opened 2025-01-20
CLDR-18236 Add `gmtUnknownFormat`

robertbastian:
CLDR-18236

ALLOW_MANY_COMMITS=true


docs/ldml/tr35-dates.md
```diff
@@ -1707,12 +1708,18 @@ Note: A generic location format is constructed by a part of time zone ID represe
...
 * "Гринуич+03:30" (long)
 
-Otherwise (when the offset from GMT is zero, referring to GMT itself) the style specified by the `<gmtZeroFormat>` element is used:
+Otherwise, when the offset from GMT is zero, the style specified by the `<gmtZeroFormat>` element is used:
 
 * "GMT"
 * "UTC"
 * "Гринуич"
 
+Otherwise, when the offset from GMT is not known, the style specified by the `<gmtUnknownFormat>` element is used:
```
macchiati, line 1717: This sequence doesn't terminate; that is, there should be an unconditional final state

```suggestion
Otherwise (that is, when the offset from GMT is not known), the style specified by the `<gmtUnknownFormat>` element is used:
```

macchiati (changes requested):

macchiati (dismissed):

macchiati (approved):

PR #4298, opened 2025-01-21
CLDR-17879 fix coverage/PathHeader for alphaNextToNumber,noCurrency; add to root & some locales

pedberg-icu:
CLDR-17879

See the ticket CLDR-17879. This PR fixes some CLDR issues with the currencyFormat/patterns alts alphaNextToNumber, noCurrency as follows:
- Fix coverage so for 'latn' these variants are at moderate (not comprehensive) for the standard currency formats (alphaNextToNumber was already at modern for the compact currency formats); for any other number system that is default/traditional in a given locale the variants were already at moderate for the standard currency formats, but fix so alphaNextToNumber is at modern (not comprehensive) for the compact currency formats.
- Add coverage tests to check
- Fix PathHeader so the compact currency formats for non-late' number systems that are default/traditional in a given locale will show in Survey Tool
- In root, add an alphaNextToNumber variant for the standard currency patterns (root already had a noCurrency alt for these). However, for the compact currency patterns, do not add an alt alphaNextToNumber; since the alt alphaNextToNumber patterns seem to inherit upward, this causes a test error due to a conflict between the number of zeros in the root compact currency pattern and the number of zeros in some locale compact currency patterns, such as for `de`.
- For the 42 top locales (defined as ones that all 4 major CLDR contributors have listed as modern), ensure that for 'latn' and any other number systems that are default/traditional in a locale, we have alts alphaNextToNumber and noCurrency for the standard number formats and compact currency formats that are present in the locale. These are added as draft="provisional"; the  alphaNextToNumber values are just `↑↑↑`, and the noCurrency values for standard currency formats are the non-alt value with currency symbol (and associated spacing) stripped. 4 of the locales needed no changes, but the others needed some additions/corrections (corrections when they specified a noCurrency that included a currency symbol).
    - These first 42 are just to check that the approach to these changes is valid, once that seems to be correct I can update other locales in a separate PR.

ALLOW_MANY_COMMITS=true


pedberg-icu:
Filed https://unicode-org.atlassian.net/browse/CLDR-18238 to add a test that the noCurrency variant does not in fact have a currency symbol.

srl295 (approved):

conradarcturus:
If I understand what happened correctly -- this (or a different PR in the past few days) is causing a lot of our locales to drop in coverage. See https://github.com/unicode-org/cldr/commit/8dfa3e2b87829007bd0cbe9fddedf7048c3bd780

pedberg-icu:
> If I understand what happened correctly -- this (or a different PR in the past few days) is causing a lot of our localees to drop in coverage. See coverage changes in [8dfa3e2](https://github.com/unicode-org/cldr/commit/8dfa3e2b87829007bd0cbe9fddedf7048c3bd780)

@conradarcturus @macchiati @srl295 @AEApple Yes, this PR likely caused many locales to drop in coverage. Not sure what to do about that. We need the changes for this in place for 48 (and ICU may very likely want this change). Can we tweak coverage to exclude these items?

PR #4300, opened 2025-01-21
CLDR-18244 Update references to DDL WG

AEApple:
CLDR-18244

Update references to DDL WG

ALLOW_MANY_COMMITS=true


AEApple:
The main question I have is if we should move the ddl page under the cldr-tc, similar to the other WGs and add a redirect. I'm fine either way, but I wanted to ask while we're updating this page.

srl295 (approved):

srl295:
> The main question I have is if we should move the ddl page under the cldr-tc, similar to the other WGs and add a redirect. I'm fine either way, but I wanted to ask while we're updating this page.

it's already under the cldr-tc in the site map, this just gives it a shorter url

srl295:
@AEApple OK to merge this?

macchiati (approved):

PR #4301, opened 2025-01-22
CLDR-16821 Fix Australian time zone names

robertbastian:
CLDR-16821

This adds exemplar cities for some islands, and aligns the `en` metazone names with the correct names from `en-AU`.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -3913,12 +3913,17 @@ annotations.
 			<regionFormat type="daylight">{0} Daylight Time</regionFormat>
 			<regionFormat type="standard">{0} Standard Time</regionFormat>
 			<fallbackFormat>{1} ({0})</fallbackFormat>
-			<zone type="Pacific/Honolulu">
-				<short>
-					<generic>HST</generic>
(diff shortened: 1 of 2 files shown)


AEApple:
Why did the exemplar cities in root get re-ordered? It makes it more difficult to read the diff, so I'm curious about the motivation.

robertbastian:
Because as the list is growing it's becoming hard to locate a zone if it's not alphabetic. It's in a separate commit so it's easier to review.

macchiati (changes requested):
The locale files are all processed for release. Any changes in the ordering will be discarded, so it just causes problems in the diff for reviewers. If you want to change the wording, then you should create a new ticket to change the ordering that the tooling produces.

robertbastian:
I've remoed the root.xml change, but I'm keeping en.xml sorted.

macchiati (approved):
Please don't sort in the future. It makes it painful to review_, and it is reversed by the tooling!_ 
If you think they should be sorted differently, please file a different ticket.

PR #4302, opened 2025-01-22
CLDR-18099 Add a few more European Englishes

AEApple:
CLDR-18099

- en_CZ
- en_HU
- en_PT

and update en_FR for consistency with the approach for other en_150 child locales.

- [x] This PR completes the ticket. I will clone off any future work into a separate ticket targeting 48.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_FR.xml
+++ b/common/main/en_FR.xml
@@ -11,6 +11,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<language type="en"/>
 		<territory type="FR"/>
 	</identity>
+	<numbers>
 		<symbols numberSystem="latn">
 			<decimal>,</decimal>
 			<group> </group>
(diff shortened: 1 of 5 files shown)


common/main/en_FR.xml
```diff
@@ -11,4 +11,10 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<language type="en"/>
 		<territory type="FR"/>
 	</identity>
+	<numbers>
+		<symbols numberSystem="latn">
+			<decimal>,</decimal>
+			<group> </group>
```
srl295, line 17: error here. is this supposed to be a different kind of space? fr.xml has NNBSP

srl295: It gets collapsed to `<group></group>` or `<group/>` without `xml:space="preserve"` so the parser sees a 0-length group string.  Probably better to fix it here.

srl295 (approved):

PR #4305, opened 2025-01-22
CLDR-18253 Add en_JP, en_KR, en_TW

AEApple:
CLDR-18253

Will add other locales as soon as en_JP passes the test.

ALLOW_MANY_COMMITS=true


AEApple:
@srl295 @btangmu @macchiati - Is there an easier way to identify where I'm missing sideways inheritance that I need to add back? I had only kept the items where something was different, but I must have removed something that I should have made inherit instead, I'm just not sure how to find it well without re-adding back things in the sections with only a couple of items.

macchiati:
Sorry, was occupied. Will check tomorrow

common/main/en_JP.xml
```diff
@@ -0,0 +1,305 @@
...
+			<metazone type="Malaysia">
+				<short>
+					<standard>MYT</standard>
+				</short>
+			</metazone>
+			<metazone type="Singapore">
+				<short>
+					<standard>SGT</standard>
+				</short>
+			</metazone>
```
robertbastian, line 295: This is definitely missing Japan's own time zone, JST.

macchiati: That sounds reasonable; perhaps similar actions for other locales. 

Annemarie, I'm happy to approve now, if we want to get it in to the first ICU integration.

PR #4306, opened 2025-01-22
CLDR-17760 4 locales, delete obsolete decimal/currencyFormats without numberSystem or add it

pedberg-icu:
CLDR-17760

Per discussion in ticket: For dv, gez, sa, tt (the remaining locales for which this is an issue): If there is a number format element without numberSystem and there is a corresponding element with numberSystem='latn', just delete the one without; otherwise add numberSystem='latn' to the one without.

This does not address root or additional tests, which will be under separate tickets to be filed.

ALLOW_MANY_COMMITS=true


srl295 (approved):

DavidLRowe (approved):
LGTM

PR #4309, opened 2025-01-22
CLDR-18255 BRS v47a0 CLDRModify no options: reorder a couple of timezone entries

pedberg-icu:
CLDR-18255

CLDRModify passes no-options and -fP (DAIP) before v47 alpha0. The only change was that the no-options pass reordered 2 timezone entries, one each in root and en.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -3928,6 +3928,9 @@ annotations.
 			<zone type="Etc/Unknown">
 				<exemplarCity>Unknown City</exemplarCity>
 			</zone>
+			<zone type="Pacific/Easter">
+				<exemplarCity>Easter Island</exemplarCity>
+			</zone>
 			<zone type="Europe/London">
 				<long>
 					<daylight>British Summer Time</daylight>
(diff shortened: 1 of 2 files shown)


robertbastian (changes requested):
This collides with https://github.com/unicode-org/cldr/pull/4301, for which I'm still waiting on a review

pedberg-icu:
> This collides with #4301, for which I'm still waiting on a review

Does #4301 address the reordering issues here, or make them moot?

robertbastian:
It reorders `en.xml` but not `root.xml`

AEApple (approved):

AEApple:
> It reorders `en.xml` but not `root.xml`

I think that may be WAI for the regular CLDRModify passes? I'm curious what would trigger re-ordering of root or if it would be a one-time/ad hoc thing that needs to be done whenever new metazones are added?

pedberg-icu:
Closed in favor of [#4310,](https://github.com/unicode-org/cldr/pull/4310) which was done after #4301 was merged (source of the conflict).

PR #4312, opened 2025-01-23
CLDR-17760 Add note about v48 deprecation of number symbols and format elements without numberSystem

pedberg-icu:
CLDR-17760

Add migration note (per ticket comment) about v48 deprecation of number symbols and format elements without numberSystem, and removal of these from root.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-47.md
+++ b/docs/site/downloads/cldr-47.md
@@ -102,7 +102,11 @@ For a full listing, see [Delta Data](https://unicode.org/cldr/charts/dev/delta/i
 
 ## Migration
 
-- TBD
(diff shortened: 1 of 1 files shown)


macchiati (dismissed):

AEApple (commented):
Any chance we can point to the relevent section in the spec for easy reference to people reviewing the release notes?

pedberg-icu:
> Any chance we can point to the relevent section in the spec for easy reference to people reviewing the release notes?

@AEApple Good idea, just updated to do that. But now I think we also need to update the spec a bit because its examples use the to-be-deprecated style without numberSyste attribute. Will do that in a separate PR (on my fork).

AEApple (approved):

PR #4313, opened 2025-01-23
CLDR-18128 Generate datetime test data using semantic skeletons

echeran:
CLDR-18128

This PR switches the implementation of date time test data generation to:
  -  use semantic skeletons to represent the input values
  -  use a more carefully designed set (called "kernel") of combinations of input values that is minimal size while maintaining a near maximal coverage of date time functionality use cases / options

A follow up PR is still needed to use this kernel set of combinations and combine it via iteration over all values in the dimensions of: locales, a larger set of input date/times, a larger set of time zones, etc. Combining this kernel set as a Cartesian product with iterations of larger value sets across different dimensions will ensure representation of those values in the different dimensions.

ALLOW_MANY_COMMITS=true


sffc (approved):
Nice work!

What I think we have left before we consider the ticket "complete":

1. Optimize the kernel a bit more to further reduce duplication in test case generation
    - Including support for TimePrecision
2. Take the cartesian product of the kernel over all of the dimensions
3. Fix TODOs and any additional bugs we find

What we currently have is a great milestone, sufficient for ICU4X testing and CLDR 47.

macchiati (approved):

sffc:
@echeran Ready to merge?

PR #4314, opened 2025-01-24
CLDR-17760 Also fix number symbols without numberSystem; update spec accordingly

pedberg-icu:
CLDR-17760

In addition to number format elements without a numberSystem attribute (PRs [#4292](https://github.com/unicode-org/cldr/pull/4292), [#4306](https://github.com/unicode-org/cldr/pull/4306)), we also need to clean up number symbols without a numberSystem attribute. And we need to update the spec for both so the examples do include a numberSystem attribute, and mention the such elements without a numberSystem attribute should not be used and will be deprecated.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -220,13 +220,8 @@ The LDML specification is divided into the following parts:
 * [References](#References)
 * [Acknowledgments](#Acknowledgments)
 * [Modifications](#Modifications)
-  * [Conformance Modifications](#conformance-modifications)
-  * [Locale Identifiers and Inheritance Modifications](#locale-identifiers-and-inheritance-modifications)
-  * [Message Format Modifications](#message-format-modifications)
-  * [Date Modifications](#date-modifications)
-  * [Units Modifications](#units-modifications)
-  * [Collation Data Changes](#collation-data-changes)
-  * [Misc. Modifications](#misc.-modifications)
+  * [Number symbols and formats without numberSystem](#number-symbols-and-formats-without-numbersystem)
+  * [Well-formed identifiers](#well-formed-identifiers)
 
 ## <a name="Introduction" href="#Introduction">Introduction</a>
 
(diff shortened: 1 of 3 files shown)


docs/ldml/tr35.md
```diff
@@ -4358,7 +4358,7 @@ Other contributors to CLDR are listed on the [CLDR Project Page](https://www.uni
     - Change the constraints into either well-formedness constraints or validity constraints.
     - Add validity constraints for base-component.
     - Reformat the EBNF to avoid using HTML tables.
-- Updated the [Unit_Preferences](tr35-info.html#Unit_Preferences) to provide well-formedness and validity definitions.
+- Updated the [Unit_Preferences](tr35-info.md#Unit_Preferences) to provide well-formedness and validity definitions.
```
AEApple, line 4361: I thought we were intentionally keeping the html versions of the links in the spec, or is this an issue before we publish the final spec?

AEApple (approved):
LGTM, I just thought we were supposed to keep the html versions of the links?

docs/ldml/tr35.md
pedberg-icu: As I understand it, the .md links get converted to .html when the spec is processed for posting.

AEApple: If that's the case, then shouldn't we bulk convert them back at the beginning of each cycle and only convert them to html when we are processing the final spec on the maint branch for that release?

pedberg-icu: These internal links using .md do not get converted in the GitHub repo, only when being processed for posting. If you look at something like tr35.md, the internal references to other parts of the spec all seem to use .md, not .html. Check out the directory to various parts:
```
*   Part 1: [Core](tr35.md#Contents) (languages, locales, basic structure)
*   Part 2: [General](tr35-general.md#Contents) (display names & transforms, etc.)
*   Part 3: [Numbers](tr35-numbers.md#Contents) (number & currency formatting)
*   Part 4: [Dates](tr35-dates.md#Contents) (date, time, time zone formatting)
*   Part 5: [Collation](tr35-collation.md#Contents) (sorting, searching, grouping)
*   Part 6: [Supplemental](tr35-info.md#Contents) (supplemental data)
*   Part 7: [Keyboards](tr35-keyboards.md#Contents) (keyboard mappings)
*   Part 8: [Person Names](tr35-personNames.md#Contents) (person names)
*   Part 9: [MessageFormat](tr35-messageFormat.md#Contents) (message format)
```

AEApple: Ah, should we have a test so that people don't use the html by accident?

pedberg-icu: Perhaps, but Steven can probably provide the best advice on that.

AEApple: SG! I'll add it to the CLDR Infra agenda

macchiati:
I think currently the spec and the site behave differently. One of them has to have .md links and the other one has to have .html links.

The. MD links are nicer because they work in GitHub rich text view.

PR #4315, opened 2025-01-27
CLDR-18258 Exclude new values from coverage

conradarcturus:
These values were added to CLDR in between submission cycles and it would be unfair to drop locale coverage if the languages did not support them. I'm not happy with the styling of all of my code here -- I'm happy to respond to suggestions.

Thereby, we're temporarily overriding their coverage as "comprehensive". All other coverage level changes are for other reasons ~~(and Haitian Creole has a separate ticket to fix it)~~. You can see changes to charts in this PR https://github.com/unicode-org/cldr-staging/pull/168 Some changes to charts are from other PRs, but the ones keeping coverage level at a good level are there.

While I was here I also had put in the fix for the ticket on Haitian Creole (https://unicode-org.atlassian.net/browse/CLDR-18105) -- because tests are failing since it thought Haitian Creole is a modern locale and thereby it should have translations in other languages (which we don't have yet because its new). So commit #3 does that change.

CLDR-18258

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/properties/coverageLevels.txt
+++ b/common/properties/coverageLevels.txt
@@ -62,6 +62,7 @@ hi ;	modern ;	Hindi
 hi_Latn ;	modern ;	Hindi [Latin]
 hr ;	modern ;	Croatian
 hsb ;	modern ;	Upper Sorbian
+ht ;	modern ;	Haitian Creole
 hu ;	modern ;	Hungarian
 hy ;	modern ;	Armenian
 ia ;	moderate ;	Interlingua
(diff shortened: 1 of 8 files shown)


macchiati (changes requested):
Looks pretty good, however to appears to go overboard with the script-specific currency data. We should only be targeting one currency, after all, and that should be done with the variables.

Also, tests fail.

Side note: 'fair' isn't the right term.

conradarcturus:
Thanks for taking a look. It takes 10 minutes to run all of the code so its been a long game of change and hope the result passes every test. Apologies there still seems to be errors. I'll do another pass to try to resolve them. Let me know if you have tips to make running tests faster.

The script specific currency changes were solved what I thought you were asking for, In the original email you indicated we needed to change these to comprehensive:

> All the ZWG currency elements
> Currency alphaNextToNumber, noCurrency
> Short Currency using XXX Digits
> light-speed

The script-specific changes are for Currency alphaNextToNumber, noCurrency, and Short currency -- since that problem was fixed in Latn but introduced for non-latin scripts.

macchiati (approved):
I was worried that you were including too much, But I reviewed it in detail (once I had more than a phone!), and it looks good. Since you are off in different timezones, I'm going to go ahead and merge.

PR #4319, opened 2025-01-28
CLDR-18268 Add timeData for GS

AEApple:
CLDR-18268

Add timeData for GS in supplementalData

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

pedberg-icu:
Nice solution!

PR #4320, opened 2025-01-29
CLDR-18268 Add HH timeFormat to en_GS

AEApple:
CLDR-18268

Add HH timeFormat to en_GS

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_GS.xml
+++ b/common/main/en_GS.xml
@@ -11,4 +11,36 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<language type="en"/>
 		<territory type="GS"/>
 	</identity>
+	<dates>
+		<calendars>
+			<calendar type="gregorian">
+				<timeFormats>
+					<timeFormatLength type="full">
+						<timeFormat>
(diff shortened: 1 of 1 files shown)


AEApple:
Note: I don't think we'll probably need this, but I created it in case we do, and will abandon if we don't. It was hard to tell what en_GB was doing since these formats were marked as inheriting so I copied the formats from en_IE just in case.

pedberg-icu (approved):
en_GB has explicit gregorian `timeformats` using H, and what you have for en_GS matches the en_GB formats; en_GB is not inheriting from en_001, which inherits h-cycle timeFormats from en. So what you have here _is_ necessary.

AEApple:
> en_GB has explicit gregorian `timeformats` using H, and what you have for en_GS matches the en_GB formats; en_GB is not inheriting from en_001, which inherits h-cycle timeFormats from en. So what you have here _is_ necessary.

Thanks! I must have mis-read the en_GB file when I originally checked it. I'll merge this in. Longer term I want to check what the most common hour cycle is and update it.

PR #4322, opened 2025-01-30
CLDR-18267 BRS v47: update logKnownIssues

srl295:
CLDR-18267

Three tickets were already closed but needed the logKnownIssues adjusted:
- CLDR-15663 to CLDR-17857
- CLDR-17756 to CLDR-17945
- CLDR-17544 to CLDR-17849

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4324, opened 2025-01-30
CLDR-18267 site: document how to handle known issues

srl295:
CLDR-18267

see #4322 for the current known issues PR

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/cldr-big-red-switch/brs-log-known-issues.md
+++ b/docs/site/development/cldr-big-red-switch/brs-log-known-issues.md
@@ -1,0 +1,54 @@
+---
+title: 'BRS: Handling Known Issues'
+---
+
+Sometimes, we skip tests during development “temporarily” rather than comment them out or remove them, expecting to reinstate them once the relevant bug is fixed.  To achieve this end, the "logKnownIssues" mechanism has been used, shared with the [ICU project](https://icu.unicode.org/setup/eclipse/time).
+
+See [Testing!](/development/development-process/#testing) for more details on how this is used during development.
+
+For this BRS task there are two goals. FIRST, make sure that any log known issue targets that are remaining issues, point to a currently open ticket.  SECOND, make sure that any issues that actually were fixed, have the "logKnownIssue" call removed to make sure the regression tests are actually run.
+
+Steps:
+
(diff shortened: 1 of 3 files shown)


docs/site/development/cldr-big-red-switch/brs-log-known-issues.md
```diff
@@ -0,0 +1,54 @@
+---
+title: 'BRS: Handling Known Issues'
+---
+
+Sometimes, we skip tests during development “temporarily” rather than comment them out or remove them, expecting to reinstate them once the relevant bug is fixed.  To achieve this end, the "logKnownIssues" mechanism has been used, shared with the [ICU project](https://icu.unicode.org/setup/eclipse/time).
+
+See [Testing!](/development/development-process/#testing) for more details on how this is used during development.
+
+For this BRS task there are two goals. FIRST, make sure that any log known issue targets that are remaining issues, point to a currently open ticket.  SECOND, make sure that any issues that actually were fixed, have the "logKnownIssue" call removed to make sure the regression tests are actually run.
```
macchiati, line 9: For the future, linebreak between sentences: https://sembr.org/

macchiati (approved):

docs/site/development/development-process.md
```diff
@@ -96,7 +96,10 @@ When a ticket is in design, the owner is responsible for bringing back to the co
 
 Skipping test failures
 
-If there is a test failure that is due to a bug that cannot be fixed right now (eg it requires a data change, but it is past the data freeze), then use the logKnownIssue() method to disable the test. See http://site.icu-project.org/setup/eclipse/time for details, and look in the CLDR test code for some examples of usage.
+If there is a test failure that is due to a bug that cannot be fixed right now (eg it requires a data change, but it is past the data freeze), then use the logKnownIssue() method to disable the test. See <https://icu.unicode.org/setup/eclipse/time> for details, and look in the CLDR test code for some examples of usage.
```
AEApple, line 99: Question about the 'but it is past the data freeze'. I think we need to be more specific here, since we can make changes after the data freeze for issues during alpha & beta if the CLDR TC and the ICU TC agree.

AEApple: Maybe something like: 'when it is past the data freeze and the CLDR TC considers it too risk or not a priority to fix in the current release'.

pedberg-icu (approved):
Looks like Annemaries's comment was addressed.

PR #4325, opened 2025-01-30
CLDR-18256 Add TimePrecision option to semantic skeletons

sffc:
CLDR-18256

As discussed in the CLDR Design WG.

The spec I ended up implementing here is almost the same as proposed in [the doc](https://docs.google.com/document/d/1SkxoitlCFiQ_KGW3dmRk7lbumGd_N1rfYJLMlkiXvh4/edit?tab=t.0), but I touched up some other things while I was in here.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -2476,29 +2476,23 @@ Note: A _calendar period_ is distinct from a _date_ because it cannot be paired 
 
 ##### <a name="Semantic_Time_Field_Sets" href="#Semantic_Time_Field_Sets">Time Field Sets</a>
 
-A _time field set_ refers to a particular time of day. Low-order fields, such as the minute or second, could be omitted.
+The **Time** field signifies the time of day.
 
-The fields that may be included in a time field set are:
+Whether to include the Hour, Minute, Second, or Fractional Second is configured with the [Time Precision](#Time_Precision) option.
 
-1. **Hour:** The hour, possibly with a day period, depending on factors such as the length, locale, and hour cycle locale keyword.
-2. **Minute:** The minute within an hour.
-3. **Second:** The second within a minute, a decimal number that may include fractional digits. See the [Fractional Second Digits](#Semantic_Skeleton_Fractional_Second_Digits) option.
+| Field Set | Example         |
+|-----------|-----------------|
+| { Time }  | 4:03 pm	/ 16:03 |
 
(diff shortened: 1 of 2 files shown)


sffc:
@richgillam You had some opinions on the TimePrecision options. PTAL at what I ended up with. It doesn't represent all possible values of min/max time fields, but it covers the use cases we know about, and we can add more when needed.

sffc:
Also CC @echeran who has been working with the resolution algorithm.

richgillam (dismissed):
This looks good.  I like where you ended up here.

sffc:
@AEApple said on Jira: "Would you mind updating the PR to update the modifications section as well as part of this PR"

Could someone point me to where to update the modifications section?

docs/ldml/tr35-dates.md
```diff
@@ -2697,13 +2705,23 @@ This means that the Year, Month, and Day semantic field mapping in en-US should
 | Month          | Japanese  | MMMM   | MMM    | M      |
 | Day            | Japanese  | d      | d      | d      |
 
+##### <a name="Semantic_Time_Precision_Skeleton_Variations" href="#Semantic_Time_Precision_Skeleton_Variations">Time Precision Skeleton Variations</a>
+
+The [time precision](#Semantic_Skeleton_Time_Precision) should change the skeleton for all lengths as follows:
+
+- Hour: No change.
+- Minute: Add "m"
```
macchiati, line 2713: Add "mm"? Same below, you're adding the padded minutes/seconds, right?

macchiati (changes requested):
Looks good except for one area.

AEApple:
> @AEApple said on Jira: "Would you mind updating the PR to update the modifications section as well as part of this PR"
> 
> Could someone point me to where to update the modifications section?

It's at the bottom of the page I linked in the PR (search for 'modifications')

docs/ldml/tr35-dates.md
sffc: availableFormats will fix the length if necessary for the locale.

macchiati:
Modifications missing still.

richgillam (approved):
Changes since my last review also look good.

macchiati (approved):

macchiati:
@sffc I think you need to squash to fix the message validation; then you can merge it.

PR #4328, opened 2025-01-30
CLDR-18138 kbd: fix DTD error for <import base>

srl295:
- DTD incorrectly had an implied value of 'cldr'
- No spec change needed.

CLDR-18138

ALLOW_MANY_COMMITS=true

### DIFF
--- a/keyboards/dtd/ldmlKeyboard3.xsd
+++ b/keyboards/dtd/ldmlKeyboard3.xsd
@@ -50,15 +50,10 @@ Note: DTD @-annotations are not currently converted to .xsd. For full CLDR file 
   <xs:element name="import">
     <xs:complexType>
       <xs:attribute name="path" use="required"/>
-      <xs:attribute name="base">
-        <xs:simpleType>
-          <xs:restriction base="xs:token">
-            <xs:enumeration value="cldr"/>
-          </xs:restriction>
-        </xs:simpleType>
-      </xs:attribute>
+      <xs:attribute name="base" type="xs:NMTOKEN"/>
     </xs:complexType>
   </xs:element>
+  
   
   <xs:element name="locales">
     <xs:complexType>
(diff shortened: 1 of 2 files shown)


keyboards/dtd/ldmlKeyboard3.xsd
```diff
@@ -50,16 +50,11 @@ Note: DTD @-annotations are not currently converted to .xsd. For full CLDR file
...
     <xs:complexType>
       <xs:attribute name="path" use="required"/>
-      <xs:attribute name="base">
-        <xs:simpleType>
-          <xs:restriction base="xs:token">
-            <xs:enumeration value="cldr"/>
-          </xs:restriction>
-        </xs:simpleType>
-      </xs:attribute>
+      <xs:attribute name="base" type="xs:NMTOKEN"/>
```
mcdurdin, line 53: The spec says:

> The base may be omitted (indicating a local import) or have the value "cldr".

Whereas this change is allowing any value in `base`, isn't it? 

As far as I can tell, the previous .xsd was actually correct? It doesn't have `use="required"` but it does restrict the allowable values to 'cldr'.

I think the DTD change looks okay though.

mcdurdin (changes requested):

keyboards/dtd/ldmlKeyboard3.xsd
miloush: I don't mind allowing different bases, it is up to implementations which sources they want to support.

srl295:
Okay I'm reviewing the XML spec, and the XSD spec.

Maybe the DTD was actually correct before, as was the XSD, and I may have had some other toolchain issue after which I incorrectly interpreted the DTD. 

@miloush 
> I don't mind allowing different bases, it is up to implementations which sources they want to support.

Formally this PR does *not* allow different bases, as the `@MATCH` comment on the DTD, and the wording of the spec, brings a restriction.  It is true that the XSD in this PR does allow different bases.

If the spec should allow other bases, that would need to be proposed separately, I think.  An implementation could also use a named extension to add a different base, such as `<import srl:importbase="something else" path="…"/>`

So I am inclined to close the PR and the ticket.

srl295:
closing for now

PR #4329, opened 2025-01-31
CLDR-18271 Update release page cldr-47.md

macchiati:
CLDR-18271

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-47.md
+++ b/docs/site/downloads/cldr-47.md
@@ -17,27 +17,31 @@ adapting software to the conventions of different languages.
 
 The most significant changes in this release are:
 
-- TBD
+   - New locales:
+      - Core data for Coptic (cop) , Haitian Creole (ht)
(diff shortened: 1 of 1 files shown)


macchiati:
We could go ahead and merge this, as discussed in the meeting.

macchiati:
Am squashing because of bad commit statement.

AEApple:
@conradarcturus - Are there other changes to the supplemental changes that we should include? Feel free to make a commit against this branch before the TC meeting tomorrow if you have time. We'll have to squash this PR in order to pass validation when we're ready to update the release page following tomorrow's TC meeting.

AEApple (approved):
I went ahead and squashed this since I still saw it failing due to a bad commit message.

I've made some more changes so if it looks good to you we should go ahead and merge it. Otherwise, I'm happy to re-review later tonight if you'd like to make more changes.

macchiati:
Squashed and deleted branch; we can follow up with other PRs

PR #4330, opened 2025-02-01
CLDR-17014 Use singleton to avoid synchronized in ExtraPaths

btangmu:
CLDR-17014

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
@@ -1,100 +1,82 @@
 package org.unicode.cldr.util;
 
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 
 public class ExtraPaths {
 
-    private static final Map<NameType, ExtraPaths> instances = new HashMap<>();
-
-    public static synchronized ExtraPaths getInstance(NameType nameType) {
-        return instances.computeIfAbsent(nameType, ExtraPaths::new);
+    public static void add(Collection<String> toAddTo) {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
```diff
@@ -1,100 +1,82 @@
...
-                break;
-            case SCRIPT:
-                codeType = StandardCodes.CodeType.script;
-                break;
-            default:
-                throw new IllegalArgumentException("TODO: CodeType.fromNameType");
+        Singleton() {
+            paths = new HashSet<>();
+            addPaths(NameType.SCRIPT);
+            addPaths(NameType.LANGUAGE);
```
macchiati, line 25: At the end of this method, you need to make paths be an ImmutableSet. Add an extra parameter to addPaths, etc. and change this to:
```suggestion
            Set<String> temp = new TreeSet(); // might as well put them in order; easier to debug
            addPaths(NameType.SCRIPT, temp);
            addPaths(NameType.LANGUAGE, temp);
            paths = ImmutableSet.copyOf(temp); // preserves order (Sets.copyOf doesn't)
```

macchiati (changes requested):
Almost there, thanks!

btangmu:
@macchiati the 2nd commit makes `paths` an ordered ImmutableSet as you suggested

macchiati (approved):
It would be slightly better to add pathsTemp as a parameter to addPaths, but this works and only costs a trifling bit of memory.

btangmu:
> It would be slightly better to add pathsTemp as a parameter to addPaths, but this works and only costs a trifling bit of memory.

Making it a parameter might reduce the data size by a few bytes, but increase the code size by a slightly larger number of bytes, since addPaths and addAltPath have dozens of callers. Anyway, this can be revisited in follow-up refactoring when moving more code from CLDRFile to ExtraPaths.

macchiati:
good point!

PR #4331, opened 2025-02-03
CLDR-18091 Add Mexican and Guatemalan languageMatch distances in languageInfo.xml

MarcAbonce:
CLDR-18091

ALLOW_MANY_COMMITS=true


macchiati (changes requested):
I added comments in the ticket, so I'd like to pause this until that's resolved.

MarcAbonce:
Based on the conversation in Jira, I'm going to close this PR.

Note, in case anyone stumbles upon this PR in the future:
These fallbacks can be inferred through [likelySubtags.xml](https://github.com/unicode-org/cldr/blob/main/common/supplemental/likelySubtags.xml).
So if I get `yua` locale, I can infer a fallback through `yua` → `yua_Latn_MX`, then crop it to `und_MX` (or `und_Script_Territory` in some cases) and then to `es_Latn_MX`.

PR #4332, opened 2025-02-03
CLDR-17014 Move more extra-path code from CLDRFile to ExtraPaths

btangmu:
-This is only refactoring

CLDR-17014

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
@@ -23,10 +23,10 @@ public class ExtraPaths {
      * values in root. NOTE: if this is changed, you'll need to modify
      * TestPaths.extraPathAllowsNullValue
      */
-    static final Set<String> CONST_EXTRA_PATHS =
+    private static final Set<String> CONST_EXTRA_PATHS =
             CharUtilities.internImmutableSet(
                     Set.of(
-                            // Individual zone overrides — were in getRawExtraPaths
+                            // Individual zone overrides
                             "//ldml/dates/timeZoneNames/zone[@type=\"Pacific/Honolulu\"]/short/generic",
                             "//ldml/dates/timeZoneNames/zone[@type=\"Pacific/Honolulu\"]/short/standard",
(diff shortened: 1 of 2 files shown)


btangmu:
I had checked "This PR completes the ticket", but have now unchecked it based on discussion in the ticket, since we are considering whether TERRITORY, VARIANT, CURRENCY, and CURRENCY_SYMBOL should get the same treatment as LANGUAGE and SCRIPT -- move them from ‎XMLSource::ResolvingSource::constructedItems into ExtraPaths and have their codes not show up as values

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
```diff
@@ -2,13 +2,71 @@
...
+            CharUtilities.internImmutableSet(
+                    Set.of(
+                            // Individual zone overrides
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Pacific/Honolulu\"]/short/generic",
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Pacific/Honolulu\"]/short/standard",
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Pacific/Honolulu\"]/short/daylight",
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Europe/Dublin\"]/long/daylight",
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Europe/London\"]/long/daylight",
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Etc/UTC\"]/long/standard",
+                            "//ldml/dates/timeZoneNames/zone[@type=\"Etc/UTC\"]/short/standard",
```
srl295, line 36: Why are these static? seems extremely fragile. 

I know this isn't new and it's from refactoring, but how/why did we end up with a static list in the first place?

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
macchiati: These were items that the TC agreed to add.

macchiati (approved):

PR #4336, opened 2025-02-03
CLDR-17014 Handle TERRITORY paths like LANGUAGE and SCRIPT

btangmu:
-Use ExtraPaths instead of XMLSource.ResolvingSource.constructedItems

-Paths constructed in this way no longer have fallback values

-Delete getPathsAllowingDuplicates and related code, now unused

CLDR-17014

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
@@ -80,8 +80,9 @@ public class ExtraPaths {
 
         Singleton() {
             pathsTemp = new TreeSet<>();
-            addPaths(NameType.SCRIPT);
             addPaths(NameType.LANGUAGE);
+            addPaths(NameType.SCRIPT);
+            addPaths(NameType.TERRITORY);
             addMetazones();
             pathsTemp.addAll(CONST_EXTRA_PATHS);
             paths = ImmutableSet.copyOf(pathsTemp); // preserves order (Sets.copyOf doesn't)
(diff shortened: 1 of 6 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
```diff
@@ -101,14 +101,19 @@ private void addPaths(NameType nameType) {
...
+                    codes.remove(LocaleNames.ROOT);
+                    codes.addAll(
+                            List.of(
+                                    "ar_001", "de_AT", "de_CH", "en_AU", "en_CA", "en_GB", "en_US",
+                                    "es_419", "es_ES", "es_MX", "fa_AF", "fr_CA", "fr_CH", "frc",
+                                    "hi_Latn", "lou", "nds_NL", "nl_BE", "pt_BR", "pt_PT", "ro_MD",
+                                    "sw_CD", "zh_Hans", "zh_Hant"));
+                    break;
+                case TERRITORY:
+                    codes.addAll(List.of("XA", "XB"));
```
macchiati, line 115: This is not adding an alt="variant" version of the paths.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
btangmu: Right. This corresponds to these two lines removed from XMLSource:
```
            addFallbackCode(NameType.TERRITORY, "XA", "XA");
            addFallbackCode(NameType.TERRITORY, "XB", "XB");`
```
Neither the old nor the new code involves alt. What change are you requesting?

btangmu: Note that the old code is calling this method:

```
        private static void addFallbackCode(NameType nameType, String code, String value) {
            addFallbackCode(nameType, code, value, null);
        }
```

Here `null` corresponds to `alt` in this method:

```
        private static void addFallbackCode(
                NameType nameType, String code, String value, String alt) {
            String fullpath = nameType.getKeyPath(code);
            String distinguishingPath = addFallbackCodeToConstructedItems(fullpath, value, alt);
            if (nameType == NameType.SCRIPT || nameType == NameType.TERRITORY) {
                allowDuplicates.put(distinguishingPath, code);
            }
        }
```

btangmu: Also note that this PR has 2 commits, and the first had a mistake:
```
                    addAltPath("XA", "variant", nameType);
                    addAltPath("XB", "variant", nameType);
```
That was corrected in the 2nd commit.

macchiati: Ah, right. We need both the main path and the alternate.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCLDRFile.java
```diff
@@ -924,7 +926,7 @@ public void TestSwissHighGerman() {
         }
     }
 
-    public void TestExtraPaths() {
+    public void TestExtraPathsCapitalT() { // CapitalT to distinguish from testExtraPaths above
```
macchiati, line 929: Minor, but better as TestExtraPaths2 to distinguish. The CapitalT is unexpected.

btangmu: You're right. I've just changed it to `testExtraPaths2`. I guess I was unconsciously expressing some frustration about the existence of two functions, testExtraPaths and TestExtraPaths. The convention is for all Java methods to start with a lowercase letter. I don't know why test functions would be exceptions to that rule; maybe tech debt. In TestFmwk.java we have:

`if (name.startsWith("Test") || name.startsWith("test")) {`

srl295 (approved):
good refactoring

macchiati (approved):

PR #4337, opened 2025-02-04
CLDR-18271 Flesh out release notes on collation, RBNF (some), segments, transforms

pedberg-icu:
CLDR-18271

Flesh out release notes on collation, RBNF (some), segments, transforms

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-47.md
+++ b/docs/site/downloads/cldr-47.md
@@ -82,7 +82,35 @@ For a full listing, see [Delta Data](https://unicode.org/cldr/charts/dev/delta/i
 
 ### Collation Data Changes
 
-- TBD
+- Two old `zh` collation variants are removed: big5han and gb2312.
+([CLDR-16062](https://unicode-org.atlassian.net/browse/CLDR-16062))
+
+### Number Spellout Data Changes
+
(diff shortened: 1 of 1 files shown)


AEApple (approved):

grhoten (approved):
I'm not sure that it's worthy of mentioning the RBNF changes, but it's also correct. It's kind of like the rest of the CLDR data changes.

On a related topic, for the next version of ICU, it would be worth mentioning the | operator in the RBNF syntax in an optional span.  That new syntax will likely be used in the next version of CLDR after this CLDR release to fix the large Lithuanian and Russian ordinals.

macchiati (approved):

macchiati:
Merging; we decided to do quick PRs for different changes.

@grhoten, can you file a ticket to add the RBNF syntax to https://unicode.org/reports/tr35/tr35-numbers.html#Rule-Based_Number_Formatting and include the '|' syntax that you mention? We want to put specs for collation, etc. into the LDML spec so that they can be referenced by more than ICU, eg ICU4X

grhoten:
@macchiati CLDR-18285 and CLDR-10544 have been submitted to update the RBNF documentation in the CLDR spec.

PR #4338, opened 2025-02-04
CLDR-18277 json: update PACKAGES.md

srl295:
- subdivisions should be tiered for consistency (and if we subset it someday)
- transforms is NOT tiered
- calculate tier from output directory, where possible
- also cleanup PACKAGES.md

CLDR-18277

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
@@ -103,7 +103,7 @@ public class Ldml2JsonConverter {
         annotationsDerived,
         bcp47(false, false),
         transforms(false, false),
-        subdivisions(false, true);
+        subdivisions(true, true);
 
         private final boolean isTiered;
         private final boolean hasLocales;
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
```diff
@@ -103,7 +103,7 @@ enum RunType {
         annotationsDerived,
         bcp47(false, false),
         transforms(false, false),
-        subdivisions(false, true);
+        subdivisions(true, true);
```
btangmu, line 106: isTiered, hasLocales

tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
```diff
@@ -1692,7 +1694,20 @@ public void writePackageList(String outputDir) throws IOException {
...
+                RunType r = RunType.valueOf(baseName);
+                if (!r.tiered()) {
+                    tiered = false;
+                }
+            } catch (IllegalArgumentException t) {
+                // ignored
+            }
+
+            // if not tiered
+            if (!tiered) {
```
btangmu, line 1710: ideally there should be a subroutine `isTiered` to encapsulate this, and it would return false as soon as it saw "core". The surrounding method `writePackageList` is too long, over 100 lines; it's writing two files at once, cldr-packages.json and PACKAGES.md. It would be helpful -- make changes easier to review -- if it were organized into subroutines

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
srl295: yes and part of the challenge is that it's not looking at RunType at that point but output directories.  But agree on subroutines.

PR #4340, opened 2025-02-05
CLDR-18287 BRS Generate test data

macchiati:
BRS task

CLDR-18287

Note: this does not generate the datetime.json file nor the MF2 files. The MF2 files are ok; they are independent of changes in cldr data. I have a message in to Elango about looping in the generator for the datetime.json file.

ALLOW_MANY_COMMITS=true


common/testData/localeIdentifiers/likelySubtags.txt
```diff
@@ -802,11 +812,6 @@ pl-AQ ;	pl-Latn-AQ ;	pl-AQ ;
 pl-Egyp ;	pl-Egyp-PL ;	pl-Egyp ;	
 pl-Latn ;	pl-Latn-PL ;	pl ;	
 pl-PL ;	pl-Latn-PL ;	pl ;	
-prg ;	prg-Latn-PL ;	prg ;
```
AEApple, line 805: Why is prg being removed? It doesn't make sense to me that it dropped coverage between CLDR 46 and 47 since it was only at Basic and AFAIK nothing changed in Basic in 47. Do you know what happened?

macchiati: I would like to go ahead and merge this; it shouldn't be a blocker for the alpha

AEApple (approved):

common/testData/localeIdentifiers/likelySubtags.txt
AEApple: Okay, approved and we can investigate the change in coverage elsewhere.

PR #4343, opened 2025-02-05
CLDR-18240 Check for null baileyValue in VoteResolver

btangmu:
-In resolveVotes, do not return INHERITANCE_MARKER if baileyValue is null

-Do not misinterpret the boolean baileySet as implying baileyValue is not null

-Always call getBaileyValue instead of reading baileyValue directly, to enforce the requirement that setBaileyValue must be called before getBaileyValue

-Make VoteResolver.getBaileyValue private; it was not called by other classes

-Comments; clarify that baileySet may be true even if baileyValue is false; it only indicates that setBaileyValue was called

CLDR-18240

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
@@ -604,7 +604,7 @@ public class VoteResolver<T> {
         private final Map<Organization, T> orgToAdd = new EnumMap<>(Organization.class);
 
         private T baileyValue;
-        private boolean baileySet; // was the bailey value set
+        private boolean baileySet; // was the bailey value set (possibly to null)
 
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/VoteResolver.java
```diff
@@ -1219,7 +1221,7 @@ private void resolveVotes() {
                         "Winning Value: '%s' with status '%s' because there were no unconflicted votes.",
                         winningValue, winningStatus);
                 // Declare the winner here, because we're about to return from the function
-            } else if (organizationToValueAndVote.baileySet) {
+            } else if (getBaileyValue() != null) {
```
btangmu, line 1224: this is the essential bug fix

macchiati (approved):

PR #4344, opened 2025-02-05
CLDR-18133 supplemental currency needs to be @ORDERED

srl295:
CLDR-18133

ALLOW_MANY_COMMITS=true


srl295:
This doesn't fix the JSON issue, but it was discovered while working on it…

srl295:
Something's wrong, because CLDRFile with a DTDComparator is returning lexical instead of file order.

```
//supplementalData/currencyData/region[@iso3166="BY"]/currency[@iso4217="BYB"][@from="1994-08-01"][@to="2000-12-31"]
//supplementalData/currencyData/region[@iso3166="BY"]/currency[@iso4217="BYN"][@from="2016-07-01"]
//supplementalData/currencyData/region[@iso3166="BY"]/currency[@iso4217="BYR"][@from="2000-01-01"][@to="2017-01-01"]
//supplementalData/currencyData/region[@iso3166="BY"]/currency[@iso4217="RUR"][@from="1991-12-25"][@to="1994-11-08"]
//supplementalData/currencyData/region[@iso3166="BY"]/currency[@iso4217="SUR"][@from="1961-01-01"][@to="1991-12-25"]
```

srl295:
OK. I fixed the DTD and added a test.

I'll fix the JSON in a followon PR.

macchiati (approved):

macchiati (approved):

srl295:
FYI this *did* fix the JSON data- but only when the dtd is also copied into cldr-staging, since the json converter uses CLDR_DIR of cldr-staging.

macchiati:
Good!

Using the cldr saying version seems fragile, unless there is a test that it
matches head of cldr

macchiati:
Makes sense. I'm just wondering how we can make sure that other people
don't stumble over this problem in the future.

macchiati:
Makes sense.

PR #4347, opened 2025-02-06
CLDR-18280 Remove gmtZeroFormat

robertbastian:
CLDR-18280

ALLOW_MANY_COMMITS=true


macchiati:
I don't see mention of the new pattern symbol here.

robertbastian:
It's in the table, with the other ones

macchiati (changes requested):
I see no explanation of "justified" in the table (2293). It only occurs further up in the spec. So when someone is looking at the table (2293), they will not know what that means. And it isn't really explained on line 1709. This needs a short sentence explaining the terminology in both places, eg so that a column of timezone offsets has roughly the same width in each line.

(Side note: justified has multiple meanings in English, including https://en.wikipedia.org/wiki/Justified_(TV_series))

robertbastian:
The other formats also aren't really explained in that paragraph.

Would you prefer "fixed-length" over "justified"?

Also, why are the "jira-ticket" and "single-commit" checks failing?

macchiati:
Yes, I think fixed-length would be better, and that doesn't need a gloss.

- I don't know why "Pull request title must start with a Jira ticket ID" is failing. Looks like you do that.
- I don't recall seeing "This PR has undefined commits; consider squashing:" before. Or maybe my gaze just passes over it, because I don't think either of these block merging.
- The "@github-actions cldr-mvn / build (pull_request) Failing after 3m" is very odd; not sure why CheckArchive is called and failing.

@srl295 Any thoughts?

AEApple:
I thought it might be due to characters in the title. I've seen it if you have extra dashes before, but I tried removing a few and it didn't fix it.

macchiati:
Yes, I recall that also. I try to stick to just  ASCII letters and digits.

robertbastian:
> I try to stick to just ASCII letters and digits.

Brought to you by the Unicode people 😅

macchiati:
> > I try to stick to just ASCII letters and digits.
> 
> Brought to you by the Unicode people 😅

#😶‍🌫️

I meant to say: 
with ASCII, I stick to just letters and digits — meaning avoid ASCII punctuation, because they sometimes trigger weird behavior in GitHub commit messages and branch names.

docs/ldml/tr35-dates.md
```diff
@@ -1706,12 +1706,16 @@ Note: A generic location format is constructed by a part of time zone ID represe
 * "UTC-3" (short)
 * "Гринуич+03:30" (long)
 
-Otherwise (when the offset from GMT is zero, referring to GMT itself) the style specified by the `<gmtZeroFormat>` element is used:
```
robertbastian, line 1709: Actually I don't agree with the assumption that a zero GMT offset refers to GMT itself. Azores Summer Time is GMT+0, but that doesn't make it GMT, it's still Azores Summer Time, just with a GMT offset of zero. I would actually prefer if this format never used the `gmtZeroFormat` and we wouldn't need to introduce a new format for this behaviour.

macchiati: We didn't have time in the last design meeting to discuss this (and I assume it is targeted to v48),

robertbastian: No this is 47

macchiati (dismissed):

macchiati:
Ok, a jira message needs fixing. Please hit the Details link to squash with a corrected message

macchiati:
Robert, please hit the Details button and squash so we can merge this!

Screenshot 2025-02-18 at 15 53 16

robertbastian:
Doesn't seem to work.

I don't actually want to merge this anyway. Introducing a new symbol for it is a quick and dirty solution that unblocks me in the short term, but I believe the existing symbols should use GMT+0 instead. I haven't had time to discuss this in any of the meetings.

common/dtd/ldml.dtd
```diff
@@ -1761,13 +1761,18 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@METADATA-->
 
 <!ELEMENT gmtZeroFormat ( #PCDATA ) >
+    <!--@METADATA-->
+    <!--@DEPRECATED-->
 <!ATTLIST gmtZeroFormat alt NMTOKENS #IMPLIED >
+    <!--@METADATA-->
```
macchiati, line 1767: If the element is deprecated, then all attributes will be. Does't hurt though.

macchiati (commented):
This should probably wait until survey tool XML integration. Otherwise importing the data from XML might restore items.

macchiati:
We are not in XML yet. If this were merged now it would be overwritten.

robertbastian:
I'm aware. A review would be nice regardless

macchiati (commented):
Note: the downside of this is that it will break clients. So we need to clear this with ICU* first. I assume that ICU4X is ok with it; please file a ticket with ICU (cf @markusicu) to make sure they they see it before we integrate, and can make the fix in this release. It also needs a migration note in the release page.

PR #4348, opened 2025-02-06
CLDR-18128 Pair classical with semantic skeletons in datetime test data

echeran:
CLDR-18128

This PR is another update to the datetime test data generation that support semantic skeletons, introduced in #4313. (A previous recent followup update PR was #4341)

The purpose of this PR's update is to include the "classical" skeleton string in test cases whenever the semantic skeleton is used as input. The reason is that i18n implementations like ICU don't have support for such a new concept as semantic skeletons. However, they can still format regular "classical" skeletons in order to execute the test case and verify the expected formatted output string is correct.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/testData/datetime/datetime.json
+++ b/common/testData/datetime/datetime.json
@@ -152,6 +152,7 @@
   {
     "semanticSkeleton": "YMDE",
     "semanticSkeletonLength": "short",
+    "classicalSkeleton": "yyMdEEE",
     "calendar": "gregorian",
     "locale": "en",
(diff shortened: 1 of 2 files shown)


macchiati (dismissed):

sffc (approved):

echeran:
Just need a rubber stamp now after applying the formatter changes and needing to pull in the latest from `main`. All tests are passing.

macchiati (approved):

PR #4349, opened 2025-02-06
CLDR-11874 Test to verify Subdivisions vs Common

srl295:
CLDR-11874

- test to verify common/subdivisions vs common/main
- tool to copy from common/main to common/subdivisions
- logKnownIssue on CLDR-18296 for any data in common/subdivisions missing from common/main

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/subdivisions/af.xml
+++ b/common/subdivisions/af.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2024 Unicode, Inc.
+<!-- Copyright © 1991-2025 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-3.0
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
@@ -421,7 +421,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<subdivision type="gbdur" draft="provisional">Durham-graafskap</subdivision>
 			<subdivision type="gbedh" draft="provisional">Edinburg</subdivision>
 			<subdivision type="gbels" draft="provisional">Buite-Hebride</subdivision>
-			<subdivision type="gbeng" draft="provisional">Engeland</subdivision>
+			<subdivision type="gbeng">Engeland</subdivision>
 			<subdivision type="gbery" draft="provisional">Oos-Yorkshire</subdivision>
 			<subdivision type="gbess" draft="provisional">Essex</subdivision>
 			<subdivision type="gbesx" draft="provisional">Oos-Sussex</subdivision>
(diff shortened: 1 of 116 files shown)


macchiati:
The common/main is primary, so the json generation code should just skip paths in a common/subdivisions locale that have values in the corresponding common/main locale.

srl295:
The generation code just takes a CLDRFile, it would have to do some
acrobatics in order to pull from the other file. We don’t have any thing
that does that sort of loading.

It might be a lot simpler and better from a data perspective to copy from
common/main into subdivisions - would that work?

common/subdivisions/af.xml
```diff
@@ -421,7 +421,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<subdivision type="gbdur" draft="provisional">Durham-graafskap</subdivision>
 			<subdivision type="gbedh" draft="provisional">Edinburg</subdivision>
 			<subdivision type="gbels" draft="provisional">Buite-Hebride</subdivision>
-			<subdivision type="gbeng" draft="provisional">Engeland</subdivision>
+			<subdivision type="gbeng">Engeland</subdivision>
```
srl295, line 424: approval status copied from common/main (via survey) into subdivisions

common/subdivisions/fr.xml
```diff
@@ -1615,7 +1615,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<subdivision type="gbwkf" draft="provisional">cité de Wakefield</subdivision>
 			<subdivision type="gbwll" draft="provisional">district métropolitain de Walsall</subdivision>
 			<subdivision type="gbwln" draft="provisional">West Lothian</subdivision>
-			<subdivision type="gbwls" draft="provisional">pays de Galles</subdivision>
+			<subdivision type="gbwls">Pays de Galles</subdivision>
```
srl295, line 1618: an actual change (and approval) from common/main

common/subdivisions/kgp.xml
```diff
@@ -0,0 +1,20 @@
...
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="kgp"/>
+	</identity>
+	<localeDisplayNames>
+		<subdivisions>
+			<subdivision type="gbeng">Ĩgratéra</subdivision>
+			<subdivision type="gbsct">Enhkósija</subdivision>
+			<subdivision type="gbwls">Vares</subdivision>
```
srl295, line 17: data in a new locale

srl295:
with this change cldr json generation works properly.

macchiati (approved):
Very good.

One request: could you file a ticket to make CLDRFile.make be thread-safe, to avoid 
```
                                synchronized (CLDRFile.class) {
                                    mainF = mainFactory.make(loc, true);
                                    subF = subFactory.make(par.getBaseName(), true);
                                }
```

I'm sure there are other places in the code that aren't as careful as you are!

srl295:
> Very good.
> 
> One request: could you file a ticket to make CLDRFile.make be thread-safe, to avoid
> 
> ```
>                                 synchronized (CLDRFile.class) {
>                                     mainF = mainFactory.make(loc, true);
>                                     subF = subFactory.make(par.getBaseName(), true);
>                                 }
> ```
> 
> I'm sure there are other places in the code that aren't as careful as you are!

Couldn't reproduce the issue now, i thought I had put a stack trace somewhere but can't find it.

PR #4350, opened 2025-02-06
CLDR-18271 Cleanup release note for alpha

macchiati:
CLDR-18271

- Removed the MF section (that was under Data when it should have been under Spec), and added a reference to that in the Spec section.
- Fixed Fredrik's items
- Bolded all TBDs.

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

pedberg-icu:
Commit messages are bad...

macchiati:
Dang, will squash. For some reason I find that harder to remember with the web github branches.

AEApple (approved):

PR #4231, opened 2024-12-08
CLDR-18149 v46.1: Pourover of MessageFormat 2.0 Final Candidate

aphillips:
CLDR-18149

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4353, opened 2025-02-07
CLDR-18295 spec: fixes to MFv2 46.1 spec 

srl295:
- corrected indents, where there were multiple top level # - indented everything in by one
- ~corrected heading names with period in them which resulted in broken links~
- ~removed "Message Format 2.0" as redundant in two headings~
- ~demoted `.abnf` to a subheading~
- correct links to /spec/ into internal links

- also ran fix-tocs on tr35-info.md

(Review note: please review ignoring whitespace changes.)

CLDR-18295

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-messageFormat.md
+++ b/docs/ldml/tr35-messageFormat.md
@@ -56,7 +56,7 @@ The LDML specification is divided into the following parts:
   * [Design Goals](#design-goals)
   * [Design Restrictions](#design-restrictions)
 * [Messages and their Syntax](#messages-and-their-syntax)
-  * [Well-formed vs Valid Messages](#well-formed-vs-valid-messages)
+  * [Well-formed vs. Valid Messages](#well-formed-vs-valid-messages)
 * [The Message](#the-message)
   * [Declarations](#declarations)
   * [Complex Body](#complex-body)
@@ -81,6 +81,7 @@ The LDML specification is divided into the following parts:
   * [Escape Sequences](#escape-sequences)
   * [Whitespace](#whitespace)
 * [Complete ABNF](#complete-abnf)
+  * [`message.abnf`](#messageabnf)
 * [Formatting](#formatting)
   * [Introduction](#introduction)
   * [Formatting Context](#formatting-context)
@@ -125,7 +126,7 @@ The LDML specification is divided into the following parts:
     * [Bad Option](#bad-option)
     * [Bad Variant Key](#bad-variant-key)
     * [Unsupported Operation](#unsupported-operation)
-* [Default Function Registry](#default-function-registry)
+* [MessageFormat 2.0 Default Function Registry](#messageformat-20-default-function-registry)
   * [String Value Selection and Formatting](#string-value-selection-and-formatting)
     * [The `:string` function](#the-string-function)
       * [Operands](#operands)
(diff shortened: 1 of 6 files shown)


macchiati (changes requested):
Two issues. 

1. If this is for the v46.1 spec, then it needs to be on the maint branch. I think that Part 9 in head is the same, so (after checking) you could put this there.
2. The master source for Part 9 is still in the message-format repo. So any changes need to be made upstream, because they'll be overwritten. (Note: we need discuss smoothing the process, but that won't happen in the near-term.)

And a question: GFM doesn't care whether a bulleted list is initiated with +, -, or * (at least not ÷ !). Does the tooling that converts to HTML in our format care?

docs/ldml/tr35-messageFormat.md
```diff
@@ -3351,7 +3353,7 @@ The following options and their values are required to be available on the funct
    -  `plural` (default)
    -  `exact`
 - `currency`
-   - well-formed [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
+   - well-formed [Unicode Currency Identifier](./tr35.md#UnicodeCurrencyIdentifier)
```
AEApple, line 3356: Why does this tr35.md have a ./ at the beginning when the rest don't?

srl295:
and to link in @aphillips 

> Two issues.

And three answers…

> 1. If this is for the v46.1 spec, then it needs to be on the maint branch. I think that Part 9 in head is the same, so (after checking) you could put this there.

The ticket was given a fixVersion of v47 as a release blocker, which is why I'm working on main and not maint. I'm not sure where 46.1 comes in.

> 2. The master source for Part 9 is still in the message-format repo. So any changes need to be made upstream, because they'll be overwritten. (Note: we need discuss smoothing the process, but that won't happen in the near-term.)

There are some automatic conversion steps in between, yes.  Perhaps there are bugs there. Again, as a release blocker, this PR is meant to correct the spec, and then we can propagate the changes.

> And a question: GFM doesn't care whether a bulleted list is initiated with +, -, or * (at least not ÷ !). Does the tooling that converts to HTML in our format care?

In the issue you quoted `+ [Date and Time Override Options](#date-and-time-override-options)`

So I think here you are referring to the table of contents as checked into the repository.  As I've mentioned, the table of contents is always removed and regenerated before the spec is deployed.  So the bulleted list starting with `+` is never seen by the markdown to HTML converter - it's always removed first.

For some reason, in #4231 the MFv2 spec (in 46.1) was checked in with a ToC that was not generated with CLDR tooling. So while the markdown has a ToC (with broken links), that's ignored and replaced on each deploy.

docs/ldml/tr35-messageFormat.md
srl295: it should not have ./

macchiati:
On the release blocker part. The MFWG content for the Beta spec has not yet
been Incorporated into the main repository. So making changes, other than
to the TOC generation tooling, will not affect the beta. Because they will
be all overridden.

The reason I mentioned 46.1, is that because the content is I believe
identical to what's in 46.1, we could use the reformed version to fix the
current release spec which is also broken.

srl295:
- periods-in-headings turned out to be a tool issue, I put a fix upstream at https://github.com/not-dalia/gfm-toc/pull/5 - so, I've restored heading names such as `### Message Format 2.0…` and `### This vs. that`



> On the release blocker part. The MFWG content for the Beta spec has not yet been Incorporated into the main repository. So making changes, other than to the TOC generation tooling, will not affect the beta. Because they will be all overridden. The reason I mentioned 46.1, is that because the content is I believe identical to what's in 46.1, we could use the reformed version to fix the current release spec which is also broken.

Okay- didn't realize that.

Perhaps this should go to @aphillips to determine what steps to take next, using this PR as a source for which issues there are. 

I'll open an upstream PR (in mfv2) of any issues I can fix there.

srl295:
Please see https://unicode-org.atlassian.net/browse/CLDR-18295?focusedCommentId=180268 for discussion on next steps.  
- Summary for this PR: proposing to merge as a fix for the 46.1 version of the spec and backport to maint-46

macchiati (approved):

PR #4358, opened 2025-02-10
CLDR-16022 Add test to detect new aliases for BCP47 keys or non-tz values

pedberg-icu:
CLDR-16022

Add a test that detects any new aliases for BCP47 keys, or for BCP47 types other than tz types (for tz types the alias contains the tzdata name so any new zones will require an alias); these should not be added. This test depends on having the cldr-archive available (since it compares BCP47 aliases in the current version to those in the previous version), and is skipped if the flag is set indicating that the archive is not available.

This also adds the unit test TestBCP47 to TestAll(); there are probably more unit tests that should be added to TestAll(), filed [CLDR-18301](https://unicode-org.atlassian.net/browse/CLDR-18301) about that.

ALLOW_MANY_COMMITS=true

[CLDR-18301]: https://unicode-org.atlassian.net/browse/CLDR-18301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAll.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAll.java
@@ -185,6 +185,7 @@ public class TestAll extends TestGroup {
                     "org.unicode.cldr.unittest.TestAnnotations",
                     "org.unicode.cldr.unittest.TestAttributeValues",
                     "org.unicode.cldr.unittest.TestBasic",
+                    "org.unicode.cldr.unittest.TestBCP47",
                     "org.unicode.cldr.unittest.TestCLDRFile",
                     "org.unicode.cldr.unittest.TestCLDRUtils",
                     "org.unicode.cldr.unittest.TestCanonicalIds",
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
@@ -19,9 +19,11 @@ import java.util.stream.Collectors;
 import org.unicode.cldr.icu.dev.test.TestFmwk;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.CLDRPaths;
 import org.unicode.cldr.util.ChainedMap;
 import org.unicode.cldr.util.CldrUtility;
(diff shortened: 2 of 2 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
```diff
@@ -422,6 +424,73 @@ public void TestMu() {
             typesFound.add(type);
         }
 
-        assertEquals("mu values are only temperature units for now", allowedBcp47Units, typesFound);
+        // Currently <convertUnits> - and consequently allowedBcp47Units - has a temperature unit
+        // rankine which is not supported for mu, and hence not in typesFound.
+        // Filed CLDR-18300 about this but meanwhile typesFound just needs to be a subset of
+        // allowedBcp47Units.
```
pedberg-icu, line 430: see [CLDR-18300](https://unicode-org.atlassian.net/browse/CLDR-18300)

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAll.java
```diff
@@ -185,6 +185,7 @@ public TestAll() {
                     "org.unicode.cldr.unittest.TestAnnotations",
                     "org.unicode.cldr.unittest.TestAttributeValues",
                     "org.unicode.cldr.unittest.TestBasic",
+                    "org.unicode.cldr.unittest.TestBCP47",
```
pedberg-icu, line 188: see  [CLDR-18301](https://unicode-org.atlassian.net/browse/CLDR-18301) "TestAll() in unittest/TestAll.java does not include all of the tests in unittest; should it?"

macchiati: Also https://unicode-org.atlassian.net/browse/CLDR-17708 — mostly the same?

macchiati (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAll.java
pedberg-icu: Yes, thanks, I resolved CLDR-18301 as a duplicate of CLDR-17708 and copied the linked issues over.

PR #4360, opened 2025-02-11
CLDR-17971 kbd: (dtd/spec) add <info attribution=…> attribute

srl295:
- and example files

CLDR-17971

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-keyboards.md
+++ b/docs/ldml/tr35-keyboards.md
@@ -923,11 +923,16 @@ _Attribute:_ `indicator`
 
 _Attribute:_ `attribution`
 
-> The `attribution` attribute describes a short string which gives some indication of the origin of the keyboard design, if different from the author of the layout file.
-> For example, an external standards or other body may have originated the layout used in the document.
+> The `attribution` attribute describes a short string which gives some indication of the originating entity of the keyboard design, if different from the author of the layout file.
+> For example, an external standards body or other entity may have originated the layout used in the document.
+> This attribute does not imply endorsement by the named entity.
 >
 > This attribute is not localized.
 > There is no requirement that an implementation display, store, or otherwise process this attribute.
+
+```xml
+<info attribution="Malta Standards Authority"/>
+```
 
 * * *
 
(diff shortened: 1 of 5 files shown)


srl295:
Review note: kbd-check will fail here until the tooling catches up with unreleased version v47. please ignore.

mcdurdin (dismissed):

miloush:
Like I mentioned at the meeting, I am not keen on saying the "attribution" attribute is to be used for "reference" purposes, which is a semantic piece of metadata. Attribution can also sound like endorsement. I am fine with adding "attribution" attribute that lists contributors, but if we want to include an information about which standard is the layout based on, maybe we could have "basedOn" attribute or "reference" or similar.

If the usage of the attribute written in this PR is what you are going for, then it should be called something else like "reference".  If the name of the attribute in this PR is what you are going for (and I have no objections to adding it), then the usage text should be rewritten to indicate it is meant for attribution, not source standards.

srl295:
> Like I mentioned at the meeting, I am not keen on saying the "attribution" attribute is to be used for "reference" purposes, which is a semantic piece of metadata. Attribution can also sound like endorsement. I am fine with adding "attribution" attribute that lists contributors, but if we want to include an information about which standard is the layout based on, maybe we could have "basedOn" attribute or "reference" or similar.
> 
> If the usage of the attribute written in this PR is what you are going for, then it should be called something else like "reference". If the name of the attribute in this PR is what you are going for (and I have no objections to adding it), then the usage text should be rewritten to indicate it is meant for attribution, not source standards.

A full semantic model might be something more like OCI annotations https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys - we could have key-value pairs for additional data. I think what you are saying is that `attribution="Malta Standards Authority"` (typo in the file, i'll update) rather than having the standard number there.

srl295:
@miloush the standards name is more appropriate as part of the keyboard name

srl295:
@miloush you said you were OK with the current formulation?

miloush:
I am OK with the attribute but the description in the spec is the opposite of what we agreed, isn't it?

srl295:
> I am OK with the attribute but the description in the spec is the opposite of what we agreed, isn't it?

Does this capture it now? Or can you suggest a change?

srl295:
@unicode-org/cldr-keyboards ping? can i get a review on this?

mcdurdin (approved):
LGTM

PR #4214, opened 2024-11-26
CLDR-17885 Add Sunu-Deva and Deva-Sunu transforms

NeilSureshPatel:
[CLDR-17885](https://unicode-org.atlassian.net/browse/CLDR-17885)

Initial commit of Sunuwar to Devanagari and Devanagari to Sunuwar transforms using the InterIndic intermediate step. Transforms are based on the SWS standard.

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4362, opened 2025-02-11
CLDR-17885 add font for Sunuwar script

NeilSureshPatel:
CLDR-17885

Added open source Sunuwar fonts for embedding

ALLOW_MANY_COMMITS=true


srl295:
isn't this already in #4214

srl295:
should use `Noto Sans Sunuwar` ?

macchiati:
Unless a better choice pops up, the Noto font is good.

srl295:
according to the notes, the noto one is the SEI one but now under Noto. All the better if we can load it via Noto.

NeilSureshPatel:
SEI Sunuwar and Noto Sans Sunuwar come from the source file so they are basically the same. They are built with different workflows and have a few differences but for the purposes of CLDR either will work well.

PR #4363, opened 2025-02-11
CLDR-18294 Avoid null Bailey values for paths with INHERITANCE_MARKER

btangmu:
-Remove some paths with INHERITANCE_MARKER in XML for eight locales

-Restore code-fallback in XMLSource.java, maybe temporarily, for some paths

-Code comments reference this ticket for follow up in v48

CLDR-18294

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/br.xml
+++ b/common/main/br.xml
@@ -7745,9 +7745,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<zone type="Asia/Phnom_Penh">
 				<exemplarCity>Phnum Pénh</exemplarCity>
 			</zone>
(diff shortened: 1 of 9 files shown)


btangmu:
@macchiati this is what I could do before leaving the office today. I didn't do anything with the personNames paths, or with those marked "Ok to remove path; inherited" in the spreadsheet.

macchiati (approved):

macchiati:
Looks good. Can you also add the test (can be separate PR) with an exception set for any remaining issues (and a logKnownIssue if that set is !isEmpty()?

PR #4364, opened 2025-02-12
CLDR-9774 Enable test for missing/extra English names for BCP47 keys/types

pedberg-icu:
CLDR-9774

Start using the test to check for missing or extra English display names for relevant BCP47 keys and types. Most of this was already in unittest/TestBCP47.java, but the tests there were not being run until I added them to TestAll() in PR [#4358](https://github.com/unicode-org/cldr/pull/4358). Then for this PR I just had to switch to using normal error logging, and add some test skips (one marked with a logKnownIssue).

Tested this by temporarily adding some new entries in common/bcp47 files without translations in English and verifying that the test caught them.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
@@ -78,9 +78,6 @@ public class TestBCP47 extends TestFmwk {
     }
 
     public void TestEnglishKeyTranslations() {
-        logKnownIssue(
-                "cldr7631",
-                "Using just warnings for now, until issues are resolved. Change WARNING/ERROR when removing this.");
         ChainedMap.M3<String, String, String> foundEnglish =
                 ChainedMap.of(
                         new TreeMap<String, Object>(), new TreeMap<String, Object>(), String.class);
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
```diff
@@ -78,9 +78,6 @@ public static void main(String[] args) {
     }
 
     public void TestEnglishKeyTranslations() {
-        logKnownIssue(
-                "cldr7631",
-                "Using just warnings for now, until issues are resolved. Change WARNING/ERROR when removing this.");
```
pedberg-icu, line 83: CLDR-7631 was resolved a while ago

macchiati (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
srl295: I don't know why my sweep didn't find this.  Is this function actually called?

srl295 (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
pedberg-icu: @srl295 Maybe because until a day or so ago the tests in this file were not being run (since they were not included in TestAl())?

PR #4366, opened 2025-02-12
CLDR-18294 Add unit test for inheriting null, and fix two items in xml

btangmu:
-Combine the test with TestCheckCLDR to avoid another cycle through factory.getAvailable

-New method checkNullWithInheritanceMark

-Skip and log known issue for paths not in LANGUAGE/SCRIPT/TERRITORY/VARIANT

CLDR-18294

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
@@ -42,6 +42,7 @@ import org.unicode.cldr.util.CLDRInfo.CandidateInfo;
 import org.unicode.cldr.util.CLDRInfo.PathValueInfo;
 import org.unicode.cldr.util.CLDRInfo.UserInfo;
 import org.unicode.cldr.util.CLDRLocale;
+import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.Counter;
 import org.unicode.cldr.util.DayPeriodInfo;
 import org.unicode.cldr.util.DayPeriodInfo.DayPeriod;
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
```diff
@@ -476,8 +478,10 @@ public void TestAllLocales() {
         LanguageTagParser ltp = new LanguageTagParser();
         Set<String> locales = new HashSet<>();
         for (String locale : getInclusion() <= 5 ? eightPointLocales : factory.getAvailable()) {
+            checkNullWithInheritanceMark(locale);
```
btangmu, line 481: Actually I've just realized, this doesn't really avoid adding a cycle through the locales, since the original test waits until after this loop to cycle with `checkLocale`

Given the hurry to get this ready, I suggest merging this anyway and following up to address the performance

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
```diff
@@ -491,6 +495,28 @@ public void TestAllLocales() {
...
+            if (CldrUtility.INHERITANCE_MARKER.equals(value)) {
+                switch (NameType.fromPath(path)) {
+                    case LANGUAGE:
+                    case SCRIPT:
+                    case TERRITORY:
+                    case VARIANT:
+                        assertNotNull(locale + " " + path, resolved.getStringValue(path));
+                        break;
+                    default:
+                        if (!logKnownIssue("CLDR-18294", "Null with inheritance mark")) {
```
srl295, line 512: needs to be a new ticket (48) for after 18294 is already fixed.

srl295 (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
btangmu: I'll follow up with that; merging now to fix the data

PR #4370, opened 2025-02-12
CLDR-18294 Fix known issue to reference future ticket

btangmu:
CLDR-18294

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
@@ -509,7 +509,7 @@ public class TestCheckCLDR extends TestFmwk {
                         assertNotNull(locale + " " + path, resolved.getStringValue(path));
                         break;
                     default:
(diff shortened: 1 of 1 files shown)


btangmu:
The previous PR also had a possible performance issue. @macchiati had written "Note: I suspect that it is best if we piggy back off of an existing related test that cycles of all locales and all paths. Otherwise we might get churn as locales are tossed out of the cache and then rebuilt from disk."

I called checkNullWithInheritanceMark from TestAllLocales but that doesn't address the performance issue well. I haven't measured the time. I can follow up on that, either in this ticket or the new v48 ticket

macchiati:
> "Note: I suspect that it is best if we piggy back off of an existing related test that cycles of all locales and all paths. Otherwise we might get churn as locales are tossed out of the cache and then rebuilt from disk."

That's for the future, because it is already widespread; this test only adds a slight increment to that (if indeed it is an issue; as you say, we have to measure first)

macchiati (approved):

macchiati:
Taking the liberty of merging in the interests of time.

PR #4371, opened 2025-02-12
CLDR-18217 Add CLDRFile Iterable interface option to include extra paths by default

btangmu:
-Private boolean CLDRFile.DEFAULT_ITERATION_INCLUDES_EXTRAS determines whether to skip extras; either way they skip paths with null values

-Rename old iterator iteratorWithoutExtras

-Iterable and iterator are distinct, as in iterableWithoutExtras versus iteratorWithoutExtras

-Call iteratorWithoutExtras/iterableWithoutExtras directly where tests would otherwise fail with DEFAULT_ITERATION_INCLUDES_EXTRAS true

CLDR-18217

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -1564,12 +1564,12 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
 
     private static class FullIterable implements Iterable<String>, SimpleIterator<String> {
         private final CLDRFile file;
-        private final Iterator<String> fileIterator; // dataSource only
+        private final Iterator<String> iteratorWithoutExtras;
         private Iterator<String> extraPaths;
 
         FullIterable(CLDRFile file) {
             this.file = file;
-            this.fileIterator = file.dataSource.iterator(); // file.iteratorWithoutExtras();
+            this.iteratorWithoutExtras = file.iteratorWithoutExtras();
         }
 
         @Override
@@ -1581,8 +1581,8 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
 
         @Override
         public String next() {
-            if (fileIterator.hasNext()) {
-                return fileIterator.next();
+            if (iteratorWithoutExtras.hasNext()) {
+                return iteratorWithoutExtras.next();
             }
             if (extraPaths == null) {
                 extraPaths = file.getExtraPaths().iterator();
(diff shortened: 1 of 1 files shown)


btangmu:
I'm not sure whether this should go into v47 or wait for v48. Another option would be to make DEFAULT_ITERATION_INCLUDES_EXTRAS false for v47 (preserving current behavior), and change it to true for v48. There are risks either way. Changing a lot of paths to extra paths (without code-fallback values) has increased the risk that some unknown bugs exist due to skipping extra paths. Including extra paths in iteration could also introduce bugs. Either way, the bugs in question aren't covered yet by unit tests, and they could be tools that are run infrequently.

macchiati:
Let's not change behavior from v46.1 this late in the game.

btangmu:
> Let's not change behavior from v46.1 this late in the game.

Understood. What do you think about making DEFAULT_ITERATION_INCLUDES_EXTRAS false, and merging it that way?

macchiati:
Only if we verify that there are no differences in resolved values.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
```diff
@@ -1497,22 +1544,38 @@ public Iterator<String> iterator(String prefix, Comparator<String> comparator) {
...
+    private static class FullIterable implements Iterable<String>, SimpleIterator<String> {
         private final CLDRFile file;
-        private final Iterator<String> fileIterator;
+        private final Iterator<String> fileIterator; // dataSource only
         private Iterator<String> extraPaths;
 
         FullIterable(CLDRFile file) {
             this.file = file;
-            this.fileIterator = file.iterator();
+            this.fileIterator = file.dataSource.iterator(); // file.iteratorWithoutExtras();
```
srl295, line 1572:
```suggestion
            this.fileIterator = file.iteratorWithoutExtras();
```

would it be better to use the function for documentation?

btangmu: Good idea. And I'll rename `fileIterator` to `iteratorWithoutExtras` to make it clear and consistent.

btangmu: I've done that, in the 2nd commit

btangmu:
> Only if we verify that there are no differences in resolved values.

Yes, that's wise. Locally, I will generate VXML, and also run GenerateProductionData and confirm no differences...

btangmu:
Locally, with these changes to CLDRFile (including DEFAULT_ITERATION_INCLUDES_EXTRAS false), I generated VXML by logging in as Admin and choosing Generate VXML from the main menu. Then I copied the VXML to common/main and ran GenerateProductionData. Then I repeated those steps on the main branch, with a different destination directory. Comparing the destination directories, both for VXML and for production, resulted in zero differences.

macchiati:
Thanks!

macchiati (approved):

PR #4372, opened 2025-02-12
CLDR-17925 da short inch-ofhg, use U+2033 not U+201C

pedberg-icu:
CLDR-17925

da, short inch-ofhg should use U+2033 (double prime) as already used in narrow inch-ofhg, not U+201C (left double quotation)

ALLOW_MANY_COMMITS=true


cgaarden (approved):
Thank you Peter

PR #4374, opened 2025-02-13
CLDR-15693 Improve CurrencyDateInfo ordering, add test for XML file ordering, fix data

pedberg-icu:
CLDR-15693

Note that per TR35 [Supplemental Currency Data](https://www.unicode.org/reports/tr35/tr35-numbers.html#Supplemental_Currency_Data), "The [xml file] *ordering* of the elements in the [currency] list tells us which was the primary currency during any period in time."

- Update TR35 to note that currency elements with tender="false" should be last in list for currencyData/region (they are never primary at any point in time).
- Order SupplementalDataInfo.CurrencyDataInfo first by isLegalTender, then by _newest_ date range first (not _older_ date range). Note that while this ordering is closer to the file ordering, it will not be identical (since we need to sort these in a unique way), so SupplementalDataInfo has never preserved info about the primary currency during any time period in the event that two currencies are legal tender during overlapping time periods.
- Add test for currencyData/region element ordering (regions must be ordered by region code, and within a region, currency elements with `tender = "false"` should be last).
- Fix supplementalData.xml currencyData to pass test (use reorder one region)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -1145,11 +1145,11 @@ The printed version of ISO-4217:2001
             <currency iso4217="DEM" from="1999-09-01" to="2002-03-09"/>
             <currency iso4217="YUM" from="1994-01-24" to="1999-09-30"/><!-- Yugoslavian history. YU was renamed to CS. -->
         </region>
-        <region iso3166="YE">
-            <currency iso4217="YER" from="1990-05-22"/><!-- You might be able to say 1964-05-08 is the starting date -->
-        </region>
         <region iso3166="YD">
             <currency iso4217="YDD" from="1965-04-01" to="1996-01-01"/>
+        </region>
+        <region iso3166="YE">
+            <currency iso4217="YER" from="1990-05-22"/><!-- You might be able to say 1964-05-08 is the starting date -->
         </region>
         <region iso3166="YT">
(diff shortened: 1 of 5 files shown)


macchiati (approved):

pedberg-icu:
@macchiati @srl295 I seem to have lost the ability to merge PRs ("You’re not [authorized](https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches) to merge this pull request"). Can one of you merge this for now, then we can figure out what is going on? Thanks.

srl295:
@pedberg-icu I sure thought I had re-added you but I will double check a little bit later tonight

pedberg-icu:
> @pedberg-icu I sure thought I had re-added you but I will double check a little bit later tonight

@srl295 Should I have rebased onto current main? Did the deleted cldr-tc group affect the actual cldr repo such that I might have branched from the bad version?

srl295:
> > @pedberg-icu I sure thought I had re-added you but I will double check a little bit later tonight
> 
> @srl295 Should I have rebased onto current main?

no impact on this

> Did the deleted cldr-tc group affect the actual cldr repo such that I might have branched from the bad version?

no, just affected your perms

> .... Does not look like it; and I do seem to be in the current version of https://github.com/orgs/unicode-org/teams/cldr-tc

yes

PR #4375, opened 2025-02-13
CLDR-17484 Update dayPeriods.xml

macchiati:
One small change from the ticket to minimize changes: keeping the boundary between evening and night at 9:00pm

CLDR-17484

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/dayPeriods.xml
+++ b/common/supplemental/dayPeriods.xml
@@ -7,6 +7,7 @@
  -->
 <supplementalData>
 	<version number="$Revision$"/>
+	<!-- Remember to make parallel changes both dayPeriodRuleSets! -->
 	<dayPeriodRuleSet>
 		<dayPeriodRules locales="root">
 			<dayPeriodRule type="am" from="00:00" before="12:00"/>	<!-- AM -->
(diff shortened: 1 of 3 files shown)


richgillam (dismissed):
Thank you!

richgillam (dismissed):

macchiati:
Still fixing tests; don't review until they complete successfully. Should have marked as draft until the tests complete.

macchiati:
Finally fixed the test failure, so we can merge this. (We already agreed that this was very unlikely to cause ICU problem [Although there might be unit tests that need adjustment, like I had to for CLDR]).

richgillam (approved):
I'm not familiar with the CLDR tests, but the changes there look plausible, and the CLDR data changes look good.

macchiati:
@markusicu, while you were out I checked with Rich and Dragan about this one, and said it was ok. But before merging I want to check with you.

markusicu:
thanks for checking; these changes are ok with me

pedberg-icu:
Also discussed in ICU TC, and agreed this is OK for CLDR 77. However ICU will likely create release candidate before this is integrated.

PR #4379, opened 2025-02-17
CLDR-13688 Correct the first day of week for Iceland from Monday to Sunday

gardar:
CLDR-13688

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -4897,7 +4897,7 @@ XXX Code for transations where no currency is involved
 			FI FJ FO FR
 			GB GE GF GP GR
 			HR HU
-			IE IS IT
+			IE IT
 			KG KZ
 			LB LI LK LT LU LV
 			MC MD ME MK MN MQ MY
@@ -4920,7 +4920,7 @@ XXX Code for transations where no currency is involved
 			ET
 			GT GU
 			HK HN
-			ID IL IN
+			ID IL IN IS
 			JM JP
 			KE KH KR
 			LA


srl295 (approved):
Change itself does what it says on the tin

srl295:
Just to note, if you go in the top right corner of the Jira ticket, there
is a button that will log you in… You can login with Gmail or GitHub I
think

gardar:
> Just to note, if you go in the top right corner of the Jira ticket, there is a button that will log you in… You can login with Gmail or GitHub I think

I did sign up to the Jira tracker but at first I was unable to comment even though I had created an account, but soon after I posted here I refreshed jira and was able to comment. Perhaps there was some manual approval or a workflow that needed to pass before I got the privilege to comment.

Since I was able to comment on the Jira issue I actually moved my post to there and deleted my previous comment here which you are responding to, I see you are replying via email so you still see my old reply.

srl295:
> > Just to note, if you go in the top right corner of the Jira ticket, there is a button that will log you in… You can login with Gmail or GitHub I think
> 
> I did sign up to the Jira tracker but at first I was unable to comment even though I had created an account, but soon after I posted here I refreshed jira and was able to comment. Perhaps there was some manual approval or a workflow that needed to pass before I got the privilege to comment.
> 
> Since I was able to comment on the Jira issue I actually moved my post to there and deleted my previous comment here which you are responding to, I see you are replying via email so you still see my old reply.

Ah ok.  That explains why I couldn't find the comment, glad you were able to get in OK.  No approval, probably just some database didn't update yet.

PR #4383, opened 2025-02-18
CLDR-13589 Add unit test that locale time cycle matches region prefs

btangmu:
-Similar to icu4j DateTimeGeneratorTest.testJjMapping

-Use current CLDR data, not ICU data

CLDR-13589

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateFromCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateFromCLDR.java
@@ -30,7 +30,7 @@ import org.unicode.cldr.util.XPathParts;
  *
  * @author markdavis
  */
-class FlexibleDateFromCLDR {
+public class FlexibleDateFromCLDR {
     DateTimePatternGenerator gen = DateTimePatternGenerator.getEmptyInstance();
     private transient ICUServiceBuilder icuServiceBuilder = new ICUServiceBuilder();
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateFromCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateFromCLDR.java
@@ -30,7 +30,7 @@ import org.unicode.cldr.util.XPathParts;
  *
  * @author markdavis
  */
-public class FlexibleDateFromCLDR {
+class FlexibleDateFromCLDR {
     DateTimePatternGenerator gen = DateTimePatternGenerator.getEmptyInstance();
     private transient ICUServiceBuilder icuServiceBuilder = new ICUServiceBuilder();
 
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
@@ -1,0 +1,68 @@
+package org.unicode.cldr.unittest;
+
+import com.ibm.icu.text.DateFormat;
+import com.ibm.icu.text.DateTimePatternGenerator;
+import com.ibm.icu.text.SimpleDateFormat;
+import com.ibm.icu.util.ULocale;
+import java.util.logging.Logger;
+import org.junit.jupiter.api.Test;
+
+public class TestTimeCycle {
+    private static final Logger logger = Logger.getLogger(TestTimeCycle.class.getName());
+
+    /**
+     * Test that a locale's standard time format matches the region's preferred time cycle
+     *
+     * <p>This is essentially the same as icu4j DateTimeGeneratorTest.testJjMapping
+     */
+    @Test
+    public void testJjMapping() {
+        final String jSkeleton = "j";
+        final char[] timeCycleChars = {'H', 'h', 'K', 'k'};
+        // First test that j maps correctly by region in a locale for which we do not have data.
+        {
+            String testLocaleID = "de_US"; // short patterns from fallback locale "de" have "HH"
+            ULocale testLocale = new ULocale(testLocaleID);
+            DateTimePatternGenerator dtpg = DateTimePatternGenerator.getInstance(testLocale);
+            String jPattern = dtpg.getBestPattern(jSkeleton);
+            String jPatSkeleton = dtpg.getSkeleton(jPattern);
+            if (jPatSkeleton.indexOf('h')
+                    < 0) { // expect US preferred cycle 'h', not H or other cycle
+                logger.severe(
+                        "DateTimePatternGenerator getBestPattern locale "
+                                + testLocaleID
+                                + ", pattern j did not use 'h'");
+            }
+        }
+
+        // Next test that in all available Locales, the actual short time pattern uses the same
+        // cycle as produced by 'j'
+        ULocale[] locales = DateFormat.getAvailableULocales();
+        for (ULocale locale : locales) {
+            String localeID = locale.getName();
+            DateTimePatternGenerator dtpg = DateTimePatternGenerator.getInstance(locale);
+            DateFormat dfmt = DateFormat.getTimeInstance(DateFormat.SHORT, locale);
+            String shortPattern = ((SimpleDateFormat) dfmt).toPattern();
+            String jPattern = dtpg.getBestPattern(jSkeleton);
+            // Now check that shortPattern and jPattern use the same hour cycle
+            String jPatSkeleton = dtpg.getSkeleton(jPattern);
+            String shortPatSkeleton = dtpg.getSkeleton(shortPattern);
+            for (char timeCycleChar : timeCycleChars) {
+                if (jPatSkeleton.indexOf(timeCycleChar) >= 0) {
+                    if (shortPatSkeleton.indexOf(timeCycleChar) < 0) {
+                        String dfmtCalType = dfmt.getCalendar().getType();
+                        logger.severe(
+                                "locale "
+                                        + localeID
+                                        + ", expected j resolved char "
+                                        + timeCycleChar
+                                        + " to occur in short time pattern "
+                                        + shortPattern
+                                        + " for "
+                                        + dfmtCalType);
+                    }
+                }
+            }
+        }
+    }
+}
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
@@ -7,11 +7,15 @@ import com.ibm.icu.text.DateTimePatternGenerator;
 import com.ibm.icu.text.SimpleDateFormat;
 import com.ibm.icu.util.ULocale;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.logging.Logger;
 import org.junit.jupiter.api.Test;
 import org.unicode.cldr.icu.LDMLConstants;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.Factory;
 import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.PreferredAndAllowedHour;
@@ -30,9 +34,14 @@ public class TestTimeCycle {
 
     private final char[] timeCycleChars = {'H', 'h', 'K', 'k'};
 
-    private final boolean USE_ICU_DATA = false;
+    private final boolean USE_ICU_DATA = false; // temporary, debugging
 
     private final String calendarType = LDMLConstants.GREGORIAN; /* or LDMLConstants.GENERIC? */
+
+    private final String shortTimePath =
+            "//ldml/dates/calendars/calendar[@type=\""
+                    + calendarType
+                    + "\"]/timeFormats/timeFormatLength[@type=\"short\"]/timeFormat[@type=\"standard\"]/pattern[@type=\"standard\"]";
 
     /**
      * Test that in all available locales, the standard time format matches the region's preferred
@@ -48,13 +57,14 @@ public class TestTimeCycle {
         SupplementalDataInfo sdi = testInfo.getSupplementalDataInfo();
         Map<String /* region */, PreferredAndAllowedHour> timeData = sdi.getTimeData();
         Map<String, String> likelySubtags = sdi.getLikelySubtags();
-        int failureNumber = 0;
+        Set<String> skippedLocales = new TreeSet<>(), worldFallback = new TreeSet<>();
+        Map<CLDRLocale, String> failedLocales = new TreeMap<>();
         for (String localeID : cldrFactory.getAvailable()) {
+            CLDRLocale loc = CLDRLocale.getInstance(localeID);
             CLDRFile cldrFile = cldrFactory.make(localeID, true);
             ULocale uloc = new ULocale(localeID);
             DateFormat dfmt = DateFormat.getTimeInstance(DateFormat.SHORT, uloc);
-            String jPattern;
-            String shortPattern;
+            String jPattern, shortPattern;
             DateTimePatternGenerator dtpg;
             if (USE_ICU_DATA) {
                 dtpg = DateTimePatternGenerator.getInstance(uloc);
@@ -62,63 +72,73 @@ public class TestTimeCycle {
                 shortPattern = ((SimpleDateFormat) dfmt).toPattern(); // e.g., "h:mm a" or "HH:mm"
             } else {
                 dtpg = DateTimePatternGenerator.getEmptyInstance();
-                jPattern = getCldrJPatternForLocale(timeData, localeID, likelySubtags);
+                jPattern = getJPattern(timeData, localeID, likelySubtags, worldFallback);
                 if (jPattern == null) {
-                    failOrLog("Got null jPattern, locale = " + localeID);
+                    failedLocales.put(loc, "Null jPattern");
                     continue;
                 }
-                String shortTimePath =
-                        "//ldml/dates/calendars/calendar[@type=\""
-                                + calendarType
-                                + "\"]/timeFormats/timeFormatLength[@type=\"short\"]/timeFormat[@type=\"standard\"]/pattern[@type=\"standard\"]";
                 shortPattern = cldrFile.getWinningValue(shortTimePath);
                 if (shortPattern == null) {
-                    failOrLog("Got null shortTimePath, locale = " + localeID);
+                    failedLocales.put(loc, "Null shortTimePath");
+                    continue;
+                }
+                if (CLDRLocale.ROOT.equals(loc.getParent()) && !cldrFile.isHere(shortTimePath)) {
+                    skippedLocales.add(localeID);
                     continue;
                 }
             }
-            // convert to skeletons to make it easier to check for hour character (eliminates
+            // Convert to skeletons to make it easier to check for hour character (eliminates
             // literals, repeats, etc.)
             String shortPatSkeleton = dtpg.getBaseSkeleton(shortPattern); // e.g., "ahm" or "Hm"
             String jPatSkeleton = dtpg.getBaseSkeleton(jPattern); // e.g., "ah" or "H"
-            System.out.println(
-                    "testJjMapping "
-                            + localeID
-                            + "\t["
-                            + jPattern
-                            + "] ["
-                            + jPatSkeleton
-                            + "] ["
-                            + shortPattern
-                            + "] ["
-                            + shortPatSkeleton
-                            + "]");
             for (char timeCycleChar : timeCycleChars) {
                 boolean has1 = jPatSkeleton.indexOf(timeCycleChar) >= 0;
                 boolean has2 = shortPatSkeleton.indexOf(timeCycleChar) >= 0;
                 if (has1 && !has2) {
                     String calType = USE_ICU_DATA ? dfmt.getCalendar().getType() : calendarType;
-                    failOrLog(
-                            ++failureNumber
-                                    + " locale "
-                                    + localeID
-                                    + ", calendar "
-                                    + calType
-                                    + ", expected "
+                    failedLocales.put(
+                            loc,
+                            calType
+                                    + " expected "
                                     + timeCycleChar
-                                    + " to occur in both patterns "
+                                    + " in both "
                                     + jPattern
                                     + " and "
                                     + shortPattern);
                 }
             }
         }
+        if (!worldFallback.isEmpty()) {
+            logger.info(
+                    "TestTimeCycle skipped "
+                            + worldFallback.size()
+                            + " world-fallback locales: "
+                            + worldFallback);
+        }
+        if (!skippedLocales.isEmpty()) {
+            logger.info(
+                    "TestTimeCycle skipped "
+                            + skippedLocales.size()
+                            + " top-level locales without explicit timeFormatLength: "
+                            + skippedLocales);
+        }
+        if (!failedLocales.isEmpty()) {
+            for (Map.Entry<CLDRLocale, String> entry : failedLocales.entrySet()) {
+                logger.severe(entry.getKey() + " " + entry.getValue());
+            }
+            fail(
+                    "TestTimeCycle failed "
+                            + failedLocales.size()
+                            + " locales: "
+                            + failedLocales.keySet());
+        }
     }
 
-    private String getCldrJPatternForLocale(
+    private String getJPattern(
             Map<String, PreferredAndAllowedHour> timeData,
             String localeID,
-            Map<String, String> likelySubtags) {
+            Map<String, String> likelySubtags,
+            Set<String> worldFallback) {
 
         PreferredAndAllowedHour prefAndAllowedHr = timeData.get(localeID);
         if (prefAndAllowedHr == null) {
@@ -128,23 +148,11 @@ public class TestTimeCycle {
                 String loc2 = likelySubtags.get(localeID);
                 if (loc2 != null && !loc2.isEmpty()) {
                     region = lp.set(loc2).getRegion();
-                    System.out.println(
-                            "getCldrJPatternForLocale "
-                                    + localeID
-                                    + " loc2 = "
-                                    + loc2
-                                    + " region = "
-                                    + region);
                 }
             }
             prefAndAllowedHr = timeData.get(region);
             if (prefAndAllowedHr == null) {
-                System.out.println(
-                        "getCldrJPatternForLocale "
-                                + localeID
-                                + " get (region "
-                                + region
-                                + ") null, falling back to 001");
+                worldFallback.add(localeID);
                 prefAndAllowedHr = timeData.get("001" /* world */);
                 if (prefAndAllowedHr == null) {
                     return null;
@@ -169,16 +177,7 @@ public class TestTimeCycle {
             case hB:
                 return "h";
             default:
-                failOrLog("Cannot get pattern for " + hourStyle);
                 return null;
-        }
-    }
-
-    private void failOrLog(String message) {
-        if (false) {
-            fail(message);
-        } else {
-            logger.severe(message);
         }
     }
 }
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
@@ -11,6 +11,7 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import org.junit.jupiter.api.Test;
 import org.unicode.cldr.icu.LDMLConstants;
 import org.unicode.cldr.util.CLDRConfig;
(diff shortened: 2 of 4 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,68 @@
...
+                logger.severe(
+                        "DateTimePatternGenerator getBestPattern locale "
+                                + testLocaleID
+                                + ", pattern j did not use 'h'");
+            }
+        }
+
+        // Next test that in all available Locales, the actual short time pattern uses the same
+        // cycle as produced by 'j'
+        ULocale[] locales = DateFormat.getAvailableULocales();
```
pedberg-icu, line 40: Here we want to use all the locales in CLDR, not just the ones in ICU. Typically we do this as follows, look at other unittest files:
```
    static CLDRConfig testInfo = CLDRConfig.getInstance();
    ...

        Factory cldrFactory = testInfo.getCldrFactory();
        for (String locale : cldrFactory.getAvailable()) {
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,68 @@
...
+        for (ULocale locale : locales) {
+            String localeID = locale.getName();
+            DateTimePatternGenerator dtpg = DateTimePatternGenerator.getInstance(locale);
```
pedberg-icu, line 43: Here we want to populate the DateTimePatternGenerator with current CLDR data, not ICU data. I will try to find an example but it would use ICUServiceBuilder.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,68 @@
...
+            DateFormat dfmt = DateFormat.getTimeInstance(DateFormat.SHORT, locale);
```
pedberg-icu, line 44: And again here we want to get the current CLDR data, using a CLDRFile method, instead of the ICU data.

pedberg-icu (changes requested):
Need to revise to use CLDR data instead of ICU data

btangmu:
> Need to revise to use CLDR data instead of ICU data

I made a 2nd commit based on your comments, but I doubt it's right. There's no usage of `DateFormat.SHORT`. I don't know if `LDMLConstants.GREGORIAN` is right. `jPattern` is used twice...

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,88 @@
...
+        CLDRConfig testInfo = CLDRConfig.getInstance();
+        Factory cldrFactory = testInfo.getCldrFactory();
+        for (String localeID : cldrFactory.getAvailable()) {
+            CLDRLocale loc = CLDRLocale.getInstance(localeID);
+            ICUServiceBuilder icuServiceBuilder = ICUServiceBuilder.forLocale(loc);
+
+            // Compare ICU data version:
+            // DateTimePatternGenerator dtpg = DateTimePatternGenerator.getInstance(uloc);
+
+            DateTimePatternGenerator dtpg = DateTimePatternGenerator.getEmptyInstance();
```
btangmu, line 55: `getEmptyInstance`? Is it OK that `dtpg` itself isn't associated with the locale? Instead it's `icuServiceBuilder` that's locale-specific.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,88 @@
...
+
+            String jPattern = dtpg.getBestPattern(jSkeleton);
+
+            // Compare ICU data version:
+            // DateFormat dfmt = DateFormat.getTimeInstance(DateFormat.SHORT, uloc);
+
+            SimpleDateFormat dfmt =
+                    icuServiceBuilder.getDateFormat(LDMLConstants.GREGORIAN, jPattern);
```
btangmu, line 63: `DateFormat.getTimeInstance` specifies "time" while `icuServiceBuilder.getDateFormat` doesn't. I didn't see any method like ``icuServiceBuilder.getTime...` 

I'm just guessing about `jPattern` here, versus `jPatSkeleton` or `shortPatSkeleton` or something completely different.

macchiati: Speaking generally here.

You have to be careful not to use "plain" ICU calls for formatting in tests or examples, since that will use ICU's (old) copy of CLDR data, not what is in the repository. So it won't test what needs to be tested.

If ICUServiceBuilder has API for what you want, that's great. Internally, it uses API code, but makes calls that load up CLDR data and then calls mechanisms that build the ICU classes using CLDR data.

There are some cases where ICUServiceBuilder doesn't have an API for what you want. In that case, do a search for the class you want to use on the ICU side within the CLDR code. For example, for DateTimePatternGenerator, there are a few places in the code that build it up, like     

CheckDates:
DateTimePatternGenerator dateTimePatternGenerator = DateTimePatternGenerator.getEmptyInstance();

(This ought to be more centralized and documented.) I have to run to a meeting, but will check later.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,88 @@
...
+     * Test that a locale's standard time format matches the region's preferred time cycle
+     *
+     * <p>This is similar to icu4j DateTimeGeneratorTest.testJjMapping. It uses current CLDR data,
+     * not ICU data.
+     */
+    @Test
+    public void testJjMapping() {
+        final String jSkeleton = "j";
+        final char[] timeCycleChars = {'H', 'h', 'K', 'k'};
+        // First test that j maps correctly by region in a locale for which we do not have data.
```
pedberg-icu, line 27: This first part is really just a test of ICU code (and using ICU data), so there is not much point to having it in CLDR as a unit test. I would just get rid of it.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
pedberg-icu: You will need to load up the DateTimePatternGenerator yourself with the CLDR data for the locale, it does not automatically do that. Take a look at tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateFromCLDR.java (in fact you may be able to use that class in this test)

btangmu:
I've added a 3rd commit, with comments and debugging, and changed this to a DRAFT PR.

Calling `FlexibleDateFromCLDR` does result in some localized values, but usually not the ones matching the original ICU test, maybe involving the lack of specifying anything like `DateFormat.SHORT`...

tools/cldr-code/src/main/java/org/unicode/cldr/test/FlexibleDateFromCLDR.java
```diff
@@ -30,7 +30,7 @@
  *
  * @author markdavis
  */
-class FlexibleDateFromCLDR {
+public class FlexibleDateFromCLDR {
```
btangmu, line 33: There's an old (?) comment above, "Temporary class while refactoring", which casts suspicion on making this public

pedberg-icu:
Hmm. It may be that using DateTimePatternGenerator to handle the 'j' character is not the right way to approach this. Basically what we are trying to do is make sure that the gregorian and/or preferred-calendar (if different) time formats in each locale - which may be inherited - use the preferred hour cycle for the region or region-language combination as specified by the supplemental `<timeData>`. I think we should probably do this without using ICU for locale data at all. I think we can do the following:
1. Get the mapping from region (or language_region) to preferred hour cycle as follows:
````
    CLDRConfig testInfo = CLDRConfig.getInstance();
    SupplementalDataInfo sdi = testInfo.getSupplementalDataInfo();
    Map<String, PreferredAndAllowedHour> timeData = sdi.getTimeData();
```
2. Get an empty DateTimePatternGenerator, we are only going to use this to parse time patterns (so no locale data)
```
    DateTimePatternGenerator dtpg = DateTimePatternGenerator.getEmptyInstance();
````
3. Then get the list of locales and loop:
````
    // CLDRConfig testInfo = CLDRConfig.getInstance(); // already done above
    Factory cldrFactory = testInfo.getCldrFactory();
    for (String localeID : cldrFactory.getAvailable()) {
```
4. For each locale, get the preferred hour:
````
        // first try the locale as is:
        PreferredAndAllowedHour prefAndAllowedHr = timeData.get(localeID);
        // if that does not work (returns null?), try the locale's region: if it has a region part, use that,
        // otherwise get the region from likelySubtags; you can use sdi.getLikelySubtags() to get a map
        String region = ...
        prefAndAllowedHr = timeData.get(region);
        // if that does not work, try "001" world
        prefAndAllowedHr = timeData.get("001");
        // then get the hour cycle from prefAndAllowedHr, have not yet figured out how to do that
```
5. Then for each locale, use CLDR file to get the short time format for Gregorian calendar:
        CLDRFile cldrFile = testInfo.getCLDRFile(locale, true);
        String gregoShortTimePath = "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/timeFormats/timeFormatLength[@type=\"short\"]/timeFormat/pattern";
        String shortTimeString = cldrFile.getStringValueWithBailey(gregoShortDatePath);
        // convert to skeleton to make it easier to check for hour character (eliminates literals etc.):
        String shortTimeSkeleton = dtpg.getBaseSkeleton(shortTimeString);
         // then see whether the shortTimeSkeleton uses the preferred hour cycle from step 4, error if it does not
6. As a bonus step, you can find the preferred calendar for the locale (based on data from SupplementalDataInfo for the region), and if it is not Gregorian then also check the shortTimeFormat for that...

btangmu:
@pedberg-icu I'm starting to have some success with that, thanks! The path you gave doesn't seem quite right. I'm using this instead, OK?

`//ldml/dates/calendars/calendar[@type="gregorian"]/timeFormats/timeFormatLength[@type="short"]/timeFormat[@type="standard"]/pattern[@type="standard"]`

btangmu:
The 4th commit uses `PreferredAndAllowedHour` and `//ldml/dates/calendars/calendar[@type="gregorian"]/timeFormats/timeFormatLength[@type="short"]/timeFormat[@type="standard"]/pattern[@type="standard"]`

It fails for locale "aa", since `timeData.get("aa")` returns null, and `new LocaleIDParser().set("aa").getRegion()` returns null.

`LocaleIDParser.setRegion` has only one caller, `FlexibleDateTime.DeprecatedCodeFixer.fixLocale`, which seemingly is never called, at least not for "aa".

Is there some other way to get a region/territory from a locale ID?

pedberg-icu:
> @pedberg-icu I'm starting to have some success with that, thanks! The path you gave doesn't seem quite right. I'm using this instead, OK?
> 
> `//ldml/dates/calendars/calendar[@type="gregorian"]/timeFormats/timeFormatLength[@type="short"]/timeFormat[@type="standard"]/pattern[@type="standard"]`

Yes you are right, I had forgotten about the (implied) type="standard"

pedberg-icu:
> Tom: It fails for locale "aa", since `timeData.get("aa")` returns null, and `new > Is there some other way to get a region/territory from a locale ID?
> Mark: You should be calling LikelySubtags to fill in the region if there isn't one.

Yes. Tom, see step 4, I had mentioned "you can use sdi.getLikelySubtags() to get a map". This returns a Map<String, String> that should mirror the content of supplemental/likelySubtags.xml, which will map e.g. "aa" to "aa_Latn_ET" and then you can use LocaleIDParser to get the region from that.

btangmu:
The 5th commit uses `LikelySubtags`. These 53 failures now occur, which is better than before:

 1 locale apc, calendar gregorian, expected h to occur in both patterns h and HH:mm
 2 locale apc_SY, calendar gregorian, expected h to occur in both patterns h and HH:mm
 3 locale arn, calendar gregorian, expected h to occur in both patterns h and HH:mm
 4 locale arn_CL, calendar gregorian, expected h to occur in both patterns h and HH:mm
 5 locale az_Arab_IQ, calendar gregorian, expected h to occur in both patterns h and HH:mm
 6 locale bal_Arab, calendar gregorian, expected H to occur in both patterns H and hh:mm a
 7 locale bal_Latn, calendar gregorian, expected H to occur in both patterns H and hh:mm a
 8 locale bgn, calendar gregorian, expected h to occur in both patterns h and HH:mm
 9 locale bgn_AE, calendar gregorian, expected h to occur in both patterns h and HH:mm
10 locale bgn_OM, calendar gregorian, expected h to occur in both patterns h and HH:mm
11 locale bgn_PK, calendar gregorian, expected h to occur in both patterns h and HH:mm
12 locale cho, calendar gregorian, expected h to occur in both patterns h and HH:mm
13 locale cho_US, calendar gregorian, expected h to occur in both patterns h and HH:mm
14 locale cop, calendar gregorian, expected h to occur in both patterns h and HH:mm
15 locale cop_EG, calendar gregorian, expected h to occur in both patterns h and HH:mm
16 locale el_POLYTON, calendar gregorian, expected H to occur in both patterns H and h:mm a
17 locale en_Dsrt_US, calendar gregorian, expected h to occur in both patterns h and HH:mm
18 locale gn, calendar gregorian, expected h to occur in both patterns h and HH:mm
19 locale gn_PY, calendar gregorian, expected h to occur in both patterns h and HH:mm
20 locale ha_Arab_SD, calendar gregorian, expected h to occur in both patterns h and HH:mm
21 locale hi_Latn, calendar gregorian, expected H to occur in both patterns H and h:mm a
22 locale hnj, calendar gregorian, expected h to occur in both patterns h and HH:mm
23 locale hnj_Hmnp_US, calendar gregorian, expected h to occur in both patterns h and HH:mm
24 locale iu_Latn_CA, calendar gregorian, expected h to occur in both patterns h and HH:mm
25 locale kok_Deva, calendar gregorian, expected H to occur in both patterns H and h:mm a
26 locale kok_Latn, calendar gregorian, expected H to occur in both patterns H and a h:mm
27 locale ks_Arab, calendar gregorian, expected H to occur in both patterns H and h:mm a
28 locale ks_Deva, calendar gregorian, expected H to occur in both patterns H and a h:mm
29 locale kxv_Deva, calendar gregorian, expected H to occur in both patterns H and h:mm a
30 locale kxv_Latn, calendar gregorian, expected H to occur in both patterns H and h:mm a
31 locale kxv_Orya, calendar gregorian, expected H to occur in both patterns H and h:mm a
32 locale kxv_Telu, calendar gregorian, expected H to occur in both patterns H and h:mm a
33 locale mni_Beng, calendar gregorian, expected H to occur in both patterns H and h:mm a
34 locale mni_Mtei, calendar gregorian, expected H to occur in both patterns H and h.mm. a
35 locale ms_Arab, calendar gregorian, expected H to occur in both patterns H and h:mm a
36 locale nv, calendar gregorian, expected h to occur in both patterns h and HH:mm
37 locale nv_US, calendar gregorian, expected h to occur in both patterns h and HH:mm
38 locale pa_Guru, calendar gregorian, expected H to occur in both patterns H and h:mm a
39 locale pis, calendar gregorian, expected h to occur in both patterns h and HH:mm
40 locale pis_SB, calendar gregorian, expected h to occur in both patterns h and HH:mm
41 locale quc, calendar gregorian, expected h to occur in both patterns h and HH:mm
42 locale quc_GT, calendar gregorian, expected h to occur in both patterns h and HH:mm
43 locale sat_Deva, calendar gregorian, expected H to occur in both patterns H and h:mm a
44 locale sat_Olck, calendar gregorian, expected H to occur in both patterns H and h:mm a
45 locale sd_Arab, calendar gregorian, expected H to occur in both patterns H and h:mm a
46 locale sdh_IQ, calendar gregorian, expected h to occur in both patterns h and HH:mm
47 locale skr, calendar gregorian, expected h to occur in both patterns h and HH:mm
48 locale skr_PK, calendar gregorian, expected h to occur in both patterns h and HH:mm
49 locale vai_Latn, calendar gregorian, expected H to occur in both patterns H and h:mm a
50 locale vai_Vaii, calendar gregorian, expected H to occur in both patterns H and h:mm a
51 locale wbp, calendar gregorian, expected h to occur in both patterns h and HH:mm
52 locale wbp_AU, calendar gregorian, expected h to occur in both patterns h and HH:mm
53 locale yue_Hant, calendar gregorian, expected H to occur in both patterns H and ah:mm

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,184 @@
...
+                            + jPatSkeleton
+                            + "] ["
+                            + shortPattern
+                            + "] ["
+                            + shortPatSkeleton
+                            + "]");
+            for (char timeCycleChar : timeCycleChars) {
+                boolean has1 = jPatSkeleton.indexOf(timeCycleChar) >= 0;
+                boolean has2 = shortPatSkeleton.indexOf(timeCycleChar) >= 0;
+                if (has1 && !has2) {
```
btangmu, line 99: this (like the original ICU code) is unsymmetrical and leaves out checking `if (has2 && !has1)` -- however, I tried `(has1 != has2)` and just got exactly twice the number of failures, for example:

105 locale yue_Hant, calendar gregorian, expected H to occur in both patterns H and ah:mm
106 locale yue_Hant, calendar gregorian, expected h to occur in both patterns H and ah:mm

I guess `(has2 && !has1)` could be more significant if `jPatSkeleton` had none of h, H, k, K, but at least currently that's not the case

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+            prefAndAllowedHr = timeData.get(region);
+            if (prefAndAllowedHr == null) {
+                System.out.println(
+                        "getCldrJPatternForLocale "
+                                + localeID
+                                + " get (region "
+                                + region
+                                + ") null, falling back to 001");
+                prefAndAllowedHr = timeData.get("001" /* world */);
```
btangmu, line 148: fallback to 001 happens for az_Cyrl, az_Latn, bal_Arab, bal_Latn, be_TARASK, ..., altogether 53 locales. Strangely not the same as the 53 failing locales, though there is some overlap, such as for el_POLYTON

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTimeCycle.java
macchiati: Just taking one example, apc.xml

It has no timeFormatLength values, so it will just inherit from root.

Lines like the following also just inherit from root.

<dateFormatItem id="hm" draft="unconfirmed">↑↑↑</dateFormatItem>

root has (correctly) HH.

But its region is:

<likelySubtag from="apc" to="apc_Arab_SY"/>		

which matches

<hours preferred="h" allowed="h hB hb H" regions="AE BH DZ EG EH HK IQ JO KW LB LY MO MR OM PH PS QA SA SD SY TN YE ar_001"/>

and results in 'h'.

So we should filter out locales that are top-level locals with no explicit value for timeFormatLength.

(Might also have to refine this further).

macchiati:
Fallback to 001 is expected, and not a problem. 53 is just a coincidence,
tho a pretty unlikely one!

pedberg-icu:
Mark wrote: "So we should filter out locales that are top-level locals with no explicit value for timeFormatLength."

I disagree, finding those is part of the point of this (otherwise they will cause errors in ICU). If a top level locale has a default region that prefers 'h' and has no standard time formats, we need to add them. That is part the criteria for Basic I believe.

macchiati:
It is. But we don't want a *unit test* to fail. It will just be at core
until correct data is supplied.

btangmu:
The 6th commit skips top-level locals with no explicit `timeFormatLength` value.

That brings the failures down to 41 locales: [apc_SY, arn_CL, az_Arab_IQ, bal_Arab, bal_Latn, bgn_AE, bgn_OM, bgn_PK, cho_US, cop_EG, el_POLYTON, en_Dsrt_US, gn_PY, ha_Arab_SD, hi_Latn, hnj_Hmnp_US, iu_Latn_CA, kok_Deva, kok_Latn, ks_Arab, ks_Deva, kxv_Deva, kxv_Latn, kxv_Orya, kxv_Telu, mni_Beng, mni_Mtei, ms_Arab, nv_US, pa_Guru, pis_SB, quc_GT, sat_Deva, sat_Olck, sd_Arab, sdh_IQ, skr_PK, vai_Latn, vai_Vaii, wbp_AU, yue_Hant]

There are 53 world-fallback locales: [az_Cyrl, az_Latn, bal_Arab, bal_Latn, be_TARASK, bm_Nkoo, bs_Cyrl, bs_Latn, el_POLYTON, en_150, en_Dsrt, en_TK, en_TV, ff_Latn, fr_HT, ha_Arab, hi_Latn, hnj_Hmnp, ht, ht_HT, iu_Latn, kaa_Cyrl, kaa_Latn, kk_Cyrl, kok_Deva, kok_Latn, ks_Arab, ks_Deva, kxv_Deva, kxv_Latn, kxv_Orya, kxv_Telu, mni_Beng, mni_Mtei, ms_Arab, pa_Guru, rhg_Rohg, root, sat_Deva, sat_Olck, sd_Arab, shi_Latn, shi_Tfng, sr_Cyrl, sr_Latn, su_Latn, uz_Cyrl, uz_Latn, vai_Latn, vai_Vaii, yue_Hant, zh_Hans, zh_Latn]

There are skipped 54 top-level locales without explicit timeFormatLength: [ann, apc, arn, az_Arab, ba, bgn, blt, bss, ce, cho, cop, en_Dsrt, en_Shaw, gn, ha_Arab, hnj, io, iu_Latn, jbo, kaa, kaa_Latn, ken, lmo, ltg, mdf, mhn, mn_Mong, myv, mzn, nnh, nqo, nr, nv, pis, quc, sdh, shi, shi_Latn, shn, skr, sma, smj, sms, ss, tok, tyv, tzm, ve, vmw, wa, wae, wbp, za, zh_Latn]

Note: I just noticed, I mistakenly wrote "...skipped world-fallback locales ..." but they're not skipped (at least not just due to getting world-fallback), they're just logged for information. I'll fix that in a subsequent commits.

btangmu:
BTW the intersection of failed and worldFallback has 22 locales: [kok_Latn, kxv_Latn, kxv_Deva, vai_Latn, pa_Guru, kxv_Orya, kok_Deva, mni_Beng, sat_Olck, yue_Hant, sd_Arab, el_POLYTON, ks_Deva, bal_Arab, ms_Arab, sat_Deva, kxv_Telu, bal_Latn, ks_Arab, mni_Mtei, vai_Vaii, hi_Latn]

That's just over half of the failed locales.

pedberg-icu:
@btangmu, some comments...

> One reason that the original test passes when using ICU data, is that it only includes locales in `DateFormat.getAvailableULocales`. However, even when looping through the locales in `cldrFactory.getAvailable()` the test passes using ICU data,...

We should not be testing ICU data here...

> Maybe root is nothing to worry about and could just be skipped

Yes, root should be skipped

> On the other hand, locale "apc_SY" passes with ICU data

Again, no point in testing here with ICU data

> If default-content locales are skipped (I don't know if they should be skipped, I'm just experimenting)

Yes, we should skip default content locales

And some more comments not specifically related to the above:

1. I am confused by an error message like "[some locale] gregorian expected H in both HH and h:mm a". There are only two things we should be checking: (a) what is the expected hour cycle from supplemental data, and (b) what is the actual hour cycle in the short date format. If they do not math there is an error. So why is the error message showing two patterns, e.g. "HH" and "h:mm a" in the example above? I would only expect "h:mm a".

2.  Why is the error logging using logger.info and logger.severe (and what are they?)? Usually in unit tests we use login, error and/or assertTrue, etc.

3. If there are locales that have an issue that should be fixed (like top-level locales), we should file a ticket about them, and then skip the test for them using a call to logKnownIssue that lists that ticket (see other such usages). That way we can track which tests to remove when the problem is fixed.

btangmu:
> We should not be testing ICU data here

Understood. That's only for comparison. Locally I'm running with USE_ICU_DATA false or true, comparing results, to understand why the test passes if using ICU data and fails otherwise

btangmu:
> 1. why is the error message showing two patterns, e.g. "HH" and "h:mm a" in the example above? I would only expect "h:mm a".

Good point. That's based on the original ICU code. The "H" being searched for in "HH" and "h:mm a" is actually derived from the "HH" in a trivial way, that is, `"H" = getBaseSkeleton("HH")`. So we can omit the "HH" from the error message. I'll revise the code accordingly.

> 2. Why is the error logging using logger.info and logger.severe (and what are they?)? Usually in unit tests we use login, error and/or assertTrue, etc.

`java.util.logging.Logger` includes methods `info` and `severe` (see https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html ). Some of the output is informational for debugging, not intended to be in the final merged version of this test.

Also, I need some clarification when we should use JUnit and when we should use TestFmwk. Some CLDR tests use one, some use the other, and maybe some use both. For example, there are both `org.junit.jupiter.api.Assertions.assertEquals` and `org.unicode.cldr.icu.dev.test.TestFmwk.assertEquals`, which don't have the same order of parameters.

Currently this test uses JUnit and the only JUnit function it uses is `fail`.

> 3. If there are locales that have an issue that should be fixed (like top-level locales), we should file a ticket about them, and then skip the test for them using a call to logKnownIssue that lists that ticket (see other such usages). That way we can track which tests to remove when the problem is fixed.

I understand. I still don't always know which test failures are due to problems with the data, and which are only due to problems with the test itself. Once that's clear, before this ticket is closed, I will incorporate `logKnownIssue` into the test.

btangmu:
Looking at this failure "az_Arab_IQ gregorian expected h in HH:mm", Survey Tool shows:

https://st.unicode.org/cldr-apps/v#/az_Arab_IQ/Gregorian/235af2cdaa201be0

`path = //ldml/dates/calendars/calendar[@type="gregorian"]/timeFormats/timeFormatLength[@type="short"]/timeFormat[@type="standard"]/pattern[@type="standard"]`

Value: HH:mm
This item is inherited from the root locale.

It fails due to not containing the expected "h" derived from this new method:

```
    private String getRegionHourFormat(
            Map<String, PreferredAndAllowedHour> timeData,
            String localeID,
            Map<String, String> likelySubtags,
            Set<String> worldFallback) {
        PreferredAndAllowedHour prefAndAllowedHr = timeData.get(localeID);
        if (prefAndAllowedHr == null) {
            LocaleIDParser lp = new LocaleIDParser();
            String region = lp.set(localeID).getRegion();
            if (region == null || region.isEmpty()) {
                String loc2 = likelySubtags.get(localeID);
                if (loc2 != null && !loc2.isEmpty()) {
                    region = lp.set(loc2).getRegion();
                }
            }
            prefAndAllowedHr = timeData.get(region);
            if (prefAndAllowedHr == null) {
                worldFallback.add(localeID);
                prefAndAllowedHr = timeData.get(StandardCodes.NO_COUNTRY /* 001, world */);
                if (prefAndAllowedHr == null) {
                    return null;
                }
            }
        }
        return prefAndAllowedHr.preferred.base.name();
    }
```

btangmu:
Presumably the mapping from "IQ" (as in "az_Arab_IQ") to "h" is derived from this in common/supplemental/supplementalData.xml:
```
    <timeData>
                ...
		<hours preferred="h" allowed="h hB hb H" regions="AE BH DZ EG EH HK IQ JO KW LB LY MO MR OM PH PS QA SA SD SY TN YE ar_001"/>
```

btangmu:
It seems like at this point, we should:

1. File a new ticket, and log a known issue in the current ticket, for the 54 skipped top-level locales without explicit timeFormatLength: [ann, apc, arn, az_Arab, ba, bgn, blt, bss, ce, cho, cop, en_Dsrt, en_Shaw, gn, ha_Arab, hnj, io, iu_Latn, jbo, kaa, kaa_Latn, ken, lmo, ltg, mdf, mhn, mn_Mong, myv, mzn, nnh, nqo, nr, nv, pis, quc, sdh, shi, shi_Latn, shn, skr, sma, smj, sms, ss, tok, tyv, tzm, ve, vmw, wa, wae, wbp, za, zh_Latn]
2. File a new ticket (or the same new ticket?), and log a known issue in the current ticket, for the 17 remaining locales with errors: [az_Arab_IQ, bal_Latn, bgn_AE, bgn_OM, el_POLYTON, ha_Arab_SD, hi_Latn, kok_Latn, ks_Deva, kxv_Deva, kxv_Orya, kxv_Telu, mni_Mtei, ms_Arab, sat_Deva, sdh_IQ, vai_Latn]

Is that right?

pedberg-icu:
@btangmu You analysis of the situation with az_Arab_IQ is correct, it would need standard time formats added (using 'h'), typically we figure out the formats to use using the (possibly inherited) availableFormats entries for time skeletons using - in this case - 'h'. The we add them as draft="contributed" so they make it into ICU. But currently az_Arab* locales are not being imported into ICU because az_Arab is not yet at Basic.

btangmu:
Per infrastructure meeting, the plan is to make this a "check" to alert vetters in ST, rather than a unit test, so I'm closing this PR.

PR #4386, opened 2025-02-19
CLDR-15388 check annotations for empty values or keyword entry without name (tts) entry

pedberg-icu:
CLDR-15388

Add a consoleCheck test (so vetters will see it) in CheckAnnotations that:
- Annotation entries do not have empty values (`<annotation cp=".."></annotation>` or `<annotation cp=".."/>`).
- If there is a keyword entry for a given emoji/symbol, there should also be a tts (name) entry.

This found many missing tts errors, about which I have filed [CLDR-18329](https://unicode-org.atlassian.net/browse/CLDR-18329). In the meantime, I added a skip for these known error cases so that CI tests will pass (I did not use logKnownIssue, as far as I know we cannot do that in consoleCheck tests, not sure what would happen in Survey Tool)

ALLOW_MANY_COMMITS=true

[CLDR-18329]: https://unicode-org.atlassian.net/browse/CLDR-18329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
@@ -1,27 +1,73 @@
 package org.unicode.cldr.test;
 
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.PatternCache;
+import org.unicode.cldr.util.XPathParts;
 
 public class CheckAnnotations extends CheckCLDR {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
```diff
@@ -1,27 +1,73 @@
...
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.PatternCache;
+import org.unicode.cldr.util.XPathParts;
 
 public class CheckAnnotations extends CheckCLDR {
     private static final Pattern ANNOTATION_PATH = Pattern.compile("//ldml/annotations/.*");
 
+    // Skip tts test for these locale/cp combinations unti
+    // https://unicode-org.atlassian.net/browse/CLDR-18329 is fixed
```
pedberg-icu, line 21: I did not use logKnownIssue, as far as I know we cannot do that in consoleCheck tests, not sure what would happen in Survey Tool.

pedberg-icu:
Please feel free to merge when approved, I cannot merge

macchiati (approved):
Lgtm

PR #4388, opened 2025-02-20
CLDR-14088 Fix Serbian script in Russia

conradarcturus:
CLDR-14088

(we can merge this to v47 or v48, it's a minor ticket)

Serbians in Russia certainly use the Cyrillic script (the main script of Russian and of Serbian in Serbia), not the Latin script. However, for some reason for the past 19 years we've listed the script as sr_Latn in Russia. While that makes sense in Montenegro and Romania -- it really should be `sr_Cyrl` (or just `sr` since `Cyrl` is the default).

Note a number of un-related changes happened, but that's because I re-generate language & test data after each of these changes. A number of prior PRs forgot to do this so some values changed. If you look at the commit history -- the first commit are these generated changes, the second is the Serbian change.

```
```

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):

macchiati:
We are past beta-data. Before merging anything, we have to verify with the ICU team that they are comfortable with it, that it won't mess them up. So please check with @markusicu , @richgillam and @DraganBesevic _before_ merging.

(IMO this is very unlikely to cause problems in ICU, but we need to let ICU be the judge of that, since they bear the consequences. Markus/Rich/Dragan, can you weigh in on whether you need this retracted from 47?)

srl295:
i think it would be better to revert - 
- it says it's minor
- it also contains other unrelated changes

> that's because I re-generate language & test data after each of these changes

had we agreed to do that? I think it would be better to NOT regenerate these every time to avoid conflict with similar issues.

common/supplemental/supplementalData.xml
```diff
@@ -3149,7 +3149,7 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="fr" populationPercent="27" officialStatus="official" references="R1352"/>	<!--French-->
 			<languagePopulation type="ff" populationPercent="26"/>	<!--Fula-->
 			<languagePopulation type="man" populationPercent="23"/>	<!--Mandingo-->
-			<languagePopulation type="man_Nkoo" literacyPercent="50" populationPercent="23" references="R1021"/>	<!--Mandingo (N’Ko)-->
+			<languagePopulation type="man_Nkoo" literacyPercent="25" populationPercent="23" references="R1021"/>	<!--Mandingo (N’Ko)-->
```
srl295, line 3152: not called out in the PR…

srl295:
if we hadn't regenerated data that should be done as a separate step ideally.

markusicu:
from me: no strong opinion on these changes; I defer to CLDR unless @richgillam feels strongly

pedberg-icu:
Discussed in ICU TC, agreed this is OK for CLDR 77. However ICU release candidate will likely be created before this is integrated (so it will go in after than, before final release)

conradarcturus:
Sorry about jumping the gun. Glad it's ok.

macchiati:
No problem

srl295:
let's keep talking about the generated data, though…

PR #4389, opened 2025-02-20
CLDR-18295 Add linkify to archive.js

aphillips:
- Enables a spec to use MessageFormat's linkification

Specs need to include an element with the id 'linkify' to activate

CLDR-18295

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/scripts/tr-archive/archive.js
+++ b/tools/scripts/tr-archive/archive.js
@@ -4,7 +4,7 @@ const jsdom = require("jsdom");
 const { JSDOM } = jsdom;
 const path = require("path");
 const markedAlert = require("marked-alert");
-
+const matter = require("gray-matter");
 // Setup some options for our markdown renderer
 marked.setOptions({
   renderer: new marked.Renderer(),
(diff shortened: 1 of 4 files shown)


aphillips:
@srl295 Have a look? I tested this with the v47 pourover (plus one change to add an id of 'linkify' in the .md, e.g. `<span id='linkify'></span>`)

tools/scripts/tr-archive/archive.js
```diff
@@ -246,3 +251,68 @@ fixall().then(
...
+    console.log("Duplicate Terms: " + duplicateCount);
+  }
+  return terms;
+}
+
+function generateId(term) {
+  const id = term.toLowerCase().replace(/\s+/g, "-"); // Replaces spaces safely
+  // TODO: do better than hardcoding the one case in message-format
+  if (id.endsWith("rategies")) {
+    return id.slice(0, -3) + "y";
```
srl295, line 313: need plural rules! 🤔

aphillips: 🤣 A better implementation would be to have a list of irregular plurals for the spec (in the YAML bits?)

srl295 (approved):

PR #4392, opened 2025-02-22
CLDR-11950 Fix Info Panel wrong-message bug in front end

btangmu:
-For constructed value, message should be noFollowAlias = This item is constructed from other values

-The wrong message was item_description_alias_same_locale = This item is inherited from another field in this locale

-In cldrInfo.mjs, addSelectedItem calls getItemDescription and getLinkUrlAndText

-In getItemDescription, if no inheritance path, show the message for constructed value

-In getLinkUrlAndText, if no inheritance path, there is no link URL, AND no link text; link text is ignored if there is no link URL

CLDR-11950

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -328,7 +328,7 @@ function addSelectedItem(theRow) {
   const { language, direction } = getLanguageAndDirection();
   selectedItemWrapper.setLanguageAndDirection(language, direction);
 
-  const description = getItemDescription(item?.pClass, theRow?.inheritedLocale);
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -399,7 +399,7 @@ function getLinkUrlAndText(theRow, item) {
     }
     if (xpstrid === theRow.xpstrid && loc === cldrStatus.getCurrentLocale()) {
       // following the alias would come back to the current item; no link
-      linkText = cldrText.get("noFollowAlias");
+      // and no link text either
```
srl295, line 402: makes sense.

srl295 (approved):

PR #4393, opened 2025-02-22
CLDR-18176 Changes in unit constants

macchiati:
CLDR-18176

Adds a restriction, because there cannot be a normalized form that fits in 8 characters.

`[ wfc: The string length of the unit constant must be less than 9 characters. ]`

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -916,13 +916,25 @@ See [Unit_Conversion](tr35-info.md#Unit_Conversion).
 <br/>   | long_unit_identifier
 
 <a name='core_unit_identifier' href='#core_unit_identifier'>core_unit_identifier</a> 
-<br/>:= product_unit ("-" per "-" product_unit)\*  
-<br/>   | per "-" product_unit ("-" per "-" product_unit)\*
-* *Examples:*  
-    * foot-per-second-per-second  
-    * per-second  
+<br/>:= product_unit ("-" per "-" product_unit)\*
+<br/>   | per "-" product_unit\*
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35-general.md
```diff
@@ -1057,8 +1075,8 @@ grouping
     * The locale data for currency display names is supplied in the `currencies` element, not in the `units` element.
 
 Note that while the syntax allows for unit_constants in multiple places, the typical use case is only one instance, after a "-per-".
-The normalized form of a unit identifier has at most one unit_constant in the numerator and one in the denominator.
-For example, `2-kilowatt-7-hour-per-3-meter-5-second` has the equivalent normalized form `14-kilowatt-hour-per-15-meter-second`.
+The normalized, non-deprecated form of a unit identifier has at most one unit_constant in the denominator immediately after the per.
+For example, `kilowatt-hour-per-3-meter-5-second` has the equivalent normalized form `kilowatt-hour-per-15-meter-second`.
```
srl295, line 1079: Should the length restriction be mentioned here as well in the prose?

srl295 (approved):

docs/ldml/tr35-general.md
macchiati: It doesn't need to be. The syntax already contains a well-formed clause:

> * [ wfc: The string length of the unit constant must be less than 9 characters. ]

macchiati:
@sffc @younies I have merged this, but if you have any further comments we can open up a new PR.

PR #4394, opened 2025-02-22
CLDR-18054 spec: Document Triple uparrow

macchiati:
CLDR-18054

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -3097,6 +3097,18 @@ Unfortunately, XML does not have the capability to contain all Unicode code poin
 
 The element `<cp>`, which was formerly used for this purpose, has been deprecated.
 
+##### Inheritance Marker
+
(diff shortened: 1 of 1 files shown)


srl295 (approved):

AEApple:
Is it also worth explaining that it is also used for lateral inheritance in addition to inheritance from a parent locale?

srl295:
Renamed PR-too terse.

srl295:
> Is it also worth explaining that it is also used for lateral inheritance in addition to inheritance from a parent locale?

True it should say inherited location (such as a parent locale) and point to the Rules.

AEApple (approved):
LGTM! That looks like a good place for it in the spec.

Are the anchors on the sections automatically generated, or will we just add one if we need it in the future?

macchiati:
Anchors are automatically generated, and the TOC is automatically updated.

PR #4395, opened 2025-02-23
CLDR-18146 Add DTD attribute value constraints

macchiati:
CLDR-18146

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/dtd_annotations.md
+++ b/docs/site/development/dtd_annotations.md
@@ -1,0 +1,61 @@
+---
+title: Attribute Value Constraints
+---
+
+# Attribute Value Constraints
+
+The following are [DTD Annotations](https://unicode.org/reports/tr35/tr35.html#DTD_Annotations) that provide constraints on attribute values.
+They are used internally in managing and testing the data in XML files.
+Because they are for internal use for CLDR tooling, 
+they are described here instead of in LDML section: [DTD Annotations](https://unicode.org/reports/tr35/tr35.html#DTD_Annotations).
+
+They are used in DTD Annotation lines such as the following;
+
+`<!--@MATCH:{attribute value constraint}-->`
+
+The following describes the options available.
+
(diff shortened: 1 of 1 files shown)


docs/ldml/tr35.md
```diff
@@ -3647,40 +3647,10 @@ and are included below the !ELEMENT or !ATTLIST line that they apply to. The cur
...
-| range/\{start_number~{end_number}} | number between (inclusive) start and end |
-| time/\{time or date or date-time pattern} | eg HH:mm |
-| unicodeset/\{unicodeset pattern} | valid unicodeset |
-| validity/\{field}         | currency, language, locale, region, script, subdivision, short-unit, unit, variant<br/>The field can be qualified by particular enums, such as:<br/>`validity/unit/regular deprecated`: matches only _deprecated_ and _regular_<br/>`validity/unit/!deprecated`: matches all but _deprecated_ |
-| version                   | 1 to 4 digit field version, such as 35.3.9 |
-| set/\{match}              | set of elements that match \{match} |
-| or/\{match1}\|\|\{match2}…  | matches at least one of \{match1}, etc |
-
-
+Because they are intended for internal use in CLDR tooling, the {attribute value constraints} are described in [DTD Attribute Value Constraints](https://cldr.unicode.org/development/tests-and-tools).
```
srl295, line 3653: This is probably good.

 However, it would be nice to improve the XSD generation to read attributes - then, there would be a broadly implemented way to verify some of these constraints (there's a ticket for this already.)

srl295, line 3653: This is probably good.

 However, it would be nice to improve the XSD generation to read attributes - then, there would be a broadly implemented way to verify some of these constraints (there's a ticket for this already.)

srl295 (approved):

docs/ldml/tr35.md
macchiati: A second implementation is always good.

PR #4396, opened 2025-02-24
CLDR-17531 Add script variant data

macchiati:
CLDR-17531

ALLOW_MANY_COMMITS=true


macchiati:
The JSON problem was a bit tricky to fix (and as I write this, not sure if I got it right)

> org.opentest4j.AssertionFailedError: set items missing from JSON configuration.
> To fix: Add to CHILD_VALUE_IS_SPACESEP_ARRAY, VALUE_IS_SPACESEP_ARRAY, or ATTRVALUE_AS_ARRAY_SET? ==> expected: <[]> but was: <[(scriptVariant,base)]>

None of those items, after looking at the code, appeared to be the right answer. So ended up doing the following:

- I downloading a released JSON data
- looked at the ldmlSupplemental.dtd to find a similar example (an @VALUE attribute with multiple items).
- Found calendarPreference@ordering
- Checked looked at the data in the release to make sure it was generated as expected (see below) as a mapping to a list.
- Added a similar line.

```
<!ATTLIST calendarPreference ordering NMTOKENS #REQUIRED >
    <!--@MATCH:set/bcp47/ca-->
    <!--@VALUE-->
```

```
    "calendarPreferenceData": {
      "001": [
        "gregorian"
      ],
      "AE": [
        "gregorian",
        "islamic-umalqura",
        "islamic",
        "islamic-civil",
        "islamic-tbla"
      ],
```

macchiati:
Also filed https://unicode-org.atlassian.net/browse/CLDR-18337

tools/cldr-code/src/test/java/org/unicode/cldr/json/LdmlConvertRulesTest.java
```diff
@@ -76,6 +76,7 @@ void testSplittableAttributes() {
         jsonSplittableAttrs.add(Pair.of("defaultContent", "locales"));
         jsonSplittableAttrs.add(Pair.of("era", "aliases"));
         jsonSplittableAttrs.add(Pair.of("calendarPreference", "ordering"));
+        jsonSplittableAttrs.add(Pair.of("scriptVariant", "base"));
```
srl295, line 79: I'll check on json soon with this .

srl295 (approved):
Needs spec text?

tools/cldr-code/src/test/java/org/unicode/cldr/json/LdmlConvertRulesTest.java
macchiati: Thanks! I filed tickets, so no immediate need.

PR #4397, opened 2025-02-24
CLDR-10008 Update Bassa likely script and Liberia population estimates

conradarcturus:
[to be landed only after version changed to v48]

We have an outstanding ticket to fix the default writing system for the Bassa language. While there is a writing system in Unicode specific to the language, it's not adopted by many people. This change recognzied Latin script as the primary script for Bassa.

While I was doing it, the way LikelySubtags is constructed I needed to add population estimates, otherwise the "primary script" is ignored. So I got the 2022 census and added language estimates from that. Unfortunately the census only has ethnic identity information -- not languages used -- but in this area it's a sufficiently close proxy to language known.

I left out literacy data (it's probably low for most of these languages) because we just don't have a by-language breakdown. The only one I included was for Dan/Gio [dnj] because [dnj] is already listed for Ivory Coast at a very low literacy -- if I didn't add a literacy rate it would change the likely subtag -- but we don't want that.

Command to regenerate xmls:
```
```

CLDR-10008

ALLOW_MANY_COMMITS=true


srl295:
Target Ddl/v48?

srl295 (approved):
should be OK to land on the branch

PR #4398, opened 2025-02-24
CLDR-14479 Fix Belarus locale demographics

conradarcturus:
[to be merged after main is v48]

For the past few decades, Russian has been the primary language in Belarus. However, our data gives the false impression that Belarusian is still the most common.

This changelist updates the population numbers to the 2019 census. This also changes the likely subtags for Belarus.

CLDR-14479

ALLOW_MANY_COMMITS=true


macchiati:
Please bring this up in the TC for approval before any action, since it affects a TC locale.

macchiati (dismissed):

ssvb:
@macchiati @srl295 @AEApple please take a look at this as well. How much time is still left until the CLDR 48 release and what can be done to prevent the damage? I believe this is very urgent.

srl295:
> @macchiati @srl295 @AEApple please take a look at this as well. How much time is still left until the CLDR 48 release and what can be done to prevent the damage? I believe this is very urgent.

Posting on a merged pull request on a closed ticket is not the right way…

Please add a comment to the ticket link at the top

srl295:
There are many months before 48 release and you can see the schedule on the top of the CLDR homepage

conradarcturus:
Er, I cannot reopen the ticket (@srl295  is that possible?) Perhaps we can open a new one.

Nonetheless I do want to share my experience - I have worked on product launches in Belarus. We were so excited to translate into the Belarusian language, making it equal to Russian in our product. However, when we launched, >99% of the users choose Russian. It was very disappointing because we were so excited and invested a lot of energy and money in it -- but the reality on the ground did not match. That incident ended up backfiring and causing the business-side to consider decreasing funding for efforts to translate into other native languages -- saying we were wasting the money translating into dying languages and that our data was lying to us. It ended up harming our group's goal to provide more translations across the world.

Having accurate data reflecting on common usage of the people in a country is important. I definitely did not appreciate the tone of the original Jira ticket creator -- but I had to swallow my pride and accept that he was right. Belarusian may be the native language of Belarusians -- but the very same census you quote also says that the language used commonly at Home is Russian -- that's where my numbers come from.

srl295:
@ssvb please continue the discussion on https://unicode-org.atlassian.net/browse/CLDR-14479?focusedCommentId=182135 which is re-opened. (not here on GitHub)

PR #4399, opened 2025-02-24
CLDR-14551 ku name in en xml

conradarcturus:
[Feel free to discuss, do not stamp approve, rather let's double check this is what we want]

In ISO `ku` signifies Kurdish in general, as a macrolanguage. However, in CLDR `ku` represents the largest constituent language, `kmr` or Northern Kurdish. So right now language names gives the options `ku` Kurdish, `ckb` Central Kurdish, and `sdh` Southern Kurdish. Since "Kurdish" is ambiguous, this changes `ku` to Northern Kurdish as we use it in CLDR.

Long-term I want to handle `ku` and `kmr` separately, but that is not the CLDR paradigm today.

CLDR-14551

ALLOW_MANY_COMMITS=true


macchiati (changes requested):
I think https://unicode-org.atlassian.net/browse/CLDR-18144 provides a better approach to this, so I don't want to merge this before discussing that.

srl295:
@conradarcturus given discussion needed, i'm moving this to draft…

PR #4401, opened 2025-02-25
CLDR-18342 merge ddl/v48 back into main

srl295:
… after v47 is branched to maint

The purpose of this PR now is to keep an eye on what's actually in the `ddl/v48` branch.

CLDR-18342

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true


macchiati:
This can happen only after we branch, right?

srl295:
> This can happen only after we branch, right?

yes, that's why it has v48 in the front to keep it from being merged.

conradarcturus (approved):
Now that main is v48 can we merge this?

srl295:
Yes, but it needs a special merge. I will do it later.

PR #4402, opened 2025-02-25
CLDR-11874 json: output subdivisions to the correct spot

srl295:
- the configuration wasn't updated after an internal path change, so the files were in the wrong place

CLDR-11874

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/json/JSON_config_subdivisions.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/json/JSON_config_subdivisions.txt
@@ -4,7 +4,7 @@
 package=subdivisions ; packageDesc=Translated versions of locale display name elements: subdivisions.
 
 # mapping paths to sections
-section=subdivisions ; path=//cldr/subdivisions/[^/]++/localeDisplayNames/subdivisions/.* ; package=subdivisions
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/resources/org/unicode/cldr/json/JSON_config_subdivisions.txt
```diff
@@ -4,7 +4,7 @@
 package=subdivisions ; packageDesc=Translated versions of locale display name elements: subdivisions.
 
 # mapping paths to sections
-section=subdivisions ; path=//cldr/subdivisions/[^/]++/localeDisplayNames/subdivisions/.* ; package=subdivisions
+section=subdivisions ; path=//cldr/subdivisions/localeDisplayNames/subdivisions/.* ; package=subdivisions
```
btangmu, line 7: What was the reason for `[^/]++/`?

btangmu (approved):

tools/cldr-code/src/main/resources/org/unicode/cldr/json/JSON_config_subdivisions.txt
srl295: The original path was one more level deep.

PR #4405, opened 2025-02-25
CLDR-18279 Preliminary refactoring of CheckDates.java

btangmu:
-Nearly all changes recommended/assisted by IntelliJ, including removal of dead code

CLDR-18279

ALLOW_MANY_COMMITS=true


macchiati (commented):
Looks pretty good. But there are enough different kinds of characters that it will take some time to verify that it has exactly the same behavior.

Can you relay how you did the refactoring. If, for example, you used Eclipse or Intellij refactoring for some things, then we don't have to check; or if you are deleting unused items, etc.

macchiati (approved):

btangmu:
I accidentally put the wrong ticket number (CLDR-18279) in the commit; it should have been CLDR-13589.

PR #4409, opened 2025-02-25
CLDR-18340 Add missing items to Modifications

macchiati:
CLDR-18340

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -4334,12 +4334,39 @@ Other contributors to CLDR are listed on the [CLDR Project Page](https://www.uni
 
 ### Number symbols and formats without numberSystem
 - Noted that [Number Symbols](tr35-numbers.md#Number_Symbols) and [Number Formats](tr35-numbers.md#Number_Formats) without a specific `numberSystem` attribute should not be used and will be deprecated in CLDR v48.
+- 
 ### Clarified `currencyData` element ordering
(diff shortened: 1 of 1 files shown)


macchiati:
@AEApple , I think once this is done, we can summarize the changes from it for the blog post and release page.

srl295 (approved):

macchiati:
We can open a new PR for the MF2 changes. @aphillips

docs/ldml/tr35.md
```diff
@@ -4334,13 +4334,40 @@ Other contributors to CLDR are listed on the [CLDR Project Page](https://www.uni
...
+    * Clarified the normalized form of units.  
+* Modified the process of formatting to handle unit\_constants  
+* [Unit Preferences Overrides](tr35-info.html#Unit_Preferences_Overrides)
+  * Clarified the process of deriving the region for overrides
+ 
+### [DTD Annotations](https://cldr.unicode.org/development/dtd_annotations)
+* Fixed the description of @ORDERED
+* Expanded the documentation for @MATCH and moved to [cldr.unicode.org](https://cldr.unicode.org/development/dtd_annotations)
+
+### [Inheritance](https://unicode.org/reports/tr35/dev/tr35.html#inheritance-marker)
```
srl295, line 4365: @macchiati why does this link to <https://unicode.org/reports/tr35/dev/tr35.html> ?

macchiati: That was to point people to the section where changes were being made, during alpha/beta. Those links all need to be fixed for release, but I don't think we had that in the BRS tasks.

macchiati: In this particular case, they should have instead been relative links, since it was within the spec! That way no update is needed.

PR #4410, opened 2025-02-26
CLDR-18347 GenerateDTD

macchiati:
CLDR-18347

ALLOW_MANY_COMMITS=true


macchiati:
Note, doesn't have to be in the beta, but best to merge before branching maint.

srl295 (approved):

PR #4411, opened 2025-02-26
CLDR-18349 Keep scriptData when generating ConvertLanguageData

conradarcturus:
ConvertLanguageData rebuilds part of the supplementalData.xml. Since the new <scriptData /> is in the middle of 2 generated sections we need to re-write the generator so it does not accidentally drop it when running the script. Long term this is only more reason to break up the supplementalData.xml - but let's plan that out first.

Can be landed in v48 or v47, no changes will affect ICU.

Code to run generator:
```
mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData
```

CLDR-18349

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
@@ -135,9 +135,6 @@ public class ConvertLanguageData {
         try (final BufferedReader oldFile = FileUtilities.openUTF8Reader(oldSupp);
                 final PrintWriter newFile = FileUtilities.openUTF8Writer(genSupp);
                 final PrintWriter newLsraw = FileUtilities.openUTF8Writer(genLsraw); ) {
-            // load elements we care about
-            CldrUtility.copyUpTo(
-                    oldFile, PatternCache.get("\\s*<languageData>\\s*"), newFile, false);
 
             Set<String> available = cldrFactory.getAvailable();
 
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
```diff
@@ -1145,7 +1158,7 @@ private static void writeTerritoryLanguageData(
         LanguageTagParser ltp = new LanguageTagParser();
 
         out.println(
-                " <!-- See http://unicode.org/cldr/data/diff/supplemental/territory_language_information.html for more information on territoryInfo. -->");
+                "\t<!-- See http://unicode.org/cldr/data/diff/supplemental/territory_language_information.html for more information on territoryInfo. -->");
```
srl295, line 1161: obsolete link?

srl295 (dismissed):
I was going to say 48, but probably worth having in 47 so we don't lose the data if an update needed to happen.

I think this is reasonable.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
conradarcturus: ahaha. Yea, I should patch this link. Let me see what it should point to...

conradarcturus: https://www.unicode.org/cldr/charts/latest/supplemental/territory_language_information.html looks like a good new target

srl295 (approved):

PR #4435, opened 2025-03-02
CLDR-18363 Fix formatting and links

AEApple:
CLDR-18363 Fix formatting and links and remove seed

ALLOW_MANY_COMMITS=true


AEApple:
Need an example for the CLDRModify with Config -> the current link didn't work.

I will update Ricks BRS tasks list in a different PR. I was reading through them and they're very out of date.

macchiati (approved):

PR #4437, opened 2025-03-02
CLDR-18242 Migrate generate-algorithmic-locales.md

AEApple:
CLDR-18242

Migrate generate-algorithmic-locales.md

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/generate-algorithmic-locales.md
+++ b/docs/site/development/generate-algorithmic-locales.md
@@ -1,0 +1,18 @@
+---
+title: Generate algorithmic locales
+---
+
+# Generate algorithmic locales
+
+* Run (and edit if necessary when?) tool/CLDRFileTransformer to generate transformed locales. 
+* Copy the generated xml files into common/{main, annotations, subdivisions}.
+
+> NOTE: Need to re-run each modified file in (main, annotations, subdivisions) through CLDRModify -fp (but NOT minimize (-r)) after generation!
+
+ * Can be done with filter, eg (but update list!!)
+
+```
+-s${workspace_loc}/cldr/common/annotations
+-fp
+m(de_CH|ha_NE|sr_Latn_BA|sr_Latn_ME|sr_Latn_XK|sr_Latn|yo_BJ|yue_Hans)\
+```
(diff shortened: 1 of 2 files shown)


AEApple:
@srl295 @macchiati - should the commands be copyable separately?

AEApple:
I added back all the trailing spaces, but it's still not working as expected, but maybe I missed one?

docs/site/development/generate-algorithmic-locales.md
```diff
@@ -0,0 +1,20 @@
...
+
+ * Can be done with filter, eg (but update list!!)
+
+```
+-s${workspace_loc}/cldr/common/annotations
+-fp
+m(de_CH|ha_NE|sr_Latn_BA|sr_Latn_ME|sr_Latn_XK|sr_Latn|yo_BJ|yue_Hans)\
+```
+
+{TODO: Steven file a ticket to automate Transform -> Modify}
```
srl295, line 20: Can we take todo out of the site ?

srl295 (approved):

srl295:
> @srl295 @macchiati - should the commands be copyable separately?

What do you mean ?

docs/site/development/generate-algorithmic-locales.md
AEApple: You no longer plan to do it? I just moved it over from the original page?

AEApple:
> > @srl295 @macchiati - should the commands be copyable separately?
> 
> What do you mean ?

When you see it in GitHub there's actually a nice copy option for that block of commands, but it would copy the whole block, so not sure if you want it to be where you just copy a command at a time? Longer term it might be nice to find a way to have the code blocks when the site is rendered to make it easier for people to copy stuff from the internal development pages.

srl295:
> > > @srl295 @macchiati - should the commands be copyable separately?
> 
> > 
> 
> > What do you mean ?
> 
> 
> 
> When you see it in GitHub there's actually a nice copy option for that block of commands, but it would copy the whole block, so not sure if you want it to be where you just copy a command at a time? Longer term it might be nice to find a way to have the code blocks when the site is rendered to make it easier for people to copy stuff from the internal development pages.

Sure we can add a button. Even better is to simplify the process so there's less to copy. Or run automatically.

srl295 (dismissed):

AEApple:
> LGTM. I can navigate to the page and it displays. (I can't comment on the content, but I think this PR is just for the migration.)

Yes this is just for migration.

PR #4438, opened 2025-03-02
Consistently use preferred nonterminal references in UTS 35

gibson042:
CLDR-_____

I'll retroactively open a CLDR ticket later, but noticed this inconsistent use of `keyword` vs. `ufield` in passing and wanted to correct it without needed to context switch.

ALLOW_MANY_COMMITS=true


gibson042:
Hmm, I guess renaming a branch closes any associated PR. Anyway, the replacement is at #4440.

PR #4439, opened 2025-03-03
CLDR-18365 investigate unit apis

macchiati:
CLDR-18365

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestConverterAPI.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestConverterAPI.java
@@ -1,0 +1,482 @@
+package org.unicode.cldr.unittest;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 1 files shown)


macchiati:
Current timing on my laptop.

```
iterations: ~100,000,000
0.791 ns: 	Doubles w/existing MeasureUnit
8.8 ns: 	Doubles w/unbounded caller cache — ConcurrentHashMap
38.2 ns: 	Doubles w/bounded caller cache — LoadingCache
46.4 ns: 	Doubles
266 ns: 	BigDecimals (64)
1,460 ns: 	Rationals
```
When you create a MeasureUnit, the constructor builds the information it needs for conversion. (I'm using the name MeasureUnit2 for clarity, but this could go into an extension of the regular ICU MeasureUnit.)

The code currently has a separate UnitConverter2, with statics. These could also be moved into MessageUnit, thus not needing a separate class for conversion.

Notes:
- The best speed is when the caller uses MeasureUnits rather than strings. All the conversion info is right there.
- The caller could also use their own cache (String ==> MeasureUnit), so I gave two variants.
- The code supports 3 variants, of increasing precision (and decreasing speed).
- For BigDecimal division, you need to limit the precision of any division; this code uses MathContext.DECIMAL64.

PR #4441, opened 2025-03-03
CLDR-18359 Create Laz [lzz] locale

conradarcturus:
CLDR-18359

ALLOW_MANY_COMMITS=true


conradarcturus:
Obviously this has errors -- I'm having a hard time adding some of the indicated characters with Carot and Circumflex diacritics.

srl295:
On the test, it should be triggering the message per #3760  but it's not.

Please run: `mvn -DSHOW_EXEMPLARS test --file tools/pom.xml '-Dorg.unicode.cldr.unittest.testArgs=-f:LikelySubtagsTest -n -q'`

and it will update the necessary .txt file

As I noted in CLDR-17395 I think this particular test serves no useful purpose but to increase friction around exemplar chars and should be removed.

roozbehp:
> I'm wondering about the t+breve (U+0074 U+0306) in the main exemplar sequence. The other characters are using caron, for example p+caron (U+0070 U+030C).

I think breve must be used in all cases. I found someone's thesis on font design for Laz and attached it to the Jira ticket. It says the breve was chosen for consistency with Turkish ğ.

roozbehp:
> I think breve must be used in all cases.
Here's my attempt at fixing the exemplars: 
```
<exemplarCharacters>[a b cç {ç̆} d {dz} e f gğ h iİ j k {k̆} l m n o p {p̆} q r sş t {t̆} {ts} {tz} u v x y z]</exemplarCharacters>
<exemplarCharacters type="auxiliary">[ı ö ü w {z̆} ʒ {ʒ̆}]</exemplarCharacters>
<exemplarCharacters type="index">[A B CÇ {Ç̆} D E F GĞ H İ J K {K̆} L M N O P {P̆} Q R SŞ T {T̆} U V W X Y Z]</exemplarCharacters>
```

conradarcturus:
LikelySubtagsTest: testGetResolvedScriptVsExemplars still doesn't like the characters with breves or carons.

In the new it's complaining about:
{k̆}{p̆}{t̆}{ç̆}

In the old its complaining about
{p̌}{t̆}{ç̌}

srl295:
> LikelySubtagsTest: testGetResolvedScriptVsExemplars still doesn't like the characters with breves or carons.
> 
> In the new it's complaining about: {k̆}{p̆}{t̆}{ç̆}
> 
> In the old its complaining about {p̌}{t̆}{ç̌}

Did you update scriptToExemplars.txt?

srl295:
filed https://unicode-org.atlassian.net/browse/CLDR-18384 to get rid of the test entirely

conradarcturus:
Thanks for the feedback everyone! That really helped. I changed the letters to use breves and ran the script that rebuilt exemplars. Looks great!

```
mvn -DSHOW_EXEMPLARS test --file tools/pom.xml '-Dorg.unicode.cldr.unittest.testArgs=-f:LikelySubtagsTest -n -q'
```

srl295 (approved):

roozbehp:
Generally looks good to me, except that I didn't think Ž and ž need to be in braces, since they are precomposed Unicode characters.

srl295:
> Generally looks good to me, except that I didn't think Ž and ž need to be in braces, since they are precomposed Unicode characters.

doesn't hurt anything (all of this gets normalized eventually anyway)

roozbehp (approved):

common/main/lzz.xml
```diff
@@ -0,0 +1,26 @@
...
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="lzz"/>
+	</identity>
+	<localeDisplayNames>
+		<languages>
+			<language type="en">İngilizuri</language>
+			<language type="laz">Lazuri</language>
```
lucaswerkmeister, line 16: I believe this should be `type="lzz"`, not `type="laz"`? This seems to be defining a language name for [Aribwatsa](https://iso639-3.sil.org/code/laz), an unrelated extinct language :/

DavidLRowe: @lucaswerkmeister Good catch! Thanks for taking the time to report it. Since this PR is already merged, I've opened a new [CLDR ticket](https://unicode-org.atlassian.net/browse/CLDR-19316) to correct this error.

macchiati:
> > Generally looks good to me, except that I didn't think Ž and ž need to be in braces, since they are precomposed Unicode characters.
> 
> doesn't hurt anything (all of this gets normalized eventually anyway)

It is actually always safest to have braces just in case. Users may not be familar with which user-characters are represented by a single code point and which are not.

The internal software (CLDRModify) will normalize the format eventually, and unneeded braces (around a string that NFC normalizes to a single character) will be discarded.

common/main/lzz.xml
lucaswerkmeister: Great, thanks for creating the ticket! (I tried but couldn’t get access even after logging in. Probably not worth debugging now ^^)

conradarcturus: Whoops! Thanks for figuring out the mistake.

Fix posted: https://github.com/unicode-org/cldr/pull/5485

PR #4443, opened 2025-03-04
CLDR-16715 Match Classical Chinese with Traditional Han writing

conradarcturus:
The Classical Chinese [lzh] language is the standardized form of Chinese that arose around the start of the Common Era (potentially as early as 5th Century BCE). It is marked as ending in common usage by the May Fourth Movement in 1919 -- it was replaced with vernacular Chinese [zh & it's constituents].

The majority of its duration coincides with Traditional Han writing [Hant]. While many writings were in the earlier scripts, the vast majority of corpora are in Traditional script.  The printed form in which "Traditional Han writing [Hant]" now most often occurs arose in the Song dynasty starting in the 10th century CE.

The ending of common usage of Classical Chinese was 37 years before the advent of Simplified Chinese (1956). Since, for historic languages, we match their writing system to the largest corpora, we should fix this and match it to Traditional Han writing.

CLDR-16715

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -441,7 +441,7 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="luz" to="luz_Arab_IR"/>		<!--Southern Luri‧?‧?	➡ Southern Luri‧Arabic‧Iran-->
 		<likelySubtag from="lv" to="lv_Latn_LV"/>		<!--Latvian‧?‧?	➡ Latvian‧Latin‧Latvia-->
 		<likelySubtag from="lwl" to="lwl_Thai_TH"/>		<!--Eastern Lawa‧?‧?	➡ Eastern Lawa‧Thai‧Thailand-->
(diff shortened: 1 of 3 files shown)


roozbehp (approved):

btangmu:
"...standardized form of Chinese that arose around the start of the Common Era (potentially as early as 5th Century BCE)" -- that's exaggerated. The brush-written "regular script", 楷體 kǎitǐ, dates back to about the 3rd century CE. Before that was an evolution from seal script, extremely different from regular script. The printed form in which "Traditional Han writing [Hant]" now most often occurs arose in the Song dynasty starting in the 10th century CE.

I'm not disagreeing with the data changes, only disagreeing with the explanation. To say any form of Chinese in use today was standardized at the start of the Common Era, let alone centuries earlier, is like saying that Julius Caesar wrote in French. All forms of Chinese writing arose gradually through many centuries of evolution and standardization.

conradarcturus:
Thanks @btangmu  for the clarification! I'm fortunate to work with such knowledgable people :)

srl295 (approved):

PR #4445, opened 2025-03-04
CLDR-5708 Remove misleading territories attribute from supplementalData

conradarcturus:
supplementalData.xml lists the top territories for languages under the `<languageData>` area -- however it's not clear and it's best for people to consume the data from `<territoryInfo>` directly.

This quickly removes the territory information -- but there is additional cleanup to do. I just wanted to get this simple change out there to see how many breaks.

CLDR-5708

Scripts ran
```
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
@@ -333,6 +333,7 @@ public class SupplementalDataInfo {
 
         private Map<String, Integer> scriptsByPopulation = new TreeMap<>();
 
+        // TODO CLDR-18087 completely remove territories
         private Set<String> territories = Collections.emptySet();
 
         public Type getType() {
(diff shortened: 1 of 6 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -333,6 +333,7 @@ public enum Type {
 
         private Map<String, Integer> scriptsByPopulation = new TreeMap<>();
 
+        // TODO CLDR-5708 completely remove territories
```
srl295, line 336: TO-DONE?

srl295 (commented):
would it be better to deprecate this and then actually remove it in a future version?

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
conradarcturus: Unfortunately, it will require significantly more changes to remove this variable. All I did is remove territories from the string output. I didn't remove the variable.

I discussed the approach in the ticket:
https://unicode-org.atlassian.net/browse/CLDR-5708?focusedCommentId=181278

srl295: where you said:

> and I recommend still landing `#1` as a first-step then making a separate Jira Ticket for `#2`.

If so, then `TODO CLDR-5708` needs to be updated to point to the new ticket for `#2`, right?

conradarcturus:
> would it be better to deprecate this and then actually remove it in a future version?

We included in the migration notes for v47 that we wanted to take it away. Can you explain more detail how you'd want to deprecate it?

macchiati:
> > would it be better to deprecate this and then actually remove it in a future version?
> 
> We included in the migration notes for v47 that we wanted to take it away. Can you explain more detail how you'd want to deprecate it?

Agreed

macchiati (approved):

PR #4454, opened 2025-03-04
CLDR-11231 Update Tokelauan language name in generated files

conradarcturus:
I was running generating files and I noticed there were a few places to auto-update Tokelauan. I made this quick PR to do this so it doesn't happen in unrelated PRs.

```
mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags
```

CLDR-11231

ALLOW_MANY_COMMITS=true


DavidLRowe (approved):
Thanks for catching these.

PR #4457, opened 2025-03-05
CLDR-18325 Make CLDRFile.DEFAULT_ITERATION_INCLUDES_EXTRAS true

btangmu:
-In Ldml2JsonConverter call iteratorWithoutExtras instead of the default iterator

-In ShowData (for GenerateAllCharts) call iterableWithoutExtras instead of the default iterable

CLDR-18325

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/Ldml2JsonConverter.java
@@ -562,7 +562,8 @@ public class Ldml2JsonConverter {
         // read paths in DTD order. The order is critical for JSON processing.
         final CLDRFile.Status status = new CLDRFile.Status();
         for (Iterator<String> it =
-                        file.iterator("", DtdData.getInstance(fileDtdType).getDtdComparator(null));
+                        file.iteratorWithoutExtras(
+                                "", DtdData.getInstance(fileDtdType).getDtdComparator(null));
                 it.hasNext(); ) {
             int cv = Level.UNDETERMINED.getLevel();
             final String path = it.next();
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowData.java
@@ -187,13 +187,13 @@ public class ShowData {
 
                 // get all of the paths
                 Set<String> allPaths = new HashSet<>();
-                file.forEach(allPaths::add);
+                file.iterableWithoutExtras().forEach(allPaths::add);
 
                 if (!locale.equals("root")) {
                     for (String childLocale : children) {
                         CLDRFile childCldrFile = cldrFactory.make(childLocale, false);
                         if (childCldrFile != null) {
-                            childCldrFile.forEach(allPaths::add);
+                            childCldrFile.iterableWithoutExtras().forEach(allPaths::add);
                         }
                         sublocales.put(childLocale, childCldrFile);
                     }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -1474,7 +1474,7 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         return s;
     }
 
-    private final boolean DEFAULT_ITERATION_INCLUDES_EXTRAS = false;
+    private final boolean DEFAULT_ITERATION_INCLUDES_EXTRAS = true;
 
     public Iterator<String> iterator() {
         if (DEFAULT_ITERATION_INCLUDES_EXTRAS) {
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -1554,7 +1554,7 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         }
     }
 
-    private Iterable<String> iterableWithoutExtras() {
+    public Iterable<String> iterableWithoutExtras() {
         return this::iteratorWithoutExtras;
     }
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -1533,7 +1533,7 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         return dataSource.iterator(pathFilter);
     }
 
-    private Iterator<String> iteratorWithoutExtras(String prefix, Comparator<String> comparator) {
+    public Iterator<String> iteratorWithoutExtras(String prefix, Comparator<String> comparator) {
         Iterator<String> it =
                 (prefix == null || prefix.isEmpty())
                         ? dataSource.iterator()


btangmu:
While doing the v47 ticket, locally, I tested making DEFAULT_ITERATION_INCLUDES_EXTRAS true or false and found it to make no difference to data produced by Generate VXML and GenerateProductionData. It is possible, however, that other tools may produce different results, and we may not discover the differences until we run those other tools.

srl295 (dismissed):
This does what it says. gave my reservation about this but no major objection!

srl295:
could be worth generating charts and json before and after.  json also used just an 'ordinary' iterator currently.

macchiati:
> could be worth generating charts and json before and after. json also used just an 'ordinary' iterator currently.

The iterator that skips nulls is fine for any processing that only cares about the populated data; so charts, json, etc. BTW I assume, @srl295 , that json converts the fully resolved data.

macchiati (approved):

srl295:
> > could be worth generating charts and json before and after. json also used just an 'ordinary' iterator currently.
> 
> The iterator that skips nulls is fine for any processing that only cares about the populated data; so charts, json, etc. BTW I assume, @srl295 , that json converts the fully resolved data.

I think json uses unresolved data (it processes each locale ID), and doesn't seem to skip nulls so it might get stuck there.

btangmu:
I think it's worthwhile for me to generate json and charts with the boolean both ways and see whether it makes any difference

btangmu:
More observations:

ST (on cldr-staging) shows examples 1 and 2 with status "Missing", and example 3 with status "Approved".

common/main/af.xml has:
```
			<language type="ckb">Sorani</language>
			<language type="ckb" alt="menu">↑↑↑</language>
			<language type="ckb" alt="variant">Koerdies Sorani</language>
```

macchiati (changes requested):
I'm a bit confused. 

There are three possible iterations
1. no extrapaths
2. extrapaths, but only if the value (**resolved, if with a resolving source**) ≠ null
3. all extrapaths

I had thought that we had access to all three, and were discussing which should be the default, and what the names of the others would be.

From a quick glance (and maybe I'm missing something), we generally want the iteration to be all extrapaths (3), but then want to easily be able to choose the others. ST would want all (\#3) and json / production would want non-null (\#2).

btangmu:
@macchiati "json / production would want non-null" -- yes, the differences here are based on including extrapaths only if the value ≠ null.

`org.unicode.cldr.json.Ldml2JsonConverter#mapPathsToSections` is where the CLDRFile iteration takes place for cldr-json:

`       for (Iterator<String> it =
                        file.iterator("", DtdData.getInstance(fileDtdType).getDtdComparator(null));
                it.hasNext(); ) {
`
I think it uses the data in cldr-staging rather than using the data in cldr directly. The data in cldr-staging is presumably filtered to exclude paths whose status is "missing". However, with the boolean in this ticket true, the iteration includes extra paths if they have values (inherited, etc.), and then there doesn't appear to be any filtering on status.

That would explain why we get values even with status "missing" (Examples 1 and 2), which seems like a bug, so it's a good thing we haven't merged this PR.

On the other hand, Example 3 shows an "Accepted" value, that's included in xml (with inheritance marker),  but currently excluded from json (unless the boolean is true), which seems inconsistent.

srl295:
> @macchiati "json / production would want non-null" -- yes, the differences here are based on including extrapaths only if the value ≠ null.
> 
> `org.unicode.cldr.json.Ldml2JsonConverter#mapPathsToSections` is where the CLDRFile iteration takes place for cldr-json:
> 
> ` for (Iterator<String> it = file.iterator("", DtdData.getInstance(fileDtdType).getDtdComparator(null)); it.hasNext(); ) {` I think it uses the data in cldr-staging rather than using the data in cldr directly.

Correct.

> The data in cldr-staging is presumably filtered to exclude paths whose status is "missing". 

I'm not sure how it would be so filtered. Paths whose status is missing aren't present by definition.

> However, with the boolean in this ticket true, the iteration includes extra paths if they have values (inherited, etc.), and then there doesn't appear to be any filtering on status.

Right. 

> That would explain why we get values even with status "missing" (Examples 1 and 2), which seems like a bug, so it's a good thing we haven't merged this PR.

Or, means that cldr-json and ST should have a code change to choose the no-extrapaths version? 

> On the other hand, Example 3 shows an "Accepted" value, that's included in xml (with inheritance marker), but currently excluded from json (unless the boolean is true), which seems inconsistent.

From the ST link you gave, this path is not included in coverage. 

`ckb alt="menu"` does not show up in the production data for afrikaans. Perhaps because it's a triple-up-arrow and so can just be dropped?

```
cldr-staging/production/common/main/ar.xml:			<language type="ckb" alt="menu">الكردية، السورانية</language>
cldr-staging/production/common/main/az.xml:			<language type="ckb" alt="menu">Kürdcə, mərkəzi</language>
cldr-staging/production/common/main/bal_Latn.xml:			<language type="ckb" alt="menu">Myáni Kordi- ckb</language>
cldr-staging/production/common/main/be.xml:			<language type="ckb" alt="menu">курдская, сарані</language>
cldr-staging/production/common/main/be_TARASK.xml:			<language type="ckb" alt="menu" draft="provisional">курдзкая, сарані</language>
cldr-staging/production/common/main/bew.xml:			<language type="ckb" alt="menu" draft="unconfirmed">Kurdi, Sorani</language>
cldr-staging/production/common/main/bn.xml:			<language type="ckb" alt="menu">কুর্দিশ, মধ্য</language>
cldr-staging/production/common/main/br.xml:			<language type="ckb" alt="menu">kurdeg kreiz</language>
cldr-staging/production/common/main/brx.xml:			<language type="ckb" alt="menu">कुर्दिस, मिरु</language>
cldr-staging/production/common/main/ca.xml:			<language type="ckb" alt="menu">sorani</language>
cldr-staging/production/common/main/cs.xml:			<language type="ckb" alt="menu">kurdština (centrální)</language>
cldr-staging/production/common/main/cv.xml:			<language type="ckb" alt="menu" draft="unconfirmed">тӗп курд</language>
cldr-staging/production/common/main/da.xml:			<language type="ckb" alt="menu">kurdisk, sorani</language>
cldr-staging/production/common/main/de.xml:			<language type="ckb" alt="menu">Kurdisch (Sorani)</language>
cldr-staging/production/common/main/el.xml:			<language type="ckb" alt="menu">Κουρδικά, Κεντρικά</language>
cldr-staging/production/common/main/en.xml:			<language type="ckb" alt="menu">Kurdish, Central</language>
```

This data is showing up in json, for example for `cs`

```json
          "ckb": "kurdština (sorání)",
          "ckb-alt-menu": "kurdština (centrální)",
```

btangmu:
> > That would explain why we get values even with status "missing" (Examples 1 and 2), which seems like a bug, so it's a good thing we haven't merged this PR.
> 
> Or, means that cldr-json and ST should have a code change to choose the no-extrapaths version?

Depends what "ST" means here. There was a bug where ST skipped paths needed for VXML generation because they weren't already in XML, but they had votes in the database. That's one of the reasons we're doing all this.

> > On the other hand, Example 3 shows an "Accepted" value, that's included in xml (with inheritance marker), but currently excluded from json (unless the boolean is true), which seems inconsistent.
> 
> From the ST link you gave, this path is not included in coverage.

Do you mean not in the coverage level required for inclusion in cldr-json? Would that be "modern"? If so, at what stage do "comprehensive" paths get filtered out?
 
> `ckb alt="menu"` does not show up in the production data for afrikaans. Perhaps because it's a triple-up-arrow and so can just be dropped?

Perhaps... There's a lot I'm still ignorant about, involving the relationships between cldr xml (common/main, etc.), production data (GenerateProductionData), cldr-staging xml, and cldr-json, in terms of what's included/excluded based on status, coverage, ..., and when triple-up-arrow gets converted to Bailey or just gets omitted...

macchiati:
Right, we absolutely need ST for vetters to include extrapaths-including-null-values, including those with no values. Otherwise the vetters wouldn't see the paths that don't have root values. 
These are important, like all plural categories other than 'other', grammatical categories, etc. The extrapaths have always included a host of like items; and in v48 they are including language, script, and territory names

And ConsoleCheck, in order to accurately reproduce what vetters see, has to do the same.

So the default needs to be `extrapaths-including-null-values`, to prevent skipping stuff that is important.

For more specialized purposes, code needs to also support (with different names) the two other kinds of iteration:
- `excluding-extrapaths`
- `extrapaths-excluding-null-values`
The more specific names can warn people which stuff is being skipped.

srl295:
> Right, we absolutely need ST for vetters to include `including-extrapaths-including-null-values`, including those with no values. Otherwise the vetters wouldn't see the paths that don't have root values.

That's done by explicitly calling extrapaths and handling them specially - not in the iterator (currently). That way these are paths that are known to not have a value and are handled as such.

btangmu:
The default iteration has been to include code-fallback, and to exclude extra-paths. That's somewhat arbitrary. There were longstanding bugs we didn't notice since relatively few paths were extra. Now a much larger set of paths are implemented with extra-paths instead of code-fallback.

btangmu:
If we assume extra paths aren't needed for generating json, there's a simple solution, which is to call `iteratorWithoutExtras` instead of `iterator` in `Ldml2JsonConverter.mapPathsToSections`.

It could be interesting/illuminating to investigate the question further, especially if there is any chance of discovering an exception -- where including extra paths actually adds something beneficial to the json data. It's not clear how likely that is or whether the investigation would justify prioritizing.

Meanwhile I want to conduct a similar test for generating charts. The situation may be similar to generating json. If differences result, we can either just call `iteratorWithoutExtras` instead of `iterator` when generating charts, or investigate further.

btangmu:
I tested generating charts (for v47) with the boolean false and true. The results are like what is shown below. The only differences are within `docs/charts/47/summary` and they are always differences in the number of Aliased items, with the number being larger if the boolean is true:

`$ diff -r -x '.DS_store' ../cldr-staging-boolean-false/docs/charts ../cldr-staging-boolean-true/docs/charts
diff -r -x .DS_store ../cldr-staging-boolean-false/docs/charts/47/summary/aa.html ../cldr-staging-boolean-true/docs/charts/47/summary/aa.html
21837c21837
< <tr><td class='a'>Aliased items: </td><td>9137</td></tr>
---
> <tr><td class='a'>Aliased items: </td><td>9162</td></tr>
diff -r -x .DS_store ../cldr-staging-boolean-false/docs/charts/47/summary/ab.html ../cldr-staging-boolean-true/docs/charts/47/summary/ab.html
25377c25377
< <tr><td class='a'>Aliased items: </td><td>9128</td></tr>
---
> <tr><td class='a'>Aliased items: </td><td>9133</td></tr>
diff -r -x .DS_store ../cldr-staging-boolean-false/docs/charts/47/summary/af.html ../cldr-staging-boolean-true/docs/charts/47/summary/af.html
29128c29128
< <tr><td class='a'>Aliased items: </td><td>13691</td></tr>
---
> <tr><td class='a'>Aliased items: </td><td>13695</td></tr>
`

btangmu:
I've added a 2nd commit, to call iteratorWithoutExtras instead of the default iterator, in Ldml2JsonConverter.

I still want to figure out how to do skip extras in GenerateAllCharts...

btangmu:
I've re-requested review. This seems ready to merge now that it skips extra paths for both cldr-json and charts.

This essentially completes the ticket, except that I'd like to do some follow-up refactoring (and of course we may discover more contexts in which extra paths should be skipped).

There are many places with code like this:
```
            for (Iterator<String> it = file.iterator(); it.hasNext(); ) {
                String path = it.next();
```

They can be simplified to:

```
            for (String path : file) {
```

which is a lot easier to read. Also there are a lot of compiler warnings in CLDRFile that should be cleaned up.

macchiati:
Let's do refactoring under their own tickets. It is way easier to review PRs that don't mix refactoring with substantive changes.

btangmu:
> Let's do refactoring under their own tickets. It is way easier to review PRs that don't mix refactoring with substantive changes.

That's a good reason for using a separate PR, not necessarily a separate ticket, right?

macchiati (approved):

macchiati:
Yes, can be separate pr

PR #4458, opened 2025-03-05
CLDR-18309 Remove inheritance marker if inherited value is missing

btangmu:
-New CLDRModify option -fm to remove such values

-Update XML files per modify -fm

-Revise/simplify test checkNullWithInheritanceMark, no more logKnownIssue

CLDR-18309

ALLOW_MANY_COMMITS=true


srl295 (approved):
All good, but I have a related question — should this be a test? I would imagine we should fail a PR that introduces (or causes) these without running this cleanup?

macchiati:
He did that with the TestCheckCLDR.java change.

macchiati (approved):

PR #4463, opened 2025-03-07
CLDR-18114 Add explicit script markers in population data

conradarcturus:
These languages had multiple primary scripts + different primary scripts in different countries. This changes it so they each have 1 primary script and countries are specifically annotated. See the analysis in the CLDR ticket. If you have any concerns I'm happy to iterate or cut out a change (reserving it for a later time).

### Changes to files
* **language_script.tsv**: one primary, rest secondary for these languages (others will be fixed in a later diff)
* **country_language_population.tsv**: most instances explicitly annotated (az_Latn, az_Arab, az_Cyrl) -- but sometimes it is not annotated because of lack of citation or declaration at source (so default kept).
* **TestSupplementalInfo.java**: added a TODO since soon we'll test for different constraints. Also the yue inequality constraint was weird -- why did we multiply by 10?
* **GenerateLikelySubtags.java**: I was able to remove one override since it is no longer necessary -- but some of the others are still necessary with the shape of the population data and since nan is official in Taiwan (where it would be Hant) but its not official in China -- where it is Hans, but also has so many more speakers, so nan_Hans should still take precedence.
* **supplementalData.xml** & **likelySubtags.xml**: all derived changes from above. 

### Likely Subtag changes
* **ku_???** -> was ku_Latn now ky_**Cyrl** for former soviet countries
* **mey_???** -> was mey_Latn now is mey_**Arab** in general, only ->_Latn for Senegal as it should be. tbh it's not always written but in reality when it is, it's usually written in the Arabic alphabet
* **nan_MO** -> was nan_Hans_MO now is nan_**Hant**_MO because of the dominance of Traditional Chinese still there
* **und_Latn_AM** -> stays **ku** but I had to add placeholder population data to get that.
* **und_Latn_GE** -> was ku now is **en** -- because the default of Kurdish in Georgia is now Cyrillic. tbd English adoption is picking up alot so its probably accurate, even though there probably are Latin-alphabet Kurdish users in Georgia.

### Changes per language
* **Azerbaijani** [az]: Latin primary (even if not most population, based on literacy + common usage in Azeribaijan).
   * Arab used in Iran, Iraq, Cyrl in Russia
* **Hakka** [hak]: Simplified Chinese primary, Traditional secondary
  * No notable changes
* **Northern Kurdish** [ku]: primary Latn
   * Cyrl used in former Soviet countries (this is also changing to Latn but there is a lack of recent data). Likely subtags changed to match.
   * Note I needed to add counts for how many people speak Russian in Armenia (and its alot!) because otherwise the automated script thought Kurdish was the biggest Cyrillic language in Armenia -- but it's definitely Russian there.
   * German Kurds probably just Latin alphabet, especially since they probably come from Turkey
* **Hassaniyya** [mey]: Primary Arabic
  * Secondary Latin ONLY in Senegal -- but that's the only place we had data, so I had to add in Ethnologue values for the other countries
  * This is the biggest likely subtag change because it really should be Arabic + Algeria not Latin + Senegal
* **Min-nan** [nan]: Simplified Han writing primary, Traditional secondary
  * AFAIK simplified in China (where the most speakers are) and traditional in outer Chinese territories
* **Panjabi** [pa]: Gurmukhi primary, Arabic secondary
  * While there are more Panjabi speakers in the world in Pakistan (where they would use Shahmuhki aka Arab letters), literacy is lower & most write in Urdu.
  * Outside of south Asia, most Panjabi speakers are Sikh immigrations, who use Gurmukhi
  * Gurmukhi also the default across
* **Cantonese/Yue** [yue]: Primary Traditional Han, secondary Simplified
  * Reflects the primary usage of Hong Kong and Macau, additionally Cantonese speakers are most likely to use Traditional writing regardless of where they live. Probably also in China (but actual numbers are hard to get)
* **Chinese** [zh]: Primary Simplified Han, secondary all others
  * This is really just a formalization of what the code was already doing.
  * Malaysia was oddly labeled with "Chinese (Traditional)" but using "zh" which was mapped to "Chinese (Simplified)". I checked and in recent times the Chinese community there has largely switched to Simplified (matching Singapore's switch too).

CLDR-18114

Generated data command:
```
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -1822,7 +1822,8 @@ XXX Code for transations where no currency is involved
 		<language type="myx" scripts="Latn"/>
 		<language type="myz" scripts="Mand"/>
 		<language type="mzn" scripts="Arab"/>
-		<language type="nan" scripts="Hans" />
+		<language type="na" scripts="Latn"/>
+		<language type="nan" scripts="Hans"/>
 		<language type="nan" scripts="Hant" alt="secondary"/>
 		<language type="nap" scripts="Latn"/>
 		<language type="naq" scripts="Latn"/>
@@ -2233,10 +2234,10 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="mk" populationPercent="0.47"/>	<!--Macedonian-->
 		</territory>
 		<territory type="AM" gdp="57730000000" literacyPercent="99.6" population="2976770">	<!--Armenia-->
-			<languagePopulation type="hy" populationPercent="99" officialStatus="official" references="R1170"/>	<!--Armenian-->
-			<languagePopulation type="ru" populationPercent="65" references="R1170"/>	<!--Russian-->
-			<languagePopulation type="ku" writingPercent="5" populationPercent="3.3" references="R1168"/>	<!--Kurdish-->
-			<languagePopulation type="ku_Cyrl" populationPercent="3.3" references="R1252"/>	<!--Kurdish (Cyrillic)-->
+			<languagePopulation type="hy" populationPercent="99" officialStatus="official" references="R1041"/>	<!--Armenian-->
+			<languagePopulation type="ru" populationPercent="65" references="R1041"/>	<!--Russian-->
+			<languagePopulation type="ku" writingPercent="5" populationPercent="3.3" references="R1346"/>	<!--Kurdish-->
+			<languagePopulation type="ku_Cyrl" populationPercent="3.3" references="R1376"/>	<!--Kurdish (Cyrillic)-->
 			<languagePopulation type="az" populationPercent="0" references="R1290"/>	<!--Azerbaijani-->
 		</territory>
 		<territory type="AO" gdp="265900000000" literacyPercent="70.4" population="37202100">	<!--Angola-->
@@ -2287,8 +2288,8 @@ XXX Code for transations where no currency is involved
 			<languagePopulation type="az" populationPercent="89" officialStatus="official" references="R1120"/>	<!--Azerbaijani-->
 			<languagePopulation type="az_Cyrl" populationPercent="9.9" officialStatus="official" references="R1121"/>	<!--Azerbaijani (Cyrillic)-->
 			<languagePopulation type="tly" populationPercent="9.8"/>	<!--Talysh-->
-			<languagePopulation type="ku" writingPercent="5" populationPercent="0.24" references="R1331"/>	<!--Kurdish-->
-			<languagePopulation type="ku_Cyrl" populationPercent="0.24" references="R1271"/>	<!--Kurdish (Cyrillic)-->
+			<languagePopulation type="ku" writingPercent="5" populationPercent="0.24" references="R1377"/>	<!--Kurdish-->
+			<languagePopulation type="ku_Cyrl" populationPercent="0.24" references="R1378"/>	<!--Kurdish (Cyrillic)-->
 			<languagePopulation type="ttt" populationPercent="0.22"/>	<!--Muslim Tat-->
 			<languagePopulation type="tkr" populationPercent="0.16"/>	<!--Tsakhur-->
 		</territory>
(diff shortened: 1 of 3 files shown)


macchiati:
I found the following a bit confusing:
> These languages had multiple primary scripts + different primary scripts in different countries. This changes it so they each have 1 primary script and countries are specifically annotated.

Do you mean that if a language xx is written in multiple scripts, that in the lang/pop data:

1. xx will always have a script subtag?
OR
2. xx will have a script subtag IFF the script = the likely script?

I'm guessing the first.

conradarcturus:
Thanks for the comments! Apologies for the delay in responding to them -- I'd love advice on setting up my email filters better since I seem to easily miss these comments in the noise.

I'll note there are many more subtasks to https://unicode-org.atlassian.net/browse/CLDR-18102 -- this particular task is focusing on the multiple primaries + changed across countries. There are separate tasks to resolve other languages with multiple primaries and other languages with multiple secondaries (no primary).

> > language_script.tsv: one primary, rest secondary for these languages (others will be fixed in a later diff)
> should have a test for this if this is to be enforced

I would, but the test will be bit convoluted because many languages still have multiple primaries and I needed to split this work up into multiple separate tasks otherwise there were too many changes to review. I originally updated the test but there were too many exceptions so I decided the test was not worth it until the work has progressed a bit more. I left a TODO linking to the master task in [TestSupplementalInfo.java](https://github.com/unicode-org/cldr/pull/4463/files#diff-36994b58edb9853f984f3f4b991c5a1acb4753a776a6c0b8adb7037e0b10292e) 

> would be good to get review of the changes particularly zh and zh-MY

Yes, I can highlight that change at the CLDR TC. The ticket has been up for review for a bit now. We don't always make it through the full agenda, I hope we can get to it this week.

I'll note the likely subtag does not change since only the label is changed. 

> Do you mean that if a language xx is written in multiple scripts, that in the lang/pop data: Malaysia's population data for Chinese was `zh`

> 1 xx will always have a script subtag?
> OR
> xx will have a script subtag IFF the script = the likely script?
> I'm guessing the first.

Correct, my aim is to make every language in `country_language_population.tsv` to have a script IF there are multiple possible options. Both for likely/primary scripts as well as secondary.

macchiati:
> Correct, my aim is to make every language in country_language_population.tsv to have a script IF there are multiple possible options. Both for likely/primary scripts as well as secondary.

I assume that does not include low-use scripts like Shavian for English, which would force all en to be en_Latn.

srl295:
> > Correct, my aim is to make every language in country_language_population.tsv to have a script IF there are multiple possible options. Both for likely/primary scripts as well as secondary.
> 
> I assume that does not include low-use scripts like Shavian for English, which would force all en to be en_Latn.

this may relate to https://unicode-org.atlassian.net/browse/CLDR-11224 where we have hardcoded data (in multiple places) on which languages are considered multi-script.

conradarcturus:
Good comments

> I assume that does not include low-use scripts like Shavian for English, which would force all en to be en_Latn

Great example. Yes, I don't think we should force it. I'm not sure a good rule of thumb is, like it would be good to enforce scripts for `mey` and `zh` but not good for `en`.

> this may relate to https://unicode-org.atlassian.net/browse/CLDR-11224 where we have hardcoded data (in multiple places) on which languages are considered multi-script.

Sounds good -- let's keep that task open and improve the test & validity rules in that task.

conradarcturus:
@srl295  @macchiati  do you have any requested changes for this particular PR? I would update the tests but there are many languages with multiple primaries & secondaries that I split by subtask. This was me focusing on 1 subset at a time (avoiding the need to review too many changes simultaneously).


Additionally, this PR is also failing a test that is seemingly unrelated, rather its related to the Maltese keyboard which was updated last month in https://github.com/unicode-org/cldr/pull/4360 .

Error from [kbd-check / keyboard-compile (push)](https://github.com/unicode-org/cldr/actions/runs/13957527248/job/39072172773?pr=4474):

mt-t-k0-47key.xml - error KM01001: Error validating LDML XML file: /keyboard3/conformsTo: enum: must be equal to one of the allowed values allowedValues="45"

srl295:
> .. Additionally, this PR is also failing a test that is seemingly unrelated, rather its related to the Maltese keyboard which was updated last month in #4360 .
> 
> Error from [kbd-check / keyboard-compile (push)](https://github.com/unicode-org/cldr/actions/runs/13957527248/job/39072172773?pr=4474):
> 
> mt-t-k0-47key.xml - error KM01001: Error validating LDML XML file: /keyboard3/conformsTo: enum: must be equal to one of the allowed values allowedValues="45"

unrelated. i've turned off that workflow. it's because there wasn't a compiler that supported that version at one time.

conradarcturus:
I added updates to the release notes

macchiati (dismissed):

srl295 (approved):

PR #4464, opened 2025-03-07
CLDR-16524 spec v47: change the Revision line to just link to Modifications

srl295:
CLDR-16524

I don't think this line needs a version number, the user can click on Modifications and see the entire history (starting with changes since the prev version).

Also, the current version is still in the "this version" URL if needed.

image

ALLOW_MANY_COMMITS=true


macchiati (approved):

macchiati:
I was confused at first because
https://www.unicode.org/reports/tr35/dev/tr35.html doesn't show the change
but
https://www.unicode.org/reports/tr35/47/tr35.html shows the change

(I thought those were always in lock-step??)

PR #4472, opened 2025-03-09
CLDR-18392 Create CLDR 48 release notes stub

AEApple:
CLDR-18392

Create stub release notes page for CLDR 48

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -1,0 +1,127 @@
+---
+title: CLDR 48 Release Note
+---
+
+# CLDR 48 Release Note
+
+| No. |    Date    | Rel. Note |  Data  |  Charts  | Spec |   Delta  | GitHub Tag | Delta DTD | CLDR JSON |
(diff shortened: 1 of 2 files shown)


macchiati:
I think we should preserve all the subheads from the previous version (eg also the Advance Warning)

AEApple:
Is this what you were envisioning?

macchiati:
Yes, looks good. Except on https://7aa8a327.cldr.pages.dev/downloads/cldr-48 we wouldn't have **V48** advance warnings, just v49

AEApple:
> Yes, looks good. Except on https://7aa8a327.cldr.pages.dev/downloads/cldr-48 we wouldn't have **V48** advance warnings, just v49

I was keeping them as advanced warnings until we make the change, then they get moved to the regular migration section. If the get delayed a release we move them to the v49 advanced warning section. That way we don't miss anything by accident that we had considered important from a migration perspective. SG?

PR #4474, opened 2025-03-10
CLDR-13440 Add language populations for Guatemala, Belize

conradarcturus:
This PR updates the locale demographics for Guatemala and Belize (and a bit for Honduras).

I found the data from the 2018 Guatemalan census -- added all of the languages (ahead of some organizations that are getting ready to contribute for those languages). Some languages also are spoken in Belize and Honduras -- to keep the likely subtags correct I needed to added data for those countries as well. Belize I just added the whole census. Honduras's census was harder to find so I ended up just following the data in Ethnologue.

Note: All new likely subtags are just upgrades -- the data comes from CLDR population data instead of SIL. None have functionally changed.

CLDR-13440

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4475, opened 2025-03-10
CLDR-18381 Add kek_GT locale

conradarcturus:
Adds a new locale, per CLDR-18381

There are some outstanding questions on orthography.

Also I'll update language's population in a separate diff updating the data from the full Guatemalan census: https://github.com/unicode-org/cldr/pull/4474

CLDR-18381

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -312,6 +312,7 @@ annotations.
 			<language type="kcg">Tyap</language>
 			<language type="kde">Makonde</language>
 			<language type="kea">Kabuverdianu</language>
+			<language type="kek">Qʼeqchiʼ</language>
 			<language type="ken">Kenyang</language>
 			<language type="kfo">Koro</language>
 			<language type="kg">Kongo</language>
--- a/common/main/kek.xml
+++ b/common/main/kek.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2025 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-3.0
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="kek"/>
+	</identity>
+	<localeDisplayNames>
+		<languages>
+			<language type="kek">Qʼeqchiʼ</language>
+		</languages>
+	</localeDisplayNames>
+	<layout>
+		<orientation>
+			<characterOrder>↑↑↑</characterOrder>
+			<lineOrder>↑↑↑</lineOrder>
+		</orientation>
+	</layout>
+	<characters>
+		<exemplarCharacters>[a {aa} {bʼ} c {ch} {ch’} e {ee} h i {ii} j k {k’} l m n o {oo} p q {q’} r s t {t’} {tz} {tz’} u {uu} w x y]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[d f g v z]</exemplarCharacters>
+		<exemplarCharacters type="numbers">↑↑↑</exemplarCharacters>
+		<exemplarCharacters type="punctuation">[“” ’ , . ! ? \: \- … ( ) ; \[ \] \{ \} /]</exemplarCharacters>
+	</characters>
+</ldml>
--- a/common/supplemental/attributeValueValidity.xml
+++ b/common/supplemental/attributeValueValidity.xml
@@ -74,7 +74,7 @@
 				 haw hnj ht
 				 ii io iu
 				 jbo jgo jmc
-				 kaa kab kaj kam kcg kde ken khq ki kkj kl kln kpe ksb ksf ksh kw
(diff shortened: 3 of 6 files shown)


common/main/en.xml
```diff
@@ -312,6 +312,7 @@ annotations.
 			<language type="kcg">Tyap</language>
 			<language type="kde">Makonde</language>
 			<language type="kea">Kabuverdianu</language>
+			<language type="kek">Qʼeqchiʼ</language>
```
srl295, line 315: we don't usually do this until later, when the language is at basic — right?

common/main/kek.xml
```diff
@@ -0,0 +1,30 @@
...
+		<orientation>
+			<characterOrder>↑↑↑</characterOrder>
+			<lineOrder>↑↑↑</lineOrder>
+		</orientation>
+	</layout>
+	<characters>
+		<exemplarCharacters>[a {aa} {bʼ} c {ch} {ch'} e {ee} h i {ii} j k {k’} l m n o {oo} p q {q’} r s t {t’} {tz} {tz’} u {uu} w x y]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[d f g v z]</exemplarCharacters>
+		<exemplarCharacters type="numbers">↑↑↑</exemplarCharacters>
+		<exemplarCharacters type="punctuation">[“” , . ! ? : - … ( ) ; [ ] { } /]</exemplarCharacters>
```
srl295, line 28:
```suggestion
		<exemplarCharacters type="punctuation">[“” , . ! ? : \- … ( ) ; \[ \] { } /]</exemplarCharacters>
```

regular `&quot;` was missing as was `\&amp;` and maybe others -seems a bit minimal

srl295 (dismissed):

common/main/en.xml
conradarcturus: You are thinking of adding it to the coverage levels.

When a locale is at Basic we add it to [coverageLevels.xml <coverageVariable key="%language80" ](https://github.com/unicode-org/cldr/blob/main/common/supplemental/coverageLevels.xml#L111). Which is [a requirement for Modern locale display names](https://github.com/unicode-org/cldr/blob/main/common/supplemental/coverageLevels.xml#L1040-L1041). Adding it here doesn't introduce any problem but introduces the opportunity to translate.

srl295: > Adding it here doesn't introduce any problem but introduces the opportunity to translate.

it is though usually a different commit from the core data.

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -197,8 +197,8 @@ public class DisplayAndInputProcessor {
     public static final Set<String> LANGUAGES_USING_MODIFIER_APOSTROPHE =
```
srl295, line 197: 😱 don't like these kinds of lists… in code…

common/main/kek.xml
```diff
@@ -0,0 +1,30 @@
...
+		<orientation>
+			<characterOrder>↑↑↑</characterOrder>
+			<lineOrder>↑↑↑</lineOrder>
+		</orientation>
+	</layout>
+	<characters>
+		<exemplarCharacters>[a {aa} {bʼ} c {ch} {ch’} e {ee} h i {ii} j k {k’} l m n o {oo} p q {q’} r s t {t’} {tz} {tz’} u {uu} w x y]</exemplarCharacters>
+		<exemplarCharacters type="auxiliary">[d f g v z]</exemplarCharacters>
+		<exemplarCharacters type="numbers">↑↑↑</exemplarCharacters>
+		<exemplarCharacters type="punctuation">[“” ’ , . ! ? \: \- … ( ) ; \[ \] { } /]</exemplarCharacters>
```
srl295, line 28: i think it's missing a backslash somewhere.

common/main/en.xml
srl295: I think our policy has been to record the English name but not to add it to en.xml at this juncture.

common/main/kek.xml
DavidLRowe: Need `\{` and `\}` for curly braces.

common/supplemental/attributeValueValidity.xml
```diff
@@ -74,7 +74,7 @@
 				 haw hnj ht
 				 ii io iu
 				 jbo jgo jmc
-				 kaa kab kaj kam kcg kde ken khq ki kkj kl kln kpe ksb ksf ksh kw
+				 kaa kab kaj kam kcg kde kek ken khq ki kkj kl kln kpe ksb ksf ksh kw
```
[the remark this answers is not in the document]
macchiati: We'd have to see what uses $languageNonTcLtBasic

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
macchiati: Agreed. We should drive this off of whether the language's exemplars include modifier apostrophe.

common/supplemental/attributeValueValidity.xml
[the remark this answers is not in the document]
conradarcturus: A test will fail if I don't update this. Every new locale needs to be added to this code.

common/main/kek.xml
conradarcturus: oh yay, that was exactly my error

srl295 (approved):

PR #4482, opened 2025-03-11
CLDR-18402 CLDRConfig to always set CLDR_DIR=. with a warning

srl295:
- .. if the current dir seems to be a CLDR dir
- we already do this for the .jar, but do this everywhere (such as from mvn:exec)

CLDR-18402

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRConfig.java
```diff
@@ -23,6 +23,7 @@
 import org.unicode.cldr.icu.dev.test.TestFmwk;
 import org.unicode.cldr.icu.dev.test.TestLog;
 import org.unicode.cldr.test.CheckCLDR.Phase;
+import org.unicode.cldr.test.SubmissionLocales;
```
srl295, line 26: this is a spotless change - SubmissionLocales is mentioned in a javadoc comment somewhere.

macchiati (approved):

PR #4483, opened 2025-03-11
CLDR-16527 v47 embrace markdown

srl295:
- auto generate additional <a id> target for 'simple' headings

CLDR-16527

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -31,7 +31,7 @@ For the latest version of the Unicode Standard see [[Unicode](https://www.unicod
 For more information see [About Unicode Technical Reports](https://www.unicode.org/reports/about-reports.html) and the [Specifications FAQ](https://www.unicode.org/faq/specifications.html).
 Unicode Technical Reports are governed by the Unicode [Terms of Use](https://www.unicode.org/copyright.html)._
 
-## <a name="Parts" href="#Parts">Parts</a>
+## Parts
 
 The LDML specification is divided into the following parts:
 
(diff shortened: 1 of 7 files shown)


docs/ldml/tr35-dates.md
```diff
@@ -31,7 +31,7 @@ For the latest version of the Unicode Standard see [[Unicode](https://www.unicod
 For more information see [About Unicode Technical Reports](https://www.unicode.org/reports/about-reports.html) and the [Specifications FAQ](https://www.unicode.org/faq/specifications.html).
 Unicode Technical Reports are governed by the Unicode [Terms of Use](https://www.unicode.org/copyright.html)._
 
-## <a name="Parts" href="#Parts">Parts</a>
+## Parts
```
srl295, line 34: @macchiati this removes "simple" cases and has the generator add a `<a id="Parts"/>` in the HTML…

docs/ldml/tr35-dates.md
```diff
@@ -79,7 +79,7 @@ The LDML specification is divided into the following parts:
   * Table: [timeZoneNames Elements Used for Fallback](#timeZoneNames_Elements_Used_for_Fallback)
   * [Metazone Names](#Metazone_Names)
 * [Supplemental Time Zone Data](#Supplemental_Time_Zone_Data)
-  * [Metazones](#Metazones)
+  * [Metazones](#metazones)
```
srl295, line 82: @macchiati I guess the downside here is that `(#Metazones)` no longer works _from within the .md file_

macchiati: It seems like we should have the "canonical" anchor match whatever github wants, to enable the markdown links, but then add other forms for compatibility.

srl295:
This would break internal links.  I think let's go with a simpler approach in #4484 for now.

Moving the `<a name` out to the prior line broke the html generation.

PR #4484, opened 2025-03-11
CLDR-16527 v47 drop unneeded href

srl295:
- we have [Anchor.js](https://www.bryanbraun.com/anchorjs/) so we do not need hrefs on anchors. They will be auto-linked in HTML.

Before: 

```md
### <a name="CLDR_Collation_Algorithm" href="#CLDR_Collation_Algorithm">CLDR Collation Algorithm</a>
```

After:

```md
### <a name="CLDR_Collation_Algorithm">CLDR Collation Algorithm</a>
```

CLDR-16527

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-collation.md
+++ b/docs/ldml/tr35-collation.md
@@ -31,7 +31,7 @@ For the latest version of the Unicode Standard see [[Unicode](https://www.unicod
 For more information see [About Unicode Technical Reports](https://www.unicode.org/reports/about-reports.html) and the [Specifications FAQ](https://www.unicode.org/faq/specifications.html).
 Unicode Technical Reports are governed by the Unicode [Terms of Use](https://www.unicode.org/copyright.html)._
 
-## <a name="Parts" href="#Parts">Parts</a>
+## <a name="Parts">Parts</a>
 
 The LDML specification is divided into the following parts:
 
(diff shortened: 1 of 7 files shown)


macchiati:
It sounds like this is just a fix for 47; that when processing 48 we can automatically generate from "CLDR Collation Algorithm" both "cldr-collation-algorithm" and "CLDR_Collation_Algorithm".

If that is so, is there any benefit to doing this partial step now?

srl295:
> It sounds like this is just a fix for 47; that when processing 48 we can automatically generate from "CLDR Collation Algorithm" both "cldr-collation-algorithm" and "CLDR_Collation_Algorithm".
> 

Different issue. 

This PR is about the self referential href= that add no value. It just removes them because the anchor.js tool puts self links in everywhere. 

> 
> If that is so, is there any benefit to doing this partial step now?

It's not just for 47 but it would need to be cherry picked. 

Maybe better to just shift this all to 48

macchiati:
If it is removing something that has no value, but the removal doesn't _increase_ value, then let's not disturb 47 at this point, and target 48 instead.

The phrase "anchor.js tool puts self links in everywhere" means that everywhere we have name=X or id=X, we don't need href="#X"; I assume that's what you mean.

PR #4485, opened 2025-03-11
CLDR-16527 v47: fix broken links in modification section

srl295:
CLDR-16527

broke in #4409 

ALLOW_MANY_COMMITS=true


srl295:
Also the ToC generator is failing on headers that are links, I filed https://github.com/not-dalia/gfm-toc/issues/6 but we might want to consider avoiding those.

AEApple (approved):

PR #4498, opened 2025-03-13
CLDR-18269 Restore timezone filtering

robertbastian:
CLDR-18269

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
```diff
@@ -5349,42 +5349,12 @@ public Set<String> getUnitPrefixes() {
...
-                                                                            x, shortId, info));
-                                                            return !info.deprecated;
-                                                        })
-                                                .collect(Collectors.toUnmodifiableSet());
-                            }
-                            return result;
+                            return availableLongTz.stream()
+                                    .filter(
+                                            x -> {
+                                                return zone_territory.get(x) != null;
```
macchiati, line 5355: I can understand dropping the hack. So that I can understand what is going on, can you explain why the old code had
```
return !info.deprecated;
```
and the new code just has 
```
return zone_territory.get(x) != null;
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCLDRFile.java
```diff
@@ -349,6 +349,7 @@ public void testExtraPaths() {
                                                     && (path.contains("[@type=\"energy-foodcalorie")
                                                             || path.contains(
                                                                     "[@type=\"graphics-dot")))))
+                            || (path.startsWith("//ldml/dates/timeZoneNames/zone") && path.endsWith("exemplarCity"))
```
macchiati, line 352: Again, can you explain the change?

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java
robertbastian: The new code tries to check whether an ID is canonical (https://github.com/unicode-org/cldr/blob/10e099457512881f27d57785bcfadf4230b913af/tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalDataInfo.java#L3470). A canonical ID is never deprecated.

PR #4505, opened 2025-03-16
CLDR-18391 Add downloads page filters

AEApple:
CLDR-18391

Add filters on downloads page which need to be updated every release.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/cldr-big-red-switch.md
+++ b/docs/site/development/cldr-big-red-switch.md
@@ -29,6 +29,11 @@ TODO: List all major types of tasks with their definitions and link to pages wit
 
 ### Release tracking
 
+#### Downloads page
+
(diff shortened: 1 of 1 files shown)


AEApple:
@unicode-org/cldr-tc - I wonder if we should:

1. Update the saved searches for each release to include the status 'Fixed in Survey Tool' for any issues that were resolved via changes made in the Survey Tool that release which had originally been requested by a Jira ticket.
2. Update the dev-tickets filter to also include any tickets which are 'In Review' or 'Review Feedback' statuses since the work on those tickets are already done and are expected to be included in the upcoming release that is currently in development.

WDYT?

macchiati:
Both sound good to me

stenshamn (changes requested):
Fairly stylistic suggestions, if you disagree I'm happy to approve as is.

stenshamn (approved):

PR #4506, opened 2025-03-18
CLDR-18356 deps unvendor org.json package

srl295:
- replace with maven version
- note also CLDR-5839 for removing this use altogether

CLDR-18356

- [x] will need FE testing

ALLOW_MANY_COMMITS=true


markusicu (approved):
thank you!

srl295:
I've confirmed that ST works with this change. (logging in, voting ,etc.) 

org.json is used for 'older' parts such as SurveyAjax, status, examples, etc, but all seems to be working.

btangmu (approved):
Wow, what a great clean-up!

PR #4515, opened 2025-03-20
CLDR-18391 Add remaining filters from BRS sheet

AEApple:
CLDR-18391

These are the remaining filters that were in the BRS sheet.

ALLOW_MANY_COMMITS=true


AEApple:
Two of these filters seem like they may be duplicates. We may want to consider whether we only want to keep one of them.

macchiati (approved):

PR #4518, opened 2025-03-20
CLDR-18215 Suppress coverage of en/ja-only units

macchiati:
CLDR-18215

See ticket for details.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -167,9 +167,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 		
 		<coverageVariable key="%unitDurationTypes" value="duration-(year|month|week|day|hour|minute|second|millisecond)"/>
 		<coverageVariable key="%unitLengths" value="(long|short|narrow)"/>
-		<coverageVariable key="%unitsEnglish" value="(length-fathom|length-furlong|mass-stone|volume-bushel|energy-british-thermal-unit)"/>
-		<coverageVariable key="%unitsOther" value="(length-(picometer|light-year)|pressure-(hectopascal|inch-ofhg|millibar)|acceleration-g-force|angle-(degree|minute|second)|area-(acre|hectare|square-(foot|kilometer|meter|mile))|power-(horsepower|kilowatt|watt)|speed-meter-per-second|volume-cubic-(mile|kilometer))"/>
-		<coverageVariable key="%unitBeaufortRegions" value="(CN|DE|GB|GR|HK|MO|MT|NL|TW|US)"/>
+		<coverageVariable key="%unitsEnglish" value="(length-fathom|length-furlong|mass-stone|volume-bushel|energy-british-thermal-unit|volume-fluid-ounce-metric|volume-pint-imperial|volume-cup-imperial|temperature-rankine|mass-slug|pressure-gasoline-energy-density|energy-british-thermal-unit-it|length-rod|length-chain|duration-fortnight)"/>
+		<coverageVariable key="%unitsJapanese" value="(area-bu-jp|area-cho|area-se-jp|length-jo-jp|length-ken|length-ri-jp|length-rin|length-shaku-cloth|length-shaku-length|length-sun|mass-fun|volume-cup-jp|volume-koku|volume-kosaji|volume-osaji|volume-sai|volume-shaku|volume-to-jp)"/>
 
 		<coverageVariable key="%variantTypes" value="([A-Z0-9]++)"/>
 		<coverageVariable key="%wideAbbr" value="(wide|abbreviated)"/>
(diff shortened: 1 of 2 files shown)


common/supplemental/coverageLevels.xml
```diff
@@ -1245,37 +1244,53 @@ For terms of use, see http://www.unicode.org/copyright.html
...
+		
+		<coverageLevel value="comprehensive" match="units/unitLength[@type='%unitLengths']/unit[@type='%unitsJapanese']/unitPattern[@count='%anyAttribute'][@case='%anyAttribute']"/>
+		<coverageLevel value="comprehensive" match="units/unitLength[@type='%unitLengths']/unit[@type='%unitsJapanese']/unitPattern[@count='%anyAttribute']"/>
+		<coverageLevel value="comprehensive" match="units/unitLength[@type='%unitLengths']/unit[@type='%unitsJapanese']/displayName"/>
+		<coverageLevel value="comprehensive" match="units/unitLength[@type='%unitLengths']/unit[@type='%unitsJapanese']/gender"/>
+		<coverageLevel value="comprehensive" match="units/unitLength[@type='%unitLengths']/unit[@type='%unitsJapanese']/perUnitPattern"/>
+		
+		<!-- Other Units, all locales -->
 
 		<coverageLevel value="modern" match="units/unitLength[@type='%unitLengths']/unit[@type='%anyAttribute']/unitPattern[@count='%anyAttribute'][@case='%anyAttribute']"/>
```
conradarcturus, line 1277: I think this code will make any non-specified units (like Beaufort) now required for modern for all languages

common/supplemental/coverageLevels.xml
```diff
@@ -167,9 +167,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 		
 		<coverageVariable key="%unitDurationTypes" value="duration-(year|month|week|day|hour|minute|second|millisecond)"/>
 		<coverageVariable key="%unitLengths" value="(long|short|narrow)"/>
-		<coverageVariable key="%unitsEnglish" value="(length-fathom|length-furlong|mass-stone|volume-bushel|energy-british-thermal-unit)"/>
-		<coverageVariable key="%unitsOther" value="(length-(picometer|light-year)|pressure-(hectopascal|inch-ofhg|millibar)|acceleration-g-force|angle-(degree|minute|second)|area-(acre|hectare|square-(foot|kilometer|meter|mile))|power-(horsepower|kilowatt|watt)|speed-meter-per-second|volume-cubic-(mile|kilometer))"/>
-		<coverageVariable key="%unitBeaufortRegions" value="(CN|DE|GB|GR|HK|MO|MT|NL|TW|US)"/>
+		<coverageVariable key="%unitsEnglish" value="(length-fathom|length-furlong|mass-stone|volume-bushel|energy-british-thermal-unit|volume-fluid-ounce-metric|volume-pint-imperial|volume-cup-imperial|temperature-rankine|mass-slug|pressure-gasoline-energy-density|energy-british-thermal-unit-it|length-rod|length-chain|duration-fortnight)"/>
```
conradarcturus, line 170: Do you intend to add speed-beaufort here since you removed the other Beaufort code?

conradarcturus (dismissed):
Changes mostly look good to me, just wondering if speed-beaufort was meant to move to modern or not.

Also it looks like there's a test failure with `volume-fluid-ounce-metric`.

common/supplemental/coverageLevels.xml
macchiati: Right, that's what I proposed to the committee. Beaufort is not only an English measure, by the way. According to Wikipedia, it is used in "Netherlands, Germany, Greece, China, Taiwan, Hong Kong, Malta, and Macau", but that list is not complete: apparently it is used more widely in shipping. So because it is fairly prevalent in shipping, like knot for shipping or foot for aviation elevation, it is simplest to go ahead and get it in all languages.

common/supplemental/coverageLevels.xml
macchiati: The %unitsEnglish is really "english only". If we put Beaufort in this group, then it wouldn't appear for "Netherlands, Germany, Greece, China, Taiwan, Hong Kong, Malta, ..." (see above).

Some English measures have wide enough usage (eg inch for monitor screen size) that they are worth translating generally, but these are pretty narrow.

macchiati:
I'll squash later to fix the commit message

macchiati:
Pinging for review

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUnits.java
```diff
@@ -4593,12 +4593,47 @@ public void testCoverage() {
...
+                        "sun",
+                        "fun",
+                        "slug",
+                        "stone",
+                        "gasoline-energy-density",
+                        "rankine",
+                        "bushel",
+                        "cup-imperial",
+                        "cup-jp",
+                        // "fluid-ounce-metric",
```
btangmu, line 4623: "fluid-ounce-metric" as a comment?

btangmu (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestUnits.java
macchiati: It was dropped from that list. Again, left in case we have to restore.

PR #4519, opened 2025-03-21
Revise era codes based on March 20 discussion

sffc:
CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -4535,8 +4535,8 @@ XXX Code for transations where no currency is involved
         <calendar type="gregorian">
            <calendarSystem type="solar" />
            <eras>
-               <era type="0" end="0-12-31" code="gregory-inverse" aliases="bc bce"/>
-               <era type="1" start="1-01-01" code="gregory" aliases="ad ce"/>
+               <era type="0" end="0-12-31" code="bce" aliases="bc"/>
+               <era type="1" start="1-01-01" code="ce" aliases="ad" primary="true"/>
            </eras>
         </calendar>
         <calendar type="generic"/>
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
         <calendar type="islamic">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15" code="islamic" aliases="ah"/>
+              <era type="0" start="622-7-15" code="ah"/> <!-- Anno Hegirae -->
           </eras>
         </calendar>
         <calendar type="islamic-civil">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-16" code="islamic-civil" aliases="islamicc ah"/>
+              <era type="0" start="622-7-16" code="ah"/> <!-- Anno Hegirae -->
           </eras>
         </calendar>
         <calendar type="islamic-rgsa">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15" code="islamic-rgsa" aliases="ah"/>
+              <era type="0" start="622-7-15" code="ah"/> <!-- Anno Hegirae -->
           </eras>
         </calendar>
         <calendar type="islamic-tbla">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15" code="islamic-tbla" aliases="ah"/>
+              <era type="0" start="622-7-15" code="ah"/> <!-- Anno Hegirae -->
           </eras>
         </calendar>
         <calendar type="islamic-umalqura">
           <calendarSystem type="lunar" />
           <eras>
-              <era type="0" start="622-7-15" code="islamic-umalqura" aliases="ah"/>
+              <era type="0" start="622-7-15" code="ah"/> <!-- Anno Hegirae -->
           </eras>
         </calendar>
         <calendar type="chinese">
           <calendarSystem type="lunisolar"/>
           <eras>
-              <era type="0" start="-2636-01-01" code="chinese"/> <!-- 'sequential' year -->
+              <era type="0" start="-2636-01-01"/> <!-- 'sequential' year -->
           </eras>
         </calendar>
         <calendar type="hebrew">
           <calendarSystem type="lunisolar"/>
           <eras>
-              <era type="0" start="-3760-10-7" code="hebrew" aliases="am"/>
+              <era type="0" start="-3760-10-7" code="am"/> <!-- Anno Mundi -->
           </eras>
         </calendar>
         <calendar type="buddhist">
           <calendarSystem type="solar" />
           <eras>
-              <era type="0" start="-542-01-01" code="buddhist" aliases="be"/>
+              <era type="0" start="-542-01-01" code="be"/> <!-- Buddhist Era -->
           </eras>
         </calendar>
         <calendar type="coptic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="284-08-28" code="coptic-inverse"/>
-              <era type="1" start="284-08-29" code="coptic"/>
+              <era type="0" end="284-08-28" code="??? TODO ???"/>
+              <era type="1" start="284-08-29" code="am" primary="true"/> <!-- Anno Martyrum -->
           </eras>
         </calendar>
         <calendar type="persian">
           <calendarSystem type="solar"/>
           <eras>
-              <era type="0" start="622-01-01" code="persian" aliases="ap"/>
+              <era type="0" start="622-01-01" code="ap" aliases="sh hs"/> <!-- Anno Persico -->
           </eras>
         </calendar>
         <calendar type="dangi">
           <calendarSystem type="lunisolar"/>
           <eras>
-              <era type="0" start="-2332-01-01" code="dangi"/> <!-- 'sequential' year -->
+              <era type="0" start="-2332-01-01"/> <!-- 'sequential' year -->
           </eras>
         </calendar>
         <calendar type="ethiopic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="8-08-28" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
-              <era type="1" start="8-08-29" code="ethiopic" aliases="incar"/>
+              <era type="0" end="8-08-28" code="aa" aliases="mundi"/> <!-- Amete Alem -->
+              <era type="1" start="8-08-29" code="am" aliases="incar" primary="true"/> <!-- Amete Mihret -->
           </eras>
         </calendar>
         <calendar type="ethiopic-amete-alem">
           <eras>
-              <era type="0" end="-5492-08-29" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
+              <era type="0" end="-5492-08-29" code="aa" aliases="mundi"/> <!-- Amete Alem -->
           </eras>
         </calendar>
         <calendar type="indian">
           <eras>
-              <era type="0" start="79-01-01" code="indian" aliases="saka"/>
+              <era type="0" start="79-01-01" code="saka"/>
           </eras>
         </calendar>
         <calendar type="roc">
           <eras>
               <era type="0" end="1911-12-31" code="roc-inverse" aliases="before-roc"/>
-              <era type="1" start="1912-01-01" code="roc" aliases="minguo"/>
+              <era type="1" start="1912-01-01" code="roc" aliases="minguo" primary="true"/>
           </eras>
         </calendar>
 


common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
+              <era type="0" start="-542-01-01" code="be"/> <!-- Buddhist Era -->
           </eras>
         </calendar>
         <calendar type="coptic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="284-08-28" code="coptic-inverse"/>
-              <era type="1" start="284-08-29" code="coptic"/>
+              <era type="0" end="284-08-28" code="??? TODO ???"/>
+              <era type="1" start="284-08-29" code="am" primary="true"/> <!-- Anno Martyrum -->
```
gibson042, line 4839: There's a suggestion at https://github.com/tc39/proposal-intl-era-monthcode/issues/25#issuecomment-2743063952:
> the Coptic calendar should have eras BD and AD, not BM and AM

However, it's at least partly at odds with other resources I can find, all of which use "AM" or "A.M.":
* https://suscopts.org/coptic-orthodox/fasts-and-feasts/
* https://www.timeanddate.com/calendar/coptic-calendar.html
* https://en.orthodoxwiki.org/Coptic_Calendar#The_Coptic_year
* https://www.stvnashville.org/coptic-new-year#yui_3_17_2_1_1742572291617_643

I do see _references_ to A.D. (from "anno Diocletiani"), but concrete usage seems rare. I have no opinion on which is more correct here.

gibson042 (approved):

common/supplemental/supplementalData.xml
```diff
@@ -4535,8 +4535,8 @@ XXX Code for transations where no currency is involved
         <calendar type="gregorian">
            <calendarSystem type="solar" />
            <eras>
-               <era type="0" end="0-12-31" code="gregory-inverse" aliases="bc bce"/>
-               <era type="1" start="1-01-01" code="gregory" aliases="ad ce"/>
+               <era type="0" end="0-12-31" code="bce" aliases="bc"/> <!-- Before Common Era, Before Christ -->
+               <era type="1" start="1-01-01" code="ce" aliases="ad" primary="true"/> <!-- Common Era, Anno Domini -->
```
robertbastian, line 4539: what does `primary` mean, and when is it ever not the last era?

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
+              <era type="0" start="-2332-01-01"/> <!-- 'sequential' year -->
           </eras>
         </calendar>
         <calendar type="ethiopic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="8-08-28" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
-              <era type="1" start="8-08-29" code="ethiopic" aliases="incar"/>
+              <era type="0" end="8-08-28" code="aa" aliases="mundi"/> <!-- Amete Alem -->
+              <era type="1" start="8-08-29" code="am" aliases="incar" primary="true"/> <!-- Amete Mihret -->
```
robertbastian, line 4858: consider removing the aliases, they're our invention afaict

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
           <eras>
-              <era type="0" start="79-01-01" code="indian" aliases="saka"/>
+              <era type="0" start="79-01-01" code="saka"/>
           </eras>
         </calendar>
         <calendar type="roc">
           <eras>
-              <era type="0" end="1911-12-31" code="roc-inverse" aliases="before-roc"/>
-              <era type="1" start="1912-01-01" code="roc" aliases="minguo"/>
+              <era type="0" end="1911-12-31" code="minguo-qian" aliases="broc before-roc"/>
```
robertbastian, line 4873:
```suggestion
              <era type="0" end="1911-12-31" code="minguo-qian"/>
```

common/supplemental/supplementalData.xml
macchiati: Our invention or not, if an ID has been in any release, it needs to be deprecated but maintained as an alias.

robertbastian: In theory yes, but if neither ICU, nor ICU4X, nor Temporal, nor any implementation we know of has used these, can we be a bit flexible?

common/supplemental/supplementalData.xml
sffc: I haven't written the spec yet but `primary` means "the era used for extended year"; it is `ce` in Japanese which is not the last era

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
-              <era type="0" end="8-08-28" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
-              <era type="1" start="8-08-29" code="ethiopic" aliases="incar"/>
+              <era type="0" end="8-08-28" code="aa" aliases="mundi"/> <!-- Amete Alem -->
+              <era type="1" start="8-08-29" code="am" aliases="incar" primary="true"/> <!-- Amete Mihret -->
           </eras>
         </calendar>
         <calendar type="ethiopic-amete-alem">
           <eras>
-              <era type="0" end="-5492-08-29" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
+              <era type="0" end="-5492-08-29" code="aa" aliases="mundi"/> <!-- Amete Alem -->
```
robertbastian, line 4863:
```suggestion
              <era type="0" start="-5492-08-29" code="aa" aliases="mundi"/> <!-- Amete Alem -->
```

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
-              <era type="0" start="-2332-01-01" code="dangi"/> <!-- 'sequential' year -->
+              <era type="0" start="-2332-01-01"/> <!-- 'sequential' year -->
           </eras>
         </calendar>
         <calendar type="ethiopic">
           <calendarSystem type="other"/>
           <eras>
-              <era type="0" end="8-08-28" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
-              <era type="1" start="8-08-29" code="ethiopic" aliases="incar"/>
+              <era type="0" end="8-08-28" code="aa" aliases="mundi"/> <!-- Amete Alem -->
```
robertbastian, line 4857:
```suggestion
              <era type="0" start="-5492-08-29" code="aa" aliases="mundi"/> <!-- Amete Alem -->
```

common/supplemental/supplementalData.xml
robertbastian, line 4858: I think the Amete Alem era should be primary, because then both ethiopic calendars have the same extended year.

common/supplemental/supplementalData.xml
robertbastian: makes sense.

common/supplemental/supplementalData.xml
sffc: Hmm interesting thought

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
         <calendar type="islamic">
           <calendarSystem type="lunar" />
           <eras>
```
robertbastian, line 4789: there is some evidence for a BH era: https://web.archive.org/web/20170202005514/http://www.sharjah.ae/registration-locations.aspx?Val=111#:~:text=A%20limited%20number%20of%20years%20before%20Hijra%20(BH)%20are%20used%20to%20date%20events%20related%20to%20Islam%2C%20such%20as%20the%20birth%20of%20Prophet%20Mohammad%20in%2053%20BH.

sffc: Yes I definitely agree that we should consider adding a `bh` era. https://github.com/unicode-org/icu4x/issues/5068#issuecomment-2741915472

I consider _changing_ the eras out-of-scope for the current PR, but I'm happy to scope-creep if we have consensus and it won't block the PR.

sffc:
@macchiati any comments on this before you disappear on vacation?

common/supplemental/supplementalData.xml
sffc: I'd like to discuss this question in the CLDR Design WG meeting. As a general principle I agree with @macchiati, but I don't have evidence that these codes are used _anywhere_, so I might prefer making an exception in this one case.

common/supplemental/supplementalData.xml
sffc: https://unicode-org.atlassian.net/browse/CLDR-18464

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
         <calendar type="ethiopic-amete-alem">
           <eras>
-              <era type="0" end="-5492-08-29" code="ethioaa" aliases="ethiopic-amete-alem mundi"/>
+              <era type="0" end="-5492-08-29" code="aa" aliases="mundi"/> <!-- Amete Alem -->
           </eras>
         </calendar>
         <calendar type="indian">
           <eras>
-              <era type="0" start="79-01-01" code="indian" aliases="saka"/>
+              <era type="0" start="79-01-01" code="saka"/>
```
robertbastian, line 4868: given that we cannot use the correct spelling "Śaka", maybe this should be "shaka".

sffc: https://github.com/tc39/proposal-intl-era-monthcode/issues/31

common/supplemental/supplementalData.xml
sffc: https://github.com/tc39/proposal-intl-era-monthcode/issues/30

pedberg-icu: > I don't have evidence that these codes are used _anywhere_, so I might prefer making an exception in this one case.

Well, it is really the opposite question. Unless we have proof they are they are *not* used anywhere I think we should keep the old codes as aliases. We do not necessarily know who all of the clients of CLDR data are and how they use it.

sffc: We discussed this in the 2025-03-31 CLDR Design WG meeting and the conclusion from that meeting (based on the people who happened to be in attendance that day) was that given how news these codes are, and given that ICU does not use them, and Temporal doesn't use them yet, and that they were added _for Temporal_, they are new enough that we can just drop the codes currently in the XML.

But I do not want to block this PR on that discussion so if adding the existing codes as aliases is going to move this forward, I'll do that

common/supplemental/supplementalData.xml
```diff
@@ -4787,91 +4787,91 @@ XXX Code for transations where no currency is involved
...
           </eras>
         </calendar>
         <calendar type="roc">
           <eras>
-              <era type="0" end="1911-12-31" code="roc-inverse" aliases="before-roc"/>
-              <era type="1" start="1912-01-01" code="roc" aliases="minguo"/>
+              <era type="0" end="1911-12-31" code="minguo-qian" aliases="broc before-roc"/>
+              <era type="1" start="1912-01-01" code="minguo" primary="true"/>
```
robertbastian, line 4874: this page https://en.wikipedia.org/wiki/Chinese_calendar_correspondence_table uses "AM", presumably that means "Anno Minguoae" (pardon my Latin).

sffc: oh my

sffc:
I recall @macchiati stating that he was concerned about removing the CLDR 43 codes from the file and preferred them as aliases, though the CLDR Design WG seemed okay with that path on 2025-03-31:

> Shane: Is it a hard requirement to retain the old codes (added in CLDR 43, [CLDR-16417](https://unicode-org.atlassian.net/browse/CLDR-16417)), or can we just discard them? No evidence of usage.
>
> Rich: If Temporal is the only user, and they want to change it, I'm fine changing it
>
> Markus: If the person proposing this change wants to change it, and they're not aware of other users, it seems fine to change it

I don't have any written documentation of Mark's position and whether it is required for landing this PR or otherwise changes the Design WG opinion. Could you clarify?

[CLDR-16417]: https://unicode-org.atlassian.net/browse/CLDR-16417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sffc:
CLDR-TC approved this with the following notes in https://unicode-org.atlassian.net/browse/CLDR-18369

> - Temporal was the original requester for the string era names, so we decided to align with Temporal (we don’t know of any adoption beyond ICU4X)
> - Need a note in the migration section of the CLDR 48 release notes

I need to write a migration note for both this and https://github.com/unicode-org/cldr/pull/4620. I intend to merge this PR to update the data and unblock integrations, but the ticket should not be closed until the migration guide is updated (which I plan to do in a separate PR).

sffc:
The CLDR-TC approved this PR for alignment with Temporal, so I am going to finally hit the merge button.

common/supplemental/supplementalData.xml
```diff
@@ -4304,8 +4304,8 @@ XXX Code for transations where no currency is involved
         <calendar type="gregorian">
            <calendarSystem type="solar" />
            <eras>
-               <era type="0" end="0-12-31" code="gregory-inverse" aliases="bc bce"/>
-               <era type="1" start="1-01-01" code="gregory" aliases="ad ce"/>
+               <era type="0" end="0-12-31" code="bce" aliases="bc"/> <!-- Before Common Era, Before Christ -->
```
justingrant, line 4307: @sffc why is `type: 0` used for the backwards-counting, earlier era here, but is used for the forwards-counting, later era in most (all?) calendars below?  Are 0 vs. 1 meaningful in any way?

robertbastian: They are generally chronological with the exception of AH/BH, which had a whole discussion in https://github.com/unicode-org/cldr/pull/4581

sffc:
Release notes in #4952

bmkkakabmkkaka0-commits:
> @macchiati any comments on this before you disappear on vacation? (I see you already left one regarding ID stability)

No

PR #4520, opened 2025-03-21
CLDR-17570 Rational formatting

macchiati:
CLDR-17570

Add new structure for rationals; put them with the rest of the number formats (decimal, scientific, etc.).
This involved various other changes:
- root and en and dtd
- Placeholder additions
- DAIP additions
- ExampleGenerator additions
- Note that often new structure needs Checks, but this uses DAIP for two cases that would otherwise be errors.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
@@ -13,6 +13,7 @@ import java.util.TreeSet;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import org.unicode.cldr.util.RegexLookup.Finder;
 
 public class PathDescription {
     /** Remember to quote any [ character! */
@@ -258,7 +259,7 @@ public class PathDescription {
                     + ".\n"
 
                     /*
-                     * Warning: the longer match must come first
+                     * Warning: the longer match and more specific match must come first!
                      */
                     + "^//ldml/units/unitLength\\[@type=\"([^\"]*)\"]/compoundUnit\\[@type=\"([^\"]*)\"]/compoundUnitPattern1"
                     + RegexLookup.SEPARATOR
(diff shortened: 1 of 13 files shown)


macchiati:
PS Filed a ticket for improvements to the tooling, to make it easier to add structure.

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
```diff
@@ -510,10 +511,45 @@ public class PathDescription {
...
+                    + "^//ldml/numbers/rationalFormats/rationalPattern"
                     + RegexLookup.SEPARATOR
-                    + "Special pattern used to compose {1} numbers. Note: before translating, be sure to read "
-                    + CLDRURLS.NUMBER_PATTERNS
+                    + "A pattern that is used to format a rational fraction (eg, ½), using the numerator and denominator"
+                    + CLDRURLS.NUMBERS_HELP
+                    + ".\n"
+                    + "^//ldml/numbers/rationalFormats/integerAndRationalPattern\\[@alt=\"([^\"]*)\"]"
+                    + RegexLookup.SEPARATOR
+                    + "A pattern that is used to “glue” an integer and a formatted rational fraction (eg, ½) together; only used when the rational fraction does not start with an un-superscripted digit"
```
btangmu, line 521: "does not start with an un-superscripted digit" -- meaning (1) it does start with a superscripted digit, or (2) it starts with a non-digit? If it starts with a non-digit, would it be something like "one/half"? Just curious...

btangmu (approved):
This makes my head spin. +1 ½ for extensive tests

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
macchiati: Formally https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7Bgc%3DNd%7D, we'll need to explain it. Once we have markdown in examples, it would be easy to link out to an explanation

PR #4523, opened 2025-03-24
CLDR-18215 Add examples for light-second

macchiati:
And so on.

CLDR-18215

Aside from adding to the example generate handler for unit formats:

- Adds convenience constructor for example generator
- Refactors a bit

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
- adds an easy way to override a CLDRFile for testing

- deletes a hack, because we can now format all parts

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
- adds a test case.

ALLOW_MANY_COMMITS=true


macchiati:
I noticed a glitch (https://unicode-org.atlassian.net/issues/CLDR-18449?filter=10033) which I believe is unconnected, where the ExampleGenerator doesn't change an example when the value changes. Not a showstopper for submission, but something we should look at.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1362,8 +1375,33 @@ private void handleFormatUnit(XPathParts parts, String unitPattern, List<String>
         if (parts.getElement(-1).equals("unitPattern")) {
             String count = parts.getAttributeValue(-1, "count");
             DecimalQuantity amount = getBest(Count.valueOf(count));
+
             if (amount != null) {
-                addFormattedUnits(examples, parts, unitPattern, shortUnitId, amount);
+                if (shortUnitId.equals("light-speed")) {
```
btangmu, line 1380: Ideally this added code should go in a subroutine. It should be possible to view an entire function on the screen without scrolling.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
```diff
@@ -71,12 +71,12 @@ public class UnitConverter implements Freezable<UnitConverter> {
...
-                    "dot-per-centimeter",
-                    "dot-per-inch",
+                    "dot-per-centimeter", "dot-per-inch"
                     // skip because a component is not translated
-                    "liter-per-100-kilometer",
-                    "millimeter-ofhg",
-                    "inch-ofhg");
+                    //                    "liter-per-100-kilometer",
+                    //                    "millimeter-ofhg",
+                    //                    "inch-ofhg"
```
btangmu, line 78: Is there a reason to keep the commented-out code?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1362,8 +1375,33 @@ private void handleFormatUnit(XPathParts parts, String unitPattern, List<String>
...
+                            cldrFileOverride, examples, parts, "light-speed-minute", amount);
+                    addFormattedUnitsConstructed(
+                            cldrFileOverride, examples, parts, "light-speed-hour", amount);
+                    addFormattedUnitsConstructed(
+                            cldrFileOverride, examples, parts, "light-speed-day", amount);
+                    addFormattedUnitsConstructed(
+                            cldrFileOverride, examples, parts, "light-speed-week", amount);
+                    addFormattedUnitsConstructed(
+                            cldrFileOverride, examples, parts, "light-speed-month", amount);
+                    examples.add("Compare with:");
```
btangmu, line 1398: Will this "Compare with:" get displayed appropriately on the front end?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public CLDRFileOverride(CLDRFile sourceFile, Map<String, String> valueOverrides) {
+        super(new DelegateXMLSource(sourceFile.dataSource, valueOverrides));
+    }
+
+    /**
+     * Overrides the values of the source. The keys of the overrides must already be in the
+     * XMLSource source
+     */
+    public static class DelegateXMLSource extends XMLSource {
```
btangmu, line 23: There's already a top-level class named DelegateXMLSource. I'm confused how this new class is related to the old one. Does it need to have the same name, and does it need to be public?

btangmu (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: I tend to agree for very long functions. But this does fit on a page.

tools/cldr-code/src/main/java/org/unicode/cldr/util/UnitConverter.java
macchiati: It is easier to see what changed during development, but I can remove that.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: yes

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
macchiati: That is in the apps directory, and inaccessible (rightfully) in the util directory. Plus, this has some changes. Will retitle.

tools/cldr-apps/.settings/org.eclipse.wst.common.component
srl295, line 1: surprising. did this get deleted automatically?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -0,0 +1,168 @@
...
+
+public class CLDRFileOverride extends CLDRFile {
+
+    /**
+     * Overrides the values of the sourceFile. The keys of the overrides must already be in the
+     * XMLSource source
+     */
+    public CLDRFileOverride(CLDRFile sourceFile, Map<String, String> valueOverrides) {
+        super(new XMLSourceMapOverride(sourceFile.dataSource, valueOverrides));
+    }
```
srl295, line 17: I don't see the value for this being a CLDRFile subclass, since there's only a call to the constructor.

I think better would be to just make the XMLSourceMapOverride public (and the top level class) and add (if even needed) a convenience factory function:

```java
    public static final CLDRFile forFile(CLDRFile sourceFile, Map<String, String> valueOverrides) {
         return new CLDRFile(new XMLSourceMapOverride(sourceFile.dataSource, valueOverrides));
     }
```


but this isn't fundamentally a new kind of CLDRFile, so I don't think the subclass is warranted.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -0,0 +1,168 @@
...
+        private Map<String, String> overrides;
+
+        public XMLSourceMapOverride(XMLSource source, Map<String, String> overrides) {
+            overrides.keySet().stream()
+                    .forEach(
+                            x -> {
+                                if (source.getValueAtDPath(x) == null) {
+                                    throw new IllegalArgumentException();
+                                }
+                            });
```
srl295, line 38: could perhaps be a subclass, ensureNoNulls

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -0,0 +1,168 @@
...
+        private XMLSource delegate;
+
+        private Map<String, String> overrides;
+
+        public XMLSourceMapOverride(XMLSource source, Map<String, String> overrides) {
+            overrides.keySet().stream()
+                    .forEach(
+                            x -> {
+                                if (source.getValueAtDPath(x) == null) {
+                                    throw new IllegalArgumentException();
```
srl295, line 36:
```suggestion
                                    throw new IllegalArgumentException("Null DPath "+x+" found in the source.");
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Overrides the values of the source. <br>
+     * The keys of the overrides must already be in the XMLSource source as DPaths. And the full
+     * paths cannot be changed <br>
+     * Note: this is similar to the top-level DelegateXMLSource in the apps directory, but not the
+     * same, since it provides for the map.
+     */
+    public static class XMLSourceMapOverride extends XMLSource {
+        private XMLSource delegate;
```
srl295, line 27: perhaps DelegateXMLSource should be modified so that this could be a subclass of it instead of largely duplicate?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -0,0 +1,168 @@
...
+            readonly();
+        }
+
+        @Override
+        public void getPathsWithValue(String valueToMatch, String pathPrefix, Set<String> result) {
+            throw new UnsupportedOperationException("support later when needed");
+            // TODO (if needed)
+            // call delegate.getPathsWithValue(valueToMatch, pathPrefix, result);
+            // remove the items that collide with the map
+            // then look through the map to see if any matches there need to be added
```
srl295, line 156: TODO with a ticket #?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -2050,6 +2052,62 @@ public VoterInfo getVoterInfo() {
         }
     }
 
+    public void testLightSpeed() {
```
srl295, line 2055: would be a great parameterized junit test!

srl295 (commented):
Merged 5 minutes before I reviewed!

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
srl295: The class in cldr-apps could certainly be moved to util if it was useful.

PR #4524, opened 2025-03-24
Add `America/Coyhaique` timezone

robertbastian:
CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/bcp47/timezone.xml
+++ b/common/bcp47/timezone.xml
@@ -130,6 +130,7 @@ For terms of use, see http://www.unicode.org/copyright.html
             <type name="chzrh" description="Zurich, Switzerland" alias="Europe/Zurich"/>
             <type name="ciabj" description="Abidjan, Côte d'Ivoire" alias="Africa/Abidjan"/>
             <type name="ckrar" description="Rarotonga, Cook Islands" alias="Pacific/Rarotonga"/>
+            <type name="clcxq" description="Aysén Region, Chile" alias="America/Coyhaique"/>
             <type name="clipc" description="Easter Island, Chile" alias="Pacific/Easter Chile/EasterIsland"/>
             <type name="clpuq" description="Punta Arenas, Chile" alias="America/Punta_Arenas" since="31"/>
             <type name="clscl" description="Santiago, Chile" alias="America/Santiago Chile/Continental"/>


common/supplemental/metaZones.xml
```diff
@@ -645,6 +645,9 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<timezone type="America/Punta_Arenas">
 				<usesMetazone to="2016-12-03 23:00" mzone="Chile"/>
 			</timezone>
+			<timezone type="America/Coyhaique">
+				<usesMetazone to="2025-03-19 23:00" mzone="Chile"/>
```
sffc, line 649: What is the new metazone?

sffc, line 649: What is the new metazone?

robertbastian: There is no new metazone. This puts the new timezone in the existing metazone so it has the same history as the timezone it was split off of.

robertbastian: I modeled this after `America/Punta_Arenas`, which left "Chile Time" in 2016 to do its own thing (permanent DST). It doesn't have a metazone since then.

sffc: This was probably fine when `America/Punta_Arenas` was the only time zone in the metazone, but now both time zones are in the same metazone, should they use the same display name.?

robertbastian: There is no official (https://www.diariooficial.interior.gob.cl/publicaciones/2025/03/20/44104/01/2624263.pdf) or unofficial (https://www.latercera.com/nacional/noticia/tras-consulta-ciudadana-region-de-aysen-mantendra-el-horario-de-verano-durante-todo-el-ano/) name for this "timezone" (non-tech meaning) that would induce a new metazone.

sffc: Should it be Chile Summer Time ?

robertbastian: No

markusicu:
@Yqwed could you please take a look here as well? And coordinate with @yumaoka and @cjchapman on a PR for ICU?

yumaoka:
@robertbastian Can you also update common/supplemental/windowsZones.xml ? This mapping is only applicable to current active zone, so the mapping should be same as `America/Punta_Arenas`.

```
			<!-- (UTC-03:00) Punta Arenas -->
			<mapZone other="Magallanes Standard Time" territory="001" type="America/Punta_Arenas"/>
			<mapZone other="Magallanes Standard Time" territory="CL" type="America/Punta_Arenas America/Coyhaique"/>
```

I think above is what we expect.

cjchapman:
I created JIRA tickets for the tzdata 2025b update:
https://unicode-org.atlassian.net/browse/CLDR-18479
https://unicode-org.atlassian.net/browse/ICU-23096

yumaoka (changes requested):
As I mentioned in the conversation thread, please uppdate windowsZones.xml.

You should also need to update tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/tzdb-version.txt. See https://cldr.unicode.org/development/updating-codes/update-time-zone-data-for-zoneparser

Yqwed:
> @Yqwed could you please take a look here as well? And coordinate with @yumaoka and @cjchapman on a PR for ICU?

I will check this (and 2025b in general) tomorrow.

common/supplemental/metaZones.xml
AEApple: The discussion was whether to wait for it to develop a name instead of inventing one unless there's a very strong case for why it needs one immediately

Yqwed: `windowsZones.xml` uses "Magallanes Standard Time" for `America/Punta_Arenas` [1]. Though there seem to be no links to sources.

[1] https://github.com/unicode-org/cldr/commit/f2ef1270a1cd8554f2d80995435a2c700d5abbd3

yumaoka: >Rule	Chile	2019	max	-	Apr	Sun>=2	3:00u	0	-
>Rule	Chile	2023	max	-	Sep	Sun>=2	4:00u	1:00	-

Chile rule uses daylight saving time from Sep to Apr.

>Zone America/Coyhaique	-4:48:16 -	LMT	1890
>...
>			-4:00	Chile	%z	2025 Mar 20
>			-3:00	-	%z

Until March 20 in this local uses UTC-4:00 + Daylight saving +1:00 = UTC-3:00. With the rule, the rule "-4:00 Chile" should cease 2025-03-20T00:00:00 in local time. And local time at that point is UTC-3:00. So the actual switch time in UTC should be 2025-03-20T03:00:00Z. I think the metazone mapping should end - 

<usesMetazone to="2025-03-20 03:00" mzone="Chile"/>

instead of 

<usesMetazone to="2025-03-19 23:00" mzone="Chile"/>

yumaoka (changes requested):
One more change please.

common/supplemental/metaZones.xml
robertbastian: I've followed what had been done for `America/Punta_Arenas`, which was also observing DST when its change happened: 

> Rule	Chile	2016	2018	-	Aug	Sun>=9	4:00u	1:00	-

> Zone America/Punta_Arenas -4:43:40 -	LMT	1890
...
			-4:00	Chile	%z	2016 Dec  4
			-3:00	-	%z

It is modeled as leaving the metazone at 2016-12-03 23:00.

It's all pretty arbitrary. The first instant where Coyhaique actually diverges from Santiago is 2025-04-05T03:00:00Z, I don't know how TZDB came up with these transition dates.

Yqwed: Second this. It seems that 20th March is just the date when the law was published: https://github.com/eggert/tz/blob/771a14bf500f8ee8d8ab32e86d19dd3998a343d4/southamerica#L1353

yumaoka: Found something interesting - https://unicode-org.atlassian.net/browse/CLDR-10089

I cannot remember what was done. The change was introduced by tzdata 2017a, and later, I pointed out the metazone mapping data should be corrected - and it looks I fixed the problem somewhere - probably in a release branch??

yumaoka: Let's use the correct transition time aligned to TZDB for Coyhaique at least. Punta_Arenas seems also wrong, but it has been there for years.. If you don't want to touch it, we can update it later.

robertbastian: ok

yumaoka: @robertbastian Thanks.

```
<timezone type="America/Punta_Arenas">
	<usesMetazone to="2016-12-03 21:00" mzone="Chile"/>
</timezone>

```

I mentioned Punta_Arenas should swithc at 21:00, but it should be actually `2016-12-04 03:00` instead. Anyway, you did not make the change in this PR, and it's fine. I approved the changes.

yumaoka (approved):
LGTM thanks!

Yqwed:
> What needs to still happen so the tests will pass?

I have following patches:

```
diff --git i/tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java w/tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
index 11a8a19650..3a50b1ef5f 100644
--- i/tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
+++ w/tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
@@ -130,6 +130,8 @@ public final class SubmissionLocales {
                             + "|localeDisplayNames/territories/territory\\[@type=\"TR\"\\].*"
                             // v43: Exemplar city for America/Ciudad_Juarez
                             + "|dates/timeZoneNames/zone[@type=\"America/Ciudad_Juarez\"]/exemplarCity"
+                            // v48: Exemplar city for America/Coyhaique
+                            + "|dates/timeZoneNames/zone\\[@type=\"America/Coyhaique\"]/exemplarCity"
                             + ")");
 
     // Pattern.compile("//ldml/units/unitLength\\[@type=\"long\"]");
diff --git i/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java w/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
index 9035df3e6f..fd5e42fb3d 100644
--- i/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
+++ w/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathHeader.java
@@ -1531,6 +1531,9 @@ public class PathHeader implements Comparable<PathHeader> {
                                     || "ZZ".equals(theTerritory)) {
                                 if ("Etc/Unknown".equals(source0)) {
                                     theTerritory = "ZZ";
+                                // TODO (ICU-23096): remove once ICU's snapshot version is uploaded.
+                                } else if ("America/Coyhaique".equals(source0)) {
+                                    theTerritory = "CL";
                                 } else {
                                     throw new IllegalArgumentException(
                                             "ICU needs zone update? Source: "
diff --git i/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java w/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
index d9e85e66f6..7c7315e7fa 100644
--- i/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
+++ w/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestBCP47.java
@@ -318,6 +318,8 @@ public class TestBCP47 extends TestFmwk {
                     "WET");
 
     public void testBcp47IdsForAllTimezoneIds() {
+        // TODO (ICU-23096): remove once ICU is updated.
+        Set<String> newlyIntroducedTimeZoneIds = Set.of("clcxq");
         Map<String, String> aliasToId = new TreeMap<>();
         Set<String> missingAliases = new TreeSet<>();
         Set<String> deprecatedAliases = new TreeSet<>();
@@ -330,7 +332,9 @@ public class TestBCP47 extends TestFmwk {
             if (itemIsDeprecated) {
                 deprecatedBcp47s.add(bcp47Type);
             }
-            bcp47IdsNotUsed.add(bcp47Type);
+            if (!newlyIntroducedTimeZoneIds.contains(bcp47Type)) {
+                bcp47IdsNotUsed.add(bcp47Type);
+            }
             if (aliasSet == null) {
                 continue;
             }

```

@robertbastian Can you please update CL's title and mark it as active? Or I can upload mine if you don't want to deal with 2025b.

PR #4507, opened 2025-03-18
CLDR-18356 st: unvendor fonts

srl295:
- remove fonts from repo
- add via NPM

CLDR-18356

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4525, opened 2025-03-24
CLDR-18325 Follow-up refactoring of CLDRFile, iteration

btangmu:
-Use the simple pattern, for (String path : cldrFile), instead of ... Iterator<String> it = cldrFile.iterator(); it.hasNext() ... String path = it.next()

-Changes to CLDRFile.java recommended and assisted by IntelliJ

CLDR-18325

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/index.js
+++ b/tools/cldr-apps/js/src/index.js
@@ -17,7 +17,7 @@ import * as cldrVue from "./esm/cldrVue.mjs";
 import * as cldrMonitoring from "./esm/cldrMonitoring.mjs";
 
 // fonts (see surveytool.css)
-import "@fontsource/noto-serif-np-hmong"; 
+import "@fontsource/noto-serif-np-hmong";
 import "glyphicons-only-bootstrap/css/bootstrap.min.css";
 
 // kick off the top level monitoring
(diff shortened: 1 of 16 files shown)


tools/cldr-apps/js/src/index.js
```diff
@@ -17,7 +17,7 @@ import * as cldrVue from "./esm/cldrVue.mjs";
 import * as cldrMonitoring from "./esm/cldrMonitoring.mjs";
 
 // fonts (see surveytool.css)
-import "@fontsource/noto-serif-np-hmong"; 
+import "@fontsource/noto-serif-np-hmong";
```
srl295, line 20: unrelated, i may have missed the prettier call.

srl295: this was leftover from #4507

srl295 (approved):
great improvements

tools/cldr-apps/js/src/index.js
btangmu: For Java, but not for JS, a PR is blocked if not normalized; would that be hard to implement for JS?

PR #4526, opened 2025-03-24
CLDR-17942 ISO8601: Fix coverage also

macchiati:
CLDR-17942

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
@@ -283,6 +283,8 @@ public class CheckDates extends FactoryCheckCLDR {
             }
         }
 
+        checkIso8601(path, value);
+
         try {
             if (path.contains("[@type=\"abbreviated\"]")) {
                 // ensures abbreviated <= wide for quarters, months, days, dayPeriods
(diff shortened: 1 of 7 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -641,6 +643,99 @@ && isTooMuchWiderThan(value, abbrValue)
...
+            case "timeFormatLength":
+            case "availableFormats":
+                break;
+            case "intervalFormats":
+                isInterval = true;
+                break;
+            default:
+                return null;
+        }
+        // verify the order is year month dow hour minute second
```
btangmu, line 695: dow = day?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -641,6 +643,99 @@ && isTooMuchWiderThan(value, abbrValue)
         return this;
     }
 
+    static final Set<Integer> expectedField =
```
srl295, line 646: seems like this could be an API on the ICU side.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
srl295: dow = day of week

It's restricting it to `expectedFields`.    So in order, YEAR QUARTER MONTH WEEKDAY(dow) DAY HOUR MINUTE SECOND ZONE

```suggestion
        // verify the order is YEAR QUARTER MONTH WEEKDAY(dow) DAY HOUR MINUTE SECOND ZONE (by the integer values)
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -641,6 +643,99 @@ && isTooMuchWiderThan(value, abbrValue)
...
+            Set.of(
+                    DateTimePatternGenerator.YEAR,
+                    DateTimePatternGenerator.QUARTER,
+                    DateTimePatternGenerator.MONTH,
+                    DateTimePatternGenerator.DAY,
+                    DateTimePatternGenerator.WEEKDAY,
+                    DateTimePatternGenerator.HOUR,
+                    DateTimePatternGenerator.MINUTE,
+                    DateTimePatternGenerator.SECOND,
+                    DateTimePatternGenerator.ZONE);
```
srl295, line 656:
```suggestion
                    DateTimePatternGenerator.YEAR,
                    DateTimePatternGenerator.QUARTER,
                    DateTimePatternGenerator.MONTH,
                    DateTimePatternGenerator.WEEKDAY,
                    DateTimePatternGenerator.DAY,
                    DateTimePatternGenerator.HOUR,
                    DateTimePatternGenerator.MINUTE,
                    DateTimePatternGenerator.SECOND,
                    DateTimePatternGenerator.ZONE);
```

put in order

srl295 (approved):
LGTM with suggestions

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: Day Of Week (will fix in later PR)

macchiati:
I'll make those changes in next PR.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: Set.of makes an unordered immutable set (Hash). ImmutableSet.of makes an immutable ordered set (LinkedHash). While generally I prefer the latter but the order, for the test, doesn't matter. 

However we want weekday to be between day and hour, because in formatting you never say 

September, Tuesday, 1.
but rather
September 1, Tuesday

PR #4534, opened 2025-03-26
CLDR-18423 st: search: improvements to search and tests

srl295:
Add new searches
- search by code
- search by parent locale
- search by xpath

Also:
- if the user is in say a top page, pass in 'en' as the current locale- this will let them search the 'baseline' content. Prevents error when searching from the main page.

Behind the scenes:
- refactor for testability
- update tests

CLDR-18423

Yes it should be using a search collator..

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
@@ -2,6 +2,8 @@ package org.unicode.cldr.web;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import java.io.Closeable;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
@@ -13,9 +15,11 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.Factory;
+import org.unicode.cldr.util.PathHeader;
 
-public class SearchManager {
+public class SearchManager implements Closeable {
     static final Logger logger = SurveyLog.forClass(SearchManager.class);
 
     /** The request of a search */
@@ -33,7 +37,7 @@ public class SearchManager {
     }
 
     /** Struct for one single result */
-    public static final class SearchResult {
+    public static final class SearchResult implements Comparable<SearchResult> {
         @Schema(description = "xpath to the resource")
         public String xpath;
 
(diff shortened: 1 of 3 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
```diff
@@ -139,9 +170,67 @@ public void begin() {
...
         }
 
+        final CLDRLocale EN = CLDRLocale.getInstance("en");
+
         @Override
         public Search call() throws Exception {
             // first, simple search of exact values
+            addExactMatches(locale);
+
+            CLDRLocale cLoc = CLDRLocale.getInstance(locale);
```
macchiati, line 180: I don't think we want this, except for inherited values.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
```diff
@@ -139,9 +170,67 @@ public void begin() {
...
+            CLDRLocale parLoc = cLoc.getParent();
+            while (response.isEmpty() && parLoc != null) {
+                addExactMatches(parLoc.getBaseName());
+                parLoc = parLoc.getParent();
+            }
```
macchiati, line 185: I don't think we want to search the parent, except for inherited values.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
```diff
@@ -139,9 +170,67 @@ public void begin() {
...
+
+            // add codes (in the locale)
+            addCodes(locale);
+
+            // Add English (if we didn't already search it)
+            if (response.isEmpty() && !cLoc.childOf(EN)) {
```
macchiati, line 191: We will probably want control over whether or not to include English, to avoid spurious hits. Eg a check box.

macchiati: But we can start with this.

macchiati (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
srl295: This uses a non resolved file and searches the child first.

btangmu (approved):
This will be very helpful! I'm curious about things like searches that take a long time, display of partial results while a search is in progress, and canceling a search in progress. It will be easier to learn about those things by actually using the interface than studying the code.

srl295:
> This will be very helpful! I'm curious about things like searches that take a long time, display of partial results while a search is in progress, and canceling a search in progress. It will be easier to learn about those things by actually using the interface than studying the code.

The API already supports long term searches and cancelling. It is supposed to cancel a previous search if someone changes their search request.

srl295:
I purposefully didn't try to do UI improvements here, pending the other issue gathering requirements. Play with it and add comments to the tickets with suggestions.

PR #4537, opened 2025-03-27
CLDR-18144 Initial investigation

macchiati:
CLDR-18144

ALLOW_MANY_COMMITS=true


AEApple:
Will languages be able to re-order the extension and the core parts? Do we know if there are languages where for certain combinations the extension may appear on different sides of the core word (kinda like in French where most adjectives follow the noun, while a handful precede the noun? https://en.wikipedia.org/wiki/French_grammar#Adjectives

macchiati:
> Will languages be able to re-order the extension and the core parts? Do we know if there are languages where for certain combinations the extension may appear on different sides of the core word (kinda like in French where most adjectives follow the noun, while a handful precede the noun? https://en.wikipedia.org/wiki/French_grammar#Adjectives

The extension use the same mechanism as the language and (script, region, variant, keys, etc). That is, someone would see in a menu (where the core/extension mechanism is used — it is optional) as the following (look at the Kurdish lines for the differences)

NEW
```
...
French
Kurdish (Kurmanji)
Kurdish (Sorani)
Serbian (Cyrillic)
Serbian (Latin)
Swedish (Sweden)
Swedish (Finland)
...
```

OLD
```
...
French
Kurdish
Serbian (Cyrillic)
Serbian (Latin)
Sorani Kurdish
Swedish (Sweden)
Swedish (Finland)
...
```

See also the regions' part of the spreadsheet in the ticket.

sffc (commented):
I looked at the spreadsheet and there are some interesting findings that should be discussed.

The tool itself looks like a proof-of-concept; is it your intent to eventually land this as part of a test?

conradarcturus:
@macchiati  do you want to land this code? Have there been any changes to the approach since you posted this?

macchiati:
I'd like to go ahead and land the tool, because I think it could be useful in the future.

macchiati:
This is just tooling, with no effect on the release. There is one unrelated test failure; I strongly suspect that rebasing will fix it, but I don't have time now. So when I get back.

PR #3700, opened 2024-05-08
CLDR-17615 Update languages (v46 SBRS)

btangmu:
-Did all comprehensible/doable steps in https://cldr.unicode.org/development/updating-codes/update-languagescriptregion-subtags

-I'm unable to do this step: https://cldr.unicode.org/development/updating-codes/external-version-metadata

-Region code CQ may require further work

-Output of GenerateEnums and mvn test, after changes in this PR, is included in PR comments

CLDR-17615

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4543, opened 2025-03-28
CLDR-18461 Update ISO languages for v48

conradarcturus:
I was building a new dataset with ISO codes and I thought "I should check to make sure the CLDR ISO codes are up to date". Turns out there are a few changes.

Followed the protocol in this document:

https://cldr.unicode.org/development/updating-codes/update-languagescriptregion-subtags

Question: @btangmu's previous PR added the languages to the XMLs but it doesn't look like the script did that for him -- do I have to do that manually? https://github.com/unicode-org/cldr/pull/3700

CLDR-18461

ALLOW_MANY_COMMITS=true


srl295:
Yes, I meant to mention that the language sub tag registry had been updated

btangmu:
@conradarcturus Looking at my notes, I see that I ran GenerateValidityXML as part of https://github.com/unicode-org/cldr/pull/3700 . Any changes to I made to XML files are more likely to be from running tools than from hand-editing, though I don't remember for sure.

conradarcturus:
Perfect! That's the script that ran it. 2 questions though:

* The guide mentions updating en.xml --- uhhh we haven't added new languages or scripts to en.xml. In fact, we only have 684 localeDisplayNames (out of 7963) and 55 variant names (out of 122 variants).
* The last ISO update was Jan 15 -- based on their cadence I wouldn't be surprised if there is another update April 15 -- do you know if there is an impending update?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language-subtag-registry
```diff
@@ -1,4 +1,4 @@
-File-Date: 2024-06-14
+File-Date: 2025-03-10
```
srl295, line 1: ok good - didn't see this at first

srl295 (approved):

conradarcturus:
Turns out because of the deprecated codes I need to actually do a few more changes anyway, not ready yet

conradarcturus:
Added language aliases for the deprecated codes. Also added Language Matches since a different document requested that update when we add new languages to macrolanguages.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateLikelySubtags.java
```diff
@@ -499,6 +499,9 @@ public static void main(String[] args) throws IOException {
                                 // Temporary additions to fix bad script mappings
                                 // See https://unicode-org.atlassian.net/browse/CLDR-18121
                                 {"sga", "sga_Latn_IE"},
+
+                                // Perserve data now that dek has been merged with sqm
+                                {"sqm", "sqm_Latn_CF"},
```
conradarcturus, line 504: If we didn't do this -- then our import of SIL data was going to merge "dek_Latn_CM" with "sqm_Latn_CF" into "sqm_Latn_CM" -- however there are <5000 speakers in Cameroon and 50,000 speakers in CAF - so I needed to add an override here. I can add more comprehensive Cameroon and CAF data later -- already this PR is doing a lot...

srl295 (dismissed):

macchiati (approved):
Looks great, thanks!

macchiati:
> The guide mentions updating en.xml --- uhhh we haven't added new languages or scripts to en.xml. In fact, we only have 684 localeDisplayNames (out of 7963) and 55 variant names (out of 122 variants).

The guide is out of date; we have never tried to add all the English language / script names to en.xml. We might do that along with the endonyms, but in another directory, not main — see the endonym ticket.
And we have only added a subset of variants (and even those there it a ticket to reduce those to the ones in more common usage). Again, we might add them all along, but not in main.

PR #4545, opened 2025-03-30
CLDR-17570 Update number-and-currency-patterns.md

macchiati:
CLDR-17570

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/translation/number-currency-formats/number-and-currency-patterns.md
+++ b/docs/site/translation/number-currency-formats/number-and-currency-patterns.md
@@ -128,3 +128,16 @@ X digit-two
 
 Xdigit-other
 
+## Rational Formatting
+
+There are special patterns for formatting rational fractions, such as in the following:
+
+![Screenshot 2025-03-29 at 17 46 03](https://github.com/user-attachments/assets/1a5a3ba5-e38e-49f5-9219-b1cd1d35ad08)
+
+As usual, different numbering systems used in your language can have different patterns. Here are the different fields for you to vote for.
+
+1. **Rational** — A pattern that is used to format a rational fraction (eg, ½), using the numerator and denominator. For most languages, there is a special kind of ‘slash’ character that is used to separate the numerator and denominator. In well-constructed fonts, this will cause the digits before that slash to be superscripted, and the digits after to be subscripted. Only change this if your language doesn't use a slash for rational fractions.
+2. **Integer + Rational** — This pattern is used to “glue” an integer and a formatted rational fraction together, such as 3 + ½, to get 3 ½ (**with** a space). This pattern is to be used when an implementation **has not** verified that the digits before the slash will be superscripted, and the number would look like 31/2 if it doesn't.
+3. **Integer + Rational** — This pattern is used to “glue” an integer and a formatted rational fraction together, such as 3 + ½, to get 3½ (**without** a space). This pattern is to be used if the implementation **has** verified that superscripting is performed.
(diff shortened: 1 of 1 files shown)


macchiati:
Please merge if you Approve

docs/site/translation/number-currency-formats/number-and-currency-patterns.md
```diff
@@ -128,3 +128,16 @@ X digit-two
 
 Xdigit-other
 
+## Rational Formatting
+
+There are special patterns for formatting rational fractions, such as in the following:
+
+![Screenshot 2025-03-29 at 17 46 03](https://github.com/user-attachments/assets/1a5a3ba5-e38e-49f5-9219-b1cd1d35ad08)
```
srl295, line 135: this is an ephemeral URL. this file needs to be brought into the repo (in this same directory is fine)

srl295:
I think the discussion of implementations won't be helpful to users here. That belongs in the spec. I'll think about a suggested reformulation

docs/site/translation/number-currency-formats/number-and-currency-patterns.md
```diff
@@ -128,3 +128,16 @@ X digit-two
...
+
+As usual, different numbering systems used in your language can have different patterns. Here are the different fields for you to vote for.
+
+1. **Rational** — A pattern that is used to format a rational fraction (eg, ½), using the numerator and denominator. For most languages, there is a special kind of ‘slash’ character that is used to separate the numerator and denominator. In well-constructed fonts, this will cause the digits before that slash to be superscripted, and the digits after to be subscripted. Only change this if your language doesn't use a slash for rational fractions.
```
srl295, line 139:
```suggestion
1. **Rational** — A pattern that is used to format a rational fraction (eg, ½), using the numerator (`{0}`) and denominator (`{1}`). For most languages, there is a special kind of ‘slash’ character that is used to separate the numerator and denominator. Only change this value if your language doesn't use a slash for rational fractions.
```

What should it be changed to? Can an example be given here?

docs/site/translation/number-currency-formats/number-and-currency-patterns.md
```diff
@@ -128,3 +128,16 @@ X digit-two
...
+2. **Integer + Rational** — This pattern is used to “glue” an integer and a formatted rational fraction together, such as 3 + ½, to get 3 ½ (**with** a space). This pattern is to be used when an implementation **has not** verified that the digits before the slash will be superscripted, and the number would look like 31/2 if it doesn't.
```
srl295, line 140:
```suggestion
2. **Integer + Rational** — This pattern is used to “glue” an integer and a formatted rational fraction together, such as `3` + `1/2`, to get `3 1/2` (**with** a space).
```

docs/site/translation/number-currency-formats/number-and-currency-patterns.md
```diff
@@ -128,3 +128,16 @@ X digit-two
...
+3. **Integer + Rational** — This pattern is used to “glue” an integer and a formatted rational fraction together, such as 3 + ½, to get 3½ (**without** a space). This pattern is to be used if the implementation **has** verified that superscripting is performed.
```
srl295, line 141:
```suggestion
3. **Integer + Rational** — This pattern is used to “glue” an integer and a formatted rational fraction (with super/subscript numbers) together, such as `3` + `½`, to get `3½` (**without** a space).
```

Implementation notes should be in the spec.

macchiati: These are not really implementation notes. There is an important distinction between the two different formats that takes some judgment. If a language favors spaces between `3` + `½`, then they can be the same. 

**However**, if a language does not favor spaces between `3` + `½`. then one of them still has to have a space (probably a narrow/thin no-break space but possibly a no-break space; or for CJK a zero-width space). 

The note about the implementation is to help give guidance as to when what is used.

AEApple:
Are we ready to merge this?

macchiati:
I'd recommend merging and then fixing later. Steven, are you okay with that.?

AEApple (approved):
Approving but we will need to update to resolve Stevens comments before we start submission.

PR #4547, opened 2025-03-30
CLDR-18329 For the known cases in which there is keyword entry without tts, re-enable the check failure

pedberg-icu:
CLDR-18329

For the known cases (in 8 locales) for which there is a keyword annotation without a tts annotation, re-enable the check failure so vetters will see the error in Survey Tool. (Wondering whether this will also cause a failure in CI tests). Later on during data resolution, if we still have such cases, we will remove the unpaired keyword annotation.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
@@ -1,14 +1,16 @@
 package org.unicode.cldr.test;
 
-import java.util.Arrays;
+// import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
+// import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
+import org.unicode.cldr.test.CheckCLDR.CheckStatus.Type;
+import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.LocaleIDParser;
 import org.unicode.cldr.util.PatternCache;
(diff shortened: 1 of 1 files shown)


pedberg-icu:
@srl295 @macchiati So, the console check fail causes the CI tests to fail, so I cannot merge this. How can I add a check that will fail in Survey Tool to alert vetters to an issue, but have the CI tests pass so I can add the check??

macchiati:
The check needs to look at the phase, and if it is BUILD, then turn it into a warning instead of an error.

The CI tests should already be set to BUILD — if not, Steven and/or Tom should fix.

pedberg-icu:
> The check needs to look at the phase, and if it is BUILD, then turn it into a warning instead of an error.

Thanks, I noticed that CheckPersonNames.java did something similar so I used its approach.

Also, I have to update a currency test skip for regions CW/SX since the preferred currency changes as of 2025-Mar-31 (had to too the same thing in PR #4549 )

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -1594,7 +1594,7 @@ public void TestSupplementalCurrency() {
                 // TODO 17397: remove isKnownIssue and the if around errln when the logknown issue
                 // goes away.
                 final boolean skipKnownIssue =
-                        currency.equals("ANG")
+                        currency.equals("XCG")
                                 && isoCountries.isEmpty()
```
pedberg-icu, line 1598: The preferred currency changes as of 2025-Mar-31, so updated th test skip, and also noted this in the CLDR ticket for the test skip.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -1594,7 +1594,7 @@ public void TestSupplementalCurrency() {
...
                                 && cldrCountries.equals(Set.of("CW", "SX"))
                                 && logKnownIssue("CLDR-17397", "Mismatched codes " + cldrCountries);
```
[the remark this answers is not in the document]
srl295: @pedberg-icu it would be better to put this in a separate commit rather than an unrelated one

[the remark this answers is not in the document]
srl295: @pedberg-icu since the other one is  merged, this one could just be rebased now

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
```diff
@@ -103,10 +106,14 @@ public CheckCLDR handleCheck(
                 }
             }
             if (ttsMissing) {
+                final Type mainType =
+                        getPhase() != Phase.BUILD
+                                ? CheckStatus.errorType
+                                : CheckStatus.warningType; // we need to be able to check this in
```
srl295, line 112: this is the right way to do it.

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
srl295, line 112: Sorry, it's not right

Perhaps instead:

```java
final CLDRConfig.Environment env = CLDRConfig.getInstance().getEnvironment();
final boolean onSurveyTool = env  != CLDRConfig.Environment.PRODUCTION && env != CLDRConfig.Environment.SMOKETEST;
```

then:

```suggestion
                final CLDRConfig.Environment env = CLDRConfig.getInstance().getEnvironment();
                final boolean onSurveyTool = env  != CLDRConfig.Environment.PRODUCTION && env != CLDRConfig.Environment.SMOKETEST;
                final Type mainType =
                        onSurveyTool
                                ? CheckStatus.errorType // error in ST
                                : CheckStatus.warningType; // warning elsewhere
```

srl295 (dismissed):
OK once tests pass

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
pedberg-icu: @srl295 Thanks! I have rebased locally and am trying out the code above. But `onSurveyTool ` seems to have the wrong polarity and I am getting failures running ConsolCheck. Should it not be
```
onSurveyTool = (env == CLDRConfig.Environment.PRODUCTION || env == CLDRConfig.Environment.SMOKETEST);
```
?

pedberg-icu:
> OK once tests pass

@srl295 Thanks, I had not actually committed my rebase and test fix at that point but I have now so please re-review, thanks.

srl295 (approved):

srl295:
> @srl295 Is there a way to get the jira-ticket check unstuck?

1. go to another PR and find the check's URL, <https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/cldr/4546>
2. update the PR # at the end to <https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/cldr/4547>
3. internal server error

maybe due to https://www.githubstatus.com ? no, that shows green.

pedberg-icu:
> 2. update the PR # at the end to https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/cldr/4547
> 3. internal server error

@srl295 Thanks, I tried this (and I remembered having figure that out myself before) and it worked for me without internal error, so things are now OK for this PR.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckAnnotations.java
```diff
@@ -103,13 +107,24 @@ public CheckCLDR handleCheck(
...
+                // we can commit
                 result.add(
                         new CheckStatus()
                                 .setCause(this)
-                                .setMainType(CheckStatus.errorType)
+                                .setMainType(mainType)
                                 .setSubtype(Subtype.ttsAnnotationMissing)
                                 .setMessage(
-                                        "Have keywords but missing the corresponding name (tts) entry"));
+                                        "Have keywords but missing the corresponding name (tts) entry; error vs warning depends on environment {0}",
```
srl295, line 126: "err vs warning depends on environment" won't make sense to vetters and probably isn't needed.

PR #4561, opened 2025-04-01
CLDR-18422 Update collation for Inari Sámi

jhsoby:
Update the collation for Inari Sámi to use the correct alphabetical order for the language's letters, and sort common extra letters of neighbouring languages in the most logical locations.

CLDR-18422

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/collation/smn.xml
+++ b/common/collation/smn.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
 <!--
-Copyright © 1991-2014 Unicode, Inc.
+Copyright © 1991-2025 Unicode, Inc.
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
 For terms of use, see http://www.unicode.org/copyright.html
 -->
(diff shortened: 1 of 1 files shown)


srl295:
structure seems ok

srl295 (approved):

markusicu (commented):
Is there an authoritative reference?

Looking at https://en.wikipedia.org/wiki/Inari_S%C3%A1mi_language#Orthography
- We still tailor à although that's not mentioned in the article nor in the ticket. Does that in fact want to be primary-equal to â?
- Do å and æ really want to be primary-equal to â? Or can they use the root order? (å secondary-after a, æ sorts like a+e)
  - In [Norwegian](https://github.com/unicode-org/cldr/blob/main/common/collation/no.xml)/[Swedish](https://github.com/unicode-org/cldr/blob/main/common/collation/sv.xml)/[Finnish](https://github.com/unicode-org/cldr/blob/main/common/collation/fi.xml), they sort after z.

Also, and maybe more importantly, an alphabet showing accented letters does not necessarily mean that they should sort primary-after their base letters. For example, the "German alphabet" includes äöüß, but they sort primary-equal to aou and ss (in the German default=dictionary sort order).

Is there any way to confirm that all of these accented letters should be primary distinct? For example, are there printed phone books? Dictionaries? Would they always be included in an "[alphabetic index](https://www.unicode.org/reports/tr35/tr35-collation.html#Collation_Indexes)"?

macchiati (changes requested):
Can you point to some sources to verify the sort order. Many seem to put some accented vowels after Z, such as https://en.wikipedia.org/wiki/S%C3%A1mi_orthography

PR #4563, opened 2025-04-01
CLDR-18350 Add dateTimeFormat type=relative, update spec, coverage, ExampleGenerator, tests

pedberg-icu:
CLDR-18350

Add a new type 'relative' for the dateTimeFormat format for combining dates and times (in addition to the existing types `standard` and `atTime`). If this exists, it is used for combining a relative date with a single time; if absent, we fall back to the `standard` format for this combination. (We do this instead of falling back to the `atTime` format in order to produce behavior that is better for locales that do not get much attention; it does mean that vetters for locales that do get attention may need to copy any existing atTime pattern to the relative pattern if they want to preserve current behavior for relative date + single time, and this info will be added to the vetter guidelines).
- add the new type in ldml.dtd.
- add `dateTimeFormat type="relative"` entries in `root` with fallback to  `type="standard"` as discussed above.
- add `dateTimeFormat type="relative"` entries in `en` that are copies of the `atTime` entries (to maintain the current atTime behavior for relative dates), and entries in  `hi`, `or`, and `ta` that explicitly fall back to the `standard` entries (per desired behavior described in the ticket).
- add coverage for the `dateTimeFormat type="relative"` formats.
- update the spec to describe the new formats and how they should be used.
- update the ExampleGenerator to generate the right example patterns based on the spec guidance.
- update PathDescription.java, prettyPath.txt, and MapComparator in DtdData.java.
- add tests (for coverage, ExampleGenerator)

The vetter guidance site pages will be updated in a separate PR for this ticket.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -1528,7 +1528,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT dateTimeFormat ( alias | ( pattern*, displayName*, special* ) ) >
 <!ATTLIST dateTimeFormat type NMTOKEN "standard" >
-    <!--@MATCH:literal/standard, atTime-->
+    <!--@MATCH:literal/standard, atTime, relative-->
 <!ATTLIST dateTimeFormat alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
 <!ATTLIST dateTimeFormat draft (approved | contributed | provisional | unconfirmed | true | false) #IMPLIED >
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -1827,12 +1827,18 @@ annotations.
 						<dateTimeFormat type="atTime">
 							<pattern>{1} 'at' {0}</pattern>
 						</dateTimeFormat>
+						<dateTimeFormat type="relative">
+							<pattern>{1} 'at' {0}</pattern>
+						</dateTimeFormat>
 					</dateTimeFormatLength>
 					<dateTimeFormatLength type="long">
 						<dateTimeFormat>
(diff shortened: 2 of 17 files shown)


richgillam (approved):
This looks good to me, AFAICT.  A couple process questions:
- The CLDR-to-ICU tooling handles this properly without having to add anything, doesn't it?  Or will somebody need to make a change there?
- The ICU data will have to be regenerated to contain these patterns.  How/when does that happen?
- We'll have to make code changes in ICU to take advantage of this.  Is there a ticket for that?

pedberg-icu:
@richgillam , thanks.

> * The CLDR-to-ICU tooling handles this properly without having to add anything, doesn't it?  Or will somebody need to make a change there?

There needs to be an additional entry in tools/cldr/cldr-to-icu/src/main/resources/ldml2icu_locale.txt, parallel to the entry already there for atTime formats.

> * The ICU data will have to be regenerated to contain these patterns.  How/when does that happen?

Usually there is a CLDR milestone-1 integration to ICU right before the opening of CLDR data submission, after new CLDR structure is in, to see if there are any integration issues with the new structure, and to get the new structure into ICU so code can be written that uses it. That is typically when we would update ldml2icu_locale.txt as above.

> * We'll have to make code changes in ICU to take advantage of this.  Is there a ticket for that?

Not currently an ICU ticket for that, I can create one.

richgillam:
> > * We'll have to make code changes in ICU to take advantage of this.  Is there a ticket for that?
> 
> Not currently an ICU ticket for that, I can create one.

Thank you!  Sounds like we need one for the CLDR-to-ICU tool as well.  Or you can just point that out in the ticket for the ICU change.

PR #4565, opened 2025-04-02
CLDR-16755 Restore data-driven approach to PathDescriptions

btangmu:
-Read descriptions from PathDescriptions.txt instead of including them in PathDescriptions.java

-Add a TODO comment about duplicated URLs here and in CLDRUrls.java

CLDR-16755

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathDescriptions.txt
```diff
@@ -0,0 +1,218 @@
...
+^//ldml/typographicNames/styleName[@type="([^"]*)"][@subtype="([^"]*)"]; A label for a typographic style, such as “Narrow” or “Semibold”. For more information, see %TYPOGRAPHIC_NAMES.
+^//ldml/typographicNames/featureName[@type="([^"]*)"]; A label for a typographic feature, such as “Small Capitals”. For more information, see %TYPOGRAPHIC_NAMES.
+^//ldml/characterLabels/characterLabelPattern\[@type="([^"]*)"]\[@count="([^"]*)"]; A label for a set of characters that has a numeric placeholder, such as “1 Stroke”, “2 Strokes”. For more information, see %CHARACTER_LABELS.
+^//ldml/characterLabels/characterLabelPattern\[@type="([^"]*)"]; A modifier composed with a label for a set of characters. For more information, see %CHARACTER_LABELS.
+^//ldml/characterLabels/characterLabel\[@type="([^"]*)"]; A label for a set of characters. For more information, see %CHARACTER_LABELS.
+^//ldml/annotations/annotation\[@cp="([^"]*)"]\[@type="tts"]; A name for a character or sequence. For more information, see %SHORT_CHARACTER_NAMES.
+^//ldml/annotations/annotation\[@cp="([^"]*)"]; A set of keywords for a character or sequence. For more information, see %SHORT_CHARACTER_NAMES.
```
btangmu, line 218: In addition to avoiding duplication of the URLs, we could save some space and ensure consistency by adding macros for "For more information, see " (sometimes, inconsistently, with "please": "For more information, please see ") and "Note: before translating, be sure to read "

srl295:
Since it's markdown, i thought it could be something like this as a single .md file.

The 'references' part would be appended to every extracted markdown blob.

```md
# Path Descriptions

Data for generating descriptions of paths in Survey Tool.
OK to write more stuff here, etc.

## Descriptions

<!-- this is a comment -->
- `^//ldml/annotations/annotation\[@cp="([^"]*)"]\[@type="tts"]`

A name for a character or sequence. For more information, see [Short Character Names]

- `^//ldml/annotations/annotation\[@cp="([^"]*)"]`
 
A set of keywords for a character or sequence. For more information, see [Short Character Names]

<!-- References -->
[Short Character Names]: https://cldr.unicode.org/translation/characters/short-names-and-keywords#h.4a8wthj27m74
[Transforms]: https://cldr.unicode.org/translation/transforms
[Typographic Names]: https://cldr.unicode.org/translation/characters/typographic-names
[TZ City Names]: https://cldr.unicode.org/translation/time-zones-and-city-names
[Units]: https://cldr.unicode.org/translation/units
[Misc Units]: https://cldr.unicode.org/translation/units
```

srl295 (approved):
approving because it's an improvement but suggest to take a more markdown centric approach (maybe as followon)

btangmu:
@srl295 and @macchiati I remember the markdown idea and I see that it has potential benefits. I'm not sure how/when/where the file would get parsed and divided -- on the front or back end, and with what code -- an existing library, or something custom-built?

Also, whether it's .txt or .md, there's a problem with keeping the URLs consistent with CLDRUrls.java. That was one of my motivations for switching from .txt to .java in the first place. (Also I don't like long lines that require horizontal scrolling, and I do like formats that support automatic formatting and syntax highlighting/checking. These are partly matters of personal preference.)

Instead of using the existing %macro substitution feature for the URLs, we could design a new custom notation/mechanism. It would be best, though, to have an overall design, taking into account the goal of merging hints.txt with PathDescription.txt/md/java, and so forth -- see the design doc https://docs.google.com/document/d/1BBm5XdOFnT7HFbhWJ55MArEq_FBVP1sJvsC1MVeNBlk/edit?tab=t.0

PR #4566, opened 2025-04-02
CLDR-18475 deps, web: remove org.json from deps, fix rtl

srl295:
CLDR-18475

- Also fixes CLDR-5839
- Drop org.json dependency completely
- Add shim objects for now to ease the transition to Gson, but no longer link against org.json
- add a standard Gson serializer object in JsonUtil.gson()

Also:

- Better handle CLA deserialization issues from prefs.
**Migration note**: this will invalidate previously signed CLAs (would only affect test data) which will get this message instead.  Now use ISO 8601 style timestamps.

> `Failed parsing 'Dec 9, 2024, 12:43:31 PM' as Date; at path $.signed`

- ST: remove setOverride() global - use an explicit override for locale directionality.
  This was causing a race condition.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
@@ -812,10 +812,6 @@ function locInfo(loc) {
   return locmap.getLocaleInfo(loc);
 }
 
-function setOverrideDir(dir) {
-  overridedir = dir;
-}
-
 /**
  * Set the dir and lang attributes for a node that represents
  * a CLDR value.
@@ -823,7 +819,7 @@ function setOverrideDir(dir) {
  * @param {Node} node DOM node
  * @param {String} loc locale
  */
-function setLang(node, loc) {
+function setLang(node, loc, overridedir) {
   var info = locInfo(loc);
 
   if (overridedir) {
@@ -979,7 +975,6 @@ export {
   localizeFlyover,
   parseStatusAction,
   setLang,
-  setOverrideDir,
   setShower,
   showLoader,
   testsToHtml,
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
@@ -1,0 +1,51 @@
+package org.unicode.cldr.web.util;
+
+import com.google.gson.Gson;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.TreeMap;
+
+/* shim */
+public class JSONObject implements JSONString {
+    private static final Gson gson = JsonUtil.gson();
+    private final Map<String, Object> m = new TreeMap<>();
+
+    @Override
+    public String toJSONString() throws JSONException {
+        return gson.toJson(m);
+    }
+
+    @Override
+    public String toString() {
+        return toJSONString();
+    }
+
+    public JSONObject put(String k, Object v) {
+        m.put(k, v);
+        return this;
+    }
+
+    public Object get(String string) {
+        return m.get(string);
+    }
+
+    public JSONArray getJSONArray(String string) {
+        return (JSONArray) (get(string));
+    }
+
+    public String getString(String k) {
+        return get(k).toString();
+    }
+
+    public int length() {
+        return m.size();
+    }
+
+    public boolean has(String ov) {
+        return m.containsKey(ov);
+    }
+
+    public Iterator<String> keys() {
+        return m.keySet().iterator();
+    }
+}
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -177,8 +177,6 @@ function insertRowsIntoTbody(theTable, reuseTable) {
   for (let i in rowList) {
     const k = rowList[i];
     const theRow = theRows[k];
-    const dir = theRow.dir;
-    cldrSurvey.setOverrideDir(dir != null ? dir : null);
     /*
      * Don't regenerate the headings if we're re-using an existing table.
      */
@@ -809,7 +807,7 @@ function updateRowProposedWinningCell(tr, theRow, cell, protoButton) {
     const flagIcon = cldrSurvey.addIcon(cell, "s-flag");
     flagIcon.title = cldrText.get("flag_desc");
   }
-  cldrSurvey.setLang(cell);
+  cldrSurvey.setLang(cell, null, theRow.dir);
   tr.proposedcell = cell;
 
   /*
@@ -844,7 +842,7 @@ function updateRowProposedWinningCell(tr, theRow, cell, protoButton) {
 function updateRowOthersCell(tr, theRow, cell, protoButton) {
   let hadOtherItems = false;
   cldrDom.removeAllChildNodes(cell); // other
-  cldrSurvey.setLang(cell);
+  cldrSurvey.setLang(cell, null, theRow.dir);
 
   /*
    * Add the other vote info -- that is, vote info for the "Others" column.
--- a/tools/cldr-apps/pom.xml
+++ b/tools/cldr-apps/pom.xml
@@ -110,12 +110,6 @@
 			<artifactId>javax.activation</artifactId>
 		</dependency>
 
-		<!-- https://mvnrepository.com/artifact/org.json/json -->
-		<dependency>
-			<groupId>org.json</groupId>
-			<artifactId>json</artifactId>
-		</dependency>
-
 		<!-- https://mvnrepository.com/artifact/org.jsoup/jsoup -->
 		<dependency>
 			<groupId>org.jsoup</groupId>
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/util/CLDRConfigImpl.java
@@ -10,15 +10,11 @@ import java.io.PrintWriter;
 import java.net.URL;
 import java.util.Date;
 import java.util.Enumeration;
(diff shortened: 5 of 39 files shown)


tools/cldr-apps/js/src/esm/cldrSurvey.mjs
```diff
@@ -812,18 +812,14 @@ function locInfo(loc) {
...
-function setOverrideDir(dir) {
-  overridedir = dir;
-}
-
 /**
  * Set the dir and lang attributes for a node that represents
  * a CLDR value.
  * Also appends the 'cldrValue' class to the node.
  * @param {Node} node DOM node
  * @param {String} loc locale
```
btangmu, line 820: Missing @param for overridedir.

More importantly, it looks like cldrSurvey.mjs still includes a top-level `let overridedir = null` which is distinct from this function parameter, and unaffected by this parameter, so parts of `cldrSurvey` may be accessing one, and other parts accessing the other. Also, what are the consequences if not all callers of `setLang` use the 3rd parameter?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
```diff
@@ -0,0 +1,51 @@
+package org.unicode.cldr.web.util;
+
+import com.google.gson.Gson;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.TreeMap;
+
+/* shim */
+public class JSONObject implements JSONString {
```
btangmu, line 9: It's a little surprising that this general-purpose code needs to be built from scratch. Is there a viable path for migrating this to a dependency on something not written by us?

btangmu (commented):
See comment about overridedir. Also, do you want to test this on staging?

Ideally some front-end tests would inspire more confidence...

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: > Also, what are the consequences if not all callers of setLang use the 3rd parameter?

if they don't set it, it will be undefined and therefore not used.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
srl295: yes, using Google's gson directly.

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: added `@param` and removed orphan and unused `let overridedir`

tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
srl295: - consider the web/api/ classes, where we just use plain objects (with jackson serialization almost completely hidden)
- JsonUtil.gson() is now used universally in the cldr-apps module, you can see examples of it searching for gson().
- cldr-code uses Gson separately in two instances.

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
btangmu: > if they don't set it, it will be undefined and therefore not used.

I'm worried that previously, `overridedir` might have been set to RTL by one method, and then kept that value for subsequent usages where it wasn't explicitly set/changed, whereas with these changes, it would effectively default back to LTR for those subsequent usages. Is that possible, and if so, is the change for better or for worse? It seems risky. Calling an n-parameter function with less than n parameters generally raises doubt whether the omission is intentional or accidental. I'll study it a bit...

btangmu: After studying it for a bit, I think the whole system is fragile, and there's a risk that this change will cause problems, though a big problem is my own ignorance.

For example, cldrVote.mjs calls `cldrSurvey.setLang(span)` and currently implicitly reuses `overridedir`, whereas with this PR it doesn't, so there's a fallback based on the current locale. It's not clear to me when/why the override differs from what would be gotten from the locale.

This PR, instead of calling `cldrSurvey.setOverrideDir` once for each row, calls `cldrSurvey.setLang(cell, null, theRow.dir)` from `updateRowProposedWinningCell` and `updateRowOthersCell`. That seems good in terms of identifying those two cells as main (only?) ones that may contain values that need localization.

`cldrTable.appendExample` is sometimes called with a `loc` parameter and sometimes without (in which case the current locale is used), and it's not clear whether the explicit parameter is just to save time by avoiding a call to `cldrStatus.getCurrentLocale`... actually, at least when called with `TRANS_HINT_ID = "en"` it overrides the current locale.

We'd better do some serious testing, especially for examples, and for submitting new values that contain mixed RTL/LTR text. I have almost zero knowledge of any RTL languages...

btangmu: I just compared cldr-smoke (without this PR) and cldr-staging (with this PR). Doing exactly the same thing (submitting a new value gotten by copying the Winning value, clicking to the right of it, and then appending 123 or abc), I got different results.

Without this PR:
image

With this PR:
image
Which is correct?

btangmu: Without this PR (cldr-smoke):
image

With this PR (cldr-staging):
image

Maybe the difference is entirely due to RIGHT_TO_LEFT vs rtl, which I understand is a bug fix

srl295: that is the bug and the fix

srl295: it should be right aligned as in the "with this fix"

srl295: > and for submitting new values that contain mixed RTL/LTR text. I have almost zero knowledge of any RTL languages.

We don't need to do that.  This is only setting the default directionality for the language, which is calculated on the Java side. Mixed text doesn't affect it.

srl295: > After studying it for a bit, I think the whole system is fragile, and there's a risk that this change will cause problems, though a big problem is my own ignorance.


I think it's working correctly.  'loc' is null meaning use the default (current locale) in most cases unless it's an English sample in which case it gets overridden to en.

btangmu: There seems to be no difference between smoke and staging as far as the input box itself is concerned:
image

As far as I can tell from DOM inspection, there are no `dir` attributes, so the rendering must be whatever is default for a mixture of RTL and LTR text.

srl295: the inputbox may not be affected here. i'll check it

btangmu: Example hover on smoke:
image

Example hover on staging:
image

I hope that's an improvement.

btangmu (approved):
Approving, and hoping people with RTL expertise will look at the screenshots and do further testing before ST opens

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: i'm assuming 123 was added at the beginning of the string? yes, it looks like it's displaying correctly in the example and the cell

btangmu: The examples are left-justified in both cases. Should they be right-justified for consistency?

srl295:
https://unicode-org.atlassian.net/browse/CLDR-17861 I found this issue while analyzing that one - your issue about inconsistent directionality across plurals.  The display on smoketest was wrong.   But if you look at the screenshots in CLDR-17861 you will see they are showing a RTL (right justified for example)  view.

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: they may need to be dir=rtl at a different level in the structure.

btangmu: > assuming 123 was added at the beginning of the string?

Yes, "beginning" meaning the right end, not the left. But if I clicked at the left end first, it was displayed exactly the same in the input box.

srl295:
@btangmu I'm going to merge this because I think it's an improvement, and then it will get to smoketest.

I'll work on the FE test, and note the example and input box issues in the issue (and leave it open)

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
btangmu: This is weird. I click at the left end of the string, and type "LEFT", and click SUBMIT. Then I click on the + button again and the WINNING button. Here's what I see:

image

What I see in the input box doesn't match what ends up being displayed as the value that I typed. Is that a bug?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
btangmu: Should our home-grown JSONObject be deprecated?

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: the bug is a lack of dir on the input box

tools/cldr-apps/src/main/java/org/unicode/cldr/web/util/JSONObject.java
btangmu: This appears related to https://unicode-org.atlassian.net/browse/CLDR-18513

PR #4567, opened 2025-04-02
CLDR-17857 coverage for scripts, update 'bal' coverage

srl295:
CLDR-17857

- add `bal` to modern
- update test which had missed bal before
- also, test that all Basic lang's scripts are in modern
- add CLDRLocale.toMaximal
- add check for variants - with a known issue CLDR-18480

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
@@ -1,6 +1,7 @@
 package org.unicode.cldr.util;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
```diff
@@ -44,7 +46,6 @@ public void TestCoveragePerf() {
     @Test
     public void TestPriorBasicLanguage() throws IOException {
```
srl295, line 47: Everything is aggregated - so that it will print out ALL errors and not just stop at the first.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
```diff
@@ -107,4 +127,71 @@ public void TestPriorBasicLanguage() throws IOException {
...
     }
+
+    private void checkLanguageCoverageForScriptsInBasic(
+            final CalculatedCoverageLevels prevCovLevel,
+            final Map<String, CoverageLevel2> covs,
+            final CLDRFile english)
+            throws MultipleFailuresError {
+        final Level failIfAbove = Level.MODERN;
+
+        final String XPATH_LANGUAGE = "//ldml/localeDisplayNames/scripts/script";
```
pedberg-icu, line 138: Odd name given that it is for scripts... and lines below refer to language when they really mean script. I know there was some copy-paste from the previous function...

srl295:
let's do this. I'll make the 'log known issue' a println and take out the refactoring. then have a separate issue for the refactoring.

srl295:
@pedberg-icu removed the logKnownIssues stuff, so this is a simpler pr again

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
```diff
@@ -42,69 +37,211 @@ public void TestCoveragePerf() {
...
-
-        // Result: locales not in en.xml
-        final Set<String> notInEnglish = new TreeSet<>();
-        // Result: locales not in coverage
-        final Set<String> notInCoverage = new TreeSet<>();
 
         final Set<String> localesToCheck =
                 SupplementalDataInfo.getInstance().getLanguageTcOrBasic();
+
         final Map<String, CoverageLevel2> covs = new HashMap<>();
```
conradarcturus, line 53: I know covs is used alot -- but the variable name is pretty short and I don't grok what it is. It is languages to their coverage level so something by coverageByLanguage/languageToCoverage? Is it always language or is it sometimes other things?

conradarcturus (approved):
Thanks for the test :)

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
srl295: It's locales

srl295: It's actually just a set to check , the keys aren't as important.

PR #4569, opened 2025-04-03
CLDR-18481 logKnownIssues for JUnit

srl295:
- only for the cldr-code project for now
- refactor TestShim to cleanup - TestShimUtils is now ShimmedMain
- improvements to TestParams

This will make it easier to move tests over to JUnit (much more flexible, modern API)

CLDR-18481

Note: this test adds a known issue on itself, CLDR-18481 - however, when CLDR-17857 #4567 is merged, the variant test will be a known issue.

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/util/ShimmedMain.java
srl295, line 1: this was TestShimUtils.java previously

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCoverageLevel2.java
srl295, line 1: only actual case of logKnownIssues in JUnit now

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
```diff
@@ -1174,6 +1161,31 @@ public static class TestParams {
...
+            if (key != null && key.length() > 0 && props != null) {
+                val = (String) props.get(key.toLowerCase());
+            }
+            return val;
+        }
+
+        public boolean getBooleanProperty(String key, boolean defVal) {
+            String s = getProperty(key);
+            if (s != null) {
+                if (s.equalsIgnoreCase("yes") || s.equals("true")) {
```
btangmu, line 1175: Maybe `s.equals("true")` should be `s.equalsIgnoreCase("true")` for consistency here?

PR #4572, opened 2025-04-03
CLDR-18437 Remove duplicate language alias

conradarcturus:
The Mandingo `man` macrolanguage has 7 members. Right now, in languageAlias, we show 2 default members that represent the whole of Mandingo. In reality, there should be only 1 default for each macrolanguage, so this updates the values to just 1.

Eastern Maninkakan `emk` has ~3.5million speakers but Mandinka `mnk` only has 1 million, so Eastern Maninkakan takes precedence.

CLDR-18437

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalMetadata.xml
+++ b/common/supplemental/supplementalMetadata.xml
@@ -304,7 +304,6 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<languageAlias type="zsm" replacement="ms" reason="macrolanguage"/> <!-- Standard Malay ⇒ Malay -->
 			<languageAlias type="zyb" replacement="za" reason="macrolanguage"/> <!-- Yongbei Zhuang ⇒ Zhuang -->
 			<languageAlias type="him" replacement="srx" reason="macrolanguage"/> <!-- Himachali ⇒ Sirmauri (= Pahari, Himachali) -->
-			<languageAlias type="mnk" replacement="man" reason="macrolanguage"/> <!-- Mandinka ⇒ Mandingo -->
 			<languageAlias type="bh" replacement="bho" reason="macrolanguage"/> <!-- Bihari ⇒ Bhojpuri -->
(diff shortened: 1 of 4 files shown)


pedberg-icu:
Hmm. Currently common/validity/language.xml shows `mnk` as deprecated, likely because until now it was supposed to be replaced by `man`. To fix the errors you will probably need to remove it from the deprecated section and add it to the regular section. @srl295 does that seem right?

srl295:
> Hmm. Currently common/validity/language.xml shows `mnk` as deprecated, likely because until now it was supposed to be replaced by `man`. To fix the errors you will probably need to remove it from the deprecated section and add it to the regular section. @srl295 does that seem right?

that would be correct.

btangmu:
```
  TestValidity {
    TestBasicValidity (0.000s) Passed
    TestBothDirections (0.009s) Passed
    TestCompatibility (0.373s) Passed
    TestLanguageTagCanonicalizer (0.001s) Passed
    TestLanguageTagParser (0.000s) Passed
    TestLstrConsistency
Error:  (TestValidity.java:623)  Error: Deprecated validity - lstr
	LstrType=	language
	size=	1
	codes=	[mnk]
	names=	[
		mnk ⇒	{Description=Mandinka, Added=2009-07-29, Macrolanguage=man}
		valid.⇒	deprecated
		alias⇒	null]

Warning:  (TestValidity.java:544)  Warning: Changed aliases from LSTR, just double-check: 3
```

conradarcturus:
Thanks for the tips everyone. I needed to rebuild the validity XML.

```
mvn package -DskipTests=true && java -jar tools/cldr-code/target/cldr-code.jar GenerateValidityXML
```

pedberg-icu (approved):

pedberg-icu:
Also need to run spotless code formatter:
```
mvn --file=tools/pom.xml spotless:apply
```

srl295 (approved):

pedberg-icu (approved):

PR #4573, opened 2025-04-03
CLDR-16755 Do not require Gulf time short/standard at moderate coverage

btangmu:
-Remove the line for ...Gulf...short/standard from coverageLevels.xml

-Do not remove the similar line for Gulf...long/standard

-This may affect all locales associated with territory AE, such as ar and en, since ar_AE and en_AE exist

CLDR-16755

ALLOW_MANY_COMMITS=true


btangmu:
Sorry, wrong ticket number. I'll make a new PR.

PR #4574, opened 2025-04-04
CLDR-18475 st: fix RTL input in input box and examples

srl295:
- rename cldrAddAlt.addButton and cldrAddValue.addButton to reduce my own confusion
- export cldrSurvey.locInfo
- import <a-config-provider> from ant
- add a <AddValue dir="rtl"/> prop to AddValue to support passing through the directionality
- update AddValue to use the configProvider to set directionality
- fix a typo in TestCldrSurvey
- update cldrVue.mount to pass through extra props to mounted components
- also fix RTL on examples

CLDR-18475

ALLOW_MANY_COMMITS=true

Fixed input page:
image

Fixed example page:
image

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
@@ -802,7 +802,7 @@ function appendExtraAttributes(container, theRow) {
  * @param loc optional
  * @returns locale bundle
  */
-function locInfo(loc) {
+export function locInfo(loc) {
   if (!loc) {
     loc = cldrStatus.getCurrentLocale();
   }
--- a/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSurvey.mjs
@@ -802,7 +802,7 @@ function appendExtraAttributes(container, theRow) {
  * @param loc optional
  * @returns locale bundle
  */
-export function locInfo(loc) {
+function locInfo(loc) {
   if (!loc) {
     loc = cldrStatus.getCurrentLocale();
   }
@@ -980,6 +980,7 @@ export {
   hideLoader,
   isInputBusy,
   localizeFlyover,
+  locInfo,
   parseStatusAction,
   setLang,
   setShower,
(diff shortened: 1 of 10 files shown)


srl295:
Filed https://unicode-org.atlassian.net/browse/CLDR-18482 for further improvement to the AddAlt / AddValue

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
```diff
@@ -802,7 +802,7 @@ function appendExtraAttributes(container, theRow) {
  * @param loc optional
  * @returns locale bundle
  */
-function locInfo(loc) {
+export function locInfo(loc) {
```
btangmu, line 805: Suggestion: put the export at the bottom of the file for consistency and to be able to see at a glance what's exported

btangmu (dismissed):

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: I'll do that for consistency (maybe in a followon PR)- but I think it's mostly for compatibility with the commonjs `module.exports = { locInfo, … }` pattern - i'm not sure this is a feature?  newer files such as cldrCla.mjs I don't use the exports block at all

btangmu: > having self contained, makes it more amenable to refactoring… You can just pick up that one function and move it and not need to adjust the bottom of the file.

True, in that respect it's easier. On the other hand, the time spent actually editing might be small compared to the time spent studying and deciding. Moving a function to another file might change whether it still needs to be exported at all, or whether other functions in the same file still need exporting or moving. Code needs rebalancing from time to time. If we see that we're adding another export to a module that already has more than about a dozen exports, it's a clue that we may need more separation of concerns, less tight coupling between modules. What's most expedient in the short term can lead to tech debt in the long term. We need to look at the forest, not just the trees.

srl295:
@btangmu export moved to bottom!

tools/cldr-apps/js/src/esm/cldrSurvey.mjs
srl295: well, they would be better with classes with true encapsulation rather than just groups of related functions- but that's a more involved refactor.

btangmu (approved):

PR #4575, opened 2025-04-06
CLDR-18447 Updated English names for iso8601 calendar, coverage, tests

pedberg-icu:
CLDR-18447

- Per ticket, updated English display names for iso8601 calendar.
    - The ticket had the primary name as “Gregorian (Year First)“ but all of the other primary calendar names include the word Calendar (there is a ticket to revamp that, but we should do them all together), so made the primary name “Gregorian Calendar (Year First)“
    - The ticket also requested an alt="variant" name “ISO 8601 Order”
- Added coverage for the alt name at moderate, same as the primary name.
    - At the same time, updated coverage for some other iso8601 calendar items that were incorrectly at basic to be moderate (per discussion in e-mail): intervalFormatFallback, and availabeFormats for %basicDateSkeletons
- Updated tests, PathDescriptions

ALLOW_MANY_COMMITS=true


AEApple:
I thought we were going to put the iso8601 calendar to comprehensive per CLDR-18473 and CLDR-18458?

pedberg-icu:
> I thought we were going to put the iso8601 calendar (other than the display name) up to comprehensive per [CLDR-18473](https://unicode-org.atlassian.net/browse/CLDR-18473) and [CLDR-18458](https://unicode-org.atlassian.net/browse/CLDR-18458)?

@AEApple Ah, I see. I only remembered a discussion about making them moderate, not the resolution about making them comprehensive., and not that thee was a ticket. Shall I use this PR to make coverage comprehensive or shall we make a separate PR corresponding to the separate ticket (and then leave this as it is)? And what about the *name* for ISO 8601, should that also be comprehensive or modern or moderate? That was specifically what I had to change to fix a test failure for this PR.

AEApple:
>  Shall I use this PR to make coverage comprehensive or shall we make a separate PR corresponding to the separate ticket (and then leave this as it is)? And what about the _name_ for ISO 8601, should that also be comprehensive or modern or moderate? That was specifically what I had to change to fix a test failure for this PR.

Whichever is easier for you. I think the calendar name at moderate makes sense, but we need to decide about what is reasonable for vetters to be able to localize in the date formats in this calender so other than the calendar name, we should have the rest of the calendar in comprehensive until that discussion is resolved.

pedberg-icu:
> > Shall I use this PR to make coverage comprehensive ...
> 
> Whichever is easier for you. I think the calendar name at moderate makes sense, but we need to decide about what is reasonable for vetters to be able to localize in the date formats in this calender so other than the calendar name, we should have the rest of the calendar in comprehensive until that discussion is resolved.

@AEApple I agree the iso8601 _formats_ should probably at least modern and maybe comprehensive, probably people should only use them in `root` or `en`. But the _name_ of the iso8601 calendar should be moderate. I think we should keep the changes in this PR as is and then use the separate ticket about iso8601 format coverage to make the formats comprehensive for now.

AEApple (approved):

conradarcturus:
Thanks for making the update Peter!

PR #4581, opened 2025-04-07
CLDR-18464 Add a Before Hijrah era to the Islamic calendars

sffc:
CLDR-18464

ALLOW_MANY_COMMITS=true

CC @roozbehp @pedberg-icu @AEApple

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -4536,30 +4536,35 @@ XXX Code for transations where no currency is involved
           <calendarSystem type="lunar" />
           <eras>
               <era type="0" start="622-7-15" code="islamic" aliases="ah"/>
+              <era type="1" end="622-7-14" code="islamic-inverse" aliases="bh"/>
           </eras>
         </calendar>
         <calendar type="islamic-civil">
           <calendarSystem type="lunar" />
           <eras>
               <era type="0" start="622-7-16" code="islamic-civil" aliases="islamicc ah"/>
+              <era type="1" end="622-7-15" code="islamic-civil-inverse" aliases="bh"/>
           </eras>
         </calendar>
         <calendar type="islamic-rgsa">
           <calendarSystem type="lunar" />
           <eras>
               <era type="0" start="622-7-15" code="islamic-rgsa" aliases="ah"/>
+              <era type="1" end="622-7-14" code="islamic-rgsa-inverse" aliases="bh"/>
           </eras>
         </calendar>
         <calendar type="islamic-tbla">
           <calendarSystem type="lunar" />
           <eras>
               <era type="0" start="622-7-15" code="islamic-tbla" aliases="ah"/>
+              <era type="1" end="622-7-14" code="islamic-tbla-inverse" aliases="bh"/>
           </eras>
         </calendar>
         <calendar type="islamic-umalqura">
           <calendarSystem type="lunar" />
           <eras>
               <era type="0" start="622-7-15" code="islamic-umalqura" aliases="ah"/>
+              <era type="1" end="622-7-14" code="islamic-umalqura-inverse" aliases="bh"/>
           </eras>
         </calendar>
         <calendar type="chinese">
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -3195,8 +3195,13 @@ annotations.
 					</monthContext>
 				</months>
 				<eras>
+					<eraNames>
+						<era type="0">Anno Hegirae</era>
+						<era type="1">Before Hijrah</era>
+					</eraNames>
 					<eraAbbr>
 						<era type="0">AH</era>
+						<era type="1">BH</era>
 					</eraAbbr>
 				</eras>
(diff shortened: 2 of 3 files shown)


common/supplemental/supplementalData.xml
```diff
@@ -4535,30 +4535,35 @@ XXX Code for transations where no currency is involved
         <calendar type="islamic">
           <calendarSystem type="lunar" />
           <eras>
+              <era type="-1" end="622-7-14" code="islamic-inverse" aliases="bh"/>
```
sffc, line 4538: I gave it the code "islamic-inverse" because I want to land the code updates, #4519, atomically once we have agreement on them.

common/supplemental/supplementalData.xml
```diff
@@ -4535,30 +4535,35 @@ XXX Code for transations where no currency is involved
...
               <era type="0" start="622-7-15" code="islamic" aliases="ah"/>
           </eras>
         </calendar>
         <calendar type="islamic-civil">
           <calendarSystem type="lunar" />
           <eras>
+              <era type="-1" end="622-7-15" code="islamic-civil-inverse" aliases="bh"/>
```
sffc, line 4545: I gave it the type -1 because type 0 is already used for the AH era. Please let me know if I should instead:

1. Use a positive number like 1 (and leave AH as 0)
2. Change BH to 0, AH to 1, and update all the XML files that currently have translations for AH at type 0 to instead be at type 1

pedberg-icu: @sffc Usually this number is used as an array index in ICU. I guess I would be in favor of your option #2 above. We would need to add a migration note for ICU (and for CLDR), and probably update some ICU code. I think we allow years to be negative, but I do not think we allow eras to be.

roozbehp (commented):
Generally looks good to me. I would also prefer the "Change BH to 0, AH to 1" option, as it seems cleanest.

sffc:
This means changing all the XML files, which I'm willing to do and can probably automate, but to make sure, before I proceed, is it sufficient for me to change something like

```xml
					<eraAbbr>
						<era type="0">যুগ</era>
					</eraAbbr>
```

to

```xml
					<eraAbbr>
						<era type="1">যুগ</era>
					</eraAbbr>
```

without adding structure for type="0"? It will inherit automatically, right?

pedberg-icu:
> This means changing all the XML files, which I'm willing to do and can probably automate...It will inherit automatically, right?

As far as I know, yes.

sffc:
The other calendars don't have data. Only `islamic`.

```bash
$ xmlstarlet sel -t -v "/ldml/dates/calendars/calendar[@type='islamic-civil']/eras/*/era[@type=0]" *.xml
$ xmlstarlet sel -t -v "/ldml/dates/calendars/calendar[@type='islamic-rgsa']/eras/*/era[@type=0]" *.xml
$ xmlstarlet sel -t -v "/ldml/dates/calendars/calendar[@type='islamic-umalqura']/eras/*/era[@type=0]" *.xml
$ xmlstarlet sel -t -v "/ldml/dates/calendars/calendar[@type='islamic-tbla']/eras/*/era[@type=0]" *.xml
$ 
```

srl295 (dismissed):
needs commit fix

sffc:
@srl295, do you prefer me to merge the PR in this shape (AH becomes index 1, BH inserted at index 0) or would you prefer keeping AH at index 0 and adding BH at index 1 as @FrankYFTang suggested?

pedberg-icu:
> @srl295, do you prefer me to merge the PR in this shape (AH becomes index 1, BH inserted at index 0) or would you prefer keeping AH at index 0 and adding BH at index 1 as @FrankYFTang suggested?

@sffc I am not @srl295 but I now prefer keeping AH at index 0 and adding BH at index 1

sffc:
OK please re-review.

srl295 (approved):
The overall PR is straightforward to review now, with changes in other locales undone

srl295:
Do we need a followup issue to make sure CLDR spec clarifies this question so that it doesn't need to get discussed again? Variants and/or lack thereof?

sffc:
This is ready to merge; should I Squash And Merge or Rebase And Merge?

I'd like to use [CLDR-18369](https://unicode-org.atlassian.net/browse/CLDR-18369) for all spec updates on this topic.

pedberg-icu:
> This is ready to merge; should I Squash And Merge or Rebase And Merge?

We usually Squash and merge except for special circumstances such as cherry-picking maint branch changes to main. But since this only has 1 commit does it make a difference? Does squash gives a different commit hash?

pedberg-icu:
This causes an IndexOutOfBoundsException at org.unicode.cldr.test.ExampleGenerator.handleEras(ExampleGenerator.java:3158), because there is a reference to a hard-coded CALENDAR_ERAS HashMap maps "islamic" to a sample date for only a single era, need to add another entry for the new islamic era. I will make a PR to do that.

PR #4582, opened 2025-04-07
CLDR-18430 Update coverage levels before submission

conradarcturus:
The coverage levels for most locales dropped significantly because of the new basic item: GMT Unknown Format	eg. "GMT+?". Note that the ISO-8601 calendar also caused drops in coverage but as of Monday morning it's no longer counted as basic: https://github.com/unicode-org/cldr/pull/4575 

This updates the coverageLevels.txt to reflect that and other vetter-data changes. See all missing data in https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/48/tsv/locale-missing.tsv

CLDR-18430

Generated by:
```
mvn package -DskipTests=true && java -jar tools/cldr-code/target/cldr-code.jar GenerateAllCharts
```

ALLOW_MANY_COMMITS=true


srl295 (approved):
sure.. we can always pull this back. I'm not sure we usually do this at this point but, i think this reflects reality for now.

conradarcturus:
I'm not going to merge this. There are many tests on fire because of so many locales dropped below basic. I would have to update the tests but I don't want to forget to update those tests.

PR #4583, opened 2025-04-07
CLDR-18486 publish Jira updater

srl295:
CLDR-18486

no impact on CLDR.

Update the jira updater so that it can be pushed to npm separately (readme, etc)

ALLOW_MANY_COMMITS=true


srl295:
you can review the result at https://www.npmjs.com/package/@unicode-org/jira-updater

btangmu (approved):
Neat!

PR #4584, opened 2025-04-07
CLDR-18463 Clean out 46 notes and update known issues

AEApple:
CLDR-18463

Clean out 46 notes and update known issues. Then we can add in updates from here.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/translation.md
+++ b/docs/site/translation.md
@@ -17,14 +17,13 @@ During Submission, please read the CLDR Training (if new to the survey tool), pl
 
 ### Updates
 
-- There are substantial changes to the guidance on dealing with Emoji. Be sure to read the updates in **New Areas**!
-- The Survey Tool Approval column has new icons in the Approval column; see **Survey Tool**.
+- TBD
 
 When a section below changes, the date will be in the header.
 
 ## Status and Schedule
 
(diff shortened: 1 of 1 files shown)


pedberg-icu (approved):

docs/site/translation.md
```diff
@@ -120,24 +63,25 @@ fvfixed date with single time (as was the case in CLDR 47 and earlier), then for
...
+TBD
 
 ## Survey Tool
 
 Once trained and up to speed on [Critical reminders](#critical-reminders-for-all-linguists) (below), log in to the [Survey Tool](https://st.unicode.org/cldr-apps/) to begin your work.
 
 ### Survey Tool Changes
 
+1. The ability to search in the Survey Tool has been added in [CLDR-18423][] and supports searching for: values, English value, and for the codes
 1. There has been substantial performance work that will show up for the first time. If there are performance issues, please file a ticket with a row URL and an explanation for what happened.
```
btangmu, line 75: Is there a problem with the numbering? It seems like all the items start with "1". Or is that a feature?

btangmu (commented):
I guess the numbering thing ("1, 1, 1, ...") is a feature I didn't know about; in preview mode it does display as "1, 2, 3" or "i, ii, iii"

srl295 (approved):

pedberg-icu:
Merging so I can start adding stuff for 48. I will check the numbering.

docs/site/translation.md
AEApple: Usually in markdown you can use 1. for everything and it will be correctly numbered. LMK if it doesn't work in this markdown. It's a lot more convenient when things are likely to change numbers as list items are added/removed.

PR #4585, opened 2025-04-08
CLDR-18144 Unit test fixes

macchiati:
Still need to fix Checks

CLDR-18144

This focuses on Kurdish just to get everything working.

TBD:
- Fix the Checks (especially collisions)
- Extend to root (probably with extra paths)
- Add more languages (perhaps); focus should be on languages where locales have with coverage = moderate or above in 47 to start; we can always add more later.
- Consider adding regions

ALLOW_MANY_COMMITS=true


macchiati:
Note: still draft, but would appreciate early sanity checks.

macchiati:
I can finish the rest in a subsequent PR

macchiati:
Ping, I'd like to get a review. If it looks ok, please merge it.

Cf @AEApple

sffc (approved):

common/main/en.xml
```diff
@@ -347,6 +349,8 @@ annotations.
 			<language type="ksf">Bafia</language>
 			<language type="ksh">Colognian</language>
 			<language type="ku">Kurdish</language>
+			<language type="ku" menu="core">Kurdish</language>
+			<language type="ku" menu="extension">Kurmanji</language>
```
conradarcturus, line 353: Kurmanji or Northern?

macchiati:
I've always heard it as Kurmanji. We can add both, with an alt, but still
have to decide on the primary one.

PR #4591, opened 2025-04-08
CLDR-18463 Update InfoHub for CLDR 48, part 1

pedberg-icu:
CLDR-18463

This is just a first pass at updating the top-level InfoHub page for things that vetters need to pay attention to in CLDR 48. It is not complete; , there may be items here that really belong in the subpages (with a link from the top-level page). Unfortunately this week I have no more time to work on this, so I hope others can take over and flesh this out as necessary and make any other updates and corrections (so unless there are serious problems maybe we can just merge this and then someone else can make a new PR for further updates).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/translation.md
+++ b/docs/site/translation.md
@@ -17,8 +17,6 @@ During Submission, please read the CLDR Training (if new to the survey tool), pl
 
 ### Updates
 
-- TBD
-
 When a section below changes, the date will be in the header.
 
 ## Status and Schedule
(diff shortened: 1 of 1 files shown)


AEApple (approved):
LGTM! Let's merge this and then clean up from there.

I think there are some polish itesm that we should do such as linking to the CLDR training from the top sections etc.

PR #4535, opened 2025-03-26
CLDR-16439 Alias South_Pole to McMurdo

robertbastian:
CLDR-16439

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4592, opened 2025-04-09
CLDR-16439 zone alias South_Pole was moved to McMurdo, so remove it from Auckland

pedberg-icu:
CLDR-16439

PR [#4535](https://github.com/unicode-org/cldr/pull/4535) added the secondary zone alias South_Pole for McMurdo, but did not remove it from its previous entry for Auckland. That caused a serious problem in the timezoneTypes array generated for ICU which tries to map legacy zone IDs to modern one; it was trying to map South_Pole to 2 different IDs, so the resource could not be read and all mapping of BCP47 keys and types was broken. So this PR removes the old alias. I will also file a ticket for a unit test to check this sort of thing.

ALLOW_MANY_COMMITS=true


conradarcturus (approved):

pedberg-icu:
Thanks Conrad!

PR #4595, opened 2025-04-09
CLDR-18375 Check time format matches region

btangmu:
-Add to existing checks in CheckDates

-New methods checkTimeFormatMatchesRegion, getRegionHourFormat

-Failure causes warning, not error

CLDR-18375

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
@@ -50,6 +50,7 @@ import org.unicode.cldr.util.PathHeader;
 import org.unicode.cldr.util.PatternCache;
 import org.unicode.cldr.util.PreferredAndAllowedHour;
 import org.unicode.cldr.util.RegexUtilities;
+import org.unicode.cldr.util.StandardCodes;
 import org.unicode.cldr.util.SupplementalDataInfo;
 import org.unicode.cldr.util.XPathParts;
 
@@ -88,6 +89,9 @@ public class CheckDates extends FactoryCheckCLDR {
     private static final Pattern YEAR_FIELDS = PatternCache.get("(y|Y|u|U|r){1,5}");
 
     private static final String CALENDAR_ID_PREFIX = "/calendar[@type=\"";
+
+    private static final String TIME_FORMAT_CHECK_PATH =
+            "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/timeFormats/timeFormatLength[@type=\"short\"]/timeFormat[@type=\"standard\"]/pattern[@type=\"standard\"]";
 
     // The following calendar symbol sets need not have distinct values
     // "/months/monthContext[@type=\"format\"]/monthWidth[@type=\"narrow\"]/month",
@@ -253,20 +257,24 @@ public class CheckDates extends FactoryCheckCLDR {
             return this; // skip paths that we don't have
         }
 
+        if (value == null) {
+            return this;
+        }
+
         if (!path.contains("/dates") || path.endsWith("/default") || path.endsWith("/alias")) {
             return this;
         }
 
         if (!accept(result)) return this;
 
+        if (TIME_FORMAT_CHECK_PATH.equals(fullPath)) {
+            checkTimeFormatMatchesRegion(value, result);
+        }
+
         String sourceLocale = getCldrFileToCheck().getSourceLocaleID(path, status);
 
         if (!path.equals(status.pathWhereFound)
                 || !sourceLocale.equals(getCldrFileToCheck().getLocaleID())) {
-            return this;
-        }
-
-        if (value == null) {
             return this;
         }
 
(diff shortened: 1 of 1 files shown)


btangmu:
@pedberg-icu I started working on this again based on your comment in https://unicode-org.atlassian.net/browse/CLDR-18495

The ticket is "pre-sub2" rather than "pre-sub", which is one reason why I hadn't gotten to it sooner.

If I run console check without "-e" (errors only), on either this branch or the main branch, it fails due to various exceptions possibly involving thread safety -- is that a known problem?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -654,6 +662,70 @@ && isTooMuchWiderThan(value, abbrValue)
...
+            Map<String, String> likelySubtags) {
+        PreferredAndAllowedHour prefAndAllowedHr = timeData.get(localeID);
+        if (prefAndAllowedHr == null) {
+            LocaleIDParser lp = new LocaleIDParser();
+            String region = lp.set(localeID).getRegion();
+            if (region == null || region.isEmpty()) {
+                String loc2 = likelySubtags.get(localeID);
+                if (loc2 != null && !loc2.isEmpty()) {
+                    region = lp.set(loc2).getRegion();
+                }
```
pedberg-icu, line 716: We need one more step here. At this point we know localeID does not have a region, and we tried to get the likelySubtags for it in loc2. However, if that fails, it might be because localeID has a script, and likelySubtags does not have an entry for that combination of language and script. In that case we have to strip the script and look up likelySubtags for the language part only, and then get the region from that.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -654,6 +662,70 @@ && isTooMuchWiderThan(value, abbrValue)
...
+                    && shortPatSkeleton.indexOf(timeCycleChar) < 0) {
+                String message =
+                        "Time format does not match region; expected "
+                                + timeCycleChar
+                                + " in the value "
+                                + value;
+                CheckStatus item =
+                        new CheckStatus()
+                                .setCause(this)
+                                .setMainType(CheckStatus.errorType)
```
pedberg-icu, line 695: A way to have this be an error in SurveyTool but only a warning in ConsoleCheck (so you can commit) was suggested to me by Steven when I had the same problem in a different PR, see lines 110-119 here: https://github.com/unicode-org/cldr/pull/4547/files

pedberg-icu:
@btangmu I analyzed the 19 errors that you listed in the table in the infrastructure notes.

13 of them are real errors; the locales in question have no Gregorian time formats at all so they are inheriting the root patterns using `HH:mm`, but their region is one that prefers `h` and the test is correctly detecting that. There is a method (suggested to me by Steven) where we can have this be an error in Survey Tool but not in ConsoleCheck; see my comment at line 695. The locales in this situation are: 

| locale   | maximized   | pref | error |
| -------- | ----------- | ---- | ----- |
| apc      | apc_Arab_SY | h    | expected h (correct) in the value HH:mm (inherited from root) |
| arn      | arn_Latn_CL | h    | " |
| bgn      | bgn_Arab_PK | h    | " |
| cho      | cho_Latn_US | h    | " |
| cop      | cop_Copt_EG | h    | " |
| gn       | gn_Latn_PY  | h    | " |
| hnj      | hnj_Hmnp_US | h    | " |
| kek      | kek_Latn_GT | h    | " |
| nv       | nv_Latn_US  | h    | " |
| pis      | pis_Latn_SB | h    | " |
| quc      | quc_Latn_GT | h    | " |
| skr      | skr_Arab_PK | h    | " |
| wbp      | wbp_Latn_AU | h    | " |

The other 6 are cases in which the locale does have Gregorian formats and they do match the preference for the locale, but the test is incorrectly determining what that preference is. That is because the test is not handling locales consisting of language+script (but not region); I have a comment at your line 716 about how to fix this. This applies to the following locales:

| locale   | maximized   | pref | error |
| bal_Latn | bal_Latn_PK | h    | expected H (wrong) in the value hh:mm a (coming from locale data) |
| kok_Latn | kok_Latn_IN | h    | expected H (wrong) in the value a h:mm (coming from locale data) |
| ks_Deva  | ks_Deva_IN  | h    | expected H (wrong) in the value a h:mm (coming from locale data) |
| kxv_Deva | kxv_Deva_IN | h    | expected H (wrong) in the value h:mm a (coming from locale data) |
| ms_Arab  | ms_Arab_MY  | h    | expected H (wrong) in the value h:mm a (coming from locale data) |
| vai_Latn | vai_Latn_LR | h    | expected H (wrong) in the value h:mm a (coming from locale data) |

pedberg-icu:
> If I run console check without "-e" (errors only), on either this branch or the main branch, it fails due to various exceptions possibly involving thread safety -- is that a known problem?

It is not known to me :-)

btangmu:
> > If I run console check without "-e" (errors only), on either this branch or the main branch, it fails due to various exceptions possibly involving thread safety -- is that a known problem?
> 
> It is not known to me :-)

I made a new ticket for that: https://unicode-org.atlassian.net/browse/CLDR-18511

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
btangmu: Thanks! I'll follow up with this soon

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
btangmu: Good to know -- I ended up using the "-1" option for "singleThread" to avoid crashing without "-e". I'm assuming for now at least we want warnings rather than errors.

btangmu:
The third commit fixes the 6 false-alarms bal_Latn -> bal_Latn_PK, etc., by calling `LikelySubtags.maximize`. I'm making this ready for review.

pedberg-icu (approved):

pedberg-icu:
> The third commit fixes the 6 false-alarms bal_Latn -> bal_Latn_PK, etc., by calling `LikelySubtags.maximize`.

Oh, that is a better idea, I did not know that we had that in CLDR.

PR #4597, opened 2025-04-09
CLDR-18190 Add Bakhtiari

conradarcturus:
CLDR-18190

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/bqi.xml
+++ b/common/main/bqi.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
+<!-- Copyright © 1991-2025 Unicode, Inc.
+For terms of use, see http://www.unicode.org/copyright.html
+SPDX-License-Identifier: Unicode-3.0
+CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="bqi"/>
+	</identity>
+	<layout>
+		<orientation>
+			<characterOrder>right-to-left</characterOrder>
+		</orientation>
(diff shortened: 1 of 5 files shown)


DavidLRowe:
I expect lineOrder should be top-to-bottom (or omitted since that is the default). I don't think you want lines going RTL on the page.

I was a little surprised to see the addition of a variant name to en.xml, but maybe that's okay since the base name is already there?

I don't have a lot of experience with Arabic script exemplars, but these seem comparable to what's in ar.xml and the tests are passing.

common/main/bqi.xml
```diff
@@ -0,0 +1,26 @@
...
+-->
+<ldml>
+	<identity>
+		<version number="$Revision$"/>
+		<language type="bqi"/>
+	</identity>
+	<layout>
+		<orientation>
+			<characterOrder>right-to-left</characterOrder>
+			<lineOrder>right-to-left</lineOrder>
```
roozbehp, line 16: Drop lineOrder. fa.xml doesn't have it either. I think what you put here means vertical text.

roozbehp (changes requested):
lineOrder must be dropped. Looks good otherwise.

roozbehp:
We also probably need a test to fail when lineOrder and characterOrder conflict like this. The LDML spec clearly says "If the value of lineOrder is one of the vertical values, then the value of characterOrder must be one of the horizontal values, and vice versa."

conradarcturus:
oh! Thanks for the catch.

roozbehp (approved):
LGTM

DavidLRowe:
@srl295 following up on this comment by @roozbehp :

> We also probably need a test to fail when lineOrder and characterOrder conflict like this. The LDML spec clearly says "If the value of lineOrder is one of the vertical values, then the value of characterOrder must be one of the horizontal values, and vice versa."

Does we need a ticket for a new test?

PR #4598, opened 2025-04-10
CLDR-18233 Remove language families from likely subtags

conradarcturus:
I noticed there are language families in the likely subtags data when we import data from Script-Metadata and SIL. It's caused because of script imports -- which we like but in this case the families are too general. The 2 items removed are for the families `pra` Prakrit and `zhx` Chinese -- they are written in a lot of scripts and a lot of regions. It's not well-defined to say which script or region is the likely one, so I am removing the language families from the generated file.

SIL inherited this error from CLDR and is passing it back to CLDR, so this also helps prevent that erroneous feedback loop.

Note that I did this in 2 commits. The first commit deletes dead code and fixes the spelling of Hierarchy, it also updates a language name that wasn't previously updated by running the likely subtag generation script. The second commit contains the functional changes.

CLDR-18233

Script to generate XML:
```
mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -890,7 +890,6 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="zh_Bopo" to="zh_Bopo_TW"/>		<!--Chinese‧Bopomofo‧?	➡ Chinese‧Bopomofo‧Taiwan-->
 		<likelySubtag from="zh_Hanb" to="zh_Hanb_TW"/>		<!--Chinese‧Han with Bopomofo‧?	➡ Chinese‧Han with Bopomofo‧Taiwan-->
 		<likelySubtag from="zh_Hant" to="zh_Hant_TW"/>		<!--Chinese‧Traditional‧?	➡ Chinese‧Traditional‧Taiwan-->
-		<likelySubtag from="zhx" to="zhx_Nshu_CN"/>		<!--Chinese (family)‧?‧?	➡ Chinese (family)‧Nüshu‧China-->
 		<likelySubtag from="zkt" to="zkt_Kits_CN"/>		<!--Kitan‧?‧?	➡ Kitan‧Khitan small script‧China-->
 		<likelySubtag from="zlm" to="zlm_Latn_MY"/>		<!--Malay (individual language)‧?‧?	➡ Malay (individual language)‧Latin‧Malaysia-->
 		<likelySubtag from="zmi" to="zmi_Latn_MY"/>		<!--Negeri Sembilan Malay‧?‧?	➡ Negeri Sembilan Malay‧Latin‧Malaysia-->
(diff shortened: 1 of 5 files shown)


srl295:
FYI @emily-roth

common/supplemental/likelySubtags.xml
```diff
@@ -1296,7 +1295,6 @@ not be patched by hand, as any changes made in that fashion may be lost.
 		<likelySubtag from="und_Newa" to="new_Newa_NP"/>		<!--?‧Newa‧?	➡ Newari‧Newa‧Nepal-->
 		<likelySubtag from="und_Nkoo" to="man_Nkoo_GN"/>		<!--?‧N’Ko‧?	➡ Mandingo‧N’Ko‧Guinea-->
 		<likelySubtag from="und_Nkoo_ML" to="bm_Nkoo_ML"/>		<!--?‧N’Ko‧Mali	➡ Bambara‧N’Ko‧Mali-->
-		<likelySubtag from="und_Nshu" to="zhx_Nshu_CN"/>		<!--?‧Nüshu‧?	➡ Chinese (family)‧Nüshu‧China-->
```
srl295, line 1299: hm. is there potentially a better target here?

emily-roth:
Thanks, I'll go ahead and make the same changes in the staging of SLDR.

srl295 (dismissed):

emily-roth:
The tests do not seem to like that und_Nshu doesn't have a language attached to it anymore. It seems like one of those two files with the Nshu script tag was somewhat load-bearing; as a way to identify a fallback language to use for the script maybe?


`(LanguageTest.java:167) Error: Script without likely language: Nshu Nüshu`

 `(LikelySubtagsTest.java:445)  Error: likelySubtags.xml missing language for script (und_Nshu). Script Metadata suggests that it should be something like:	 {"und_Nshu", "zhx_Nshu_CN"}`

conradarcturus:
Ah, I'll have to update the test to except this writing system. [Nüshu](https://en.wikipedia.org/wiki/N%C3%BCshu) is an historic writing system for [Xiangnan Tuhua](https://en.wikipedia.org/wiki/Xiangnan_Tuhua). There does not seem to be consensus on language code the dialect should apply to. There is no [ISO-639 change request](https://iso639-3.sil.org/code_changes/change_request_index/data?title=&field_change_instance_af_idnt_cd_value=&field_change_instance_afctd_name_value=&field_change_request_region_grp_tid=All&field_change_request_lf_group_tid=138986&field_change_instance_chnge_type_tid=All&field_change_instance_act_status_tid=All&items_per_page=50) to add a language code for it. Without a language code, "zhx" for "the family of Chinese languages" was used instead.

From wikipedia:
> Some scholars classify it under [Xiang Chinese](https://en.wikipedia.org/wiki/Xiang_Chinese) or [Pinghua](https://en.wikipedia.org/wiki/Pinghua), and other scholars consider it a hybrid dialect.[[3]](https://en.wikipedia.org/wiki/Xiangnan_Tuhua#cite_note-FOOTNOTEZhao2006162-3) Most Jiangyong residents are bilingual in Xiangnan Tuhua and the Hunan dialect of [Southwestern Mandarin](https://en.wikipedia.org/wiki/Southwestern_Mandarin).[[3]](https://en.wikipedia.org/wiki/Xiangnan_Tuhua#cite_note-FOOTNOTEZhao2006162-3)[[4]](https://en.wikipedia.org/wiki/Xiangnan_Tuhua#cite_note-FOOTNOTEChiang199522-4) Xiangnan Tuhua was only written using [Nüshu](https://en.wikipedia.org/wiki/N%C3%BCshu),[[5]](https://en.wikipedia.org/wiki/Xiangnan_Tuhua#cite_note-FOOTNOTEChiang199520-5) and Nüshu was not used to write other languages, such as the Southwestern Mandarin spoken in Hunan, or the local [Yao language](https://en.wikipedia.org/wiki/Hmong%E2%80%93Mien_languages).[[6]](https://en.wikipedia.org/wiki/Xiangnan_Tuhua#cite_note-FOOTNOTEZhao2006247-6)[[4]](https://en.wikipedia.org/wiki/Xiangnan_Tuhua#cite_note-FOOTNOTEChiang199522-4)

srl295:
phx maybe seems like the right answer.

I would recommend instead just adding an exception for Nüshu and makign it a known issue in the test (with a followup issue).

conradarcturus:
> phx maybe seems like the right answer.

What's `phx`?

> with a followup issue

What is the followup? I don't think there is any better language code we can use for Nüshu. It's just a fact its a script without a corresponding ISO 639-3 language.

conradarcturus:
Updated LikelySubtagsTest -- fortunately there already is code for adding overrides. While I was there, I was able to remove some overrides that were no longer necessary.

```
mvn test --file tools/pom.xml '-Dorg.unicode.cldr.unittest.testArgs=-f:LikelySubtagsTest -n -q'
```

srl295:
> KNOWN_SCRIPTS_WITHOUT_LIKELY_SUBTAGS

perfect.

srl295 (approved):

PR #4600, opened 2025-04-12
CLDR-18495 Set Kurdish (Syria) time format to follow Turkey

conradarcturus:
Add this override since ku_SY follows the ku.xml hour pattern not the standard Turkish hour pattern. Confirmed by our Kurdish submitters.

CLDR-18495

ALLOW_MANY_COMMITS=true


pedberg-icu (approved):
Looks great!

PR #4601, opened 2025-04-13
CLDR-18144 Add to languagelocale-names.md for core/extension

macchiati:
CLDR-18144

ALLOW_MANY_COMMITS=true


docs/site/translation/displaynames/languagelocale-names.md
```diff
@@ -35,7 +35,37 @@ For languages that are part of a larger family, the Survey Tool may request tran
...
-## Other variants
+### Core and Extensions
 
-Some languages may have other variant forms. For example, “ckb” may in English be called “Central Kurdish” or “Sorani Kurdish”; the former is used as the standard name for English, and the latter is the variant. In other languages the equivalent of “Sorani Kurdish” may be used as the standard name; if there is also an equivalent for “Central Kurdish” it may be supplied as the variant. If there is only one form in your language, please use it for both the standard and the variant form.
+Sometimes languages and other fields need further qualitification in the context of a menu. For example, when there is only one variant of French shown in a menu, it is fine to just list the language as follows.
+* English
+* French
+* German
 
+However, when there are multiple variants, it is clearer to list each one, such as:
```
conradarcturus, line 45: Minor copy-edit to clarify pronouns but you can leave as-is if you like.
```suggestion
However, when there are multiple variants, it is clearer to show the variant information, such as:
```

docs/site/translation/displaynames/languagelocale-names.md
```diff
@@ -35,7 +35,37 @@ For languages that are part of a larger family, the Survey Tool may request tran
...
+* French (Canada)
+* French (France)
+* German
+
+When the variants differ not by region or script, but by another qualifier, then you may see codes with "-core" or "-extension", as in the following.
+
+<img src="kurdish-example.png" />
+
+In these cases, supply the translated name of the qualifier (Kurmanji), and of the core.
+The core should omit the qualification (if any), and the extension should include it. For example, for Sorani Kurdish, you would see:
```
conradarcturus, line 56: The algorithm in prose is a bit hard to follow. This makes me wonder why we have the base (plain `ckb` value) in the first place -- for versions that don't support algorithms?

1. Check if _lang_-core exists
   a. Use _lang_-core if unique
   b. Use _lang_-core (_lang_-extension) if it conflicts and extension clarifies
   c. Use _lang_-core (_region_) if it conflicts and the region clarifies
2. Otherwise, use plain _lang_
   a. Use _lang_ (_region_) if it conflicts and the region clarifies

docs/site/translation/displaynames/languagelocale-names.md
```diff
@@ -35,7 +35,37 @@ For languages that are part of a larger family, the Survey Tool may request tran
...
+* French
+* German
 
+However, when there are multiple variants, it is clearer to list each one, such as:
+* English
+* French (Canada)
+* French (France)
+* German
+
+When the variants differ not by region or script, but by another qualifier, then you may see codes with "-core" or "-extension", as in the following.
```
conradarcturus, line 51: You should probably also indicate that variants may appear with different language codes. It's not a matter of variants for Kurdish since the language codes are "ku" and "ckb" -- rather the issue is menu item collision since the core for each is "Kurdish".

However, Kurdish may be a bad example since the languages are so different (and ku is in Latin script, ckb is in Arabic script) it may always be preferable to show Kurdish (Kurmanji) and Kurdish (Sorani).

conradarcturus (commented):
Thanks for updating the documentation. I left some notes but I'm happy to accept as-is if you are happy with it.

docs/site/translation/displaynames/languagelocale-names.md
macchiati: This isn't intended to be an algorithm for when to show what — that will be in the spec, and is too detailed for most vetters. There is an outline of that in the bug (but again, not appropriate for vetters).

The text is just to (1) set the expectation for the vetter as to what the appearance is if the regular value is used vs what it is when the core + extension is used, and (2) give an example of when the core + extension could appear.

I'll play with the wording a bit.

docs/site/translation/displaynames/languagelocale-names.md
macchiati: I do this in the example now.

macchiati: I also revised the other text; please see if that is clear enough.

macchiati:
Could I please get an approval so it can be merged? We can always refine later.

DavidLRowe (approved):
RSLGTM

PR #4604, opened 2025-04-15
CLDR-17254 site: add Keyboard Process Document to site

srl295:
CLDR-17254

We'd like to get this posted to the site so that we can start pointing people at the process.  It's not enough for keyboards to start coming in, we are working on other pieces of that.

ALLOW_MANY_COMMITS=true


mcdurdin (approved):
LGTM

srl295:
this is what was approved by the CLDR TC, we can continue to refine it.

PR #4615, opened 2025-04-16
CLDR-18518 abstracts: work around CORS, improve logging on ui

srl295:
CLDR-18518

  - example, az-short https://st.unicode.org/cldr-apps/v#/ab/Languages_A_D/700f7be0e77ae39b had no abstract but az did
  - fix the error message on deferHelp (dbPedia extract)
  - add logError, logException, logInfo, logWarning to cldrNotify for low level datadog logging without showing a message popup (these will go to datadog)
  - cleanup some noise in initial gui load and ajax load (Browser debugger panel for network shows better information here anyway)
  - workaround CORS issue with dbpedia by adding the hostname of the request to the query, this way the cache is individual for, say, smoketest vs. production (vs local) - otherwise they aren't shareable.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
@@ -2,6 +2,8 @@
  * cldrDeferHelp: encapsulate code related to showing language descriptions in the Info Panel
  */
 import { marked } from "./cldrMarked.mjs";
+import * as cldrNotify from "./cldrNotify.mjs";
+
 const defaultEndpoint = "https://dbpedia.org/sparql/";
 const format = "JSON";
 const abstractLang = "en";
@@ -56,15 +58,18 @@ function subloadAbstract(resource) {
 
   // This query simply returns the abstract result for the specific resource.
   // The query is so small that browsers persistently cache the GET request.
+  //
+  // We include the hostname so that the cache is specific to this instance.
   sparqlQuery(
     `PREFIX  dbo:  <http://dbpedia.org/ontology/>
-  PREFIX foaf: <http://xmlns.com/foaf/0.1/>
-  SELECT ?abstract ?primaryTopic
-  WHERE {
-      <${resource}> dbo:abstract ?abstract .
-      <${resource}> foaf:isPrimaryTopicOf ?primaryTopic
-      FILTER langMatches(lang(?abstract), "${abstractLang}")
-  } LIMIT 1`
+PREFIX foaf: <http://xmlns.com/foaf/0.1/>
+SELECT ?abstract ?primaryTopic
+WHERE {
+    <${resource}> dbo:abstract ?abstract .
+    <${resource}> foaf:isPrimaryTopicOf ?primaryTopic
+    FILTER langMatches(lang(?abstract), "${abstractLang}")
+} LIMIT 1
+# Query from ${window.location.hostname} - cldr.unicode.org`
   ).then(
     ({ results }) => {
       let seeAlso = resource;
@@ -85,7 +90,12 @@ function subloadAbstract(resource) {
       );
     },
     (err) => {
-      absContent.text(`Err loading ${resource}: ${err}`);
+      cldrNotify.logException(err, `Loading ${resource}`);
+      absContent.text(
+        `Error: Could not load ${resource} : ${
+          err?.status || err?.message || ""
+        }`
+      );
     }
   );
 
--- a/tools/cldr-apps/js/src/esm/cldrNotify.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrNotify.mjs
@@ -51,6 +51,14 @@ function open(message, description, onClick) {
   });
 }
 
+function logInfo(message, description) {
+  if (hasDataDog) {
+    datadogLogs.logger.info(message, { description });
+  } else {
+    console.log(message);
+  }
+}
+
 /**
  * Display a warning notification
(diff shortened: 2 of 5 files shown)


btangmu (approved):

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
```diff
@@ -56,15 +58,18 @@ function subloadAbstract(resource) {
...
-      FILTER langMatches(lang(?abstract), "${abstractLang}")
-  } LIMIT 1`
+PREFIX foaf: <http://xmlns.com/foaf/0.1/>
+SELECT ?abstract ?primaryTopic
+WHERE {
+    <${resource}> dbo:abstract ?abstract .
+    <${resource}> foaf:isPrimaryTopicOf ?primaryTopic
+    FILTER langMatches(lang(?abstract), "${abstractLang}")
+} LIMIT 1
+# Query from ${window.location.hostname} - cldr.unicode.org`
```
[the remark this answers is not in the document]
srl295: > I had to study this a long time.

sorry, let's make it clearer!

>  I'm not sure if I understand it; here's my theory. The browser may cache the response to the request, keyed to the request including the query string. Including hostname in a comment in the query causes the browser to cache responses separately for each hostname (if it caches them at all). It should be safe to assume that the browser doesn't parse the query (and strip comments); only the endpoint parses it. Bypassing the cache in this way prevents CORS errors.

True, but not the entire story. The browser caches the **Cors OPTIONS query**.  What seems to be happening is that before the browser does the request, it does an OPTIONS query for that exact URL.   dbpedia "helpfully" puts whatever you put in the Referer header into their CORS allowable hosts.

however, then if smoketest vs production makes the exact same query, the CORS OPTIONS are cached (not talking about the content) and the browser says, Nope. the client is not on the allow list, so is denied. it doesn't retry the OPTIONS call, assuming its cacheable.

So yes. we break the caching so that it makes a separate OPTIONS call (invisible to you) that has the correct allow header, and it works.

A  side effect is that it does not cache the GET response either. whereas before it would have shared them (this data is cached quite aggressively). but that is a minor inconvenience.
 
> Is that right? If so I suggest expanding the comment. "We include the hostname so that the cache is specific to this instance" doesn't make it clear that the purpose is to prevent CORS errors, that it's a _browser_ cache, and that the hostname is in a _comment_ in the query. Not being familiar with sparql, I wasted some time wondering if the endpoint parses comments like "# Query from ..." and interprets them.

It's a comment. Yeah, i'll expand the description.
 
> Also, how about moving the whole query string to a named constant `abstractLangQuery` at the top of the file along with `defaultEndpoint`, `format`, and `abstractLang`? That would make the function more legible, especially given the indentation.
 
Will do.

> What's the reason for including "cldr.unicode.org"?

in case dbpedia wonders what these queries are from, since they come from clients.

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
```diff
@@ -110,7 +96,29 @@ WHERE {
  */
 function sparqlQuery(query, endpoint) {
   endpoint = endpoint || defaultEndpoint;
-  return $.getJSON(endpoint, { query, format });
+  const url = new URL(endpoint);
+  url.searchParams.append("query", query);
+  url.searchParams.append("format", "JSON");
+  return fetch(url);
```
btangmu, line 102: I like to use `cldrAjax.doFetch` but I guess it's not appropriate for a non-unicode.org server. We might consider a new method cldrAjax for other servers. It's good to go thru a single module for http.

btangmu (dismissed):

tools/cldr-apps/js/src/esm/cldrDeferHelp.mjs
srl295: That would leak the session key! I would be fine with a generic fetch, it could even just be re-exported fetch() would that work

btangmu (approved):

tools/cldr-apps/js/src/esm/cldrNotify.mjs
```diff
@@ -51,33 +51,51 @@ function open(message, description, onClick) {
...
+ *
+ * @param {String} message the title
+ * @param {String} description the more detailed description
+ */
+function logWarning(message, description) {
+  if (hasDataDog) {
+    datadogLogs.logger.warn(message, { description });
+  } else {
+    console.warn(message);
+  }
```
btangmu, line 88: When we're testing on production or cldr-smoke or cldr-staging, it might still be helpful if warnings/errors would show up in the browser console, even if they also show up in datadog

tools/cldr-apps/js/src/esm/cldrNotify.mjs
```diff
@@ -101,12 +131,22 @@ function errorWithCallback(message, description, callback) {
...
  * @param {Object|String} e the Error that was thrown and caught
  * @param {String} context a description of where it was caught
  */
 function exception(e, context) {
+  e = logException(e, context);
+  notification.error({
+    message: "Internal error: " + e.name + " " + context,
+    description: e.message,
+    duration: NO_TIMEOUT,
+  });
```
btangmu, line 145: Just a thought for the future: we might in general have two error messages, one for ordinary users and one including "technical details", and the latter might be available by clicking a button for "technical details"

btangmu (approved):

PR #4626, opened 2025-04-22
CLDR-14996 Test for forbidden time formats in en_001

btangmu:
CLDR-14996

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEn001Time.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEn001Time.java
@@ -7,12 +7,17 @@ import org.unicode.cldr.util.*;
 
 public class TestEn001Time {
 
+    // TODO: for short timezone names, check that en_001 has “∅∅∅” iff en has a value, and
+    // that en_001 otherwise has no short timezone names (except through inheritance).
+    // Reference: https://unicode-org.atlassian.net/browse/CLDR-14996
+    static final boolean CHECK_SHORT_TIMEZONE_NAMES = false;
     static final String TIME_FORMAT_PREFIX =
             "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/timeFormats/timeFormatLength";
+
     static final String TIMEZONE_NAME_PREFIX = "//ldml/dates/timeZoneNames";
     static final String TIMEZONE_NAME_SHORT = "short";
 
-    /** en_001 should not have standard time formats or any short timezone names */
+    /** en_001 should not have standard time formats */
     @Test
     void testTime() {
         Factory factory = CLDRConfig.getInstance().getCommonAndSeedAndMainAndAnnotationsFactory();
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEn001Time.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+import org.junit.jupiter.api.Test;
+import org.unicode.cldr.util.*;
+
+public class TestEn001Time {
+
+    static final String TIME_FORMAT_PREFIX =
+            "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/timeFormats/timeFormatLength";
```
btangmu, line 11: Note that the ticket refers to "standard time formats". I'm not sure whether that applies to any calendars other than Gregorian; currently only Gregorian is checked. Also, I'm not sure whether all paths starting with this prefix qualify as "standard". This part of the test currently passes. The test could be revised later if and when the need arises to add paths that would fail.

btangmu: Currently, en_001.xml contains no paths starting with "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/timeFormats/timeFormatLength".

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEn001Time.java
```diff
@@ -0,0 +1,34 @@
...
+                    path.startsWith(TIME_FORMAT_PREFIX),
+                    "Path starts with " + TIME_FORMAT_PREFIX + ": " + path);
+            assertFalse(
+                    path.startsWith(TIMEZONE_NAME_PREFIX) && path.contains(TIMEZONE_NAME_SHORT),
+                    "Path starts with "
+                            + TIMEZONE_NAME_PREFIX
+                            + " + and contains "
+                            + TIMEZONE_NAME_SHORT
+                            + ": "
+                            + path);
```
btangmu, line 31: Currently en_001.xml does contain the path "//ldml/dates/timeZoneNames/metazone[@type=\"Hawaii_Aleutian\"]/short/daylight".

https://cldr-smoke.unicode.org/cldr-apps/v#/en_001/NAmerica/39b5dd8bc9bd28b2

Its value is “∅∅∅”.

The ticket description includes, “Check for short timezone names (other than GMT/UTC) and exclude”. Should inclusion of this path with value “∅∅∅” trigger a test failure? Testing locally, this new test does fail on that path.

btangmu: Also note that the test doesn't (yet) treat GMT/UTC differently from other zones

btangmu: As expected, the test fails:
` Error:    TestEn001Time.testTime:24 Path starts with //ldml/dates/timeZoneNames + and contains short: //ldml/dates/timeZoneNames/metazone[@type="Hawaii_Aleutian"]/short/daylight ==> expected: <false> but was: <true>`

Should I make another commit that removes the path in question from en_001.xml? Or, revise the test to allow the value “∅∅∅”?

I just found related documentation here:

https://github.com/unicode-org/cldr/blob/main/docs/site/development/development-process/design-proposals/english-inheritance.md

"Time zone names: en_001 should us “∅∅∅” to cancel the short metazone names inherited from en, for example:"

tools/cldr-code/src/test/java/org/unicode/cldr/test/TestEn001Time.java
pedberg-icu: Yes, and we want to keep it that way.

pedberg-icu (approved):
Loos good as far as it goes (so far it only addresses one of the two tests requested in the ticket, as noted - completing the other is a TODO).

PR #4628, opened 2025-04-22
CLDR-18516 GenerateBirths v48

srl295:
CLDR-18516

ALLOW_MANY_COMMITS=true


macchiati (approved):
Approving, but note that the tool also generates human readable files that can be checked in on CLDR staging for review.

PR #4629, opened 2025-04-22
CLDR-18464 Fix IndexOutOfBoundsException in ExampleGenerator.handleEras

pedberg-icu:
CLDR-18464

Addition of the new islamic era caused an IndexOutOfBoundsException in ExampleGenerator.handleEras, since it uses a hard-coded map to a list of sample dates for each era and it did not have an entry for the new islamic era. Fixed by adding the entry.

ALLOW_MANY_COMMITS=true


btangmu (approved):

sffc (commented):
LGTM

PR #4060, opened 2024-09-17
CLDR-5854 Add-examples-for-relative-dates-and-times

haytenf:
CLDR-5854

| Locale               | Before               | After |
| -------------- | -------------- |  -------------- |
| cy |  Screenshot 2024-08-27 at 4 33 56 PM |  Screenshot 2024-08-27 at 4 34 32 PM |
| cy |  Screenshot 2024-08-27 at 4 35 20 PM | Screenshot 2024-08-27 at 4 50 53 PM |
| cy | Screenshot 2024-08-27 at 4 36 47 PM | Screenshot 2024-08-27 at 4 37 19 PM |
| cy | Screenshot 2024-08-27 at 4 43 45 PM | Screenshot 2024-08-27 at 4 46 19 PM |
| cy | Screenshot 2024-08-27 at 4 53 20 PM | Screenshot 2024-08-27 at 4 39 28 PM |
| ro | Screenshot 2024-08-27 at 4 56 03 PM |Screenshot 2024-08-27 at 4 57 18 PM |
| ro | Screenshot 2024-08-27 at 4 56 21 PM |Screenshot 2024-08-27 at 4 57 54 PM |
| ro | Screenshot 2024-08-27 at 5 02 07 PM | (future-few) Screenshot 2024-08-27 at 5 01 05 PM |
| ro | Screenshot 2024-08-27 at 4 59 38 PM  | Screenshot 2024-08-27 at 5 03 43 PM |
| it | Screenshot 2024-08-27 at 5 05 34 PM | Screenshot 2024-08-27 at 5 06 23 PM |
| it | Screenshot 2024-08-27 at 5 06 51 PM  | Screenshot 2024-08-27 at 5 07 40 PM |

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4630, opened 2025-04-22
CLDR-18025 Relative examples should use locale-appropriate num value/format and contextTransform

pedberg-icu:
CLDR-18025

This is a followup to PR [#4060](https://github.com/unicode-org/cldr/pull/4060) to address issues mentioned in (post-merge) review comments there, and other issues:
- The numeric values used as examples for different plural forms need to depend on locale; especially for the few, many and other cases we need a locale-specific curated value.
- The two examples need to illustrate mid-sentence and stand-alone (not beginning of sentence) capitalization, with the latter determined by contextTransform data, and with guidance on how to fix it if wrong. Any titlecasing needs to be done in a locale-appropriate way.
- Numeric values need to be formatted in a locale-appropriate way.

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -3230,25 +3251,74 @@ private void handleRelative(
...
+                    // example at the end of the list
+                    exampleCount = exampleCounts.get(exampleCountSize - 1);
+                } else {
+                    Map<Count, String> countToStringExamplesMap =
+                            pluralInfo.getCountToStringExamplesMap();
+                    String stringExamples = countToStringExamplesMap.get(Count.valueOf(count));
+                    // skip the default value already set
+                    for (int i = 1; i < exampleCountSize; i++) {
+                        String exampleCountTest = exampleCounts.get(i);
+                        if (stringExamples.contains(exampleCountTest)) {
```
pedberg-icu, line 3301: This is a little hokey (searching for a string value in the string of sample values) but it works with the values used here. Could also convert the string value to a number and get the count from PluralRules and see whether that matches.

pedberg-icu:
@srl295 @macchiati Ping? This is something that might help vetters...

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -344,19 +345,39 @@ public void setCachingEnabled(boolean enabled) {
...
+                                    "21" /*kw*/,
+                                    "0.5" /*cs sk lt*/));
+                    put(
+                            "other",
+                            List.of(
+                                    "100" /*ar cs de en es fr he it*/,
+                                    "22" /*lv prg*/,
+                                    "23" /*gv*/,
+                                    "14" /*ceb fil tl*/,
+                                    "0.5" /*pl ru uk*/));
```
srl295, line 375: Couldn't this call `Map<Count, Set<Double>> SupplementalDataInfo.PluralInfo.getCountToExamplesMap()` ?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathDescriptions.txt
```diff
@@ -193,14 +193,14 @@
...
+^//ldml/dates/fields/field\[@type="([^"]*)"]/relative\[@type="-1"]; Provide a name for “last {1}”. For more information, please see %DATE_TIME_NAMES_RELATIVE. The lettercasing should be appropriate for the top example. If the lettercasing is then wrong for the bottom example, please file a ticket to fix contextTransforms/relative/stand-alone.
+^//ldml/dates/fields/field\[@type="([^"]*)"]/relative\[@type="1"]; Provide a name for “next {1}”. For more information, please see %DATE_TIME_NAMES_RELATIVE. The lettercasing should be appropriate for the top example. If the lettercasing is then wrong for the bottom example, please file a ticket to fix contextTransforms/relative/stand-alone.
+^//ldml/dates/fields/field\[@type="([^"]*)"]/relativeTime\[@type="future"]/relativeTimePattern\[@count="([^"]*)"]; Provide a pattern used to display times in the future. For more information, please see %DATE_TIME_NAMES. The lettercasing should be appropriate for the top example. If the lettercasing is then wrong for the bottom example, please file a ticket to fix contextTransforms/relative/stand-alone.
```
srl295, line 202: good, i've conflicted with it in #4633 but that's OK

srl295 (approved):
LGTM with suggestions

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
pedberg-icu: But then we still need to pull one example from the set, and what I was trying to do here was find examples that would contrast in an obvious way with the examples for other plural forms (and that could also be used across multiple locales), instead of e.g. just pulling the first example from the set returned by `SupplementalDataInfo.PluralInfo.getCountToExamplesMap`. Anyway I can explore other options in a future PR, this PR at least gets correct behavior in for the vetters, even if the code is not as clean as it could be.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathDescriptions.txt
pedberg-icu: Sorry about that

PR #4631, opened 2025-04-23
CLDR-13555 Add Pali locale

conradarcturus:
CLDR-13555

This PR starts the process to support Pali in CLDR.

There was significant debate for the default script & region but we have moved that resolution to a separate ticket https://unicode-org.atlassian.net/browse/CLDR-18545

ALLOW_MANY_COMMITS=true


common/supplemental/supplementalData.xml
```diff
@@ -1893,8 +1893,7 @@ XXX Code for transations where no currency is involved
 		<language type="peo" scripts="Xpeo"/>
 		<language type="pfl" scripts="Latn"/>
 		<language type="phn" scripts="Phnx"/>
-		<language type="pi" scripts="Mymr"/>
-		<language type="pi" scripts="Sinh Deva Thai" alt="secondary"/>
+		<language type="pi" scripts="Deva Mymr Sinh Thai" alt="secondary"/>
```
AEApple, line 1896: do only the secondary scripts get listed in this section of the supplementalData?

macchiati (approved):

PR #4632, opened 2025-04-23
CLDR-18546 Format XML source using xmlstarlet

sffc:
CLDR-18546

I generated this PR with the following command: `find . -name *.xml -exec xmlstarlet ed --inplace --ps {} \;`

ALLOW_MANY_COMMITS=true


sffc:
@srl295 said on Jira:

> wrong time in the cycle. If we do this it should be after “merge from vxml” months from now
>
> wrong part of the toolchain. These files are generated - output should be changed in CLDRFile etc. (see CLDR-14496)

PR #4633, opened 2025-04-23
CLDR-17829 move PathDescriptions.txt to PathDescriptions.md

srl295:
- add lychee build task to link check PathDescriptions.md
- added PathDescriptionParser and unit tests
- fixes many descriptions, link targets, and placeholders

CLDR-17829

ALLOW_MANY_COMMITS=true


srl295:
Note: hard build failure if PathDescriptions.md contains an invalid link:

image

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
```diff
@@ -93,22 +106,32 @@ public enum Status {
...
+            markdown = MISSING_DESCRIPTION;
+            description = null;
+        } else {
+            description = entry.getFirst();
+            markdown = entry.getSecond();
+        }
+
+        if (description == null || description.isEmpty()) {
             description = MISSING_DESCRIPTION;
-        } else if ("SKIP".equals(description)) {
```
macchiati, line 111: If it starts with SKIP, but has other characters, it is an error; so best to raise an exception.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestPathDescription.java
```diff
@@ -1,13 +1,34 @@
...
-        RegexLookup<String> l = PathDescription.parseLookupString();
-        assertNotNull(l);
+        final PathDescriptionParser parser = new PathDescriptionParser();
+        RegexLookup<Pair<String, String>> lookup =
+                parser.parse(PathDescription.getPathDescriptionString());
+        assertNotNull(lookup);
+        assertFalse(lookup.size() == 0, "lookup is empty");
+        String references = parser.getReferences();
+        assertNotNull(references);
+        assertFalse(parser.getReferences().trim().isBlank());
```
macchiati, line 20: Very minor: .trim().isBlank() is redundant, I think.

macchiati: BTW: we should check all of our .md files for URLs containing "#h.", and raise errors when found. Do you want me to file a separate tickets?

srl295: yes would be separate -- could just be a search though?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathDescriptions.md
```diff
@@ -0,0 +1,1115 @@
...
+-->
+
+[Characters]: https://cldr.unicode.org/translation/characters
+[Character Labels]: https://cldr.unicode.org/translation/characters/character-labels
+[Compound Units]: https://cldr.unicode.org/translation/units/unit-names-and-patterns#compound-units
+[Country Names]: https://cldr.unicode.org/translation/displaynames/countryregion-territory-names
+[Currency Names]: https://cldr.unicode.org/translation/currency-names-and-symbols
+[Date Time]: https://cldr.unicode.org/translation/date-time/date-time-names#h.ewzjebmpoi4k
+[Date Time Names]: https://cldr.unicode.org/translation/date-time/date-time-names
+[DATE_TIME_NAMES_CYCLIC]: https://cldr.unicode.org/translation/date-time/date-time-names#h.h0vy2eyzcj0n
```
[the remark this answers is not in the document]
srl295: thanks good catch. It's easiest to fix this with a refactoring (fn-F2 to rename the link)

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathDescriptions.md
```diff
@@ -0,0 +1,1115 @@
...
+[Date Time Fields]: https://cldr.unicode.org/translation/date-time/date-time-names#h.8wfk3599ck9x
```
macchiati, line 1089: When you looked up the #h. fragments to get the header names, you should have also fixed the fragment to be right. Right now it fails and just goes to the top of the page.
```suggestion
[Date Time Fields]: https://cldr.unicode.org/translation/date-time/date-time-names#date-field-names
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
srl295: the description here is more rich, it's for you to name (or in the future group) these entries.

as below, where First Day of the Week is a comment.

    ### First Day of the Week

    - `^//ldml/localeDisplayNames/types/type\[@key="fw"]\[@type="([^"]*)"]`

    The name of “first day of the week is {1}”. For more information, please see [Key Names].

macchiati (changes requested):
Mostly minor except for the "#h."

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathDescriptions.md
srl295: didn't look them up, just updated the titles.

srl295:
@macchiati some of the other targets don't seem to exist.

srl295:
"Cyclic Name Sets" isn't an anchor (defect in date-time-names.md )

srl295:
I thought this task was to move the format over, for future fixup. I will revisit this tomorow

srl295:
I'll add a test against the relevant sites page… Probably for a future PR

macchiati:
> I thought this task was to move the format over, for future fixup. I will revisit this tomorow

I just thought that as long as you were getting the anchor text, you could fix the links as well. It can be a separate PR though.

srl295:
> > I thought this task was to move the format over, for future fixup. I will revisit this tomorow
> 
> 
> 
> I just thought that as long as you were getting the anchor text, you could fix the links as well. It can be a separate PR though.

No, i had just title cased the variable names.

macchiati (dismissed):

macchiati:
I was thinking that for this file and similar ones, it would be useful to have a tool that generated a sample of each message, that would be used to build a web page; we could then review to make sure that all the parameters were right, etc.  

Not part of this ticket — I'll file a different one for it.

macchiati:
https://unicode-org.atlassian.net/browse/CLDR-18555

macchiati (approved):

srl295:
sorry, i was running the wrong test locally.

This also now fixes a number of cases where the placeholders `{}` were wrong and were being passed through to users. and adds a test so it doesn't happen again.

srl295:
> I was thinking that for this file and similar ones, it would be useful to have a tool that generated a sample of each message, that would be used to build a web page; we could then review to make sure that all the parameters were right, etc.
> 
> Not part of this ticket — I'll file a different one for it.

i recommend deduping cases where the message is the same for multiple xpaths.

macchiati:
> > I was thinking that for this file and similar ones, it would be useful to have a tool that generated a sample of each message, that would be used to build a web page; we could then review to make sure that all the parameters were right, etc.
> > Not part of this ticket — I'll file a different one for it.
> 
> i recommend deduping cases where the message is the same for multiple xpaths.

That's what I meant by one message per line, eg per regex path in the <key, values> pair.

macchiati (approved):

PR #4636, opened 2025-04-23
CLDR-18548 BRS 48 CLDRModify passes 2025-04-23

pedberg-icu:
CLDR-18548

I decided to go ahead and do the CLDRModify passes since they cleaned up some recent DDL locale updates:
- the no-options pass fixed formatting (spaces->tabs, comments) for `bqi bua_RU bua ku_Arab`
- the -fP pass reordered/reformatted exemplar sets for `blo bqi bua sgs`

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/blo.xml
+++ b/common/main/blo.xml
@@ -432,8 +432,8 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		</codePatterns>
 	</localeDisplayNames>
 	<characters>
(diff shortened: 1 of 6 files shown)


macchiati (approved):

DavidLRowe:
sgs.xml has an error:  Error: Auxiliary characters also exist in main: ‎[ā ē ī ō ū]‎
I believe that these five characters should be removed from the auxiliary exemplar. (I'm also curious whether the index exemplar is correct, or whether it should contain only unaccented capitals. But that could be a question for another time.)

pedberg-icu:
> sgs.xml has an error: Error: Auxiliary characters also exist in main: ‎[ā ē ī ō ū]‎ I believe that these five characters should be removed from the auxiliary exemplar. (I'm also curious whether the index exemplar is correct, or whether it should contain only unaccented capitals. But that could be a question for another time.)

@DavidLRowe Yes, I noticed that, I think the CLDRModify may have done some normalization that made it evident. But also: Even removing the duplicate NFC forms of those characters from the aux exemplars, there were still NFD forms of these same characters  in curly braces; I have removed those as well, since exemplar characters are supposed to be in NFC form (I guess the duplicate test doesn't catch NFD forms in brackets that duplicate NFC forms not in brackets. So in commit 3 here I have removed both sets of duplicate ‎[ā ē ī ō ū], and that fixes the test error. That still leaves one slight oddity in the main exemplars: the bracketed element in the sequence e{ė̄}ē. That bracketed element is actually { U+0117 U+0304 }, that is LATIN SMALL LETTER E WITH DOT ABOVE plus COMBINING MACRON (above the dot); is that intended? (If not we can remove it in a separate PR, because that is actually removing something from the exemplars that is not a dup of something already there).

PR #4639, opened 2025-04-24
CLDR-18463 Add exemplar city and language name information

AEApple:
CLDR-18463

Update Info Hub for language names and exemplar cities

ALLOW_MANY_COMMITS=true


macchiati (approved):
Very minor; would be nicer as a table. But not worth the effort now.

PR #4642, opened 2025-04-25
CLDR-17851 SupplementalCalendarData parser, update ExampleGenerator

srl295:
- minimal, modern parser
- minimal test case
- update example generator

CLDR-17851 

- Fixes original root cause of #4629 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -3104,10 +3104,10 @@ public class ExampleGenerator {
     private void handleEras(XPathParts parts, String value, List<String> examples) {
         String calendarId = parts.getAttributeValue(3, "type");
         String type = parts.getAttributeValue(-1, "type");
-        String id =
-                (calendarId.startsWith("islamic"))
-                        ? "islamic"
-                        : calendarId; // islamic variations map to same sample
+        String id = calendarId;
+        if (id.equals("generic") || id.equals("iso8601")) {
+            id = "gregorian"; // use Gregorian eras, 'generic' is not in the data
+        }
         final SupplementalCalendarData.CalendarData calendarData =
                 supplementalDataInfo.getCalendarData().get(id);
 
@@ -3123,21 +3123,48 @@ public class ExampleGenerator {
         GregorianCalendar startCal = forDateString(eraData.getStart());
         GregorianCalendar endCal = forDateString(eraData.getEnd());
 
-        if (startCal == null && endCal == null) return; // no data
+        final SupplementalCalendarData.EraData prevEra = calendarData.get(typeIndex - 1);
+        final SupplementalCalendarData.EraData nextEra = calendarData.get(typeIndex + 1);
 
-        if (startCal != null && endCal != null) {
-            // go 2 days before end
-            endCal.roll(Calendar.JULIAN_DAY, -2);
-            if (endCal.before(startCal)) {
-                endCal = startCal;
-            }
-        } else if (startCal == null) {
-            endCal.roll(Calendar.JULIAN_DAY, -2);
-        } else if (endCal == null) {
-            endCal = forDateString("2002-07-15"); // CLDR repo root commit
+        if (startCal == null && prevEra != null && prevEra.getEnd() != null) {
+            startCal = forDateString(prevEra.getEnd());
+            // shift forward so we are in the next era
+            startCal.setTimeInMillis(startCal.getTimeInMillis() + (DateConstants.MILLIS_PER_DAY));
+        }
+        if (endCal == null && nextEra != null && nextEra.getStart() != null) {
+            endCal = forDateString(nextEra.getStart());
+            // shift backward so we are in the prev era
+            endCal.setTimeInMillis(endCal.getTimeInMillis() - (DateConstants.MILLIS_PER_DAY));
         }
 
-        final Date sample = endCal.getTime();
+        GregorianCalendar sampleDate = null;
+
+        if (startCal != null && endCal != null) {
+            // roll back a day to not hit the edge
+            sampleDate = endCal;
+            sampleDate.setTimeInMillis(
+                    sampleDate.getTimeInMillis() - (DateConstants.MILLIS_PER_DAY));
+        } else if (startCal == null && endCal != null) {
+            // roll back a day to not hit the edge
+            sampleDate = endCal;
+            sampleDate.setTimeInMillis(
+                    sampleDate.getTimeInMillis() - (DateConstants.MILLIS_PER_DAY));
+        } else if (startCal != null && endCal == null) {
+            sampleDate = forDateString("2002-07-15"); // CLDR repo root commit
+            if (sampleDate.before(startCal)) {
+                sampleDate = startCal;
+                sampleDate.setTimeInMillis(
+                        sampleDate.getTimeInMillis() + (DateConstants.MILLIS_PER_DAY));
+            }
+        } else {
+            // System.err.println("No good date for " + eraData);
+            // TODO: should be an error in TestSupplementalDataInfo
+            sampleDate = null;
+        }
+
+        if (sampleDate == null) return; // could not find the time
+
+        final Date sample = sampleDate.getTime();
 
         String skeleton = "Gy";
         String checkPath =
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -3120,19 +3120,34 @@ public class ExampleGenerator {
         if (eraData == null) {
             return; // no era data
         }
-        GregorianCalendar startCal = forDateString(eraData.getStart());
-        GregorianCalendar endCal = forDateString(eraData.getEnd());
+        GregorianCalendar startCal = eraData.getStartCalendar();
+        GregorianCalendar endCal = eraData.getEndCalendar();
 
-        final SupplementalCalendarData.EraData prevEra = calendarData.get(typeIndex - 1);
-        final SupplementalCalendarData.EraData nextEra = calendarData.get(typeIndex + 1);
+        final SupplementalCalendarData.EraData eminusone = calendarData.get(typeIndex - 1);
+        final SupplementalCalendarData.EraData eplusone = calendarData.get(typeIndex + 1);
+
+        SupplementalCalendarData.EraData prevEra = null;
+        SupplementalCalendarData.EraData nextEra = null;
+
+        // see if we can find the 'prev' and 'next' era by date
+        if (eminusone != null && eminusone.compareTo(eraData) < 0) {
+            prevEra = eminusone;
+        } else if (eplusone != null && eplusone.compareTo(eraData) < 0) {
+            prevEra = eplusone;
+        }
+        if (eminusone != null && eminusone.compareTo(eraData) > 0) {
+            nextEra = eminusone;
+        } else if (eplusone != null && eplusone.compareTo(eraData) > 0) {
+            nextEra = eplusone;
+        }
 
         if (startCal == null && prevEra != null && prevEra.getEnd() != null) {
-            startCal = forDateString(prevEra.getEnd());
+            startCal = prevEra.getEndCalendar();
             // shift forward so we are in the next era
             startCal.setTimeInMillis(startCal.getTimeInMillis() + (DateConstants.MILLIS_PER_DAY));
         }
         if (endCal == null && nextEra != null && nextEra.getStart() != null) {
-            endCal = forDateString(nextEra.getStart());
+            endCal = nextEra.getStartCalendar();
             // shift backward so we are in the prev era
             endCal.setTimeInMillis(endCal.getTimeInMillis() - (DateConstants.MILLIS_PER_DAY));
         }
@@ -3150,7 +3165,7 @@ public class ExampleGenerator {
             sampleDate.setTimeInMillis(
                     sampleDate.getTimeInMillis() - (DateConstants.MILLIS_PER_DAY));
         } else if (startCal != null && endCal == null) {
-            sampleDate = forDateString("2002-07-15"); // CLDR repo root commit
+            sampleDate = new GregorianCalendar(2002, 6, 15); // CLDR repo root commit
             if (sampleDate.before(startCal)) {
                 sampleDate = startCal;
                 sampleDate.setTimeInMillis(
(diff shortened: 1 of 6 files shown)


srl295:
@sffc this gives us a data driven approach that should be less crashy on era data changes

sffc (commented):
I don't understand enough about how CLDR works to be able to give this a thorough review

srl295:
> I don't understand enough about how CLDR works to be able to give this a thorough review

mostly just a FYI that CLDR *could* now have data quality checks on `<calendarData>` (nothing, sadly, parsed it before)

srl295:
example generator issue, will bring back soon

srl295:
alright, BH now shows `AH 0` as its example.  Which is better than `AH 1` i guess.

anyway, everyone gets examples now

image

AEApple:
Thank you for doing this! I'm not familiar enough either to review this unfortunately, although I could try applying it in my branch and check if it fixed the error if that works?

srl295:
Already did and it does work

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,229 @@
+package org.unicode.cldr.util;
+
+import com.ibm.icu.impl.locale.XCldrStub.ImmutableMap;
```
macchiati, line 3: This is wrong; we should never have XCldrStub in any CLDR code.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,229 @@
...
+public class SupplementalCalendarData implements Iterable<String> {
+    /** an <era> element */
+    public static class EraData implements Comparable<EraData> {
+        static final int INDEX = 4; // index of our element
+
+        /** the <era> xpath */
+        private final XPathValue xpath;
+
+        EraData(final XPathValue xpath) {
+            this.xpath = xpath;
```
macchiati, line 21: In general, for sup. data we have parsed out the data from each path and cached it in a friendly data structure, rather than parsing on every single call to access data.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,229 @@
...
+            }
+        }
+
+        private GregorianCalendar getLatest() {
+            GregorianCalendar l = getEndCalendar();
+            if (l == null) l = getStartCalendar();
+            return l;
+        }
+
+        /** only works within the same cal system */
```
macchiati, line 86: This is then too fragile; add a calendar ID to the EraData, then compare first by the calendarID, then by the era num ber.

macchiati: BTW, ComparisonChain is less error-prone.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
macchiati: If you don't have time to do those changes now, please file a ticket to fix in the future.

macchiati (changes requested):
A few items

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: It actually doesn't re parse, the only overhead is a hash table look up on the frozen Xpath.

srl295: I was trying to keep the parser  simple as a model for others that need to be added

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: done

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: done

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: done

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -3151,30 +3100,87 @@ private void handleDateRangePattern(String value, List<String> examples) {
...
+        if (endCal == null && nextEra != null && nextEra.getStart() != null) {
+            endCal = nextEra.getStartCalendar();
+            // shift backward so we are in the prev era
+            endCal.setTimeInMillis(endCal.getTimeInMillis() - (DateConstants.MILLIS_PER_DAY));
+        }
+
+        GregorianCalendar sampleDate = null;
+
+        if (startCal != null && endCal != null) {
+            // roll back a day to not hit the edge
```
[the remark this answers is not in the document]
srl295: done

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,269 @@
...
+        public GregorianCalendar getEndCalendar() {
+            return endCalendar;
+        }
+
+        public String getCode() {
+            return code;
+        }
+
+        public String[] getAliases() {
+            final String s = xpath.getAttributeValue(INDEX, LDMLConstants.ALIASES);
```
macchiati, line 66: Didn't notice this earlier. That's an awkward way to get an empty string array. Also, we should avoid String::split.

Use a static Splitter with splitToList, and trimming, and change the return value to a List

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,269 @@
...
+        private final String calendarType; // for comparison
+        private final GregorianCalendar startCalendar;
+        private final GregorianCalendar endCalendar;
+
+        EraData(final XPathValue xpath) {
+            calendarType = xpath.getAttributeValue(2, LDMLConstants.TYPE);
+            type = Integer.parseInt(xpath.getAttributeValue(INDEX, LDMLConstants.TYPE));
+
+            start = xpath.getAttributeValue(INDEX, LDMLConstants.START);
+            startCalendar = forDateString(start);
```
macchiati, line 35: A Calendar is pretty heavy-weight for this. Would be simplest to just extract the millis.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,269 @@
...
+import com.ibm.icu.util.GregorianCalendar;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import org.unicode.cldr.icu.LDMLConstants;
+
+public class SupplementalCalendarData implements Iterable<String> {
+    /** an <era> element */
```
macchiati, line 16:
```suggestion
    /** an <era> element, which represents the era information from a calendar such as:
    * <pre>         
    * <calendarData><supplementalData>
    *    <calendar type="islamic">
    *      <calendarSystem type="lunar" />
    *      <eras>
    *          <era type="0" start="622-7-15" code="islamic" aliases="ah"/>
    *           <era type="1" end="622-7-14" code="islamic-inverse" aliases="bh"/>
    *      </eras>
    *   </calendar>
    * </pre>
    */
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
macchiati: Actually, an even better reason is that then you (and your clients) can avoid complications with null. The earliest start date is Integer.MIN_VALUE and the latest end date is Integer.MAX_VALUE.

There is one further complication. Normally missing start/end dates get the above values. However, for the Japanese calendar, the end date is set based on the **next** line's start data - 1.
So
```
              <era type="0" start="645-6-19"/>
              <era type="1" start="650-2-15"/>
...
              <era type="235" start="1989-1-8" code="heisei"/>
              <era type="236" start="2019-5-1" code="reiwa"/>
```
Is equivalent to 
```
              <era type="0" start="645-6-19" end="650-2-14" />
              <era type="1" start="650-2-15" end="671-12-31"/>
...
              <era type="235" start="1989-1-8" end="2019-4-30" code="heisei"/>
              <era type="236" start="2019-5-1" code="reiwa"/>
```
So that means you don't know the end data for an era until the next one gets added. And that's assuming that you read them in file order. Otherwise you need to read them in, and sort by start date.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        private GregorianCalendar getLatest() {
+            GregorianCalendar l = getEndCalendar();
+            if (l == null) l = getStartCalendar();
+            return l;
+        }
+
+        /** only works within the same cal system */
+        @Override
```
macchiati, line 103:
```suggestion
        @Override
        /* Normally comparisons require all of the fields to be listed, to ensure that no 
         * two unequal values a and b will never have a.compareTo(b) = 0.
         * However, the structure of the survey tool guaranties that no two eras (for different calendarTypes)
         * will overlap in start/end dates.
         */
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
```diff
@@ -0,0 +1,269 @@
...
+        public int compareTo(EraData o) {
+            return ComparisonChain.start()
+                    .compare(calendarType, o.calendarType)
+                    .compare(getLatest(), o.getLatest())
+                    .compare(getType(), o.getType())
```
macchiati, line 108:
```suggestion
                    .compare(getEnd(), o.getEnd())
```
Because of the above feature, you will never need the type for comparison; just the calendarType and end.

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: do we need more clarity in the spec here? 

before this, i couldn't find any code in CLDR that read these lines.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: or just link to the spec?

macchiati: The behavior of start/end is documented in https://www.unicode.org/reports/tr35/tr35-69/tr35-dates.html#Calendar_Data

(It could use some wordsmithing to clarify that "open-ended" applies when there isn't a series; probably start that first. I'll file a ticket for that.)

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: > So that means you don't know the end data for an era until the next one gets added. And that's assuming that you read them in file order. Otherwise you need to read them in, and sort by start date.

That's why i had the end time be null there.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
srl295: the survey tool doesn't guarantee this - do you mean the spec? the unit tests ought to validate this. But I'm not sure what all of the constraints are.

macchiati: Right, we don't make that clear. Filed https://unicode-org.atlassian.net/browse/CLDR-18576

macchiati: You're right. It is the spec not the Survey Tool.

tools/cldr-code/src/main/java/org/unicode/cldr/util/SupplementalCalendarData.java
macchiati: Ticket: https://unicode-org.atlassian.net/browse/CLDR-18576

srl295:
@macchiati this is getting pretty heavy weight. -and not working. Could I go back to the lightweight parser that succeeds for purposes of examples and split out the heavier api questions?

srl295:
> @macchiati this is getting pretty heavy weight. -and not working. Could I go back to the lightweight parser that succeeds for purposes of examples and split out the heavier api questions?

Specifically, back to Friday's version (but memoize it so the object is only created once).  And split out a PR for the more heavyweight processor that tries to sort out the situation for Japanese, etc.

I was trying to make a parser that just reflected what was in the XML, to enable unit testing, but it is kind of turning into a general purpose implementation of the algorithm.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -3151,30 +3100,87 @@ private void handleDateRangePattern(String value, List<String> examples) {
...
+        } else if (startCal != null && endCal == null) {
+            sampleDate = new GregorianCalendar(2002, 6, 15); // CLDR repo root commit
+            if (sampleDate.before(startCal)) {
+                sampleDate = startCal;
+                sampleDate.setTimeInMillis(
+                        sampleDate.getTimeInMillis() + (DateConstants.MILLIS_PER_DAY));
             }
+        } else {
+            // System.err.println("No good date for " + eraData);
+            // TODO: should be an error in TestSupplementalDataInfo
```
btangmu, line 3176: @srl295 does this TODO comment indicate more to be done on this ticket?

srl295: was supposed to be handled in #4648  - probably needs a clone.

srl295: Cloned tp CLDR-19018

PR #4646, opened 2025-04-28
CLDR-18370 Prevent alt=ascii values from being exported for datetime test data

echeran:
This PR fixes an issue that is occurring in datetime test data generation, in which `alt="ascii"` alternate data is being picked up, even though no extra configuration was set to request it. So this is undesirable yet happening by default in some cases.

[CLDR-18370](https://unicode-org.atlassian.net/browse/CLDR-18370)

ALLOW_MANY_COMMITS=true

[CLDR-18370]: https://unicode-org.atlassian.net/browse/CLDR-18370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
@@ -157,6 +157,9 @@ public class DateTimeFormats {
     /**
      * Set a CLDRFile and calendar. Must be done before calling addTable.
      *
+     * <p>Note: currently, this method will skip `alt="ascii"` elements of `dateFormatItem`. This
+     * may be configurable in the future. See CLDR-18580
+     *
      * @param file
      * @param calendarID
      * @return
@@ -167,6 +170,9 @@ public class DateTimeFormats {
 
     /**
      * Set a CLDRFile and calendar. Must be done before calling addTable.
+     *
+     * <p>Note: currently, this method will skip `alt="ascii"` elements of `dateFormatItem`. This
+     * may be configurable in the future. See CLDR-18580
      *
      * @param file
      * @param calendarID
@@ -270,6 +276,10 @@ public class DateTimeFormats {
                                         + calendarID
                                         + "\"]/dateTimeFormats/availableFormats/dateFormatItem"))) {
             XPathParts parts = XPathParts.getFrozenInstance(path);
+            if ("ascii".equals(parts.findAttributeValue("dateFormatItem", "alt"))) {
+                // TODO(CLDR-18580): Make this configurable.
+                continue;
+            }
             String key = parts.getAttributeValue(-1, "id");
             String value = file.getStringValue(path);
             if (key.equals(DEBUG_SKELETON)) {
--- a/common/testData/datetime/datetime.json
+++ b/common/testData/datetime/datetime.json
@@ -590,7 +590,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2000-01-01T00:00Z[Etc/GMT]",
-    "expected": "12:00:00 AM"
+    "expected": "12:00:00 AM"
   },
   {
     "semanticSkeleton": "T",
@@ -599,7 +599,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2024-07-01T08:50:07Z[Etc/GMT]",
-    "expected": "8:50:07 AM"
+    "expected": "8:50:07 AM"
   },
   {
     "semanticSkeleton": "T",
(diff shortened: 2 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -270,6 +270,9 @@ public DateTimeFormats set(CLDRFile file, String calendarID, boolean useStock) {
                                         + calendarID
                                         + "\"]/dateTimeFormats/availableFormats/dateFormatItem"))) {
             XPathParts parts = XPathParts.getFrozenInstance(path);
+            if ("ascii".equals(parts.findAttributeValue("dateFormatItem", "alt"))) {
```
macchiati, line 273: This doesn't allow us to build the ascii formats. I don't think it is important, but please add  javadoc note for this function, and file a ticket to (if needed!) extend the function to allow the choice of ascii or not.

sffc: DateTimeFormats is filling in data to a DateTimePatternGenerator. The DTPG allows only one pattern per skeleton. The current behavior on main is that the pattern being populated into the DTPG is arbitrarily determined via the XPath iteration order. This change makes it such that we reliably add the non-ascii version of the pattern to the DTPG, which is the behavior before the alt=ascii was added.

It seems good to file a follow-up ticket to make this behavior configurable.

Were you okay with the PR besides this?

macchiati:
Yes, I was.

BTW, We don't need an ordering. If we had the parameter, then if useAscii
is off, we would discard any alt='ascii'. If it were on, then in collecting
the data elements, we'd replace any non-alt=ascii element with the
corresponding alt=ascii version, and never replace the alt=ascii version
with the corresponding alt=ascii.

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
echeran: Added Javadoc and filed https://unicode-org.atlassian.net/browse/CLDR-18580 as requested. PTAL

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -168,6 +172,10 @@ public DateTimeFormats set(CLDRFile file, String calendarID) {
     /**
      * Set a CLDRFile and calendar. Must be done before calling addTable.
      *
+     * Note: currently, this method will initialize the datetime pattern in `dateFormatItem`
+     * by taking "default" values only. That is to say, any alternate values, patterns annotated
+     * with an attribute like `alt="ascii"` will be skipped over and not returned.
+     *
```
sffc, line 178:
```suggestion
     * Note: currently, this method will skip `alt="ascii"` elements of `dateFormatItem`.
     * This may be configurable in the future. See CLDR-18580
     *
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -270,6 +278,9 @@ public DateTimeFormats set(CLDRFile file, String calendarID, boolean useStock) {
                                         + calendarID
                                         + "\"]/dateTimeFormats/availableFormats/dateFormatItem"))) {
             XPathParts parts = XPathParts.getFrozenInstance(path);
+            if ("ascii".equals(parts.findAttributeValue("dateFormatItem", "alt"))) {
+                continue;
```
sffc, line 282:
```suggestion
                // TODO(CLDR-18580): Make this configurable.
                continue;
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -157,6 +157,10 @@ enum MyOptions {
     /**
      * Set a CLDRFile and calendar. Must be done before calling addTable.
      *
+     * Note: currently, this method will initialize the datetime pattern in `dateFormatItem`
+     * by taking "default" values only. That is to say, any alternate values, patterns annotated
+     * with an attribute like `alt="ascii"` will be skipped over and not returned.
+     *
```
sffc, line 163:
```suggestion
     * Note: currently, this method will skip `alt="ascii"` elements of `dateFormatItem`.
     * This may be configurable in the future. See CLDR-18580
     *
```

sffc (approved):

PR #4650, opened 2025-04-29
CLDR-14139 Filename collisions

macchiati:
CLDR-14139

ALLOW_MANY_COMMITS=true


macchiati:
This is still draft, and we could postpone to 49, but I want to see what y'all think.

btangmu:
LGTM but I don't see what this has to do with the ticket https://unicode-org.atlassian.net/browse/CLDR-14139

PR #4651, opened 2025-04-30
CLDR-17885 Sunuwar - add suz, suz_Deva, suz_Sunu

srl295:
CLDR-17855

includes font  from #4214 @NeilSureshPatel 

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true


DavidLRowe:
@srl295 In supplementalData.xml, I would have expected parentLocale="root" to have "suz_Sunu" rather than "suz_Deva". AFAICT `suz-Deva` = `suz`, but `suz-Sunu` needs to fall back to root, not to `suz`, thus `suz_Sunu` needs to be listed here (not `suz_Deva`).

DavidLRowe:
In the log, there's an error :

`Error:  (LikelySubtagsTest.java:574)  Error: : Deva: expected "[ँ-ःअ-ऍए-ऑओ-नपफभ-रलळव-ह़-ॅे-ॉो-्ॐ-॒ॠ-ॣॻॼॾॿ‌‍𑯚]", got "[ँ-ःअ-ऍए-ऑओऔखघङछझञठढणथधफभळशष़ऽिू-ॅैॉौ्ॐ-॒ॠ-ॣॻॼॾॿ‌‍]"
`

I don't know where that error is generated or what to do to fix it. AFAICT the second set of characters is a subset of the first. The first uses a number of ranges and includes more characters than the second.

It seems there is a Sunuwar character (U+11BDA Sunuwar letter ngar) just before the closing square bracket in the "expected" string, which seems odd.

srl295:
> In the log, there's an error :
> 
> `Error: (LikelySubtagsTest.java:574) Error: : Deva: expected "[ँ-ःअ-ऍए-ऑओ-नपफभ-रलळव-ह़-ॅे-ॉो-्ॐ-॒ॠ-ॣॻॼॾॿ‌‍𑯚]", got "[ँ-ःअ-ऍए-ऑओऔखघङछझञठढणथधफभळशष़ऽिू-ॅैॉौ्ॐ-॒ॠ-ॣॻॼॾॿ‌‍]" `
> 
> I don't know where that error is generated or what to do to fix it. AFAICT the second set of characters is a subset of the first. The first uses a number of ranges and includes more characters than the second.
> 
> It seems there is a Sunuwar character (U+11BDA Sunuwar letter ngar) just before the closing square bracket in the "expected" string, which seems odd.

yeah, there's an untranslated sunuwar char, i will check that. 

It's "expected" because it expected to find that etter in the static list but didn't.

NeilSureshPatel:
I just took look at the Sunuwar-InterIndic_SWS transform and the ngar is defined in the setup but it is missing in the transforms. I am guessing it is just passing through unchanged.

srl295:
> I just took look at the Sunuwar-InterIndic_SWS transform and the ngar is defined in the setup but it is missing in the transforms. I am guessing it is just passing through unchanged.

OK. I will fix it.

I had some issues with the transforms, that's why I made a new PR that has heavy hand editing.   Part might be in the transforms, btu there are also serious CLDR/ICU tooling issues.

srl295:
> I just took look at the Sunuwar-InterIndic_SWS transform and the ngar is defined in the setup but it is missing in the transforms. I am guessing it is just passing through unchanged.

the test file https://github.com/unicode-org/cldr/pull/4214/files#diff-1684eb9afa19fb9944dc3bf883a450979eff9787c04ac8ffd506c835b359322dR29 shows it unchanged also

DavidLRowe:
See my previous comment about `suz_Deva` in `parentLocale parent="root"` in supplementalData.xml. I think it should be `suz_Sunu` (or else I'd like to know why I'm wrong, which is entirely possible!).

DavidLRowe (approved):

PR #4654, opened 2025-04-30
CLDR-18544 Add transforms for translteration Adlam <-> Latin

sven-oly:
CLDR-18544

ALLOW_MANY_COMMITS=true


macchiati (changes requested):
Test files go into https://github.com/unicode-org/cldr/tree/main/common/testData/transforms and must follow the structure in those files. 
- must be named <transformId>.txt
- lines must be in the format: <source text><TAB><expected result text>
- should be extensive, eg 
    - https://github.com/unicode-org/cldr/blob/main/common/testData/transforms/und-Arab-t-und-beng.txt
    - https://github.com/unicode-org/cldr/blob/main/common/testData/transforms/cs-fonipa-t-cs.txt

sven-oly:
Thanks for the notes on this PR.

I've moved test files to the right directory and added test strings including region names, days, months, some geographic features, and a few other values.

sven-oly:
Looking at this again, what can I do to move this forward? I am not sure what to do about the maven failure and the Jira ticket failure.

Thanks.

PR #4665, opened 2025-05-03
CLDR-18370 Refactor dimension set definitions to be global constants

echeran:
CLDR-18370

ALLOW_MANY_COMMITS=true


sffc:
Three comments so far:

1. Try to move more code into methods, specifically the cloning of the test case.
2. Need to track which dimensions to multiply. Everything multiplies by locale, but only some things multiply by time zone or datetime.
3. Prevent duplicate test cases: the locales already in the kernel shouldn't be added again.
4. Since we're putting calendars and locales into the same dimension, make sure to consider this when forming your multiplication set.

AEApple:
@echeran - Is this still planned to be merged? Or will you clone the ticket and continue the work in CLDR 49?

PR #4669, opened 2025-05-05
CLDR-18475 dashboard fix: file/tab names were wrong

srl295:
Not sure where this regressed, but this fixes three issues:

1. XLSX download file name was always for example Dash_es_default.xlsx (default instead of es.moderate)
2. XLSX tab name was also es.default instead of es.moderate, ex
3. URL broken internally, https://st.unicode.org/cldr-apps/v#/undefined/T_NAmerica/77b7368701e0333e instead of /es/…  (undefined instead of locale)

CLDR-18475

this was under CLDR-18475 because I originally thought it was a JSON issue, but it's not.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDashData.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDashData.mjs
@@ -16,7 +16,7 @@ class DashData {
    *
    * @returns the new DashData object
    */
-  constructor() {
+  constructor(locale, coverageLevel) {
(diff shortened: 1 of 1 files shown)


srl295:
### Before: `Dash_es_default.xlsx`

image

### After: `Dash_es_basic.xlsx`

image

macchiati (approved):

PR #4670, opened 2025-05-05
CLDR-18574 Fix name YMD calendar section name

macchiati:
CLDR-18574

There is a related PR for the Info Hub change: https://github.com/unicode-org/cldr/pull/4672

Shouldn't say "8601" or "ISO"

ALLOW_MANY_COMMITS=true


macchiati:
Although this is a code change, it just modifies the user-visible names in the ST.

macchiati:
The documentation is updated, so it would be good to get this in soon.

pedberg-icu (approved):

PR #4671, opened 2025-05-05
CLDR-14409 run autotest for FE

srl295:
CLDR-14409

- restructures maven.yml a bit, new jobs for packaging and also building w/ docker.
  - also adds names to some of the steps
- move cldr-apps-webdriver into main cldr repo - not linked into the same maven tree just yet
  - the webdriver is somewhat limited in what it runs automatically -right now it checks that the dashboard works (but see https://unicode-org.atlassian.net/browse/CLDR-18614 ) 
  - updated the webdriver to be more resilient and to take configuration options
  - I didn't ~delete~ archive the other repo, but we might consider it if we move it into cldr
- add a new client-test subdirectory under js
- add docs

This adds a new step "Docker Test cldr-apps" which **is not required for passing** in fact it's not even required for smoketest deploy -- yet.

ALLOW_MANY_COMMITS=true


tools/cldr-apps-webdriver/scripts/selenium-grid-start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+VER=4.32.0
+URL=https://github.com/SeleniumHQ/selenium/releases/download/selenium-${VER}/selenium-server-${VER}.jar
+FILE=selenium-server.jar
```
srl295, line 4: I deleted the jar file that was previously checked into the repo.

This script isn't used within docker - we spin up a standalone chrome browser in a container.

tools/cldr-apps/js/test/client/test-client.mjs
srl295, line 1: First of a new class of tests - Node.js based, but runs against a running surveytool. Yes, you can run it against a local survey tool (see README just prior)

macchiati (dismissed):
Assuming this does not block PRs and has no effect on ST

macchiati (approved):

PR #4673, opened 2025-05-06
CLDR-16765 Add right regex pattern from UnicodeSet

macchiati:
CLDR-16765

ALLOW_MANY_COMMITS=true


macchiati:
@srl295 I think a problem with the last PR was that a UnicodeSet pattern is not the same as a regex pattern (in general). I tried testing it, and as expected it failed.

This adds a method (and test) to make a regex string from a UnicodeSet, that you can use in your cldrapp functions. Assuming ok, you use this in your code with a patch to this PR and merging. (I can't debug cldrapp, or I'd do it.)

srl295 (approved):
Yup exactly

srl295 (approved):
should this go into ICU's UnicodeSet?

macchiati:
Eventually, yes.

PR #4676, opened 2025-05-06
CLDR-16765 use-right-regex in front end

srl295:
- use new 'use right regex' function #4673
 - remove a debug printout
 - pass inherited text through the escaper, as well as translated

CLDR-16765

depends on #4673

image

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrEscaper.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrEscaper.mjs
@@ -10,10 +10,7 @@ const staticInfo = {
 
 /** updates content and recompiles regex */
 export function updateInfo(escapedCharInfo) {
-  const updatedRegex = escapedCharInfo.forceEscapeRegex
-    .replace(/\\ /g, " ")
-    .replace(/\\U[0]*([0-9a-fA-F]+)/g, `\\u{$1}`);
-  console.log(updatedRegex);
+  const updatedRegex = escapedCharInfo.forceEscapeRegex;
   const forceEscape = new RegExp(updatedRegex, "u");
   data = { escapedCharInfo, forceEscape };
 }
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -921,7 +921,7 @@ function addVitem(td, tr, theRow, item, newButton) {
       cldrText.get("voteInfo_baseline_desc")
     );
   }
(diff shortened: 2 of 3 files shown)


macchiati:
Looks good so far.

tools/cldr-apps/js/src/esm/cldrTable.mjs
```diff
@@ -921,7 +921,7 @@ function addVitem(td, tr, theRow, item, newButton) {
       cldrText.get("voteInfo_baseline_desc")
     );
   }
-  checkLRmarker(choiceField, item.value);
+  checkLRmarker(choiceField, displayValue);
```
srl295, line 924: this is an old bug, `displayValue` is updated above to be the inherited value instead of cldrSurvey.INHERITANCE_MARKER

tools/cldr-apps/js/src/esm/cldrEscaper.mjs
```diff
@@ -10,10 +10,7 @@ const staticInfo = {
 
 /** updates content and recompiles regex */
 export function updateInfo(escapedCharInfo) {
-  const updatedRegex = escapedCharInfo.forceEscapeRegex
-    .replace(/\\ /g, " ")
-    .replace(/\\U[0]*([0-9a-fA-F]+)/g, `\\u{$1}`);
-  console.log(updatedRegex);
+  const updatedRegex = escapedCharInfo.forceEscapeRegex;
```
srl295, line 13: nice and simple!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/CharInfo.java
```diff
@@ -53,7 +53,7 @@ public EscapedCharEntry(final CodePointEscaper c) {
 
     public static final class EscapedCharInfo {
         public final String forceEscapeRegex =
-                CodePointEscaper.ESCAPE_IN_SURVEYTOOL.toPattern(true);
+                CodePointEscaper.regexPattern(CodePointEscaper.ESCAPE_IN_SURVEYTOOL, "\\u{", "}");
```
srl295, line 56: @macchiati i'm assuming this is how you expected this to be used

macchiati: yes. As it turns out, there are only really two choices, \u style or \x style, so the API can be simplified, but TBD for now.

macchiati:
Going ahead and merging.

PR #4678, opened 2025-05-06
CLDR-18586 Update Managing Users

stenshamn:
CLDR-18586

First stab at adding section about vote forwarding, along with some general editing. But there are some sections that make no sense to me, including with at least one external dead link, so need some advice here.

ALLOW_MANY_COMMITS=true


srl295 (approved):

docs/site/index/survey-tool/managing-users.md
```diff
@@ -14,68 +14,83 @@ Open the survey tool, and click on the gear icon in the top left, then pick **Li
...
+2. Choose the user's predecessor (i.e. the user you want to copy votes *from*).
+3. Choose what locale you want to copy votes *from*. (In case the old user has been active in multiple locales.)
+4. Choose what locale you want to copy the old user's votes *to*. (In almost all cases, that will be the same locale, which is the default choice.)
+5. Vote copying will take place.
+6. Tell the new user to exit and re-enter the survey tool.
+
+At this time, the survey tool will consider the old user's voting history the same as the new user's. When the new user logs in, survey tool will automatically apply all the old user's votes **when they match the currently shipping value**. In any case where the old vote does not match the currently shipping value OR where the value is new, the vote will be abstained and the new user will have to submit a new vote.
+
+Remind the new user that **they are still responsible for inherited votes**, and should review them carefully to ensure they agree with them. The vote inheritance is provided for efficiency and convenience, but does not shift responsibility.
```
AEApple, line 51: Should we also mention that votes can be imported from another organization?

docs/site/index/survey-tool/managing-users.md
```diff
@@ -14,68 +14,83 @@ Open the survey tool, and click on the gear icon in the top left, then pick **Li
...
 
 ## To resend a password
```
AEApple, line 53: @srl295 - How does the new SSO functionality fit with this?

srl295: we don't have SSO for actually logging in yet… but we could, do you want it?   Could do google, etc SSO also

docs/site/index/survey-tool/managing-users.md
stenshamn: Can they? I don't know that they can from within the confines of the survey tool though? Either way, it seems like an unusual case.

stenshamn:
To be clear: this is **not ready to merge**. We have three sections at the end that need work, but I don't have the info needed to complete them.

srl295:
> To be clear: this is **not ready to merge**. We have three sections at the end that need work, but I don't have the info needed to complete them.

Make it draft then

docs/site/index/survey-tool/managing-users.md
macchiati: @srl295 In discussion, I thought you said that a Manager of organization X could import votes to user A from a user B that is in organization Y, in the Survey tool. Is that true?

docs/site/index/survey-tool/managing-users.md
```diff
@@ -14,68 +14,83 @@ Open the survey tool, and click on the gear icon in the top left, then pick **Li
...
+*Known bug: if you try to create a user and that email address already has an account, you get an unknown error message. Often this is because the user is locked. Click on "Show locked users", find the person, and change the status (see below).*
+
+## To forward old votes to a new user
+When replacing a previous user with a new one, you will likely want that new user to inherit the votes from their predecessor. Otherwise the new user would have to submit all votes from scratch manually, which can be several thousands.
+
+1. Once the new user has been created, find them in the list of users and click on the link "Copy votes to this vetter"
+2. Choose the user's predecessor (i.e. the user you want to copy votes *from*).
+3. Choose what locale you want to copy votes *from*. (In case the old user has been active in multiple locales.)
+4. Choose what locale you want to copy the old user's votes *to*. (In almost all cases, that will be the same locale, which is the default choice.)
```
macchiati, line 45: This seems very strange; copying votes from say, German to Spanish.

docs/site/index/survey-tool/managing-users.md
```diff
@@ -14,68 +14,83 @@ Open the survey tool, and click on the gear icon in the top left, then pick **Li
...
+5. Vote copying will take place.
+6. Tell the new user to exit and re-enter the survey tool.
+
+At this time, the survey tool will consider the old user's voting history the same as the new user's. When the new user logs in, survey tool will automatically apply all the old user's votes **when they match the currently shipping value**. In any case where the old vote does not match the currently shipping value OR where the value is new, the vote will be abstained and the new user will have to submit a new vote.
```
macchiati, line 49: This makes it sounds like I can only import votes into a new user (before they have logged in to the current cycle).

If the vetter has logged in area, can a Manager still import votes and have them stick. (All those that don't conflict with a blue star, right??)

docs/site/index/survey-tool/managing-users.md
```diff
@@ -87,23 +102,23 @@ Later on, in the vetting phase, you can send messages with the outstanding dispu
 
 1. Click on the magnifying glass in the left column. You zoom in on that user, and have more Actions in the action column.
 2. Click the one you want, and hit \[Do Action]
-3. Normally you don't want to delete users \-\- you want to Lock them. (Deleting removes information about what they've submitted so far.)
+3. Normally you don't want to delete users \-\- you want to lock them. (Deleting removes information about what they've submitted so far.)
```
macchiati, line 105: This should be stronger. (And we should disallow deletion — there is a ticket for that and we should prioritize)
```suggestion
3. **Never** delete users — you lock them instead. (Deleting removes information about what they've submitted so far.)
```

docs/site/index/survey-tool/managing-users.md
```diff
@@ -112,17 +127,16 @@ Later on, in the vetting phase, you can send messages with the outstanding dispu
...
-	4. Go to Manage Users (top right), and look at the users signed up to these locales
-	5. When they are bogus fix them (work with translator to replace by real locale ids)
-	6. When they are valid, you may propose to the committee that we add a locale:
-		1. Valid ones are under languages in en.xml
-		2. OR (unusual case) under languages in http://www.ietf.org/internet\-drafts/draft\-ietf\-ltru\-4645bis\-04\.txt
-		3. However, watch for mistakes, eg someone using "tw" to mean Taiwanese, when it actually means "Twi"
+## NEEDS WORK: To check if you have "bad" locales (locales assigned that aren't in CLDR)
+
+1. Go to Coverage (up at the top right in the survey tool).
+2. If you see the following message, then some of your users have bogus locales OR locales that aren't in CLDR: **Warning: locale designations not matching locales in CLDR: X....**
```
macchiati, line 133: This also needs a ticket to disallow bogus/non-CLDR locales.

docs/site/index/survey-tool/managing-users.md
```diff
@@ -112,17 +127,16 @@ Later on, in the vetting phase, you can send messages with the outstanding dispu
...
+4. Go to Manage Users (top right), and look at the users signed up to these locales.
+5. When they are bogus fix them (work with translator to replace by real locale IDs).
+6. When they are valid, you may propose to the committee that we add a locale:
+	1. Valid ones are under languages in en.xml
+	2. OR (unusual case) under languages in http://www.ietf.org/internet\-drafts/draft\-ietf\-ltru\-4645bis\-04\.txt
+	3. However, watch for mistakes, eg someone using "tw" to mean Taiwanese, when it actually means "Twi"
```
macchiati, line 139:
```suggestion
5. When they are bogus or invalid, fix them (work with translator to replace by real locale IDs).
```

macchiati (changes requested):

docs/site/index/survey-tool/managing-users.md
srl295: They are disallowed from voting in bogus/non-CLDR locales, and disallowed from creating users with such.  Do you mean to remove this warning from the UI? That is, suppose `iw` was renamed `he` (hey, it could happen), what happens to users assigned to `iw`? This warning says, you need to go in and choose a different locale for them. We could have no warning, and then there would be no indication that something was inconsistent.

So, please clarify 'disallow'.

docs/site/index/survey-tool/managing-users.md
srl295: that sounds more like _moving_ user from organization Y to organization X.  Probably an admin can copy votes between organization.  But I think you have to have permissions to manage both A and B to do the copy.

docs/site/index/survey-tool/managing-users.md
srl295: We've already disallowed deletion...

srl295:
```suggestion
3. Normally you don't delete users, you lock them. (Deleting would remove information about what they've submitted so far.)
```

docs/site/index/survey-tool/managing-users.md
macchiati: If we reversed an alias, then we would have to make substantial changes to both data on disk **and** fix the ST database.

So Managers should **never** see that message, and we should drop any discussion of it instead of taking up their time trying to understand what to do in a case that won't happen.

docs/site/index/survey-tool/managing-users.md
macchiati:
```suggestion
4. Choose what locale you want to copy the old user's votes *to*.
The default is the same locale.
**Never** copy to a locale with a different language code, eg from German (de) to French (fr).
```

docs/site/index/survey-tool/managing-users.md
macchiati: Great. So here is the change then.
```suggestion
3. If a vetter is no longer active, then lock them.
(We don't allow deletion because it removes information about what they've submitted so far.)
```

docs/site/index/survey-tool/managing-users.md
srl295: they can import votes to a non-new user, but they won't "stick" for xpaths that:
- they have already done an explicit vote on
- they already have an existing import that they've performed
a message comes up saying 'skipping this xpath' or some such

docs/site/index/survey-tool/managing-users.md
srl295: @macchiati actually IDs do change, in 0a79e9f8c3bfc107a1c326ab0cbae9039dfe4807 ajp.xml was renamed to apc.xml (and there have been more major ones, such as iw to he in the distant past). Also, I think we used this vote function specifically not too long ago to move votes between nb/nn/no.

docs/site/index/survey-tool/managing-users.md
srl295: fully agree for this discussion.

docs/site/index/survey-tool/managing-users.md
macchiati: Clearly **the TC** needs to be able to do this — along with more substantial changes that need to be coordinated with that.

But **Managers** are more likely to make mistakes than to do the right thing.

stenshamn:
> > To be clear: this is **not ready to merge**. We have three sections at the end that need work, but I don't have the info needed to complete them.
> 
> Make it draft then

TIL

docs/site/index/survey-tool/managing-users.md
stenshamn: I ended up removing this entire section. I wasn't able to find anything in the Survey Tool that resembles this feature. If you can point it out to me, I can try and document it.

docs/site/index/survey-tool/managing-users.md
stenshamn: I had a use case for this back when a vetter accidentally used `ur_IN` to submit all their initial voting instead of just `ur`. I would think someone with the keys to management and TC would know to treat this carefully.

docs/site/index/survey-tool/managing-users.md
stenshamn: In my later draft I removed this entire section because I couldn't find this interface in the survey tool. There is no "Coverage" link that I could find.

docs/site/index/survey-tool/managing-users.md
AEApple: Should we limit this ability to the members of the TC and admin then?

docs/site/index/survey-tool/managing-users.md
```diff
@@ -86,43 +100,25 @@ Later on, in the vetting phase, you can send messages with the outstanding dispu
 ## To change users' status
 
 1. Click on the magnifying glass in the left column. You zoom in on that user, and have more Actions in the action column.
-2. Click the one you want, and hit \[Do Action]
-3. Normally you don't want to delete users \-\- you want to Lock them. (Deleting removes information about what they've submitted so far.)
+2. Click the one you want, and click \[Do Action]
+3. If a user is inactive, lock their account. (We don't delete accounts since it removes information about what they've submitted so far.)
```
AEApple, line 104:
```suggestion
3. If a user is inactive, lock their account. (We don't delete accounts since it removes information about what they've submitted so far, and can be helpful for auditing changes in response to bug reports.)
```

docs/site/index/survey-tool/managing-users.md
stenshamn: Since Apple doesn't use Manager, we just have TC and vetters, I'll leave that for others to say.

docs/site/index/survey-tool/managing-users.md
stenshamn: Not a bad suggestion, but we don't really need to motivate our choices publicly, better to just keep it short IMHO.

docs/site/index/survey-tool/managing-users.md
stenshamn: We don't intend to cover every case under the moon either. The most common scenario for this feature is when we introduce a new vetter, and we don't say in this text that this is the only scenario this feature could be used, so I think we can stick with the current wording for now.

docs/site/index/survey-tool/managing-users.md
stenshamn: Adopted similar wording in a later commit.

AEApple (approved):
LGTM! We can always continue to update in a different PR

macchiati (approved):

PR #4679, opened 2025-05-06
CLDR-18603 Add GyM, GyMEd availableFormats to root, en

pedberg-icu:
CLDR-18603

Add `GyM`, `GyMEd` availableFormats for gregorian, generic, and some other calendars to `root`, `en` (there are already intervalFormats for these), based on the existing availableFormats in the same calendars for `GyMd`, `yM`, and `yMEd`. As far as I can tell these new availableFormats will already be included in appropriate coverage levels.

Leaving the ticket open in case we find other such omissions to add.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -912,7 +912,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					</dateFormatLength>
 					<dateFormatLength type="short">
 						<dateFormat>
-							<pattern>GGGGG y-MM-dd</pattern>
+							<pattern>G y-MM-dd</pattern>
 							<datetimeSkeleton>GGGGGyMMdd</datetimeSkeleton>
 						</dateFormat>
 					</dateFormatLength>
@@ -979,7 +979,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="Ehms">E h:mm:ss a</dateFormatItem>
 						<dateFormatItem id="EHms">E HH:mm:ss</dateFormatItem>
 						<dateFormatItem id="Gy">G y</dateFormatItem>
-						<dateFormatItem id="GyMd">GGGGG y-MM-dd</dateFormatItem>
+						<dateFormatItem id="GyM">G y-MM</dateFormatItem>
+						<dateFormatItem id="GyMd">G y-MM-dd</dateFormatItem>
+						<dateFormatItem id="GyMEd">G y-MM-dd, E</dateFormatItem>
 						<dateFormatItem id="GyMMM">G y MMM</dateFormatItem>
 						<dateFormatItem id="GyMMMd">G y MMM d</dateFormatItem>
 						<dateFormatItem id="GyMMMEd">G y MMM d, E</dateFormatItem>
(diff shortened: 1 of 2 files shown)


pedberg-icu:
Any TC member, please merge when approved

common/main/root.xml
```diff
@@ -979,7 +979,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="Ehms">E h:mm:ss a</dateFormatItem>
 						<dateFormatItem id="EHms">E HH:mm:ss</dateFormatItem>
 						<dateFormatItem id="Gy">G y</dateFormatItem>
+						<dateFormatItem id="GyM">GGGGG y-MM</dateFormatItem>
```
macchiati, line 982: The root and english locales should use G consistently instead of sometimes G and sometimes GGGGG (that is, default to the "medium" format instead of a mix of medium and narrow formats). This would then be overridden in the various languages.

macchiati (changes requested):
OK except for one thing.

common/main/root.xml
pedberg-icu: Actually, the procedure that we usually try to follow is to have GGGGG if there is a numeric month and G otherwise. I will convert accordingly (in the first version of this I just followed whatever convention was already used for GyMd, but that did not necessarily accord with the above)

sffc (commented):
Thanks!

common/main/root.xml
pedberg-icu: Well, let me amend that, looking at the actual eras. `root` only has abbreviated eras so I think only G should be used formats. For `en` we did want to use GGGGG for numeric formats (with numeric month) in the japanese calendar so we just have the single-letter era name. For the other calendars in `en` I agree we should use G.

pedberg-icu: @macchiati OK, looking at the eras: in `root` we only want G (we only have abbreviated eras). In `en` we want G in most cases, certainly gregorian; however in the japanese calendar, for formats with numeric month we definitely want GGGGG to get the one-letter abbreviated era names. Made those changes, please take a look (and if you approve, please also merge if tests have passed).

AEApple:
Should I update the en_JP and en_001 in that case or will we pick up the fixes when we raise the en_GB changes? I had kept consistent with the generic calendar so if generic was wrong, then the newer calendars will have the same issue.

pedberg-icu:
> Should I update the en_JP and en_001 in that case or will we pick up the fixes when we raise the en_GB changes? I had kept consistent with the generic calendar so if generic was wrong, then the newer calendars will have the same issue.

@AEApple for `en_JP` I think we should change all GGGGG to G except for these 3 in the japanese calendar:
```
						<dateFormatItem id="yyyyM">GGGGG y/MM</dateFormatItem>
						<dateFormatItem id="yyyyMd">GGGGG y/MM/dd</dateFormatItem>
						<dateFormatItem id="yyyyMEd">GGGGG y/MM/dd EEEE</dateFormatItem>
```

for `en_001` I think we should go ahead now and change all GGGGG to G for the benefit of other `en`s that inherit from it

pedberg-icu:
(this is blocked until Mark approves and removes his block)

macchiati (approved):

pedberg-icu:
Other omissions are covered by other tickets

PR #4682, opened 2025-05-07
CLDR-16765 escape English column, link to docs

srl295:
CLDR-_____

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/translation/getting-started/guide.md
+++ b/docs/site/translation/getting-started/guide.md
@@ -84,6 +84,20 @@ The main panel uses icons to indicate important information and possible problem
 
 (Icons for the **Dashboard** are listed in the [Dashboard Icons](#dashboard-icons) section below.)
 
+### Special Characters
+
+Some items may use special characters, especially invisible characters. 
+
+These will be shown underneath the regular value in an additional box. For example:
+
+> h:mm:ss a z
+>
+> h:mm:ss❰NBTSP❱a z
+
+This indicates that there is a special character that was not visible in the upper line, in this case a no-break thin space, U+202F.
+
+See [Key to Escapes](/translation/core-data/exemplars#key-to-escapes) for a list with some of the escape characters.
+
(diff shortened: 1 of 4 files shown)


docs/site/translation/getting-started/guide.md
```diff
@@ -84,6 +84,20 @@ The main panel uses icons to indicate important information and possible problem
 
 (Icons for the **Dashboard** are listed in the [Dashboard Icons](#dashboard-icons) section below.)
 
+### Special Characters
+
+Some items may use special characters, especially invisible characters.
```
macchiati, line 89: We should align this and the info hub changes. Can be done after merging.

tools/cldr-apps/src/main/webapp/css/redesign.css
```diff
@@ -243,17 +243,39 @@ h3.collapse-review > span:first-child {position: relative; top: -220px;display:
 	border: 1px solid #0FF;
 }
 .data .visible-mark::before{
-	content: '<';
+	content: '❰';
```
macchiati, line 246: The downside of having this not in the text is that people can't search on page for it.

macchiati: same with other one.

macchiati (approved):
Needs some changes, but approving in case you want to get this updated today.

tools/cldr-apps/src/main/webapp/css/redesign.css
srl295: it's to make the display more flexible.

macchiati: Right. But people can't search for it. And the slight advantage of our being able to update with a CSS change instead of a code change isn't worth sacrificing that.

However, in the code they can be constants, so it is easy to change in one place.

PR #4683, opened 2025-05-08
CLDR-18572 Add en_EE, en_GE and en_UA

AEApple:
CLDR-18572

- Add en_EE, en_GE and en_UA
- Remove unnecessary time formats from en_LT and en_LT

ALLOW_MANY_COMMITS=true


AEApple:
Starting with adding the data items with the most value so that we can start analyzing what all the en_150 child locales have in common.

PR #4684, opened 2025-05-08
CLDR-18616 Revise the Key to Escapes

macchiati:
CLDR-18616

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/translation/core-data/exemplars.md
+++ b/docs/site/translation/core-data/exemplars.md
@@ -25,38 +25,31 @@ Certain fields have _**sets**_ of characters (and strings) as values, called **U
 
 ### Key to Escapes
 
-| Abbr. | Code Point | Name |
-|---|---|---|
-| ❰TAB❱ | U+0009 | tab |
-| ❰LF❱ | U+000A | line feed |
-| ❰CR❱ | U+000D | carriage return |
-| ❰SP❱ | U+0020 | space |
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/CodePointEscaper.java
```diff
@@ -10,80 +10,99 @@
...
-    LRM(
-            0x200E,
-            "left-right mark",
-            "For BIDI, invisible character that behaves like Hebrew letter."),
-    RLM(0x200F, "right-left mark", "For BIDI, invisible character that behaves like Latin letter."),
 
-    LRO(0x202D, "left-right override"),
-    RLO(0x202E, "right-left override"),
-    PDF(0x202C, "end override"),
+    // Oddball Cf characters that are in CLDR data
```
srl295, line 71:
```suggestion
    // Other Cf characters that are in CLDR data
```

Are they here because they are Cf?  The only practical reason I could think of to include them is confusabillity (that ؂ might look like another character, etc)

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/CodePointEscaper.java
macchiati: Yes

PR #4694, opened 2025-05-08
CLDR-16765 unescape improvements

srl295:
1. escape multiple chars, not just first
2. move prefix/suffix out of CSS and into plain text
3. don't escape items on the FE that DAIP already escaped on the BE

CLDR-16765

ALLOW_MANY_COMMITS=true

#### 1 - multiple escape

image

#### 2 - plain text brackets

showing selected as plain text

image

#### 3 - no double escape

I'm a little surprised that we didn't have double escaping before for Arabic exemplars. In any event, this is suppressed now - however, I actually think the new escaping display is superior to the server-generated HTML.  Ideally we should have a _widget_ for working with exemplars on the server side. Anyway.

image

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrEscaper.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrEscaper.mjs
@@ -11,7 +11,7 @@ const staticInfo = {
 /** updates content and recompiles regex */
 export function updateInfo(escapedCharInfo) {
   const updatedRegex = escapedCharInfo.forceEscapeRegex;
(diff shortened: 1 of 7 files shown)


srl295:
> > however, I actually think the new escaping display is superior to the server-generated HTML. Ideally we should have a widget for working with exemplars on the server side. Anyway.
> 
> I fully agree. The basic hold-up is inputing the escaped characters — we need to explore good ways to do that in the 'input' ticket.

this ticket's API gives the front end the data it needs to work with it.. it could even present all of the escaped chars as a well of chips to draw from!

PR #4702, opened 2025-05-14
CLDR-18638 Add examples for available/intervalFormats

macchiati:
CLDR-18638 

WIP

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -1784,6 +1784,17 @@ public class ExampleGenerator {
             return;
         }
         examples.add(intervalFormat.format(FIRST_INTERVAL, later));
+        Set<String> relatedPatterns = RelatedPathValues.getRelatedPathValues(cldrFile, parts);
+        if (!relatedPatterns.isEmpty()) {
+            examples.add("Comparisons:");
+            for (String pattern : relatedPatterns) {
+                SimpleDateFormat sdf =
+                        icuServiceBuilder.getDateFormat(
+                                parts.getAttributeValue(RelatedPathValues.calendarElement, "id"),
+                                pattern);
+                examples.add(sdf.format(DATE_SAMPLE));
+            }
+        }
     }
 
     private void handleDelimiters(
@@ -2037,7 +2048,7 @@ public class ExampleGenerator {
         examples.add(format(value, setBackground(sdf.format(DATE_SAMPLE)), setBackground(zone)));
     }
 
-    private class IntervalFormat {
+    public class IntervalFormat {
         @SuppressWarnings("deprecation")
         DateTimePatternGenerator.FormatParser formatParser =
                 new DateTimePatternGenerator.FormatParser();
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/RelatedPathValues.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/RelatedPathValues.java
@@ -1,0 +1,79 @@
+package org.unicode.cldr.test;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.XPathParts;
+
+public class RelatedPathValues {
+    public static final int calendarElement = 3;
+    public static final int dateTypeElement = 5;
+    public static final int idElement = 6;
+
+    static final XPathParts interval =
+            XPathParts.getFrozenInstance(
+                    "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\"MMMd\"]");
+
+    // samples
+    // ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/intervalFormats/intervalFormatItem[@id="Bhm"]/greatestDifference[@id="B"]
+    // ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/availableFormats/dateFormatItem[@id="MMMd"]
+    // ldml/dates/calendars/calendar[@type="gregorian"]/dateFormats/dateFormatLength[@type="full"]/dateFormat[@type="standard"]/pattern[@type="standard"]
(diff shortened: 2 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/RelatedDatePathValues.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+        XPathParts newPath = xparts.cloneAsThawed();
+        Set<String> result = new LinkedHashSet<>();
+        for (String item : skeletons) {
+            addAvailable(cldrFile, newPath, item, result);
+        }
+        return result;
+    }
+
+    private static String addRelated(String skeleton, String letter, Set<String> skeletons) {
```
btangmu, line 95: addRelated should be void not String; currently it always returns null and the return value is unused

tools/cldr-code/src/main/java/org/unicode/cldr/test/RelatedDatePathValues.java
```diff
@@ -0,0 +1,226 @@
...
+    //            cores = new TreeSet<>(SKELETON_COMPARE);
+    //            String newItem = addCore(skeleton, "G", cores);
+    //            addCore(skeleton, "E", cores);
+    //            addCore(skeleton, "v", cores);
+    //        }
+    //        skeletonToCores.put(skeleton, ImmutableSet.copyOf(cores));
+    //        return cores;
+    //    }
+
+    private static String addCore(String skeleton, String letter, Collection<String> cores) {
```
btangmu, line 188: This should be void as well

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1784,6 +1784,18 @@ private void handleIntervalFormats(XPathParts parts, String value, List<String>
             return;
         }
         examples.add(intervalFormat.format(FIRST_INTERVAL, later));
+        Set<String> relatedPatterns = RelatedDatePathValues.getRelatedPathValues(cldrFile, parts);
+        if (!relatedPatterns.isEmpty()) {
```
btangmu, line 1788: I suggest checking relatedPatterns != null, or else adding @NotNull annotation to getRelatedPathValues. It looks like it currently can't return null, but better to make sure

tools/cldr-code/src/main/java/org/unicode/cldr/test/RelatedDatePathValues.java
```diff
@@ -0,0 +1,226 @@
...
+import java.util.TreeSet;
+import java.util.concurrent.ExecutionException;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.CldrIntervalFormat;
+import org.unicode.cldr.util.XPathParts;
+
+public class RelatedDatePathValues {
+    public static final int calendarElement = 3;
+    public static final int dateTypeElement = 5;
+    public static final int idElement = 6;
```
btangmu, line 22: A comment like this would help:

In a path like `//ldml/dates/calendars/calendar[@type=\"gregorian\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\"MMMd\"]`,

element 0 = ldml
element 3 = calendarElement is like `calendar[@type=\"gregorian\"]`
element 5 = dateTypeElement is like `availableFormats`
element 6 = idElement is like `dateFormatItem[@id=\"MMMd\"]`

btangmu (approved):
It would take me many hours to understand all this. For now, LGTM, with a couple minor suggestions

tools/cldr-code/src/main/java/org/unicode/cldr/test/RelatedDatePathValues.java
srl295: could actually do it this way:

```js
XPathParts x = XPathParts.cloneAsThawed("//ldml/dates/calendars/calendar[@type=\"gregorian\"]/dateTimeFormats/availableFormats/dateFormatItem[@id=\"MMMd\"]);
public static final int calendarElement = x.findElement(LDMLConstants.CALENDAR);
public static final int dateTypeElement = x.findElement(LDMLConstants.DATE_TIME_FORMATS);
…
```

macchiati: I looked at this. For now I'm going to add comments. BTW LDMLConstants is itself kinda clunky, and requires manual update. Better would be to build constants (even better, an enum) from DtdData (but not now).

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -57,7 +57,10 @@
 import org.unicode.cldr.util.XPathParts;
 
 public class CheckDates extends FactoryCheckCLDR {
+    private static final boolean DEBUG = false;
+
```
srl295, line 61:
```suggestion
```

these shouldn't be needed, just use a debugger?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -179,6 +182,9 @@ public CheckCLDR handleSetCldrFileToCheck(
             try {
                 flexInfo.checkFlexibles(path, value, fullPath);
             } catch (Exception e) {
+                if (DEBUG) {
+                    e.printStackTrace();
+                }
```
srl295, line 187:
```suggestion
```

or use a logger…

tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
```diff
@@ -468,7 +468,7 @@ public static void main(String[] args) throws Throwable {
 
         String user = options[USER].value;
 
-        System.out.println("Source directories:\n");
+        System.out.println("# Source directories:\n");
```
srl295, line 471: unrelated change (not bad but unrelated

srl295 (approved):
Code all seems fine (so I'm going to approve it since it affectes vetters) but it seems like it should be about four different PRs, plus includes a lot of unneeded stuff like dead debugging code.

macchiati:
I'm going to go ahead and merge this, but follow up with a couple of fixes in a separate PR

PR #4706, opened 2025-05-15
CLDR-18642 Sicilian CLDR-18643 Sunuwar CLDR-18644 Shan

srl295:
CLDR-18642

- CLDR-18642 Sicilian
- CLDR-18643 Sunuwar
- CLDR-18644 Shan

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true


srl295:
@dpaleino

dpaleino:
> @dpaleino

Thanks!
I don't fully understand the code, but out of curiosity, may I ask why sicilian_cademia and not just sicilian? Btw the correct translation would be "Sicilian Academy", not "Sicilian Cademia".

srl295:
> > @dpaleino
> 
> Thanks! I don't fully understand the code, but out of curiosity, may I ask why sicilian_cademia and not just sicilian, or just cademia? Btw the correct translation would be "Sicilian Academy", not "Sicilian Cademia".
> 
> If it's the short name I chose in the form, and was changed to make it unambiguous, would it be possible to make it "cademiasiciliana" instead? Because I did choose "cademia" thinking there was a limitation on characters.
> 
> Thank you!

Hi,
 I have been starting them with the language/region so that it's clearer for the technical committee what it pertains to. Inside the tool and for users it will show up as "Cademia Siciliana". `sicilian_cademia` is just the internal database key.

srl295:
@dpaleino i updated the alternate name , see if that is better

conradarcturus (approved):

PR #4707, opened 2025-05-15
Reject invalid locale ID before setting/using

btangmu:
CLDR-18573

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLocales.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrLocales.mjs
@@ -4,8 +4,20 @@
 import * as cldrEvent from "./cldrEvent.mjs";
 import * as cldrLoad from "./cldrLoad.mjs";
 import * as cldrMenu from "./cldrMenu.mjs";
+import * as cldrNotify from "./cldrNotify.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
+
+// The current locale may be "USER" temporarily, meaning the back end should choose an appropriate locale
+// for the current user. The real locale ID should be set when a server response contains it.
+const USER_LOCALE_ID = "USER";
+
+const VALIDATE_ALL_LOCALES = true;
+const VALIDATE_ALL_LOCALES_VERBOSE = true;
+
+const LOCALE_REGEX = new RegExp("^[a-zA-Z0-9_]*$");
+
+let validatedAllLocales = false;
 
 // called as special.load
 function load() {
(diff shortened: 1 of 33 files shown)


btangmu:
One concern I have about this PR is that maybe `cldrStatus.setCurrentLocale` could be called with a valid non-null non-empty value, before `cldrLoad.localeMapReady` becomes true. I haven't seen that happen in local testing, but I'd like to confirm that it's impossible. Otherwise users might get the notification "The list of locales is unavailable", which would be annoying though probably not a big problem.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
```diff
@@ -4,8 +4,25 @@
...
+
+const VALIDATE_ALL_LOCALES = true;
+const VALIDATE_ALL_LOCALES_VERBOSE = false;
+
+const LOCALE_REGEX = new RegExp("^[a-zA-Z0-9_]*$");
+
+let didValidateAll = false;
+
+// For unit testing, inhibit calling cldrNotify for locale-validation failures,
+// to avoid "global leak(s) detected: '__VUE__', '__VUE_DEVTOOLS_HOOK_REPLAY__'"
```
srl295, line 24: instead, change mocha.setup in Test.html to include any symbols that are global. 

probably change `mocha.setup("bdd")` to `mocha.setup({ui:"bdd",global:['__VUE__','__VUE_DEVTOOLS_HOOK_REPLAY__']);`  see https://mochajs.org/#browser-configuration

i have hit this before, it's either on one of the PRs awaiting ticket acceptance or I ended up not seeing the leak.

srl295 (commented):
i know this is draft but just commenting

instead of handling the mode where "locales aren't available" we should instead load the locale list, and anything else we need, in  `cldrGui.initialSetup()` which loads **in parallel** any one-time loads.  The locmap and menus are (i think) cached after first load so should be quick.  There's even a to-do in that function.

all of this is done at the same time as the session look-up so should be very quick on the user's side. then we can avoid this state where the page is up but some things are broken.

btangmu:
> instead of handling the mode where "locales aren't available" we should instead load the locale list

Yes, that's what I was alluding to in my first comment here. I'm not sure if the failure in question is even possible, and if it possible, we should fix it so it's impossible -- the locale list should be loaded before setCurrentLocale (for non-null, non-empty value)

tools/cldr-apps/js/test/TestCldrForumFilter.mjs
```diff
@@ -1,6 +1,8 @@
 import * as cldrForum from "../src/esm/cldrForum.mjs";
 import * as cldrForumFilter from "../src/esm/cldrForumFilter.mjs";
 
+import * as chai from "chai";
```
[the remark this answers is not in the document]
btangmu: Without that import, VS Code shows a warning message that chai isn't defined

tools/cldr-apps/js/test/TestCldrLocales.mjs
```diff
@@ -0,0 +1,46 @@
...
+export const TestCldrLocales = "ok";
+
+const assert = chai.assert;
+
+describe("cldrLocales.ping", function () {
+  const result = cldrLocales.ping();
+  it("should return pong", function () {
+    assert(result === "pong", "result equals pong");
+  });
+});
```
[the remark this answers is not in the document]
btangmu: Removed.

btangmu:
> confirm that it's impossible

I've made more changes to this PR, in particular `cldrGui.initialSetup` now calls `cldrLocales.fetchMap` so I believe it's impossible for the locale to be set before the locale map is ready.

btangmu:
@srl295 now the locale map is loaded at the beginning, separately from the menus. There's still a lot of room for improvement, but maybe this is enough changes for one PR. BTW I had problems with bad commit messages and couldn't figure out how to fix them other than squashing.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
```diff
@@ -55,4 +100,59 @@ function parseHash(pieces) {
...
+  }
+  notifyUnusableLocale(map, loc);
+  return false;
+}
+
+function notifyUnusableLocale(map, loc) {
+  const problem = cldrText.get("locale_id_unusable");
+  let explanation;
+  if (!map) {
+    explanation = cldrText.get("locale_id_list_unavailable");
```
btangmu, line 123: This should be impossible, but if it happens we'll report it

tools/cldr-apps/js/src/esm/cldrLocales.mjs
```diff
@@ -1,11 +1,56 @@
...
+    .doFetch(url)
+    .then(cldrAjax.handleFetchErrors)
+    .then((r) => r.json())
+    .then(setMap)
+    .catch(console.error);
+}
+
+function getUrl() {
+  const p = new URLSearchParams();
+  p.append("what", "menus"); // cf. WHAT_GET_MENUS in SurveyAjax.java
```
btangmu, line 38: WHAT_GET_MENUS is now used completely separately for getting locales and for getting menus. It could be separated (and modernized to api) on the back end. Or, possibly it should be recombined for fewer http requests, however that would require refactoring of the menu code since that's very crufty with `cldrLoad.myLoad` and `flipper`

tools/cldr-apps/js/src/esm/cldrStatus.mjs
```diff
@@ -232,11 +233,14 @@ function getCurrentLocale() {
 }
 
 function setCurrentLocale(loc) {
+  if (loc && loc !== cldrLocales.USER_LOCALE_ID && !cldrLocales.isValid(loc)) {
+    return;
```
btangmu, line 237: This is where setting an invalid locale is prevented

tools/cldr-apps/js/test/client/test-client.mjs
```diff
@@ -27,14 +27,14 @@ describe(`client test`, async function () {
...
+  const STATUS_URL = new URL(`SurveyAjax?what=status`, baseURI);
+  it(`Should be able to retrieve status from ${STATUS_URL}`, async () => {
+    const r = await fetch(STATUS_URL);
+    // console.dir(r);
+    expect(r).to.be.ok;
+    const body = await r.json();
+    expect(r).to.be.ok;
+    // const {surveyOK, isBusted,isSetup,status,triedToStartUp,wasInitCalled} = body;
+    // console.dir(body);
+  });
```
btangmu, line 39: Only formatting is changed in this file, must have been skipped in an earlier merged PR

srl295 (approved):

PR #4709, opened 2025-05-16
CLDR-18539 root,en: Add availFmts Hv,hv to match intervalFmts; add availFmts EBh,Eh,EH to match ones with m

pedberg-icu:
CLDR-18539

- Add availableFormats for hv,Hv where we have intervalFormats for those skeletons in the same calendar (formats can come from the intervalFormat)
- Add availableFormats for
    - EBh where we have an availableFormat for EBhm in the same calendar
    - Eh where we have an availableFormat for Ehm in the same calendar
    - EH where we have an availableFormat for EHm in the same calendar
    - The formats can come from the corresponding existing format, just dropping the ":mm", except that...
- When adding formats for H without m (or s) as for the skeletons Hv and EH, suffix the "HH" in the pattern with literal 'h' for disambiguation as was done for my earlier PR today for already-existing formats, PR [#4708](https://github.com/unicode-org/cldr/pull/4708) for [CLDR-18538](https://unicode-org.atlassian.net/browse/CLDR-18538)

ALLOW_MANY_COMMITS=true

[CLDR-18538]: https://unicode-org.atlassian.net/browse/CLDR-18538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -1916,9 +1916,13 @@ annotations.
 						<dateFormatItem id="Bhms">h:mm:ss B</dateFormatItem>
 						<dateFormatItem id="d">d</dateFormatItem>
 						<dateFormatItem id="E">ccc</dateFormatItem>
+						<dateFormatItem id="EBh">E h B</dateFormatItem>
 						<dateFormatItem id="EBhm">E h:mm B</dateFormatItem>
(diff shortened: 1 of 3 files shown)


sffc (approved):
Seems about right; thanks!

PR #4710, opened 2025-05-16
CLDR-18655 Enable people at Meta to contribute to other locales provisionally at Moderate Level

cgaarden:
CLDR-18655

Enable people at Meta to contribute to other locales provisionally at Moderate Level

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/test/resources/org/unicode/cldr/web/data/TestSTFactory.xml
+++ b/tools/cldr-apps/src/test/resources/org/unicode/cldr/web/data/TestSTFactory.xml
@@ -24,7 +24,7 @@
     <verify status="approved" locale="el" xpath="//ldml/localeDisplayNames/languages/language[@type='fr']" >the price of tea in greece</verify>
 
     <!-- cannot vote in locale ssy since this locale isn't listed for this org in Locales.txt -->
-    <user name="v_org_not_ssy" org="facebook" level="vetter" locales="*"/>
+    <user name="v_org_not_ssy" org="cldr" level="vetter" locales="*"/>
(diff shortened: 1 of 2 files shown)


AEApple:
We're trying to get rid of the star, it would be better if you update with the list of actual locales you plan to contribute to.

srl295:
> We're trying to get rid of the star, it would be better if you update with the list of actual locales you plan to contribute to.

without the star, it adds it to the list of "TC Org Locales" - is that desired?

AEApple:
> > We're trying to get rid of the star, it would be better if you update with the list of actual locales you plan to contribute to.
> 
> without the star, it adds it to the list of "TC Org Locales" - is that desired?

So when we're doing resolution it is helpful to know which orgs have a contact for which language, so having an accurate representation in locales.txt is helpful.

srl295:
> > > We're trying to get rid of the star, it would be better if you update with the list of actual locales you plan to contribute to.
> > 
> > 
> > without the star, it adds it to the list of "TC Org Locales" - is that desired?
> 
> So when we're doing resolution it is helpful to know which orgs have a contact for which language, so having an accurate representation in locales.txt is helpful.

That information should come dynamically from the list of who actually has vetters assigned. And I think there should be a view that shows just that.

A "tc org locale" has a higher bar for changes and other coverage implications, that may not be meant here if they are not making it part of their general coverage.

srl295:
@cgaarden Yes, indeed, there was a unit test that assumed that Facebook (sic) had no vetters who could vote for `ssy` (i.e. no asterisk).  I moved it to the Cldr organization.

srl295 (approved):
if nobody is seeing stars this LGTM

AEApple (approved):

PR #4711, opened 2025-05-16
CLDR-14409 fe test: update docs, add novnc

srl295:
- explain how to vnc into the webdriver container
- explain how to run the client (BE) tests

CLDR-14409

ALLOW_MANY_COMMITS=true

with this change you only need maven and docker and a browser, and you can view the UI tests

image


pedberg-icu (approved):
Helpful, thanks!

PR #4721, opened 2025-05-20
CLDR-18647 IllegalArgumentException in ExampleGenerator

btangmu:
-Do not throw exception from XMLSourceMapOverride if locale is en = COMPARISON_LOCALE

-Enlarge testLightSpeed to include locale am (Amharic) and path that threw the exception

-Make delegate and overrides final per compiler suggestion

CLDR-18647

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
@@ -8,7 +8,12 @@ import org.unicode.cldr.util.XPathParts.Comments;
 
 public class CLDRFileOverride extends CLDRFile {
 
-    private static boolean threwException = false;
+    /**
+     * Silently ignore missing paths for the locale used for the Comparison Value column in Survey
+     * Tool. This happens for some light-speed unit paths with @case="accusative", which have values
+     * for some locales such as Amharic ("am") but not for English.
+     */
+    static final String COMPARISON_LOCALE = "en";
 
     /**
      * Overrides the values of the sourceFile. The keys of the overrides must already be in the
@@ -26,20 +31,19 @@ public class CLDRFileOverride extends CLDRFile {
      * same, since it provides for the map.
      */
     public static class XMLSourceMapOverride extends XMLSource {
-        private XMLSource delegate;
+        private final XMLSource delegate;
 
-        private Map<String, String> overrides;
+        private final Map<String, String> overrides;
 
         public XMLSourceMapOverride(XMLSource source, Map<String, String> overrides) {
             overrides.keySet().stream()
                     .forEach(
                             x -> {
                                 if (source.getValueAtDPath(x) == null) {
-                                    System.out.println(
-                                            "source.getValueAtDPath null in XMLSourceMapOverride");
-                                    if (!threwException) { // only throw once
-                                        threwException = true;
-                                        throw new IllegalArgumentException();
+                                    String loc = source.getLocaleID();
+                                    if (!COMPARISON_LOCALE.equals(loc)) {
+                                        throw new IllegalArgumentException(
+                                                "loc=" + loc + "; path=" + x);
                                     }
                                 }
                             });
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
@@ -8,6 +8,8 @@ import org.unicode.cldr.util.XPathParts.Comments;
 
 public class CLDRFileOverride extends CLDRFile {
 
+    private static boolean threwException = false;
+
     /**
      * Overrides the values of the sourceFile. The keys of the overrides must already be in the
      * XMLSource source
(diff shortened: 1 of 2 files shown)


btangmu:
This PR, as of the first commit, does not solve the bug, but it logs the exception only once rather than a hundred or more times as in recent log files. Also, it adds a problematic locale/path to an existing unit test.

As I mentioned in the ticket, the bug seems to involve an assumption that if a path is present in a CLDRFile, it will also be present in the corresponding XMLSource, that is `sourceFile.dataSource` here:

```
public CLDRFileOverride(CLDRFile sourceFile, Map<String, String> valueOverrides) {
    super(new XMLSourceMapOverride(sourceFile.dataSource, valueOverrides));
}
```

That assumption seems not to hold. It's not clear to me yet how to solve that. We could skip the paths for which the assumption fails; that would be pretty much equivalent to catching and ignoring the exception, and not a very satisfactory solution.

It's unclear to me why the additional unit test data doesn't result in a test failure or exception or even printing "source.getValueAtDPath null in XMLSourceMapOverride" to the terminal.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -33,7 +35,12 @@ public XMLSourceMapOverride(XMLSource source, Map<String, String> overrides) {
                     .forEach(
                             x -> {
                                 if (source.getValueAtDPath(x) == null) {
-                                    throw new IllegalArgumentException();
+                                    System.out.println(
+                                            "source.getValueAtDPath null in XMLSourceMapOverride");
```
btangmu, line 39: This happens when running the server and going to http://localhost:9080/cldr-apps/v#/am/SpeedAcceleration/6e5a3d520195cbe2

Mysteriously it doesn't happen when running tests in spite of the new data in testLightSpeed

btangmu: The plot thickens: the failure occurs for locale en, when loading the page for locale am!

btangmu: The caller is:

```
        public String getDisplayExample() {
            String displayExample = null;
            if (displayName != null) {
                displayExample =
                        sm.getComparisonValuesExample().getNonTrivialExampleHtml(xpath, rawEnglish);
            }
            return displayExample;
        }
```

btangmu:
The 2nd commit fixes it.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -8,6 +8,13 @@
 
 public class CLDRFileOverride extends CLDRFile {
 
+    /**
+     * Silently ignore missing paths for the locale used for the Comparison Value column in Survey
+     * Tool. This happens for some light-speed unit paths with @case="accusative", which have values
+     * for some locales such as Amharic ("am") but not for English.
+     */
+    static final String COMPARISON_LOCALE = "en";
```
srl295, line 16: if it's fixed, is this still needed?

srl295 (approved):
I wonder if this should be a known issue instead?  But if this quells for now, sounds good

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
srl295: This class is currently only used in this situation, but it's named as if it were more generic (and indeed should be merged with delegatingxmlsource in cldr-apps).  

so i am wondering if this should be a TODO with a future ticket.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
```diff
@@ -24,16 +31,20 @@ public CLDRFileOverride(CLDRFile sourceFile, Map<String, String> valueOverrides)
...
 
-        private Map<String, String> overrides;
+        private final Map<String, String> overrides;
 
         public XMLSourceMapOverride(XMLSource source, Map<String, String> overrides) {
             overrides.keySet().stream()
                     .forEach(
                             x -> {
                                 if (source.getValueAtDPath(x) == null) {
-                                    throw new IllegalArgumentException();
```
srl295, line 36: @macchiati @btangmu  why is this check even here? The class comment says "the (overridden) keys must already be in the source"… but, why why is it an error to override a null DPath with a non-null one from `overrides`?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
btangmu: I don't know the implications well enough to address the TODO question. Meanwhile it's worthwhile merging something like this PR to avoid needlessly throwing exceptions and cluttering the logs.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFileOverride.java
srl295: I added a comment on https://unicode-org.atlassian.net/browse/CLDR-18453 linking back here.

PR #4729, opened 2025-05-21
CLDR-18670 ui: show progress on auto import

srl295:
CLDR-18670

ALLOW_MANY_COMMITS=true

### Review images

image

image

image

### DIFF
--- a/tools/cldr-apps/js/src/views/AutoImport.vue
+++ b/tools/cldr-apps/js/src/views/AutoImport.vue
@@ -10,11 +10,19 @@
 
   <p><a-spin v-if="!haveResult" :delay="100" /> {{ message }}</p>
 
+  <div class="progressSection" v-if="versionsTotal">
+    <a-progress type="dashboard" showInfo="false" :percent="percent" />
+    <p v-if="versionsTotal > versionsDone">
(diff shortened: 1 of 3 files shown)


srl295:
the progress bar is one step per CLDR version so it doesn't move very quickly but it at least shows some status (several minute process) instead of a dead page

macchiati:
For the future: could we build a cache map of the users' votes over all versions somewhere in the database before submission starts? That way that cache could be used for import, making it much faster. Of course, the cache would only keep the latest value for each path.

macchiati (approved):
Looks reasonable to me; but I'm not familiar with vue, so if Tom is around..

srl295:
> For the future: could we build a cache map of the users' votes over all versions somewhere in the database before submission starts? That way that cache could be used for import, making it much faster. Of course, the cache would only keep the latest value for each path.

we don't need to build a cache map in Java, we need to do the same work as a procedure on the SQL side. anyway - yes, it can be optimized quite a bit.

macchiati:
I meant "cache map" in an abstract sense, eg the equivalent in database.

Robbie086 (commented):
Ok

PR #4730, opened 2025-05-21
CLDR-16665 Missing English core values

macchiati:
CLDR-16665

English was missing some key-type pairs (mostly core).
Also some touchups.

Also found out why some paths were not showing up to users of the ST. The ExtraPaths don't replicate all of the static paths that were created by the old code in https://github.com/unicode-org/cldr/blob/maint/maint-46/tools/cldr-code/src/main/java/org/unicode/cldr/util/XMLSource.java#L1455C1-L1473C37

A more thorough-going fix to the 2nd problem needs to be done, but this should solve the immediate problem. (fingers crossed)

ALLOW_MANY_COMMITS=true


macchiati:
Finger-crossing didn't work, so I'll dig into the test failures.

AEApple (dismissed):

macchiati:
Annemarie approved this, but then I had to fix a test. So I'd appreciate a stamp since she may be occupied.

PR #4732, opened 2025-05-22
CLDR-18574 Make error messages clearer for YMD

macchiati:
CLDR-18574

The error messages should be much clearer now. It is easiest to review by looking at the test cases in tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java

I left a comment showing the structure.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
@@ -748,9 +748,9 @@ public class CheckDates extends FactoryCheckCLDR {
                     DateTimePatternGenerator.YEAR,
                     DateTimePatternGenerator.QUARTER,
                     DateTimePatternGenerator.MONTH,
-                    DateTimePatternGenerator.DAY,
                     DateTimePatternGenerator.WEEK_OF_MONTH,
                     DateTimePatternGenerator.WEEK_OF_YEAR,
+                    DateTimePatternGenerator.DAY,
                     DateTimePatternGenerator.WEEKDAY,
                     DateTimePatternGenerator.HOUR,
                     DateTimePatternGenerator.MINUTE,
@@ -780,22 +780,18 @@ public class CheckDates extends FactoryCheckCLDR {
         switch (key) {
             case "dateTimeFormatLength":
                 {
+                    // should be something like
+                    // ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/dateTimeFormatLength[@type="full"]/dateTimeFormat[@type="standard"]/pattern[@type="standard"]
+                    // {1}, {0}
+
                     int index0 = value.indexOf("{0}");
                     int index1 = value.indexOf("{1}");
                     if (index0 < index1) {
-                        return "Field out of order: {0}…{1}";
+                        return "The {0} field (=time) cannot come before the {1} field (=date), in a YMD calendar.";
                     }
                     return null;
                 }
             case "appendItem":
-                {
-                    int index0 = value.indexOf("{0}");
-                    int index1 = value.indexOf("{1}");
-                    if (index0 > index1) {
-                        return "Field out of order: {1}…{0}";
-                    }
-                    return null;
-                }
             case "dateFormatLength":
             case "timeFormatLength":
             case "availableFormats":
@@ -806,6 +802,8 @@ public class CheckDates extends FactoryCheckCLDR {
             default:
                 return null;
         }
+
+        String intervalPosition = "1st";
         // verify
         //  the order is the same as in expectedField
         //  there is no other field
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
```diff
@@ -403,10 +409,15 @@ public void testIso8601() {
...
         String[][] tests = {
             // only the first 6 elements of the path matter
+            {
+                "//ldml/dates/calendars/calendar[@type=\"iso8601\"]/dateFormats/dateFormatLength",
+                "M yy",
+                "Field yy is incorrect. For a YMD calendar, the year field cannot be truncated to 2 digits."
+            },
             {
                 "//ldml/dates/calendars/calendar[@type=\"iso8601\"]/dateFormats/dateFormatLength",
                 "M y",
```
macchiati, line 419: Sample test message: the path, the value, and then an error message (or null, if there is no problem)

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -780,22 +780,18 @@ public static String checkIso8601(String path, String value) {
...
                 {
+                    // should be something like
+                    // ldml/dates/calendars/calendar[@type="gregorian"]/dateTimeFormats/dateTimeFormatLength[@type="full"]/dateTimeFormat[@type="standard"]/pattern[@type="standard"]
+                    // {1}, {0}
+
                     int index0 = value.indexOf("{0}");
                     int index1 = value.indexOf("{1}");
                     if (index0 < index1) {
-                        return "Field out of order: {0}…{1}";
+                        return "The {0} field (=time) cannot come before the {1} field (=date), in a YMD calendar.";
```
AEApple, line 790: Consider saying YMD (Year-first) or vice-versa

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
```diff
@@ -440,18 +451,18 @@ public void testIso8601() {
             },
             {
                 "//ldml/dates/calendars/calendar[@type=\"iso8601\"]/dateTimeFormats/dateTimeFormatLength",
-                "{1} {0}",
-                "Field out of order: {1}…{0}"
+                "{0} {1}",
+                "The {0} field (=time) cannot come before the {1} field (=date), in a YMD calendar."
```
srl295, line 455:
```suggestion
                "In a YMD calendar, the {1} field must come before the {0} field."
```

- =time and =date are interesting but possibly superfluous
- changed to say what TO do, instead of emphasizing what went wrong.

srl295 (dismissed):
think the wording can be clearer but it's an improvement

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
srl295:
```suggestion
                        return "In a YMD calendar, the {1} field must come before the {0} field.";
```

see comment on test below

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -806,6 +802,8 @@ public static String checkIso8601(String path, String value) {
             default:
                 return null;
         }
+
+        String intervalPosition = "1st";
```
srl295, line 806: `RuleBasedNumberFormat(Locale.ENGLISH, RuleBasedNumberFormat.SPELLOUT).format(1, "%spellout-ordinal")`

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -821,18 +819,28 @@ public static String checkIso8601(String path, String value) {
...
+                }
+            }
+
+            // No year truncation
+
+            if (type == DateTimePatternGenerator.YEAR) {
+                if (field.toString().length() == 2) {
+                    return "Field "
+                            + field
+                            + " is incorrect. For a YMD calendar, the year field cannot be truncated to 2 digits.";
```
srl295:
```suggestion
                    return "Field "
                            + field
                            + " is incorrect. For a YMD calendar, the year field must be a single 'y'.";
```

Shouldn't it be always 'y'?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -841,7 +849,15 @@ public static String checkIso8601(String path, String value) {
...
-                    return "Field out of order: " + lastField + "…" + field;
+                    return "Field "
+                            + lastField
+                            + " cannot come before field "
+                            + field
+                            + (isInterval
+                                    ? " in the " + intervalPosition + " part of the range"
+                                    : "")
+                            + ". A YMD calendar is special: bigger fields must come before smaller ones even when it feels unnatural in your language. "
+                            + " Change the text separating the fields as best you can.";
```
srl295:
```suggestion
                    return "Field "
                            + field
                            + " must come before field "
                            + lastField
                            + (isInterval
                                    ? " in the " + intervalPosition + " part of the range"
                                    : "")
                            + ". A YMD calendar is special: bigger fields must come before smaller ones even when it feels unnatural in your language. "
                            + " Change the text separating the fields as best you can.";
```

> A YMD calendar is special: bigger fields must come before smaller ones even when it feels unnatural in your language.
> Change the text separating the fields as best you can.

This is like an easter egg - trigger this error, and you get two sentences explaining what a YMD calendar is and giving guidance!

Perhaps all of the YMD errs should simply link somewhere on the site and not try to explain YMD?

macchiati:
Made the recommended changes, so needs a rereview.

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: Not necessary for this test; we always want English.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: No, someone could have yyyy (minimum 4 digits), which is actually better for gregorian (but only slightly; not that many people's dates are before 1000 AD)

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: From the discussion in the meeting, it sounded like we really wanted to add more info directly in the error message, rather than have a link that people don't follow.

Another question: can we have markdown in error messages? If so, we could make many of them more informative.

macchiati:
Please merge when checks pass, if you see this; I have to step away.

PR #4733, opened 2025-05-22
CLDR-18229 Add tts keycap names

macchiati:
CLDR-18229

The tts forms were missing, so added them. Also added to root, so that other locales can add to them.

Minor tool changes.

Note for the future: these are the easiest kinds of changes to make:

Ran SearchXML with options:

* -p\\d\\u20E3.*tts
* -s/Users/markdavis/github/cldr/common/annotationsDerived

(needed refinement in the -p regular expression to get exactly the right paths)

Paste generated lines into modify_config.txt

(sometimes the data needs fixing; a spreadsheet is easiest, since the lines are tabbed)

Ran CLDRModify with options:

* -fk
* -s/Users/markdavis/github/cldr/common/annotations

Spot-checked the result, then copied the altered files into the right directory

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/annotations/af.xml
+++ b/common/annotations/af.xml
@@ -3898,15 +3898,25 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="₿">bitcoin | elektroniese geldeenheid | kriptogeld</annotation>
 		<annotation cp="₿" type="tts">bitcoin</annotation>
 		<annotation cp="0⃣">0 | nul | knoppiesimbool</annotation>
+		<annotation cp="0⃣" type="tts">knoppiesimbool: 0</annotation>
 		<annotation cp="1⃣">1 | een | knoppiesimbool</annotation>
+		<annotation cp="1⃣" type="tts">knoppiesimbool: 1</annotation>
 		<annotation cp="2⃣">2 | twee | knoppiesimbool</annotation>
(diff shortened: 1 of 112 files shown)


macchiati:
Not quite ready yet: French is having a problem.

macchiati:
Squashing, to fix a commit message.

AEApple (approved):

srl295 (approved):
Approving with several concerns and questions:

- should the new annotations be provisional?
- for that matter, should the new annotations be derived? It seems they are derived from `keycap: #` - do we expect them to be voted on?
- why is `modify_config.txt` even checked in? It seems like a temporary work file. Could we `.gitignore` it and then just attach it to the PR if you want a record?

macchiati:
> Approving with several concerns and questions:
> 
> * should the new annotations be provisional?
They don't have to be: they are build from already approved data.

> * for that matter, should the new annotations be derived? It seems they are derived from `keycap: #` - do we expect them to be voted on?

They are derived from RBNF data also. However, by adding them we allow locales without RBNF data to have better values. And it's a small number of paths.

> * why is `modify_config.txt` even checked in? It seems like a temporary work file. Could we `.gitignore` it and then just attach it to the PR if you want a record?

It has been useful to look back through PRs to see what was done; checking it in makes that frictionless.

macchiati:
Merging so that I can test smoke.

PR #4736, opened 2025-05-23
CLDR-14409 Make cldr-apps-webdriver a project module; minor fixes

btangmu:
-Add cldr-apps-webdriver as a module in tools/pom.xml, so it can be debugged in IntelliJ with right-click, Debug All Tests

-Fix comment for location of surveydriver.properties, now expected directly inside cldr-apps-webdriver directory

-Avoid some NullPointerExceptions in SurveyDriverCredentials.java if properties are not defined, by checking for null before calling toString

-Define PROPS_TIME_OUT_SECONDS_KEY consistently with other keys

-In pom.xml, change the name cldr-apps-webdriver to CLDR Apps Webdriver, for consistency with cldr-apps, cldr-code

-In pom.xml, add skipIfEmpty for maven-jar-plugin, maven-install-plugin, and maven-deploy-plugin

-In pom.xml, specify execution goal test-jar to prevent error: ... packaging for this project did not assign a file ...

CLDR-14409

ALLOW_MANY_COMMITS=true


srl295 (approved):
Looks great!

btangmu:
@srl295 Before merging this, I'm looking at the failures. "cldr-mvn / build (pull_request)Failing after 22m" links to a log including:
```
[INFO] Running org.unicode.cldr.surveydriver.AppTest
While reading surveydriver.properties java.io.FileNotFoundException: surveydriver.properties (No such file or directory)
java.io.FileNotFoundException: surveydriver.properties (No such file or directory)
```
Then, "cldr-mvn / Docker test cldr-apps ... Failing after 3m" links to a log including:
```
SurveyDriverDashboard.test i = 0
...
org.openqa.selenium.StaleElementReferenceException: 
stale element reference: stale element not found in the current frame
```

Looking at other PRs, there are different failures, so seemingly surveydriver.properties is found sometimes but not always. The inconsistencies are puzzling. For PR 4718, there is:

```
❌ Test failed, maybe timed out, waiting for id DashboardScroller to exist for http://cldr-apps:9080/cldr-apps/v#/cs//
```

srl295:
it's a test, so it's trying to test it.

maybe unlinking it is better so that it only runs when requested.

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriverCredentials.java
```diff
@@ -15,15 +15,15 @@ public class SurveyDriverCredentials {
     private static final String EMAIL_PREFIX = "driver-";
 
     /**
-     * cldr-apps-webdriver/src/test/resources/org/unicode/cldr/surveydriver/surveydriver.properties
-     * -- not in version control; contains a line WEBDRIVER_PASSWORD=...
+     * A file should exist: cldr-apps-webdriver/surveydriver.properties -- not in version control;
+     * contains WEBDRIVER_PASSWORD=..., etc.
```
srl295, line 19: i'd revert this

srl295:
> Actually the FileNotFoundException for surveydriver.properties was blocking this PR. That failure was in the required cldr-mvn / build, not in the currently optional cldr-mvn / Docker test.

but we do not *want* to run the test in this situation.

> The 3rd commit adds surveydriver.properties. It's not clear though whether that should be necessary. Dockerfile has:
> 
> `COPY surveydriver-docker.properties /workarea/surveydriver.properties`
> 
> Does SurveyDriverCredentials.java look for it in `/workarea/`?

looks for it in the current working directory

> 
> `InputStream stream = new FileInputStream(PROPS_FILENAME)`
> 
> Testing without Docker that means it's looked for at `tools/cldr-apps-webdriver/surveydriver.properties`...

right, the working diretory.

Another option is that we could throw an `AssumptionViolatedException` if the .properties file is missing - then the webdriver test will just get skipped.  However, then the test will 'silently fail' in cases where we want to assert that the test passes.

but i think it might be best to make the webdriver not be a unit test and make it an explicitly run executable- what do you think?  that will reach the goal of including it in the project hierarchy - which makes sense as a goal.

srl295:
@btangmu ok if i make changes to this PR as mentioned? maybe i can make a sub PR , i'll do that

btangmu:
> ok if i make changes to this PR as mentioned? maybe i can make a sub PR , i'll do that

Yes, anything you'd like. I'm mostly out of office today. I agree webdriver probably doesn't need to be a "test"

btangmu:
@srl295 I merged your sub-PR. That resulted in a "Commit message for 18c1d60 fails validation" which I guess could be fixed by squashing; I don't know if there's another way to fix it. Also there's a merge conflict on SurveyDriverDashboard.java which I'll look at now...

srl295:
> @srl295 I merged your sub-PR. That resulted in a "Commit message for 18c1d60 fails validation" which I guess could be fixed by squashing; I don't know if there's another way to fix it. Also there's a merge conflict on SurveyDriverDashboard.java which I'll look at now...

It needed to be a rebase merge - you could check it out locally and run "git commit --amend" and then fix it and then force push.

btangmu:
See https://github.com/unicode-org/cldr/pull/4800

PR #4738, opened 2025-05-26
CLDR-18561 hora peninsular

robertbastian:
CLDR-18561

Adds "hora peninsular" to es, es_ES (Spain), es_EA (Ceuta & Melilla), es_IC (Canaries).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/es.xml
+++ b/common/main/es.xml
@@ -4031,6 +4031,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<exemplarCity>↑↑↑</exemplarCity>
 			</zone>
 			<zone type="Europe/Madrid">
+				<long>
+					<generic>hora peninsular</generic>
+				</long>
 				<exemplarCity>↑↑↑</exemplarCity>
 			</zone>
 			<zone type="Africa/Addis_Ababa">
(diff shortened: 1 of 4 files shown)


macchiati (commented):
Apparently Spain has daylight savings. I think we might need 3 strings for that in ICU. @richgillam ?

robertbastian:
We could add "hora de verano peninsular" and "hora estándar peninsular", the ticket just only asked for "hora peninsular".

Also, there seems to be inconsistent usage of "hora" vs "horario", most metazone names use "hora", some use "horario", the specific region formats use "horario", but the generic one uses "hora".

AEApple:
> We could add "hora de verano peninsular" and "hora estándar peninsular", the ticket just only asked for "hora peninsular".
> 
> Also, there seems to be inconsistent usage of "hora" vs "horario", most metazone names use "hora", some use "horario", the specific region formats use "horario", but the generic one uses "hora".

- We probably should add for daylight and standard since those will be shorter than the automatically generated ones. I think they're not really using them which is why they aren't requesting them.
- Should we ask the vetters for feedback on the inconsistency?

robertbastian:
> Should we ask the vetters for feedback on the inconsistency?

Yes!

robertbastian:
We should also ask vetters whether to use "hora de verano peninsular" or "hora peninsular de verano" (and same for standard).

macchiati:
Looks like we can merge. Robert, do you have permissions for that?

robertbastian:
Do we want to finish the discussion first?

macchiati:
sure

robertbastian:
@AEApple can you check with vetters what the correct specific names are?

PR #4741, opened 2025-05-27
CLDR-9814 Recover Turkey metazone names

robertbastian:
CLDR-9814

ALLOW_MANY_COMMITS=true


AEApple:
@macchiati - Is there a way to add this to coverage instead of introducing English into the ar.xml file?

robertbastian:
also, why is it complaining about missing values for root? no metazone names are in root

macchiati:
I think we need to have metazones in root / or in ExtraPaths/Codefallback (@btangmu )

Note: adding the following doesn't have much effect. If Modern coverage locales should have it (which it appears, since you are adding so many), then they should go in the modern coverage metazone variable.

<coverageVariable key="%metazone100" value="(Acre|Almaty|Anadyr|Aqtau|Aqtobe|Kamchatka|Macau|Qyzylorda|Samara|Turkey)"/>

robertbastian:
> Note: adding the following doesn't have much effect. 

Did you mean to paste a link or something?

> If Modern coverage locales should have it (which it appears, since you are adding so many), then they should go in the modern coverage metazone variable.

I'm recovering what was deleted in 2012. The set of modern locales has probably changed since then.

macchiati:
We can do on staging.

macchiati (approved):

macchiati:
Let me know if you want me to merge. (If you don't have permissions we need to change that!)

PR #4743, opened 2025-05-27
CLDR-18573 Make locale validation contingent on cldrGui.LOAD_LOCALES_EARLY

btangmu:
-For now, prevent validation error due to locale map not being loaded yet

CLDR-18573

ALLOW_MANY_COMMITS=true


btangmu:
@srl295 @macchiati this should also get in before a push to production

PR #4748, opened 2025-05-28
CLDR-18683 Make ExampleCache thread-safe

btangmu:
-New TestExampleCache non-parallel (sequential) test passes, and parallel test fails (sometimes)

-There is no guarantee that the parallel test will fail, due to randomness

-With PATH_COUNT=100000 the parallel test will usually fail, taking about 1/3 second

CLDR-18683

ALLOW_MANY_COMMITS=true


btangmu:
> I expect it to fail here

As expected based on local testing, the test still fails after the 3rd commit:

```
   TestExampleCache {
    testNonParallel (0.163s) Passed
    testParallel


Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e449962", got null
Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e62604", got null
Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e684144", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e684182", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e822216", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e822250", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e822455", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e450065", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e62952", got null

Error:  (TestExampleCache.java:53)  Error: : Example should match immediately: expected "e62974", got null

...
```

macchiati:
I think the test (which is great) could just stop at the first failure.

btangmu:
> I think the test (which is great) could just stop at the first failure.

Good idea. Meanwhile, the 4th commit implements your suggestion to remove ExampleCacheItem. The test still fails.

btangmu:
```
  TestExampleCache {
    testNonParallel (0.018s) Passed
    testParallel

Error:  (TestExampleCache.java:68)  Error: : Example should match immediately: expected "e125008", got null
Error:  (TestExampleCache.java:68)  Error: : Example should match immediately: expected "e875011", got null
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/ThreadSafeMapOfMapOfMap.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        return map3.containsKey(key3);
+    }
+
+    /**
+     * Checks if the nested map is empty.
+     *
+     * @return true if the map contains no key-value mappings, false otherwise.
+     */
+    public boolean isEmpty() {
```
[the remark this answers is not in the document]
btangmu: We might not need isEmpty...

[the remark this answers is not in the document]
macchiati: I was thinking we probably don't — and thus don't need to think about it!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleCache.java
```diff
@@ -166,10 +101,34 @@ <T extends ClearableCache> T registerCache(T clearableCache, String... starredPa
...
+     * ExampleGenerator we're connected to.
+     *
+     * @return the example html or null
+     */
+    public String getExample(String xpath, String value) {
+        if (!cachingIsEnabled) {
+            return null;
+        }
+        String starredPath = pathStarrer.set(xpath);
+        String result = cache.get(starredPath, xpath, value);
```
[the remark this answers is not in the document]
btangmu: That makes sense. I'll study how to revise the unit test to fit that structure. Currently the unit test depends on separation between `get` and `put` to test 3 assertions:

```
    private void doOnePath(ExampleCache exCache, boolean firstPass, int i) {
        if (alreadyReportedFailure) {
            return;
        }
        String xpath = "//p" + i;
        String value = "v" + i;
        String example = "e" + i;
        String result = exCache.getExample(xpath, value);
        if (firstPass) {
            assertNull("Example should be null before putExample", result);
            if (result != null) {
                alreadyReportedFailure = true;
            }
            exCache.putExample(xpath, value, example);
            result = exCache.getExample(xpath, value);
            assertEquals("Example should match immediately", example, result);
        } else {
            assertEquals("Example should match later", example, result);
        }
        if (!example.equals(result)) {
            alreadyReportedFailure = true;
        }
    }
```

`assertEquals("Example should match immediately", example, result)` would be straightforward with `putIfAbsent`, with a version of `constructExampleHtml` returning "e123" for the path "//p123".

Equivalents of the other two, `assertNull("Example should be null before putExample", result)` and `assertEquals("Example should match later", example, result)` could be tested by passing in different versions of `constructExampleHtml` returning, say, "f123" and "g123" respectively, for the path "//p123".

btangmu:
The 7th commit removes the get/put methods, using computeIfAbsent instead

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleCache.java
```diff
@@ -124,7 +124,12 @@ public String computeIfAbsent(
         if (!cachingIsEnabled) {
             return null;
         }
-        String starredPath = pathStarrer.set(xpath);
+        // Temporary work-around for PathStarrer thread-safety bug: do not call
+        // pathStarrer.set in the trivial case where xpath does not contain brackets.
+        // This assumes starredPath equals xpath if xpath does not contain "[...]".
+        // TODO: Make PathStarrer thread-safe
+        // Reference: https://unicode-org.atlassian.net/browse/CLDR-18683
+        String starredPath = xpath.contains("[") ? pathStarrer.set(xpath) : xpath;
```
macchiati: That will not handle the vast majority of cases — so it isn't worth doing.

macchiati: I think the temporary fix is along the lines I suggested; it is enough for the ExampleCache to work right.

```
    private static ConcurrentHashMap<String, String> STAR_CACHE = new ConcurrentHashMap<>();
    /**
     * Thread-safe version, only STAR_PATTERN
     */
    static final String get(String path) {
        return STAR_CACHE.computeIfAbsent(path, x -> {
        XPathParts parts = XPathParts.getFrozenInstance(x).cloneAsThawed();
                    parts.setAttribute(i, key, STAR_PATTERN);
            }
        }
        return parts.toString();});
    }
```
(I haven't checked if this compiles or works, but something along these lines should)

btangmu: That last commit, with `String starredPath = xpath.contains("[") ? pathStarrer.set(xpath) : xpath;`, was to get the ExampleCache test passing, in case we want to merge the improvements for ExampleCache before we fix PathStarrer.

I'll try your proposed fix for PathStarrer.

btangmu: Mark, your proposed fix for PathStarrer seems to work well. I've added it to this PR. I've made a new ticket https://unicode-org.atlassian.net/browse/CLDR-18697 for making PathStarrer completely thread-safe.

btangmu:
@macchiati Tests pass locally with the last commit, temporarily disabling TestPersonNameFormatter.TestExampleDependencies

As discussed in the infrastructure meeting, maybe this is worth getting into production in spite of remaining problems.

macchiati (approved):

PR #4753, opened 2025-05-29
CLDR-18682 Make sure Dashboard is displayed if wanted when page is reloaded

btangmu:
-Call cldrDashContext.insert from cldrGui.completeStartupWithSession if appropriate

-Calling cldrDashContext.insert redundantly (when already visible) is harmless

CLDR-18682

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrGui.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrGui.mjs
@@ -127,6 +127,9 @@ function completeStartupWithSession() {
   // TODO CLDR-18681: Here is where auto import should be scheduled.
   cldrSurvey.updateStatus();
   cldrEvent.startup();
+  if (!cldrDashContext.isVisible() && cldrDashContext.shouldBeShown()) {
+    cldrDashContext.insert();
+  }
 }
 
 /**


srl295 (approved):
Do you think this is what was making it not show in the web driver?

btangmu:
> Do you think this is what was making it not show in the web driver?

I'm wondering the same thing! Maybe we'll find out...

btangmu:
> > Do you think this is what was making it not show in the web driver?
> 
> I'm wondering the same thing! Maybe we'll find out...

Maybe so, if surveydriver doesn't go through the left-side menu to choose a locale, but just chooses it implicitly in the URL.

srl295:
> > > Do you think this is what was making it not show in the web driver?
> > 
> > 
> > I'm wondering the same thing! Maybe we'll find out...
> 
> Maybe so, if surveydriver doesn't go through the left-side menu to choose a locale, but just chooses it implicitly in the URL.

oh so you mean going through the left side was required to set up the mechanism?

btangmu:
> oh so you mean going through the left side was required to set up the mechanism?

Not necessarily. There are various circumstances in which the dashboard gets opened... hard to generalize...

PR #4758, opened 2025-06-01
CLDR-18691 Search for annotations

srl295:
- add a new search type looking for cp= (tts)
- fix unit test factory issue

CLDR-18691

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
@@ -211,6 +211,12 @@ public class SearchManager implements Closeable {
             logger.finest(() -> "AddCodes made resolved " + locale);
 
             for (final String x : resolvedFile.fullIterable()) {
+                if (x.startsWith("//ldml/annotations/annotation")
+                        && x.contains("[@cp=\"" + request.value + "\"]")
+                        && x.contains("[@type=\"tts\"]")) {
+                    response.addResult(new SearchResult(x, "tts: " + request.value, locale));
+                }
+
                 final PathHeader ph = phf.fromPath(x);
                 if (!ph.getSurveyToolStatus().visible()) continue; // skip invisible paths
 
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
@@ -217,6 +217,7 @@ public class SearchManager implements Closeable {
                     if (x.contains("[@cp=\"" + cp + "\"]") && x.contains("[@type=\"tts\"]")) {
                         response.addResult(new SearchResult(x, "tts: " + cp, locale));
                     }
+                    continue; // Do not try to match otherwise code for annotation
                 }
 
                 final PathHeader ph = phf.fromPath(x);
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
@@ -17,6 +17,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.unicode.cldr.util.AnnotationUtil;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRLocale;
 import org.unicode.cldr.util.Factory;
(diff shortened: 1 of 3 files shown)


srl295:
passes unit tests, not able to try w/ server yet. could try staging.

AEApple:
> passes unit tests, not able to try w/ server yet. could try staging.

happy to try testing it on staging

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
```diff
@@ -211,6 +211,12 @@ private void addCodes(final String locale) {
             logger.finest(() -> "AddCodes made resolved " + locale);
 
             for (final String x : resolvedFile.fullIterable()) {
+                if (x.startsWith("//ldml/annotations/annotation")
+                        && x.contains("[@cp=\"" + request.value + "\"]")
```
macchiati: The cp will always be present.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
```diff
@@ -211,6 +211,12 @@ private void addCodes(final String locale) {
...
+                        && x.contains("[@type=\"tts\"]")) {
+                    response.addResult(new SearchResult(x, "tts: " + request.value, locale));
```
macchiati: Note that when you search for emoji that the user enters, you have to remove all the emoji variation selectors. You might already do this somewhere else.

tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSearchManager.java
```diff
@@ -128,6 +128,26 @@ void TestKorean() throws InterruptedException {
                 () -> testNoResult("3월3월", "ko"));
     }
 
+    /** <annotation cp="🕗" draft="contributed"> //ldml/annotations/annotation[@cp="🕗"] */
+    @Test
+    void TestEmoji() throws InterruptedException {
```
macchiati, line 135: You'll want to test both with and without the tts.

srl295: Do you mean the expected result is both with and without tts?

macchiati: No, what i mean is that searching for a value (like "cook") should find the tts version, and find it in the non-tts version. And searching for an emoji character (like "👩🏼‍🍳" should find both paths (with and without tts attribute) and searching cp _without_ emoji variant selector.

srl295: > searching for an emoji character (like "👩🏼‍🍳" should find both paths (with and without tts attribute)

That's what I meant about the expected result 

Ok on the keyword search- will work on that. Maybe use English for stability

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
srl295: I don't, thanks! I think there's a routine for that..

tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSearchManager.java
AEApple: Can we do the keyword search part on a follow on PR? If you can search the emoji and find its tts entry at least it takes you to the right part of the survey tool since they tts and keywords are next to each other.

I think it probably also needs to give a message if you search for a derived tts or keywords?

srl295: Sure, i'll test this and make it ready for review..

> I think it probably also needs to give a message if you search for a derived tts or keywords?

Um.. not sure I follow at all?

Can you make this a note on the ticket?

srl295: > "👩🏼‍🍳" should find both paths (with and without tts attribute

Does it really need to match both paths? Or just take you to the TTS one? Won't the TTS always be present?

srl295:
seems to work in unit tests and in the survey tool.

tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSearchManager.java
srl295: for now it just matches the TTS (per @AEApple ) - please add comments to the issue

tools/cldr-apps/src/test/java/org/unicode/cldr/web/TestSearchManager.java
```diff
@@ -128,6 +129,33 @@ void TestKorean() throws InterruptedException {
...
+        final String CP = Character.toString(0x1F557);
```
macchiati, line 136: Use an emoji that is composed, and has EVS characters (FE0F). An example is:

2764 FE0F 200D 1F525                                   ; fully-qualified     # ❤️‍🔥 E13.1 heart on fire

srl295: done

tools/cldr-apps/src/main/java/org/unicode/cldr/web/SearchManager.java
```diff
@@ -211,6 +212,14 @@ private void addCodes(final String locale) {
             logger.finest(() -> "AddCodes made resolved " + locale);
 
             for (final String x : resolvedFile.fullIterable()) {
+                if (x.startsWith("//ldml/annotations/annotation")) {
+                    final String cp = AnnotationUtil.removeEmojiVariationSelector(request.value);
+                    if (x.contains("[@cp=\"" + cp + "\"]") && x.contains("[@type=\"tts\"]")) {
+                        response.addResult(new SearchResult(x, "tts: " + cp, locale));
+                    }
+                    continue; // Do not try to match otherwise code for annotation
```
srl295, line 220: This prevents the  `X -keywords` "code" from  path header from matching as a substring.

PR #4768, opened 2025-06-01
CLDR-17710 Add kok and kok_Latn day periods

AEApple:
CLDR-17710

- Add day periods for kok and kok_Latn
- Merge languages with same day periods for readability since the comments don't get updated from the ST.
- Fix error due to the partial addition of the scn day periods in CLDR-17626

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/dayPeriods.xml
+++ b/common/supplemental/dayPeriods.xml
@@ -271,21 +271,6 @@
 			<dayPeriodRule type="evening1" from="18:00" before="20:00"/>	<!-- شام -->
 			<dayPeriodRule type="night1" from="20:00" before="04:00"/>	<!-- رات -->
 		</dayPeriodRules>
-		<dayPeriodRules locales="hi">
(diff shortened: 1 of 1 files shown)


AEApple:
It's pretty hard to read the day periods for which languages share day periods right now. Could we refactor at some point to make sure that if languages have the same day period they're grouped?

Also the charts look wrong. Hindi and Marathi have midnight in the Survey Tool but not charts: https://www.unicode.org/cldr/charts/46/supplemental/day_periods.html#hi

AydaTolaymat:
Approved

macchiati:
Yes, we could have a unit test to detect when rules should be grouped: good idea.

I looked at the latest charts https://www.unicode.org/cldr/charts/48/supplemental/day_periods.html#hi. It looks like we are omitting it from them. (The only way it shows up is in gsw, which looks strange anyway.

PR #4771, opened 2025-06-02
CLDR-18692 Fix DAIP annotation delimiters

macchiati:
CLDR-18692

- Broadened the number of look-alikes for | in annotations.
- Added test
- Ran over current annotations (to fix blue star)

Characters are now: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%EF%B8%B3%EF%B8%B1%E0%A5%A4%7C%EF%BD%9C%E2%8E%B8%E2%8E%B9%E2%8F%90%EF%BF%A8%E2%9D%98%5D

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/annotations/ja.xml
+++ b/common/annotations/ja.xml
@@ -1697,7 +1697,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 		<annotation cp="🧑‍🧒‍🧒" type="tts">家族: 大人一人と子供二人</annotation>
 		<annotation cp="👣" draft="contributed">裸足 | 足 | 足あと</annotation>
 		<annotation cp="👣" type="tts">足あと</annotation>
-		<annotation cp="🫆">指紋｜鑑識｜生体認証｜セキュリティ</annotation>
+		<annotation cp="🫆">セキュリティ | 指紋 | 生体認証 | 鑑識</annotation>
 		<annotation cp="🫆" type="tts">指紋</annotation>
 		<annotation cp="🦰">赤毛 | 赤毛の人 | 赤茶</annotation>
 		<annotation cp="🦰" type="tts">赤毛</annotation>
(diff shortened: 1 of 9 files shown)


AEApple:
Looks like this is the same error I ran into on https://github.com/unicode-org/cldr/pull/4768, and I suspect it's due to a scn change that merged recently. Checking to see if my PR fixed it or if we need to revert the day periods until the plurals are merged?

We need a way where if the build fails, the change that caused it to fail is immediately reverted. WDYT?

srl295:
> Looks like this is the same error I ran into on https://github.com/unicode-org/cldr/pull/4768, and I suspect it's due to a scn change that merged recently. Checking to see if my PR fixed it or if we need to revert the day periods until the plurals are merged?
> 
> 
> 
> We need a way where if the build fails, the change that caused it to fail is immediately reverted. WDYT?

The err is CLDR/TestAnnotations/TestUniqueness seems unrelated to scn?

AEApple:
Oh, wow! There's a lot of duplicate names in zh-HK. Is that new?

macchiati:
> Oh, wow! There's a lot of duplicate names in zh-HK. Is that new?

The change in separator causes the annotations to be sorted differently; possibly also removing redundant terms. So expected, given how many they had.

AEApple:
> > Oh, wow! There's a lot of duplicate names in zh-HK. Is that new?
> 
> The change in separator causes the annotations to be sorted differently; possibly also removing redundant terms. So expected, given how many they had.

Do we just remove the derived data for now that has collisions?

macchiati:
Needed to fix tests. Deleted one character that had collisions.

AEApple (approved):

macchiati:
Yes, the derived ones will be rebuilt later on.

PR #4774, opened 2025-06-03
CLDR-18637 st: don't show 'undefined' for rows without examples

srl295:
- was testing for truthiness of a selector result, need to check length

CLDR-18637

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true


srl295:
fixed ticket #

PR #4776, opened 2025-06-03
CLDR-18641 Add plural rules for kok and kok_Latn

AEApple:
CLDR-18641

- Add one form for kok and kok_Latn for Survey Tool will need to fix rules after push to production

ALLOW_MANY_COMMITS=true


btangmu (approved):
Rubber-stamp LGTM

PR #4778, opened 2025-06-03
CLDR-18706 update component info

michelleperham:
Update page to let people know that Components are now mandatory in Jira when filing a new issue

CLDR-18706

ALLOW_MANY_COMMITS=true


AEApple:
I think we should add a page with a description of what the different components are used for

macchiati (approved):

macchiati:
To fix the Jira ticket, the status needs to be changed, and then hit the ... and then update status.

michelleperham:
Accepted the Jira ticket. I agree the Annemarie's comment and will open a new ticket for it.

PR #4782, opened 2025-06-04
CLDR-5717 Remove unused metazones

robertbastian:
CLDR-5717

This PR removes all metazones that haven't been used for over 20 years, which is the cutoff decided in CLDR-5130.

It also fixes the coverage of some metazones that were in the inactive set, but weren't actually inactive.

We currently have two inactive metazones (used in the last 20 years, but not currently).

ALLOW_MANY_COMMITS=true


macchiati (approved):

AEApple (approved):

AEApple:
After this merges I'll spot check Smoke Test.

PR #4784, opened 2025-06-04
CLDR-18712 Don't include grammar for new units

macchiati:
CLDR-18712

Visible on staging,  https://cldr-staging.unicode.org/cldr-apps/v#/de/OtherUnitsMetric/4cb2101b01962374

Screenshot 2025-06-04 at 14 16 36

Katal is one of the new units. So appears to be working correctly.

ALLOW_MANY_COMMITS=true


macchiati:
@AEApple just adding you for reference.

macchiati:
I need a review on this (the build&test checks passed earlier, before a spotless formatting fix was checked in),

macchiati:
Steven and Tom are probably gone for the day; if someone else can review that would be great.

PR #4786, opened 2025-06-05
CLDR-18703 Run GenerateExampleDependencies; fix ExampleCache

btangmu:
-Fix ExampleCache for use when cachingIsEnabled is false, as it is for GenerateExampleDependencies

-Replace ExampleDependencies.java with newly generated version

-In GenerateExampleDependencies, use newer PathStarrer.computeIfAbsent in place of PathStarrer.set

-Improve RecordingCLDRFile performance with new methods enableRecording and disableRecording, to avoid unneeded recording such as when  GenerateExampleDependencies calls getStringValue directly

CLDR-18703

ALLOW_MANY_COMMITS=true


btangmu:
I tried re-enabling `TestPersonNameFormatter.TestExampleDependencies` and it still failed. Fixing that seems like a high priority.

btangmu:
GenerateExampleDependencies took 3.5 hours!

srl295 (dismissed):

btangmu:
Problem: the new `PathStarrer.computeIfAbsent` uses:
```
    public static final String STAR_PATTERN = "([^\"]*+)";
```

Other PathStarrer uses have the simpler pattern like:
```
   new PathStarrer().setSubstitutionPattern("*");
```

However, we still have:
```
/** Register the cache, so that it gets cleared when any of the paths change */
PersonNamesCache personNamesCache =
        exCache.registerCache(
                new PersonNamesCache(),
                "//ldml/personNames/sampleName[@item=\"*\"]/nameField[@type=\"*\"]",
                "//ldml/personNames/initialPattern[@type=\"*\"]",
                "//ldml/personNames/foreignSpaceReplacement",
                "//ldml/personNames/nativeSpaceReplacement",
                "//ldml/personNames/personName[@order=\"*\"][@length=\"*\"][@usage=\"*\"][@formality=\"*\"]/namePattern");
```

Which assumes the old pattern; it's not compatible. Let's not merge this until that's fixed!

btangmu:
The reason it's not compatible is `ExampleCache.update` relies on the precise starred path. It fails to update (clear) person-names example dependencies for this reason. There are also problems with `ThreadSafeMapOfMapOfMap`...

btangmu:
For example, `"//ldml/personNames/sampleName[@item=\"([^\"]*+)\"]/nameField[@type=\"([^\"]*+)\"]"`
is not same as `"//ldml/personNames/sampleName[@item=\"*\"]/nameField[@type=\"*\"]"`

macchiati (dismissed):
Looks good!

macchiati:
Note: we could have 2 caches, to keep from having to change all the generated files. 

    private static final ConcurrentHashMap<String, String> ASTERISK_CACHE = new ConcurrentHashMap<>();

    /** Thread-safe version, only ASTERISK='*' */
    public static String computeIfAbsentAsterisk(String path) {
        return ASTERISK_CACHE.computeIfAbsent(
                path,
                x -> {
                    XPathParts parts = XPathParts.getFrozenInstance(x).cloneAsThawed();
                            parts.setAttribute(i, key, ASTERISK);
                        }
                    }
                    return parts.toString();
                });
    }

macchiati:
I just checked, and the only substitutions are:

1. ([^"]*+) — the default, intended for regex
2. * — vast majority of explicit sets; these are cases where the starred pattern will never be used in a regex.
3. %A — a few,  intended for regex, notably for RegexLookup
4. × — only one (GenerateEnglishChanged), and could probably be easily converted to *

In a refactoring for thread-safety, we could have an enum of styles, then have a a cache for each enum.

btangmu:
The 2nd commit enables `TestPersonNameFormatter.TestExampleDependencies` to pass. Whatever problems remain, I think this is an improvement.

btangmu:
> 2 caches, to keep from having to change all the generated files

I'll study that... which generated files do you mean?

I think I'm done for today, though. You're welcome to merge if you approve it.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateExampleDependencies.java
```diff
@@ -99,15 +97,18 @@ private void addDependenciesForLocale(Multimap<String, String> dependencies, Str
...
             }
-            String starredB = pathStarrer.set(pathB);
-            cldrFile.clearRecordedPaths();
+
+            String starredB = PathStarrer.computeIfAbsent(pathB);
+            cldrFile.enableRecording();
             egTest.getExampleHtml(pathB, valueB);
             HashSet<String> pathsA = cldrFile.getRecordedPaths();
+            cldrFile.clearRecordedPaths();
+            cldrFile.disableRecording();
```
btangmu, line 106: We could merge getRecordedPaths, clearRecordedPaths, and disableRecording into a single method. Currently this is the only usage, but who knows how RecordingCLDRFile might be used for other purposes.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateExampleDependencies.java
macchiati: Approved. My comment above was referencing the other ticket for improving PathStarrer (later)

PR #4787, opened 2025-06-05
CLDR-18718 Disable DateOrder usage for now

macchiati:
CLDR-18718

Disable the test for now (I verified that it was never reaching any point where it could actually produce results).
In a later PR it can be re-enabled.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
@@ -58,6 +58,7 @@ import org.unicode.cldr.util.XPathParts;
 
 public class CheckDates extends FactoryCheckCLDR {
     private static final boolean DEBUG = false;
+    private static final boolean DISABLE_DATE_ORDER = false;
 
     static boolean GREGORIAN_ONLY = CldrUtility.getProperty("GREGORIAN", false);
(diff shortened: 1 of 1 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -58,6 +58,7 @@
 
 public class CheckDates extends FactoryCheckCLDR {
     private static final boolean DEBUG = false;
+    private static final boolean DISABLE_DATE_ORDER = false;
```
btangmu: @macchiati do you want it true (disabled), or false (not disabled)?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: Damn, I'd switched back and forth to compare times. Will reset.

macchiati: Good catch, thanks!

srl295:
```java
    private static final boolean DISABLE_DATE_ORDER = CldrUtility.getProperty("DISABLE_DATE_ORDER", true);
```

PR #4789, opened 2025-06-06
CLDR-18683 Remove inclusion of non-existent manage.jsp

btangmu:
CLDR-18683

ALLOW_MANY_COMMITS=true


btangmu:
I just realized I made this PR on the wrong ticket. It should have been https://unicode-org.atlassian.net/browse/CLDR-18651

PR #4792, opened 2025-06-06
CLDR-18721 Disable Regional Variants (sideways) menu

btangmu:
-Temporary change to address performance issues, if CLDR_LOCALE_CACHE_MAX increase turns out to be problematic or insufficient

CLDR-18721

ALLOW_MANY_COMMITS=true


btangmu:
I suspect this isn't necessary, and I suggest waiting to see the results of recent performances fixes, and keeping this for part of a back-up plan.

macchiati (approved):

srl295 (approved):

PR #4793, opened 2025-06-06
CLDR-7459 make sideways view be by-request

srl295:
CLDR-7459

also see #4792

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrSideways.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrSideways.mjs
@@ -8,7 +8,9 @@ import * as cldrLoad from "./cldrLoad.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
 
-const SIDEWAYS_DEBUG = false;
+const SIDEWAYS_DEBUG = true;
+/** If true only load when clicked */
+const LOAD_ON_DEMAND = true;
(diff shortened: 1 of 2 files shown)


btangmu:
Looks good on first glance; this is for later, not today, right?

macchiati:
How does it work? When you hit the button, does it load the sublocales and then change the button to a pulldown?

macchiati (dismissed):
Need more information on how it functions.

srl295:
> Need more information on how it functions.

You get the button instead of the menu if you push the button on the menu appears that's it

macchiati:
ok, sounds good.

macchiati:
Withdrew my review; will leave it up to Tom to review the JS

srl295:
> Withdrew my review; will leave it up to Tom to review the JS

Sounds good… I moved it back to draft because we also need to revert the earlier disablement or it won't have any effect

I should add also that the button disappears once clicked

btangmu:
"also need to revert the earlier disablement" -- that can be done by setting DISABLE_SIDEWAYS_MENU = false in cldrInfo.mjs (and maybe simplifying the code to remove that boolean if we don't anticipate needing it again, but that could be postponed)

srl295:
@btangmu ready to review

srl295:
confirmed OK after rebase

btangmu (approved):

PR #4795, opened 2025-06-09
CLDR-18731 BRS Generate charts for 48

macchiati:
Includes some formatting fixes, and generation of a TSV

CLDR-18731

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/chart_messages.html
```diff
@@ -45,44 +45,47 @@ <h1 align="center">Chart Messages</h1>
...
-						{O}. Hovering with the mouse shows a short description.</p>
+						processed. Languages with a small population in a territory may be omitted.</p>
+					<p>Official status is supplied where available, formatted as an abbreviation such as {O}. 
+						Hovering with the mouse shows a short description.</p>
 					<ul>
 						<li><b>Likely languages and scripts:</b>To see (and verify)
-							the likely languages and scripts for this subtag, click on the
-							country code.</li>
+							the likely languages, scripts, and regions, see <a  target='likely_subtags' 
+							href='https://www.unicode.org/cldr/charts/48/supplemental/likely_subtags.html'>Likely Subtags</a>.</li>
```
conradarcturus, line 77: use /48/ or /latest/?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/chart_messages.html
```diff
@@ -45,44 +45,47 @@ <h1 align="center">Chart Messages</h1>
 			</tr>
 			<tr>
 				<td>territory_language_information</td>
```
conradarcturus, line 47: For the moment, these edits are all fine. My only 2 notes are:
* You may want to add a clearer warning on usage "We strive to provide good data but these values are not necessarily comprehensive. We recommend using this as one of multiple sources of the ground truth for the languages used in each territory. <link>".
* We use the CIA Factbook less and less & we also use a UN source. Here's my old slide for this. I'm happy to make updates to it if you want to put it in the website.
Screenshot 2025-06-10 at 11 44 08


I haven't gotten a chance to send the email but I would like to completely revamp this page. Ranjit has a lot of good feedback about the table components. With Translation Commons I've been working on an open source website with language demographics and I plan to re-use some of those components when redoing this page. See my work-in-progress website https://translation-commons.github.io/lang-nav/?objectType=Language&view=Table&languageSchema=CLDR

macchiati:
@conradarcturus Is this ok to merge, and then follow up with improvements? (That is, is it better than what is there, even if not optimal?)

I have a limited time to make improvements since I'm out in October.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/chart_messages.html
macchiati: During development, it wants to be /dev/, but for the release it wants to be /latest/.

PR #4796, opened 2025-06-09
CLDR-18615 Detect page from xpath if obsolete page name is used

btangmu:
-Extend the existing usage of page=auto to handle obsolete/invalid page names

-On the front end, always include xpstrid in the page request, if available

-On the back end, do not reject the combination of page and xpstrid; instead use xpstrid as a fallback to determine the page

-New PageId.fromString is the same as PageId.forString, except it returns null instead of throwing ICUException; PageId.forString still has other callers; changing them seems risky and out of scope for this PR, except for WebContext.getPageId

-Fix some compiler warnings

-Comments

CLDR-18615

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -399,10 +399,16 @@ function getSingleRowUrl(theRow) {
 
 function getPageUrl(curLocale, curPage, curId) {
   let p = null;
-  if (curId && !curPage) {
+  if (curId) {
+    if (!curPage) {
+      curPage = "auto";
+    }
+    // xpstrid is normally only used on the server if page is "auto". However, sometimes a row is bookmarked
+    // and the page name changes, but xpstrid is still valid, and the bookmark can still be used. In that
+    // case the server will treat the obsolete page name the same as "auto", and use xpstrid to determine
+    // the correct current page name.
     p = new URLSearchParams();
     p.append("xpstrid", curId);
-    curPage = "auto";
   }
   const api = "voting/" + curLocale + "/page/" + curPage;
   return cldrAjax.makeApiUrl(api, p);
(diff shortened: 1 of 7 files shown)


tools/cldr-apps/js/src/esm/cldrTable.mjs
```diff
@@ -399,10 +399,16 @@ function getSingleRowUrl(theRow) {
 
 function getPageUrl(curLocale, curPage, curId) {
   let p = null;
-  if (curId && !curPage) {
+  if (curId) {
+    if (!curPage) {
+      curPage = "auto";
```
srl295, line 404: so `#/mt//49b63607f40d4344`  (current usage) gets mapped to `#/mt/auto/49b63607f40d4344` ?

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
```diff
@@ -155,7 +158,7 @@ private static Response handleGetRows(ArgsForGet args) {
             return Auth.noSessionResponse();
         }
         try {
-            /** if true, hide emails. TODO: CLDR-16829 remove this parameter */
+            /* if true, hide emails. TODO: CLDR-16829 remove this parameter */
             final boolean redacted =
                     ((mySession.user == null) || (!mySession.user.getLevel().isGuestOrStronger()));
```
srl295, line 163: FWIW this was the line recently discussed about emails hidden for guest

tools/cldr-apps/js/src/esm/cldrTable.mjs
btangmu: This change only affects the http request, not (directly) what appears in the address bar, since curPage here is a local variable. The significant change in this function is that the back end always gets curId (if defined), to use as a fallback using the same mechanism as "auto", if curPage turns out to be obsolete/invalid.

PR #4802, opened 2025-06-10
CLDR-18744 Revise flagged item message in Survey Tool

btangmu:
-If vr.requiredVotes is 50 or higher, show new message

-Link to https://cldr.unicode.org/translation/getting-started/guide#changing-protected-items

CLDR-18744

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -717,12 +717,11 @@ function updateRowVoteInfo(tr, theRow) {
       )
     );
   } else if (vr.requiredVotes) {
-    var msg = cldrText.sub("explainRequiredVotes", {
-      requiredVotes: vr.requiredVotes,
-    });
-    tr.voteDiv.appendChild(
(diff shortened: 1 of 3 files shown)


btangmu:
image

btangmu:
The 2nd commit fixes a unit test

btangmu:
I just realized a possible problem with this PR. What if `requiredVotes` is nonzero but less than 50?

srl295:
This may not be right. Won't this show up even for items that only need 8?

btangmu:
I'll fix it to show new message only if requiredVotes >= 50

btangmu:
The third commit restores the old behavior for values less than 50

srl295:
> I'll fix it to show new message only if requiredVotes >= 50


sounds good because


> https://st.unicode.org/cldr-apps/v#/mt/Relative/11e7a3c102439781
>
>> Changes to this item require 4 votes.



> The third commit restores the old behavior for values less than 50

hard coded 50 isn't right but i don't see right off where it's passed down

it should be "≥TC vote" but lets leave that for future

macchiati:
Since Tom is out, I'll merge

PR #4807, opened 2025-06-11
CLDR-18697 Use static PathStarrer.computeIfAbsent where possible

btangmu:
-Assume that it is safe to use STAR_PATTERN instead of simple *

-Change starrer.set to static PathStarrer.computeIfAbsent if pattern is * or unchanged from default STAR_PATTERN, unless getAttributes is called on an instance

-Remove implements Transform and the transform method, apparently unused for PathStarrer

-Remove unused method getSubstitutionPattern

-Remove already-unused ExampleCache.pathStarrer; revise related comments

-Rename the 2-argument set method to setSkippingAttributes, to disambiguate at least for now; the 2-argument method has only a single caller, in TestAlt

-In the 1-argument set method, avoid calling the 2-argument method, for simplicity and performance, at least for now

CLDR-18697

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowInconsistentAvailable.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowInconsistentAvailable.java
@@ -54,7 +54,6 @@ public class ShowInconsistentAvailable {
     static FormatParser fp = new DateTimePatternGenerator.FormatParser();
     static Factory f = CONFIG.getCldrFactory();
     static PathHeader.Factory phf = PathHeader.getFactory();
-    static PathStarrer ps = new PathStarrer();
     static int counter = 0;
     static Set<String> nullErrors = new LinkedHashSet<>();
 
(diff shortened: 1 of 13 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/tool/ShowInconsistentAvailable.java
```diff
@@ -286,7 +285,7 @@ static void showInconsistenciesInLocale(String locale) {
                 continue;
             }
             if (SHOW_PROGRESS_RAW) {
-                ps.set(path);
+                PathStarrer.computeIfAbsent(path);
```
macchiati: This is strange, looks like the value is never used.

btangmu: Seems like that was already the case. Also, SHOW_PROGRESS_RAW is false, so maybe this is part of an incomplete unused feature.

btangmu: I've removed that line in the 2nd commit

macchiati (approved):

PR #4808, opened 2025-06-11
CLDR-18223 ui: update heading

srl295:
CLDR-18323

- [ ] This PR completes the ticket. (no, there may be other bugs)

image

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/MainHeader.vue
+++ b/tools/cldr-apps/js/src/views/MainHeader.vue
@@ -3,14 +3,21 @@
     <a-spin v-if="!loaded" :delay="250" />
     <ul>
       <li>
-        {{ stVersion }} {{ stPhase }}
-        <span
+        {{ stVersion }}
+        <a
+          href="https://cldr.unicode.org/translation/getting-started/survey-tool-phases"
+          :title="explainPhase"
+          _target="CLDR_ST_DOCS"
+        >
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/js/src/views/MainHeader.vue
```diff
@@ -189,9 +196,22 @@ export default {
       this.specialHeader = cldrStatus.getSpecialHeader();
       this.stVersion = "Survey Tool " + cldrStatus.getNewVersion();
       this.extendedException = cldrLoad.getLocaleInfo(loc)?.extended;
-      if (!loc || !this.extendedException) {
+      if (!loc) {
```
macchiati, line 199: This looks like it is using the "/" notation that I had in my email, for the "locale-less" Front page. I'd adjusted that in the ticket description to be nothing instead (June 11, 2025 at 2:36 PM). But it is ok to have the slash combo since that gives us something to hang the link on.

srl295: I think i opened this PR before I saw your other message.  (it's the front page or anywhere else where htere's no locale

macchiati: It works, so let's go with what you have here, with the /

PR #4813, opened 2025-06-13
CLDR-17707 plural rules for Chuvash "cv"

DavidLRowe:
CLDR-17707

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/cv.xml
+++ b/common/main/cv.xml
@@ -17023,7 +17023,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<pattern type="range" draft="unconfirmed">↑↑↑</pattern>
 		</miscPatterns>
 		<minimalPairs>
-			<pluralMinimalPairs count="other" draft="contributed">{0} кунсем</pluralMinimalPairs>
+			<pluralMinimalPairs count="zero"></pluralMinimalPairs>
+			<pluralMinimalPairs count="one"></pluralMinimalPairs>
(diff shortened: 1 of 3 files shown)


DavidLRowe:
Waiting on confirmation of ticket author and acceptance of ticket. (see ticket for details)

DavidLRowe:
@srl295 Based on the discussion in the ticket, I've added "cv" to the "zero,one,other" case for "ksh". 

I'm not totally clear on what significance the numbers in the pluralRule element have. I noticed that the ones for "blo" changed to match the "ksh" ones. I'm not sure if that was intentional or accidental.

srl295 (approved):

DavidLRowe:
It seems that the tests require the minimal pair examples be included in the PR. (I first tried making them blank, but that failed as well.)

AEApple:
> It seems that the tests require the minimal pair examples be included in the PR. (I first tried making them blank, but that failed as well.)

It may be something else and/or the phase we're in? I was able to add kok without the minimal pairs.

Oh, it's because you added them as empty instead of just not adding them at all. If you add them as empty they will collide :)

AEApple (approved):

DavidLRowe:
Thanks, @AEApple . I originally didn't include any minimal pair information, but evidently the examples that already existed for "other" were enough to cause the test failures. I tried making the examples empty, but (as you note) that didn't work. I suppose I could have removed them entirely and maybe that would have worked. I'm definitely still learning!

PR #4816, opened 2025-06-16
CLDR-17599 Test changing implementation of Hints to resemble PathDescription

btangmu:
-New PathDescriptionHints.md could replace hints.txt, similar format to PathDescription.md

-Work in progress: still use hints.txt rather than PathDescriptionHints.md, except for the test showing they match

-Possibly the data will be combined in a single file but for now PathDescriptionHints.md and PathDescription.md are separate files

-New PathDescription.getHintRawDescription serves the same purpose as old TranslationHints.get

-New unit test TestPathDescription.testHints confirms that PathDescription.getHintRawDescription returns the same value as TranslationHints.get for all paths in hints.txt

-Fix compiler warnings by deleting final from static final for two methods

CLDR-17599

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/PathDescription.java
@@ -23,6 +23,11 @@ public class PathDescription {
                     .lines()
                     .collect(Collectors.joining("\n"));
 
+    private static final String pathDescriptionHintsString =
+            CldrUtility.getUTF8Data("PathDescriptionHints.md")
+                    .lines()
+                    .collect(Collectors.joining("\n"));
+
     private static final Logger logger = Logger.getLogger(PathDescription.class.getName());
 
     public enum ErrorHandling {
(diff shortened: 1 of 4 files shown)


macchiati:
I think it's the right path to have two different files for now, we will probably want to have other different files for the HTML messages that currently exist( there are two of those and they want to be separate also.)

btangmu:
> I think it's the right path to have two different files for now, we will probably want to have other different files for the HTML messages that currently exist( there are two of those and they want to be separate also.)

In that case, maybe this ticket should be renamed. Currently "Merge Hints and PathDescription". If the distinction is to be preserved, a more accurate name is "Change implementation of Hints to be like PathDescription".

btangmu:
This draft PR is not meant for merging. It demonstrates by means of a temporary unit test that PathDescriptionHints.md can replace hints.txt and still match all the same paths.

See https://github.com/unicode-org/cldr/pull/4817 for a PR that actually removes hints.txt and TranslationHints.java.

macchiati:
agreed!

> In that case, maybe this ticket should be renamed. Currently "Merge Hints and PathDescription". If the distinction is to be preserved, a more accurate name is "Change implementation of Hints to be like PathDescription".

btangmu:
Closing this PR; it served its purpose as a demonstration. The subsequent PR https://github.com/unicode-org/cldr/pull/4817 was merged.

PR #4817, opened 2025-06-16
CLDR-17599 Change implementation of Hints to resemble PathDescription

btangmu:
-New PathDescriptionHints.md replaces hints.txt, similar format to PathDescription.md

-New PathDescription.getHintRawDescription serves the same purpose as old TranslationHints.get

-Remove hints.txt and TranslationHints.java

-Make DataPage.config static for usage by pathHintFactory

-Fix compiler warnings by deleting final from static final for two methods

CLDR-17599

ALLOW_MANY_COMMITS=true


btangmu:
See the earlier draft PR https://github.com/unicode-org/cldr/pull/4816 for confirmation that PathDescriptionHints.md can replace hints.txt and still match all the same paths.

btangmu:
The first commit still has the entire References section in PathDescriptionHints.md copied from PathDescription.md. I'll make another commit that removes that section, after making sure everything still works without it.

macchiati:
So it sounds like you verified that the hints.txt and PathDescriptionHints.md match precisely, right?

For both of these, I think we should take the time now to fix the format to the easier-to-maintain form that doesn't require backslashing so much, and uses variables. 

Note, the ### HINT doesn't need to be repeated.

- `^//ldml/units/unitLength\[@type="(narrow|short)"\]/unit\[@type="energy-foodcalorie"\]/(displayName|unitPattern\[@count="one"\]|unitPattern\[@count="other"\])$`

==>

- units/unitLength[@type='%unitLengths']/unit[@type='energy-foodcalorie']/displayName
- units/unitLength[@type='%unitLengths']/unit[@type='energy-foodcalorie']/unitPattern[@count='%anyAttribute']

btangmu:
> So it sounds like you verified that the hints.txt and PathDescriptionHints.md match precisely, right?

Yes, in the sense that all the paths in hints.txt match the appropriate regex in PathDescriptionHints.md. There's still a chance that a regex might also match a path missing from hints.txt. I'll add a test to the previous PR, cycling through all paths to make sure that doesn't happen.

btangmu:
> easier-to-maintain form that doesn't require backslashing so much

Do you mean a form that isn't used yet in PathDescriptions.md? Like in PathHeader.txt?

btangmu:
@macchiati assuming you do mean to employ syntax more like PathHeader.txt, that makes sense, though it seem like there's a risk of mission creep, relative to the original mission of requiring each user to check a box saying "I have read this..." for each hint and/or path description.

Currently, PathDescription.md and PathHeader.txt do both employ RegexLookup, but they employ it differently. Compare:

```
PathHeader.java:

public static class Factory implements Transform<String, PathHeader> {
    static final RegexLookup<RawData> lookup =
            RegexLookup.of(new PathHeaderTransform())
                    .setPatternTransform(RegexLookup.RegexFinderTransformPath)
                    .loadFromFile(PathHeader.class, "data/PathHeader.txt");
```
and

```
PathDescription.java:

private static final String pathDescriptionString =
        CldrUtility.getUTF8Data("PathDescriptions.md")
                .lines()
                .collect(Collectors.joining("\n"));

private static final RegexLookup<Pair<String, String>> pathHandling =
        parser.parse(pathDescriptionString);
```

It looks like we're talking about adding PathHeaderTransform and RegexFinderTransformPath to PathDescription...

macchiati:
This is just the format of the regexes, nothing would change in the
operation.

Mark

— Il meglio è l’inimico del bene —

btangmu:
> This is just the format of the regexes, nothing would change in the operation.

Would it make sense to merge this PR first, and then make further changes? Note that the draft PR tests did pass with the 2nd commit --

https://github.com/unicode-org/cldr/pull/4816

-- so there's a high degree of confidence that this PR is only verified safe refactoring.

macchiati:
Yes, we could merge this first.

Mark

— Il meglio è l’inimico del bene —

macchiati (approved):
Approving this, assuming a follow-on PR with the other changes.

PR #4819, opened 2025-06-16
CLDR-14737 Update adding-locales.md

DavidLRowe:
CLDR-14737

ALLOW_MANY_COMMITS=true


DavidLRowe:
Note: The ku-SY example under the "Country locales" heading is okay, however, script tags have been added to many of the current "ku*" locales now. There is no longer a ku_SY.xml, but rather ku_Latn_SY.xml. If anyone knows of a better example, please let me know.

DavidLRowe:
Or maybe this should reference [CLDR-16611](https://unicode-org.atlassian.net/browse/CLDR-16611) ?

[CLDR-16611]: https://unicode-org.atlassian.net/browse/CLDR-16611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,75 @@ title: Adding a new locale to CLDR
...
-		- Once he has added, run the tool in [LikelySubtags and Default Content](/development/updating-codes/likelysubtags-and-default-content) to add the new language and its associated subtags to common/supplemental/likelySubtags.xml
-	- Also add the English translation for any new languages in **common/main/en.xml**
-- If requested, add to vendor targets (Locale.txt), and to Cldr where resources are committed.
-- Run the tests (will be done automatically when a PR is created)
-- Commit your work to a branch and create a Pull Request.
-- The new locale will be included in Smoketest when the PR is merged, and will be in production once a push to production occurs.
+## Country locales
+
+If you are just adding a new country locale (that is, CLDR already has a locale for the language), you just need to add the new empty locale file, and Update Survey Tool. 
+[This example](https://github.com/unicode-org/cldr/pull/4335/files) adds Kurdish in Syria ("ku-SY").
```
srl295:
```suggestion
## Regional locales

If you are just adding a new region locale (that is, CLDR already has a locale for the language), you just need to add the new empty locale file, and Update Survey Tool. 
[This example](https://github.com/unicode-org/cldr/pull/4335/files) adds Kurdish in Syria ("ku-SY"). 
```

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,75 @@ title: Adding a new locale to CLDR
...
+	- time cycle (12-hour vs 24-hour, 0-based vs 1-based)
+	- orientation of characters on a line and lines on a page
+	- English name for this language (which won't be used until the language reaches Basic level, but needs to be documented in the CLDR ticket which adds the locale)
+
+### Add Data to the GitHub repository
+
+#### Verify or update country/population/language/script Data
+
+See [Update Language Script Info](https://cldr.unicode.org/development/updating-codes/update-language-script-info) and [LikelySubtags and Default Content](https://cldr.unicode.org/development/updating-codes/likelysubtags-and-default-content) for information on how to use the language, script, region and population data to verify that the information in [country_language_population.tsv](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv) and [language_script.tsv](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv) is correct. If changes are needed, correct an existing entry or add a new one as appropriate. Include references.
```
srl295:
```suggestion
See [Update Language Script Info](/development/updating-codes/update-language-script-info) and [LikelySubtags and Default Content](/development/updating-codes/likelysubtags-and-default-content) for information on how to use the language, script, region and population data to verify that the information in [country_language_population.tsv](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv) and [language_script.tsv](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv) is correct. If changes are needed, correct an existing entry or add a new one as appropriate. Include references. 
```

srl295 (approved):
Suggested improvements but its' an improvement so hitting approve.

The site infrastructure ought to have an issue to automatically remove a leading  `https://cldr.unicode.org` from URLs.

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,75 @@ title: Adding a new locale to CLDR
...
+	- NB: Even though the attribute is "regions", it can contain a region or a locale. Consider the example of "fr_CA" in the "regions" attribute of an "hours" element where the "preferred" attribute is "H", in contrast to "CA" which is in the "regions" attribute of an "hours" element where "preferred" attribute is "h". This corresponds to Canadian French (fr_CA) using 24-hour time even though 12-hour time is preferred for Canada (region CA).
+	- For the locale submission for "pap" (ticket [CLDR-14872](https://unicode-org.atlassian.net/browse/CLDR-14872) and pull request [PR#2542](https://github.com/unicode-org/cldr/pull/2542)), the region "CW" did not appear in the "timeData" information and needed to be added to the appropriate location.
+
+#### Optional additions
+- Plural rules are not needed until the locale needs to move beyond Basic level. If plural rule information is available, it could be added at the same time the locale is established. Add them to **common/supplemental/plurals.xml** and **common/supplemental/ordinals.xml**.
```
AEApple: Can we actually put that Plural rules are strong recommended? We've had several locales which have started contributing beyond basic without plural rules recently.

We may want to include what the impact is if plural rules are not available.

DavidLRowe: I added a "strongly recommended" line, but didn't quite know how to phrase the impact of not having them.

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,75 @@ title: Adding a new locale to CLDR
...
+	- NB: Even though the attribute is "regions", it can contain a region or a locale. Consider the example of "fr_CA" in the "regions" attribute of an "hours" element where the "preferred" attribute is "H", in contrast to "CA" which is in the "regions" attribute of an "hours" element where "preferred" attribute is "h". This corresponds to Canadian French (fr_CA) using 24-hour time even though 12-hour time is preferred for Canada (region CA).
+	- For the locale submission for "pap" (ticket [CLDR-14872](https://unicode-org.atlassian.net/browse/CLDR-14872) and pull request [PR#2542](https://github.com/unicode-org/cldr/pull/2542)), the region "CW" did not appear in the "timeData" information and needed to be added to the appropriate location.
+
+#### Optional additions
+- Plural rules are not needed until the locale needs to move beyond Basic level, but it is strongly recommended that plural rule information be collected as early as possible. If plural rule information is available, it can be added at the same time the locale is established. Add the rules to **common/supplemental/plurals.xml** and **common/supplemental/ordinals.xml**.
```
AEApple:
```suggestion
- Plural rules are not needed until the locale needs to move beyond Basic level, but it is **strongly recommended** that plural rule information be collected as early as possible.
   - The addition of new plural forms to a language after data collection at higher levels has begun will cause the coverage percentage to drop significantly until the data for plurals can be provided. The addition of new plural forms can also be disruptive for any localized applications which are relying on CLDR plural data.
   - If plural rule information is available, it can be added at the same time the locale is established. Add the rules to **common/supplemental/plurals.xml** and **common/supplemental/ordinals.xml**.
```

DavidLRowe:
@macchiati Would you have a few minutes to review this "Adding a locale" page?

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,77 @@ title: Adding a new locale to CLDR
...
+	- country / population data
+	- time cycle (12-hour vs 24-hour, 0-based vs 1-based)
+	- orientation of characters on a line and lines on a page
+	- English name for this language (which won't be used until the language reaches Basic level, but needs to be documented in the CLDR ticket which adds the locale)
+
+### Add Data to the GitHub repository
+
+#### Verify or update country/population/language/script Data
+
+See [Update Language Script Info](/development/updating-codes/update-language-script-info) and [LikelySubtags and Default Content](/development/updating-codes/likelysubtags-and-default-content) for information on how to use the language, script, region and population data to verify that the information in [country_language_population.tsv](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/country_language_population.tsv) and [language_script.tsv](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/language_script.tsv) is correct. If changes are needed, correct an existing entry or add a new one as appropriate. Include references.
```
macchiati: We should discuss guidelines for markdown. I strongly favor [semantic line breaks](https://sembr.org/) and [link references](https://github.github.com/gfm/#link-reference-definitions).

Both make future diffing and reading the raw text much easier. The semantic line breaks are very easy to to (except that we can't do them in tables).

AEApple: If you want to enforce them, we either need a lint or a tool that automatically processes the markdown otherwise it is too hard to ensure consistency.

DavidLRowe:
@macchiati @srl295 I think I've incorporated changes in response to feedback. Looking for reapproval.

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,113 @@ title: Adding a new locale to CLDR
...
+#### Supply metadata
+- In **common/supplemental/attributeValueValidity.xml**, add the language subtag to
+	- `<variable id\="$language" type\="choice">` for TC locales, or
+	- `<variable id\="$languageNonTcLtBasic" type\="choice">` for non-TC locales
+- In **common/supplemental/supplementalMetadata.xml**, add the appropriate default content locale to \<defaultContent locales\="..."\>
+	- The default content locale will include language and region (and possibly script, for multi-script languages).
+- Consider whether the locale needs parent locale information (which it probably won't)
+	- If parent inheritance can be determined by truncating subtags (and then falling back to root), no parent locale information is needed.
+In the majority of cases this will be true.
+	- If the locale is written in the non-preferred script for the language, it needs to have the parent specified as root,
```
macchiati:
```suggestion
	- If the locale is written in the non-default script for the language, it needs to have the parent specified as root,
```

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,113 @@ title: Adding a new locale to CLDR
...
+and pull request [PR#2542](https://github.com/unicode-org/cldr/pull/2542)),
+the region `"CW"` did not appear in the `"timeData"` information and needed to be added to the appropriate location.
+
+#### Optional additions
+- Plural rules are not needed until the locale needs to move beyond Basic level,
+but it is **strongly recommended** that plural rule information be collected as early as possible.
+   - The addition of new plural forms to a language after data collection at higher levels has begun
+will cause the coverage percentage to drop significantly until the data for plurals can be provided.
+The addition of new plural forms can also be disruptive for any localized applications which are relying on CLDR plural data.
+   - If plural rule information is available, it can be added at the same time the locale is established.
```
macchiati, line 107:
```suggestion
   - If plural rule information is available, it can be added at the same time the locale is established.
   - The simplest case is when the rules are identical to another locale's: it then just needs to be added to the list of locales with those rules.
```

docs/site/development/adding-locales.md
```diff
@@ -6,37 +6,113 @@ title: Adding a new locale to CLDR
...
+Add the rules to **common/supplemental/plurals.xml** and **common/supplemental/ordinals.xml**.
+- Similarly day period rules can be added to **common/supplemental/dayPeriods.xml**.
+
+## Submit the pull request
+
+- If requested, add to vendor targets in **tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt**,
+and to Cldr where resources are committed.
```
macchiati:
```suggestion
and to CLDR where resources are committed.
```

macchiati (changes requested):
Some small changes, OW looks good.

docs/site/development/adding-locales.md
DavidLRowe: "Cldr" was in the original text. I think it refers to the entry in Locales.txt (where it is "Cldr" rather than "CLDR"). I'll go ahead and include this change. It seems pretty clear either way.

macchiati (approved):

DavidLRowe:
Sorry, I added a bullet point so that the "Add the rules" line would be more obvious.

DavidLRowe:
@macchiati Needs (hopefully final!) reapproval after minor tweaks.

DavidLRowe:
@macchiati Need reapproval after incorporating your changes.

AEApple (approved):

AEApple:
This looks good to merge. We can always put in a follow on PR if needed if there are further changes after this initial update.

PR #4820, opened 2025-06-17
CLDR-18761 Indicate in the Info Panel whether a vote was imported

btangmu:
-New cldrInfo.voteCountTypeCell adds hover text to vote count

-Minor refactoring/cleanup/modernization on front end

-Enhance server methods for getting vote overrides to get vote type and date (new class VoteDetails)

-Fix PerUserData setting of VoteType.UNKNOWN per compiler warning

-Fix unrelated warning for explainPhase in MainHeader.vue

-Comments

CLDR-18761

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -536,22 +536,17 @@ function updateRowVoteInfo(tr, theRow) {
     console.error("theRow is null or undefined in updateRowVoteInfo");
     return;
   }
-  const vr = theRow.votingResults;
   tr.voteDiv = document.createElement("div");
   tr.voteDiv.className = "voteDiv";
   const surveyUser = cldrStatus.getSurveyUser();
-  if (theRow.voteVhash && theRow.voteVhash !== "" && surveyUser) {
+  if (theRow.voteVhash && surveyUser) {
     const voteForItem = theRow.items[theRow.voteVhash];
-    if (
-      voteForItem &&
-      voteForItem.votes &&
-      voteForItem.votes[surveyUser.id] &&
-      voteForItem.votes[surveyUser.id].overridedVotes
-    ) {
+    if (voteForItem?.votes[surveyUser.id]?.voteDetails?.override) {
       tr.voteDiv.appendChild(
         cldrDom.createChunk(
           cldrText.sub("override_explain_msg", {
-            overrideVotes: voteForItem.votes[surveyUser.id].overridedVotes,
+            overrideVotes:
+              voteForItem.votes[surveyUser.id].voteDetails.override,
             votes: surveyUser.votecount,
           }),
           "p",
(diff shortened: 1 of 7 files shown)


btangmu:
This adds a hover with the vote type and day(s) ago:

image

It doesn't add an icon. That could be done in a follow-up PR. There's some difficulty, though, involving the crufty DOM-based implementation with styles in surveytool.css. It would be easier if this part of the interface were modernized to use Vue.

macchiati (dismissed):
Lgtm. If @srl295 happened to have a few minutes to review it would be good, otherwise we could merge and test on smoke

btangmu:
@macchiati I looked at the changes again this morning.

I made a second commit, fixing an incorrect comment -- per our discussion the date is included after all, as "day(s) ago".

I think this is safe to merge.

tools/cldr-apps/js/src/esm/cldrInfo.mjs
```diff
@@ -773,32 +770,29 @@ function makeVoteExplainerDiv(voteTranscript) {
  * Update the vote info for one candidate item in this row, looping through all the orgs.
  * Information will be displayed in the Information Panel (right edge of window).
  *
- * @param theRow the row
- * @param vr the vote resolver
- * @param value the value of the candidate item
- * @param item the candidate item
- * @param vdiv a table created by the caller as vdiv = cldrDom.createChunk(null, "table", "voteInfo_perValue table table-vote")
+ * @param {Object} vr the vote resolver
```
srl295, line 773: resolver's results, technically.

srl295 (approved):
good improvements all round

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
```diff
@@ -60,6 +67,43 @@ public VoteEntry(User u, Integer override, boolean redacted) {
...
+        public final long daysAgo;
+
+        public VoteDetails(Integer override, VoteType voteType, Date date) {
+            this.override = override;
+            this.voteType = voteType;
+            this.daysAgo = daysSinceDate(date);
+        }
+
+        private long daysSinceDate(Date date) {
+            ZoneId zone = ZoneId.of("UTC+0");
```
srl295, line 93: I think there's a function for GMT? 

anyway this should go into DateUtils but could be a followup

tools/cldr-apps/js/src/esm/cldrInfo.mjs
btangmu: can update that later

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/VoteAPIHelper.java
btangmu: Which DateUtils do you mean?

PR #4824, opened 2025-06-20
CLDR-18765 Update development-process.md with phase changes

michelleperham:
CLDR-18765

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/development-process.md
+++ b/docs/site/development/development-process.md
@@ -58,11 +58,16 @@ When a ticket is in design, the owner is responsible for bringing back to the co
 
 |   |   |
 |---|---|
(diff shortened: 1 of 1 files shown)


michelleperham:
I updated https://cldr.unicode.org/development/development-process to list the new phases. I noticed the phases were out of date when I was looking to add component info.

DavidLRowe:
Under the "Assignees" heading, I think `http://unicode.org/cldr/trac/report/61` is a old link (that is redirecting to `https://cldr.unicode.org/requesting_changes#TOC-Filing-a-Ticket`). I'm not sure what the proper link should be.

PR #4827, opened 2025-06-23
CLDR-18370 Make atTime default for DateTimeFormatType per UTS 35 clarification

echeran:
According to discussion in today's CLDR design meeting, which clarifies ambiguities around "standard time" vs. "at time" for the "glue pattern" style of formatting when a datetime object contains both a date and a time, the default value should be "at time" in our case.

More specifically, "at time" should be considered the default for formatting a single datetime at a time, and "standard time" should be used in contexts in which multiple datetimes are formatted. In the context of how our test data cases are created, "at time" for a single datetime is appropriate description of this context, so that should be the default.

This change removes explicit references to `atTime` values in the test cases because that will be redundant. We are thus testing that implementations not only are capable of this behavior, but also perform this behavior by default. 

https://github.com/unicode-org/conformance/issues/469

CLDR-18370

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/testData/datetime/README.md
+++ b/common/testData/datetime/README.md
@@ -12,3 +12,14 @@ The test generator constructs the expected value using the various pieces
 (date format pattern, time format pattern, datetime "glue" pattern)
 and `SimpleDateFormat`s to combine them together.
 Each test case reports the inputs for the test case and the expected value.
+
+The UTS 35 LDML spec for datetime formatting will be updated in CLDR v48 to specify
+that the default value for dateTimeFormatType will be "atTime".
+dateTimeFormatType represents the value that indicates the type of
+datetime "glue" pattern, ex: indicating "atTime" or "standard" pattern.
+By datetime "glue" pattern, we mean the pattern that is used to combine the result
+of date-only formatting and time-only formatting to arrive at the overall combined
+formatting for the datetime object containing both a date and a time.
+Therefore, for test cases in the dataset in which a date and a time are both present
+in the datetime object, if a dateTimeFormatType is not specified explicitly, the
+value should be assumed to be "atTime".
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
@@ -59,7 +59,8 @@ public class GenerateDateTimeTestData {
 
     /**
      * The known set of values used to indicate the type of "glue pattern" aka the dateTimeFormat
-     * type.
+     * type. The default value should be assumed to be AT_TIME if a variable of type
+     * DateTimeFormatType is not set.
      *
      * <p>atTime = the word "at" is inserted between the date and time when formatting both date &
      * time together, at least for long and full dates.
@@ -353,9 +354,10 @@ public class GenerateDateTimeTestData {
         } else if (timeLength == null) {
             formattedDateTime = dateFormatter.format(zdt);
         } else {
-            assert (dateTimeGluePatternFormatType != null);
             String formattedDate = dateFormatter.format(zdt);
             String formattedTime = timeFormatter.format(zdt);
+
+            assert dateTimeGluePatternFormatType != null;
 
             formattedDateTime =
                     localeCldrFile.glueDateTimeFormat(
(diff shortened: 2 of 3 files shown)


sffc (commented):
Also please add the `"dateTimeFormatType": "standard"` on the semantic skeleton patterns that use glue

sffc (commented):
Whoops. It looks like currently the semantic skeleton test cases are unconditionally using the ICU4J glue pattern, which is almost certainly `atTime`, due to the TODO on line 111: "TODO: fix CLDR DateTimeFormats constructor to use CLDRFile to get the dateTimeFormat glue pattern rather than use ICU to get it". So I think we need to revert the change where we add `"standard"` here.

pedberg-icu (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -604,7 +604,10 @@ private static ImmutableSet<FieldStyleComboInput> getFieldStyleComboInputs() {
         elem.fieldStyleCombo = new FieldStyleCombo();
         elem.fieldStyleCombo.dateStyle = DateStyle.FULL;
         elem.fieldStyleCombo.timeStyle = TimeStyle.SHORT;
-        elem.fieldStyleCombo.dateTimeFormatType = DateTimeFormatType.STANDARD;
+        // TODO: uncomment once CLDR DateTimeFormats constructor to use CLDRFile to get the dateTimeFormat glue pattern,
+        //     since we are currently using ICU to get the dateTimeFormat pattern, which defaults to the behavior
+        //     of DateTimeFormatType.AT_TIME
+        // elem.fieldStyleCombo.dateTimeFormatType = DateTimeFormatType.STANDARD;
```
sffc: This comment only applies to the skeleton ones. You can use STANDARD for dateStyle/timeStyle.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDateTimeTestData.java
```diff
@@ -700,7 +703,10 @@ private static ImmutableSet<FieldStyleComboInput> getFieldStyleComboInputs() {
         elem.fieldStyleCombo.semanticSkeleton = SemanticSkeleton.MDTZ;
         elem.fieldStyleCombo.zoneStyle = ZoneStyle.SPECIFIC;
         elem.fieldStyleCombo.semanticSkeletonLength = SemanticSkeletonLength.SHORT;
-        elem.fieldStyleCombo.dateTimeFormatType = DateTimeFormatType.STANDARD;
+        // TODO: uncomment once CLDR DateTimeFormats constructor to use CLDRFile to get the dateTimeFormat glue pattern,
+        //     since we are currently using ICU to get the dateTimeFormat pattern, which defaults to the behavior
+        //     of DateTimeFormatType.AT_TIME
+        // elem.fieldStyleCombo.dateTimeFormatType = DateTimeFormatType.STANDARD;
```
sffc: It's not just "uncomment"; we want to add both atTime and standard once that feature is available, possibly in an extension (not in the kernel). I think just write a single comment explaining the plan and remove these 8 big block comments.

macchiati (dismissed):
Add to whatever describes the format of the file that what is being tested is the atTime form. Not sure where that is.

echeran:
> Add to whatever describes the format of the file that what is being tested is the atTime form. Not sure where that is.

Done. Added the docs to the existing Readme file that documents the output file. (We keep that Readme in the output folder since regular JSON as a file format doesn't allow comments.)

sffc (approved):

sffc:
@echeran Is this ready to merge?

common/testData/datetime/README.md
```diff
@@ -11,4 +11,15 @@ and various input datetime values.
 The test generator constructs the expected value using the various pieces
 (date format pattern, time format pattern, datetime "glue" pattern)
 and `SimpleDateFormat`s to combine them together.
-Each test case reports the inputs for the test case and the expected value.
\ No newline at end of file
+Each test case reports the inputs for the test case and the expected value.
+
+The UTS 35 LDML spec for datetime formatting will be updated in CLDR v48 to specify
+that the default value for dateTimeFormatType will be "atTime".
+We use dateTimeFormatType to represent the value that indicates tthe type of
```
macchiati: Minor, but we tend to avoid the imperial "we". ;)

These can be simply restated as descriptions:

The dateTimeFormatType represents the value...

macchiati (commented):
Just small wording recommendation.

common/testData/datetime/README.md
echeran: Good point. I reworded accordingly.

sffc (approved):

PR #4828, opened 2025-06-24
CLDR-18385 update langtags.json data

DavidLRowe:
CLDR-18385

This PR only updates the langtags.json data. Further processing is needed to extract the likely subtag information.

ALLOW_MANY_COMMITS=true


macchiati:
With a spot check, it looks good. I think the tool does ignore all fields we don't understand, so that should be ok. We should however, run the generation also so that we see the effect on likely subtags; that will help us to see what changes would happen. @conradarcturus do you have a bit of time to do that?

macchiati (approved):

conradarcturus:
Thanks for waiting, I ran the scripts to regenerate likely subtags and updated the PR. Now you can clearly see the LikelySubtag changes in the file diff.

macchiati:
@DavidLRowe I assume the deletions like the following are intentional (due to deprecated codes or similar).

`<likelySubtag from="tcn" to="tcn_Tibt_NP" origin="sil1"/> <!--Tichurong‧?‧? ➡ Tichurong‧Tibetan‧Nepal-->`

DavidLRowe:
@macchiati Yes, I think so. In the case of "tcn", the default script changed from Tibt to Deva, so the likely subtag entry changed accordingly.

PR #4830, opened 2025-06-26
CLDR-17707 adjust "cv" plural rule

DavidLRowe:
CLDR-17707

- [X] This PR completes the ticket (building on the previous PR)

The previous PR on this ticket added plural rules for Chuvash "cv" after identifying the need for "zero", "one", "other", but in the interests of quickly enabling the extra categories, "cv" was added to the existing "ksh" entry. This PR removes "cv" from the "ksh" entry and makes a separate "cv" entry.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/plurals.xml
+++ b/common/supplemental/plurals.xml
@@ -81,7 +81,12 @@ CLDR data files are interpreted according to the LDML specification (http://unic
             <pluralRule count="one">n = 1 @integer 1 @decimal 1.0, 1.00, 1.000, 1.0000</pluralRule>
             <pluralRule count="other"> @integer 2~17, 100, 1000, 10000, 100000, 1000000, … @decimal 0.1~0.9, 1.1~1.7, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, …</pluralRule>
         </pluralRules>
-        <pluralRules locales="cv ksh">
+        <pluralRules locales="cv">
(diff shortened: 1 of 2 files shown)


DavidLRowe:
Initially (after the first commit), the PR failed with the information:
```
TestDigitPluralCompleteness
Error:  (TestSupplementalInfo.java:747)  Error: : cv, 2, zero: expected true, got false
Error:  (TestSupplementalInfo.java:747)  Error: : cv, 3, zero: expected true, got false
Error:  (TestSupplementalInfo.java:747)  Error: : cv, 4, zero: expected true, got false
To fix the above, review the following, then replace in TestDigitPluralCompleteness
{"cv", "zero", "0"},	// n = 0

```
After adding the indicated line to TestSupplementalInfo.java, the tests passed. Hopefully that was the correct action to take to correct the test failure!

DavidLRowe:
Discussion of the categories and the numeric values for each is in the ticket [CLDR-17707](https://unicode-org.atlassian.net/browse/CLDR-17707)

[CLDR-17707]: https://unicode-org.atlassian.net/browse/CLDR-17707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4831, opened 2025-06-26
CLDR-18610 Revise VettingParticipation

btangmu:
-New VettingParticipation.vue, displays table instead of chits

-Revise cldrVettingParticipation.mjs extensively

-Revise SurveyVettingParticipation.java to skip if userIsManagerOrStronger or locale=*; minor refactoring

CLDR-18610

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyVettingParticipation.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyVettingParticipation.java
@@ -5,10 +5,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
 import org.unicode.cldr.util.CLDRLocale;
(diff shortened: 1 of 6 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/SurveyVettingParticipation.java
```diff
@@ -87,10 +80,10 @@ public void getJson(SurveyJSONWrapper r)
                 int theirId = rsu.getInt("id");
                 final User theUser = sm.reg.getInfo(theirId);
 
-                if ((theUser.userlevel > GUEST_STLEVEL)
-                        || UserRegistry.userIsLocked(theUser) // skip these
-                ) {
-                    continue;
+                if (UserRegistry.userIsLocked(theUser)
```
srl295, line 83: Ideally should be filtered on the client side or an option. Fine for now..

btangmu: For performance, filtering on the server side may be preferable to including unused data in the http response

srl295: Not likely to make a difference for this operation but fine for now

macchiati:
High bit; I think this is good enough to push now, even if there are some later improvements. But a few more people should try it out.

https://cldr-smoke.unicode.org/cldr-apps/v#vetting_participation///

BTW, I like the explicit button to generate, just in case someone ends up on the page by accident

It first appeared to just hang in smoke, in 'waiting'. Logged in as a google TC. 

Screenshot 2025-06-26 at 09 31 33

After a couple of minutes it started. 

**Minor suggestions:** at least for now print the reason it is waiting (someone else?), the time it took to wait, and the time it took to finish.

Notes: 

1. I think there is a way to get the detailed votes for a particular user, like the following. In a subsequent PR we should add a link on the Name cell to that view of the individual vetters' votes.

Locale | XpathId | XpathCode | Value | When
-- | -- | -- | -- | --
Finnish | 38295668bf23817d | Units	Weather	bar	long-other-partitive | {0} baaria | 6/25/2025
Finnish | 65eb0f4513b1e15 | Units	Weather	ofhg	narrow-other-nominative | {0} Hg | 6/25/2025
Finnish | 5f98282b06cc4e2f | Units	Weather	ofhg	narrow-one-nominative | {0} Hg | 6/25/2025

I myself am not sure how to actually see this view with the current active UI — nothing I click on in the https://st.unicode.org/cldr-apps/v#vetting_participation/// gets to it.

2. Longer term it would be great to have a chron job to autogenerate these daily during submission/vetting, at least for the TC org. That can be a separate ticket.

cf @AEApple

macchiati:
BTW, the link to put on the Name will be https://st.unicode.org/cldr-apps/v#recent_activity///3142 (where the number comes from the Vetter ID. (The link could go on Vetter ID instead).

PR #4834, opened 2025-06-26
CLDR-11977 Add Asia/Hanoi timezone

robertbastian:
CLDR-11977

ALLOW_MANY_COMMITS=true


robertbastian:
So the error here is because ICU doesn't include `backzone` zones. Should it?

macchiati (approved):

PR #4835, opened 2025-06-26
CLDR-16716 Missing Braille script metadata

robertbastian:
CLDR-16716

ALLOW_MANY_COMMITS=true


macchiati (commented):
Looks good to me, however, Markus should also review.

We could also add a couple of other changes, IMO

- Shaping is not required, so F8 = NO
- There is also discussion of about LB, so LB might become NO. Markus will know more of the status, but this might be coming in Unicode 17.

srl295:
it's also my understanding that braille is always LTR

macchiati:
Good catch, right

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Aname%09%3D%2FBRAILLE%2F%3A%5D-%5Cp%7Bbc%3Dl%7D

markusicu:
1. The source for this data is in a spreadsheet. For changes, we need to agree on what the data should be, and then I can update the spreadsheet and turn the crank.
2. Braille is a bit like Morse notation: We don't know what symbols mean without specifying a convention, like Unified English Braille. As such it's a bit odd to define some of the script properties.
3. This is why every Braille character is classified as a symbol. There are no letter characters in this script. We usually derive script properties from the letters of the script.
4. In the spreadsheet Brai looks like Zzzz -- almost all data values are "n/a".
5. What is the problem we are trying to solve?
6. The ID Usage is supposed to come from UAX31. It's currently "n/a". If we set it to Exclusion, then we should also propose to the UTC to add it to [table 4 there](https://www.unicode.org/reports/tr31/proposed.html#Table_Candidate_Characters_for_Exclusion_from_Identifiers).

PR #4837, opened 2025-06-26
CLDR-18789 Remove en-CA short time zone names

robertbastian:
CLDR-18789

ALLOW_MANY_COMMITS=true


AEApple:
Please hold off on this change until after the Survey Tool data from 48 has been imported into XML, otherwise it will get overwritten by the current data.

macchiati (approved):

PR #4839, opened 2025-06-28
CLDR-18175 update mg exemplars

DavidLRowe:
CLDR-18175

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/mg.xml
+++ b/common/main/mg.xml
@@ -286,7 +286,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	</localeDisplayNames>
 	<characters>
 		<exemplarCharacters>[aà b d eè f g h iì j k l m n oòô p r s t v yỳ z]</exemplarCharacters>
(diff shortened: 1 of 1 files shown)


DavidLRowe:
Since "ã" is found in the LDML data ("São Tomé-et-Príncipe"), I'm adding it to the auxiliary exemplar.

moyogo:
"São Tomé-et-Príncipe" is a French translation ("Sao Tomé-et-Principe" is the current French translation in CLDR but it is one of many).

The Malagasy translation seems to be ["Saô Tôme e Prinsipe"](https://mg.wikipedia.org/wiki/Sa%C3%B4_T%C3%B4me_e_Prinsipe) on the Malagasy Wikipedia, or "Sao Tome sy Principe" which redirects to it.

I don’t know to what extent this make a difference either way as it’s possible the French translation has been used at some point, just like in CLDR. It’s just odd to have the Portuguese-like French translation in Malagasy being used to define the auxiliary.

PR #4840, opened 2025-06-28
CLDR-18683 Investigate more general nested maps

macchiati:
CLDR-18683

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
@@ -1,0 +1,524 @@
+package org.unicode.cldr.util;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * ENGINE CLASS A generic, nested map structure that can handle a variable number of levels. This
+ * class provides the core logic and is wrapped by the type-safe shims.
+ *
+ * <p>Various nested Multimaps can be created by using an extra layer with Boolean. Multimap2<K1,
+ * K2, V> == NestedMap3<K1, K2, V, Boolean>
+ */
+public class NestedMap {
+
+    protected final Map<Object, Object> root;
+    private final Supplier<Map<Object, Object>> mapFactory;
+
+    // Examples: TreeMap::new, HashMap::new, ConcurrentHashMap::new
+    private NestedMap(Supplier<Map<Object, Object>> mapFactory) {
+        this.mapFactory = Objects.requireNonNull(mapFactory);
+        this.root = this.mapFactory.get();
+    }
+
+    public NestedMap(Map<Object, Object> immutableRoot) {
+        this.root = immutableRoot;
+        this.mapFactory = null; // will never change
+    }
+
+    @SuppressWarnings("unchecked")
+    // TODO: add comparators for TreeMaps
+    //    public static NestedMap createWithSortedMaps(Comparator<Object>... comparator) {
+    //        return new NestedMap(() -> (Map<Object, Object>) (Map<?, ?>) new
+    // TreeMap<>(comparator));
+    //    }
+
+    // --- CORE API METHODS ---
+    public void put(Object... keysAndValue) {
+        if (keysAndValue.length < 2) {
+            throw new IllegalArgumentException("At least one key and one value must be provided.");
+        }
+        Map<Object, Object> currentMap = this.root;
+        for (int i = 0; i < keysAndValue.length - 2; i++) {
+            Object key = keysAndValue[i];
+            currentMap =
+                    (Map<Object, Object>) currentMap.computeIfAbsent(key, k -> mapFactory.get());
+        }
+        Object finalKey = keysAndValue[keysAndValue.length - 2];
+        Object value = keysAndValue[keysAndValue.length - 1];
+        currentMap.put(finalKey, value);
+    }
+
+    public Object get(Object... keys) {
+        if (keys.length == 0) return null;
+        Map<Object, Object> currentMap = this.root;
+        for (int i = 0; i < keys.length - 1; i++) {
+            Object nextLevel = currentMap.get(keys[i]);
+            if (nextLevel instanceof Map) {
+                currentMap = (Map<Object, Object>) nextLevel;
+            } else {
+                return null;
+            }
+        }
+        return currentMap.get(keys[keys.length - 1]);
+    }
+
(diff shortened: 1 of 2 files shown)


btangmu:
This reminds me of ThreadSafeMapOfMapOfMap

macchiati:
> This reminds me of ThreadSafeMapOfMapOfMap

Yes, they are a generalization of that. The additional features are:
* the engine supports arbitrary levels (but for type-safety we'd add the generic shims at each level)
* It allows for any map as the basis, not just Concurrent
* You can even have different maps at different levels, eg 
    * `Map2<String, String, Integer> map2 = Map2.create(LinkedHashMap::new, TreeMap::new);`
* It corrects a problem with ThreadSafeMapOfMapOfMap.remove, since one needs to code it differently to avoid thread problems when one thread is trying to remove and another is adding

I anticipate replacing ThreadSafeMapOfMapOfMap, plus various other code, once this is available.

macchiati:
NOTE: I change the ticket number, but the branch has the old number in its messages

robertbastian:
Are there no libraries with a collection like this that we could use? Writing a correct, performant, concurrent collection in Java is hard, and we shouldn't reinvent the wheel.

macchiati:
I hadn't been able to find one that handles multiple levels. I did verify the usage of computeIfAbsent and computeIfPresent in order to handle put and remove.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
```diff
@@ -0,0 +1,689 @@
...
+    private Object getInternal(Object... keys) {
+        int limit = keys.length;
+        Map<Object, Object> currentMap = this.root;
+        Object nextLevel = null;
+        for (int i = 0; i < limit; i++) {
+            nextLevel = currentMap.get(keys[i]);
+            if (nextLevel instanceof Map) {
+                currentMap = (Map<Object, Object>) nextLevel;
+            } else {
+                return nextLevel;
```
robertbastian: if this is always called with the correct number of keys, then this branch should be unreachable?

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
```diff
@@ -0,0 +1,689 @@
...
+    @SuppressWarnings("unchecked")
+    private void putInternal(Object... keysAndValue) {
+        if (keysAndValue.length < 2) {
+            throw new IllegalArgumentException("At least one key and one value must be provided.");
+        }
+        Map<Object, Object> currentMap = this.root;
+        for (int i = 0; i < keysAndValue.length - 2; i++) {
+            Object key = keysAndValue[i];
+            Supplier<Map<Object, Object>> factory = mapFactories.get(i);
+            currentMap = (Map<Object, Object>) currentMap.computeIfAbsent(key, k -> factory.get());
```
robertbastian, line 173: `Supplier::get` is not guaranteed to return a new object every time:

https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html#:~:text=There%20is%20no%20requirement%20that%20a%20new%20or%20distinct%20result%20be%20returned%20each%20time%20the%20supplier%20is%20invoked.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
```diff
@@ -0,0 +1,689 @@
...
+                .computeIfPresent(
+                        key,
+                        (k, value) -> {
+                            // BASE CASE: We are at the last key or value
+                            if (index == keys.length - 1) {
+                                // We return null to remove the leaf object (C) from the map (B)
+                                return null;
+                            }
+                            // Recursive Case: The value is another map
+                            if (value instanceof Map) {
```
robertbastian: this will break if the user value implements `Map`. keep track of the depth instead

srl295:
> Are there no libraries with a collection like this that we could use? Writing a correct, performant, concurrent collection in Java is hard, and we shouldn't reinvent the wheel.

Maybe donate it to Guava? or at least make sure there's a bug report somewhere requesting it?

srl295:
> > Are there no libraries with a collection like this that we could use? Writing a correct, performant, concurrent collection in Java is hard, and we shouldn't reinvent the wheel.
> 
> Maybe donate it to Guava? or at least make sure there's a bug report somewhere requesting it?

What about https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Table.html and make the 'column' type be a composite (with multiple fields, etc. )

macchiati:
You can't choose it to be threadsafe, nor the type of map, and it is only 3 levels, and you can't have a 2 level multimap.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
srl295: true, though that is how  Collector.supplier seems to use it … https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collector.html#supplier--

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
macchiati: good catch, thanks

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
macchiati: Documented what happens: it returns the keyset for the last map. The shims make this type-safe.

Eg with <String, enum, String> the keys are normally <String, enum>. If you call get("foo") you get a set of the enums at the second level.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
macchiati: Yes, if someone makes a supplier that returns the same map, almost any usage of that will blowup.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
macchiati: I rewrote the code to fix that. It now uses the depth to figure out whether or not there should be a map. 

The one exception is NestedMapSpliterator, which I still need to fix. Will work on that.

macchiati:
I would like to merge this tomorrow. Can one of you review?

robertbastian:
There doesn't seem to be any changed code since my last review.

macchiati:
@robertbastian I had to squash, so you can't see the changes since your last review. There are substantial changes, and I think all of your comments have been resolved.

@srl295 @btangmu Robert is going to be OOO for a week. Can one of you review?

tools/cldr-code/src/main/java/org/unicode/cldr/util/NestedMap.java
srl295: guava's `Suppliers.memoize(Supplier<T>)` is purposefully a supplier that always returns the same object, that is how I usually use Supplier.  I think this is fine, though perhaps should be clarified.

macchiati:
Thanks!

tools/cldr-code/src/test/java/org/unicode/cldr/util/NestedMapTest.java
srl295, line 1: @macchiati i missed this before… NestedMapTest is both an old-style TestFmwk test, AND isn't linked in properly. would you like me to do a followon to enable this test?

PR #4857, opened 2025-07-02
CLDR-18790 Fetch initial ST status before other tasks

btangmu:
-New cldrSurvey.fetchInitialStatus, similar to updateStatus but async

-Await completion of cldrSurvey.fetchInitialStatus in cldrGui.initialSetup

CLDR-18790

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrGui.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrGui.mjs
@@ -72,6 +72,7 @@ async function initialSetup() {
     cldrLocales.fetchMap(),
     cldrEscaperLoader.updateEscaperFromServer(),
   ]);
+  await cldrSurvey.fetchInitialStatus();
   // This needs to be done before initial menus.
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/js/src/esm/cldrGui.mjs
```diff
@@ -72,6 +72,7 @@ async function initialSetup() {
     cldrLocales.fetchMap(),
     cldrEscaperLoader.updateEscaperFromServer(),
   ]);
+  await cldrSurvey.fetchInitialStatus();
```
srl295, line 75: can't this be moved up to line 73 so they are all in parallel? 

but approving because it's an improvement

btangmu: I don't think they can all be in parallel. Status requires session. If request status before getting response for session, it will fail.

srl295: that makes sense.

PR #4860, opened 2025-07-03
CLDR-18610 Add daysAgo to VettingParticipation; remove unused code/data

btangmu:
-Send daysAgo instead of last_mod; use daysAgo instead of user.time (which is when seen, not when voted)

-Remove some (not all) http response data that was unused

-Remove code that generated/manipulated unused data, including OrgCoverageLevelCounter.java

-Move function daysSinceDate from VoteAPIHelper.java to TimeDiff.java

-Change message from Starting... to Waiting for initial data...

-Comments

CLDR-18610

ALLOW_MANY_COMMITS=true


btangmu:
This removes OrgCoverageLevelCounter.java but we might want to bring some of its code back as part of further revision.

As it stands, when Admin requests the table for all orgs, there are over 1000 http requests (for each user+locale) and each one essentially causes the entire Dashboard to be generated even though only a few bytes of data (votedPathCount and votablePathCount) are actually used from each response. Most of the work is probably wasted/redundant. The table could probably be generated much more quickly.

For this ticket, we still need to add two columns:
* EMP is the sum of errors + missing + provisional (for locale)
* MP is the sum of missing + provisional (for locale)

If I understand correctly, those columns are not specific to one user or organization, but they do depend on the coverage level (which depends on the org).

BTW I've noticed the existence of VettingParticipation2.vue and localesTxtGenerator.mjs but I haven't studied yet what they are, or how they might relate to the current VP table.

srl295 (approved):

PR #4861, opened 2025-07-08
CLDR-18280 Change localized offset format to not use gmtZeroFormat

robertbastian:
CLDR-18280

ALLOW_MANY_COMMITS=true


macchiati (commented):
We need to say that the gmtZeroFormat has been deprecated; the regular format should be applied to zero instead.

macchiati:
And this is a major migration issue because it could break software; we should consider for 1 release maintaining the old data — but _generating_ it from an application to 0.

robertbastian:
> We need to say that the gmtZeroFormat has been deprecated

In a changelog or something?

macchiati:
Added https://unicode-org.atlassian.net/browse/CLDR-18280?focusedCommentId=185441 noting that.

macchiati (approved):

PR #4862, opened 2025-07-08
CLDR-18796 Fix bogus left sidebar locale links

btangmu:
-New function cldrLoad.linkToLocaleOnly for hash not including any page or row identifier

-Comments, including caution for the original error-prone cldrLoad.linkToLocale which is still used for everything but the left sidebar

CLDR-18796

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLoad.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrLoad.mjs
@@ -1009,6 +1009,7 @@ function myLoad(url, message, handler, postData, headers) {
 }
 
 function appendLocaleLink(subLocDiv, subLoc, subInfo, fullTitle) {
+  // Note: subLoc, etc., do not necessarily refer to sub-locales; they may also refer to top-level locales.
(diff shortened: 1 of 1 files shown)


btangmu:
Please note that this PR assumes that choosing a locale from the left sidebar should result in navigation to "General Info" for that locale, rather than to a different page or row.

The behavior of ST (prior to this PR) is inconsistent. Normally, choosing a locale from the left sidebar goes to General Info. However, if the URL in the address bar includes a page and row ID, and then you reload the page using the browser ⟳ button, and then choose a different locale, the page and row are preserved. I don't know if that was accidental or a half-broken intentional feature, but it only worked if you did the reload, so it seems unlikely anyone was making use of it intentionally.

Also, this PR is minimal, to fix the left sidebar. There are probably similar bugs remaining. A general solution would probably entail making a clear distinction between (at least) three categories of ID: row ID, user ID, and forum-post ID. An ID would never be set or used to construct a link without taking its category into account to avoid nonsensical combinations of special/locale/page/id.

PR #4863, opened 2025-07-08
CLDR-18697 Change more PathStarrer.set to computeIfAbsent

btangmu:
-Remove PathStarrer.getAttributes; instead use new XPathParts.getPathAttributes and XPathParts.getPathAttributesJoined

-Comments

CLDR-18697

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CopySubdivisionsIntoMain.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CopySubdivisionsIntoMain.java
@@ -15,9 +15,9 @@ import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.CLDRPaths;
 import org.unicode.cldr.util.Factory;
-import org.unicode.cldr.util.PathStarrer;
 import org.unicode.cldr.util.SimpleFactory;
 import org.unicode.cldr.util.SimpleXMLSource;
+import org.unicode.cldr.util.XPathParts;
 
 public class CopySubdivisionsIntoMain {
 
@@ -91,8 +91,6 @@ public class CopySubdivisionsIntoMain {
                     "Should probably discontinue this, leaving translated subdivisions in main");
         }
 
-        PathStarrer pathStarrer = new PathStarrer();
-
         for (String localeId : mainFactory.getAvailable()) {
 
             boolean isRoot = "root".equals(localeId);
@@ -132,8 +130,8 @@ public class CopySubdivisionsIntoMain {
 
                 String oldValue = subdivisionFile.getStringValue(path);
                 if (!Objects.equal(oldValue, value)) {
-                    pathStarrer.set(path);
-                    String firstAttributeValue = pathStarrer.getAttributes().iterator().next();
+                    String firstAttributeValue =
+                            XPathParts.getPathAttributes(path).iterator().next();
                     if (Objects.equal(firstAttributeValue, value)) {
                         System.out.println(
                                 localeId
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListCoverageLevels.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListCoverageLevels.java
@@ -34,7 +34,6 @@ import org.unicode.cldr.util.ChainedMap.M3;
 import org.unicode.cldr.util.ChainedMap.M4;
 import org.unicode.cldr.util.ChainedMap.M5;
 import org.unicode.cldr.util.Factory;
-import org.unicode.cldr.util.LanguageTagParser;
 import org.unicode.cldr.util.Level;
 import org.unicode.cldr.util.Organization;
 import org.unicode.cldr.util.PathStarrer;
@@ -63,9 +62,6 @@ public class ListCoverageLevels {
         CLDRConfig config = CLDRConfig.getInstance();
         StandardCodes sc = StandardCodes.make();
         SupplementalDataInfo sdi = config.getSupplementalDataInfo();
-        LanguageTagParser ltp = new LanguageTagParser();
-        Set<String> defaultContents = sdi.getDefaultContentLocales();
-        PathStarrer starrer = new PathStarrer().setSubstitutionPattern("*");
         Factory mainAndAnnotationsFactory = config.getMainAndAnnotationsFactory();
 
         Locales localesToTest = Locales.specific;
@@ -116,7 +112,7 @@ public class ListCoverageLevels {
                         coverageLeveler == null
                                 ? Level.COMPREHENSIVE
                                 : coverageLeveler.getLevel(path);
-                String skeleton = starrer.set(path);
+                String skeleton = PathStarrer.computeIfAbsent(path);
                 levelToCounter.get(level).add(skeleton);
                 if (path.startsWith("//ldml/units/unitLength")
                         && !path.contains("coordinateUnit")
@@ -190,8 +186,8 @@ public class ListCoverageLevels {
             if (path.endsWith("/alias")) {
                 continue;
             }
-            String starred = starrer.set(path);
-            List<String> plainAttrs = starrer.getAttributes();
+            String starred = PathStarrer.computeIfAbsent(path);
+            List<String> plainAttrs = XPathParts.getPathAttributes(path);
             for (String locale : toTest) {
                 CLDRLocale cLoc = CLDRLocale.getInstance(locale);
                 CoverageLevel2 coverageLeveler = CoverageLevel2.getInstance(locale);
@@ -201,27 +197,6 @@ public class ListCoverageLevels {
                 starredToLevels.put(starred, level, cLoc, plainAttrs, Boolean.TRUE);
             }
         }
-
-        //        for (String locale : toTest) {
-        //            if (!ltp.set(locale).getRegion().isEmpty()
-        //                //  || locale.equals("root")
-        //                || locale.equals("ceb")
-        //                || defaultContents.contains(locale)) {
-        //                continue;
-        //            }
-        //            CLDRLocale cLoc = CLDRLocale.getInstance(locale);
-        //            ALL.add(cLoc);
-        //            CoverageLevel2 coverageLeveler = CoverageLevel2.getInstance(locale);
-        //            //Level desiredLevel = sc.getLocaleCoverageLevel(Organization.cldr, locale);
-        //            CLDRFile testFile = mainAndAnnotationsFactory.make(locale, false);
-        //            for (String path : testFile.fullIterable()) {
-        //                Level level = coverageLeveler.getLevel(path);
-        //                String starred = starrer.set(path);
-        //                Attributes attributes = new Attributes(cLoc, starrer.getAttributes());
-        //                data.put(level, starred, attributes, Boolean.TRUE);
-        //                starredToLevels.put(starred, level, cLoc, Boolean.TRUE);
-        //            }
-        //        }
 
         System.out.println("ALL=" + getLocaleName(null, ALL));
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
@@ -6,6 +6,7 @@
  */
 package org.unicode.cldr.util;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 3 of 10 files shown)


btangmu:
This is a draft PR for now, since I have not yet tested these utilities:
- CopySubdivisionsIntoMain
- GenerateEnglishChanged
- ListCoverageLevels
- ShowLocaleCoverage
- ShowStarredCoverage

btangmu:
I ran CopySubdivisionsIntoMain with this branch, and also with the main branch, and confirmed the results were identical.

btangmu:
I've run

`mvn -DCLDR_DIR=$(pwd) --file=tools/pom.xml -pl cldr-code compile -DskipTests=true exec:java -Dexec.mainClass=org.unicode.cldr.tool.GenerateEnglishChanged`

but it fails with:

`org.unicode.cldr.util.SimpleFactory$NoSourceDirectoryException: Unable to determine the source directory for locale en`

I have run CheckoutArchive. I haven't been able figure out what's needed.

btangmu:
I ran ListCoverageLevels with this branch, and also with the main branch, and confirmed the results were identical, except that the paths have `"([^"]*+)"` rather than `"*"`. For example:

```
< basic	3	//ldml/dates/timeZoneNames/metazone[@type="([^"]*+)"]/long/standard
---
> basic	3	//ldml/dates/timeZoneNames/metazone[@type=]/long/standard
```

That's due to:

```
public class PathStarrer {
    public static final String STAR_PATTERN = "([^\"]*+)";
```

Is this a problem? I suspect it isn't.

btangmu:
I ran ShowLocaleCoverage with this branch, and also with the main branch, and confirmed the resulting files `common/properties/coverageLevels.txt` were identical.

btangmu:
I ran ShowStarredCoverage (distinct from ShowLocaleCoverage) with this branch, and also with the main branch, and the resulting output did not match. In order to fix this, I did `replace(PathStarrer.STAR_PATTERN, "*")` for each starred path. Apparently it is assumed somewhere in ShowStarredCoverage, or in code called by ShowStarredCoverage, that the starred paths will have "*" rather than PathStarrer.STAR_PATTERN (which is more complex).

Also, for a long time this unrelated error has appeared whenever recompiling the code:

```
[WARNING] /Users/tbishop/Documents/WenlinDocs/Organizations/Unicode/CLDR_job/cldr/tools/cldr-code/src/main/java/org/unicode/cldr/util/LocaleValidator.java:[162,46] non-varargs call of varargs method with inexact argument type for last parameter;
  cast to java.lang.Object for a varargs call
  cast to java.lang.Object[] for a non-varargs call and to suppress this warning
```

I've fixed that in LocaleValidator.java. I tried both solutions, with and without `[]`. Tests pass with `[]` and fail without it.

btangmu:
I have now tested all these utilities, except for GenerateEnglishChanged:
-  CopySubdivisionsIntoMain 
-  GenerateEnglishChanged 
-  ListCoverageLevels 
-  ShowLocaleCoverage 
-  ShowStarredCoverage

btangmu:
I need advice on how to test GenerateEnglishChanged. Once that is confirmed to run the same as before, this PR should be OK to merge.

srl295:
> I need advice on how to test GenerateEnglishChanged. Once that is confirmed to run the same as before, this PR should be OK to merge.

Does it generate the same binary before and after ?

You should be able to just run that binary class file on Main, and also run it on your own Branch and even if it generates a different file from what's checked in, it should be the same as each other… Anyway, that's the way I would check it

btangmu:
> Does it [GenerateEnglishChanged] generate the same binary before and after ?

It didn't generate anything when I ran it, due to this:

```
org.unicode.cldr.util.SimpleFactory$NoSourceDirectoryException: Unable to determine the source directory for locale en
    at org.unicode.cldr.util.SimpleFactory.handleMake (SimpleFactory.java:561)
    at org.unicode.cldr.util.Factory.make (Factory.java:133)
    at org.unicode.cldr.util.Factory.make (Factory.java:143)
    at org.unicode.cldr.tool.GenerateEnglishChanged.main (GenerateEnglishChanged.java:54)
```

I have run CheckoutArchive. I've tried adding "48.0" to CLDR_VERSIONS, or changing DEV_VERSION from "48" to "47",  but the result is always the same. When is GenerateEnglishChanged ordinarily run? I don't see it in BRS.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListCoverageLevels.java
```diff
@@ -116,7 +112,7 @@ public static void main(String[] args) {
                         coverageLeveler == null
                                 ? Level.COMPREHENSIVE
                                 : coverageLeveler.getLevel(path);
-                String skeleton = starrer.set(path);
+                String skeleton = PathStarrer.computeIfAbsent(path);
```
macchiati: This is a kind of confusing method name (computeIfAbsent); it is exposing the internals of how PathStarrer is working.

macchiati: would be better as "get" or "make".

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1273,4 +1274,19 @@ private XPathParts removeAttribute(String attribute) {
...
+    public static List<String> getPathAttributes(String path) {
+        List<String> attributes = new ArrayList<>();
+        XPathParts parts = getFrozenInstance(path);
+        for (int i = 0; i < parts.size(); ++i) {
+            for (String key : parts.getAttributeKeys(i)) {
+                attributes.add(parts.getAttributeValue(i, key));
+            }
+        }
+        return attributes;
+    }
```
macchiati: Better as:

```suggestion
    public List<String> getPathAttributes() {
        List<String> attributes = new ArrayList<>();
        for (int i = 0; i < parts.size(); ++i) {
            for (String key : parts.getAttributeKeys(i)) {
                attributes.add(parts.getAttributeValue(i, key));
            }
        }
        return attributes;
    }
    
        public static List<String> getPathAttributes(String path) {
           return getFrozenInstance(path).getPathAttributes();
           }
           // and then refactor away the static
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListCoverageLevels.java
btangmu: Shall I change it to `get`?

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1273,4 +1274,19 @@ private XPathParts removeAttribute(String attribute) {
         }
         return this;
     }
+
+    public static String getPathAttributesJoined(String path, String separator) {
+        return Joiner.on(separator).join(getPathAttributes(path));
+    }
```
macchiati: It sounds like this has only 2 uses. Better in that case to refactor away.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ListCoverageLevels.java
btangmu: Any idea what's going on with `GenerateEnglishChanged`?

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
btangmu: OK, I'll make that change and the others you proposed

btangmu:
@macchiati The 8th commit implements the changes you suggested

btangmu:
The changes to GenerateEnglishChanged.java are few, and don't actually change the way PathStarrer is used, other than using XPathParts instead of PathStarrer to get the original attributes. GenerateEnglishChanged still uses the old non-thread-safe PathStarrer.set rather than the new thread-safe PathStarrer.get.

Still, it would be reassuring to be able to test GenerateEnglishChanged. How can it be tested?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CopySubdivisionsIntoMain.java
```diff
@@ -132,8 +134,11 @@ private static void doAfter() {
 
                 String oldValue = subdivisionFile.getStringValue(path);
                 if (!Objects.equal(oldValue, value)) {
-                    pathStarrer.set(path);
-                    String firstAttributeValue = pathStarrer.getAttributes().iterator().next();
+                    String firstAttributeValue =
+                            XPathParts.getFrozenInstance(path)
+                                    .getPathAttributes()
```
macchiati: FYI, I just recently found out about IterableUtils.first(myCollection); (from Guava)

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1273,4 +1262,14 @@ private XPathParts removeAttribute(String attribute) {
         }
         return this;
     }
+
+    public List<String> getPathAttributes() {
```
macchiati: Minor, and we could do later.

We should call this getPathAttributeValues, because it is the attribute **values** that we are getting, not the attributes.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
```diff
@@ -1273,4 +1262,14 @@ private XPathParts removeAttribute(String attribute) {
...
+        List<String> attributes = new ArrayList<>();
+        for (int i = 0; i < size(); ++i) {
+            for (String key : getAttributeKeys(i)) {
+                attributes.add(getAttributeValue(i, key));
+            }
```
macchiati: I think could be slightly simpler/faster as
```suggestion
            attributes.addAll(getAttributes().values());
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CopySubdivisionsIntoMain.java
btangmu: I noticed it was strange to use an iterator here to get only the first value...

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
btangmu: Or just getAttributeValues now that it's an instance method -- there's already a getAttributeValue (singular) method

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
btangmu: I'll do that

btangmu: that is, `attributes.addAll(getAttributes(i).values());`

macchiati:
> Still, it would be reassuring to be able to test GenerateEnglishChanged. How can it be tested?

I debugged it, and there was a bug.

The following fails because it is searching for ...47.0.0 in cldr-archive
>         addStandardSubdirectories =
                 CLDR_CONFIG.addStandardSubdirectories(
                         CLDR_CONFIG.fileArrayFromStringArray(RELEASE_DIRECTORY, base));

The fix is to change RELEASE_DIRECTORY to drop the extra ".0"

>     private static final File RELEASE_DIRECTORY =
            new File(
                    CLDRPaths.ARCHIVE_DIRECTORY
                            + "cldr-"
                            + ToolConstants.LAST_RELEASE_VERSION
                            + "/");

To compare, run the file before your change, capturing the console. Then run after and diff.

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
btangmu: Done in the 9th commit

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
macchiati: Yes, even better!

tools/cldr-code/src/main/java/org/unicode/cldr/util/XPathParts.java
macchiati: Yup. thanks

btangmu:
```
$ git diff
diff --git a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateEnglishChanged.java b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateEnglishChanged.java
index 7cc409f9f7..5ab222bc53 100644
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateEnglishChanged.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateEnglishChanged.java
@@ -32,7 +32,6 @@ public class GenerateEnglishChanged {
                     CLDRPaths.ARCHIVE_DIRECTORY
                             + "cldr-"
                             + ToolConstants.LAST_RELEASE_VERSION
-                            + ".0"
                             + "/");
     private static final boolean TRIAL = false; 

$ diff -s ../tmp-GenerateEnglishChanged-main.txt ../tmp-GenerateEnglishChanged-branch.txt
Files ../tmp-GenerateEnglishChanged-main.txt and ../tmp-GenerateEnglishChanged-branch.txt are identical
```

btangmu:
The 10th commit fixed GenerateEnglishChanged. Confirmed output is the same as with main branch (if main is fixed in the same way).

PR #4864, opened 2025-07-09
CLDR-14876 on vote for removal, update TestCache

srl295:
- TestCache will get notified if a value goes from null to non-null or vice versa (i.e. removal or non-removal)

CLDR-14876

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
@@ -889,7 +889,10 @@ public class STFactory extends Factory implements BallotBoxFactory<UserRegistry.
 
             String newVal = dataBackedSource.getValueAtDPath(distinguishingXpath);
             String newFullPath = dataBackedSource.getFullPathAtDPath(distinguishingXpath);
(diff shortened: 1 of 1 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/STFactory.java
```diff
@@ -889,7 +889,10 @@ public void voteForValueWithType(
 
             String newVal = dataBackedSource.getValueAtDPath(distinguishingXpath);
             String newFullPath = dataBackedSource.getFullPathAtDPath(distinguishingXpath);
-            if (newVal != null && (!newVal.equals(oldVal) || !oldFullPath.equals(newFullPath))) {
+            if ((newVal == null && oldVal != null)
+                    || (oldVal == null && newVal != null)
```
btangmu, line 893: (newVal == null && oldVal != null) || (oldVal == null && newVal != null)

could be shortened slightly to

((newVal == null) != (oldVal == null))

but I'm not sure if that would make it easier or harder to understand

srl295: I thought about that. I also think there's at least one function that compares strings including null strings.  But this is at least explicit about what it's checking for.

PR #4866, opened 2025-07-09
CLDR-18745 add validate_currencies.py

preetsojitra2712:
CLDR-18745

Adds `validate_currencies.py` under `tools/scripts/llm/` that:

1. Downloads the CLDR `currencies.json` for a given locale.
2. Flattens each currency entry into lines of the form  
   `CODE: one='…', other='…', symbol='…', alt_narrow='…'`.
3. Splits the lines into small chunks to stay under model context limits.
4. Runs two LLM-powered sanity checks per chunk:  
   - Correctness of display names/symbols  
   - Missing symbols detection  
5. Counts entries locally and emits a single JSON report.

### DIFF
--- a/tools/scripts/llm/validate_currencies.py
+++ b/tools/scripts/llm/validate_currencies.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+"""
+validate_currencies.py
+
+Run AI‐powered sanity checks on CLDR currency data and emit a JSON report.
+
+Usage:
+  python validate_currencies.py [locale] [output.json]
+
+Defaults:
+  locale      = en-001
+  output.json = currencies_report.json
+"""
+
+import os
+import json
+import requests
+import argparse
+from transformers import pipeline
+
+def fetch_currencies(locale: str) -> dict:
+    """Download the CLDR currencies.json for a given locale."""
+    url = (
+        "https://raw.githubusercontent.com/unicode-org/cldr-json"
+        "/main/cldr-json/cldr-numbers-full/main/"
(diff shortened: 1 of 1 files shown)


preetsojitra2712:
Hello @younies , please review this PR adding `validate_currencies.py` under `tools/scripts/llm/`, which implements LLM-powered sanity checks and JSON reporting for CLDR currency data.

tools/scripts/llm/validate_currencies.py
```diff
@@ -0,0 +1,126 @@
...
+    resp = requests.get(url)
+    resp.raise_for_status()
+    return resp.json()
+
+def flatten(cdata: dict, locale: str) -> list[str]:
+    """Turn the nested currencies object into lines of text."""
+    cur = cdata["main"][locale]["numbers"]["currencies"]
+    out = []
+    for code, info in cur.items():
+        one    = info.get("displayName-count-one", "")
```
younies, line 37: this is good for english, but for other locales, there are more than one count.
it is better to fetch all "displayName-count-.*" in a list

younies (approved):

tools/scripts/llm/validate_currencies.py
younies: Add this in a following PR

preetsojitra2712: Thanks for the suggestion! . I’ll update the flatten() function in the next PR to collect all displayName-count-* entries into a list instead of just “one” and “other.” I’ll also include an example sample JSON so you can verify it – stay tuned!

PR #4869, opened 2025-07-11
CLDR-18745 add sample_currencies_report.json for currency validator

preetsojitra2712:
CLDR-18745

Adds `validate_currencies.py` under `tools/scripts/llm/` that:

1. Fetches the CLDR `currencies.json` for a given locale (default `en-001`).
2. Flattens each currency’s `displayName-count-*`, `symbol` and `symbol-alt-narrow` into simple lines.
3. Splits the lines into 5-item chunks so we stay under the model’s 512-token limit.
4. Invokes two LLM checks (“Check correctness” and “Find missing symbols”).
5. Emits a JSON report containing each check’s results and a local count of entries.

Also adds `examples/sample_currencies_report.json` as a concrete example of the output format.


preetsojitra2712:
@younies  please review this sample output.

younies (approved):

PR #4870, opened 2025-07-11
CLDR-18745 update validate_currencies.py

preetsojitra2712:
CLDR-18745

- Updates `validate_currencies.py` to collect *all* `displayName-count-*` fields
- Breaks each locale’s currency lines into 5-item chunks for the LLM
- Runs “Check correctness” and “Find missing symbols” prompts
- Emits a JSON report with the AI results and a local count


preetsojitra2712:
Thanks for the feedback, @younies   I’ll follow up with a PR that updates `flatten()` to collect **all** `displayName-count-*` entries into a list and include a sample JSON output for you to review.

preetsojitra2712:
Hello @younies ,

I got this error in our Docker/WebDriver CI job:

```
unknown error: net::ERR_CONNECTION_REFUSED
```

It’s happening because the CLDR app never starts before Selenium runs, so the browser can’t connect. Additionally, our `setup-config.sh` has a typo (`Usage::`), is missing `-f` when removing the jar, and has a bad `expr` call.

younies (approved):

PR #4872, opened 2025-07-14
CLDR-18610 VettingParticipation add EMP and MP; revise UI

btangmu:
-New columns EMP and MP for Errors, Missing, and Provisional counts, based on organization-specific coverage level for locale

-Add tooltip descriptions (comments) for table column headers, matching those in spreadsheet

-Revise column headers: change Targ. Cover. to Coverage, change Vetter# to User#

-Make table header sticky so it remains visible when scroll down to see lowest rows

-On the back end, revise ReviewOutput and ParticipationResults to include errors/missing/provisional counts from LocaleCompletionData; the data was already computed but not included in the http response

-Revise LocaleCompletion Java comment and API description to clarify what was already true, that LocaleCompletion API uses the target coverage level, which is not specific to any organization (and therefore is unsuitable for the vetting participation table)

CLDR-18610

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrVettingParticipation.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrVettingParticipation.mjs
@@ -32,19 +32,15 @@ const NORMAL_RETRY = 1 * SECONDS_IN_MS;
 const COLUMN_TITLE_ORG = "Org";
 const COLUMN_TITLE_LOCALE_NAME = "Locale";
 const COLUMN_TITLE_LOCALE_ID = "Code";
-const COLUMN_TITLE_COVERAGE_LEVEL = "Targ. Cover.";
+const COLUMN_TITLE_COVERAGE_LEVEL = "Coverage";
 const COLUMN_TITLE_LEVEL = "Level";
 const COLUMN_TITLE_PROGRESS_PERCENT = "Done";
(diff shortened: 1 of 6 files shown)


btangmu:
@macchiati This almost finishes the goals in the ticket, except:

- drop "guest" -- straightforward; I will follow up with that
- drop "* in Code (except for Orgs that have just one language. In that case, reset the Code to that language)" -- maybe should be a separate ticket?
- drop "A language in Code that is not in Locales.txt for that Org" -- maybe should be a separate ticket?

The last two "drops", if I understand them, might belong in a separate ticket. I think we've discussed disallowing locales=* for vetters, and removing any connection between a locale and a vetter whose org doesn't have that locale. Those changes would be more general than just filtering the Vetting Participation table. If we clean-up/filter the cldr_users db table, that will automatically keep inappropriate data out of Vetting Participation and other reporting.

btangmu:
Actually looking through the comments there are 2 more TODO items:

- After finishing, print the time it took to wait, and the time it took to finish (after the wait ended)
- Add a link on Name or Vetter# to the individual recent activity (https://cldr-smoke.unicode.org/cldr-apps/v#recent_activity///1067).

macchiati (approved):

PR #4876, opened 2025-07-15
CLDR-18195 Import Exemplars from SIL

emily-roth:
CLDR-18195

ALLOW_MANY_COMMITS=true


DavidLRowe:
Not sure about `con_Latn.xml`. Eventually this will be `con.xml`. (Only because "con.xml" is an illegal name under Windows 10, did we use "con_Latn.xml". That restriction is gone for Windows 11.) 

We could either

- change the name from `con_Latn.xml` to `con.xml` (which might cause problems for anyone dealing with this data on a Windows 10 platform)
- omit the file completely from this PR (and pick it up later when the name is updated)

@macchiati Any thoughts, Mark?

macchiati:
I looked through the build&test log to find the problems, that is:

https://github.com/unicode-org/cldr/actions/runs/16300797065/job/46034594976?pr=4876

Here are some examples, and what to do:

```
  TestCLDRFile {
    TestConstructedValue (0.001s) Passed
    TestExtraPaths13954 (0.179s) Passed
    TestFileIds
Error:  (TestCLDRFile.java:778)  Error: Can only one have 1 instance of bqi in main, but have in [common, exemplars]
```

You have to remove all of the files that are already present in common/main. There appear to be just 4 of these (searching for "Can only one have 1 instance of").

```
Error:  (TestCLDRFile.java:808)  Error: Missing parent (aii) for aii_Syrc  in exemplars/main; likely=aii
```

Whenever we have xxx_yyy, we also have to have xxx alone. Typically that is done by removing the script from the locale that is the default. For example, if we look at

https://www.unicode.org/cldr/charts/48/supplemental/likely_subtags.html#aii

We find that the target ID is aii_Syrc_IQ, so your aii_Syrc.xml should be aii.xml
(This can be done programmatically by accessing the likelySubtags.xml file, or manually with that chart.)

There are appear to be  4 of these (searching for "Missing parent")

```
Error:  (TestCLDRFile.java:821)  Error: Missing simple parent (con) for con_Latn  in exemplars/main; likely=null
```
The same goes for this one. I think the 'simple' is added if there is only one script, so the process is simpler; just drop the script (eg "_Latn")

There appear to be 22 of these (searching for "Missing simple parent").


----
```
  TestIdentity {
    TestIdentityVsFilename
Error:  (TestIdentity.java:104)  Error: Locale "fuc_Arab" uses a non-canonical language tag: fuc
	Located in : /home/runner/work/cldr/cldr/exemplars/main
	Canonical form would be : ff
```
the fuc needs to change to ff.

```
Error:  (TestIdentity.java:104)  Error: Locale "ttq_Tfng" uses a non-canonical language tag: ttq
	Located in : /home/runner/work/cldr/cldr/exemplars/main
	Canonical form would be : tmh
```
the ttq needs to change to tmh

```
  TestPaths {
    TestNonLdml
Error:  (TestPaths.java:629)  Error: main/ttq_Tfng.xml/language@type, expected match to: ⟪validity/locale-for-names⟫ actual: «ttq»

Error:  (TestPaths.java:353)  Error: Deprecated attribute in data: ldml:orientation:characters 	;/home/runner/work/cldr/cldr/exemplars/main/mts.xml

Error:  (TestPaths.java:629)  Error: main/fuc_Arab.xml/language@type, expected match to: ⟪validity/locale-for-names⟫ actual: «fuc»

Error:  (TestPaths.java:353)  Error: Deprecated attribute in data: ldml:orientation:characters 	;/home/runner/work/cldr/cldr/exemplars/main/les.xml
```

This will be fixed by the same 2 changes above.

macchiati:
There is a failure with the ticket. I went ahead and accepted it (you'll see a comment), but the commit message needs fixing.

All of your commit messages have to start with "CLDR-18195 " (and the space is important). (I know that's awkward, but we do it to make sure that we have a connection to the ticket.) The way to fix this is to squash; David can help with that — you can do it as a final step once the test failures are handled.

emily-roth:
Okay, I think I fixed all of the ones that were non-minimal tags and removed the ones that had overlap with the CLDR. 
For the remaining issues with tags, the ones labeled "missing simple parent", those are already the minimal tag. For example, `bji_Ethi` isn't `bji`, but rather, `bji` is `bji_Latn`. It would be misleading to label the file with ethiopic data as `bji`

@macchiati Does this mean we need to also include empty files that just have an identity element for the shorter tag as well? We don't have any data on `bji` using Latn script in the SLDR. I see an empty file in the exemplars folder called `ife_TG`, is this why that file exists?

macchiati:
Yes, you should only remove the script IF it is the default script; good catch.

We are not set up to have a locale with a language and script where we don't have that locale without a script. So there are two possibilities:

1. The script **is actually** the likely script for the language, but the likely subtags are wrong. In that case we need to correct the likely subtags, OR
2. The script **is not** the likely script for the language, and we don't have data for the language with the likely subscript. In that case we need to remove the locale until we have data for the language with the likely subscript. (We can always add it later once we have the locale for the language with the likely subscript

Locales with no parent but root should never have empty main exemplars; I'll file a ticket for fixing that.

emily-roth:
Alright. I've updated the script I use to prepare SLDR files for exemplars/main, telling it to skip the files missing a non-root parent in either common/main or exemplars/main, unless said parent is also part of the files being imported. In the process, I also caught a few files that shouldn't have been in this batch but managed to sneak in through a loophole in that script. 🤦‍♀️ Assuming I caught everything, this commit should get rid of the remaining "missing parent" errors.

macchiati:
**In the home stretch!** 
Going to the very end of the Files tab, you see just 2 errors remaining!

 Check failure on line 353 in tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java


GitHub Actions / build
 (TestPaths.java:353)  Error: Deprecated attribute in data: ldml:orientation:characters 	;/home/runner/work/cldr/cldr/exemplars/main/les.xml


GitHub Actions / build
 (TestPaths.java:353)  Error: Deprecated attribute in data: ldml:orientation:characters 	;/home/runner/work/cldr/cldr/exemplars/main/mts.xml
 Check failure on line 353 in tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestPaths.java

Here is the problem in les.xml (probably similar in mts)
```
	<layout>
		<orientation characters="left-to-right" lines="top-to-bottom"/>
	</layout>
```

Instead it should be the following:
```
	<layout>
		<orientation>
			<lineOrder>top-to-bottom</lineOrder>
		</orientation>
	</layout>
```

You don't need characterOrder, since left-to-right is the default. For example, Arabic is:
```
	<layout>
		<orientation>
			<characterOrder>right-to-left</characterOrder>
		</orientation>
	</layout>
```

emily-roth:
Hurray, it finally builds! This has given me some good ideas for things we can add to our own tests at SIL for the data we collect, so that next time, those errors would have been caught long before they arrived in the exemplars folder. What about the Docker test? Why is that failing @macchiati?

macchiati (approved):

macchiati:
Docker is a flaky test, and it is not required to pass. If you confirm that it is "Ready for Review" (hit the button) I can merge it.

Sounds great to make upstream changes for next time!

emily-roth:
Thanks for your help, @macchiati!

macchiati:
Thanks for all your work on this!

PR #4877, opened 2025-07-16
CLDR-18697 New PathStarrer.getWithPattern; make STAR_PATTERN private

btangmu:
-New getWithPattern gets default starred path, then replaces STAR_PATTERN with given pattern

-Use getWithPattern in GenerateEnglishChanged and ShowStarredCoverage

-Assume STAR_PATTERN would never occur (literally, not as regex) in an unstarred path

-Make STAR_PATTERN private for encapsulation, leaving open options for refactoring

-Fix compiler warnings in GenerateEnglishChanged, mostly automatic refactoring

-Comments

CLDR-18697

ALLOW_MANY_COMMITS=true


btangmu:
I confirmed that GenerateEnglishChanged produces same output as before

btangmu:
The next steps would be to replace remaining `starrer.set()` with `PathStarrer.get()` or `PathStarrer.getWithPattern()`, and to get the attributes, where needed, from XPathParts instead of PathStarrer.

macchiati (approved):

srl295 (approved):

PR #4878, opened 2025-07-16
CLDR-18798 bad locales: call out bad locales in user list

srl295:
- we are already normalizing away 'bad' (invalid or missing locales) from the user list
- this code _preserves_ anything that was normalized away and presents it in the user list
- error can be corrected by calling 'set locales' and changing the locales, as the 'set locale' function will overwrite the locales list and does not allow invalid or missing locales to be set.

CLDR-18798

ALLOW_MANY_COMMITS=true

image

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
@@ -451,6 +451,7 @@ public class UserList {
         boolean havePermToChange = me.isAdminFor(user);
         boolean userCanDeleteUser = UserRegistry.userCanDeleteUser(me, user.id, user.userlevel);
         VoteResolver.Level level = VoteResolver.Level.fromSTLevel(user.userlevel);
+        // TODO: why doesn't this use u.toJSONObject()?
         shownUsers.put(
                 new JSONObject()
                         .put("actions", u.ua)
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
```diff
@@ -451,6 +451,7 @@ private void putShownUser(JSONArray shownUsers, UserSettings u) throws JSONExcep
         boolean havePermToChange = me.isAdminFor(user);
         boolean userCanDeleteUser = UserRegistry.userCanDeleteUser(me, user.id, user.userlevel);
         VoteResolver.Level level = VoteResolver.Level.fromSTLevel(user.userlevel);
+        // TODO: why doesn't this use u.toJSONObject()?
```
btangmu, line 454: "TODO" comment should include a ticket number

btangmu (approved):
Great improvements! Ideally, locales would be private, accessed only with getLocales, setLocales. Minor quibble: TODO comment should reference a ticket

tools/cldr-apps/src/main/java/org/unicode/cldr/web/UserList.java
srl295: good. i'll update this in a followon

macchiati:
I ran across two other locale codes that should not occur — nobody should be able to vote in them.

Norwegian Bokmål	nb → should be no
Portuguese (Brazil)	pt_BR → should be pt

PR #4888, opened 2025-07-22
CLDR-18850 v48 vxml merge to main

srl295:
CLDR-18850

Instructions:
- to fix, make PRs against the branch `brs/v48vxml`
- Use my ticket CLDR-18850 for every commit, please.

ALLOW_MANY_COMMITS=true


macchiati:
The branch should no longer have any test failures, so you should be able to merge it now. Cf @AEApple

macchiati (approved):

macchiati:
🎉🎉🎉🎉

common/main/sl.xml
```diff
@@ -5866,11 +6108,11 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					<pattern type="100000" count="other">000 tisoč</pattern>
 					<pattern type="1000000" count="one">0 milijon</pattern>
 					<pattern type="1000000" count="two">0 milijona</pattern>
-					<pattern type="1000000" count="few">0 milijone</pattern>
+					<pattern type="1000000" count="few">0 milijoni</pattern>
 					<pattern type="1000000" count="other">0 milijonov</pattern>
 					<pattern type="10000000" count="one">00 milijon</pattern>
 					<pattern type="10000000" count="two">00 milijona</pattern>
-					<pattern type="10000000" count="few">00 milijoni</pattern>
+					<pattern type="10000000" count="few">00 milijona</pattern>
```
naotoj, line 6115: This one seems suspicious. Should this be modified to "milijona"?

AEApple: One of the vetters requested the change to 00 milijona and two other vetters agreed with the reasoning of:

8 digits -few should in practice never be used for whole numbers since 11.000.000, 12.000.000, 13.000.000 and 15.000.000 all fall within -other, but at least would cover the case like 12.500.000 as 12,5 milijona. Same logic could perhaps be applied to 8-digits-long-one and -two

naotoj: Thanks for the clarification of the background!

PR #4889, opened 2025-07-22
CLDR-18850 Fix en, ckb, pap, etc.

macchiati:
CLDR-18850

Fix en, ckb, pap, etc.

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4891, opened 2025-07-23
CLDR-18850 v48 BRS vxml: logknownissue for unit tests

srl295:
CLDR-18850

- move one issues to logKnownIssues

add logknownissue for Gujr transform to CLDR-18852 for U+0AF0
the workaround was in https://github.com/unicode-org/cldr/pull/4889 but was subsumed under #4888

ALLOW_MANY_COMMITS=true


macchiati:
Can you rebase? I merged the other PR, and don't see any remaining test failures.

srl295:
The only remaining quesiton here is whether the suppressScriptHack should be log known issue. Closing this otherwise.

macchiati:
I think having both of those be logKnownIssues would be good.

PR #4892, opened 2025-07-23
CLDR-18745 Create pipeline_compare_units.py

preetsojitra2712:
CLDR-18745

This script begins by parsing the CLDR `units.xml` file to pull out locale-specific preferences for two measurements—person height and road distance—into JSON maps keyed by ISO country codes. It then constructs neutral, user-facing prompts asking GPT-4 which units are commonly used for those measurements in each country, captures the model’s bullet-list responses, and normalizes them into corresponding JSON files.

Next, rather than hand-coding diff logic, it wraps both the CLDR-derived and LLM-predicted JSON objects into a single “compare these for me” GPT-4 prompt. The model returns a clear, human-readable report listing each country, indicating whether the units match, and explaining any discrepancies. The end result is an automated pipeline that extracts, predicts, and cross-validates locale unit preferences with minimal manual effort.

ALLOW_MANY_COMMITS=true


preetsojitra2712:
hello @younies , please review this: I’ve added `pipeline_compare_units.py`, which extracts person-height and road-distance preferences from `units.xml`, prompts GPT-4 (no CLDR mention) for predictions, parses and normalizes both outputs, and generates human-readable comparison reports.

PR #4894, opened 2025-07-23
CLDR-18430 Update coverageLevels post v48 submissions

conradarcturus:
Now that all of the v48 submissions are in XML, we can see the changes to coverage levels.

CLDR-18430

ALLOW_MANY_COMMITS=true


AEApple:
Can we set sr_Latn and yue_Hans back to match sr and yue respectively since they are algorithmically generated?

PR #4895, opened 2025-07-24
CLDR-18462 BRS remove apple core emoji

markusicu:
CLDR-18462

- [184-C26] Consensus: U+1FADD APPLE CORE, approved for encoding in Unicode Version 17.0 (ref. Consensus 181-C5), is removed from Unicode Version 17.0 and is targeted instead for Unicode Version 18.0.
- [184-A64] Action Item for Markus Scherer, CLDR: Update data files for CLDR 48, removing U+1FADD APPLE CORE [ref. Consensus 184-C26].

ALLOW_MANY_COMMITS=true


AEApple (approved):
Thank you for doing this!

I'd like to understand the changes at the end of the file, although I'm assumed they are done automatically by the emoji related tooling.

markusicu:
> I'd like to understand the changes at the end of the file, although I'm assumed they are done automatically by the emoji related tooling.

The end of which file? emoj-test.txt? I hand-edited that one for now. At first I thought that I just needed to remove this one character from the annotations, but then I found a few more files. And I didn't want to wait until @nedley has time to turn the upstream emoji data files crank.

Once we have near-final Unicode 17 data files, I can update emoji-test.txt again. That will then also update the common/uca/ files. I wonder why a [GitHub search for 1FADD](https://github.com/search?q=repo%3Aunicode-org%2Fcldr%201FADD&type=code) didn't find those, nor emoji-test.txt.

PR #4897, opened 2025-07-28
CLDR-18283 remove Chisoi script

markusicu:
CLDR-18283

Mostly removing the Chisoi script and its associated data.
The Script Metadata import also fixes a data bug, making Gunjala Gondi an Excluded script [as UAX31 says](https://www.unicode.org/reports/tr31/tr31-42.html#Table_Candidate_Characters_for_Exclusion_from_Identifiers).

From the UTC-184 meeting minutes:

- [[184-C2](https://www.unicode.org/cgi-bin/GetL2Ref.pl?184-C2)] Consensus: Of characters approved for encoding in Unicode 17.0 (ref. 181-C60, 181-C61), the following 43 characters are removed from Unicode 17.0 and are targeted instead for Unicode 18.0:
  - 09FF BENGALI LETTER SANSKRIT BA
  - 0B53 ORIYA SIGN DOT ABOVE and 0B54 ORIYA SIGN DOUBLE DOT ABOVE
  - 40 Chisoi script characters at 16D80..16DA9 and the Chisoi block at 16D80..16DAF.
- [[184-A75](https://www.unicode.org/cgi-bin/GetL2Ref.pl?184-A75)] Action Item for Markus Scherer, PAG: Work with CLDR-TC to align the CLDR script metadata UAX31-derived “ID Usage” with a contemporary version of UAX31, especially for Gunjala Gondi. See [L2/25-183](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-183) item 6.4.

ALLOW_MANY_COMMITS=true


markusicu:
FYI:

I was a bit surprised by how much data I had to remove... especially, a few languages have data for things like `<symbols numberSystem="chis">` which look like they might be unnecessary, especially when all of their data values are `↑↑↑`. "ba" actually overides the grouping separator, but I don't know if that's real. How much do Bashkir speakers care about the grouping separator when writing numbers with Chisoi-script digits?

AEApple:
> FYI:
> 
> I was a bit surprised by how much data I had to remove... especially, a few languages have data for things like `<symbols numberSystem="chis">` which look like they might be unnecessary, especially when all of their data values are `↑↑↑`. "ba" actually overides the grouping separator, but I don't know if that's real. How much do Bashkir speakers care about the grouping separator when writing numbers with Chisoi-script digits?

Mark and I have both commented that we may need to put a harder limit on allowing people to contribute values in comprehensive. Several DDL communities are very enthusiastic and may not realize when there is diminishing returns from contributing data.

cc @conradarcturus @srl295 @DavidLRowe

AEApple (approved):

PR #4900, opened 2025-07-29
CLDR-18861 Migrate Adding/Fixing Plural and Ordinals page 

AEApple:
CLDR-18861

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/updating-codes/update-plurals.md
+++ b/docs/site/development/updating-codes/update-plurals.md
@@ -1,0 +1,31 @@
+---
+title: Adding/Fixing Plural and Ordinals
+---
+
+# Adding/Fixing Plural and Ordinals
+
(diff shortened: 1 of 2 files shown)


DavidLRowe:
I'm not familiar enough with the tooling to comment on that section.

In line 27:
```
 - Diff the plurals chart against the old one: eg v.s. https://unicode-org.github.io/cldr-staging/charts/38/supplemental/language_plural_rules.html
```
Is "v.s." necessary?
Is "/38/" the number of the previous version? Is it worth adding "(where "38" is the version number of the old chart)"

AEApple:
This was just to migrate the page. Someone should clean it up in a different PR. Can someone please approve so that I can merge this PR?

PR #4901, opened 2025-07-29
Fix typo on common/supplemental/units.xml:506

SpikyC:
Change `usage="visiblty"` to `usage="visibility"`.

ALLOW_MANY_COMMITS=true


AEApple:
This actually isn't a typo, and was reported in https://unicode-org.atlassian.net/browse/CLDR-18496 with what the proposed change would be.

SpikyC:
> This actually isn't a typo, and was reported in https://unicode-org.atlassian.net/browse/CLDR-18496 with what the proposed change would be.

Thanks Annemarie!

PR #4902, opened 2025-07-29
CLDR-18209 spec: split Modifications section into Appendix A

srl295:
CLDR-18209

- regen ToC on numbers
- split Modifications from part 1 to its own Appendix A
- update fix-tocs so that the **Parts** section from part 1, `tr35.md` is copied over to all the other parts.
- re-run fix-tocs to copy that Parts section over.

ALLOW_MANY_COMMITS=true


docs/ldml/tr35-collation.md
```diff
@@ -31,7 +31,7 @@ For the latest version of the Unicode Standard see [[Unicode](https://www.unicod
 For more information see [About Unicode Technical Reports](https://www.unicode.org/reports/about-reports.html) and the [Specifications FAQ](https://www.unicode.org/faq/specifications.html).
 Unicode Technical Reports are governed by the Unicode [Terms of Use](https://www.unicode.org/copyright.html)._
 
-## <a name="Parts" href="#Parts">Parts</a>
+## Parts
```
srl295, line 34: this whole section is now copied from tr35.md to every part, automatically

docs/ldml/tr35-modifications.md
```diff
@@ -0,0 +1,83 @@
...
+|-------|----------|
+|Editors|Steven R. Loomis (<a href="mailto:srloomis@unicode.org">srloomis@unicode.org</a>) and <a href="tr35.md#Acknowledgments">other CLDR committee members|
+
+For the full header, summary, and status, see [Part 1: Core](tr35.md).
+
+### _Summary_
+
+This is a partial document, describing only the changes to the LDML since the previous release. For the other parts of the LDML see the [main LDML document](tr35.md) and the links above.
+
+### _Status_
```
srl295, line 15: is this status area needed for an appendix?

AEApple: I think we need it for modifications, but I don't think we need it for the acknowledgements. @macchiati - WDYT?

macchiati: We don't need them in modification or acks.

I actually don't think we need the Status for _any_ of the Parts except the main part. However, since we are talking about making some of the Parts be separate documents, we should leave them in  for now.

docs/ldml/tr35-collation.md
macchiati: Note: this changes the href to parts instead of Parts. 

Do you automatically regenerate redirects when these change?

srl295: I thought I had… well, it does now (in this PR)

docs/ldml/tr35-modifications.md
srl295: OK, i will leave it as is then.

srl295:
@AEApple I'd like to merge this before I do the error-without-modification…

PR #4906, opened 2025-07-30
CLDR-18492 BRS currency update

srl295:
CLDR-18492

- update latest data from SIX
- amendments include 179 (XAD) and 178 (CUC)
- fixed a problem with CountItems tool where it would silently succeed if a -Dmethod was not chosen

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -355,7 +355,7 @@ The printed version of ISO-4217:2001
         </region>
         <region iso3166="CU">
             <currency iso4217="CUP" from="1859-01-01"/>
-            <currency iso4217="CUC" from="1994-01-01" to="2021-01-01"/>
+            <currency iso4217="CUC" from="1994-01-01" to="2021-06-01"/>
             <!-- currency iso4217="USD" from="1994"/ --><!-- Unofficially used -->
             <currency iso4217="USD" from="1899-01-01" to="1959-01-01"/>
         </region>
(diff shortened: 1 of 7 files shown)


common/supplemental/supplementalData.xml
```diff
@@ -78,6 +78,7 @@ For terms of use, see https://www.unicode.org/copyright.html
             <info iso4217="VND" digits="0" rounding="0"/>
             <info iso4217="VUV" digits="0" rounding="0"/>
             <info iso4217="XAF" digits="0" rounding="0"/>
+			<info iso4217="XAU" digits="2" rounding="0"/>
```
btangmu, line 81: tabs versus spaces? Do we have any normalization for xml?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CountItems.java
```diff
@@ -125,9 +126,11 @@ public static void main(String[] args) throws Exception {
         try {
             String methodName = System.getProperty("method");
             if (methodName != null) {
+                System.err.println("Calling -Dmethod=" + methodName);
```
btangmu, line 129: `System.err` -- is this an error?

btangmu (approved):
rslgtm

common/supplemental/supplementalData.xml
srl295: probably should have a normalizer for it.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CountItems.java
srl295: some of the tools output to stdout, so i'm outputting to stderr so if someone is redirecting it they won't get different output.

PR #4907, opened 2025-07-30
CLDR-18531 Add Hawaii metazone

robertbastian:
CLDR-18531

ALLOW_MANY_COMMITS=true


AEApple:
I thought you were adding Aleutian as well, or are we just expecting it to not have a metazone name going forward and/or we're going to use the existing metazone but only for Aleutian?

robertbastian:
The old metazone stays and Adak continues using it.

PR #4909, opened 2025-07-30
CLDR-18862 Update th abbreviated days format to match common usage

AEApple:
CLDR-18862 Update th abbreviated days format to match common usage

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/th.xml
+++ b/common/main/th.xml
@@ -2711,13 +2711,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<days>
 					<dayContext type="format">
 						<dayWidth type="abbreviated">
-							<day type="sun">อา.</day>
-							<day type="mon">จ.</day>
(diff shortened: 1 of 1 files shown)


AEApple:
@macchiati @pedberg-icu - Can you confirm I did the inheritance updates correctly? I checked root here: https://github.com/unicode-org/cldr/blob/d62c0a9ab8d67227b2267ebe64196100a986ecf3/common/main/root.xml#L1227

I kept the current short values since the linguists had not proposed changing them.

macchiati:
The safest course is to put in the explicit values; the software will take care of putting in ^^^ when we do the CLDRModify pass. So you could use that to get the narrows right, no matter what the inheritance.

AEApple:
Thanks! I made the abbreviated and short values explicit but didn't touch the narrow or wide values. Please review when you have a chance.

macchiati (approved):

PR #4921, opened 2025-08-04
CLDR-18608 Announce Japanese era removal

robertbastian:
CLDR-18608

ALLOW_MANY_COMMITS=true


docs/site/downloads/cldr-48.md
```diff
@@ -148,7 +148,7 @@ The following changes are planned for CLDR 48. Please plan accordingly to avoid
 
 ### V49 advance warnings
 
-- TBD
+- Pre-Meiji Japanese eras have been found to be incorrect and will be removed in a future release. Please file a ticket if you use this data. [CLDR-11400]
```
macchiati:
```suggestion
- There is too much uncertainty in the exact values for pre-Meiji Japanese eras,
and there is feedback that the general practice for exact dates is to use Gregorian for pre-Meiji dates.
These are slated for removal in a future release.
Please file a ticket if you use this data. [CLDR-11400]
```

macchiati: I think we want a bit more nuanced; the previous statement would elicit the question: "well, why don't you just make them accurate"

docs/site/downloads/cldr-48.md
```diff
@@ -148,7 +148,10 @@ The following changes are planned for CLDR 48. Please plan accordingly to avoid
 
 ### V49 advance warnings
 
-- TBD
+- There is too much uncertainty in the exact values for pre-Meiji Japanese eras,
+and there is feedback that the general practice for exact dates is to use Gregorian for pre-Meiji dates.
+These are slated for removal in a future release.
+Please file a ticket if you use this data. [CLDR-11400]
```
AEApple:
```suggestion
Please add a comment to [CLDR-11400] if you use this data and explain your use case if possible.
```

I'd prefer that they comment on the existing ticket instead of filing a new one. Unless there's a reason they shouldn't?

AEApple (approved):

PR #4896, opened 2025-07-28
CLDR-18840 Remove YMD (8601) data

macchiati:
CLDR-18840

(Pretty simple, just deletes all the paths with `calendar[@type="iso8601"]`)

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #4922, opened 2025-08-04
CLDR-18840 Remove ISO8601 calendar coverage requirement

conradarcturus:
Since we removed the YMD strings -- all locales are failing in coverage. https://github.com/unicode-org/cldr/pull/4896

It wasn't noticed since that change did not re-generate charts. This change regens charts and coverage levels maintained (only sr_Latn and yue_Hans which haven't been generated for v48 yet have dropped in coverage).

CLDR-18840

I'm happy to delete the coverage level stubs altogether rather than just commenting them out -- let me know what you prefer.

ALLOW_MANY_COMMITS=true


AEApple (dismissed):
SGTM!

Commenting these out for this release seems like the best path forward since the data was removed.

conradarcturus:
Hmmm, I hit an error with the TestCoverageLevel test but I don't understand the test code to figure out the remediation.

Command to run the test:
`mvn test --file tools/pom.xml '-Dorg.unicode.cldr.unittest.testArgs=-f:TestCoverageLevel -n -q'`

macchiati:
I'll take a look at it now

macchiati:
Here's what I found out (for the record):

1. iso8601 originally only meant "Gregorian, but force 8601 week number". That usage disappeared, because we changed Gregorian to have that meaning.
2. We gathered data in v48, re-using that ID, where the ordering of the time fields was strictly YMDEHms, with the intention of renaming the id used for it afterwards (to something like gregorian-ymd) — and deprecating the id iso8601 since it was now identical to Gregorian.
3. We had problems with data quality for Gregorian YMD, and decided to hold off including in the release.
4. The retraction needed to change coverage, which PR is intended to fix.

So the fix is either to 
1. completely deprecate the id iso8601 (aliasing it to 'gregorian'), or
2. restore the data to v47 (but have it be still inappropriate, since the data like the following will be misleading, since it is not ISO 8601.
    * fr: <type key="calendar" type="iso8601">calendrier ISO 8601</type> 

Due to the schedule, I suggest that the latter is the right approach for now, but then clean it up in v49.

Note for the future: bcp47/calendar would change as follows:
```<type name="gregory" description="Gregorian calendar" alias="gregorian"/>```
=>
```<type name="gregory" description="Gregorian calendar" alias="gregorian iso8601"/>```
...
```<type name="iso8601" description="ISO calendar (Gregorian calendar using the ISO 8601 calendar week rules)" since="2.0"/>```
=>
```<type name="iso8601" description="ISO calendar (Gregorian calendar using the ISO 8601 calendar week rules)" deprecated="true" preferred="gregorian" since="2.0"/>```

conradarcturus:
Mark's advised change to use "ISO 8601 Weeks" in `en.xml` ended up causing other test failures so I re-ran `GenerateTestData` -- which ended up causing more updates because of updates to the xml after incorporating the v48 survey tool data.

So this change ended up doing a bit more than originally planned -- but hopefully this puts the tests and coverage all in a better place.

PR #4931, opened 2025-08-05
CLDR-18721 update TODO denying VP for non-TC-org

srl295:
- move the TODO from CLDR-18721 to CLDR-18868

CLDR-18721

ALLOW_MANY_COMMITS=true


srl295:
fix the known issues so we can close the original ticket

macchiati (approved):

PR #4932, opened 2025-08-06
CLDR-18869 Use latest (Aug 5) version of ICU4J, update tests/code

pedberg-icu:
CLDR-18869

Update CLDR to use the latest published version of ICU4J (from 2025-08-05). Adjust code and test as necessary:
- ICU4J com.ibm.icu.impl.Utility changed its appendToRule methods to use StringBuilder instead of StringBuffer per ICU-23054 "Fix all 'JdkObsolete' errorprone issues"; as a result, CLDR util/Quoter.java had to update accordingly.  StringBuilder is faster but also not thread-safe; however it looks like that should not be an issue for CLDR Quoter (unless I missed something).
- I thought that the current 2025-08-05 version of ICU4J had removed script Chisoi. However it still exists in UScript and UCharacter, so I filed https://unicode-org.atlassian.net/browse/ICU-23171 about that and then added lokgKnownIssue test skips for places where tests were failing because ICU4J UScript/UCharacter still had some support for Chisoi (though I removed the earlier test skirts that Markus had added).
- Also, with the new ICU4J, some of the data driven personName tests for `km` are failing, so I filed https://unicode-org.atlassian.net/browse/CLDR-18870 and added a logKnownIssue test skip for those.

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/LikelySubtagsTest.java
```diff
@@ -310,13 +310,17 @@ public void TestForMissingScriptMetadata() {
...
+                logKnownIssue("ICU-23171", "ICU4J UScript still supports CHISOI");
+            } else {
+                Info i = ScriptMetadata.getInfo(shortName);
+                if (i == null) {
+                    errln("Script Metadata is missing: " + shortName);
+                    continue;
+                }
+                if (i.likelyLanguage.equals("und") && !exceptions.contains(shortName)) {
+                    errln("Script has no likely language: " + shortName);
+                }
```
markusicu: Let's make this easier to revert.
And we should refer to the still open ICU-23038 rather than 23171 which was unnecessary as a separate ticket.
```suggestion
            Info i = ScriptMetadata.getInfo(shortName);
            if (i == null) {
                if (shortName.equals("Chis")) {
                    logKnownIssue("ICU-23038", "ICU4J UScript still supports CHISOI");
                } else {
                    errln("Script Metadata is missing: " + shortName);
                }
                continue;
            }
            if (i.likelyLanguage.equals("und") && !exceptions.contains(shortName)) {
                errln("Script has no likely language: " + shortName);
            }
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestScriptMetadata.java
```diff
@@ -113,16 +113,13 @@ public void TestScripts() {
         for (int i = UScript.COMMON; i < UScript.CODE_LIMIT; ++i) {
             String longName = UScript.getName(i);
             String shortName = UScript.getShortName(i);
+            if (shortName.equals("Chis")) {
+                logKnownIssue("ICU-23171", "ICU4J UScript still supports CHISOI");
```
markusicu:
```suggestion
                logKnownIssue("ICU-23038", "ICU4J UScript still supports CHISOI");
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
```diff
@@ -2118,7 +2118,11 @@ public void TestNumberingSystemDigitCompleteness() {
         List<Integer> unicodeDigits = new ArrayList<>();
         for (int cp = UCharacter.MIN_CODE_POINT; cp <= UCharacter.MAX_CODE_POINT; cp++) {
             if (UCharacter.getType(cp) == UCharacterEnums.ECharacterCategory.DECIMAL_DIGIT_NUMBER) {
-                unicodeDigits.add(cp);
+                if (UScript.getShortName(UScript.getScript(cp)).equals("Chis")) {
+                    logKnownIssue("ICU-23171", "ICU4J UCharacter still supports CHISOI");
+                } else {
+                    unicodeDigits.add(cp);
+                }
```
markusicu:
```suggestion
                if (UScript.getShortName(UScript.getScript(cp)).equals("Chis")) {
                    logKnownIssue("ICU-23038", "ICU4J UCharacter still supports CHISOI");
                    continue;
                }
                unicodeDigits.add(cp);
```

macchiati (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestScriptMetadata.java
pedberg-icu: Done

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestSupplementalInfo.java
pedberg-icu: Done

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/LikelySubtagsTest.java
pedberg-icu: I did change the ticket number. As for rearranging the core as you suggest: I originally had the code as something like that but then the test hung. I did not bug but assumed it had something to do with the partial support for Chisoi, so I moved the test for Chisoi further out as it currently is in the PR. I would prefer not to change it and debug exactly what is going on, it will all be moot when Icu updates to fully remove Chisoi.

pedberg-icu:
@markusicu @macchiati Per comments from Markus, updated the logKnownIssue ticket number for removing ICU4J Chisoi support (I did not know about the older ticket), and rearranged code in one place as suggested (but not in the other place, see comment in resolved issues). Please TAL and (re)stamp if OK, thanks.

markusicu:
Please don't squash until approved -- makes it harder to track comments and review changes after feedback

markusicu (approved):

PR #4933, opened 2025-08-06
CLDR-11043 Include Greenland, Urumqi, Turkey metazones in modern coverage

robertbastian:
CLDR-11043, CLDR-19425

ALLOW_MANY_COMMITS=true


conradarcturus:
If you want to do this, you will edit `coverageLevels.xml`

That said, we don't know how many languages have translated it so it may be better for v50 when the survey tool opens again next year.

macchiati:
We don't add items that require translation once we are out of vetting, except in extreme circumstances. If it is important to get these in soon, we would add them at the start of the next cycle, and because there is no submission, do it manually (without the normal Survey Tool voting & resolution).

robertbastian:
Yeah I'll probably have to save this for 50. Annoying, according to the comment it should have been done in 46.

AEApple:
> Yeah I'll probably have to save this for 50. Annoying, according to the comment it should have been done in 46.

We could get translations for most of the languages Google supports and just put them in Comprehesive coverage until CLDR 50? @conradarcturus @macchiati  - WDYT?

macchiati:
That works.

robertbastian:
Greenland currently has names in ba, ca, cv, en, et, gd, ia, ku, nl, sc, scn, shn.

Urumqi doesn't have any names, but that's less problematic because the single location is also "Ürümqi" (the locations for the Greenland metazone are "Nuuk" and "Ittoqqortoormiit").

macchiati (commented):
Looks good; just need to resolve merge conflicts.

robertbastian:
can someone approve this please

macchiati (approved):

PR #4935, opened 2025-08-06
CLDR-18591 Restore CLDRFile API for unicodetools

btangmu:
-Restore three methods and three constants, only for backward compatibility

-Based on inspection only, these appear to be the only getName codes and methods currently needed by unicodetools, but testing may prove that more are needed

CLDR-18591

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CLDRFile.java
@@ -3338,4 +3338,36 @@ public class CLDRFile implements Freezable<CLDRFile>, Iterable<String>, LocaleSt
         }
         return value;
     }
+
+    // The following deprecated constants and methods are implemented only for backward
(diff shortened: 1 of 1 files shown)


btangmu:
@srl295 I don't know how to test building/running unicodetools. Maybe it will be more efficient to merge this and see if it fixes the problems, than to test it before merging...

btangmu:
> cross link the original commit

@srl295 What do you mean?

srl295:
> > cross link the original commit
> 
> @srl295 What do you mean?

link to the commit where these APIs were removed.. I think it was in https://unicode-org.atlassian.net/browse/CLDR-15830 which is linked from this ticket. That's all I meant. Probably linking here to the ticket is enough.

markusicu:
Thank you!
What are next steps?
- I think we will need a CLDR snapshot release; who makes those? @pedberg-icu ?
- Then @nedley could try to update the Unicode Tools to that snapshot and see if the tools build.
- @mihnita and @srl295 might be able to help with config issues

PR #4936, opened 2025-08-06
CLDR-11586 change fraction digits to 0 for COP,HUF,IDR,PKR; cashRounding to 5 for HUF

pedberg-icu:
CLDR-11586

Change fraction digits to 0 for COP, HUF, IDR, PKR; also cashRounding to 5 for HUF.

ALLOW_MANY_COMMITS=true


roozbehp:
@pedberg-icu, since you are doing this, would you please also look at https://unicode-org.atlassian.net/browse/CLDR-18611?

pedberg-icu:
> @pedberg-icu, since you are doing this, would you please also look at https://unicode-org.atlassian.net/browse/CLDR-18611?

At first glance that may require some additional structure and thus probably not for 48, but I will take a closer look when I have a chance.

PR #4940, opened 2025-08-06
CLDR-17990 Update GenerateSubdivisions

btangmu:
CLDR-17990

ALLOW_MANY_COMMITS=true


common/supplemental/supplementalMetadata.xml
```diff
@@ -304,6 +304,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<languageAlias type="zsm" replacement="ms" reason="macrolanguage"/> <!-- Standard Malay ⇒ Malay -->
 			<languageAlias type="zyb" replacement="za" reason="macrolanguage"/> <!-- Yongbei Zhuang ⇒ Zhuang -->
 			<languageAlias type="him" replacement="srx" reason="macrolanguage"/> <!-- Himachali ⇒ Sirmauri (= Pahari, Himachali) -->
+			<languageAlias type="mnk" replacement="man" reason="macrolanguage"/> <!-- Mandinka ⇒ Mandingo -->
```
btangmu: @conradarcturus Looking at the history of this file, I see this restores a line you deleted (in https://github.com/unicode-org/cldr/pull/4572), so I've probably made a mistake.

I was trying to follow "Never remove an ID" in https://cldr.unicode.org/development/updating-codes/updating-subdivision-codes

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateSubdivisions.java
```diff
@@ -84,6 +84,7 @@ static final class SubdivisionInfo {
 
     public static void main(String[] args) throws IOException {
         CLDRConfig.getInstance().getSupplementalDataInfo();
+        String outputDirectoryName = CLDRPaths.GEN_DIRECTORY + "subdivision";
```
btangmu, line 89: The changes in this file are just to print out "Writing ..." to indicate what files are being created. Also, `FileUtilities.openUTF8Writer` should get a directory name plus a file name; it was getting part of the directory in the directory name and the other part in the file name

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/SubdivisionNode.java
```diff
@@ -93,17 +91,15 @@ public SubdivisionNode addName(String lang, String value) {
         return this;
     }
 
-    static class SubdivisionSet {
+    public static class SubdivisionSet {
```
btangmu, line 94: There was a compiler warning about visibility, used out of scope, so added "public"

btangmu: Likewise the other changes in this file are only refactoring to fix compiler warnings, and are nearly all automatically corrected by IntelliJ

common/validity/language.xml
```diff
@@ -12,7 +12,7 @@
 <supplementalData>
 	<version number="$Revision$"/>
 	<idValidity>
-		<id type='language' idStatus='regular'>		<!-- 7964 items -->
+		<id type='language' idStatus='regular'>		<!-- 7963 items -->
```
btangmu: Something is suspicious here: decrease from 7964 to 7963, but an item (mna~j) was added, not removed?

Also, the numbers are suspiciously similar to numbers 796, 4, 3, below:

```
$ java -DCLDR_DIR=$(pwd) -jar tools/cldr-code/target/cldr-code.jar generate-validity-data
>> org.unicode.cldr.tool.GenerateValidityXml
Skipping non-Unicode scripts: Afak Blis Chis Cirt Egyd Egyh Geok Inds Jurc Kitl Kpel Leke Loma Maya Moon Nkdb Nkgb Pcun Pelm Phlv Piqd Psin Ranj Roro Sara Shui Teng Visp Wole
Found difference in :.../common/validity/language.xml, .../common/validity/32867-language.xml
 File1: '<id type='language' idStatus='regular'>		<!-- 796', '4 items -->'
 File2: '<id type='language' idStatus='regular'>		<!-- 796', '3 items -->'
```

btangmu: That seems to involve TempPrintWriter.java

common/supplemental/supplementalMetadata.xml
btangmu: I'm going to remove the line, especially since there are related test failures

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/SubdivisionNode.java
btangmu: Except for one essential bug fix: `"\t<!-- deprecated -->"` changed to `"deprecated"`

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/SubdivisionNode.java
```diff
@@ -800,9 +719,8 @@ public void printEnglish(PrintWriter output) throws IOException {
                 }
                 for (String sdCode : remainder) {
                     String name = sdset.getBestName(sdCode, true);
-                    if (!name.equals("???")) {
-                        SubdivisionSet.appendName(
-                                fileSubdivisions, sdCode, name, "\t<!-- deprecated -->");
+                    if (name != null && !name.equals("???")) {
+                        SubdivisionSet.appendName(fileSubdivisions, sdCode, name, "deprecated");
```
btangmu, line 723: Presumably `SubdivisionSet.appendName` was revised at some time, necessitating this change to the last argument

common/supplemental/supplementalMetadata.xml
btangmu: After removing mnk from supplementalMetadata.xml (2nd commit), tests were still failing, all involving mnk. I re-ran generate-validity-data, which changed language.xml (3rd commit)

macchiati:
[This is a long comment just to also provide some background; we should add it to our development pages.]

A line like the following is only valid when the replacement ("man") is a valid macrolanguage and the type ("mnk") is a valid encompassed language, AND is the predominant encompassed language of the macrolanguage.
```
<languageAlias type="mnk" replacement="man" reason="macrolanguage"/> <!-- Mandinka ⇒ Mandingo -->
```
Moreover, every macrolanguage must be the replacement value in exactly one line of that form. So two lines with replacement="man" is a failure, and zero lines with replacement="man" is a failure. (Cf. [Macrolanguage](https://cldr.unicode.org/index/cldr-spec/picking-the-right-language-code#macrolanguages))

There are APIs for this in the Iso639Data class, but the data is also in the /cldr-code/src/main/resources/org/unicode/cldr/util/data/language-subtag-registry, where we have the following:

```
Type: language
Subtag: man
Description: Mandingo
Description: Manding
Added: 2005-10-16
Scope: macrolanguage
```
So "man" is a macrolanguage (from the **scope:**). So we need to verify that mnk is suitable. We do find the following in that file:

```
Type: language
Subtag: mnk
Description: Mandinka
Added: 2009-07-29
Macrolanguage: man
```
It is encompassed by 'man' (from the **Macrolanguage:**).

Now, the choice of which language encompassed by 'man' is the predominant form is not determined mechanically — but there has to be exactly one. But there is already one (changed by Conrad's PR):
```
<languageAlias type="emk" replacement="man" reason="macrolanguage"/> <!-- Maninkakan, Eastern ⇒ Mandingo -->
```
Note: This does not explain why that line was regenerating by one of the tools you ran. However, given the schedule we don't want to invest time in investigating that. So please just file a ticket to fix the tooling, and we'll have to handle later.

----

BTW, here is another example:
```
<languageAlias type="arb" replacement="ar" reason="macrolanguage"/> <!-- Standard Arabic ⇒ Arabic -->

Type: extlang
Subtag: arb
Description: Standard Arabic
Added: 2009-07-29
Preferred-Value: arb
Prefix: ar
Macrolanguage: ar

Type: language
Subtag: ar
Description: Arabic
Added: 2005-10-16
Suppress-Script: Arab
Scope: macrolanguage
```

The AI overview for this is actually pretty good:

In the context of the Unicode Common Locale Data Repository (CLDR), "predominant encompassed language" refers to a specific individual language within a broader macrolanguage that is considered the most common or representative form for that macrolanguage. 

Here's a further explanation:

* Macrolanguages: Some languages have many variations or dialects. To manage this, ISO 639-2 defines "macrolanguages" that encompass these variations. For instance, "ar" represents the Arabic macrolanguage, which includes different Arabic dialects like Standard Arabic ("arb").  
* Predominant Form: Within a macrolanguage, one particular language or dialect might be the most widely used or the one typically associated with the macrolanguage's code in common practice. CLDR designates this as the "predominant encompassed language".  
* Alias Relationship: CLDR establishes an alias relationship where the macrolanguage code and its predominant encompassed language are treated as equivalent in some contexts. This ensures that using the more general macrolanguage code still resolves to the commonly understood or expected form. 

Example

* The Konkani macrolanguage uses the code "kok".  
* Previously, the individual language "knn" (Konkani) was considered the predominant form for "kok".  
* However, based on industry practice and local government information, CLDR changed the predominant encompassed language for "kok" to "gom" (Goan Konkani).  
* This means that while "kok" is the macrolanguage, in many CLDR-based applications and contexts, it will effectively represent "gom".

macchiati:
However, there is something fishy going on. Language validity has the invariant that all of the aliased languages are deprecated. (Remember, deprecated in that context doesn't mean 'can't use'; for aliased languages it just indicates that they are not the canonical form for CLDR.)

Example, arb in:

```
		<id type='language' idStatus='deprecated'>		<!-- 298 items -->
			aam adp agp ais ajp ajt~u als aoh arb asd aue ayr ayx~y azj
```

But language codes can also be deprecated if they are deprecated in the language-subtag-registry.

So your third commit to make mnk be deprecated when it isn't deprecated in the language-subtag-registry is problematic .  That is, _unless the language subtag registry in CLDR is out of date_. Normally the subdivisions need to be done after the other update codes, including the language subtag registry. So I thought; yes, that's it. So I looked at the language subtag registry at https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry, and voila, it has a later date than our internal copy.

File-Date: 2025-07-15

But (sadly) that isn't the answer, because the newest language-subtag-registry has not deprecated mnk; it still has:

Type: language
Subtag: mnk
Description: Mandinka
Added: 2009-07-29
Macrolanguage: man

So you need to un-deprecate mnk, and we do actually need to figure out why the tooling is failing. What were the test errors when you had removed the alias on mnk, but it was still deprecated in validity/?

common/validity/language.xml
btangmu: After the 3rd commit, there are no longer any changes in language.xml, so this puzzle is moot

btangmu:
> What were the test errors when you had removed the alias on mnk, but it was still deprecated in validity/?

Error: (LikelySubtagsTest.java:749) Error: Allowed subtag failure: mnk_Latn_GM [Disallowed language=mnk, status=deprecated]

TestAttributeValues {

TestValid

Warning: (TestAttributeValues.java:301) Warning: Invalid /home/runner/work/cldr/cldr/common/supplemental/likelySubtags.xml likelySubtag from mnk

Warning: (TestAttributeValues.java:301) Warning: Invalid /home/runner/work/cldr/cldr/common/supplemental/likelySubtags.xml likelySubtag to mnk_Latn_GM

Error: (TestAttributeValues.java:415) Error: For [supplemental, supplemental-temp, transforms, validity]

If the test fails, use -v for details. Also look at https://cldr.unicode.org/development/updating-codes/testattributevalues for guidance.

file Count: supplementalData 408

element Count: supplementalData 24171

attribute Count: supplementalData 60690

Status DtdType Element Attribute Match expression #Failures Failing values

invalid supplementalData likelySubtag from validity/locale-for-likely 1 mnk

invalid supplementalData likelySubtag to validity/locale-for-likely 1 mnk_Latn_GM

testLanguageTagParserIsValid

Error: (TestLocale.java:939) Error: : mnk: expected "", got "Disallowed language=mnk, status=deprecated"

Error: (TestLocale.java:945) Error: : mnk_Latn_GM: expected "", got "Disallowed language=mnk, status=deprecated"

TestPaths {

TestNonLdml

Error: (TestPaths.java:629) Error: supplemental/likelySubtags.xml/likelySubtag@from, expected match to: ⟪validity/locale-for-likely⟫ actual: «mnk»

Error: (TestPaths.java:629) Error: supplemental/likelySubtags.xml/likelySubtag@to, expected match to: ⟪validity/locale-for-likely⟫ actual: «mnk_Latn_GM»

Error: (TestPaths.java:629) Error: supplemental/likelySubtags.xml/likelySubtag@from, expected match to: ⟪validity/locale-for-likely⟫ actual: «mnk»

Error: (TestPaths.java:629) Error: supplemental/likelySubtags.xml/likelySubtag@from, expected match to: ⟪validity/locale-for-likely⟫ actual: «mnk»

Error: (TestPaths.java:629) Error: supplemental/likelySubtags.xml/likelySubtag@to, expected match to: ⟪validity/locale-for-likely⟫ actual: «mnk_Latn_GM»

TestLstrConsistency

Error: (TestValidity.java:626) Error: Deprecated validity - lstr

LstrType= language

size= 1

codes= [mnk]

names= [

mnk ⇒ {Description=Mandinka, Added=2009-07-29, Macrolanguage=man}

valid.⇒ deprecated

alias⇒ null]

macchiati:
Is this ready for review now?

btangmu:
> Is this ready for review now?

Yes. Sorry, I was out of office yesterday. When I responded to your question Friday about the earlier test errors (mnk_Latn_GM etc.), after the 3rd commit, the tests were already passing. However, you had written that the 3rd commit was "problematic". I'm making this ready for review with the caveat that I don't understand all the implications yet.

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateSubdivisions.java
macchiati: These changes are unnecessary. There is an environment variable, SHOW_FILES, which if present causes the various utilities in FileUtilities to log when files are opened. That can be used with any of the tools, not just this one.

The real problem is that we don't document SHOW_FILES in the file or in https://cldr.unicode.org/

macchiati (changes requested):
Only problem is with the addition of print statements to GenerateSubdivisions, otherwise LGTM

tools/cldr-rdf/src/main/java/org/unicode/cldr/tool/GenerateSubdivisions.java
btangmu: That's fixed by the 4th commit

macchiati (approved):

PR #4942, opened 2025-08-07
CLDR-18334 adding-the-new-saudi-riyal

stenshamn:
CLDR-18334

ALLOW_MANY_COMMITS=true


common/main/root.xml
```diff
@@ -5260,6 +5260,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<currency type="RWF">
 				<symbol alt="narrow">RF</symbol>
 			</currency>
+			<currency type="SAR" alt="variant">
+    			<symbol alt="variant">&#20C1;</symbol>
```
markusicu:
```suggestion
    			<symbol alt="variant">&#x20C1;</symbol>
```
see https://en.wikipedia.org/wiki/Numeric_character_reference

stenshamn: Thanks, I thought it looked odd...

macchiati (dismissed):

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestCodePointEscaper.java
```diff
@@ -88,6 +89,11 @@ void testInRootAndEnglish() {
                         continue;
                     case Currencies:
                         if (ph.getCode().startsWith("XOF")) {
+                            continue; // TODO: log known issue?
+                        }
+                        if (ph.getCode().startsWith("SAR")
+                                && logKnownIssue(
+                                        "CLDR-18334", "some random test failure, revisit")) {
```
markusicu, line 96: The logKnownIssue() should point to a CLDR ticket that will be open after you close this one for adding the currency symbol.

srl295: Before the ticket closes yes, but it should be reworded because I have no idea what the test is trying to do. Did it this way so the fix can merge first awaiting further input.

macchiati (approved):

macchiati:
Sorry I wasn't around to help. I will have to check to make sure, but I think the test problem _might_ be that the code point is not assigned*, which CLDR disallows in the main directory. I can look at it when I'm back at work, and give a better analysis.

\* as far as CLDR knows, because I don't think it is up on the latest ICU with Unicode 17 yet. If that is the case, then it is something that we should probably prepare for in the future, since it seems like there are going to be more cases like this.

PR #4943, opened 2025-08-08
CLDR-18840 EH/H problems

macchiati:
CLDR-18840

This makes changes in v48 to revert problems with the introduction of 'h' into EH and H patterns.

- It uses the v47 data where there was a difference.
- Removes EH for this release since it was new in CLDR 48 and we are not confident in the data
- There are some tweaks, because the patterns in the data have some variance.

It also adds some minor improvements in ListCoverageLevels 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/be.xml
+++ b/common/main/be.xml
@@ -1383,7 +1383,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="EBhms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Ed">E, d</dateFormatItem>
 						<dateFormatItem id="Eh">↑↑↑</dateFormatItem>
-						<dateFormatItem id="EH">↑↑↑</dateFormatItem>
+
 						<dateFormatItem id="Ehm">E h:mm a</dateFormatItem>
 						<dateFormatItem id="EHm">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Ehms">E h:mm:ss a</dateFormatItem>
@@ -1936,7 +1936,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="EBhms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Ed">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Eh">↑↑↑</dateFormatItem>
-						<dateFormatItem id="EH">E HH 'г'</dateFormatItem>
+
 						<dateFormatItem id="Ehm">E h:mm a</dateFormatItem>
 						<dateFormatItem id="EHm">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Ehms">E h:mm:ss a</dateFormatItem>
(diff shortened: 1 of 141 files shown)


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
```diff
@@ -2579,7 +2579,7 @@ public void testAvailableAndIntervalExamples() {
             {"Hmv", "〖13:25 EST〗〖03:25 EST〗〖Related formats:〗〖13:25〗"},
             {"Hv", "〖13h EST〗〖03h EST〗"},
             {"Eh", "〖Sun 1 PM〗〖Sun 3 AM〗"},
-            {"EH", "〖Sun 13h〗〖Sun 03h〗"},
+            {"EH", "〖Sun 13〗〖Sun 03〗"},
```
conradarcturus: Apologies if I don't understand it -- "Sun 13" looks less clear than "Sun 13h".

macchiati: In that combination, it probably is. But we decided to revert the changes for EH, H pending a better solution. We need to think through what the best solution is, and make sure that the vetters follow through consistently.

macchiati: That being said, if people think it is better to retain the EH with "h", etc combinations for now, I'm ok with removing their reversions. @AEApple @stenshamn , what do you think?

stenshamn: > That being said, if people think it is better to retain the EH with "h", etc combinations for now, I'm ok with removing their reversions. @AEApple @stenshamn , what do you think?

What was it in 47? Or didn't we have this in 47?

macchiati: We didn't have EH paths in v47.

AEApple: I think we should revert them until we know how they will be used.

common/main/be.xml
```diff
@@ -1936,7 +1936,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="EBhms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Ed">↑↑↑</dateFormatItem>
 						<dateFormatItem id="Eh">↑↑↑</dateFormatItem>
-						<dateFormatItem id="EH">E HH 'г'</dateFormatItem>
+						<dateFormatItem id="EH">E H</dateFormatItem>
```
AEApple: @macchiati - Shouldn't this line disappear? I didn't think we had EH in CLDR 47?

macchiati:
Good catch. I tried to make items consistent, but it would be better to just remove those for now, because we have little confidence in the resulting data.  I'll make a follow-up commit.

They are (in root gregorian) of the form:
```
<dateFormatItem id="EH">E HH'h'</dateFormatItem>
```

AEApple:
> Good catch. I tried to make items consistent, but it would be better to just remove those for now, because we have little confidence in the resulting data. I'll make a follow-up commit.
> 
> They are (in root gregorian) of the form:
> 
> ```
> <dateFormatItem id="EH">E HH'h'</dateFormatItem>
> ```

I am going to get UX to help me figure out what the patterns would be that someone would use the EH format. If it turns out that these are items where we don't think we can have clear localization, and it is not in a way that UX would generally recommend I wonder if we should take a different approach where developers get warned if they try using them instead of giving them problematic results?

macchiati:
Yes, I think that for a 'naked' EH (or even Eh), it is likely that languages use some better disambiguation, such as
"Tuesday at 3"
"3 on Tuesday"
rather than something like
"Tuesday, 3"
"Tuesday 3"
"Tuesday 3h"

macchiati:
Just merged the removals. It is leaving blank lines, but those will disappear when the files are normalized by the CLDRModify pass.

macchiati:
Note: there is a bad commit message; once we get all other issues resolved, I'll squash.

macchiati:
You can switch to Workspace for large ones. I can follow up with the how-to.

macchiati:
Just squashed.

PR #4946, opened 2025-08-09
CLDR-18840 Regenerate currency formats

macchiati:
CLDR-18840

This modifies CLDRModify to add an -fC option, which adds derived currency path/value pairs to the locales. It also processes the files with that option. It uses the standard currency format and the accounting currency format to generate alternate formats: alphaNextToNumber, noCurrency, and the currency format for compact numbers. This ensures consistency across all of these formats.

For example:
* common/main/af.xml had <pattern type="1000" count="other" alt="alphaNextToNumber">↑↑↑</pattern>, but that resolves to ¤0 k, which for alphaNextToNumber needs a space before the 0.

It also fixes some significant other glitches in the data. For example, one line in ba.xml had two ¤ in the negative format, one on both sides of the number. I had to fix this manually so that the tool could run. (Filed a ticket to detect.)

Notes:
1. I don't expect more than spot-checks; there are quite a number of changes. Reviewers should just spot-check a few sample locales.
   * To see what changed, sometimes you have to look at what a value like ↑↑↑ or missing would have been. Remember that count="x" will inherit from count="other", and alt="x" will inherit from the no-alt version.
   * The noCurrency values will not have a currency, and the alphaNextToNumber will have a space between the currency and number.
   * Sometimes you won't see a difference, where the only difference is in a space (eg, no-breaking vs breaking).
3. A recent change to the Hawaii metazones changed the ordering of elements in the current main file to be non-canonical. If we had time to first do a CLDRModify pass on main to clean up the format, then some files would disappear once rebased, but for now, just ignore the Hawaii changes.
    * For example, common/main/am.xml has no changes other than to the Hawaii metazones
4. A current limitation is that the tool needs to be run more than once, in case items are created the first time that also generate others.  (TODO recorded in CLDR-18840)
5. One other limitation is that sometimes it generates an explicit value when ↑↑↑ would be sufficient. Those have identical effects for the production software, so that is not a serious issue.
6. Many files like ba.xml have an excessive number of comprehensive numbering systems, most or all of which just inherit.

Code

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDerivedMain.java
* Most of the code is here. The main method is just is just for development. A later PR will move this to a test class (TODO recorded in CLDR-18840).
* It gathers data from each file, then uses that to generate the new paths, and returns those that are different (from the resolved value).

* The change is to hook up a new -fC to use GenerateDerivedMain.getPathValuesToAdd() 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/cldr-big-red-switch/cldrmodify-passes.md
+++ b/docs/site/development/cldr-big-red-switch/cldrmodify-passes.md
@@ -67,7 +67,7 @@ There are a number of "one time fixes" that are in the CLDRModify code. The typi
 #### Inactive options
 The following are some of the **inactive** options. The code remains in case we want to adapt for future cases, but don't use them unless you fix the code to do what you want, and carefully diff the results. You will have to enable them in checkSuboptions(): limits the ones that are active to prevent mistakes.
 
-1. \-fc: Transition from an old currency to a new currency.  This fix is quite useful when a country introduces a new currency code ( usually due to a devaluation ), but the name remains the same.  In order to use this fix, modify the following values in the CLDRModify code under "**fixList.add('c', "Fix transiton from an old currency code to a new one**"
+1. \-fc: Transition from an old currency to a new currency. This fix is quite useful when a country introduces a new currency code (usually due to a devaluation), but the name remains the same. In order to use this fix, modify the following values in the CLDRModify code under "**fixList.add('c', "Fix transiton from an old currency code to a new one**"
     1. Change the String oldCurrencyCode and newCurrencyCode to reflect the currency codes you are transitioning.  
     2. Change the int fromDate and toDate to reflect the dates that the old currency was in circulation.
 These will be used to create the date range in the old currency string.  
--- a/docs/site/development/cldr-big-red-switch/cldrmodify-passes.md
+++ b/docs/site/development/cldr-big-red-switch/cldrmodify-passes.md
@@ -23,7 +23,7 @@ For the purpose of this document, we'll assume you are generating into {cldrdata
 ### Passes
 
   * Each pass will go from a sourceDir (default=CLDRPaths.***MAIN\_DIRECTORY)*** to a targetDir (default \= CLDRPaths.***GEN\_DIRECTORY*** \+ "cldrModify/")  
-  * Empty the contents of targetDir before running.  (If you forget, just remember after building to delete the old files.)
+  * Empty the contents of targetDir before running. (If you forget, just remember after building to delete the old files.)
   * Each time you will run CLDRModify with the standard options, plus one of the listed [Options](#options) bulleted below.
   * The console will list changes made, such as:  
      * Creating File: {cldrdata}/dropbox\\gen\\main\\zh\_Hant.xml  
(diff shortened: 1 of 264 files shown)


AEApple:
Can we add the docs updates to describe the new option -fC to cldrmodify-passes.md as part of this PR?

macchiati:
ok

macchiati:
Made updates to the CLDRModify doc. (It needed quite a bit of work, so should be generally better.)

docs/site/development/cldr-big-red-switch/cldrmodify-passes.md
```diff
@@ -16,28 +16,27 @@ You will run CLDRModify with different options, in multiple passes.
...
 
 For the purpose of this document, we'll assume you are generating into {cldrdata}/dropbox/gen/main/ as the target directory. Change any instance below to the directory that you actually use.
 
 ### Passes
 
-* Each pass will go from a sourceDir (default=CLDRPaths.***MAIN\_DIRECTORY)*** to a targetDir (default \= CLDRPaths.***GEN\_DIRECTORY*** \+ "cldrModify/")  
-  * Empty the contents of targetDir before running.  
-  * Each time you will run CLDRModify with the standard options, plus one of the listed [Options](#options) bulleted below.  
+  * Each pass will go from a sourceDir (default=CLDRPaths.***MAIN\_DIRECTORY)*** to a targetDir (default \= CLDRPaths.***GEN\_DIRECTORY*** \+ "cldrModify/")  
+  * Empty the contents of targetDir before running.  (If you forget, just remember after building to delete the old files.)
```
stenshamn:
```suggestion
  * Empty the contents of targetDir before running. (If you forget, just remember after building to delete the old files.)
```

docs/site/development/cldr-big-red-switch/cldrmodify-passes.md
```diff
@@ -48,46 +47,35 @@ Standard Options: add to your regular preferences \-DSHOW\_FILES *plus* your cho
...
+    * Currently this needs to be run multiple times, until there are no more changes. This is because modifications trickle down due to inheritance.
+3. \-fE: Fix null/inherited values in en.xml.
+
+#### Config file
+1. \-fk: use a configuration file; see the details on [CLDRModify Config](cldrmodify-using-config-file).
+
+#### Inactive options
+The following are some of the **inactive** options. The code remains in case we want to adapt for future cases, but don't use them unless you fix the code to do what you want, and carefully diff the results. You will have to enable them in checkSuboptions(): limits the ones that are active to prevent mistakes.
+
+1. \-fc: Transition from an old currency to a new currency.  This fix is quite useful when a country introduces a new currency code ( usually due to a devaluation ), but the name remains the same.  In order to use this fix, modify the following values in the CLDRModify code under "**fixList.add('c', "Fix transiton from an old currency code to a new one**"
```
stenshamn:
```suggestion
1. \-fc: Transition from an old currency to a new currency. This fix is quite useful when a country introduces a new currency code (usually due to a devaluation), but the name remains the same. In order to use this fix, modify the following values in the CLDRModify code under "**fixList.add('c', "Fix transiton from an old currency code to a new one**"
```

macchiati:
Can someone approve and merge this? I can followup with other changes, but don't want to have conflicts creep in.

PR #4947, opened 2025-08-09
CLDR-18873 Revert Myanmar main exemplars to v47

macchiati:
CLDR-18873

ALLOW_MANY_COMMITS=true


macchiati:
This simply reverts 1 line to the v47 value.

macchiati:
Had to update the scriptToExemplars also (mechanical fix)

PR #4948, opened 2025-08-09
CLDR-11763 Remove GST again

AEApple:
CLDR-11763

It looks like GST came back again via the SurveyTool. I didn't think to vote for removal since that feature was only added at the end of resolution this release.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ar.xml
+++ b/common/main/ar.xml
@@ -5567,9 +5567,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<long>
 					<standard>توقيت الخليج</standard>
 				</long>
-				<short>
-					<standard draft="contributed">GST</standard>
-				</short>
 			</metazone>
 			<metazone type="Guyana">
 				<long>


robertbastian (commented):
What about `en-GB` and `en-IN`?

macchiati:
Yes, something to remember is that all removals in XML have to happen either before Shakedown or after ST-to-XML. Otherwise they will just be nuked when we bring the data in.

macchiati (approved):

macchiati:
I see one other instance of GST, in https://github.com/unicode-org/cldr/blob/e2e3ca6e68ae6602d617185558abf916f7ad8b0d/common/main/en_AE.xml#L343 However, that may be ok, given that AE is on the Gulf.

AEApple:
I'm going to merge this removal we should file another ticket to discuss en_GB and en_IN and whether they should have GST or not. I would expect that GST is used in en_IN considering that many people fly through Dubai from India.

I'm not sure about en_GB.

However there are only two abbreviations for GST, so maybe this is intentional? https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations

PR #4949, opened 2025-08-09
CLDR-18876 Fix long compact decimal format in en_IN

AEApple:
CLDR-18876

Fix long compact decimal format in en_IN based on CLDR-13759. Used inheritance instead of T for values below a lakh since  CLDR-17232, and it should have enough space if thousand is what's used in en.

We should investigate why this wasn't raised as an issue when the vetters reviewed the numbers report.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_IN.xml
+++ b/common/main/en_IN.xml
@@ -6145,10 +6145,10 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 					<pattern type="100000000" count="other" draft="contributed">00 crore</pattern>
 					<pattern type="1000000000" count="one" draft="contributed">000 crore</pattern>
 					<pattern type="1000000000" count="other" draft="contributed">000 crore</pattern>
-					<pattern type="10000000000" count="one" draft="contributed">0T crore</pattern>
-					<pattern type="10000000000" count="other" draft="contributed">0T crore</pattern>
-					<pattern type="100000000000" count="one" draft="contributed">00T crore</pattern>
-					<pattern type="100000000000" count="other" draft="contributed">00T crore</pattern>
+					<pattern type="10000000000" count="one" draft="contributed">0000 crore</pattern>
+					<pattern type="10000000000" count="other" draft="contributed">0000 crore</pattern>
+					<pattern type="100000000000" count="one" draft="contributed">00000 crore</pattern>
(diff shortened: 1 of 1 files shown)


common/main/en_IN.xml
```diff
@@ -6131,30 +6131,30 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+					<pattern type="100000" count="other" draft="contributed">0 lakh</pattern>
+					<pattern type="1000000" count="one" draft="contributed">00 lakh</pattern>
+					<pattern type="1000000" count="other" draft="contributed">00 lakh</pattern>
+					<pattern type="10000000" count="one" draft="contributed">0 crore</pattern>
+					<pattern type="10000000" count="other" draft="contributed">0 crore</pattern>
+					<pattern type="100000000" count="one" draft="contributed">00 crore</pattern>
+					<pattern type="100000000" count="other" draft="contributed">00 crore</pattern>
+					<pattern type="1000000000" count="one" draft="contributed">000 crore</pattern>
+					<pattern type="1000000000" count="other" draft="contributed">000 crore</pattern>
+					<pattern type="10000000000" count="one" draft="contributed">0T crore</pattern>
```
stenshamn: The short formats use K to designate a thousand, not T. 
But since this is not a short format, should we even abbreviate it, or call it `0 thousand crore`?

macchiati: Yes, I'm a bit suspicious of using T (because of the confusion with trillion). Would it be better with either

case 1

0 T crore -> 0K crore
00T crore -> 00K crore

or, to avoid the K:

case 2
0 T crore -> 0000 crore
00T crore -> 00000 crore

macchiati (commented):
Whoops, didn't submit review comment, so it was Pending.

common/main/en_IN.xml
AEApple: I was planning to discuss in the TC meeting tomorrow since we have more than one option, but I did want people to take a look at this ahead of time in case there was a strong preference since 00T crore gets quite long if we don't use 00k crore

AEApple:
Updated the PR per the TC's feedback and it is ready for re-review.

PR #4953, opened 2025-08-11
CLDR-18879 Update MessageFormat for v48

eemeli:
CLDR-18879

The tooling used to map the MessageFormat WG's editable copy of the spec to this form was recently updated in unicode-org/message-format-wg#1096, and a rendered version of the editable spec is now available at https://unicode-org.github.io/message-format-wg/.

That currently corresponds to this PR, but may change if and as the WG makes further changes. No further changes are planned for inclusion in LDML 48.

ALLOW_MANY_COMMITS=true


srl295 (commented):
no production or generation issues i can see.

srl295:
> no production or generation issues i can see.

As discussed during the CLDR TC meeting, the commit message numbers could be better `unicode-org/message-format-wg#1096` or the full PR URL instead of just `#1096`

PR #4954, opened 2025-08-11
CLDR-17443 Fix person-name problem in Spanish

macchiati:
The alt and plain forms should differ by {surname2}, as pointed out in the ticket. (It doesn't matter which, so just following the medium choice). Look at the ticket for details.

CLDR-17443

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/es.xml
+++ b/common/main/es.xml
@@ -11780,7 +11780,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<namePattern alt="1">{surname} {surname2}, {title} {given} {given2}</namePattern>
 		</personName>
 		<personName order="sorting" length="long" usage="referring" formality="informal">
-			<namePattern>{surname} {surname2}, {given-informal}</namePattern>
+			<namePattern>{surname}, {given-informal}</namePattern>
 			<namePattern alt="1">{surname} {surname2}, {given-informal}</namePattern>
 		</personName>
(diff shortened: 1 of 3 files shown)


macchiati:
Whoops, neglected to regenerate the test data. Will follow up with commit.

macchiati:
There are three items mentioned:

1. The two namePatterns for es locale for order="sorting" length="long" usage="referring" formality="informal" are the same
   * This was fixed
3. Using that namePattern with a data record with no surname results a combination of two strings that seems incorrect being " , " using the combining rules in the specification
   * The fix was to produce the desired output
5. The CLDR test cases generate a result that seems non-compliant with the spec, and possible also incorrect.
   * I don't think the test cases are non-compliant (I regenerated with the fix); I think the problem was with the data. However, I can hive off that part into a separate subtask for us to check whether it is the case, and if so, fix in the spec. (The deadline for data changes is fast approaching, whereas we have more time for spec changes.)

macchiati:
Adding SRL (I think Rich and Peter are both out)

AEApple (approved):

AEApple:
Is there a ticket tracking adding a test for this if they need to be different?

macchiati:
> Is there a ticket tracking adding a test for this if they need to be different?

Good idea. I should file a ticket for 49 before we close the ticket

PR #4956, opened 2025-08-12
CLDR-18620 Add support for regional English in 419

AEApple:
CLDR-18620

Creates support for regional en-419 locales.

Excludes locales which have closer ties to en-GB and prefer direct en-001 inheritance

ALLOW_MANY_COMMITS=true


common/supplemental/supplementalData.xml
```diff
@@ -5318,6 +5318,7 @@ XXX Code for transations where no currency is involved
...
 		<parentLocale parent="en_001" locales="en_150 en_AG en_AI en_AU en_BB en_BM en_BS en_BW en_BZ en_CC en_CK en_CM en_CX en_CY en_DG en_DM en_ER en_FJ en_FK en_FM en_GB en_GD en_GG en_GH en_GI en_GM en_GS en_GY en_HK en_ID en_IE en_IL en_IM en_IN en_IO en_JE en_JM en_KE en_KI en_KN en_KY en_LC en_LR en_LS en_MG en_MO en_MS en_MT en_MU en_MV en_MW en_MY en_NA en_NF en_NG en_NR en_NU en_NZ en_PG en_PK en_PN en_PW en_RW en_SB en_SC en_SD en_SG en_SH en_SL en_SS en_SX en_SZ en_TC en_TK en_TO en_TT en_TV en_TZ en_UG en_VC en_VG en_VU en_WS en_ZA en_ZM en_ZW"/>
 		<parentLocale parent="en_150" locales="en_AT en_BE en_CH en_CZ en_DE en_DK en_EE en_ES en_FI en_FR en_GE en_HU en_IT en_LT en_LV en_NL en_NO en_PL en_PT en_RO en_SE en_SI en_SK en_UA"/>
+		<parentLocale parent="en_419" locales="en_AR en_AW en_BL en_BO en_BQ en_BR en_BV en_CL en_CO en_CR en_CU en_CW en_DO en_EC en_GF en_GP en_GT en_HT en_MF en_MQ en_MX en_NI en_PA, en_PE en_PR en_PY en_SR en_SV en_UY en_VE"/>
```
AEApple:
```suggestion
		<parentLocale parent="en_419" locales="en_AR en_AW en_BL en_BO en_BQ en_BR en_BV en_CL en_CO en_CR en_CU en_CW en_DO en_EC en_GF en_GP en_GT en_HT en_MF en_MQ en_MX en_NI en_PA en_PE en_PR en_PY en_SR en_SV en_UY en_VE"/>
```

conradarcturus:
I wonder if this will confuse people with es_419. To avoid that I may recommend sub-continents rather than regions, so: 



en_005 (South America)
en_FK
en_GY

en_013 (Central America)
en_BZ

en_029 (Caribbean)
en_AG
en_AI
en_BB
en_BS
en_DM
en_GD
en_JM
en_KN
en_KY
en_MS
en_SX
en_TC
en_TT


FWIW see all territories here: https://translation-commons.github.io/lang-nav/data?objectType=Territory&view=Hierarchy&searchBy=ID&searchString=019&sortBy=Code

AEApple:
@conradarcturus - you want to treat en differently than we treat es? We don't have the breakdown in es that you are proposing. Also that goes against what was agreed on in https://unicode-org.atlassian.net/browse/CLDR-18809 where the goal is to group inheritance based on shared locale data instead of based on region. FWIW, es_JP is actually Latin American Spanish (Japan) as well :)

Happy to make a es-029 which is a child of en-419 if you think it would be better? We do have a ticket requesting it, I'm just trying to avoid many levels of data that we have to maintain consistency in. https://unicode-org.atlassian.net/browse/CLDR-15031

conradarcturus:
Ah sorry for not following up on my comment & then leaving for vacation. Do you plan to continue this PR (presumably for v49 now?)

To reply to your comment: Your reasoning is sound :) I withdraw my comment. I was overly cautious about people mixing en_419 and es_419 but I don't think that's necessary.

PR #4962, opened 2025-08-12
CLDR-18891 update TestTransforms tests to run register() first

srl295:
CLDR-18891

Allows this to run where it failed before.

```shell
$ mvn --file=tools/pom.xml -pl cldr-code test  '-Dorg.unicode.cldr.unittest.testArgs=-e10 -n -filter:TestLocales' -Dtest=org.unicode.cldr.unittest.TestShim
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTransforms.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestTransforms.java
@@ -115,7 +115,6 @@ public class TestTransforms extends TestFmwkPlus {
 
     public void TestCyrillicLatin() {
         // this method only works for 'leaf' rule-based translators
-        register();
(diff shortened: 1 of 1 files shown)


macchiati (dismissed):

macchiati:
BTW, I think you can add this to the test instance with the following, avoiding a repeat
{
  // do in constuctor
  register();
}

AEApple:
Can we just adjust this as an infra ticket since this is just for tests and shouldn't affect data?

srl295:
> BTW, I think you can add this to the test instance with the following, avoiding a repeat { // do in constuctor register(); }

constructor might be too early for some cases.

in fact there's an `init()` call that can be used instead!

macchiati (approved):

macchiati:
You can accept the ticket as infrastructure

PR #4974, opened 2025-08-18
CLDR-18722 Explicit numberSystem attribute

btangmu:
-Revise ldml.dtd to allow numberSystem attribute for pattern element

-Add numberSystem (latn) for pattern/decimal/group elements where needed in locales ca, el, en_150, en_AT, eu, gl, it, tr

-Re-enable the check for numberSystem in CheckNumbers.java

-Use parts.containsAttribute instead of parts.getAttribute(2, ...) since the element index may not be 2

CLDR-18722

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -2064,7 +2064,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST decimal references CDATA #IMPLIED >
     <!--@METADATA-->
 <!ATTLIST decimal numberSystem CDATA #IMPLIED >
-    <!--@DEPRECATED-->
+    <!--@MATCH:bcp47/nu-->
 
 <!ELEMENT group ( #PCDATA ) >
 <!ATTLIST group alt NMTOKENS #IMPLIED >
@@ -2075,7 +2075,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST group references CDATA #IMPLIED >
     <!--@METADATA-->
 <!ATTLIST group numberSystem CDATA #IMPLIED >
-    <!--@DEPRECATED-->
+    <!--@MATCH:bcp47/nu-->
 
 <!ELEMENT list ( #PCDATA ) >
 <!ATTLIST list alt NMTOKENS #IMPLIED >


btangmu:
Locally, after making these changes, I ran console check as follows:
```
java -DCLDR_DIR=$(pwd) -jar tools/cldr-code/target/cldr-code.jar check -S common,seed -1 -e -z FINAL_TESTING -c comprehensive
```
Note that `-c comprehensive` was required, otherwise the paths that previously had `numberSystem` errors would be skipped. There were 62 errors, but none of them involved `numberSystem`.

common/main/gl.xml
```diff
@@ -5385,13 +5385,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
 			</currency>
 			<currency type="ESP">
-				<pattern>#,##0.00 ¤</pattern>
+				<pattern numberSystem="latn">#,##0.00 ¤</pattern>
 				<displayName draft="contributed">Peseta española</displayName>
 				<displayName count="one" draft="contributed">peseta</displayName>
 				<displayName count="other" draft="contributed">pesetas</displayName>
 				<symbol>₧</symbol>
-				<decimal>,</decimal>
-				<group>❰NBTSP❱</group>
```
macchiati, line 5394: There is something wrong here; unrelated to this ticket. The [❰❱] characters should never occur in values for our data files. It looks like vetters might be pasting them into values. I'll file a separate ticket about this.

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
```diff
@@ -2051,7 +2051,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST decimal references CDATA #IMPLIED >
```
macchiati, line 2051: I'm wondering why this file exists; I've noticed that it causes errors when it drifts way from the regular ldml.dtd. If we keep it, we should have a test that it is identical (and maybe we do, and that's what alerted you to it?)

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckNumbers.java
```diff
@@ -259,12 +259,8 @@ public CheckCLDR handleCheck(
         }
```
macchiati, line 259: I take it that NumericType still gets these items, so they aren't skipped by line 258.

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
macchiati: In eclipse I see the following error with that file:

Description	Resource	Path	Location	Type
The element 'gmtUnknownFormat' has not been declared.	ldml.dtd	/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd	line 1723	DTD Problem

btangmu: I'm confused by that too. There are 3 ldml.dtd files:

1. common/dtd/ldml.dtd
2. tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
3. tools/cldr-code/target/test-classes/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd

I was accidentally editing the wrong one before I noticed that...

macchiati:
Changing to Draft so we don't mistakenly merge into 48 main.

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
btangmu: In fact I did that again even after realizing there are 3 files; this PR still fails maybe for that reason. I'll make another (4th) commit shortly.

@srl295 do you have insight about how the ldml.dtd files relate to each other?

common/main/gl.xml
btangmu: @macchiati Did you file that ticket yet?

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
btangmu: @srl295 I forget if we discussed this in the infrastructure meeting or slack. Although it's tangential to this ticket, maybe there should be a new ticket for it?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckNumbers.java
btangmu: Sorry for the late response; this PR was put on hold pending branching maint/maint-48.

I don't quite understand the question.

This is more than 100 lines into `handleCheck` and this PR only does two things here, a few lines further down:

1. It re-enables the "Number formats must have an explicit numberSystem attribute." check.
2. It makes the check more general, since the `numberSystem` attribute can occur in different places, not necessary index 2.

Is the concern that some paths that would now match would be skipped prematurely because they have `NumericType.NOT_NUMERIC` or `NumericType.RATIONAL`? I could investigate that...

btangmu:
Now that we've branched maint/maint-48, I've made this ready for review again, and added Steven and Peter as reviewers, and responding to a couple of comments

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
srl295: No it's already on my list, just ignore that file i have an issue to correct it

btangmu:
I see the ticket is now done, with this follow-up ticket for v49: https://unicode-org.atlassian.net/browse/CLDR-18963

I don't know if there's a way to move the PR to a different ticket; should I just make a new PR and refer back to this one for discussion?

srl295:
> I see the ticket is now done, with this follow-up ticket for v49: https://unicode-org.atlassian.net/browse/CLDR-18963
> 
> 
> 
> I don't know if there's a way to move the PR to a different ticket; should I just make a new PR and refer back to this one for discussion?

Just retitle the pr and then either use the auto squash link or manually reword the git commits and force push

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
srl295: i recommend to revert changes to this file

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
```diff
@@ -2062,7 +2062,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST group references CDATA #IMPLIED >
     <!--@METADATA-->
 <!ATTLIST group numberSystem CDATA #IMPLIED >
-    <!--@DEPRECATED-->
+    <!--@MATCH:bcp47/nu-->
```
srl295, line 2065:
```suggestion
    <!--@DEPRECATED-->
```

tools/cldr-code/src/test/resources/org/unicode/cldr/unittest/data/common/dtd/ldml.dtd
```diff
@@ -2051,7 +2051,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 <!ATTLIST decimal references CDATA #IMPLIED >
     <!--@METADATA-->
 <!ATTLIST decimal numberSystem CDATA #IMPLIED >
-    <!--@DEPRECATED-->
+    <!--@MATCH:bcp47/nu-->
```
srl295, line 2054:
```suggestion
    <!--@DEPRECATED-->
```

common/main/gl.xml
btangmu: @macchiati I created a ticket, just in case: https://unicode-org.atlassian.net/browse/CLDR-19075

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckNumbers.java
btangmu: @macchiati i I tried commenting out these lines:

```
         if (type == NumericType.NOT_NUMERIC || type == NumericType.RATIONAL) {
             return this; // skip
         }
```

Console check then reported 3,175,054 "missing numbering system" errors. That's a lot. If there are any cases where the "skip" should not occur, meaning that a "numberSystem" attribute should be required even though the type is NOT_NUMERIC or RATIONAL, I don't know how to find such needles in such a large haystack. I'm pretty sure, though, that I just don't understand your question.

btangmu:
I decided to make a new PR instead of trying to change the ticket number and commit messages and force push: https://github.com/unicode-org/cldr/pull/5149

PR #4977, opened 2025-08-19
CLDR-18901 Fix escapes

macchiati:
CLDR-18901

ALLOW_MANY_COMMITS=true


common/main/as.xml
```diff
@@ -925,13 +925,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<type key="em" type="emoji" scope="core">ইম’জী</type>
 			<type key="em" type="text" scope="core">পাঠ্য</type>
 			<type key="hc" type="h11">১২ ঘণ্টীয়া প্ৰণালী (০–১১)</type>
-			<type key="hc" type="h11" scope="core">১২ (০–১১) ১২ (০❰NDASH❱১১)</type>
+			<type key="hc" type="h11" scope="core">১২ (০–১১) ১২ (০ ১১)</type>
```
AEApple: Shouldn't this be replaced with an n-dash?

macchiati: Yes, my mistake!

macchiati: Good catch. Out of my meeting, so fixing now.

common/main/as.xml
```diff
@@ -925,13 +925,13 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<type key="em" type="emoji" scope="core">ইম’জী</type>
 			<type key="em" type="text" scope="core">পাঠ্য</type>
 			<type key="hc" type="h11">১২ ঘণ্টীয়া প্ৰণালী (০–১১)</type>
-			<type key="hc" type="h11" scope="core">১২ (০–১১) ১২ (০ ১১)</type>
```
macchiati: Note: the as.xml data is weird. G Translate gives it as:
<type key="hc" type="h11">12-hour system (0–11)</type>
<type key="hc" type="h11" scope="core">12 (0–11) 12 (0–11)</type>

So the vetters clearly didn't understand the meaning of core. But too late to fix for this release (since this may just be the tip of the iceberg).

macchiati:
There was a bad commit message. Will squash after reviews pass.

PR #4981, opened 2025-08-20
CLDR-17016 Add missing likely subtags

conradarcturus:
For some reason these languages weren't successfully imported from langtags.json.

Therefore, I found records of populations for all 4 languages and updated the script information, then re-generated likely subtags.

```
mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags
```

CLDR-17016

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/likelySubtags.xml
+++ b/common/supplemental/likelySubtags.xml
@@ -16,6 +16,9 @@ not be patched by hand, as any changes made in that fashion may be lost.
     <likelySubtags>
 		<likelySubtag from="aa" to="aa_Latn_ET"/>		<!--Afar‧?‧?	➡ Afar‧Latin‧Ethiopia-->
 		<likelySubtag from="ab" to="ab_Cyrl_GE"/>		<!--Abkhazian‧?‧?	➡ Abkhazian‧Cyrillic‧Georgia-->
+		<likelySubtag from="abq" to="abq_Cyrl_RU"/>		<!--Abaza‧?‧?	➡ Abaza‧Cyrillic‧Russia-->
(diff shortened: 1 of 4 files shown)


macchiati:
Looks good to me. Any reason this is still draft?

conradarcturus:
Thanks @macchiati ! I'm not sure why I only posted this as draft -- but I wrote it right before going on vacation so I probably just forgot to followup :p 

I'll rebase this & double check the regen scripts, and if that's good I will update the data (although let me know if you'd prefer to punt this to v49).

conradarcturus:
No issues rebasing, I'll merge this if I confirm its still valid in the v48 window.

AEApple:
@markusicu @richgillam - It looks like this missed getting merged. Are you okay with this late change being picked up into the next (alpha3 integration early next week) or would you prefer we defer this until CLDR 49?

richgillam:
> @markusicu @richgillam - It looks like this missed getting merged. Are you okay with this late change being picked up into the next (alpha3 integration early next week) or would you prefer we defer this until CLDR 49?

I'll defer to Markus here.  This looks like a pretty extensive change, and likely-subtags is used all over the place.  On the other hand, if it integrates cleanly, it might not actually be a big deal.

markusicu:
Looks like a small change for obscure languages. Fine with me.

conradarcturus:
Cool, thanks all for commenting. I'll merge the change now.

PR #4991, opened 2025-08-25
CLDR-18880 Add metazone variant overrides to match rearguard

robertbastian:
CLDR-18880

ALLOW_MANY_COMMITS=true


sffc (approved):

common/supplemental/metaZones.xml
```diff
@@ -187,7 +187,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<timezone type="Africa/Windhoek">
 				<usesMetazone to="1990-03-20 22:00" mzone="Africa_Southern"/>
 				<usesMetazone to="1994-03-20 22:00" from="1990-03-20 22:00" mzone="Africa_Central"/>
-				<usesMetazone to="2017-10-23 22:00" from="1994-03-20 22:00" mzone="Africa_Western"/>
+				<usesMetazone to="2017-10-23 22:00" from="1994-03-20 22:00" mzone="Africa_Western" stdOffset="+01" dstOffset="+02"/>
```
sffc, line 190: Question: Where did the Morocco changes go? You are changing only Namibia and Ireland.

robertbastian: morocco only started doing negative dst after 2017-10-23 22:00, which is why it got removed from the metazone

PR #5007, opened 2025-09-02
CLDR-18916 Unexpected date formats

btangmu:
-New tool ShowUnexpectedDateFormats.java to find the paths of interest

CLDR-18916

ALLOW_MANY_COMMITS=true


btangmu:
I've added the tool's output (after the 2nd commit, which includes all calendars, not only Gregorian), to the ticket

macchiati (approved):
This needs to be expanded to include any paths that are not in root; can you file a ticket for that (for v49)

btangmu:
> can you file a ticket for that (for v49)

I see that @AEApple has already done that: https://unicode-org.atlassian.net/browse/CLDR-18931

macchiati:
That is for a different ticket. See the comments.

PR #5013, opened 2025-09-03
CLDR-18920 Update modify_config.txt for en_001

AEApple:
CLDR-18920

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en_001.xml
+++ b/common/main/en_001.xml
@@ -181,6 +181,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<dateFormatItem id="Ehms" alt="ascii">↑↑↑</dateFormatItem>
 						<dateFormatItem id="EHms">↑↑↑</dateFormatItem>
 						<dateFormatItem id="GyMd">dd/MM/y GGGGG</dateFormatItem>
+						<dateFormatItem id="GyMEd">E, dd/MM/y G</dateFormatItem>
 						<dateFormatItem id="GyMMMd">d MMM y G</dateFormatItem>
 						<dateFormatItem id="GyMMMEd">E, d MMM y G</dateFormatItem>
 						<dateFormatItem id="M">LL</dateFormatItem>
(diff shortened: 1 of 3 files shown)


macchiati:
What we would normally do is run the CLDRModify tool with -k, as part of the same PR. I'm guessing that you don't have access to an IDE to do that; if so, I can jump in and do it.

BTW, I've been wondering whether we could run tools with github workspace; that might be worth investigating!

AEApple:
> What we would normally do is run the CLDRModify tool with -k, as part of the same PR. I'm guessing that you don't have access to an IDE to do that; if so, I can jump in and do it.
> 
> BTW, I've been wondering whether we could run tools with github workspace; that might be worth investigating!

I was going to reset up my IDE but I've been trying to figure out where to edit JAVA_HOME in the GitHub Codespace, but I haven't found the right answer yet..... I was planning to switch to a local IDE if I don't find the answer soon, but I'm unfortunately much more familiar with the internal tooling than GitHub.

It may be faster if you do this, and I'll keep digging around for the answer. I want to figure out how to set up and use the Codespaces as for my IDE. It SEEMS like it should be possible.

macchiati:
Ok, can do.

Setting up your IDE is not something you want to do under pressure!

AEApple:
> Ok, can do. Setting up your IDE is not something you want to do under pressure!
> […](#)

Makes sense, but I will add instructions to our docs once I figure out how to set up a cloud workspace.

macchiati:
I generated the two files based on the modify_config. **Please spot check them for completeness.**

NOTE: I should have remembered that currently we need to modify the annotations and main directories separated, so I ended up doing two passes, modifying the modify_config for each.

macchiati (approved):
Stamping so that you can merge once you're reviewed my changes.

common/main/en_001.xml
```diff
@@ -435,7 +437,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 							<greatestDifference id="M">d MMM – d MMM</greatestDifference>
 						</intervalFormatItem>
 						<intervalFormatItem id="MMMEd">
-							<greatestDifference id="d">E d – E d MMM</greatestDifference>
```
AEApple, line 438: @macchiati - I think I probably shouldn't have changed this one. Interestingly how both en and en-GB have the longer forms. WDYT?

macchiati: Putting in values, this looks fine to me. So I'd leave this for now. 

Tue. 10 - Thu. 12 April

I think next release we need to take a good hard look at all the English interval formats, to make sure that they are as short as possible while still being customary.

macchiati: By the way, I mistakenly checked this off as done in the BRS so if you want to not merge this version then we should take that done off.

AEApple:
Went ahead and merged. We can update that one item if people have concerns about it later. I think we need to have a discussion about the formatting of the very small difference intervals.

common/main/en_001.xml
AEApple: Oh, I missed your comment. Let me just revert that one item back.

PR #5014, opened 2025-09-03
CLDR-18852 Gujr transform

srl295:
- remove log known issue
- rename ExemplarUtilities.getScript() to getLikelyScript() (and use it)
- clarify the transliterator used

And the actual fix:
- add U+0AF0 to the filter at the top of the Gujarati-Latin transform.

CLDR-18852

ALLOW_MANY_COMMITS=true


srl295:
Note, also filed https://unicode-org.atlassian.net/browse/CLDR-18937 because the way we structure these transform rules may need improvement

PR #5015, opened 2025-09-03
CLDR-18462 Update the test to disable the logKnownIssue

macchiati:
CLDR-18462

We needed new code. I also added some instructions for what to do with new cases. See how that looks.

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
```diff
@@ -683,51 +698,66 @@ public void testCompleteness() {
...
+
+            for (Pair<String, String> pathAndValue : listXmlEmoji) {
+                XPathParts parts = XPathParts.getFrozenInstance(pathAndValue.getFirst());
+                if (!"annotation".equals(parts.getElement(-1))) {
+                    continue;
+                }
+                String cp = parts.getAttributeValue(-1, "cp");
+                boolean isTts = parts.getAttributeValue(-1, "type") != null;
+                UnicodeSet set = isTts ? namesFound : searchKeywordsFound;
+                set.add(cp);
```
btangmu: "Attempt to modify frozen object" here at line 718? Strange since both namesFound and searchKeywordsFound are initialized with `new UnicodeSet()`; `freeze()` isn't called until lines 720 and 721 below

btangmu: Oh, I see, there's now an outer loop, so the error occurs the second time through! Either `freeze()` needs moving outside the outer loop, or else namesFound and searchKeywordsFound need to be initialized to new unfrozen sets inside the outer loop

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
```diff
@@ -683,51 +698,66 @@ public void testCompleteness() {
...
+                            .removeAll(TEMPORARY_SKIP_COMPOUNDS)
+                            .toPattern(false));
+            UnicodeSet onlyAllowedBecauseOfTEMPORARY_SKIP_COMPOUNDS =
+                    new UnicodeSet(missingNames)
+                            .addAll(missingKeywords)
+                            .retainAll(TEMPORARY_SKIP_COMPOUNDS)
+                            .freeze();
+            if (!onlyAllowedBecauseOfTEMPORARY_SKIP_COMPOUNDS.isEmpty()) {
+                logKnownIssue(
+                        "CLDR-18462",
```
srl295: not ready to close issue if this is still referenced

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
macchiati: thanks, my oversight. (I'm surprised, because it doesn't fail locally, but definitely a problem.)

macchiati (commented):
I found that a problem in my code was masking a test failure. Cleaned it up a bit, but I still need to add the code for deriving names for the missing ones.

srl295 (dismissed):
Changes look good

macchiati:
Thanks. I'm going to put this on the back burner for a bit and focus on the stuff needed for alpha. (It will take a bit of time to fix the derivations for the new emoji.)

macchiati:
For review, best to 

1. check the code in the 3 files at the end
2. just spot-check en.xml for results. They should all be of the form 
"people/men/women wrestling/with bunny ears: <0 or more skin tones>

Note that at least with Eclipse the cp="..." can appear wrong, since nobody's fonts have the characters yet.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAnnotations.java
```diff
@@ -679,15 +694,88 @@ public void testSymbols() {
...
+                List.of("en.xml")) { // for testing, can add others like "de.xml", "fr.xml"
+            UnicodeSet namesFound = new UnicodeSet();
+            UnicodeSet searchKeywordsFound = new UnicodeSet();
+            fillNamesAndSearchKeywords(
+                    file, namesFound, searchKeywordsFound); // freezes the results
+
+            warnln(
+                    Joiner.on('\t')
+                            .join(
+                                    "FYI, RGI:",
```
btangmu, line 709: "RGI: Recommended for General Interchange" -- I had to do a web search to find that

tools/cldr-code/src/main/java/org/unicode/cldr/util/EmojiConstants.java
```diff
@@ -83,6 +83,17 @@ public static String getEmojiFromSubdivisionCodes(String string) {
                     .add(MALE_SIGN)
                     .add(FEMALE_SIGN)
                     .freeze();
+    public static final UnicodeSet REM_PERSON_SKIP_SET =
+            new UnicodeSet(REM_SKIP_SET)
```
btangmu, line 87: I'm not sure what "REM" stands for; guessing "remainder" since "rem" sometimes stands for that in related code. Some comments might help to clarify what these items share in common

btangmu (approved):
I can only rubber-stamp, for lack of knowledge about emoji combinations

macchiati:
Thanks Tom,

I'll file a ticket for improving the documentation

macchiati:
https://unicode-org.atlassian.net/browse/CLDR-18949 for continuation

PR #5017, opened 2025-09-04
CLDR-18934 fix hv hz in German

macchiati:
German hv and hz (gregorian) are now patterned after precedent of h, H, and adding v.
I left zh alone, because it looks like an informed choice by the zh vetters.

See the ticket for more information.

CLDR-18934

ALLOW_MANY_COMMITS=true


richgillam (approved):
LOKTM.

PR #5018, opened 2025-09-05
CLDR-18938 BRS Likely subtags

macchiati:
CLDR-18938

Only names (in xml comments)

ALLOW_MANY_COMMITS=true


AEApple:
It looks like it's just updating the names of some locales? Is that expected?

AEApple (approved):

PR #5019, opened 2025-09-05
CLDR-13542 Fix Swiss grouping separator

macchiati:
The goal is to change the decimal separator from ’ to ASCII ' for the Swiss locales: to see 

## 1'234,00
instead of
## 1’234,00

Note: gsw is Swiss German, and wal is a kind of Swiss German

CLDR-13542

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/de_LI.xml
+++ b/common/main/de_LI.xml
@@ -31,7 +31,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 	<numbers>
 		<symbols numberSystem="latn">
 			<decimal>.</decimal>
-			<group>’</group>
+			<group>'</group>
 		</symbols>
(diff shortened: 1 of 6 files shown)


AEApple (dismissed):

AEApple:
Can we file a ticket so this doesn't happen in the future?

macchiati:
I had to fix DAIP also. It was doing a conversion from ' to ’, and we should leave it alone.

macchiati:
Had to squash to fix commit message error.

AEApple:
Would this get caught by a golden data test in ICU? If not, should we add one?

macchiati:
I think these might be flagged in the survey tool, if so we'd see any
changes.

PR #5024, opened 2025-09-08
CLDR-18837 Follow-up per comments in last-merged PR

btangmu:
-Remove debugging code

-Use APPLICATION_ZIP (locally defined as application/zip) instead of MediaType.APPLICATION_OCTET_STREAM

CLDR-18837

ALLOW_MANY_COMMITS=true


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/GenerateVxml.java
```diff
@@ -20,6 +20,13 @@
 @Path("/vxml")
 @Tag(name = "Generate VXML", description = "APIs for Survey Tool VXML (Vetted XML) generation")
 public class GenerateVxml {
+
+    /**
+     * Substitute for not-found MediaType.APPLICATION_ZIP. An alternative would be
```
srl295, line 25: doesn't need to be in MediaType

PR #4961, opened 2025-08-12
CLDR-18889 v48 BRS generate algorithmic

srl295:
CLDR-18889

- [X] That’s all folks.

- regenerate GenerateLocaleIDTestData
- I would update the docs to recommend running this tool after generating algorithmic, but the docs  aren't migrated over <https://unicode-org.atlassian.net/browse/CLDR-18606?focusedCommentId=186098> **update** I did update the docs. 
- TestLocale doesn't look like it is a useful test, filed https://unicode-org.atlassian.net/browse/CLDR-18890 - but until then, I updated the test to suggest how to correct the problem.
- adds a `@CDATA` tag to rbnf - CLDR-8909 - noting that the element is expected to be a CDATA element

ALLOW_MANY_COMMITS=true

PR #5027, opened 2025-09-08
CLDR-13542 Fix remaining apostrophe groupings

macchiati:
CLDR-13542

When I looked into this, I found that it was DAIP that was automatically changing the grouping separator. So aside from CH, the other cases (DDL locales) were probably originally entered as straight apostrophes.

In addition, somehow the following removed the grouping separator from de_CH, so this restores it (with the straight/ASCII) version.

- [v48 BRS generate algorithmic](https://github.com/unicode-org/cldr/pull/4961)
- https://github.com/unicode-org/cldr/commit/7fa7166e5fddf418739f35c06ee41eb16aeab5dd#diff-77acfe3eceed0aced87a3ee1334bb6499a63279f711f8733362d1e602b71d138L4932 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/de_CH.xml
+++ b/common/main/de_CH.xml
@@ -41,6 +41,9 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<exemplarCharacters type="punctuation-auxiliary" draft="contributed">↑↑↑</exemplarCharacters>
 	</characters>
 	<numbers>
+		<symbols numberSystem="latn">
+			<group>'</group>
+		</symbols>
 		<currencies>
 			<currency type="BYN">
 				<displayName>Weissrussischer Rubel</displayName>
--- a/common/main/lmo.xml
+++ b/common/main/lmo.xml
@@ -129,7 +129,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		</otherNumberingSystems>
 		<symbols numberSystem="latn">
 			<decimal draft="contributed">,</decimal>
-			<group draft="contributed">’</group>
+			<group draft="contributed">'</group>
 			<percentSign draft="contributed">↑↑↑</percentSign>
 			<plusSign draft="contributed">↑↑↑</plusSign>
 			<minusSign draft="contributed">↑↑↑</minusSign>
--- a/common/main/tn.xml
+++ b/common/main/tn.xml
@@ -655,7 +655,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<minimumGroupingDigits draft="contributed">↑↑↑</minimumGroupingDigits>
 		<symbols numberSystem="latn">
 			<decimal>↑↑↑</decimal>
-			<group>’</group>
+			<group>'</group>
 			<percentSign>↑↑↑</percentSign>
 			<plusSign>↑↑↑</plusSign>
 			<minusSign>↑↑↑</minusSign>
--- a/common/main/wal.xml
+++ b/common/main/wal.xml
@@ -346,7 +346,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<traditional>ethi</traditional>
 		</otherNumberingSystems>
 		<symbols numberSystem="latn">
-			<group draft="unconfirmed">’</group>
+			<group draft="unconfirmed">'</group>
 		</symbols>
 		<currencyFormats numberSystem="latn">
 			<currencyFormatLength>


AEApple:
wal wasn't from DAIP surprisingly: https://github.com/unicode-org/cldr/commit/ac78fb369febe2fece5a2b51cc5bb516804cfecb

macchiati:
> wal wasn't from DAIP surprisingly: [ac78fb3](https://github.com/unicode-org/cldr/commit/ac78fb369febe2fece5a2b51cc5bb516804cfecb)

Hard to tell; that commit was from applying CLDRModify, probably including DAIP.
And if not, using ’ for consistency with others (which had DAIP).

AEApple:
> > wal wasn't from DAIP surprisingly: [ac78fb3](https://github.com/unicode-org/cldr/commit/ac78fb369febe2fece5a2b51cc5bb516804cfecb)
> 
> Hard to tell; that commit was from applying CLDRModify, probably including DAIP. And if not, using ’ for consistency with others (which had DAIP).

Okay, but it didn't change from a straight apostrophy. It made the original group separator an alt:

<group draft="unconfirmed">ወ</group

macchiati:
> Okay, but it didn't change from a straight apostrophy. It made the original group separator an alt:

I'm not exactly sure what happened, but I think the ASCII apostrophe is safer; it can always be changed later.

AEApple:
> > Okay, but it didn't change from a straight apostrophy. It made the original group separator an alt:
> 
> I'm not exactly sure what happened, but I think the ASCII apostrophe is safer; it can always be changed later.

Okay, makes sense

DarioViva42:
@macchiati  Why was the ’ (U+2019) changed to a straight apostrophe (U+0027) ?
As a Swiss-German myself I think the situation now is rather strange.
Both the German Wikipedia article (https://de.wikipedia.org/wiki/Dezimaltrennzeichen#Schweiz) and the English Wikipedia article (https://en.wikipedia.org/wiki/Decimal_separator#Examples_of_use) would approve of this change.
But java.text.NumberFormat still uses U+2019 as of java 25.

```java
java.text.DecimalFormatSymbols javaSymbols = java.text.DecimalFormatSymbols
				.getInstance(Locale.of("de", "CH"));
System.out.println(Character.getName(javaSymbols.getMonetaryGroupingSeparator())); // APOSTROPHE
System.out.println(Character.getName(javaSymbols.getGroupingSeparator())); // APOSTROPHE

com.ibm.icu.text.DecimalFormatSymbols icuSymbols = com.ibm.icu.text.DecimalFormatSymbols
				.getInstance(Locale.of("de", "CH"));
System.out.println(Character.getName(icuSymbols.getMonetaryGroupingSeparator())); // RIGHT SINGLE QUOTATION MARK
System.out.println(Character.getName(icuSymbols.getGroupingSeparator())); // RIGHT SINGLE QUOTATION MARK
```
This leads different behaviour for the NumberFormat from the jdk and NumberFormatter from icu.
```java
System.out.println(NumberFormat.getInstance(Locale.of("de", "CH")).format(1000)); // 1’000
System.out.println(NumberFormatter.withLocale(Locale.of("de", "CH")).format(1000)); // 1'000
```
Now I just want to make sure that the numbers in our applications are formatted correctly. Admittedly the difference is pretty minute, but at the same time it’s not completly irrelevant because otherwise you wouldn’t have bothered to make this change.

TLDR: Would using icu over jdk default tools now lead to a more correct behaviour when formatting numbers?

DarioViva42:
@macchiati The Swiss Federal Chancellery (Schweizerische Bundeskanzlei) dictates the usage of the non breaking space instead of an apostrophe for official federal texts. 

See: [Schreibweisungen](https://www.bk.admin.ch/bk/de/home/dokumentation/cd-bund/terminologie/schreibweisungen.html) Page 79 Zahlangaben Chapter 5.1.2 Schreibweise von Zahlen in Ziffern. [direct Link to PDF](https://www.bk.admin.ch/dam/bk/de/dokumente/sprachdienste/sprachdienst_de/schreibweisungen.pdf.download.pdf/schreibweisungen.pdf)

> Zur Gliederung wird ein Festabstand → 251 ver-
wendet, damit die Zahlengruppen beim Zeilen-
sprung nicht auseinandergerissen werden.
> 
> Die früher gebräuchliche Schreibung mit Apos-
troph → 209 sollte nicht mehr angewendet wer-
den, weil sie für die Drucklegung wieder rück-
gängig gemacht werden muss. Nicht korrekt ist
die Gliederung mit Punkten oder Kommas.

English Translation:
> A fixed space → 251 is used for structuring
so that the groups of numbers are not separated
when line breaks occur.
>
> The previously common spelling with an apostrophe
→ 209 should no longer be used because it has to
be reversed for printing. It is not correct to use
periods or commas for structuring.

Also important to notice is, that in Switzerland it’s very common to only apply thousands grouping when a number has more than four digits. The Chancellery writes the following:

> Besteht eine Zahl aus vier Ziffern, so wird die
erste nicht abgesetzt, sondern eine Vierergruppe
gebildet.
>
> Zahlen, die aus mehr als vier Ziffern bestehen,
werden von der Endziffer aus in Dreiergruppen
zerlegt.

English Translation:

> If a number consists of four digits, the
first is not separated, but a group of four is
formed.
>
> Numbers consisting of more than four digits
are divided into groups of three, starting from 
the last digit.

This leads to 1000 and 10 000.

Honestly I am not sure how strongly everyone else should adapt the style guides from the federal chancellery. But everyone from the Bund (the government) should apply them.

TLDR: Maybe using U+00A0 instead of U+0027 makes more sense.

srl295:
> @macchiati Why was the ’ (U+2019) 

I linked back here from https://unicode-org.atlassian.net/browse/CLDR-13542 but it would be better to discuss this on the issue or on a new issue rather than on a merged PR.

PR #5028, opened 2025-09-09
CLDR-18924 Update Hawaii_Aleutian in sv

robertbastian:
CLDR-18924

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/sv.xml
+++ b/common/main/sv.xml
@@ -6220,9 +6220,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			</metazone>
 			<metazone type="Hawaii_Aleutian">
 				<long>
-					<generic>Honolulutid</generic>
-					<standard>Honolulu, normaltid</standard>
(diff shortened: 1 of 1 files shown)


common/main/sv.xml
```diff
@@ -6220,9 +6220,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			</metazone>
 			<metazone type="Hawaii_Aleutian">
 				<long>
-					<generic>Honolulutid</generic>
-					<standard>Honolulu, normaltid</standard>
-					<daylight>Honolulu, sommartid</daylight>
+					<generic>aleutisk tid</generic>
+					<standard>aleutisk, normaltid</standard>
```
AEApple: It looks like commas are only used when it is a location name where the location is capitalized. Not sure why, but I think we should match the existing patterns.

AEApple (commented):
LGTM! Thanks for removing the commas!

PR #5029, opened 2025-09-09
CLDR-18253 Remove eras in en_JP and update Japanese calendar to match an existing distribution

AEApple:
CLDR-18253

- Removes eras from en* so they inherit directly from root
- Remove vetter data (we should make sure the calendar is in comprehensive next cycle
- Update data to match [an existing distribution of en_JP and Japanese calendars](https://github.com/apple-oss-distributions/ICU/blob/rel/ICU-76104/icu/icu4c/source/data/locales/en_JP.txt) for availableFormats and intervalFormats
- 

ALLOW_MANY_COMMITS=true


AEApple:
Waiting to see how this does on tests

macchiati:
Looks like the build failures are due to edits that leave elements unterminated:

file:///home/runner/work/cldr/cldr/common/main/en.xml; lineNumber: 3480; columnNumber: 7; The element type "calendar" must be terminated by the matching end-tag "</calendar>

macchiati:
is this moot for v48?

AEApple:
Yes

AEApple:
Yes, this is moot for 48. Will revisit after we branch for 49. I would like to update the instructions about running the console check at comprehensive since it is not described here: https://cldr.unicode.org/development/adding-locales#regional-locales

Often when regional locales are added they are added in xml.

PR #5030, opened 2025-09-09
CLDR-18253 Remove unnecessary data and revert Japanese eras to inherit from root

AEApple:
[CLDR-18253](https://unicode-org.atlassian.net/browse/CLDR-18253)

- Remove Japanese eras in en and all child locales
- Remove Date & Time Combination Formats in en_JP (will inherit en formats)
- Remove Japanese calendar for child locales where it is unnecessary
  - We should update it so that for calendars that are not necessary for most child locales they don't need to vet (make those calendars comprehensive in CLDR 50 

ALLOW_MANY_COMMITS=true

[CLDR-18253]: https://unicode-org.atlassian.net/browse/CLDR-18253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


common/main/en_AU.xml
```diff
@@ -2913,596 +2913,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
-						<era type="233">↑↑↑</era>
-						<era type="234">↑↑↑</era>
-						<era type="235">↑↑↑</era>
-						<era type="236">↑↑↑</era>
-					</eraNarrow>
-				</eras>
-				<dateFormats>
-					<dateFormatLength type="full">
-						<dateFormat>
-							<pattern>EEEE, d MMMM, y G</pattern>
```
robertbastian, line 3400: is this diff intended?

common/main/en_IN.xml
```diff
@@ -2708,594 +2708,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
-						</dateFormat>
-					</dateFormatLength>
-					<dateFormatLength type="medium">
-						<dateFormat>
-							<pattern>↑↑↑</pattern>
-						</dateFormat>
-					</dateFormatLength>
-					<dateFormatLength type="short">
-						<dateFormat>
-							<pattern>d/M/y GGGGG</pattern>
```
robertbastian, line 3210: and this

AEApple:
Okay, I went back and checked this:
- en_GB did not have the Japanese calendar in CLDR 46 (last submission cycle)
- en_CA did not have the Japanese calendar in CLDR 46
- en_AU was complete inheritance - see file https://github.com/unicode-org/cldr/blob/maint/maint-46/common/main/en_AU.xml#L2811
- en_IN did not have the Japanese calendar in CLDR 46

Therefore we should go ahead and remove them for now.

common/main/en_IN.xml
AEApple: We can remove. This was not in the previous CLDR 46 or 47 releases.

common/main/en_AU.xml
AEApple: We can remove, en_AU inherited in the previous CLDR 46 and 47 releases which is what will happen.

PR #5031, opened 2025-09-09
CLDR-18962 BRS v48 derived annotations

srl295:
CLDR-18962


btangmu (approved):
It looks like all the changes are just in combining two comments at the top of each xml file. Is that right? If so, LGTM

macchiati (approved):

srl295:
> It looks like all the changes are just in combining two comments at the top of each xml file. Is that right? If so, LGTM

Sorry I missed that. 

Well let's merge it so there aren't diffs in the future.

macchiati:
Right, the files are normalized when written out (by CLDRModify, but really by anything using CLDRFile), which merges adjacent comments.

PR #5038, opened 2025-09-11
CLDR-18889 v48 BRS: re-run algo de_CH

srl295:
- revert to generated de_CH per CLDR-18957
- generated against c0beb8a5487b293df30c47735bd43fc14da8deb7 which was the base of #4961
- tagged this as `release-48-alpha2b`

CLDR-18889

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/de_CH.xml
+++ b/common/main/de_CH.xml
@@ -4926,7 +4926,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 		<minimumGroupingDigits>↑↑↑</minimumGroupingDigits>
 		<symbols numberSystem="latn">
 			<decimal>.</decimal>
-			<group>’</group>
+			<group>'</group>
 			<percentSign>↑↑↑</percentSign>
 			<plusSign>↑↑↑</plusSign>
 			<minusSign>↑↑↑</minusSign>


srl295:
FWIW CopyMainToSubdivisions was the result of a warning. This just meant that common/subdivisions was out of date with what was in ST.  We never know which is correct at unit test time (there's a separate ticket on this and why I think the data shouldn't be duplicated) however in this case `gu` was clearly updated from survey tool.

```
Warning:  (TestSubdivisions.java:276)  Warning: gu: common/main differs from common/subdivisions: //ldml/localeDisplayNames/subdivisions/subdivision[@type="gbsct"] - સ્કૉટલૅન્ડ vs સ્કોટલેન્ડ - run CopyMainToSubdivisions

Warning:  (TestSubdivisions.java:276)  Warning: gu: common/main differs from common/subdivisions: //ldml/localeDisplayNames/subdivisions/subdivision[@type="gbeng"] - ઈંગ્લૅન્ડ vs ઈંગ્લેન્ડ - run CopyMainToSubdivisions

Error:  (TestSubdivisions.java:291)  Error: de_CH: common/main has path not in common/subdivisions: //ldml/localeDisplayNames/subdivisions/subdivision[@type="gbwls"] - run CopyMainToSubdivisions

Error:  (TestSubdivisions.java:291)  Error: de_CH: common/main has path not in common/subdivisions: //ldml/localeDisplayNames/subdivisions/subdivision[@type="gbeng"] - run CopyMainToSubdivisions

Error:  (TestSubdivisions.java:291)  Error: de_CH: common/main has path not in common/subdivisions: //ldml/localeDisplayNames/subdivisions/subdivision[@type="gbsct"] - run CopyMainToSubdivisions
```

common/main/de_CH.xml
```diff
@@ -39,62 +1245,10028 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
 	<numbers>
+		<defaultNumberingSystem>↑↑↑</defaultNumberingSystem>
+		<otherNumberingSystems>
+			<native>↑↑↑</native>
+		</otherNumberingSystems>
+		<minimumGroupingDigits>↑↑↑</minimumGroupingDigits>
 		<symbols numberSystem="latn">
-			<group>'</group>
+			<decimal>.</decimal>
+			<group>’</group>
```
macchiati: This is a reversion, and needs to be changed to the ASCII '

macchiati: The ICU golden-data test failure for ' needs to be fixed to change ’ to '.

srl295: I actually ended up already doing that, because I had overwritten the de_CH override here

srl295:
@macchiati commit [f037322](https://github.com/unicode-org/cldr/pull/5038/commits/f037322e3b5891502deb88028f6e8e85714aba14) should revert the reversion PTAL

AEApple:
What is the impact if it doesn't have explicit inheritance in the algorithmically generated locales? Wouldn't it just inherit the same value either way?

macchiati:
The up-arrows are kinda superfluous, given that we don't have any de-CH vetters. But also harmless; they'll disappear in production data.

PR #5041, opened 2025-09-12
CLDR-18985 Fix Monrovia metazone date

robertbastian:
CLDR-18985

ALLOW_MANY_COMMITS=true


robertbastian:
So do we have to wait until Wednesday to have the TC formally accept the ticket?

AEApple:
> So do we have to wait until Wednesday to have the TC formally accept the ticket?

Just send an email to the TC asking for approval and as long as no one raises any concerns we can just accept and merge on Monday before the alpha3 integration.

yumaoka:
@robertbastian I will run full ICU time zone test with this change. After it's confirmed, I will approve this PR.

PR #5042, opened 2025-09-12
CLDR-18955 Fix inconsistency in new GyMEd available formats

AEApple:
CLDR-18955 

- This fixes German to be consistent with related available and interval formats
- We really should check the rest of the locales to see if they also have issues

ALLOW_MANY_COMMITS=true


AEApple:
> Ok but it would be better to systematically fix this

Yes, but I'm OOO and don't have time for it. If you want to find a way to systematically fix this, please go ahead.

PR #5048, opened 2025-09-16
CLDR-18369 Update root era names

robertbastian:
CLDR-18369

Now that we've aligned all the codes with the correct abbreviations, we should also align the names in root. This finally gets rid of ERA0 and ERA1 for Ethiopic. 

ALLOW_MANY_COMMITS=true


common/main/root.xml
```diff
@@ -771,11 +771,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</dayPeriods>
 				<eras>
 					<eraNames>
-						<alias source="locale" path="../eraAbbr"/>
+						<era type="0">Amete Alem</era>
+						<era type="1">Amete Mihret</era>
```
sffc: Issue: Root should have the fallback

common/main/en.xml
```diff
@@ -2137,10 +2137,10 @@ annotations.
 			<calendar type="coptic">
 				<eras>
 					<eraNames>
-						<era type="1">Anno Martyrum</era>
```
sffc, line 2140: Issue: Long names should go in en.xml

common/main/root.xml
```diff
@@ -663,7 +663,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</dayPeriods>
 				<eras>
 					<eraNames>
-						<alias source="locale" path="../eraAbbr"/>
```
macchiati, line 666: Cannot add or remove aliases without TC approval. Affects all of inheritance.

common/main/en.xml
robertbastian: these are locale agnostic names, they're Latin

AEApple (changes requested):
We are past data freeze. This will need approval from both the CLDR and ICU TCs at this point in the release or you can wait until CLDR 49.

robertbastian:
So the first CLDR-JSON we could use to test with ICU4X was released after the data freeze, so now I have to go begging to two TCs for any issue we find?

AEApple:
> So the first CLDR-JSON we could use to test with ICU4X was released after the data freeze, so now I have to go begging to two TCs for any issue we find?

I think this means that we need to refine the schedule to consider ICU4X. Historically ICU4X's release timelines were not aligned with the CLDR or ICU timelines but were independent.

robertbastian:
> Historically ICU4X's release timelines were not aligned with the CLDR or ICU timelines but were independent.

ICU4X has tested every CLDR release since I've been on the project as soon as a CLDR-JSON was available. This has always been way later than ICU could start integration, which I point out every time, but I don't think it has ever been after data freeze.

AEApple:
> > Historically ICU4X's release timelines were not aligned with the CLDR or ICU timelines but were independent.
> 
> ICU4X has tested every CLDR release since I've been on the project as soon as a CLDR-JSON was available. This has always been way later than ICU could start integration, which I point out every time, but I don't think it has ever been after data freeze.

It has always been after data freeze at least according to the CLDR BRS. In general we only start generating CLDR JSON data after alpha2 and sometimes after public alpha but always before public beta.

It may have been happening but since it isn't currently reflected that way in the BRS checklist, we should make sure that is updated so that you can test earlier. Otherwise it's likely to get missed if we are running late for any reason.

macchiati:
Moved to draft to avoid getting inadvertently merged before we settle remaining issues.

AEApple:
Saka change was discussed in https://unicode-org.atlassian.net/browse/CLDR-18487

common/main/root.xml
```diff
@@ -809,7 +809,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<alias source="locale" path="../eraAbbr"/>
 					</eraNames>
 					<eraAbbr>
-						<era type="0">ERA0</era>
+						<alias source="locale" path="../../../calendar[@type='ethiopic']/eras/eraAbbr/era[@type='0']"/>
```
macchiati: I looked this over, and I think it is ok if we change to:
```suggestion
						<alias source="locale" path="../../../calendar[@type='ethiopic']/eras/eraAbbr"/>
```

The aliases take the asked-for path, plus the location of the alias itself, in order to construct the new path. You can see that in the other cases. That being said, I haven't tested this out.

robertbastian: We don't want to pull in both eras, just the one with type=0

macchiati: That's not the way that they work: https://unicode.org/reports/tr35/tr35.html#Alias_Elements 
That would map any missing item (eg type='1') to the 0 value.

However, I see that you've removed this, so it's moot.

macchiati (approved):
You're now removing all the values, even those that change script. I understand that they are probably incorrect in that they include "0", but the Latin abbreviation is probably no better. However, to get this moving I'm approving.

gclapps0612-cmd:
Need a better under standing on whats going on

common/main/root.xml
robertbastian: I removed it from this PR so it can get merged, but we need to figure out some mechanism to keep these era names in sync.

We might be able to just remove translations for `ethipic-amete-alem` altogether if patterns also match `ethiopic`.

macchiati: Agreed, we do that for the Islamic variants!

robertbastian:
ping @AEApple

robertbastian:
autosubmit please

PR #5052, opened 2025-09-16
CLDR-18900 Add Hntl to scriptData

macchiati:
CLDR-18900

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-modifications.md
+++ b/docs/ldml/tr35-modifications.md
@@ -55,11 +55,14 @@ The LDML specification is divided into the following parts:
 
 **Changes in LDML Version 48 (Differences from Version 47)**
 
+### Locale Identifiers
+* [Special Script Codes](tr35.html#special-script-codes) The `Hntl` compound script was added. (This is also reflected in the `<scriptData>` elements in supplementalData.xml.)
+  
 ### DateTime formats
 
-* In [Element dateTimeFormat](tr35-dates.md#dateTimeFormat), added a new type `relative` and updated the
+* [Element dateTimeFormat](tr35-dates.md#dateTimeFormat) Added a new type `relative` and updated the
 guidelines on how to use the different `dateTimeFormat` types.
-* In [Using Time Zone Names](tr35-dates.md#Using_Time_Zone_Names), the "specific location format" was removed
+* [Using Time Zone Names](tr35-dates.md#Using_Time_Zone_Names) The "specific location format" was removed.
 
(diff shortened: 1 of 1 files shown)


AEApple:
When do we update https://github.com/unicode-org/cldr/blob/main/common/properties/scriptMetadata.txt#L63 ?

macchiati:
We don't add any of the ones in <scriptData> to scriptMetaData. The <scriptData ones are special, and no Unicode characters have them natively. There is a bit of documentation in https://unicode.org/reports/tr35/tr35.html#special-script-codes

AEApple:
Why are Hani and Hanb in the list then? https://github.com/unicode-org/cldr/blob/main/common/properties/scriptMetadata.txt#L60

srl295:
See also https://unicode-org.atlassian.net/browse/CLDR-18950 — where I proposed to add all script codes (that aren't Unicode scripts) to scriptData.

docs/ldml/tr35-modifications.md
```diff
@@ -55,11 +55,14 @@ The LDML specification is divided into the following parts:
 
 **Changes in LDML Version 48 (Differences from Version 47)**
 
+### Locale Identifiers
+* [Special Script Codes](tr35.html#special-script-codes) The `Hntl` compound script was added. (This is also reflected in the `<scriptData>` elements in supplementalData.xml.)
```
markusicu, line 59: The element is called scriptVariant, not scriptData.

macchiati: It is called `<scriptData>`. See https://github.com/unicode-org/cldr/blob/main/common/supplemental/supplementalData.xml#L2189

AEApple:
I think I need to spend a lot of time reading the spec and asking questions about the spec. Do you think we could have a 15 min overview of each of the major sections of the spec once a month over the next few months (and record them) since there are a lot of newer TC members and I'm sure I'm not the only one with a lot of questions about the spec and how it connects to CLDR data as different BRS tasks and CLDR tickets come up :)

docs/ldml/tr35-modifications.md
markusicu: > It is called `<scriptData>`.

Aha, that's the _outer_ element. The element that actually carries the data is scriptVariant, and that's visible in the PR diffs. I guess scriptData is not wrong then.

macchiati:
sure, sounds good.

PR #5053, opened 2025-09-16
CLDR-18369 Remove "incarnation" translations of Amete eras

robertbastian:
CLDR-18369

ALLOW_MANY_COMMITS=true


sffc (approved):

AEApple (changes requested):
It is past data freeze. This needs TC approval from both the CLDR and ICU TCs at this stage.

AEApple (approved):

robertbastian:
This should not have been merged without https://github.com/unicode-org/cldr/pull/5048

PR #5057, opened 2025-09-17
CLDR-18370 Use onlyAvailableFormats patterns for skeletons in datetime test data

echeran:
CLDR-18370

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/testData/datetime/datetime.json
+++ b/common/testData/datetime/datetime.json
@@ -1839,7 +1839,7 @@
     "calendar": "buddhist",
     "locale": "th-TH",
     "input": "2000-01-01T00:00Z[Etc/GMT]",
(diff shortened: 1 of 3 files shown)


sffc:
Thank you!

This change should definitely land in CLDR testData since it reflects the current CLDR data and the current CLDR specification. I had proposed landing this change in ICU (ICU-23209), but the ICU-TC preferred to wait until CLDR-14993 was resolved first.

As explained in CLDR-18881, these patterns in Thai and Japanese occur only in the standard patterns, not in the availableFormats patterns.

I commented in CLDR-14993 that we might want to investigate whether we can restore the Japanese Kanji behavior in skeleton patterns.

macchiati:
Note: this is only working because ICU isn't using the test file. We have to have develop a solution for when ICU does.

PR #5058, opened 2025-09-17
CLDR-18996 fix GenerateLanguageContainment

srl295:
CLDR-18996

This is just fixes to the tool - data change will be a separate PR.

- fix a less-than-helpful message on the `DtdData.DtdComparator` to tell what was being compared
- convert over the markdown page for GenerateLanguageContainment
- use `TempPrintWriter` (including an additional signature thereto) to prevent writing broken files
- document `TempPrintWriter.dontReplaceFile()`
- fix in `CLDRFile`: includes a fix for CLDR-18997 "CLDRFile shouldn't add extra paths to non-ldml DTDs (such as supplemental)"

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
@@ -1023,8 +1023,16 @@ public class DtdData extends XMLFileReader.SimpleHandler {
             String baseA = a.getElement(0);
             String baseB = b.getElement(0);
             if (!ROOT.name.equals(baseA) || !ROOT.name.equals(baseB)) {
+                final XPathParts oddity = (ROOT.name.equals(baseA) ? b : a);
+                final String baseOddity = oddity.getElement(0);
+                final String elementOddity = oddity.getElement(-1);
                 throw new IllegalArgumentException(
-                        "Comparing different DTDs: " + ROOT.name + ", " + baseA + ", " + baseB);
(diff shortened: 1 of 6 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
```diff
@@ -1023,8 +1023,16 @@ public int xpathComparator(XPathParts a, XPathParts b) {
...
+                final String baseOddity = oddity.getElement(0);
+                final String elementOddity = oddity.getElement(-1);
                 throw new IllegalArgumentException(
-                        "Comparing different DTDs: " + ROOT.name + ", " + baseA + ", " + baseB);
+                        "Comparing different DTDs: This comparator is for DTD "
+                                + ROOT.name
+                                + ", but attempted compare with DTD "
+                                + baseOddity
+                                + " in: "
+                                + oddity.toString());
```
srl295, line 1035: Example message:

> `Caused by: java.lang.IllegalArgumentException: Comparing different DTDs: This comparator is for DTD supplementalData, but attempted compare with DTD ldml in: //ldml/dates/timeZoneNames/metazone[@type="Uruguay"]/long/generic`

macchiati:
I think it uses it as a quick way to get it to sort last

srl295:
> I think it uses it as a quick way to get it to sort last
> 
>

The omega? Ok, this changes it to mul

PR #5060, opened 2025-09-18
CLDR-18956 cldr-json: RBNF new format

srl295:
- add _rbnfRulesFile key under each rule with the name of the text file to load with the rules.

CLDR-18956

### Review Note

output is planned to appear in the forthcoming 48.0.0-ALPHA3 cldr-json release. You can see the new format at:

- https://github.com/unicode-org/cldr-json/tree/v48/cldr-json/cldr-rbnf/rbnf

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -17,7 +17,7 @@ adapting software to the conventions of different languages.
 
 CLDR 48 was an open submission cycle allowing contributors to supply data for their languages via the CLDR Survey Tool —
 data that is widely used to support much of the world’s software.
-This data is also a factor in determining which languages are supported on mobile phones and computer operating systems. 
+This data is also a factor in determining which languages are supported on mobile phones and computer operating systems.
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


macchiati:
Question: would it be easier for clients to have the files in a differently named folder, so that all the files in a folder have the same format?

srl295:
> Question: would it be easier for clients to have the files in a differently named folder, so that all the files in a folder have the same format?

the .txt files are referenced by the .json files, clients don't need to (and should not) "process" the .txt files independently. 

~I can add a manifest file, though, listing all of them for convenience.~

I’m not sure the manifest file is needed— clients should use the available locales list from the `cldr-core` package. 

Also, the transforms are in a similar format, see https://github.com/unicode-org/cldr-json/tree/v48/cldr-json/cldr-transforms/transforms

srl295:
Hm - some of the RBNF rules are identical, so we could hash the output names

```
        "_rbnfRulesFile": "rbnf-bc4fdbcf2285984.txt"
```

not as friendly to humans inspecting it, though.

(not for this PR)

PR #5063, opened 2025-09-19
CLDR-18219 v48 BRS Subdivisions

srl295:
CLDR-18219

- update docs on how to update subdivision translations
- make subdivision sparql to sort as a 3rd level on the URL to be stable. Running it twice no longer has random changes.
- exclude entries from wiki that are 3166-1 codes
- exclude entries from existing XML that are 3166-1 codes 

ALLOW_MANY_COMMITS=true


srl295:
ah, i thought CLDR-18219 was assigned to me. the second part of it (excluding overlong) was already fixed, just needed to also exclude ISO-3166-1

macchiati (approved):

PR #5065, opened 2025-09-21
CLDR-18994 languageMatch: Make Kinyarwanda (rw) fallback to English (en)

musayann:
CLDR-18994

ALLOW_MANY_COMMITS=true


macchiati:
The PR looks good now. We will just be waiting until after the spec beta so that we can merge this for version 49. (It's a bit too late for 48.)

macchiati:
I converted to draft, so we don't mistakenly commit to v48. We can merge right after the branch.

musayann:
@macchiati marking the PR as ready for review again, since 48 has been released

AEApple:
@musayann - You need to squash the branch with an appropriate commit message so it passes the PR checker. The current PR title would work. I tried squashing to fix the commit message but I was not able to based on the permissions of your repo.

musayann:
@AEApple I’ve just done that; the PR checker should pass now. Thanks

AEApple (approved):

PR #5068, opened 2025-09-22
CLDR-19002 BRS currency update amd 180 BGN/EUR

srl295:
CLDR-19002

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/supplementalData.xml
+++ b/common/supplemental/supplementalData.xml
@@ -222,7 +222,8 @@ The printed version of ISO-4217:2001
             <currency iso4217="XOF" from="1984-08-04"/>
         </region>
         <region iso3166="BG">
-            <currency iso4217="BGN" from="1999-07-05"/>
+            <currency iso4217="EUR" from="2026-01-01" />
+            <currency iso4217="BGN" from="1999-07-05" to="2026-01-31"/>
             <currency iso4217="BGL" from="1962-01-01" to="1999-07-05"/>
             <currency iso4217="BGM" from="1952-05-12" to="1962-01-01"/>
             <currency iso4217="BGO" from="1879-07-08" to="1952-05-12"/>
--- a/docs/site/development/updating-codes/update-currency-codes.md
+++ b/docs/site/development/updating-codes/update-currency-codes.md
@@ -10,8 +10,8 @@ title: Update Currency Codes
 - **Use git diff to sanity check the two XML files against the old, and check them in.**
   - **"git diff \-w" is helpful to ignore whitespace. If there are only whitespace changes, there's no need to check them in.**
 - **Check the** [ISO amendments] **to get changes that will happen during the current cycle.**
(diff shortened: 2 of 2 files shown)


docs/site/development/updating-codes/update-currency-codes.md
```diff
@@ -10,8 +10,8 @@ title: Update Currency Codes
 - **Use git diff to sanity check the two XML files against the old, and check them in.**
   - **"git diff \-w" is helpful to ignore whitespace. If there are only whitespace changes, there's no need to check them in.**
 - **Check the** [ISO amendments] **to get changes that will happen during the current cycle.**
-- Example: <https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/amendments/dl-currency-iso-amendment-179.pdf>
-- It appears right now like there is no good way to collect all the amendments that are applicable, except to change "179" in the above link by incrementing until error \#404 results. So:
+- Example: <https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/amendments/dl-currency-iso-amendment-180.pdf>
+- It appears right now like there is no good way to collect all the amendments that are applicable, except to change the number (before `.pdf`) in the above link by incrementing until error \#404 results. So:
```
srl295, line 14: this way this 2nd line doesn't need to be updated.

common/supplemental/supplementalData.xml
```diff
@@ -222,7 +222,8 @@ The printed version of ISO-4217:2001
             <currency iso4217="XOF" from="1984-08-04"/>
         </region>
         <region iso3166="BG">
-            <currency iso4217="BGN" from="1999-07-05"/>
+            <currency iso4217="EUR" from="2026-01-01" />
+            <currency iso4217="BGN" from="1999-07-05" to="2026-02-01"/>
```
AEApple: Why is this 2026-02-01? I didn't see where the February 1st date comes from?

roozbehp: Correct. LDML says ‘`to` defaults to "24:00:00" (midnight at the end of the day).’ (https://unicode.org/reports/tr35/tr35.html#Date_Ranges)

While the PDF attached to the issue says 2026-01-31.

mihnita: _"**January 1-31, 2026**

A dual circulation period will be in effect. Both levs and euros will be accepted for cash payments, and businesses must accept both currencies."_

https://ibank.bg/en/bulgaria-and-the-euro

srl295: Right. Inclusive not exclusive

srl295:
> The end date for BGN needs to be corrected to 2026-01-31.

Please take a look

srl295:
> Still problematic. Now it says 2026-12. Should be 2026-01.

Thanks for checking. Sorry, I made a quick fix *too* quick.

PR #5069, opened 2025-09-23
CLDR-19003 Add aliases for million and portion

AEApple:
CLDR-19003

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DtdData.java
@@ -1687,6 +1687,8 @@ public class DtdData extends XMLFileReader.SimpleHandler {
                                         "concentr-portion", // deprecated
                                         "concentr-part",
                                         "concentr-million", // deprecated
+                                        "concentr-1e6", // NOTE: if we decide to add more, we should do the comparison programmatically
+                                        "concentr-1e9",
                                         "concentr-permillion", // deprecated
                                         "concentr-part-per-million", // deprecated
                                         "concentr-part-per-1e6",


common/supplemental/units.xml
```diff
@@ -627,7 +627,9 @@ For terms of use, see http://www.unicode.org/copyright.html
             <unitAlias type="pound-per-square-inch" replacement="pound-force-per-square-inch" reason="deprecated"/>
             <unitAlias type="milligram-per-deciliter" replacement="milligram-ofglucose-per-deciliter" reason="deprecated"/>
             <unitAlias type="metric-ton" replacement="tonne" reason="deprecated"/>
+			<unitAlias type="million" replacement="1e6" reason="deprecated"/>
             <unitAlias type="permillion" replacement="part-per-1e6" reason="deprecated"/>
+			<unitAlias type="portion" replacement="part" reason="deprecated"/>
             <unitAlias type="portion-per-1e9" replacement="part-per-1e9" reason="deprecated"/>
```
markusicu, line 633: do we need this kind of alias if we have aliases for the pieces?

macchiati: Formally not if all the parts are handled. But for ICU, maybe

macchiati: And doesn't hurt

AEApple:
Should I give a reason other than deprecated since these parts aren't in the deprecated list for https://github.com/unicode-org/cldr/blob/main/common/validity/unit.xml

I don't think it makes sense to add them as deprecated if they're not deprecated?

macchiati:
Ah, everything that is aliased should be deprecated, good catch. (Deprecated means that it should be avoided, not that it doesn't work.)

AEApple:
> Ah, everything that is aliased should be deprecated, good catch. (Deprecated means that it should be avoided, not that it doesn't work.)

I realized I tried updating this backwards, but I have another question before updating again. Should million and portion have the concentration prefix?

macchiati:
Hi, working on some fixes — the code didn't allow for a replacement to have an integer constant (1e6). That is, in the DTD, unitAlias@replacement @MATCH regex didn't allow for the replacement to start with an integer (it does allow it in subtags like ...-per-1e6).

macchiati:
Filed https://unicode-org.atlassian.net/browse/CLDR-19008 to document better how to handle adding/deprecating units.

macchiati:
(Approving preemptively, fingers-crossed that the tests now pass.)

AEApple:
Looks like there's still a test failure, but not one I would be able to resolve myself.

macchiati:
@AEApple This turns out to be too tricky to handle with the web interface; making a change, then waiting for the tests each time. (Sorry!) 

Unfortunately, I'm not set up to work with forks, so I'll make a new PR to take care of this.

PR #5073, opened 2025-09-24
CLDR-18758 Make the ordering clear plus cleanup

macchiati:
CLDR-18758

ALLOW_MANY_COMMITS=true


macchiati:
@shane, @eggrobin FYI

PR #5095, opened 2025-09-28
CLDR-18576 start/end in era supplemental data

macchiati:
CLDR-18576

This needed a bit more surgery than I thought.
- replaced outdated examples
- cleaned up the `type`
- connected start/end to to/from
- defined some illegal cases
- dropped a bunch of outdated conditions (the 1/2/3) in the diff

**Question**
why do some calendars have a code and others do not?
```xml
    <calendar type="persian">
      <calendarSystem type="solar"/>
      <eras>
          <era type="0" start="622-03-21" code="ap"/> <!-- Anno Persico -->
      </eras>
    </calendar>
    <calendar type="dangi">
      <calendarSystem type="lunisolar"/>
      <eras>
          <era type="0" start="-2332-02-15"/> <!-- 'sequential' year -->
      </eras>
    </calendar>
```

I'll also file a ticket for the following:
- https://github.com/unicode-org/cldr/blob/main/docs/ldml/tr35.md#date-and-date-ranges
  - Should mention start/end (used in eras)
  - The DTD should allow what the spec has, eg 1900-09-30 10:45:00 (optionally omitting minute, or minute-second)
  - The spec has "The dates and times are specified in local time, unless otherwise noted." But we don't define how an implementation is to compute _which_ local time is meant. And should we support timezones in start/end?
- To clean up edge cases, we should:
  - deprecate `end` attributes, and only have start dates.
  - force the `type` values to have the same order as the start dates. That is, if we have
    - \<era type=X start=Y>
	- \<era type=Z start=W>
	- then x < Z iff Y < W

That is in https://unicode-org.atlassian.net/browse/CLDR-19023, with a few more details.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -1068,44 +1068,66 @@ As in other cases, **narrow** may be ambiguous out of context.
 <!ATTLIST era aliases NMTOKENS #IMPLIED >
 ```
 
-The `<calendarData>` element now provides only locale-independent data about calendar behaviors via its `<calendar>` subelements, which for each calendar can specify the astronomical basis of the calendar (solar, lunar, etc.) and the date ranges for its eras.
-
-Era start or end dates are specified in terms of the equivalent proleptic Gregorian date (in "y-M-d" format). Eras may be open-ended, with unspecified start or end dates. For example, here are the eras for the Gregorian calendar:
-
-```xml
-<era type="0" end="0-12-31" code="gregory-inverse" aliases="bc bce"/>
-<era type="1" start="1-01-01" code="gregory" aliases="ad ce"/>
-```
-
-For a sequence of eras with specified start dates, the end of each era need not be explicitly specified (it is assumed to match the start of the subsequent era). For example, here are the first few eras for the Japanese calendar:
+The `<calendarData>` element provides locale-independent data about calendar behaviors via its `<calendar>` subelements,
+which for each calendar can specify the astronomical basis of the calendar (solar, lunar, etc.) and the date ranges for its eras.
+For example:
 
 ```xml
-<era type="0" start="645-6-19" />
-<era type="1" start="650-2-15" />
-<era type="2" start="672-1-1" />
-…
-```
-
-Some eras have additional `code` and `aliases` attributes that define invariant strings for identifying the eras. The `code` is a single globally unique identifier, and `aliases` are space-separated identifiers unique within the calendar. The code and aliases follow the following rules:
-
-1. Every calendar has either an era with a `code` that is the same as the BCP-47 name of that calendar or an `inheritEras` element pointing to another calendar with such an era. This era should be used for anchoring the "extended year" in the calendar (`u` in the date format pattern).
-2. Eras that count backwards (larger numbers for older years) are suffixed with `-inverse`.
-3. If the same era code is used in multiple calendars, then the calculations for year, month, and day in that era must be the same in all calendars in which it is used. For example, the `ethioaa` era is used in two calendar systems.
-
-If a `<calendar>` contains an `<inheritEras/>` element, all eras from the specified calendar should be inserted in order into the sequence of eras for the current calendar and follow the same start and end date rules. For example:
-
-```xml
-<calendar type="japanese">
-    <inheritEras calendar="gregorian" />
-    <eras>
-        <era type="0" start="645-6-19"/>
-        <era type="1" start="650-2-15"/>
-        <!-- ... -->
-    </eras>
+<calendar type="gregorian">
+  <calendarSystem type="solar" />
+  <eras>
+    <era type="0" end="0-12-31" code="bce" aliases="bc"/> <!-- Before Common Era, Before Christ -->
+    <era type="1" start="1-01-01" code="ce" aliases="ad"/> <!-- Common Era, Anno Domini -->
+  </eras>
 </calendar>
 ```
 
-This means that the two eras from calendar "gregorian" should be inserted into the era list for "japanese" for calculations and formatting.
+If a `<calendar>` contains an `<inheritEras/>` element, all eras from the specified calendar should be inserted in order into the sequence of eras for the current calendar, as described below.
+For example, the following means that the two eras from calendar "gregorian" should be inserted into the era list for "japanese" for calculations and formatting.
+
+```xml
+<calendar type="japanese">
+  <inheritEras calendar="gregorian" />
+  <eras>
+    <era type="232" start="1868-10-23" code="meiji"/>
+    <era type="233" start="1912-07-30" code="taisho"/>
+    <era type="234" start="1926-12-25" code="showa"/>
+    <era type="235" start="1989-01-08" code="heisei"/>
+    <era type="236" start="2019-05-01" code="reiwa"/>
+  </eras>
+</calendar>
+```
+
+Each `era` element has a `code` attribute and optional `aliases` attributes that define invariant strings for identifying the eras. These are more mnemonic than the `type` identifiers (see below).
+The `code` is unique within the calendar, and the `aliases` are space-separated identifiers, each also unique within the calendar.
+
+The `start` date is specified in terms of the equivalent _proleptic_ Gregorian date in the format "yyyy-MM-dd", such as 1842-01-01.
+An omitted start date behaves as if start=-∞.
+
+The order for the eras is given by the following algorithm:
+- Include all eras from the inheritEras calendar, if there is one.
+- An omitted start date behaves as if start=-∞
+- All elements are ordered by their start dates.
+- No two elements can have the same start date (otherwise the data is invalid).
+
+Note that the order of the eras is _not_ necessarily the order in the XML file, nor is it based on the numeric value of the `type`s.
+
+For a given _proleptic_ Gregorian date D and calendar C, the era code for D is in the `era` element in C with the greatest start date ≤ the given date.
+It is also the _first_ `era` element with start date ≤ the given date in C, given the above ordering for `era` elements.
+
+The `type` has an integer value. 
+The type values do not have to start at 0, nor do they need to be in chronological order.
+They are used to access the era names in locale files.
+For example:
+
+```xml
+<era type="232">Meiji</era>
+<era type="233">Taishō</era>
+<era type="234">Shōwa</era>
+<era type="235">Heisei</era>
+<era type="236">Reiwa</era>
+
+The `end` attribute is unused, and is slated for deprecation in the future.
 
 **Note:** The `territories` attribute in the `calendar` element is deprecated. It was formerly used to indicate calendar preference by territory, but this is now given by the _[Calendar Preference Data](#Calendar_Preference_Data)_ below.
 
--- a/docs/ldml/tr35-modifications.md
+++ b/docs/ldml/tr35-modifications.md
@@ -59,6 +59,9 @@ The LDML specification is divided into the following parts:
 
 ### Locale Identifiers
 * [Special Script Codes](tr35.md#special-script-codes) Added the `Hntl` compound script. (This is also reflected in the `<scriptData>` elements in supplementalData.xml.)
+* [Likely Subtags](tr35.md#likely-subtags) Changed the Canonicalize step to point to the section on canonicalization.
+* [Unicode Locale Identifier](tr35.md#unicode-locale-identifier) Changed the `attribute` component in the EBNF to be `uattribute` for consistency with `ufield`, etc.
+and to reduce confusion with XML attributes.
   
 ### Misc.
 * [Character Elements](tr35-general.md#character-elements) Added new exemplar types.
@@ -75,6 +78,7 @@ and updated the guidelines for using the different `dateTimeFormat` types.
 * [Time Zone Format Terminology](tr35-dates.md#time-zone-format-terminology) Added the **Localized GMT format** (replacing the **Specific location format**).
 This affects the behavior of the `z` timezone format symbol.
 There is also now a mechanism for finding the region code from short timezone identifier, which is used for the _non-location formats (generic or specific)_
+* [Calendar Data](tr35-dates.md#calendar-data) Specified more precisely the meaning of the `era` attributes in supplemental data, and how to determine the transition point in time between eras.
 
 ### Numbers
 * [Plural rules syntax](tr35-numbers.md#plural-rules-syntax) Added substantial clarifications and new examples.
--- a/docs/ldml/tr35.md
+++ b/docs/ldml/tr35.md
@@ -511,14 +511,14 @@ A _Unicode locale identifier_ is composed of a Unicode language identifier plus 
 | ----------------------------------------------------------------------------------------------------- | ----------------------------------------------- | ------------------- |
 | <a name="unicode_locale_id" href="#unicode_locale_id">`unicode_locale_id`</a>                         | `= unicode_language_id`<br/>  `extensions*`<br/>  `pu_extensions? ;` |
 | <a name="extensions" href="#extensions">`extensions`</a>                                              | `= unicode_locale_extensions`<br/>`\| transformed_extensions`<br/>` \| other_extensions ;` |
-| <a name="unicode_locale_extensions" href="#unicode_locale_extensions">`unicode_locale_extensions`</a> | `= sep [uU]`<br/>  `((sep ufield)+`<br/>  `\|(sep attribute)+ (sep ufield)*) ;` |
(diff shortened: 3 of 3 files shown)


docs/ldml/tr35-dates.md
```diff
@@ -1068,40 +1068,62 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+  <calendarSystem type="solar" />
+  <eras>
+    <era type="0" end="0-12-31" code="bce" aliases="bc"/> <!-- Before Common Era, Before Christ -->
+    <era type="1" start="1-01-01" code="ce" aliases="ad"/> <!-- Common Era, Anno Domini -->
+  </eras>
+</calendar>
 ```
+The `type` has an integer value. 
+The type values do not have to start at 0, nor be contiguous.
+However, they must be unique for the calendar, and define the ordering of `era`s within that calendar.
```
robertbastian: this sounds like it defines a chronological ordering, which it does not

docs/ldml/tr35-dates.md
```diff
@@ -1068,40 +1068,62 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+
+Era `start` or `end` dates are specified in terms of the equivalent _proleptic_ Gregorian date in the format "yyyy-MM-dd", such as 1842-01-01.
+These behave the same as the `to` and `from` attributes described in [Date and Date Ranges](tr35.md#date-and-date-ranges).
 
-For a sequence of eras with specified start dates, the end of each era need not be explicitly specified (it is assumed to match the start of the subsequent era). For example, here are the first few eras for the Japanese calendar:
+For the first era, the start date is -∞ (to the limit of the datetime datatype) whether or not there is a `start` attribute.
```
robertbastian: the "whether or not there is a `start` attribute" is just needed if we accidentally put a start attribute on the first era? I think we don't need this safety rail

robertbastian: define "first era"? it's not the one with the lowest `type`

docs/ldml/tr35-dates.md
```diff
@@ -1068,40 +1068,62 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+    <era type="235" start="1989-01-08" code="heisei"/>
+    <era type="236" start="2019-05-01" code="reiwa"/>
 …
 ```
 
-Some eras have additional `code` and `aliases` attributes that define invariant strings for identifying the eras. The `code` is a single globally unique identifier, and `aliases` are space-separated identifiers unique within the calendar. The code and aliases follow the following rules:
+The following are illegal:
+- an undefined transition: one era with no `end`, with the following era having no `start` attribute
+- an overlap: one era with an `end`, with the following era having a `start` before that `end`
+- a discontinuity: one era with an `end`, with the following era having a `start` that is not immediately after the end.
```
robertbastian: why even consider this, and not just make the start attribute implicitly end the previous era?

docs/ldml/tr35-dates.md
```diff
@@ -1068,40 +1068,62 @@ As in other cases, **narrow** may be ambiguous out of context.
...
 
-1. Every calendar has either an era with a `code` that is the same as the BCP-47 name of that calendar or an `inheritEras` element pointing to another calendar with such an era. This era should be used for anchoring the "extended year" in the calendar (`u` in the date format pattern).
-2. Eras that count backwards (larger numbers for older years) are suffixed with `-inverse`.
-3. If the same era code is used in multiple calendars, then the calculations for year, month, and day in that era must be the same in all calendars in which it is used. For example, the `ethioaa` era is used in two calendar systems.
+Eras may have additional `code` and `aliases` attributes that define invariant strings for identifying the eras.
+The `code` is a single globally unique identifier, and `aliases` are space-separated identifiers unique within the calendar.
```
robertbastian: it's not globally unique anymore, it's unique per calendar

docs/ldml/tr35-dates.md
macchiati: That's what I've been wrestling with, and why I filed https://unicode-org.atlassian.net/browse/CLDR-19023.

IMO, we should deprecate the `end` attribute, and say that all of the eras must have a start date (except optionally the first). That way the ordering is firmly established by the ordering of the start dates, and the ordering of the types is irrelevant.

We have to have some ordering *because the ordering in the XML file* is arbitrary; the elements can be interpreted in any order because there is no @ORDERED dtd annotation.

Now, if it is already the case that all but possibly one era has a start date, we could move to that this release. It would be much, much simpler.

I didn't respond to the next few comments, because they all related to this point.

docs/ldml/tr35-dates.md
macchiati: Ok, I can drop that; it was a leftover from the previous release. (And I wasn't sure what you all intended.

docs/ldml/tr35-dates.md
robertbastian: > IMO, we should deprecate the end attribute, and say that all of the eras must have a start date (except optionally the first).

That makes sense. Currently only inverse eras have an end date, but there it's always the day before the next start, so it's kind of redundant.

> Now, if it is already the case that all but possibly one era has a start date, we could move to that this release.

That's the case.

> That way the ordering is firmly established by the ordering of the start dates, and the ordering of the types is irrelevant.

That's the ordering I also have in mind, but here you say that the type defines an ordering, which is wrong.

macchiati:
I took a look at the data, searching for `end`s
```
4,323: <era type="0" end="0-12-31" code="bce" aliases="bc"/> <!-- Before Common Era, Before Christ --> 
4,577: <era type="1" end="622-07-18" code="bh"/> <!-- Before Hijrah --> 
4,584: <era type="1" end="622-07-18" code="bh"/> <!-- Before Hijrah --> 
4,591: <era type="1" end="622-07-18" code="bh"/> <!-- Before Hijrah --> 
4,598: <era type="1" end="622-07-17" code="bh"/> <!-- Before Hijrah --> 
4,605: <era type="1" end="622-07-18" code="bh"/> <!-- Before Hijrah --> 
4,663: <era type="0" end="1911-12-31" code="broc"/> 
```
It looks like we can ignore the `end` attributes entirely, and just order by start date, with elements without start date coming before all others.

Note the order in the islamic cases is bizarre; both the order of the type values and the ordering in the file are reverse chronological order.
```
          <eras>
              <era type="0" start="622-07-19" code="ah"/> <!-- Anno Hegirae -->
              <era type="1" end="622-07-18" code="bh"/> <!-- Before Hijrah -->
          </eras>
```

robertbastian:
> Note the order in the islamic cases is bizarre; both the order of the type values and the ordering in the file are reverse chronological order.

Yes, I noted this in the issue. This is intentional.

docs/ldml/tr35-dates.md
macchiati: I redid https://unicode-org.atlassian.net/browse/CLDR-19023 to make it clearer, and will bring this up in the Design meeting today to see if we can fast-track it.

macchiati: @robertbastian please let me know if there are any problems you see in that.

robertbastian: lg, just the timezone thing I already commented on

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+    <era type="233" start="1912-07-30" code="taisho"/>
+    <era type="234" start="1926-12-25" code="showa"/>
+    <era type="235" start="1989-01-08" code="heisei"/>
+    <era type="236" start="2019-05-01" code="reiwa"/>
+  </eras>
+</calendar>
 ```
 
-Some eras have additional `code` and `aliases` attributes that define invariant strings for identifying the eras. The `code` is a single globally unique identifier, and `aliases` are space-separated identifiers unique within the calendar. The code and aliases follow the following rules:
+Each `era` element has a `code` attribute and optional `aliases` attributes that define invariant strings for identifying the eras.
```
richgillam: Is it worth it to say something about who uses the `code` or `aliases` attributes?  Are APIs expected to use or return these, and are they intended to be the primary, everyday era identifiers in those APIs?  I think ICU uses the `type` attribute in its calendar-math APIs, for example, and I'm not sure it uses the `code` attribute at all (or only in BCP 47-oriented stuff).  I notice you took out a reference to BCP 47-- was that intentional?

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
-    </eras>
-</calendar>
-```
+Note that the order of the eras is _not_ necessarily the order in the XML file, nor is it based on the numeric value of the types.
+
+For a given _proleptic_ Gregorian date D and calendar C, the era code for D is in the `era` element in C with the greatest start date ≤ the given date.
+It is also the _first_ `era` element with start date ≤ the given date in C, given the above ordering for `era` elements.
+
+The `type` has an integer value. 
+The type values do not have to start at 0.
```
richgillam: Didn't we have requirements that the `type` values all have to be in sequence, whatever value they start at, and that they don't necessarily have to be in chronological order?  Do you want to say those things?

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+They are maintained for use by older software, for identifying the eras.
```
richgillam: Are you implying here that newer software shouldn't use them?

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
 
-This means that the two eras from calendar "gregorian" should be inserted into the era list for "japanese" for calculations and formatting.
+The `end` attribute is unused, and may be deprecated in the future.
```
richgillam: s/may/will ?

docs/ldml/tr35-modifications.md
```diff
@@ -59,6 +59,9 @@ The LDML specification is divided into the following parts:
 
 ### Locale Identifiers
 * [Special Script Codes](tr35.md#special-script-codes) Added the `Hntl` compound script. (This is also reflected in the `<scriptData>` elements in supplementalData.xml.)
+* [Likely Subtags](tr35.md#likely-subtags) Changed the Canonicalize step to point to the section on canonicalization.
+* [Unicode Locale Identifier](tr35.md#unicode-locale-identifier) Changed the `attribute` component in the EBNF to be `uattribute` for consistency with `ufield`, etc.
+and to reduce confusion with XML attributes.
```
sffc, line 64: Unrelated?

sffc (commented):
Era changes LGTM. There are some things about `uattribute` in here which don't seem related to eras.

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+Each `era` element has a `code` attribute and optional `aliases` attributes that define invariant strings for identifying the eras.
+The `code` is unique within the calendar, and the `aliases` are space-separated identifiers, each also unique within the calendar.
 
-1. Every calendar has either an era with a `code` that is the same as the BCP-47 name of that calendar or an `inheritEras` element pointing to another calendar with such an era. This era should be used for anchoring the "extended year" in the calendar (`u` in the date format pattern).
-2. Eras that count backwards (larger numbers for older years) are suffixed with `-inverse`.
-3. If the same era code is used in multiple calendars, then the calculations for year, month, and day in that era must be the same in all calendars in which it is used. For example, the `ethioaa` era is used in two calendar systems.
+The `start` date is specified in terms of the equivalent _proleptic_ Gregorian date in the format "yyyy-MM-dd", such as 1842-01-01.
+An omitted start date behaves as if start=-∞.
+See also [Date and Date Ranges](tr35.md#date-and-date-ranges).
```
robertbastian: maybe don't "see Date and Date Ranges" as that pull in times

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
+An omitted start date behaves as if start=-∞.
+See also [Date and Date Ranges](tr35.md#date-and-date-ranges). 
 
-If a `<calendar>` contains an `<inheritEras/>` element, all eras from the specified calendar should be inserted in order into the sequence of eras for the current calendar and follow the same start and end date rules. For example:
+The order for the eras is given by the following algorithm:
+- Include all eras from the inheritEras calendar, if there is one.
+- At most one element can have the start date omitted (otherwise the data is invalid).
+If there is such an element, it sorts before all other eras.
+- All other elements are ordered by their start dates.
+- No two elements can have the same start date (otherwise the data is invalid).
```
robertbastian, line 1111: I think this is a bit simpler, if two eras are missing a start date that's the same start date, so it's covered by the last requirement

```suggestion
- Include all eras from the inheritEras calendar, if there is one.
- An omitted start date behaves as if start=-∞
- All elements are ordered by their start dates.
- No two elements can have the same start date (otherwise the data is invalid).
```

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,60 @@ As in other cases, **narrow** may be ambiguous out of context.
...
-<calendar type="japanese">
-    <inheritEras calendar="gregorian" />
-    <eras>
-        <era type="0" start="645-6-19"/>
-        <era type="1" start="650-2-15"/>
-        <!-- ... -->
-    </eras>
-</calendar>
-```
+Note that the order of the eras is _not_ necessarily the order in the XML file, nor is it based on the numeric value of the types.
```
robertbastian:
```suggestion
Note that the order of the eras is _not_ necessarily the order in the XML file, nor is it based on the numeric value of the `type`s.
```

docs/ldml/tr35-dates.md
macchiati: Will drop for now

docs/ldml/tr35-dates.md
macchiati:
```suggestion
They are used to access the era names in locale files.
For example:

```xml
<era type="232">Meiji</era>
<era type="233">Taishō</era>
<era type="234">Shōwa</era>
<era type="235">Heisei</era>
<era type="236">Reiwa</era>
```
The `code` can also be used to access the era names, by mapping to the `type`.
```

File: docs/ldml/tr35-dates.md (hunk 8 above)
Comment (macchiati) [reply docs/ldml/tr35-dates.md]:
```suggestion
The `end` attribute is unused, and is slated for deprecation in the future.
```

docs/ldml/tr35-modifications.md
macchiati: Yes, but minor and too much trouble to disentangle with the deadline looming.

docs/ldml/tr35-dates.md
macchiati: fixing

docs/ldml/tr35-dates.md
macchiati:
```suggestion
The type values do not have to start at 0, nor do they need to be in chronological order.
```

docs/ldml/tr35-dates.md
macchiati:
```suggestion
```

docs/ldml/tr35-dates.md
macchiati:
```suggestion
Each `era` element has a `code` attribute and optional `aliases` attributes that define invariant strings for identifying the eras. These are more mnemonic than the `type` identifiers (see below).
```

macchiati:
I had to resolve a conflict to main; can someone restamp?

DavidLRowe (approved):
RSLGTM

richgillam:
Belated LGTM.  Thanks for the changes.

docs/ldml/tr35-dates.md
```diff
@@ -1068,44 +1068,66 @@ As in other cases, **narrow** may be ambiguous out of context.
...
 ```xml
-<calendar type="japanese">
-    <inheritEras calendar="gregorian" />
-    <eras>
-        <era type="0" start="645-6-19"/>
-        <era type="1" start="650-2-15"/>
-        <!-- ... -->
-    </eras>
-</calendar>
-```
```
srl295, line 1106: FWIW this trailing code fence got dropped and so broke the following links. Will fix in CLDR-19006

srl295: > #5151

PR #5103, opened 2025-09-30
CLDR-19031 Add c12, c24 to hour cycle identifier list

sffc:
CLDR-19031, a sub-item of CLDR-18894

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -164,6 +164,7 @@ The definition was unclear, and prone to misunderstanding — the more detailed 
 #### ldmlBCP47
 - `type` — adds a new attribute `region`.
 - `keyboard3@conformsTo` — is updated to allow "48".
+- `hc` — adds values `c12` and `c24` as Technical Preview. Also see the note about `h24` in "V49 advance warnings". ([CLDR-18894][])
 
 ### BCP47 Data Changes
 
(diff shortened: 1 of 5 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
```diff
@@ -109,6 +109,8 @@ public class ExtraPaths {
                             "//ldml/localeDisplayNames/types/type[@key=\"hc\"][@type=\"h12\"][@scope=\"core\"]",
                             "//ldml/localeDisplayNames/types/type[@key=\"hc\"][@type=\"h23\"][@scope=\"core\"]",
                             "//ldml/localeDisplayNames/types/type[@key=\"hc\"][@type=\"h24\"][@scope=\"core\"]",
+                            "//ldml/localeDisplayNames/types/type[@key=\"hc\"][@type=\"c23\"][@scope=\"core\"]",
+                            "//ldml/localeDisplayNames/types/type[@key=\"hc\"][@type=\"c24\"][@scope=\"core\"]",
```
sffc: Should I include these entries now, or add them after 48?

macchiati (approved):

docs/site/downloads/cldr-48.md
```diff
@@ -164,6 +164,7 @@ The definition was unclear, and prone to misunderstanding — the more detailed
 #### ldmlBCP47
 - `type` — adds a new attribute `region`.
 - `keyboard3@conformsTo` — is updated to allow "48".
+- `hc` — adds values `c12` and `c24` as Technical Preview. Also see the note about `h24` in "V49 advance warnings". ([CLDR-18894][])
```
AEApple, line 167: Why do we sometimes write this as H24 and other times h24. Which is more accurate?

sffc: `h24` is the identifier. "H24" is a way to refer abstractly to the concept

PR #5102, opened 2025-09-30
CLDR-19021 Misc items missing from spec

macchiati:
CLDR-19021

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #5104, opened 2025-09-30
CLDR-19006 regen spec ToC

srl295:


macchiati:
I think we might want to wait until tomorrow morning; I'm waiting on a review for https://github.com/unicode-org/cldr/pull/5102, which will change the TOC (a bit)

macchiati (approved):
But we might want to regen again before we release the beta.

PR #5116, opened 2025-10-01
CLDR-18843 Remove organization Special

btangmu:
-Remove Organization.special and special.visibleOnFrontEnd

-Remove all lines from Locales.txt referencing special organization

-Remove lines from Locales.txt referencing Cldr organization where same locale was included for special, with 3 exceptions: sr_Latn, qu, to, which are needed to pass test TestCLDRLocaleCoverage.TestCldrSuperset

-Remove SubmissionLocales.SPECIAL_ORG_LOCALES, no longer need to subtract from CLDR locales

-Revise SubmissionLocales.HIGH_LEVEL_LOCALES, removing some special locales

-Remove ChartLocaleGrowth.SpecialLocales, isSpecial

-No longer need to subtract special from CLDR locales in ListCoverageLevels, ShowInconsistentAvailable, DiffLanguageGroups, ListProblemDates, TestUnits

-In ShowLocaleCoverage, remove specialFlag

-In VettingViewer, remove HC = high coverage for special locales

-In TestStandardCodes.testTargetCoverageLevel, remove assertion for locale br

CLDR-18843

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
@@ -37,10 +37,6 @@ public final class SubmissionLocales {
     public static final Set<String> CLDR_LOCALES =
             StandardCodes.make().getLocaleToLevel(Organization.cldr).keySet();
 
-    /** This is the 'special' list from Locales.txt */
-    public static final Set<String> SPECIAL_ORG_LOCALES =
-            StandardCodes.make().getLocaleToLevel(Organization.special).keySet();
-
     /**
      * Non-CLDR Locales, but consistently have high level of engagement from volunteers to keep at
      * modern level. Reevaluate for each release based on meeting 95+% of modern, moderate, and
@@ -49,16 +45,10 @@ public final class SubmissionLocales {
     public static Set<String> HIGH_LEVEL_LOCALES =
             ImmutableSet.of(
                     // Note: ALL of these were found in Locales.txt under cldr.
-                    "chr", // Cherokee
                     "gd", // Scottish Gaelic, Gaelic
-                    "fo", // Faroese
                     "kok", // Konkani
                     "pcm", // Nigerian Pidgin
-                    "ha", // Hausa
-                    "hsb", // Upper Sorbian
-                    "dsb", // Lower Sorbian
-                    "yue_Hans", // Cantonese (Simplified)
-                    "to" //  Tongan
+                    "ha" // Hausa
                     );
 
     public static final Set<String> CLDR_OR_HIGH_LEVEL_LOCALES =
@@ -98,7 +88,6 @@ public final class SubmissionLocales {
         LOCALES_FOR_LIMITED = ImmutableSortedSet.copyOf(temp);
 
         Set<String> temp2 = new HashSet<>(CLDR_LOCALES);
-        temp2.removeAll(SPECIAL_ORG_LOCALES);
         TC_ORG_LOCALES = ImmutableSortedSet.copyOf(temp2);
     }
 
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
@@ -540,9 +540,6 @@ Cldr	;	be	;	modern	;	T5	Belarusian
 Cldr	;	cy	;	modern	;	T5	Welsh
 Cldr	;	yue	;	modern	;	T5	Cantonese
 
-#Cldr	Tier	generated
-Cldr	;	sr_Latn	;	modern		;	Tx	Serbian (auto-generated)
-
 #Cldr regional (from Google)
 Cldr	;	es_US	;	modern	; TR Spanish (United States)
 Cldr	;	es_MX	;	modern	; TR Spanish (Mexico)
@@ -553,8 +550,6 @@ Cldr	;	fr_CA	;	modern	; TR French (Canada)
 Cldr	;	de_CH	;	modern	; TR High German (Switzerland)
 
 #Cldr	other (from Apple)
-#Cldr    ;   fo  ;   modern
-#Cldr    ;   to  ;   moderate	# maybe different level from Apple's
 Cldr	;	en_AU	;	modern
 Cldr	;	es_MX	;	modern
 Cldr	;	fr_CA	;	modern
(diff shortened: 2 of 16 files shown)


tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
```diff
@@ -540,9 +540,6 @@ Cldr	;	be	;	modern	;	T5	Belarusian
 Cldr	;	cy	;	modern	;	T5	Welsh
 Cldr	;	yue	;	modern	;	T5	Cantonese
 
-#Cldr	Tier	generated
-Cldr	;	sr_Latn	;	modern		;	Tx	Serbian (auto-generated)
```
btangmu, line 544: This line for "sr_Latn" is moved, not removed. It is essentially unchanged, except for the addition of a comment explaining the relation to `TestCLDRLocaleCoverage.TestCldrSuperset`. The same applies to "qu" and "to".

btangmu:
The strategy that resulted in these changes was essentially: (1) remove all "special" from Locales.txt; (2) remove those same locales from "Cldr" in Locales.txt, except for sr_Latn, qu, to, which are needed to pass test `TestCLDRLocaleCoverage.TestCldrSuperset`; (3) revise Java code as needed for tests to pass and to eliminate references to special org

tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
```diff
@@ -49,16 +45,10 @@ public final class SubmissionLocales {
     public static Set<String> HIGH_LEVEL_LOCALES =
             ImmutableSet.of(
                     // Note: ALL of these were found in Locales.txt under cldr.
-                    "chr", // Cherokee
                     "gd", // Scottish Gaelic, Gaelic
-                    "fo", // Faroese
```
AEApple, line 54: I'm confused by this since ha, kok, pcm are TC locales.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
```diff
@@ -609,32 +604,18 @@ Cldr    ; doi ; basic ; Dogri
...
-Cldr  ; hsb ; modern ; Upper Sorbian
-Cldr  ; dsb ; modern  ; Lower Sorbian
 Cldr  ; gd  ; modern  ; Scottish Gaelic
-Cldr  ; chr ; modern  ; Cherokee
-Cldr	;	to	;	moderate ; Tongan
-Cldr  ; yue_Hans  ; modern  ; Cantonese (Simplified)
-
-Cldr  ; fo  ; moderate  ; Faroese
-Cldr  ; qu ; modern  ; Quechua
-Cldr  ; br  ; moderate  ; Breton
```
AEApple, line 621: I thought that the CLDR list was a superset of all the locales across all of the organizations?

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestStandardCodes.java
```diff
@@ -19,7 +19,6 @@ public class TestStandardCodes {
         "doi,BASIC", // CLDR locale
         "nn,MODERN", // CLDR locale
         "hnj,MODERN", // Maximum coverage (hmong)
-        "br,MODERATE", // Maximum (Breton)
```
AEApple, line 22: I don't understand this change?

tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
btangmu: Is the confusing thing (1) that chr and fo were removed, or (2) that ha, kok, and pcm were not removed (or were present in the first place)?

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestStandardCodes.java
btangmu: br was removed from Cldr in Locales.txt because it was one of the special locales. Consequently this test failed, unless br was removed from the test.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Locales.txt
btangmu: The test `TestCldrSuperset` ensures that, if I understand it.

These are deleted from Cldr because they were "special", and there was a lot of code that relied on the special locales being subtracted from the Cldr locales. The general strategy I employed was to remove the special locales from Cldr, with the 3 exceptions mentioned

tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
btangmu: chr and fo were removed since they were special.

ha, kok, and pcm were not removed since they weren't special. Why they were present in the first place, I'm not sure, though I suspect they might be required for `TestCldrSuperset` to pass.

macchiati: There is a test that CLDR org is a superset of the TC orgs' locales. We were tracking Specials as part of that, but that was a hack so we're removing it. So locales that were only in CLDR org by virtual of being Special are being removed.

macchiati: I think the HIGH_LEVEL_LOCALES should also go away; they were intended to align with the Specials above basic, but we are no longer tracking specials. However, that can be in a different PR.

macchiati:
Marking as draft because we don't want to merge before branch.

AEApple:
Okay, in that case in a follow-on PR, can we move the CLDR org to be the first organization in the locales.txt file and add an explanation of what it contains in the comment at the top of the file?

macchiati:
We should add the comment, but what I'd like to do is move to using the reformatting tool.

macchiati:
That is, copy in https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/48/tsv/coverage_goals.tsv instead (it reformats with each charts build), and do the ticket to clean out the deadwood orgs (that haven't contributed in (say) 5 years.

btangmu:
@macchiati wrote last week, "Marking as draft because we don't want to merge before branch." Now that we have branched with maint/maint-48, I've turned off draft and will merge this now.

PR #5119, opened 2025-10-02
CLDR-19026 ConsoleCheck default to -c comprehensive

srl295:
CLDR-19026

- [X] this completes it


AEApple (approved):
Thank you!

Do we have a dashboard measuring how fast tests run so that we can see if how quickly the tests pass and quickly know if test completion speed gets worse than anticipated in the future?

srl295:
> Thank you!
> 
> Do we have a dashboard measuring how fast tests run so that we can see if how quickly the tests pass and quickly know if test completion speed gets worse than anticipated in the future?

We don't, but I do notice that the 'check' job took 6 minutes (in this PR) whereas 'build' took 26.   So ConsoleCheck isn't the long pole.

PR #5120, opened 2025-10-03
CLDR-17846 Survey Tool shows LTR text in RTL locales with punctuation in wrong place

btangmu:
-Specify the direction for the flagged-for-review hover text (s-flag span)

-New constant TRANS_HINT_DIRECTION=ltr (left-to-right) goes with existing constant TRANS_HINT_ID=en; change the scope to the whole cldrText.mjs file since it is now used twice, not only in a single function

CLDR-17846

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrTable.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrTable.mjs
@@ -43,6 +43,9 @@ const NO_WINNING_VALUE = "no-winning-value";
  */
 const EMPTY_ELEMENT_VALUE = "❮EMPTY❯";
 
+const TRANS_HINT_ID = "en"; // expected to match SurveyMain.TRANS_HINT_ID
+const TRANS_HINT_DIRECTION = "ltr"; // English is left-to-right
+
 /**
  * Remember the element (HTMLTableCellElement or HTMLDivElement) that was most recently
  * shown as "selected" (displayed with a thick border). When a new element gets selected,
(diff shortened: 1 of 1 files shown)


btangmu:
The HTML in question was like this:
```
<td title="Winning value" class="d-win proposedcell cldrValue" dir="rtl" lang="ur">
  <span class="s-flag" title="This item has been flagged for review by the CLDR Technical Committee.">&nbsp; &nbsp;</span>
  <div class="d-item">
    <div class="choice-field">
      <label title="" class="btn btn-default" data-original-title="Vote">
        <input class="ichoice-o" type="radio" title="click to vote" value="arabext" id="vYXJhYmV4dA__xgorkb">
      </label>
      <span class="subSpan" data-original-title="" title="">
        <span class="winner cldrValue" dir="rtl" lang="ur">arabext</span>
      </span>
      <div class="i-star" title="This is the “baseline” data. See http://cldr.unicode.org/translation/getting-started/guide#TOC-Icons"></div>
    </div>
    <div class="d-example well well-sm cldrValue" dir="rtl" lang="ur">
      <div class="cldr_example">۲۳۴۵</div>
    </div>
  </div>
</td>
```
This PR makes a local change by adding `dir="ltr" lang="en"` to the "flagged" span. A more far-reaching change would be to remove rtl from the td element and make sure it is instead included on all the inner elements that need it (which may or may not already be the case; there are already nested elements with rtl). I decided to take the more cautious approach by fixing what clearly needed fixing, and leaving the rest as-is (except that I also added TRANS_HINT_DIRECTION to the setLang call in updateRowEnglishComparisonCell, assuming if lang=TRANS_HINT_ID is right then dir=TRANS_HINT_DIRECTION must also be right).

srl295 (approved):

PR #5122, opened 2025-10-04
CLDR-18209 spec: require modifications section to be updated

srl295:
CLDR-18209

- [ ] After this is merged, must require the **Formatter + Style Checker** check.

ALLOW_MANY_COMMITS=true

## Review Notes

In the summary area, you will see these:

### No spec changes

- Run: https://github.com/unicode-org/cldr/actions/runs/18251384566

image

### Failing: Update without mod

Note copy button!

- Run: https://github.com/unicode-org/cldr/actions/runs/18251524590

image

### Good: mod was updated!

- Run: https://github.com/unicode-org/cldr/actions/runs/18251557231

image

Note that it's OK for a change (such as this present PR) to _only_ change the modification section.

### DIFF
--- a/tools/scripts/tr-archive/check-mod-section.sh
+++ b/tools/scripts/tr-archive/check-mod-section.sh
@@ -40,7 +40,7 @@ then
     exit 0
 else
     msg "⚠️ Spec changes but tr35-modifications.md section was not updated!"
-    msg "At least add an HTML comment such as:"
+    msg "For non-substantive changes that do not require an item in modifications such as fixing typos, add an HTML comment such as:"
     msg ""
     msg "     <!-- updated whitespace -->"
     msg ""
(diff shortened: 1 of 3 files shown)


tools/scripts/tr-archive/check-mod-section.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+if echo "${CHANGED}" | fgrep -s '/tr35-modifications.md';
+then
+    msg "✅ Spec changes: tr35-modifications.md was updated"
+    list_changed
+    exit 0
+else
+    msg "⚠️ Spec changes but tr35-modifications.md section was not updated!"
+    msg "At least add an HTML comment such as:"
```
AEApple:
```suggestion
    msg "For non-substantive changes that do not require an item in modifications such as fixing typos, add an HTML comment such as:"
```

PR #5126, opened 2025-10-07
CLDR-19050 Now always use latest published version of ICU4J 78rc

pedberg-icu:
CLDR-19050

Switch to using ICU4J version 78.1-SNAPSHOT now that ICU is at rc and has branched (so now we just always use the latest version of 78.1 published to the maven GitHub repo, not a specific dated one)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/pom.xml
+++ b/tools/pom.xml
@@ -62,8 +62,8 @@
 		       between CLDR and the updated version of ICU, which would need to be fixed as part
 		       of the PR.
 			-->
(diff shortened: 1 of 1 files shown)


pedberg-icu:
Oops, this one targets the wrong branch, cerated another for the right branch: https://github.com/unicode-org/cldr/pull/5127

But now I see how to retarget an existing PR to a different branch, so I could have just edited this one. Oh well.

PR #5154, opened 2025-10-22
CLDR-18451 Update TerritoryInfo references after v48 changes

conradarcturus:
This is a followup from https://github.com/unicode-org/cldr/pull/5148 -- checking other references in code to languageData/language/territories and updating them. The biggest change is updating the release notes.

https://unicode-org.atlassian.net/browse/CLDR-18451

CLDR-18451

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -25,9 +25,7 @@ This data is also a factor in determining which languages are supported on mobil
 
 Some of the most significant changes in this release are:
 - Updated for Unicode 17, including new names and search terms for new emoji, new sort-order, Han → Latin romanization additions for many characters.
-- Many additions to language data including:
-    - Likely Subtags, for deriving the likely script and region from the language (used in many processes).
-    - Language populations in countries: significant updates to improve accuracy and maintainability.
+- Many additions to language data supporting Likely Subtags, for deriving the likely script and region from the language (used in many processes).
 - Updated to the latest external standards and data sources, such as the language subtag registry, UN M49 macro regions, ISO 4217 currencies, etc.
 - New formatting options
     - Rational number formats added, allowing for formats like 5½.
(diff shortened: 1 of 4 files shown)


docs/site/downloads/cldr-48.md
```diff
@@ -313,7 +311,7 @@ The following files are new in the release:
 ## Migration
 
 - Number patterns that did not have a specific numberSystem (such as `latn` or `arab`) had been deprecated for many releases, and were finally removed.
-- Additionally, language and territory data in `languageData` and `territoryInfo` data received significant updates to improve accuracy and maintainability [CLDR-18087]
+- Additionally, language and territory data in `languageData` and `territoryInfo` data received significant updates to improve accuracy and maintainability [CLDR-18087] In particular, the `territories` attribute in `languageData` was deprecated and removed, as it was unclear and prone to misunderstanding. Implementations that used this data may need to adjust accordingly, using `territoryInfo`.
```
btangmu, line 314: Need a period here before "In particular"?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
```diff
@@ -352,14 +352,16 @@ private static void writeBasicLanguageData(PrintWriter out, Set<RowData> sortedI
                                         TreeSet.class));
             }
             if (languageInCountryData.officialStatus.isMajor()) {
-                // Output will look like <language type="sw" territories="TZ"/>
+                // This is no longer saved to output
+                // It used to appear like: <language type="sw" territories="TZ"/>
```
btangmu, line 356: This is mysterious. Does the code still serve a purpose? There can be a tendency to keep code around for historical reference and/or in case it might be useful again someday. However, that leads to crufty hard-to-maintain code. So I'd rather see the code removed if unnecessary, and otherwise a comment explaining why the code is here, rather than a comment saying what it used to do or doesn't do...

btangmu (dismissed):
See comments

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
conradarcturus: I agree, it's not great to leave this code in. I started this refactor last year but I punted the more consequential work until I had a proper design to replace the old class completely with a new format.

Removing this code ended up affecting other classes that consume the data and ended up causing a big rewrite that I wasn't prepared. Would you like me to change the comment to instead say "this is not saved to the output but it is used in the BasicLanguageData class"?

btangmu: That sounds good. Also, maybe (optional!) there could be a ticket for the larger refactoring, and you could put in a TODO comment referencing that ticket. Unless that's just wishful thinking :-)

macchiati (approved):
If the only issue is the period, it would be good to get this finished soon (rather than have possible future conflicts)

macchiati:
I don't know why this PR is still hanging. I'm going to dismiss Tom's requested changes and merge.

PR #5157, opened 2025-10-23
CLDR-19053 Some zone-related updates

robertbastian:
CLDR-19053

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -31,7 +31,7 @@ Some of the most significant changes in this release are:
 - Updated to the latest external standards and data sources, such as the language subtag registry, UN M49 macro regions, ISO 4217 currencies, etc.
 - New formatting options
     - Rational number formats added, allowing for formats like 5½.
-    - For timezones, `usesMetazone` adds two new attributes `stdOffset` and `dstOffset` so that implementations can use either "vanguard" or "rearguard" TZDB data sources.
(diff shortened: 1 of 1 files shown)


docs/site/downloads/cldr-48.md
```diff
@@ -31,7 +31,7 @@ Some of the most significant changes in this release are:
 - Updated to the latest external standards and data sources, such as the language subtag registry, UN M49 macro regions, ISO 4217 currencies, etc.
 - New formatting options
     - Rational number formats added, allowing for formats like 5½.
-    - For timezones, `usesMetazone` adds two new attributes `stdOffset` and `dstOffset` so that implementations can use either "vanguard" or "rearguard" TZDB data sources.
+    - For timezones, `usesMetazone` adds two new attributes `stdOffset` and `dstOffset` so that implementations can use either "main" or "rearguard" TZDB data sources.
```
AEApple, line 34: I'm less familiar with TZDB. How do they refer to the non-rearguard version? The entire time this was discussed it was referred to as 'vanguard' so it's a bit unexpected to see it being changed to 'main' now?

I think we should match however this is referred to in TZDB (and we can just add a reference to that in the release notes)? WDYT?

robertbastian: they call it main. there are three versions, both vanguard and rearguard are opt-in. the negative dst diff is between main and rearguard, the diff between main an vanguard is not relevant to us

robertbastian: there is some information here but this is not documented very clearly https://github.com/eggert/tz/blob/main/ziguard.awk

AEApple (approved):

PR #2769, opened 2023-03-09
CLDR-16292 Update supplemental/ordinals.xml for Spanish ordinal numbers of ~1 and ~3

gosplit:
Spanish ordinal numbers First and Third will drop `o` in masculine adjective forms. 
e.g.,  1st day, 2nd day, 3rd day, 4th day, 5th day (English)= 1ᵉʳ día, 2º día, 3ᵉʳ día, 4º día, 5º día (Spanish)

The [supplemental/ordinals.xml](https://github.com/unicode-org/cldr/pull/2769/files#diff-41bbf64dc219b18617d1a124e4b6c0f8cfa47b2458afb7e2a092f077fc083b2b) should be updated accordingly. 

Reference 1: CLDR rbnf of `es`:  https://github.com/unicode-org/cldr/blob/835e0c982ed37f994c6363909d958e1731693c03/common/rbnf/es.xml#L289-L297

Reference 2: https://www.spanish.academy/blog/ordinal-numbers/

CLDR-16292

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #5163, opened 2025-10-26
CLDR-16292 Plural rules in Spanish should return different ordinal types for number 1 and 3

grhoten:
CLDR-16292

This is a revised submission of #2769. It changes RBNF to use the new plural format values.

### DIFF
--- a/common/main/es.xml
+++ b/common/main/es.xml
@@ -7731,7 +7731,8 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<pluralMinimalPairs count="one">{0} día</pluralMinimalPairs>
 			<pluralMinimalPairs count="many">{0} de días</pluralMinimalPairs>
 			<pluralMinimalPairs count="other">{0} días</pluralMinimalPairs>
-			<ordinalMinimalPairs ordinal="other">Toma la {0}.ª a la derecha.</ordinalMinimalPairs>
+            <ordinalMinimalPairs ordinal="one">El {0}.ᵉʳ equipo clasificado ganó.</ordinalMinimalPairs>
+			<ordinalMinimalPairs ordinal="other">El equipo {0}.º clasificado ganó.</ordinalMinimalPairs>
 			<genderMinimalPairs gender="feminine">La {0} es...</genderMinimalPairs>
 			<genderMinimalPairs gender="masculine">El {0} es…</genderMinimalPairs>
 		</minimalPairs>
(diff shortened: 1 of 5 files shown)


grhoten:
@srl295 @AEApple for review as per the CLDR TC meeting. I can't request a reviewer here.

AEApple:
Can you fix the minimal pairs for the ordinals? https://github.com/unicode-org/cldr/blob/e1ac2aea30b8ce94bce0b538786e37f776545cee/common/main/es.xml#L7734

I don't think they use RBNF logic right now and I think they have to be updated manually.

grhoten:
Err, that's grammatically correct. The singular masculine, singular feminine, plural masculine, and plural feminine all have a single grammatical state.  Only the adjective form of the ordinal uses the er suffix, which has 2 grammatical states.

I mean an alternate exemplar translation could be this translation. I'm not a native Spanish speaker. So I'm not 100% sure about the grammatical correctness. Is this the one that you want?

```
<ordinalMinimalPairs ordinal="one">El {0}.ᵉʳ equipo clasificado ganó.</ordinalMinimalPairs>
<ordinalMinimalPairs ordinal="other">El {0}.º equipo clasificado ganó.</ordinalMinimalPairs>
```

jbezos:
I’m a native Spanish speaker and “El {0}.ᵉʳ equipo clasificado ganó” is correct. However, “El {0}.º equipo clasificado ganó” is *not* correct. But if the ordinal is placed after the noun, then it is “El equipo {0}.º“. They are masculine. Feminine is “{0}.ª”.

AEApple (dismissed):
I'm going to approve this one, but let's update the plurals minimal pairs before resolving the ticket since this change updates the plurals.

I think one of the open questions for me is what to do about the grammatical gender since we do have support for certain unit types even if it's not available everywhere.

grhoten:
@AEApple I've updated the minimal pairs for both Spanish variants. The digit ordinals have also been fixed to use the correct localized form of the digit ordinals between the base es and es_419. The es_419 locale must override this usage due to differing punctuation usage.

The existing gender usage of genderMinimalPairs in Spanish seems correct. There are traditionally only 2 grammatical genders. If you want to go into a deeper philosophical debate about that, I recommend separating such discussions out to a separate ticket.

AEApple:
> @AEApple I've updated the minimal pairs for both Spanish variants. The digit ordinals have also been fixed to use the correct localized form of the digit ordinals between the base es and es_419. The es_419 locale must override this usage due to differing punctuation usage.
> 
> The existing gender usage of genderMinimalPairs in Spanish seems correct. There are traditionally only 2 grammatical genders. If you want to go into a deeper philosophical debate about that, I recommend separating such discussions out to a separate ticket.

no, I meant right now I don't think the plural charts had grammatical gender at all in any language, but I could have missed something.

p.s. You should be able to merge changes now.

AEApple (approved):

PR #5164, opened 2025-10-26
CLDR-19089 Improve parsing of numbers with Irish RBNF rules

grhoten:
CLDR-19089

While these changes don't fully fix parsing with Irish RBNF rules. It's one step closer. I tested these rules with [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp?locale=ga).

Here are the important changes:

* The year is now pronounced correctly. It's not supposed to use English style spelling. It uses a more traditional form.
* The cardinals are now separated from counting numbers.
* Numbers like 2000 and 2001 are now correctly parsing, but numbers like 212212 still struggle to parse correctly. It's due to Irish's unusual ordering for the tens range to have the tens after the unit.  The order is like "two hundred two thousand ten two hundred two ten" for the value 212212.

### DIFF
--- a/common/rbnf/ga.xml
+++ b/common/rbnf/ga.xml
@@ -15,31 +15,10 @@ For terms of use, see http://www.unicode.org/copyright.html
             <rbnfRules><![CDATA[
 %%lenient-parse:
 & ' ' , ',' ;
-%%2d-year:
-0: agus =%spellout-numbering=;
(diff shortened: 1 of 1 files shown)


grhoten:
@srl295 @AEApple for review as per the CLDR TC meeting. I can't request a reviewer here.

srl295 (approved):
structure looks good

PR #5167, opened 2025-10-27
CLDR-19090 Ordinal plural rules are missing for af and bg

grhoten:
CLDR-19090

Afrikaans and Bulgarian are missing the following plural rules. Afrikaans mirrors the current RBNF ordinal rules. Bulgarian is similar to the current RBNF rules, but it follows the more grammatically correct/explicit forms for 7th, and 8th.

Bulgarian also has additional grammatical states for thousands, millions, billions, trillions and quadrillions. Unfortunately that's too many states for plural format. The revised rules aren't any worse than the current rules. It should provide the same formatted results, but the parsing work.

These updates are needed to allow the RBNF ordinal rules to roundtrip with formatting and parsing. RBNF also needs to be updated to use these new rules.

### DIFF
--- a/common/rbnf/af.xml
+++ b/common/rbnf/af.xml
@@ -170,27 +170,14 @@ x.x: =#,##0.#=;
         </rulesetGrouping>
         <rulesetGrouping type="OrdinalRules">
             <rbnfRules><![CDATA[
-%%digits-ordinal-indicator:
-0: ste;
-1: ste;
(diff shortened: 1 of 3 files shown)


macchiati (approved):

grhoten:
@srl295 @AEApple Can one of you merge this? I don't have merge permissions.

PR #5173, opened 2025-10-30
CLDR-11400 Remove historical Japanese eras

robertbastian:
CLDR-11400

ALLOW_MANY_COMMITS=true


AEApple:
@robertbastian - The pr checker seems to struggle with punctuation. You'll need to remove the colon following the CLDR ticket number and squash. I've also see it get caught by a hyphen else where in commit message even when the ticket is valid at the start.

macchiati (dismissed):

macchiati (approved):

PR #5194, opened 2025-11-14
CLDR-19035 GenerateExampleDependencies, performance, refactoring

btangmu:
-Update ExampleDependencies.java by running GenerateExampleDependencies

-Improve GenerateExampleDependencies performance roughly 30% by conditionally skipping intern() in XMLSource: new XMLSource.setDoInternStrings()

-Add options to GenerateExampleDependencies for testing with subset of locales (e.g., -f^[^_]+$ to skip locales whose names contain underscore)

-Report how many minutes it takes to run GenerateExampleDependencies

-Fix a bug in GenerateExampleDependencies: call clearRecordedPaths after loop

-Refactor ICUServiceBuilder and related code: new constructor with CLDRFile, instead of calling setCldrFile separately

-Remove ICUServiceBuilder.setCldrFile and the no-arg ICUServiceBuilder constructor

-Only call the ICUServiceBuilder constructor locally from forLocale, except for fictional test locales xx, mul

-Do not call icuServiceBuilder.clearCache when it is superfluous immediately after constructing

-Call ICUServiceBuilder.clearCache instead of duplicating the code; remove ISB_CAN_CLEAR_CACHE

-New constructors for DateTimeFormats and ZoneFormats obviate calling set()

-New FlexibleDateFromCLDR constructor with CLDRFile replaces FlexibleDateFromCLDR.set

-Change CheckNumbers.defaultNumberingSystem to a local variable

-New constant LocaleNames.XX_TEST (compare existing LocaleNames.MUL)

-Make data private/final/static where possible; ICUServiceBuilder.supplementalData can be static since it does not depend on the locale

-Some automated refactoring by IntelliJ, from accepting suggestions linked to compiler warnings

-Comments

CLDR-19035

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleDependencies.java
```diff
@@ -487,9 +498,6 @@ public class ExampleDependencies {
                     .putAll(
                             "//ldml/dates/timeZoneNames/zone[@type=\"([^\"]*+)\"]/short/daylight",
                             "//ldml/personNames/personName[@order=\"([^\"]*+)\"][@length=\"([^\"]*+)\"][@usage=\"([^\"]*+)\"][@formality=\"([^\"]*+)\"]/namePattern")
-                    .putAll(
-                            "//ldml/dates/timeZoneNames/zone[@type=\"([^\"]*+)\"]/short/generic",
-                            "//ldml/personNames/personName[@order=\"([^\"]*+)\"][@length=\"([^\"]*+)\"][@usage=\"([^\"]*+)\"][@formality=\"([^\"]*+)\"]/namePattern")
```
btangmu, line 492: This is the only example dependency removed. I don't know why. The implication is that examples for these namePattern paths no longer depend on values for these timeZoneNames paths.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -907,6 +899,13 @@ private DecimalFormatSymbols _getDecimalFormatSymbols(String numberSystem) {
             symbols.setMonetaryGroupingSeparator(symbols.getGroupingSeparator());
         }
 
+        // Note: as of 2025-11-05, none of these six paths occur anywhere except in root.xml:
+        // .../beforeCurrency/currencyMatch
+        // .../beforeCurrency/surroundingMatch
+        // .../beforeCurrency/insertBetween
+        // .../afterCurrency/currencyMatch
+        // .../afterCurrency/surroundingMatch
+        // .../afterCurrency/insertBetween
```
btangmu, line 915: If these paths occur only in root.xml and can't be changed in ST, then their values are ordinarily effectively constants and need not be taken into account for example dependencies. There may be other such paths. That implies:

- ExampleDependencies.java could be smaller
- Maybe ICUServiceBuilder caching needn't be completely disabled for GenerateExampleDependencies, so it could run faster

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -40,6 +40,18 @@ public class ICUServiceBuilder {
         iso.setTimeZone(utc);
     }
 
+    public ICUServiceBuilder() {
+        // This constructor without CLDRFile requires setCldrFile to be called subsequently.
```
btangmu: The no-arg ICUServiceBuilder constructor still has 4 remaining usages. For example, CheckDates.icuServiceBuilder has an instance initializer and icuServiceBuilder.setCldrFile is called from CheckDates.handleSetCldrFileToCheck. The same ICUServiceBuilder instance gets handleSetCldrFileToCheck called repeatedly with a different CLDRFile (and locale) each time.

macchiati: Is the idea to make ICUServiceBuilder be immutable, thus no set() method. (Seems like a good idea).

Also, splitting out the functionality of ICUServiceBuilder into separate classes with smaller chunks of data that they need would allow us to be finer-grained when flushing a cache.

btangmu: It could be helpful to make ICUServiceBuilder immutable. I'm still trying to understand it better, especially in relation to caching and example dependencies. It was mostly being constructed like `new ICUServiceBuilder().setCldrFile(cldrFile)`. It seems better to use the constructor like `new ICUServiceBuilder(cldrFile)` where possible. Sometimes `clearCache()` was needlessly called between `new ICUServiceBuilder()` and `setCldrFile(cldrFile)`.

However, for some purposes like `CheckDates()`, currently the same ICUServiceBuilder is re-used with many different CLDRFiles. I don't know whether that gives a significant performance benefit. I'd like to measure that before deciding whether to make ICUServiceBuilder immutable.

btangmu: I've just made a third commit, making ICUServiceBuilder almost immutable. Specifically to find any performance difference from calling `new ICUServiceBuilder(cldrFile)` for each locale instead of calling `icuServiceBuilder.setCLdrFile(cldrFile)` on the same instance, I tested performance of `TestCheckCLDR.TestAllLocales` which is an expensive test taking between 91 and 107 seconds locally. On average it seems to run slightly faster with that change, but I suspect that's only noise and the change makes no significant difference to performance.

One part of ICUServiceBuilder that is still not immutable is collationFile, and there's something strange related to that. `ICUServiceBuilder.collationFile` is only initialized in `ICUServiceBuilder.forLocale()`. Instances of ICUServiceBuilder that are constructed without `ICUServiceBuilder.forLocale()` never get `ICUServiceBuilder.collationFile`.

Also, `ICUServiceBuilder.forLocale()` uses a cache of ICUServiceBuilder instances, one for each locale. Presumably this cache has a large performance benefit. What's puzzling, then, is why is the cache not always used? There are many callers of `new ICUServiceBuilder()` that bypass the cache.

btangmu: The fourth commit replaces nearly all `new ICUServiceBuilder(cldrFile)` with `ICUServiceBuilder.forLocale()` for consistency and improved performance. The only exception is for some unit tests that use fictitious locale names "xx" and "mul", with special test data, for which `forLocale` would not work.

There is a performance improvement for `TestCheckCLDR.TestAllLocales` which was taking between 91 and 107 seconds locally and now takes between 75 and 79 seconds.

btangmu: Before 4th commit, "Test with maven" took 22m 28s; after 4th commit, "Test with maven" took 21m 25s. Only 63 seconds shorter, so not a big performance improvement as far as running tests is concerned. The code is cleaner though, and these changes may pave the way for more improvements to GenerateExampleDependencies and other features.

macchiati: > It could be helpful to make ICUServiceBuilder immutable. I'm still trying to understand it better, especially in relation to caching and example dependencies.

This is a good topic for discussion in the infra group.

When the underlying CLDRFile is immutable (or frozen), then the ICUServiceBuilder can be cached. So for charts, unit tests, etc that can be the case. We should check, however, that CLDRFile::isFrozen! (In general, we should not be caching mutable/unfrozen objects.

The problem is when the CLDRFile can change underneath, as in the survey tool. In that case, it cannot be cached, since the change to CLDRFile can require that the data in the ICUServiceBuilder be refreshed. Now, ideally, we would have objects like ICUServiceBuilder have Listeners on their mutable CLDRFile; then we would know to refresh the ICUServiceBuilder only when that CLDRFile changes a value that affects the ICUServiceBuilder; that would cut down on needless recreation. (But before undertaking anything like that, we should measure the performance cost of needless recreation first!)

The other nice cleanup would be to split ICUServiceBuilder into different parts, based on the CLDRData that they need. For example, we could have ICUNumberBuilder that just fetched the number information and ICUDateBuilder that just fetched the date information (but also has-a ICUNumberBuilder, since that is used in building dates). But again, only worth doing if we know it would improve performance significantly.

> Only 63 seconds shorter

That is actually pretty good! That is 5% of 22m 28s. With ten 63s reductions, we'd be taking half the time for the tests (and probably a measurable effect on ST; since checks and examples would run faster).

tools/cldr-code/src/main/java/org/unicode/cldr/icu/TestCldr.java
```diff
@@ -223,7 +224,7 @@ void set(AttributeName attributeName, String attributeValue) {
         }
 
         void checkResult(String value) {
-            ULocale ul = new ULocale("xx");
+            ULocale ul = new ULocale(LocaleNames.XX_TEST);
```
srl295, line 227: why XX and not `ZZ`?

srl295, line 227:
```suggestion
            ULocale ul = null;
```

or, just move the try inside the loop.

`ULocale("XX")` doesn't make sense. 

this is simply an initialization and shouldn't even be a locale object.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -438,7 +437,12 @@ public ExampleGenerator(CLDRFile resolvedCldrFile, CLDRFile englishFile) {
                 supplementalDataInfo.getGrammarInfo(localeId); // getGrammarInfo can return null
         this.englishFile = englishFile;
         this.typeIsEnglish = (resolvedCldrFile == englishFile);
-        icuServiceBuilder.setCldrFile(resolvedCldrFile);
+        boolean locIsFake =
+                LocaleNames.XX_TEST.equals(localeId) || LocaleNames.MUL.equals(localeId);
```
srl295, line 441: MUL isn't fake, it's the sandbox and can still be built.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/LanguageCodeConverter.java
```diff
@@ -52,7 +53,7 @@ public class LanguageCodeConverter {
                     .put("sit-NP", "lif")
                     .put("ut", "und")
                     .put("un", "und")
-                    .put("xx", "und")
+                    .put(LocaleNames.XX_TEST, "und")
```
srl295, line 56: this makes me think that `und` should be used instead of XX_TEST

srl295, line 56:
```suggestion
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
```diff
@@ -46,7 +47,7 @@ public void TestDateImportance() {
         try {
             // Build test file
 
-            XMLSource source = new SimpleXMLSource("xx");
+            XMLSource source = new SimpleXMLSource(LocaleNames.XX_TEST);
```
srl295, line 50:
```suggestion
            XMLSource source = new SimpleXMLSource("mul");
```

mul seems good for here

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestLocale.java
```diff
@@ -457,7 +458,7 @@ public void TestBrackets() {
             },
         };
         // load up a dummy source
-        SimpleXMLSource dxs = new SimpleXMLSource("xx");
+        SimpleXMLSource dxs = new SimpleXMLSource(LocaleNames.XX_TEST);
```
srl295, line 461:
```suggestion
        SimpleXMLSource dxs = new SimpleXMLSource(LocaleNames.MUL);
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -40,6 +40,40 @@ public class ICUServiceBuilder {
...
+     * @param cldrFile the CLDRFile
+     */
+    private ICUServiceBuilder(CLDRFile cldrFile) {
+        this(cldrFile, false /* locIsFake */);
+    }
+
+    /**
+     * This constructor is meant to be called only by the private constructor (with locIsFake =
+     * false) and by unit tests with fictional locale ID such as "xx" or "mul" (with locIsFake =
+     * true) for which ICUServiceBuilder.forLocale won't work
```
srl295, line 58:
```suggestion
     * false) and by unit tests with fictional locale ID such as "mul" (with locIsFake =
     * true) for which ICUServiceBuilder.forLocale won't work
```

drop reference to xx, don't use it.

srl295 (approved):
approving but we should drop references to locale id "xx" - no reason for it.

also 'mul' is the valid sandbox locale.

tools/cldr-code/src/main/java/org/unicode/cldr/icu/TestCldr.java
btangmu: I have no idea! I don't like to see a mysterious locale ID like "xx" hidden away as a string; defining a constant in LocaleNames.java is a step towards clarifying what it is. If it can be removed, so much the better. I can address that in a follow-up PR.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
btangmu: I can change the name from `locIsFake` (here, and in the ICUServiceBuilder constructor) to something more appropriate in a follow-up PR. The point here is that `ICUServiceBuilder.forLocale(CLDRLocale.getInstance(localeId))` would fail on MUL...

tools/cldr-code/src/main/java/org/unicode/cldr/tool/LanguageCodeConverter.java
btangmu: Let's discuss that for a follow-up. I don't know what "und" is for exactly, or even what it's an abbreviation for ("undetermined"? "undefined"? "undecided"?); it should have a constant of its own, presumably...

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
btangmu: Again, let's consider that (with `LocaleNames.MUL` not a string literal) for a follow-up -- such a change is somewhat out-of-scope for this ticket, which is about example dependencies...

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
srl295: > The fourth commit replaces nearly all `new ICUServiceBuilder(cldrFile)` with `ICUServiceBuilder.forLocale()` for consistency and improved performance. The only exception is for some unit tests that use fictitious locale names "xx" and "mul", with special test data, for which `forLocale` would not work.
> 
> There is a performance improvement for `TestCheckCLDR.TestAllLocales` which was taking between 91 and 107 seconds locally and now takes between 75 and 79 seconds.

It's also causing CLDR-19409 because some clients should be using SurveyTool's factory instead of Factory.make

tools/cldr-code/src/main/java/org/unicode/cldr/icu/TestCldr.java
macchiati: I wouldn't remove anything without knowing what it is doing; ULocale("xx") might just be a locale made to never overlap with a real CLDR locale.

PR #5253, opened 2025-12-27
CLDR-19169 Refinements for new keyboard layouts for Sanskrit, Tibetan, and Gandhari 

xadxura:
CLDR-19169

ALLOW_MANY_COMMITS=true

These keyboard updates implement refinements for new LDML layouts for Sanskrit, Tibetan, and Gandhari including introducing backspace transforms and better use of set/uset.

### DIFF
--- a/keyboards/3.0/pgd-Khar-t-k0-qwerty.xml
+++ b/keyboards/3.0/pgd-Khar-t-k0-qwerty.xml
@@ -3,7 +3,7 @@
   conformsTo="47" draft="contributed" >
   <info author="Andrew Glass" name="Gandhari (Phonetic)"
     layout="QWERTY" indicator="\u{10A12}\u{10A0E}" />
-  <version number="1.0.1"/>
+  <version number="1.0.2"/>
 
   <keys>
     <import base="cldr" path="47/keys-Zyyy-punctuation.xml"/>
@@ -18,7 +18,7 @@
       <key id="3"          output="\u{10A42}" /> <!-- Three -->
       <key id="4"          output="\u{10A43}" /> <!-- Four -->
       <key id="5"          output="\m{FIVE}"  /> <!-- Five -->
-      <key id="6"          output="\m{SIXX}"  /> <!-- Six -->
+      <key id="6"          output="\m{SIX}"   /> <!-- Six -->
       <key id="7"          output="\m{SEVE}"  /> <!-- Seven -->
       <key id="8"          output="\m{EIGH}"  /> <!-- Eight -->
       <key id="9"          output="\m{NINE}"  /> <!-- Nine -->
@@ -59,13 +59,13 @@
       <key id="alt_k"      output="\u{10A32}\m{A}" />
       <key id="K"          output="\u{10A11}\m{A}" />
       <key id="q"          output="\u{10A11}\m{A}" />
-      <key id="Q"          output="\u{10A11}\m{A}" />
+      <key id="Q"          output="\u{10A11}\m{A}" /> <!--intentional duplicate -->
       <key id="g"          output="\u{10A12}\m{A}" />
       <key id="G"          output="\u{10A13}\m{A}" />
       <key id="c"          output="\u{10A15}\m{A}" />
       <key id="C"          output="\u{10A16}\m{A}" />
       <key id="j"          output="\u{10A17}\m{A}" />
-      <key id="J"          output="\u{10A17}\m{A}" />
+      <key id="J"          output="\u{10A17}\m{A}" /> <!--intentional duplicate -->
       <key id="alt_y"      output="\u{10A19}\m{A}" />
       <key id="alt_t"      output="\u{10A1A}\m{A}" />
       <key id="alt_T"      output="\u{10A1B}\m{A}" />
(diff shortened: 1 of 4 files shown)


miloush:
We should probably have some guidance/policy on when to use codepoints (possibly with comments) and when to use characters themselves, these are difficult to review just in code.

srl295:
> We should probably have some guidance/policy on when to use codepoints (possibly with comments) and when to use characters themselves, these are difficult to review just in code.

As usual, it depends on fonts, layout engines, RTL status, etc as to which is easier. 

Tool wise, we could have something like cldrmodify which automatically enforces such a policy.

The simplest policy is of course to leave it to author's discretion.

keyboards/3.0/pgd-Khar-t-k0-qwerty.xml
```diff
@@ -59,13 +59,13 @@
       <key id="alt_k"      output="\u{10A32}\m{A}" />
       <key id="K"          output="\u{10A11}\m{A}" />
       <key id="q"          output="\u{10A11}\m{A}" />
-      <key id="Q"          output="\u{10A11}\m{A}" />
+      <key id="Q"          output="\u{10A11}\m{A}" /> <!--intentional duplicate -->
```
srl295: Perhaps something such as "q and Q produce the same output"?

For that matter, why not just use the same key in two spots in the layout?

keyboards/3.0/pgd-Khar-t-k0-qwerty.xml
```diff
@@ -77,49 +77,60 @@
...
+      <key id="open-angle"  output="\u{10A55}" />
+      <key id="pipe"        output="\u{10A56}" />
+      <key id="alt_pipe"    output="\u{10A57}" />
+      <key id="equal"       output="\u{10A58}" />
   </keys>
 
   <layers formId="us">
     <layer modifiers="none">
       <row keys="grave 1 2 3 4 5 6 7 8 9 0 barabove equal"/>
       <row keys="q w e r t y u i o p open-square close-square backslash"/>
```
srl295, line 131: This could have `Q` etc.

keyboards/3.0/pgd-Khar-t-k0-qwerty.xml
mcdurdin:
```suggestion
      <key id="Q"          output="\u{10A11}\m{A}" /> <!-- q and Q intentionally produce the same output -->
```

This is easier to document in the key bag than in the row, where the case difference in the Q key would be confusing. And means that the duplication only needs to be tracked in one place (it is harder to keep all forms consistent otherwise).

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
```diff
@@ -3,112 +3,120 @@
   conformsTo="47" draft="contributed" >
   <info author="Andrew Glass" name="Sanskrit (Phonetic)"
     layout="QWERTY" indicator="\u{0938}\u{0902}" />
-  <version number="1.0.1"/>
+  <version number="1.0.2"/>
   <keys>
     <import base="cldr" path="47/keys-Zyyy-punctuation.xml"/>
     <import base="cldr" path="47/keys-Zyyy-currency.xml"/>
+    <!-- \m{A} — Holds the inherent vowel during normal input -->
+    <!-- \m{B} — The inherent vowel that arrises from deleting a dependent vowel -->
```
mcdurdin:
```suggestion
    <!-- \m{B} — The inherent vowel that arises from deleting a dependent vowel -->
```

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
```diff
@@ -152,17 +160,23 @@
...
     <transformGroup> <!-- Avagraha -->
-      <transform from="\u{0947} \u{0905}" to="\u{0947} \u{093D}" />
-      <transform from="\u{094B} \u{0905}" to="\u{094B} \u{093D}" />
+      <transform from="\u{0947} \u{0905}" to="\u{0947} \u{093D}" /> <!-- े अ >>> े ऽ -->
+      <transform from="\u{094B} \u{0905}" to="\u{094B} \u{093D}" /> <!-- ो अ >>> ो ऽ -->
     </transformGroup>
 
     <transformGroup> <!-- Candrabindu -->
-      <transform from="\u{0020}\u{0901}"  to="\u{A8F2}" />
+      <transform from="\u{0020}\u{0901}"  to="\u{A8F2}" />
```
mcdurdin: Editor settings should trim line endings, to reduce whitespace noise in PRs.

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
```diff
@@ -197,58 +216,69 @@
...
-      <transform from="\u{094D}\u{0960}" to="\u{0944}" />
-      <transform from="\u{094D}\u{090C}" to="\u{0962}" />
-      <transform from="\u{094D}\u{0961}" to="\u{0963}" />
-      <transform from="\u{094D}\u{090F}" to="\u{0947}" />
-      <transform from="\u{094D}\u{0910}" to="\u{0948}" />
-      <transform from="\u{094D}\u{0913}" to="\u{094B}" />
-      <transform from="\u{094D}\u{0914}" to="\u{094C}" />
+      <transform from="\u{094D}($[fullvowels])" to="$[1:dependentvowels]" />
+      <transform from="\m{B}($[fullvowels])" to="$[1:dependentvowels]" />
+      <transform from="($[consonants])\u{0906}" to="$1\u{093E}" /> <!--Not working with map-->
```
mcdurdin: Can you explain the comment 'Not working with map'?

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
```diff
@@ -197,58 +216,69 @@
...
+      <transform from="\m{SUPER}\u{096E}" to="\u{A8E8}" />          <!-- 8 -->
+      <transform from="\m{SUPER}\u{096F}" to="\u{A8E9}" />          <!-- 9 -->
+      <transform from="\m{SUPER}\u{0905}" to="\u{A8EA}" />          <!-- A -->
+      <transform from="\m{SUPER}\u{0909}" to="\u{A8EB}" />          <!-- U -->
+      <transform from="\m{SUPER}\u{0915}\u{094D}" to="\u{A8EC}" />  <!-- KA -->
+      <transform from="\m{SUPER}\u{0928}\u{094D}" to="\u{A8ED}" />  <!-- NA -->
+      <transform from="\m{SUPER}\u{092A}\u{094D}" to="\u{A8EE}" />  <!-- PA -->
+      <transform from="\m{SUPER}\u{0930}\u{094D}" to="\u{A8EF}" />  <!-- RA -->
+      <transform from="\m{SUPER}\u{0935}\u{094D}" to="\u{A8F0}" />  <!-- VI -->
+      <transform from="\m{SUPER}\u{0027}" to="\u{A8F1}" />          <!-- Avagraha -->
```
mcdurdin: These should be reducible to a single transform rule with mapping

keyboards/3.0/xct-Tibt-t-k0-qwerty.xml
```diff
@@ -51,105 +51,105 @@
...
+      <key id="a"            output="\u{0F68}" />
+      <key id="A"            output="\u{0F68}\u{0F71}" />
+      <key id="alt_a"        output="\u{0F68}\u{0F71}" />
+      <key id="alt_A"        output="\u{0F68}\u{0F71}" />
+      <key id="i"            output="\u{0F68}\u{0F72}" />
+      <key id="I"            output="\u{0F68}\u{0F71}\u{0F72}" />
+      <key id="alt_i"        output="\u{0F68}\u{0F71}\u{0F72}" />
+      <key id="alt_I"        output="\u{0F68}\u{0F71}\u{0F72}" />
+      <key id="u"            output="\u{0F68}\u{0F74}" />
+      <key id="U"            output="\u{0F68}\u{0F7}\u{0F74}" />
```
mcdurdin: Clearly missing a digit `\u{0F7}`?

mcdurdin (changes requested):
One clear bug, some other comments.

One of the challenges with these data files is that without deep knowledge of the script, it's very hard to validate accuracy. Ideally, each would include a spec layout that any Unicode-aware engineer could compare it to, alongside a description of what the transforms are supposed to achieve. This would work best if it was accompanied by tests (which makes me think that the Keyman team's plan to write test tooling for keyboards would be very useful, and gives me some ideas on how we could structure that).

For example, the [Khmer LDML keyboard](https://github.com/keymanapp/keyboards/blob/fd0479ceed428316b4aa1f3b7bc97bef40c17a7a/experimental/k/khmer_ldml/source/khmer_ldml.xml) is implemented in reference to [UTN#61 Khmer Encoding Structure](https://www.unicode.org/notes/tn61/utn61-Khmer_Encoding_Structure_V2.pdf) spec, which makes it possible for any engineer to validate, with enough patience. That's kinda my current 'gold standard' -- or it will be once we can automatically test it.

keyboards/3.0/pgd-Khar-t-k0-qwerty.xml
xadxura: Yes, I felt this was easier to track in the keybag than the layout where the comment is clearer.

xadxura:
> We should probably have some guidance/policy on when to use codepoints (possibly with comments) and when to use characters themselves, these are difficult to review just in code.

Really, having both is clearest. If only one, the code points are easiest to read with confidence.

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
xadxura: @mcdurdin, I had tried collapsing these by extending the vowel groups:
`<transform from="$[consonants])($[fullvowels])" to="$1$[2:dependentvowels]" />`

AEApple:
This doesn't seem like part of the spec? Why is LDML in the title of the related issue?

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
xadxura: Adjusted my settings.

keyboards/3.0/xct-Tibt-t-k0-qwerty.xml
xadxura: Good catch, thank you!

keyboards/3.0/sa-Deva-t-k0-qwerty.xml
```diff
@@ -249,24 +251,8 @@
...
-      <transform from="\m{SUPER}\u{096F}" to="\u{A8E9}" />          <!-- 9 -->
-      <transform from="\m{SUPER}\u{0905}" to="\u{A8EA}" />          <!-- A -->
-      <transform from="\m{SUPER}\u{0909}" to="\u{A8EB}" />          <!-- U -->
-      <transform from="\m{SUPER}\u{0915}\u{094D}" to="\u{A8EC}" />  <!-- KA -->
-      <transform from="\m{SUPER}\u{0928}\u{094D}" to="\u{A8ED}" />  <!-- NA -->
-      <transform from="\m{SUPER}\u{092A}\u{094D}" to="\u{A8EE}" />  <!-- PA -->
-      <transform from="\m{SUPER}\u{0930}\u{094D}" to="\u{A8EF}" />  <!-- RA -->
-      <transform from="\m{SUPER}\u{0935}\u{094D}" to="\u{A8F0}" />  <!-- VI -->
-      <transform from="\m{SUPER}\u{0027}" to="\u{A8F1}" />          <!-- Avagraha -->
+      <transform from="\m{A}\m{SUPER}($[cantillation_bases])" to="$[1:cantillation_marks]" />
```
mcdurdin, line 254: This is an additional change to the keyboard behavior in this commit; I guess it is just tidying up the inherent vowel marker in case of backspace later?

mcdurdin (dismissed):
LGTM

srl295:
this will need to merge fairly soon to make it to CLDR 49

srl295:
ping?

mcdurdin:
What are we waiting on here? Just another review?

mcdurdin (approved):
Changes since first review look good; especially appreciate the Gandhari documentation -- great work.

PR #5350, opened 2026-02-10
CLDR-19226 Add more specificity around using appendItems in date+time patterns

sffc:
CLDR-19226

This adds the spec. We should update the CLDR code and add lots and lots _and lots_ of tests, since this algorithm has lots of edge cases. I think this code uses ICU4J DateTimePatternGenerator so we would need to either implement this over there or make a new DTPG in CLDR code before we can add more test data. (Thought: An AI might be able to do a decent job of generating a cleanroom DTPG in CLDR Java.)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -838,7 +838,11 @@ Finally: If the requested skeleton included both seconds and fractional seconds 
 If a client-requested set of fields includes both date and time fields, and if the `availableFormats` data does not include a `dateFormatItem` whose skeleton matches the same set of fields, then the request should be handled as follows:
 
 1. Divide the request into a date fields part and a time fields part.
+    * Date fields are: [year](#dfst-year), [month](#dfst-month), [day](#dfst-day), [era](#dfst-era), [week](#dfst-week), [quarter](#dfst-quarter), and [week day](#dfst-weekday).
+    * Time fields are: [hour](#dfst-hour), [minute](#dfst-minute), [second](#dfst-second), [period](#dfst-period), and [zone](#dfst-zone).
 2. For each part, find the matching `dateFormatItem`, and expand the pattern as above.
+    * If there is still no `dateFormatItem` whose skeleton matches the same set of fields, select the one with the greatest number of matching fields (but no extra fields), then use `appendItems` to append any missing fields (see below).
+    * If multiple `dateFormatItem`s with missing fields have the same distance, rank them by their matching fields in the order listed in step 1. For example, if the request is for "HBv", and the locale has `dateFormatItem`s for only "HB" and "Hv", select the "HB" pattern, because "B" has a higher weight than "v", and then use the `appendItem` for "v" (time zone).
 3. Combine the patterns for the two `dateFormatItem`s using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
    * If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
    * Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
@@ -852,6 +856,8 @@ If a client-requested set of fields includes both date and time fields, and if t
 ```
 
 In case the best match does not include all the requested calendar fields, the `appendItems` element describes how to append needed fields to one of the existing formats. Each `appendItem` element covers a single calendar field. In the pattern, {0} represents the format string, {1} the data content of the field, and {2} the display name of the field (see [Calendar Fields](#Calendar_Fields)).
+
+Note: as described above `appendItems` for date fields should be appended to the date, and `appendItems` for time fields should be appended to the time, _before_ combining them with the `dateTimeFormat`.
 
 #### <a name="intervalFormats" href="#intervalFormats">Element intervalFormats</a>
 
--- a/docs/ldml/tr35-modifications.md
+++ b/docs/ldml/tr35-modifications.md
@@ -43,6 +43,10 @@ This is a partial document, describing only the changes to the LDML since the pr
 * [Locale Display Name Algorithm](tr35-general.html#locale_display_name_algorithm) updated to use the nested bracket replacement data and avoid nested parentheses by flattening `-t-` (transform) language names
 * The section "Enhanced Language Matching" is retitled as [Language Matching Variables](tr35.html#enhanced-language-matching) and clarified.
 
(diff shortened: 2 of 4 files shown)


pedberg-icu:
Need to add something to the modifications section

docs/ldml/tr35-dates.md
```diff
@@ -838,7 +838,11 @@ Finally: If the requested skeleton included both seconds and fractional seconds
 If a client-requested set of fields includes both date and time fields, and if the `availableFormats` data does not include a `dateFormatItem` whose skeleton matches the same set of fields, then the request should be handled as follows:
 
 1. Divide the request into a date fields part and a time fields part.
+    * Date fields are: [year](#dfst-year), [month](#dfst-month), [day](#dfst-day), [era](#dfst-era), [week](#dfst-week), [quarter](#dfst-quarter), and [week day](#dfst-weekday).
```
macchiati, line 841: This is not a complete set, and dow is a different beast, shared between them.

Field | Pattern letters | Grouping
-- | -- | --
era | "G" | date
year | "yYurU" | date
quarter | "Qq" | date
month | "MLl" | date
week_of_year | "w" | date
week_of_month | "W" | date
day | "dg" | date
day_of_year | "D" | date
day_of_week_in_month | "F" | date
**weekday** | "Ece" | **weekday**
dayperiod | "abB" | time
hour | "HkhK" | time
minute | "m" | time
second | "sA" | time
fractional_second | "S" | time
zone | "vzZOVXx" | zone

macchiati: We don't currently have data for week_of… and day_of_…; but if we were to add them they would go in dates. Zone is also a bit odd, and in a way shared.

docs/ldml/tr35-dates.md
```diff
@@ -838,7 +838,11 @@ Finally: If the requested skeleton included both seconds and fractional seconds
...
+    * Time fields are: [hour](#dfst-hour), [minute](#dfst-minute), [second](#dfst-second), [period](#dfst-period), and [zone](#dfst-zone).
 2. For each part, find the matching `dateFormatItem`, and expand the pattern as above.
+    * If there is still no `dateFormatItem` whose skeleton matches the same set of fields, select the one with the greatest number of matching fields (but no extra fields), then use `appendItems` to append any missing fields (see below).
+    * If multiple `dateFormatItem`s with missing fields have the same distance, rank them by their matching fields in the order listed in step 1. For example, if the request is for "HBv", and the locale has `dateFormatItem`s for only "HB" and "Hv", select the "HB" pattern, because "B" has a higher weight than "v", and then use the `appendItem` for "v" (time zone).
```
macchiati, line 845: The ICU code has a series of distances that let us do a better much better job of matching. 
- Suppose you ask for Ehmv and there is only Ehm or hmv — ordinal order might not be the best choice.
- Even if you have the same codes, we also need to decide among the choices base on lengths or kinds. In particular, a match between a numeric field (like M) and a non-numeric (like MMMM) is a much larger difference than between d and dd.

We can probably do a simpler job than ICU's weight table, but we should not omit it.

macchiati (changes requested):

docs/ldml/tr35-dates.md
sffc: The set is complete. It links to every rowspan in the Date Field Symbol Table. For example, the "week" rowspan in the table covers both week-of-year and week-of-month.

I intentionally put Zone and DOW into the Time and Date categories because that's where the algorithm is currently ambiguous. The primary objective of this PR is to harden the ambiguity. Based on your research and discussion earlier, it seems that we can be confident putting Zone in the time group and DOW in the date group.

docs/ldml/tr35-dates.md
sffc: If you ask for Ehmv and there is only Ehm and hmv, then this algorithm states fairly clearly, both before and after, that you should split the E from the hmv and glue them together using the dateTimeFormat.

sffc: > Even if you have the same codes, we also need to decide among the choices base on lengths or kinds. In particular, a match between a numeric field (like M) and a non-numeric (like MMMM) is a much larger difference than between d and dd.

The length distances are already specified a couple sections up. You only get to this section of the spec if you already tried everything and you weren't able to get a match on field set.

sffc:
I discussed this with @macchiati and my approach is fine because

1. It is a fallback for weird cases
2. The field order proposed here works for a large majority of locales 
3. Locales can include skeletons that prevent the fallback from being reached

pedberg-icu:
@sffc Spec changes look good but can you also update the spec date (i.e. the Date entry in the table near the top of docs/ldml/tr35.md)? Thanks. If that is done I will approve.

docs/ldml/tr35-modifications.md
```diff
@@ -43,6 +43,9 @@ This is a partial document, describing only the changes to the LDML since the pr
 * [Locale Display Name Algorithm](tr35-general.html#locale_display_name_algorithm) updated to use the nested bracket replacement data and avoid nested parentheses by flattening `-t-` (transform) language names
 * The section "Enhanced Language Matching" is retitled as [Language Matching Variables](tr35.html#enhanced-language-matching) and clarified.
 
+**Changes in LDML Version 49 (Differences from Version 48.2)**
+* Clarified the process of selecting the best `dateFormatItem` when there is no exact match, and how to use `appendItems` to add missing fields. This includes a clarification of what are date fields and what are time fields, and a note that `appendItems` for date and time fields should be appended before combining them.
+
```
pedberg-icu, line 49:
```suggestion
<!-- Updated spec date -->

```

pedberg-icu (approved):

pedberg-icu:
@macchiati It sounds from Shane's comment 2 weeks ago that he discussed the latest version with you and your objections were resolved. If so, can you approve this? Or @sffc if that is true you can also dismiss the blocking review.

macchiati (changes requested):
This is still a work in progress; separately Shane is working on a reference implementation of the spec. That is turning up more ambiguities in the spec, so we should delay this PR until we can incorporate any changes from that.

sffc:
I'm not aware of any remaining action items on this PR. I believe I resolved all the open comments. There is a merge conflict I need to resolve after @macchiati approves it.

Once #5394 is landed, we can also land code changes and tests for this spec change.

sffc:
I don't believe that any code changes are necessary.

I could add tests, but I should follow up with a PR that adds a lot more conformance tests for DTPG anyway.

sffc:
Integrated into https://github.com/unicode-org/cldr/pull/5660

PR #5394, opened 2026-02-24
CLDR-19258 Add a clean CLDR implementation of DateTimePatternGenerator

sffc:
CLDR-19258

- [ ] This PR completes the ticket. (not yet; we should migrate more call sites to use it)

---

I used Gemini 3.1 Pro to write most of this code. My prompt:

> In DateTimeFormats.java, it imports DateTimePatternGenerator from ICU4J. Please write a new DateTimePatternGenerator that exactly follows the specification, replace the imports in DateTimeFormats.java, and ensure the relevant tests pass, such as GenerateDateTimeTestData. You can find the specification in tr35-dates.md under the heading "Elements availableFormats, appendItems".

After the first round, it produced working code, but it was failing some tests. It said its code was more correct than the old code, which I verified was incorrect. My second prompt:

> You are wrong: (1) The short dateTimeFormat in English Gregorian has a comma. (2) The skeleton has a j, which means select the default hour cycle, which is h23 for Japanese. (3) The appendItem for time zone in Arabic inherits from root, which does not have ├ markers. I committed your java code but not the test output. Please fix the errors in the java code.

It worked for another hour and eventually made code that passed the test. It seems the bug was using `getWinningValue` instead of `getStringValueWithBailey`.

Then I asked it to add more comments and docs.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrDateTimePatternGeneratorCompare.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrDateTimePatternGeneratorCompare.java
@@ -1,0 +1,403 @@
+package org.unicode.cldr.tool;
+
+import com.ibm.icu.text.DateTimePatternGenerator;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+import org.unicode.cldr.draft.FileUtilities;
+import org.unicode.cldr.util.CLDRConfig;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.CLDRPaths;
+import org.unicode.cldr.util.CLDRTool;
+import org.unicode.cldr.util.CLDRURLS;
+import org.unicode.cldr.util.CldrDateTimePatternGenerator;
+import org.unicode.cldr.util.Factory;
+
+/**
+ * Compares the behavior of CldrDateTimePatternGenerator against ICU4J's DateTimePatternGenerator.
+ */
+@CLDRTool(
+        alias = "compare-cldr-dtpg",
+        description =
+                "Compares CldrDateTimePatternGenerator against ICU4J DateTimePatternGenerator",
+        url = CLDRURLS.TOOLSURL)
+public class CldrDateTimePatternGeneratorCompare {
+    private static final String[] SKELETONS = {
+        // Era (G): 1..5
+        "G",
+        "GG",
+        "GGG",
+        "GGGG",
+        "GGGGG",
+        "GyMd",
+        "GGyMd",
+        "GGGyMd",
+        "GGGGyMd",
+        "GGGGGyMd",
+
+        // Year (y): 1..4
+        "y",
+        "yy",
+        "yyyy",
+        "yMd",
+        "yyMd",
+        "yyyyMd",
+
+        // Month (M): 1..5
+        "M",
+        "MM",
+        "MMM",
+        "MMMM",
+        "MMMMM",
+        "yM",
+        "yMM",
+        "yMMM",
+        "yMMMM",
+        "yMMMMM",
+
+        // Day (d): 1..2
+        "d",
+        "dd",
+        "yMd",
+        "yMdd",
+
+        // Day of week (E): 1..6
+        // Skip EE and EEE since they canonicalize to E
+        "E",
+        "EEEE",
+        "EEEEE",
+        "EEEEEE",
+        "yMdE",
+        "yMdEEEE",
+        "yMdEEEEE",
+        "yMdEEEEEE",
+
+        // Hour 1-12 (h): 1..2
+        "h",
+        "hh",
+        "hm",
+        "hhm",
+
+        // Hour 0-23 (H): 1..2
+        "H",
+        "HH",
+        "Hm",
+        "HHm",
+
+        // Hour preferred (j): 1..6
+        "j",
+        "jj",
+        "jjj",
+        "jjjj",
+        "jjjjj",
+        "jjjjjj",
+        "jm",
+        "jjm",
+        "jjjm",
+        "jjjjm",
+        "jjjjjm",
+        "jjjjjjm",
+
+        // Hour preferred, no am/pm (J): 1..6
+        "J",
+        "JJ",
+        "JJJ",
+        "JJJJ",
+        "JJJJJ",
+        "JJJJJJ",
+        "Jm",
+        "JJm",
+        "JJJm",
+        "JJJJm",
+        "JJJJJm",
+        "JJJJJJm",
+
+        // Hour preferred, context-dependent (C): 1..6
+        "C",
+        "CC",
+        "CCC",
+        "CCCC",
+        "CCCCC",
+        "CCCCCC",
+        "Cm",
+        "CCm",
+        "CCCm",
+        "CCCCm",
+        "CCCCCm",
+        "CCCCCCm",
+
+        // Minute (m): 1..2
+        "m",
+        "mm",
+        "hm",
+        "hmm",
+
+        // Second (s): 1..2
+        "s",
+        "ss",
+        "hms",
+        "hmss",
+
+        // Timezone (z): 1..5
+        "z",
+        "zz",
+        "zzz",
+        "zzzz",
+        "zzzzz",
+        "hmsz",
+        "hmszz",
+        "hmszzz",
+        "hmszzzz",
+        "hmszzzzz",
+
+        // --- Other fields not explicitly mentioned but previously included ---
+        "U",
+        "UUUU",
+        "Q",
+        "QQ",
+        "QQQ",
+        "QQQQ",
+        "QQQQQ",
+        "yQ",
+        "yQQQ",
+        "yQQQQ",
+        "B",
+        "BBBB",
+        "BBBBB",
+        "Bh",
+        "Bhh",
+        "Bhm",
+        "BBBBhm",
+        "BBBBBhm",
+        "v",
+        "vvvv",
+        "yMv",
+        "yMMMMdv",
+        "yMMMMEEEEdvvvv",
+        "yMdHmsv",
+        "yMMMMdhmsvvvv"
+    };
+
+    private static final String[] CALENDARS = {
+        "gregorian", "buddhist", "japanese", "roc", "islamic", "chinese"
+    };
+
+    public static void main(String[] args) throws IOException {
+        String filter = args.length > 0 ? args[0] : ".*";
+
+        CLDRConfig config = CLDRConfig.getInstance();
+        Factory factory = config.getCldrFactory();
+        Set<String> locales = factory.getAvailableLanguages();
+
+        String outputDir = CLDRPaths.CHART_DIRECTORY + "/verify/dates/";
+        java.io.File dir = new java.io.File(outputDir);
+        if (!dir.exists()) {
+            dir.mkdirs();
+        }
+        String filename = "dtpg_comparison.csv";
+
+        System.out.println(
+                "Generating comparison to " + outputDir + filename + " with filter " + filter);
+
+        try (PrintWriter out = FileUtilities.openUTF8Writer(outputDir, filename)) {
+            // Header
+            out.println(
+                    "Locale,Calendar,Skeleton,CldrDTPG Pattern,ICU4J DTPG Pattern,Match?,Trace");
+
+            int count = 0;
+            List<String> sortedLocales = new ArrayList<>(locales);
+            Collections.sort(sortedLocales);
+
+            for (String localeID : sortedLocales) {
+                if (localeID.equals("root")) continue;
+                if (!localeID.matches(filter)) continue;
+
+                System.out.println("Processing locale: " + localeID);
+                CLDRFile cldrFile = factory.make(localeID, true);
+
+                boolean localeHasDiffs = false;
+                List<String> rowsForThisLocale = new ArrayList<>();
+
+                for (String calendar : CALENDARS) {
+                    CldrDateTimePatternGenerator cldrGen =
+                            new CldrDateTimePatternGenerator(cldrFile, calendar, false);
+
+                    DateTimePatternGenerator icuGen = cldrGen.getIcu4jGenerator();
+
+                    for (String skeleton : SKELETONS) {
+                        List<String> trace = new ArrayList<>();
+                        String cldrPattern = cldrGen.getBestPattern(skeleton, trace);
+                        String icuPattern = normalizePattern(icuGen.getBestPattern(skeleton));
+
+                        String matchResult = getDifferenceReason(cldrPattern, icuPattern);
+                        if (!matchResult.equals("YES")) {
+                            localeHasDiffs = true;
+                            String traceStr = String.join(" | ", trace);
+                            StringBuilder sb = new StringBuilder();
+                            sb.append(escapeCsv(localeID)).append(",");
+                            sb.append(escapeCsv(calendar)).append(",");
+                            sb.append(escapeCsv(skeleton)).append(",");
+                            sb.append(escapeCsv(cldrPattern)).append(",");
+                            sb.append(escapeCsv(icuPattern)).append(",");
+                            sb.append(escapeCsv(matchResult)).append(",");
+                            sb.append(escapeCsv(traceStr));
+                            rowsForThisLocale.add(sb.toString());
+                        }
+                    }
+                }
+
+                if (localeHasDiffs) {
+                    for (String row : rowsForThisLocale) {
+                        out.println(row);
+                        count++;
+                    }
+                } else {
+                    out.print(escapeCsv(localeID));
+                    out.print(",");
+                    out.print(escapeCsv("ALL"));
+                    out.print(",");
+                    out.print(escapeCsv("ALL"));
+                    out.print(",");
+                    out.print(escapeCsv(""));
+                    out.print(",");
+                    out.print(escapeCsv(""));
+                    out.print(",");
+                    out.print(escapeCsv("YES"));
+                    out.print(",");
+                    out.print(escapeCsv(""));
+                    out.println();
+                    count++;
+                }
+                out.flush();
+            }
+            System.out.println("Done! Generated " + count + " comparisons.");
+        }
+    }
+
+    private static String getDifferenceReason(String cldr, String icu) {
+        if (cldr.equals(icu)) return "YES";
+
+        List<String> cldrFields = getFields(cldr);
+        List<String> icuFields = getFields(icu);
+
+        // 1. Check for Era mismatch
+        boolean cldrHasG = cldrFields.stream().anyMatch(f -> f.startsWith("G"));
+        boolean icuHasG = icuFields.stream().anyMatch(f -> f.startsWith("G"));
+        if (cldrHasG != icuHasG) {
+            return icuHasG ? "NO: Era added" : "NO: Era removed";
+        }
+
+        // 2. Check for Field Set mismatch (excluding Era)
+        Set<Character> cldrFieldSet = new TreeSet<>();
+        for (String f : cldrFields) if (!f.startsWith("G")) cldrFieldSet.add(f.charAt(0));
+        Set<Character> icuFieldSet = new TreeSet<>();
+        for (String f : icuFields) if (!f.startsWith("G")) icuFieldSet.add(f.charAt(0));
+
+        if (!cldrFieldSet.equals(icuFieldSet)) {
+            StringBuilder cldrOnly = new StringBuilder();
+            for (char ch : cldrFieldSet) {
+                if (!icuFieldSet.contains(ch)) cldrOnly.append(ch);
+            }
+            StringBuilder icuOnly = new StringBuilder();
+            for (char ch : icuFieldSet) {
+                if (!cldrFieldSet.contains(ch)) icuOnly.append(ch);
+            }
+            return "NO: Field set mismatch: " + cldrOnly + " -> " + icuOnly;
+        }
+
+        // 3. Check for Field Order mismatch
+        List<Character> cldrOrder = new ArrayList<>();
+        for (String f : cldrFields) cldrOrder.add(f.charAt(0));
+        List<Character> icuOrder = new ArrayList<>();
+        for (String f : icuFields) icuOrder.add(f.charAt(0));
+
+        if (!cldrOrder.equals(icuOrder)) {
+            return "NO: Field order mismatch";
+        }
+
+        // 4. Check for Field Length mismatch
+        if (cldrFields.size() == icuFields.size()) {
+            for (int i = 0; i < cldrFields.size(); i++) {
+                String cF = cldrFields.get(i);
+                String iF = icuFields.get(i);
+                if (!cF.equals(iF)) {
+                    return "NO: Field length mismatch: " + cF + " -> " + iF;
+                }
+            }
+        }
+
+        // 5. If everything else is the same, it must be literals or separators
+        return "NO: Separator/Literal mismatch";
+    }
+
+    private static List<String> getFields(String pattern) {
+        List<String> fields = new ArrayList<>();
+        boolean inQuote = false;
+        for (int i = 0; i < pattern.length(); ) {
+            char ch = pattern.charAt(i);
+            if (ch == '\'') {
+                inQuote = !inQuote;
+                i++;
+                continue;
+            }
+            if (inQuote) {
+                i++;
+                continue;
+            }
+            if ((ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z')) {
+                int start = i;
+                while (i < pattern.length() && pattern.charAt(i) == ch) {
+                    i++;
+                }
+                fields.add(pattern.substring(start, i));
+            } else {
+                i++;
+            }
+        }
+        return fields;
+    }
+
+    private static String normalizePattern(String pattern) {
+        if (pattern == null) return null;
+        StringBuilder sb = new StringBuilder();
+        boolean inQuote = false;
+        for (int i = 0; i < pattern.length(); ) {
+            char ch = pattern.charAt(i);
+            if (ch == '\'') {
+                sb.append(ch);
+                inQuote = !inQuote;
+                i++;
+            } else if (inQuote) {
+                sb.append(ch);
+                i++;
+            } else if (ch == 'E') {
+                int count = 0;
+                while (i < pattern.length() && pattern.charAt(i) == 'E') {
+                    count++;
+                    i++;
+                }
+                if (count <= 3) {
+                    sb.append("E");
+                } else {
+                    for (int j = 0; j < count; j++) sb.append('E');
+                }
+            } else {
+                sb.append(ch);
+                i++;
+            }
+        }
+        return sb.toString();
+    }
+
+    private static String escapeCsv(String s) {
+        if (s == null) return "";
+        if (s.contains(",") || s.contains("\"") || s.contains("\n")) {
+            return "\"" + s.replace("\"", "\"\"") + "\"";
+        }
+        return s;
+    }
+}
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
@@ -1,0 +1,1107 @@
+package org.unicode.cldr.util;
+
+import com.ibm.icu.text.DateFormat;
+import com.ibm.icu.text.DateTimePatternGenerator;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A generator that produces the best localized date/time pattern for a given skeleton.
+ *
+ * <p>This implementation strictly follows the LDML TR35 specification for DateTimePatternGenerator,
+ * utilizing CLDRFile for direct data access and implementing the Matching Skeletons algorithm.
+ */
+public class CldrDateTimePatternGenerator {
+    private final CLDRFile file;
+    private final String calendarID;
+    private final boolean useStock;
+
+    private char defaultHourFormatChar = 'H';
+    private String[] allowedHourFormats = {"H"};
+    private String decimal = "?";
+    private final Map<String, String> availableFormats = new LinkedHashMap<>();
+    private final Map<String, String> appendItems = new LinkedHashMap<>();
+    private final Map<String, String> fieldNames = new LinkedHashMap<>();
+
+    private String dateTimeFormatFull = "{1} {0}";
+    private String dateTimeFormatLong = "{1} {0}";
+    private String dateTimeFormatMedium = "{1} {0}";
+    private String dateTimeFormatShort = "{1} {0}";
+
+    private static final String[] STOCK = DateTimeFormats.STOCK;
+
+    /**
+     * The canonical order of date/time fields as defined by the LDML specification (TR35).
+     * Skeletons are normalized to this order to ensure consistent matching.
+     */
+    private static final String CANONICAL_ORDER = DateTimeFormats.CANONICAL_ORDER;
+
+    /** Characters representing date fields. */
+    private static final String DATE_FIELDS = DateTimeFormats.DATE_FIELDS;
+
+    /** Characters representing time fields. */
+    private static final String TIME_FIELDS = DateTimeFormats.TIME_FIELDS;
+
+    /** Characters that are always numeric. */
+    private static final String ALWAYS_NUMERIC_FIELDS = DateTimeFormats.ALWAYS_NUMERIC_FIELDS;
+
+    /** Characters that are numeric if length is 1 or 2, and text otherwise. */
+    private static final String NUMERIC_OR_TEXT_FIELDS = DateTimeFormats.NUMERIC_OR_TEXT_FIELDS;
+
+    /**
+     * Sets of related field characters that represent the same semantic field (e.g., M and L for
+     * Month).
+     */
+    private static final String[] RELATED_FIELD_SETS = DateTimeFormats.RELATED_FIELD_SETS;
+
+    private static final Map<Character, String> RELATED_CHAR_MAP = new HashMap<>();
+
+    static {
+        for (String set : RELATED_FIELD_SETS) {
+            for (int i = 0; i < set.length(); i++) {
+                RELATED_CHAR_MAP.put(set.charAt(i), set);
+            }
+        }
+    }
+
+    /**
+     * Constructs a new generator for the given CLDRFile and calendar.
+     *
+     * @param file the CLDRFile to read data from (should be a resolved file for proper inheritance)
+     * @param calendarID the ID of the calendar (e.g., "gregorian", "japanese")
+     * @param useStock if true, the generator will also include the standard stock date and time
+     *     formats (short, medium, long, full) as defined in the dateFormats/timeFormats sections
+     *     when matching skeletons.
+     */
+    public CldrDateTimePatternGenerator(CLDRFile file, String calendarID, boolean useStock) {
+        this.file = file;
+        this.calendarID = calendarID;
+        this.useStock = useStock;
+        init();
+    }
+
+    /**
+     * Initializes the generator by loading preferred hour formats, stock patterns, field display
+     * names, and available format patterns from the CLDR data.
+     */
+    private void init() {
+        initHourFormat();
+        initDecimal();
+        initStockPatterns();
+
+        Set<String> allIds = new LinkedHashSet<>();
+        Set<String> allAppendRequests = new LinkedHashSet<>();
+
+        collectFormatIdsAndAppendRequests(allIds, allAppendRequests);
+        initFieldNames();
+        initAvailableFormats(allIds);
+        initAppendItems(allAppendRequests);
+        initDateTimeFormats();
+
+        // Sort available formats to eliminate tie-breaking at runtime
+        sortAvailableFormats();
+    }
+
+    /**
+     * Determines the preferred hour format (h, H, K, or k) for the locale from supplemental data.
+     */
+    private void initHourFormat() {
+        // TR35, Section 3.8.3: "it ['j'] requests the preferred hour format for the locale
+        // (h, H, K, or k), as determined by the preferred attribute of the hours element
+        // in supplemental data."
+        SupplementalDataInfo sdi = SupplementalDataInfo.getInstance();
+        String localeID = file.getLocaleID();
+        PreferredAndAllowedHour pref = sdi.getTimeData().get(localeID);
+        if (pref == null) {
+            CLDRLocale loc = CLDRLocale.getInstance(localeID);
+            String region = loc.getCountry();
+            if (region == null || region.isEmpty()) {
+                CLDRLocale max = loc.getMaximal();
+                if (max != null) {
+                    region = max.getCountry();
+                }
+            }
+            pref = sdi.getTimeData().get(region);
+        }
+        if (pref == null) {
+            pref = sdi.getTimeData().get("001");
+        }
+        if (pref != null) {
+            defaultHourFormatChar = pref.preferred.toString().charAt(0);
+            allowedHourFormats = pref.allowed.stream().map(Object::toString).toArray(String[]::new);
+        }
+    }
+
+    private void initDecimal() {
+        String ns = file.getStringValueWithBailey("//ldml/numbers/defaultNumberingSystem");
+        if (ns == null) {
+            ns = "latn";
+        }
+        String decimalPath = "//ldml/numbers/symbols[@numberSystem='" + ns + "']/decimal";
+        String decimalValue = file.getStringValueWithBailey(decimalPath);
+        if (decimalValue != null) {
+            this.decimal = decimalValue;
+        }
+    }
+
+    /**
+     * If useStock is true, includes standard stock date and time patterns (short, medium, long,
+     * full) in the available formats.
+     */
+    private void initStockPatterns() {
+        if (!useStock) return;
+
+        for (String stock : STOCK) {
+            String dBase =
+                    String.format(
+                            "//ldml/dates/calendars/calendar[@type=\"%s\"]/dateFormats/dateFormatLength[@type=\"%s\"]/dateFormat[@type=\"standard\"]",
+                            calendarID, stock);
+            String tBase =
+                    String.format(
+                            "//ldml/dates/calendars/calendar[@type=\"%s\"]/timeFormats/timeFormatLength[@type=\"%s\"]/timeFormat[@type=\"standard\"]",
+                            calendarID, stock);
+
+            String dp = file.getStringValueWithBailey(dBase + "/pattern[@type=\"standard\"]");
+            String ds = file.getStringValueWithBailey(dBase + "/datetimeSkeleton");
+            if (dp != null && ds != null) {
+                availableFormats.put(canonicalizeSkeleton(ds), dp);
+            }
+
+            String tp = file.getStringValueWithBailey(tBase + "/pattern[@type=\"standard\"]");
+            String ts = file.getStringValueWithBailey(tBase + "/datetimeSkeleton");
+            if (tp != null && ts != null) {
+                availableFormats.put(canonicalizeSkeleton(ts), tp);
+            }
+        }
+    }
+
+    /** Efficiently scans the calendar subtrees for all available format IDs and append requests. */
+    private void collectFormatIdsAndAppendRequests(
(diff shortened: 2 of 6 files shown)


sffc:
@pedberg-icu Let me know if you like what I'm trying to do with this PR.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
```diff
@@ -0,0 +1,529 @@
...
+            if (val == null || (localeWhereFound.value != null && (localeWhereFound.value.equals("root") || localeWhereFound.value.equals(XMLSource.CODE_FALLBACK_ID)))) {
+                String gregPath = path.replaceFirst("\\[@type=\"[^\"]+\"\\]", "[@type=\"gregorian\"]");
+                String gregVal = file.getStringValueWithBailey(gregPath, null, localeWhereFound);
+                if (gregVal != null && localeWhereFound.value != null && !localeWhereFound.value.equals("root") && !localeWhereFound.value.equals(XMLSource.CODE_FALLBACK_ID)) {
+                    return gregVal;
+                }
+            }
+        }
+        return val;
+    }
```
sffc: Would like one of the reviewers more familiar with CLDRFile to look at this code. It is AI-generated and tries to obey the mixed vertical and horizontal fallback.

sffc (commented):
I reviewed the changes and prompted the AI to fix some things. This is ready for someone else's review now.

macchiati:
I very much like this plan. However, I'd like to see a test added for it, that compares the results over all locales between the ICU implementation and the proposed CLDR one.

It should be exhaustive: testing all the reasonable combinations of input skeletons, over all locales and calendars. While this is a large number of combinations, it should be relatively quick to program and we don't care if it takes some minutes to run.

I think the simplest way to do that is to take all of the semantic skeletons (plus all singletons) and using all the width combinations, which I think includes the following:

G - 1..4 (plus u, U, r??)
y - 1..4
M - 1..5
d - 1..2
E - 1..4
h - 1..2 (plus H)
m - 1..2 
s - 1..2
z - 1..5 (plus Z, O, v, V, X, x)

It does mean using ICUs code by overriding with CLDR data. That's fairly straightforward and I can point you to examples. 

I'd suggest starting with just gregorian and generic calendars, and first just English and Japanese, then expanding the calendars and locales.
* Where there are differences, we can then determine whether they are material or not, and either tune the algorithm or exclude them from the test before moving to more combinations.

Of course, once we replace the use of ICU's code inside of CLDR, we can drop the test at the same time.

@pedberg-icu are there other symbols used in some other calendars that we should add?

pedberg-icu:
> @pedberg-icu are there other symbols used in some other calendars that we should add?

@macchiati @sffc For Chinese calendar we should add U and r. Chinese calendar is also an interesting test because it does not inherit formats sideways, it inherits only up.

sffc:
I'm happy to add more tests, noting that:

1. Conformance is intended to cover this already; this implementation aligns for the ones covered in datetime test data (which isn't as comprehensive as I'd like it to be, but it does cover a number of edge cases)
2. It is expected that there might be slight differences between this code and the ICU4J code. The objective here is to implement the spec based on a clean reading of the spec. If ICU4J behavior differs, it does not necessarily indicate that there is a bug in the CLDR code; it could be a bug in ICU4J code or perhaps ill-defined spec language.
3. Other implementations like ICU4X also exist.

I'm assuming that you are asking more for a spreadsheet to illustrate the differences rather than a test that actually lands in CLDR. That seems like something reasonable to generate.

macchiati:
I'm saying something different. The spec ( as in some other cases) is a somewhat hand wavy capture of what ICU is doing. I'm worried that a simple implementation of the spec will not catch cases that should be caught. By comparing icu's output to your implantation., we can figure out whether any of the differences are ones that should be accounted for **by enhancing the spec _and_ your implementation.**

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
```diff
@@ -0,0 +1,697 @@
...
+    private static final String[] STOCK = {"short", "medium", "long", "full"};
+    
+    /**
+     * The canonical order of date/time fields as defined by the LDML specification (TR35).
+     * Skeletons are normalized to this order to ensure consistent matching.
+     */
+    private static final String CANONICAL_ORDER = "GyYuUrQqMLwWEecdDFgabBhHKkmsSAzZOvVXx";
+
+    /** Characters representing date fields. */
+    private static final String DATE_FIELDS = "GyYruUQqMLwWEdDFg";
```
srl295: please, not more copies of this same information…

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
srl295: :-1: Just use a resolving CLDRFile and getStringValue. All of the rest should fall out automatically.

sffc: It didn't work when I tried that. I asked AI to figure out how to get the correct data and it spent about an hour getting code that would work here.

srl295: Okay- well, this code does not belong here. I'm very -1 on propagating this kind of thing outside of CLDRFile innards.

srl295 (changes requested):
Fallbacks to gregorian are in root.xml, OR there's a bug.  

I'm -1 on the internals of the fallback mechanism being replicated in this generator.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
```diff
@@ -694,4 +733,75 @@ private String canonicalizeSkeleton(String skel) {
...
+            case "Day-Of-Week": return DateTimePatternGenerator.WEEKDAY;
+            case "Hour": return DateTimePatternGenerator.HOUR;
+            case "Minute": return DateTimePatternGenerator.MINUTE;
+            case "Second": return DateTimePatternGenerator.SECOND;
+            case "Timezone": return DateTimePatternGenerator.ZONE;
+            default: return -1;
+        }
+    }
+
+    private static int mapFieldName(String key) {
```
srl295, line 1104: we already have metadata on field names. why yet more copies of this information?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
srl295: @sffc please give an exact example (or more examples) of the precise XPath that "didn't work" when using a resolving xpath and string value.  If true, there's a data and/or code bug (and/or maybe a spec bug).  In no event does this function need to exist as written.

sffc: I took a closer look. I think what happened is it took Gemini a really long time to figure out to use the `Bailey` version of the getter functions, and then it just forgot to go and clean up the workarounds it had previously written. The tests seem to pass if I call `Bailey` functions directly. I've pushed a fix. Thanks!

sffc:
Spreadsheet illustrating the differences between the new CLDR DTPG and the ICU4J DTPG:

https://docs.google.com/spreadsheets/d/1IAcbd1ylKfUePZSbBt2cwhxDjS9eZUgHeBJ6bK-0zl0/edit?gid=1157632567#gid=1157632567

sffc:
Some quick observations on the most common types of differences:

- E vs EEE (I changed the comparison script to skip these)
- M vs L when it is the only field -- should that substitution be automatic?
- M vs MM (and d vs dd, ...) -- not sure who is right here
- whether or not G is included if the skeleton contains `y` but not `G`, especially in Buddhist and Japanese (CLDR likes to include it, ICU4J likes to remove it)
- whether to use `y` or `r(U)` in the Chinese calendar

That seems to cover the vast majority of differences. Then there are some that are more material, usually in smaller locales and non-gregorian calendars. Examples:

| Locale | Calendar | Skeleton | CLDR Result | ICU4J Result |
|---|---|---|---|---|
| aa |	gregorian |	yMd |	y-MM-dd |	d/M/y |
| fr |	chinese |	yMMMEd |	EEE d MMM U |	r(U) MMM d, EEE |
| hsb |	gregorian |	j |	H 'hodź'. |	H |
| ja |	gregorian |	yMEd |	y/M/d(EEE) |	y年M月d日(EEE) |

This is just a sampling of some of the other types of differences I see. I haven't investigated in detail. I suspect some of these differences might be attributable to whether or not we include the "stock patterns" in the DTPG.

srl295 (dismissed):
Marking approve to unblock.  (seeing that it doesn't pass yet)
I think most of my concerns were addressed.  I still don't like the extra metadata here that ought to be centralized,

sffc:
> Marking approve to unblock. (seeing that it doesn't pass yet) I think most of my concerns were addressed. I still don't like the extra metadata here that ought to be centralized,

Fixed in https://github.com/unicode-org/cldr/pull/5394/commits/75ebdd41e3aab1e3b17fc179983f6c1238c79ffc ?

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
```diff
@@ -0,0 +1,707 @@
...
+     * Creates an ICU4J DateTimePatternGenerator populated with the exact same data 
+     * loaded by this CldrDateTimePatternGenerator. This is useful for algorithmic comparisons.
+     */
+    public DateTimePatternGenerator getIcu4jGenerator() {
+        DateTimePatternGenerator icuGen = DateTimePatternGenerator.getEmptyInstance();
+        icuGen.setDefaultHourFormatChar(defaultHourFormatChar);
+        icuGen.setDateTimeFormat(DateFormat.FULL, dateTimeFormatFull);
+        icuGen.setDateTimeFormat(DateFormat.LONG, dateTimeFormatLong);
+        icuGen.setDateTimeFormat(DateFormat.MEDIUM, dateTimeFormatMedium);
+        icuGen.setDateTimeFormat(DateFormat.SHORT, dateTimeFormatShort);
```
macchiati, line 1062: To isolate the differences, I suggest making the following change:

```suggestion
    public DateTimePatternGenerator getIcu4jGenerator(boolean addStock) {
        DateTimePatternGenerator icuGen = DateTimePatternGenerator.getEmptyInstance();
        icuGen.setDefaultHourFormatChar(defaultHourFormatChar);
        if (addStock) {
          icuGen.setDateTimeFormat(DateFormat.FULL, dateTimeFormatFull);
          icuGen.setDateTimeFormat(DateFormat.LONG, dateTimeFormatLong);
          icuGen.setDateTimeFormat(DateFormat.MEDIUM, dateTimeFormatMedium);
          icuGen.setDateTimeFormat(DateFormat.SHORT, dateTimeFormatShort);
        }
```

And run the comparisons between the two _without_ adding the stock values (eg addStock==false).

That way we can check that the algorithm is right or wrong.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrDateTimePatternGeneratorCompare.java
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * Compares the behavior of CldrDateTimePatternGenerator against ICU4J's DateTimePatternGenerator.
+ */
+@CLDRTool(
+    alias = "compare-cldr-dtpg",
+    description = "Compares CldrDateTimePatternGenerator against ICU4J DateTimePatternGenerator",
+    url = CLDRURLS.TOOLSURL)
+public class CldrDateTimePatternGeneratorCompare {
+    private static final String[] SKELETONS = {
```
macchiati, line 29: We need to not only test these patterns, but also test patterns where the user is choosing common variants. For example, as well as Hms, add HHms. That will check whether the matching & length adjustments are the same.

Note: we had a useful discussion in the design meeting today about why the stock formats don't always match the skeletons. Our (very rough & preliminary) conclusion is that we may have to enhance the skeleton format.

macchiati (dismissed):

macchiati:
Thanks for creating the differences spreadsheet, https://docs.google.com/spreadsheets/d/1IAcbd1ylKfUePZSbBt2cwhxDjS9eZUgHeBJ6bK-0zl0/edit?gid=1157632567#gid=1157632567

I suggest first changing the code to compare apples to apples (see my comments), and adding another sheet to that spreadsheet with the new differences; I think that will resolve many of the differences.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
sffc: I tried running with this but it didn't change anything. Note that this is only setting the four glue patterns, which is not where most of the differences are.

sffc: There's a flag in the constructor on whether to include the stock patterns. I turned it off and the number of exact matches increased slightly. I added the results to the spreadsheet.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrDateTimePatternGeneratorCompare.java
sffc: Good point. I don't think we need this until we fully understand the current diffs, which "should be" easier.

srl295:
just curious is Gemini able to follow the commit message guidelines? so that this won't need to be squashed until merge.

sffc:
> just curious is Gemini able to follow the commit message guidelines? so that this won't need to be squashed until merge.

hm? I'm assuming we will squash at the end.

I can give it guidelines for commit messages. I've already asked it to list itself as a co-author, which it started doing for some of the more recent commits.

srl295:
> > just curious is Gemini able to follow the commit message guidelines? so that this won't need to be squashed until merge.
> 
> hm? I'm assuming we will squash at the end.

As you know the branch doesn't need squash if all of the commits follow the guidelines.

Also, we get the false positive of the failing test case ❌ due to the commit message.

> I can give it guidelines for commit messages. I've already asked it to list itself as a co-author, which it started doing for some of the more recent commits.

I saw that, thanks. ✨ anyway just a question for the future.

sffc:
I got it to 98.8% alignment. I think this is getting good enough to land.

I asked Gemini to investigate root causes of the remaining differences. Here is what Gemini has to say. WARNING: The following output is LLM-generated and I have not checked it for hallucinations.

---

### Root Cause: "Era Removed" Differences

The root cause of the **"Era removed"** category of differences is a discrepancy in
how the two implementations handle **unrequested fields during pattern expansion**.

When a requested skeleton (like `yMMMM`) is not present in a locale's
`availableFormats`, the `CldrDateTimePatternGenerator` attempts to resolve it by
finding the "closest" matching skeleton (such as `yyyyMMMM`). For non-Gregorian
calendars, these expanded skeletons typically map to patterns that contain an
**Era** field (e.g., `G y MMMM`).

In `CldrDateTimePatternGenerator.java`, the `expandPattern` method iterates through
the fields of the available pattern and tries to map them to the requested fields.
When it encounters a field in the available pattern that is **not present** in the
requested skeleton (like `G` for the Era), the current logic implicitly preserves it
and appends it to the result:

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
(~line 638)

```
              if (reqF != null) {
                  // ... maps and replaces requested field ...
              } else {
                  // If the field wasn't in the requested skeleton, just append it
                  res.append(patField);
              }
```

By contrast, **ICU4J's implementation**
(`DateTimePatternGenerator.adjustFieldTypes`) contains sophisticated logic that
explicitly identifies format items that were not requested in the target skeleton.
Instead of preserving them, ICU4J actively **strips unrequested fields** (along with
adjoining literal characters, such as spaces or commas). 

As a result:
- **CLDR's Generator:** Resolves `yMMMM` against a pattern like `G y MMMM` and
incorrectly keeps the `G`, producing `G y MMMM`.
- **ICU4J's Generator:** Drops the `G` and its surrounding spaces, correctly
producing `MMMM y` (or similar variants).

### Root Cause: "Field Set Mismatch" Differences

The **"Field set mismatch"** category (226 occurrences) is primarily driven by two
missing pieces of context-sensitive substitution logic in
`CldrDateTimePatternGenerator` compared to ICU4J's implementation.

#### 1. Format vs. Standalone Context (Months & Days)
The vast majority of these mismatches occur because ICU4J intelligently converts
format characters (like `M` for month, or `E` for day of week) into their
standalone equivalents (like `L` and `c`) based on the presence of other fields in
the pattern.
- **Example:** For the requested skeleton `yMMM` in the `ab` locale, the
`CldrDateTimePatternGenerator` finds the closest matching pattern and produces `MMM
y 'ш'. G` (using the format month `M`). 
- **ICU4J's Behavior:** Because the pattern does *not* contain a day field (`d`),
ICU4J's `DateTimePatternGenerator` recognizes that the month is standing alone in
the context of a date and actively converts the format month `M` to the standalone
month `L`, producing `LLL y 'ш'. G`. CLDR currently lacks this contextual
substitution step.

#### 2. Flexible Day Periods (`B`) vs. AM/PM (`a`)
The second cluster of mismatches relates to how 12-hour clock skeletons (like `j`
and `h`) are resolved with day periods. 
- **Example:** For the requested skeleton `j` (or `h`) in `zh_Hant`, CLDR's
`zh_Hant.xml` explicitly defines an available format `<dateFormatItem
id="h">Bh時</dateFormatItem>`. The `CldrDateTimePatternGenerator` correctly looks
this up and uses the flexible day period `B` (e.g., "morning", "afternoon"),
producing `Bh時`.
- **ICU4J's Behavior:** ICU4J evaluates these fields differently during fallback or
skeleton resolution and often opts for the standard AM/PM marker `a`, generating
`ah時` instead of the `Bh時` explicitly mapped in the CLDR data. This represents a
discrepancy in how flexible day periods are prioritized or normalized during the
pattern generation pipeline between the two libraries.

### Root Cause: "Field Length Mismatch" Differences

The **"Field length mismatch"** category primarily stems from differences in how
unrequested fields are appended (eras) and how conceptually related fields (like
days of the week) are scaled during pattern expansion.

#### 1. Implicit Era Insertion (`G` vs. `GGGGG`)
This is the most common length mismatch (63 occurrences).
- **Scenario:** A skeleton that *lacks* an Era is requested (e.g., `yM`) for a
non-Gregorian calendar (like `buddhist` or `japanese`).
- **ICU4J's Behavior:** When ICU4J determines that a calendar requires an era to be
unambiguous but the user didn't request one, it relies on its internal
`appendItemFormat` fallback to inject an era. By default, this implicit fallback
uses the **narrow** era format (`GGGGG`), resulting in patterns like `M/y GGGGG`.
- **CLDR's Behavior:** `CldrDateTimePatternGenerator` resolves the `yM` skeleton by
matching it against a larger available format (like `yyyyMM`, which maps to `M-y
G`). Because the matched pattern explicitly included an **abbreviated** era (`G`),
CLDR preserves it (as seen in the "Era removed" bug) and produces `M-y G`.

#### 2. Scaling Cross-Character Fields (`ccc` vs. `cccc`, `EEEE` vs. `EEE`)
This occurs when the matched pattern uses a different character for a field than
the requested skeleton (e.g., using `c` for standalone day-of-week when `E` was
requested).
- **Scenario:** The skeleton `MMMMEEEEd` (requesting a wide day-of-week, `EEEE`) is
matched against the available format `MMMEd`. In the `ab` locale, `MMMEd` maps to
the pattern `ccc, d MMM`. 
- **CLDR's Behavior:** The `expandPattern` method in
`CldrDateTimePatternGenerator.java` strictly requires the length of the field in
the pattern string to match the length of the field in the available skeleton
before it attempts to scale it (`patField.length() == availF.length()`). Because
the pattern uses `ccc` (length 3) to represent `E` (length 1), this condition fails
(`3 != 1`). CLDR bails out of scaling and incorrectly preserves the original `ccc`.
- **ICU4J's Behavior:** ICU4J recognizes the conceptual equivalence between `E` and
`c` as day-of-week fields. When it sees that the user requested a width of 4
(`EEEE`), it correctly forces the `ccc` field to scale up to `cccc`.

#### 3. Preservation of Locale Preferences (`M` vs. `MM`)
A smaller subset of mismatches occurs when a locale explicitly prefers a longer
numeric format (like `MM` for Polish `yM` skeletons).
- **Scenario:** In the Polish `buddhist` calendar, the user requests `yM` (length 1
for month). 
- **ICU4J's Behavior:** ICU4J relies on cross-calendar fallback, pulling the `yM`
pattern from the `pl` Gregorian calendar, which explicitly defines `<dateFormatItem
id="yM">MM.y</dateFormatItem>`. It respects this locale-specific preference for a
2-digit month.
- **CLDR's Behavior:** Lacking robust cross-calendar fallback in this test setup,
CLDR finds `yyyyMM` in the Buddhist calendar (which maps to `MM.y G`). Because the
user requested `yM` (length 1), it strictly scales the `MM` down to `M`, generating
`M.y G` and erasing the locale's preference for `MM`.

### Root Cause: "Separator/Literal Mismatch" Differences

The **"Separator/Literal mismatch"** category (where the two implementations choose
patterns with different spaces, commas, or slashes) is fundamentally caused by two
distinct shortcomings in `CldrDateTimePatternGenerator`'s architecture compared to
ICU4J's `DateTimePatternGenerator`.

#### 1. Simplistic Skeleton Distance Metric
The most significant cause of mismatched literals (such as missing commas in `el` or
`fa` locales) is a flaw in `CldrDateTimePatternGenerator.getSingleFieldDistance()`. 

When comparing numeric fields (like `y` vs `yyyy` or `d` vs `dd`), CLDR's
implementation currently assigns a constant penalty:

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java

```
          if (isNumeric) {
              diff += 2;
              return diff;
          }
```

Because the distance penalty does not scale with the actual difference in field
length, `yMMMMd` calculates the exact same distance (8) to the available skeleton
`yMMMd` as it does to `yyyyMMMd`. 
- **Scenario:** In the Greek (`el`) Japanese calendar, the translator explicitly
defined `yMMMd` as `d MMM, y G` (with a comma). The skeleton `yyyyMMMd` implicitly
falls back to a generic `d MMM y G` (no comma).
- **CLDR's Behavior:** Because the calculated distance to both skeletons is
identical, CLDR arbitrarily selects `yyyyMMMd` (due to iteration order) and
produces `d MMMM y G` (lacking the locale-specific comma). 
- **ICU4J's Behavior:** ICU4J's distance metric correctly penalizes numeric length
differences. It identifies `yMMMd` as a strictly closer match to `yMMMMd` than
`yyyyMMMd`. Therefore, it uses the explicitly localized `d MMM, y G` pattern and
scales the month, correctly producing `d MMMM, y G`.

#### 2. Lack of Cross-Calendar Fallback
The second cause of separator mismatches (e.g., in `bn` and `ccp` where CLDR
produces `M-y G` but ICU4J produces `M/y G`) is the absence of cross-calendar
fallback.
- **Scenario:** In the Bengali (`bn`) Buddhist calendar, the requested skeleton is
`yM`. 
- **ICU4J's Behavior:** ICU4J looks for `yM` in the Buddhist calendar, doesn't find
it, and correctly falls back to the Gregorian calendar for that locale. In `bn`
Gregorian, `yM` is explicitly defined as `M/y`. ICU4J takes `M/y` and appends the
unrequested (but necessary) era, producing `M/y G`.
- **CLDR's Behavior:** `CldrDateTimePatternGenerator` does not fall back to the
Gregorian calendar. Forced to stay within the Buddhist available formats, it
matches against `yyyyMM`. In `bn.xml`, `yyyyMM` is explicitly defined as `MM-y G`
(using a hyphen). CLDR scales the `MM` down to `M`, producing `M-y G`.

### Root Cause: "Field Order Mismatch" Differences

The **"Field order mismatch"** category (24 occurrences) is almost exclusively
driven by discrepancies in where the **Era (`G`) field** is placed when a requested
skeleton requires an era but doesn't explicitly define one (like `yM` in
non-Gregorian calendars).

This is another side effect of how the two implementations handle fallback and
missing fields.

- **Scenario:** The user requests `yM` for a non-Gregorian calendar (e.g., `gu`
Buddhist or `hi` Buddhist). 
- **ICU4J's Behavior:** Lacking an exact `yM` match in the Buddhist calendar, ICU4J
falls back to the Gregorian calendar and finds `yM` mapped to `M/y`. Because it's
formatting a non-Gregorian date, it algorithmically forces an Era (`G`) into the
pattern to prevent ambiguity. ICU4J uses internal heuristics to inject the Era
either via `appendItemFormat` (e.g., `{1} {0}`) or by placing it adjacent to the
Year field, producing synthetic patterns like `M/G y` or `G d/M/y`. 
- **CLDR's Behavior:** `CldrDateTimePatternGenerator` stays within the
non-Gregorian calendar and matches a broader skeleton that *does* explicitly define
the Era's position according to locale data (e.g., `yyyyMM` mapped to `M-y G`). By
simply scaling down the matched pattern, CLDR preserves the explicitly localized
position of the Era at the end of the string (`M-y G`), rather than algorithmically
injecting it into a Gregorian fallback pattern.

In summary, ICU4J's algorithmic injection of missing eras into Gregorian fallback
patterns frequently contradicts the explicit era placement defined in the broader,
native non-Gregorian patterns that CLDR's generator successfully matches.

sffc:
This is an interesting finding, which comes up multiple times:

> * **Scenario:** In the Bengali (`bn`) Buddhist calendar, the requested skeleton is
>   `yM`.
> * **ICU4J's Behavior:** ICU4J looks for `yM` in the Buddhist calendar, doesn't find
>   it, and correctly falls back to the Gregorian calendar for that locale. In `bn`
>   Gregorian, `yM` is explicitly defined as `M/y`. ICU4J takes `M/y` and appends the
>   unrequested (but necessary) era, producing `M/y G`.
> * **CLDR's Behavior:** `CldrDateTimePatternGenerator` does not fall back to the
>   Gregorian calendar. Forced to stay within the Buddhist available formats, it
>   matches against `yyyyMM`. In `bn.xml`, `yyyyMM` is explicitly defined as `MM-y G`
>   (using a hyphen). CLDR scales the `MM` down to `M`, producing `M-y G`.

Looking myself, here's what I see:

bn.xml:

```
			<calendar type="buddhist">
				<eras>
					<eraNames>
						<era type="0" draft="contributed">↑↑↑</era>
					</eraNames>
					<eraAbbr>
						<era type="0" draft="contributed">↑↑↑</era>
					</eraAbbr>
					<eraNarrow>
						<era type="0" draft="contributed">↑↑↑</era>
					</eraNarrow>
				</eras>
			</calendar>
```

There are no availableFormats, so we check root.xml:

```
			<calendar type="buddhist">
				...
				<dateFormats>
					<alias source="locale" path="../../calendar[@type='generic']/dateFormats"/>
				</dateFormats>
				<timeFormats>
					<alias source="locale" path="../../calendar[@type='gregorian']/timeFormats"/>
				</timeFormats>
				<dateTimeFormats>
					<alias source="locale" path="../../calendar[@type='generic']/dateTimeFormats"/>
				</dateTimeFormats>
			</calendar>
```

Now we go back to bn.xml and check generic:

```
			<calendar type="generic">
				...
				<dateTimeFormats>
					...
					<availableFormats>
						...
						<dateFormatItem id="y">y G</dateFormatItem>
						<dateFormatItem id="yyyy">y G</dateFormatItem>
						<dateFormatItem id="yyyyM">M/y G</dateFormatItem>
						<dateFormatItem id="yyyyMd">d/M/y G</dateFormatItem>
						<dateFormatItem id="yyyyMEd">E, d/M/y G</dateFormatItem>
						<dateFormatItem id="yyyyMM">MM-y G</dateFormatItem>
						<dateFormatItem id="yyyyMMM">MMM y G</dateFormatItem>
						<dateFormatItem id="yyyyMMMd">d MMM, y G</dateFormatItem>
						<dateFormatItem id="yyyyMMMEd">E, d MMM, y G</dateFormatItem>
						<dateFormatItem id="yyyyMMMM">MMMM y G</dateFormatItem>
						<dateFormatItem id="yyyyQQQ">QQQ y G</dateFormatItem>
						<dateFormatItem id="yyyyQQQQ">QQQQ y G</dateFormatItem>
					</availableFormats>
					...
				</dateTimeFormats>
			</calendar>
```

And from there, both `yyyyM` and `yyyyMM` are very close matches, but `yyyyM` is the best, based on the skeleton `yM`. So it appears that the correct answer is ICU4J's `M/y G`. Worth fixing.

macchiati:
It will take time to review all of these cases. For the first I noticed

> CLDR's Generator: Resolves yMMMM against a pattern like G y MMMM and
incorrectly keeps the G, producing G y MMMM.

That is *correct* and desired behavior. The skeleton to pattern mapping is allowed to change fields: adding, removing, or changing widths or types. Once the best match for the skeleton is found, the adjustment phase has to fix the pattern as necessary. The G must be retained.

sffc:
Down to 247 differences and up to 74249 agreements.

macchiati:
Can you put a link to the updated spreadsheet here? I'll review on the bus.

sffc:
Same spreadsheet link.

https://docs.google.com/spreadsheets/d/1IAcbd1ylKfUePZSbBt2cwhxDjS9eZUgHeBJ6bK-0zl0/edit?gid=655580735#gid=655580735

I'm also looking through it. I see two main classes of errors so far:

1. Skeleton yMMMM, CLDR correctly selects availableFormat `yyyyMMMM`, but ICU4J seems to select `yMMM` and expands the 3M to 4M.
2. In locale `eu`, when requesting `yMEd`, there is an availableFormat `yMEd` that maps to pattern "y/MM/dd, EEEE". Since the skeleton is an exact match, CLDR keeps the weekday as 4E, but ICU4J seems to shorten it to 3E to match the requested skeleton.

sffc:
Another case:

3. Locale ko, calendar chinese, skeleton yMMMMEEEEd. The XML is

```
						<dateFormatItem id="y">r년(U년)</dateFormatItem>
						<dateFormatItem id="yMd">r. M. d.</dateFormatItem>
						<dateFormatItem id="yyyy">r년(U년)</dateFormatItem>
						<dateFormatItem id="yyyyM">r. M.</dateFormatItem>
						<dateFormatItem id="yyyyMd">r. M. d.</dateFormatItem>
						<dateFormatItem id="yyyyMEd">r. M. d. (E)</dateFormatItem>
						<dateFormatItem id="yyyyMEEEEd">r. M. d. EEEE</dateFormatItem>
						<dateFormatItem id="yyyyMMM">r년(U년) MMM</dateFormatItem>
						<dateFormatItem id="yyyyMMMd">r년 MMM d일</dateFormatItem>
						<dateFormatItem id="yyyyMMMEd">r년(U년) MMM d일 (E)</dateFormatItem>
						<dateFormatItem id="yyyyMMMEEEEd">r년(U년) MMM d일 EEEE</dateFormatItem>
						<dateFormatItem id="yyyyMMMM">r년(U년) MMMM</dateFormatItem>
						<dateFormatItem id="yyyyMMMMd">r년(U년) MMMM d일</dateFormatItem>
						<dateFormatItem id="yyyyMMMMEd">r년(U년) MMMM d일(E)</dateFormatItem>
						<dateFormatItem id="yyyyQQQ">r년(U년) QQQ</dateFormatItem>
						<dateFormatItem id="yyyyQQQQ">r년(U년) QQQQ</dateFormatItem>
```

CLDR thinks the closest skeleton match is yyyyMMMEEEEd, so it selects that pattern and changes 3M to 4M.

ICU4J seems to think the closest skeleton match is yyyyMMMMEd, so it selects that pattern and thanges 3E to 4E.

In this particular case I think CLDR's behavior makes more sense because 3E is shown in `()` while 4E is not, and meanwhile 3M and 4M doesn't seem to impact the formatting, but I'm not sure which behavior is more correct according to the algorithm.

macchiati:
I added a filter for TC locales, so we could focus on those. Gets it down
to 122

macchiati:
for #3, EEEE vs E is just as  distance of wide vs abbr. Same distance as MMMM vs MMM

And the month length should be more important than the dow length, all other things being equal. So I think ICU4j is right

macchiati:
> 1. Skeleton yMMMM, CLDR correctly selects availableFormat `yyyyMMMM`, but ICU4J seems to select `yMMM` and expands the 3M to 4M.

Agreed; I think CLDR is correct. Wide vs abb is more important than numeric width differences.

> 2. In locale `eu`, when requesting `yMEd`, there is an availableFormat `yMEd` that maps to pattern "y/MM/dd, EEEE". Since the skeleton is an exact match, CLDR keeps the weekday as 4E, but ICU4J seems to shorten it to 3E to match the requested skeleton.

I don't understand "shorten it to 3E to match the requested skeleton", but the requested skeleton just has E.

sffc:
One more difference category:

4. zh_Hant, gregorian, j. CLDR finds "Bh時"; ICU4J finds "ah時". @macchiati agrees that CLDR is correct here because this is the data for zh_Hant.

Problem case 1 comes up the most often. For example, locale `om`, calendar buddhist, skeleton yMMM, CLDR matches yyyyMMM, ICU4J appears to match yMMMM. @macchiati agrees that CLDR is correct.

sffc:
Updated the spreadsheet. Adding the month precedence fixed a bunch more differences. We are down to 93 diffs, which are:

- The Basque ones, case 2, where we said CLDR is right
    - Also includes some diffs in syr, ie
- The Traditional Chinese ones, case 4, where we said CLDR is right

It turned up one new class of diffs, which occurs just once:

5. Locale fi, calendar gregorian, skeleton yMMMMEEEEd. CLDR: "EEEE d. MMMM y". ICU4J: "cccc d. MMMM y". The relevant data is:

```
						<dateFormatItem id="yMMMMccccd">cccc d. MMMM y</dateFormatItem>
						<dateFormatItem id="yMMMMd">d. MMMM y</dateFormatItem>
						<dateFormatItem id="yMMMMEd">E d. MMMM y</dateFormatItem>
```

CLDR is selecting yMMMMEd and then widening the weekday.

ICU4J appears to be selecting yMMMMccccd, which is a closer match because of lengths, even though the weekday symbol in the available skeleton is not the same as the one in the requested skeleton.

@macchiati thoughts on this case?

macchiati:
The cccc should never be in the skeleton, just in the pattern. Try
temporarily patching the skeleton to use EEEE instead.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrDateTimePatternGeneratorCompare.java
sffc: I added more.

sffc:
Latest difference classes:

1. Skeleton "UUUUU", CLDR says "yyyyy", ICU4J says "y". (The pattern is just "y".)
    - Not sure who is right.
2. Skeleton "k", CLDR says "h a", ICU4J says "HH".
    - Neither one seems right; the result should be "kk" I think. ICU4J seems closer.
3. Skeleton "yQ", CLDR says "y ('Quarter': Q)", ICU4J says "y QQQ". (There is no yQ or yQQ skeleton here.)
    - Not sure who is right.
4. Skeleton "jjj", CLDR says "h a", ICU4J says "h aaaa"
    - ICU4J looks right. CLDR needs to learn about this special case
5. Skeleton "hhB", CLDR says "hh B", ICU4J says "h B"
    - CLDR seems right
6. Skeleton "yyyyy", CLDR says "G y", ICU4J says "G yyyyy"
    - I might drop this skeleton. It seems like a bad skeleton.

The updated spreadsheet is available in a new tab, https://docs.google.com/spreadsheets/d/1IAcbd1ylKfUePZSbBt2cwhxDjS9eZUgHeBJ6bK-0zl0/edit?gid=1832895706#gid=1832895706

AEApple:
> Latest difference classes:
> 
> 1. Skeleton "UUUUU", CLDR says "yyyyy", ICU4J says "y". (The pattern is just "y".)
>    
...
> 6. Skeleton "yyyyy", CLDR says "G y", ICU4J says "G yyyyy"
>    
>    * I might drop this skeleton. It seems like a bad skeleton.
> 

What is the narrow form of the year even look like in practice? Does anyone use it?

macchiati:
We'd have to see what the patterns in the source are that involved?y and G to know which one is correct.

As to the use of narrow forms: 

I don't think either cldr.or ICU have made any effort to test the narrow forms in calendars, except perhaps the months and eras: and we don't present them to vetters.

We should take a look at some resulting formatted dates, but that can be done later.

sffc:
I consider "yyyyy" to be an AI hallucination but I listed it because it is in the spreadsheet.

I am much more interest on opinions for cases 1-5.

macchiati:
I want to confirm that none of them the stock (<timeFormat> or <dateFormat>) patterns; we need to drop those from comparison.

Note: I reinserted a column with the TC Locale name so I could filter out the non-TC locales, and Basic locales. We just don't want to worry about those.

> Latest difference classes:
> 
> 1. Skeleton "UUUUU", CLDR says "yyyyy", ICU4J says "y". (The pattern is just "y".)
>    
>    * Not sure who is right.

I picked one of the locales to look at, 'af'. Since the only y (or y variant) in af is 
<dateFormatItem id="y" draft="unconfirmed">↑↑↑</dateFormatItem>
I followed to root, and got  
<dateFormatItem id="y">y</dateFormatItem>

CLDR is clearly right.

> 2. Skeleton "k", CLDR says "h a", ICU4J says "HH".
>    
>    * Neither one seems right; the result should be "kk" I think. ICU4J seems closer.

"k" should not be in a desired skeleton at all; the input should be either h or H. That being said, if you are normalizing the input, 'k' should map to H, since both are 24 hour.

> 3. Skeleton "yQ", CLDR says "y ('Quarter': Q)", ICU4J says "y QQQ". (There is no yQ or yQQ skeleton here.)

This is the closest match, and has all the fields. So it wins; ICU4J is right.
<dateFormatItem id="yQQQ">QQQ y</dateFormatItem>

>    
>    * Not sure who is right.
> 4. Skeleton "jjj", CLDR says "h a", ICU4J says "h aaaa"
>    
>    * ICU4J looks right. CLDR needs to learn about this special case

Yup, quoting scripture: jjj -> Numeric hour (minimum digits), wide dayPeriod if used

> 5. Skeleton "hhB", CLDR says "hh B", ICU4J says "h B"
>    
>    * CLDR seems right

Yes, unless this is for a locale that forces no leading zero. That would happen if there was a pattern:

hhB -> h B

> 6. Skeleton "yyyyy", CLDR says "G y", ICU4J says "G yyyyy"
>    
>    * I might drop this skeleton. It seems like a bad skeleton.

Agreed. Should drop this from the comparison. Along with Gyyyy

> 
> The updated spreadsheet is available in a new tab, https://docs.google.com/spreadsheets/d/1IAcbd1ylKfUePZSbBt2cwhxDjS9eZUgHeBJ6bK-0zl0/edit?gid=1832895706#gid=1832895706

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
macchiati, line 304: We need to stop using stock. The easiest way to do this is to just ignore the useStock flag. I'd suggest a constant 

```suggestion
    static boolean IGNORE_STOCK = true; // this shhou 

/**
    * Set a CLDRFile and calendar. Must be done before calling addTable.
    *
    * <p>Note: currently, this method will skip `alt="ascii"` elements of `dateFormatItem`. This
    * may be configurable in the future. See CLDR-18580
    *
    * @param file
    * @param calendarID
    * @return
    */
public DateTimeFormats(CLDRFile file, String calendarID, boolean useStock) {
    if (IGNORE_STOCK) {
        useStock = false;
    }

```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CldrDateTimePatternGeneratorCompare.java
```diff
@@ -0,0 +1,388 @@
...
+            int count = 0;
+            List<String> sortedLocales = new ArrayList<>(locales);
+            Collections.sort(sortedLocales);
+
+            for (String localeID : sortedLocales) {
+                if (localeID.equals("root")) continue;
+                if (!localeID.matches(filter)) continue;
+
+                System.out.println("Processing locale: " + localeID);
+                CLDRFile cldrFile = factory.make(localeID, true);
```
macchiati, line 220: Steven, I don't know what your comment about inheritance is directed towards. This is using a resolved CLDRFile.

sffc:
I'm trying to fix this one: jjj -> h aaaa

There are no "ha" skeletons. Only "h" and "hB". So if I try to transform "jjj" to "haaaa" for lookup, it matches "hB".

Can you walk through step by step how "jjj" should resolve against data?

sffc:
Getting to the bottom... here's another one

English, japanese, skeleton "yMdEEEEE" (those are 5 E's). Data is

```xml
<dateFormatItem id="y">y G</dateFormatItem>
<dateFormatItem id="yyyy">y G</dateFormatItem>
<dateFormatItem id="yyyyM">M/y GGGGG</dateFormatItem>
<dateFormatItem id="yyyyMd">M/d/y GGGGG</dateFormatItem>
<dateFormatItem id="yyyyMEd">E, M/d/y GGGGG</dateFormatItem>
<dateFormatItem id="yyyyMEEEEd">EEEE, MMMM d, y G</dateFormatItem>
<dateFormatItem id="yyyyMM">MM y GGGGG</dateFormatItem>
<dateFormatItem id="yyyyMMM">MMM y G</dateFormatItem>
<dateFormatItem id="yyyyMMMd">MMM d, y G</dateFormatItem>
<dateFormatItem id="yyyyMMMEd">E, MMM d, y G</dateFormatItem>
<dateFormatItem id="yyyyMMMEEEEd">EEEE, MMMM d, y G</dateFormatItem>
<dateFormatItem id="yyyyMMMM">MMMM y G</dateFormatItem>
```

ICU4J matches "yyyyMEd" closest. CLDR matches "yyyyMEEEEd" closest.

Technically speaking CLDR is correct, because 5E is closer to 4E than 1E. However, since 5E is narrow, it is more like 1E (abbreviated) than 4E (long).

macchiati:
This is one of those areas where we need to augment the spec to have different levels of distance.

sffc:
> This is one of those areas where we need to augment the spec to have different levels of distance.

I interpret this to mean that you think the "yMdEEEEE" behavior in this PR is acceptable.

Could you respond to the question above it about "jjj"?

macchiati:
We only have h in the ids; what happens is that the 'a' is added in the patterns in each case (and I believe we verify that in the tooling).

So all of the j's will map to an id of h+, B+h+, or H+, with widths depending on the number of j's. Does that make sense?

sffc:
To resolve jjj, I manually implemented this line of the spec:

> Skeletons for 12-hour-cycle time formats (using h or K) may include a day period field using one of a, b, or B. If they do not, the skeleton will be treated as implicitly containing a.

macchiati:
Yes good.

> To resolve jjj, I manually implemented this line of the spec:
> 
> > Skeletons for 12-hour-cycle time formats (using h or K) may include a day period field using one of a, b, or B. If they do not, the skeleton will be treated as implicitly containing a.

Needs to have a coda. If when matching desired skeletons to available ids, most if not all ids will omit the 'a' but have it in the pattern. Thus the absence of a 'a' in the id has a minimal effect on the distance matching; less that length or type  differences in any other field, and much less than a missing field.

sffc:
Good news! On my latest run, all diffs between CLDR and ICU4J appear to be accounted for. I put explanations in the spreadsheet:

https://docs.google.com/spreadsheets/d/1IAcbd1ylKfUePZSbBt2cwhxDjS9eZUgHeBJ6bK-0zl0/edit?gid=604117187#gid=604117187

If you agree, then we can proceed to merge this, I think?

macchiati:
Nearly there. 

I wanted to look at the U behavior, and added a new tab in your spreadsheet. It appears that it is typically just in the pattern, not in the skeleton (id). However, there are some instances where it is in the ID also (but always then in the value). I'll add some comments in the sheet.

sffc:
Discussion with @macchiati:

- For now, don't replace `y` by `U` in patterns, until we investigate further with @pedberg-icu 
- We tweaked some distance metrics in order to make MMMMM be closer to MMM than MMMM

macchiati:
You need to amend the commit message with Github Desktop so that the failure disappears. I'll stamp it afterwards.

See https://docs.github.com/en/desktop/managing-commits/amending-a-commit-in-github-desktop#amending-a-commit

sffc:
Ping @macchiati, please let me know what else there is to do on this PR. It has been open for 2 months and has gone through a great deal of testing and iteration. Thank you

sffc:
@macchiati I squashed the branch.

sffc:
The original branch with 65 commits is backed up here: https://github.com/sffc/cldr/tree/archive/cldr-dtpg

macchiati (approved):

macchiati:
Taking the liberty of merging, for time.

PR #5407, opened 2026-02-26
CLDR-18832 Add singular cardinal rules and ordinal rules for Slovenian

grhoten:
CLDR-18832

This pull request includes most of the changes from the proposed changes from this ticket. It excludes the dual and plural cardinal rules, since it was conceptually hard to verify, and the provided rules were too incomplete to use. Most of the ordinal rules were included, but if there were scenarios where the gender or grammatical number didn't affect the ordinal rule choice, I just collapsed them to one rule.

If the plural or dual cardinal rules are still desired, I recommend submitting a separate ticket, and making a contribution on top of these rules, which has roundtrippable simpler rules. As it stands, I could not validate or cross reference such information from what I could find online, and these rules are complicated enough.

The provided rules from the ticket did not roundtrip correctly. It had many gaps in the rules. These rules do roundtrip. In order to roundtrip, the latest rule syntax from ICU must be used, like the kind that are already in Polish and Lithuanian. I replaced all of the contributed rules that were 100 and larger with this syntax.

I used the soft hyphen for the ordinals. Some of the other languages use a soft hyphen to allow better line wrapping of long ordinals.

I did not redundantly transform these to the old RBNF XML syntax, and these rules use the latest rule format, which will should be going into CLDR 49. This should not be merged into CLDR 48 or earlier.

ALLOW_MANY_COMMITS=true


srl295 (dismissed):

conradarcturus (dismissed):
I don't speak Slovenian but I trust your expertise.

grhoten:
The diff between force merges is due to the end of sl.xml needing a little cleanup after merging from upstream. There are only whitespace differences at the end of the file.

grhoten:
@srl295 @conradarcturus can you please approve again. Merging upstream caused a whitespace difference to reset the approval.

srl295 (approved):

PR #5411, opened 2026-02-27
CLDR-19181 Correct RBNF rules for letter based numbering systems: Armenian, Georgian

grhoten:
CLDR-19181

The changes to turn 0 into words was rejected, since the intent of these numeral systems is to use number digits and not words to be translated. If such a digit didn't exist in a given numbering system, then we default to the digit 0. The logic is consistent with fractional and negative numbers too, which should use the same digit 0.

ALLOW_MANY_COMMITS=true


common/rbnf/root.xml
```diff
@@ -292,22 +364,78 @@ x.x: =#,##0.00=;
...
+10000: ჵ[>>];
+20000: <<ჵ[>>];
+100000000: ჵჵ[>>];
+200000000: <<ჵჵ[>>];
+1000000000000: ჵჵჵ[>>];
+2000000000000: <<ჵჵჵ[>>];
+10000000000000000: ჵჵჵჵ[>>];
+20000000000000000: <<ჵჵჵჵ[>>];
+1000000000000000000: =#,##0=;
+%georgian-upper:
```
grhoten, line 390: This is a new rule. The lower cased one is left unchanged.

common/rbnf/root.xml
```diff
@@ -292,22 +364,78 @@ x.x: =#,##0.00=;
 200: ს[>>];
 300: ტ[>>];
 400: უ[>>];
-500: ჳ[>>];
```
grhoten, line 295: This is a typo. The numbers are shifted from here on out.

common/rbnf/root.xml
```diff
@@ -292,22 +364,78 @@ x.x: =#,##0.00=;
...
+1000: ჩ[>>];
+2000: ც[>>];
+3000: ძ[>>];
+4000: წ[>>];
+5000: ჭ[>>];
+6000: ხ[>>];
+7000: ჴ[>>];
+8000: ჯ[>>];
+9000: ჰ[>>];
+10000: ჵ[>>];
```
grhoten, line 381: This range falls into the theoretical range. Also this number is obsolete.

common/rbnf/root.xml
```diff
@@ -53,7 +53,43 @@ x.x: =#,##0.00=;
 7000: ւ[>>];
 8000: փ[>>];
 9000: ք[>>];
-10000: =#,##0=;
+10000: ա︦[>>];
```
grhoten, line 56: We're expanding the range based on the examples from https://en.wikipedia.org/wiki/Armenian_numerals.

conradarcturus (approved):
nice! Thanks for cleaning up the code.

common/rbnf/root.xml
```diff
@@ -1381,24 +821,6 @@ x.x: =0.0=;
 -x: −>>;
 0: =#,##0.#=.;
 ]]></rbnfRules>
-            <!-- The following redundant ruleset elements have been deprecated and will be removed in the next release. Please use the rbnfRules contents instead. -->
```
srl295: won't this and others conflict with #5405 ? 

Maybe that should be merged first?

grhoten: I agree

PR #5443, opened 2026-03-08
CLDR-19294 Improve Russian ordinal RBNF rules

grhoten:
CLDR-19294

The Russian ordinal numbers were contributed prior to the existence of the or-pipe operator syntax used in the optional brackets in the RBNF syntax. Those ordinal rules were truncated at 21000 because the ICU implementation could not replicate how those ordinal numbers actually worked. We should fix that, and extend the ordinal rules beyond 21000.

These rules were tested with the [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp)

ALLOW_MANY_COMMITS=true


common/rbnf/ru.xml
```diff
@@ -846,38 +785,32 @@ x.x: =0.#=;
...
+1000000000000: триллион[ >>|ное];
+2000000000000: [<%spellout-cardinal-masculine< $(cardinal,one{триллион}few{триллиона}other{триллионов})$ >>|<%%thousandsprefixconjoined<триллионное];
+1000000000000000: квадриллион[ >>|ное];
+2000000000000000: [<%spellout-cardinal-masculine< $(cardinal,one{квадриллион}few{квадриллиона}other{квадриллионов})$ >>|<%%thousandsprefixconjoined<квадриллионное];
+1000000000000000000: =0=-е;
 %spellout-ordinal-feminine:
 -x: минус >>;
 x.x: =0.#=;
-0: ноль;
+0: нулевая;
```
grhoten, line 813: Not an important fix, but at least the spelling is consistent with the other ordinal rules.

common/rbnf/ru.xml
```diff
@@ -1103,41 +1012,35 @@ x.x: =0.#=;
...
+2000000000000000: [<%spellout-cardinal-masculine< $(cardinal,one{квадриллион}few{квадриллиона}other{квадриллионов})$ >>|<%%thousandsprefixconjoined<квадриллионных];
+1000000000000000000: =0=-х;
 %spellout-ordinal-masculine-dative:
 -x: минус >>;
 x.x: =0.#=;
 0: нулевому;
 1: первому;
 2: второму;
-3: третьому;
+3: третьему;
```
grhoten, line 1043: Oops! That was a typo.

common/rbnf/ru.xml
```diff
@@ -705,73 +703,20 @@ x.x: << запятая >>;
...
-91: девяностоодно;
-92: девяносто[>>];
+20: двадцати[>>];
+30: тридцати[>>];
+40: сорока[>>];
+50: пятидесяти[>>];
+60: шестидесяти[>>];
+70: семидесяти[>>];
+80: восьмидесяти[>>];
+90: девяносто[>>];
```
srl295, line 717: is this one intended to have the old value for 92 instead of spellout-cardinal-feminine?

grhoten: The rule as previously written was a little convoluted. 90 is also девяносто in spellout-cardinal-feminine. So the new rule behavior is the same as the old rule, but it's simpler.

srl295 (approved):

PR #5444, opened 2026-03-09
CLDR-15169 Italian RBNF improvements

grhoten:
CLDR-15169

The following RBNF issues in Italian are being fixed:

- The numbering is the masculine form in isolation and in front of some forms of masculine nouns. The current masculine form remains the masculine apocopic form.
- The feminine cardinals after 1 should be the same as the masculine forms. That’s the primary issue that is being addressed.
- The ordinals in the 2000 and above range were incorrect, and it's now fixed. The right way to do it requires the latest RBNF syntax. This was likely an omission due to technical barriers that no longer exist.
- The plural forms switch to the singular forms when they should stay plural.

The number of public rules remain unchanged.

ALLOW_MANY_COMMITS=true


common/rbnf/it.xml
```diff
@@ -123,16 +115,6 @@ x.x: << virgola >>;
 0: a;
 1: ­un;
 2: =%%msco-with-a=;
-%%msc-with-o:
```
grhoten, line 126: Remove redundant ruleset that was a duplciate of %%msco-with-o

common/rbnf/it.xml
```diff
@@ -154,48 +136,15 @@ x.x: << virgola >>;
...
 3: o­tre;
-4: =%%msc-with-o=;
+4: =%%msco-with-o=;
 %spellout-cardinal-feminine:
 -x: meno >>;
 x.x: << virgola >>;
 0: zero;
 1: una;
 2: =%spellout-numbering=;
-20: vent>%%fem-with-i>;
```
grhoten, line 166: This removal was the requested change in this ticket.

common/rbnf/it.xml
```diff
@@ -230,8 +179,7 @@ x.x: =#,##0.#=;
 100: cent>%%ordinal-esimo-with-o>;
 200: <%spellout-cardinal-masculine<­cent>%%ordinal-esimo-with-o>;
 1000: mille­>%%ordinal-esimo>;
-2000: <%spellout-cardinal-masculine<­mille­>%%ordinal-esimo>;
-2001: <%spellout-cardinal-masculine<­mila­>%%ordinal-esimo>;
+2000: <%spellout-cardinal-masculine<[­mila­>%%ordinal-esimo>|­millesimo];
```
grhoten, line 182: This was never quite right until this new syntax became available.

common/rbnf/it.xml
```diff
@@ -507,18 +453,17 @@ x.x: =#,##0.#=;
...
+2000000: <%spellout-cardinal-masculine<milione­>%%ordinal-esime>;
 1000000000: miliard­>%%ordinal-esime-with-o>;
-2000000000: <%spellout-cardinal-feminine<miliard­>%%ordinal-esime-with-o>;
+2000000000: <%spellout-cardinal-masculine<miliard­>%%ordinal-esime-with-o>;
 1000000000000: bilione­>%%ordinal-esime>;
-2000000000000: <%spellout-cardinal-feminine<bilion­>%%ordinal-esime>;
+2000000000000: <%spellout-cardinal-masculine<bilion­>%%ordinal-esime>;
 1000000000000000: biliard­>%%ordinal-esime-with-o>;
-2000000000000000: <%spellout-cardinal-feminine<biliard­>%%ordinal-esime-with-o>;
+2000000000000000: <%spellout-cardinal-masculine<biliard­>%%ordinal-esime-with-o>;
```
grhoten, line 466: The apocopic form was really meant.

common/rbnf/it.xml
```diff
@@ -531,7 +476,7 @@ x.x: =#,##0.#=;
 7: ­settesime;
 8: ­ottesime;
 9: ­novesime;
-10: =%spellout-ordinal-feminine=;
+10: =%spellout-ordinal-feminine-plural=;
```
grhoten, line 479: This should remain the plural form instead of switching to the singular form.

srl295 (approved):
plural syntax makes it a lot more compact

PR #5445, opened 2026-03-09
CLDR-5108 Maltese RBNF improvements

grhoten:
CLDR-5108

These changes improve the Maltese RBNF rules. No roundtripping errors were observed when testing these rules with the Number Format Tester.

ALLOW_MANY_COMMITS=true


common/rbnf/mt.xml
```diff
@@ -19,8 +19,6 @@ For terms of use, see http://www.unicode.org/copyright.html
 x.x: =0.0=;
 0: =%spellout-numbering=;
 %spellout-numbering:
-0: =%%spellout-cardinal-type-b-masculine=;
-%%spellout-cardinal-type-b-masculine:
```
grhoten, line 23: This private rule doesn't contribute much. So it's just collapsed into the `%spellout-numbering` form.

common/rbnf/mt.xml
```diff
@@ -54,80 +52,23 @@ x.x: << punt >>;
 100: mija[ u >>];
 200: mitejn[ u >>];
 300: <%spellout-cardinal-masculine< mija[ u >>];
-1000: elf[>%%and-type-b-masculine>];
```
grhoten, line 57: This was a convoluted way to self reference and insert u when >> is not zero. The brackets already do that, so it's best to just use the power of the square brackets, and simplify this rule.

common/rbnf/mt.xml
```diff
@@ -54,80 +52,23 @@ x.x: << punt >>;
...
+2000000000000: <%spellout-cardinal-masculine< triljuni[ u >>];
+11000000000000/1000: <%spellout-cardinal-masculine< triljun[ u >>];
+1000000000000000: kvadriljun[ u >>];
+2000000000000000: <%spellout-cardinal-masculine< kvadriljuni[ u >>];
+11000000000000000/1000: <%spellout-cardinal-masculine< kvadriljun[ u >>];
 1000000000000000000: =#,##0=;
-%%and-type-b-masculine:
-0: ;
-1: ' u =%%spellout-cardinal-type-b-masculine=;
-%%spellout-cardinal-type-b-feminine:
```
grhoten, line 77: This private rule is unused. It can be omitted.

common/rbnf/mt.xml
```diff
@@ -248,6 +196,22 @@ x.x: << punt >>;
 8: tmint;
 9: disat;
 10: għaxart;
+11: =%spellout-cardinal-feminine=;
+%spellout-ordinal:
```
grhoten, line 200: I added the indefinite ordinals.  The definite ordinals seem a little too complicated to write. Some of the other languages do have both the indefinite and definite forms.  Bulgarian has that distinction too.

It sounds like it switches to the numbering form at 11. If I'm wrong, I can extend it out further, if you provide some guidance.

common/rbnf/mt.xml
```diff
@@ -206,40 +144,50 @@ x.x: << punt >>;
...
+1000000000000000: kvadriljun[ u >>];
+2000000000000000: <%spellout-cardinal-masculine< kvadriljuni[ u >>];
+11000000000000000/1000: <%spellout-cardinal-masculine< kvadriljun[ u >>];
 1000000000000000000: =#,##0=;
-%%and-type-a-feminine:
-0: ;
-1: ' u =%spellout-cardinal-feminine=;
-%%thousands:
-0: ERROR-=0=;
+%spellout-cardinal-masculine-connective:
```
grhoten, line 176: The cardinal forms have the attributive non-connective forms, or short forms. I'm exposing these new attributive forms that connect with the referring noun, or the long form. It's kind of like the "a" and "an" allomorphs of the same indefinite article in English. In Maltese, it's a little different. When these forms are attributive forms are used, they can modify the spelling of the noun. Since it's a collection of rules on when to use this form, I've named it "connective" because it's used when modifying the noun when it's needed. Such a name might be reusable as a name for other languages that have elided forms, like French, and Italian.

This form was previously being used for the thousands form.

See the following for a description.
https://en.wiktionary.org/wiki/tlieta#Maltese

srl295 (approved):
Naħseb kollox sewwa.

PR #5458, opened 2026-03-11
CLDR-10024 Add tests for RBNF

grhoten:
CLDR-10024

We're adding 2 types of new tests. This is a lot easier to do with the flattened RBNF rules.

- One test involves conformance tests. These were generated from the existing rules. I didn't check them for accuracy.
- One test involves roundtripping every public ruleset in every language with a set of values.

The current set of tests depend on several other open pull requests being merged before merging these changes. That includes important fixes for Serbian (#5456), and relevant new rules.

ALLOW_MANY_COMMITS=true


common/rbnf/ar.xml
```diff
@@ -139,7 +139,7 @@ x.x: << فاصل >>;
 1000000000000000000: =#,##0=;
 %spellout-cardinal-masculine:
 -x: ناقص >>;
-x.x: <%%spellout-numbering-m< فاصل >> ;
+x.x: <%%spellout-numbering-m< فاصل >>;
```
grhoten, line 142: Whitespace typo getting fixed.

common/rbnf/el.xml
```diff
@@ -255,6 +255,7 @@ x.x: =#,##0.#=;
 1000000000000000000: =#,##0=;
 %spellout-ordinal-neuter:
 -x: μείον >>;
+x.x: =0.#=;
```
grhoten, line 258: The exact rule doesn't matter. It just needs to sort of roundtrip.

common/rbnf/he.xml
```diff
@@ -265,14 +265,15 @@ x.x: << נקודה >>;
...
-x.x: <%%spellout-numbering-m< נקודה >> ;
+x.x: <%%spellout-numbering-m< נקודה >>;
 0: =%spellout-cardinal-masculine=;
 2: שניים;
 3: =%spellout-cardinal-masculine=;
 %spellout-cardinal-feminine-standalone:
 0: =%spellout-numbering=;
 %spellout-construct-masculine:
 -x: מינוס >>;
+x.x: << נקודה >>;
```
grhoten, line 276: Oops! The rule was missing.

common/rbnf/hr.xml
```diff
@@ -59,20 +59,11 @@ x.x: << zarez >>;
...
-2000000000000: <%spellout-cardinal-masculine< bilijuna[ >>];
-5000000000000: <%spellout-cardinal-masculine< bilijuna[ >>];
-1000000000000000: <%spellout-cardinal-feminine< bilijarda[ >>];
-2000000000000000: <%spellout-cardinal-feminine< bilijarde[ >>];
-5000000000000000: <%spellout-cardinal-feminine< bilijardi[ >>];
+2000: <%spellout-cardinal-feminine< $(cardinal,one{tisuću}few{tisuće}other{tisuća})$[ >>];
+1000000: <%spellout-cardinal-masculine< $(cardinal,one{milijun}few{milijuna}other{milijuna})$[ >>];
+1000000000: <%spellout-cardinal-feminine< $(cardinal,one{milijarda}few{milijarde}other{milijardi})$[ >>];
+1000000000000: <%spellout-cardinal-masculine< $(cardinal,one{bilijun}few{bilijuna}other{bilijuna})$[ >>];
+1000000000000000: <%spellout-cardinal-feminine< $(cardinal,one{bilijarda}few{bilijarde}other{bilijardi})$[ >>];
```
grhoten, line 66: Known roundtrip issue that is fixed by switching to cardinal plural rules.

common/rbnf/ja.xml
```diff
@@ -14,11 +14,13 @@ For terms of use, see http://www.unicode.org/copyright.html
         <rulesetGrouping type="SpelloutRules">
             <rbnfRules><![CDATA[
 %spellout-numbering-year-latn:
+-x: マイナス>>;
```
grhoten, line 17: Not an important rule, but it's needed for roundtripping.

common/rbnf/vi.xml
```diff
@@ -55,6 +55,7 @@ x.x: << phẩy >>;
 1000000000: << tỷ[ >%%after-hundred>];
 1000000000000000000: =#,##0=;
 %spellout-ordinal:
+-x: âm >>;
```
grhoten, line 58: Not an important rule, but it's needed for roundtripping.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
```diff
@@ -0,0 +1,314 @@
...
+        }
+        for (String xmlFile : xmlFiles) {
+            checkRoundtrip(xmlFile);
+        }
+    }
+
+    private static final Set<String> KNOWN_BROKEN_LOCALES =
+            new TreeSet<>(Arrays.asList("ga", "ky", "lt", "su"));
+    private static final Set<String> ALIASES =
+            new TreeSet<>(Arrays.asList("nb", "en_001"));
```
grhoten: The known broken locales have roundtrip issues. We will just warning about them.

We skip over the known aliases where we are acknowledging their existence, but we don't really care that they're empty.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestRBNF.java
```diff
@@ -0,0 +1,314 @@
...
+    private static final Map<String, Number[]> ROUNDTRIP_VALUES = new HashMap<>();
+
+    static {
+        ROUNDTRIP_VALUES.put(
+                "SpelloutRules",
+                new Number[] {
+                        -1L, 0L, 0.2, 1L, 1.1, 2L, 3L, 10L, 99L, 100L, 101L, 999L, 1000L, 1001L,
+                        1999L, 2000L, 2001L, 2100L, 2200L, 10000L, 20000L, 100000L, 200000L, 1000000L,
+                        2000000L
+                });
```
grhoten, line 284: These are values that we're roundtripping for the spellout rules.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestRBNF.java
```diff
@@ -0,0 +1,314 @@
...
+            checkRoundtrip(xmlFile);
+        }
+    }
+
+    private static final Set<String> KNOWN_BROKEN_LOCALES =
+            new TreeSet<>(Arrays.asList("ga", "ky", "lt", "su"));
+    private static final Set<String> ALIASES =
+            new TreeSet<>(Arrays.asList("nb", "en_001"));
+
+    private static final Number[] ROUNDTRIP_LONG_VALUES = {
```
grhoten, line 246: These are values that we're roundtripping for the non-spellout rules. Those are normally positive integers.

common/rbnf/ak.xml
```diff
@@ -40,7 +40,7 @@ x.x: << pɔw >>;
 20: aduonu[->%%spellout-cardinal-tens>];
 30: aduasa[->%%spellout-cardinal-tens>];
 40: adu<<[->%%spellout-cardinal-tens>];
-100: ­ɔha[-na-­>>];
+100: ɔha[-na->>];
```
grhoten, line 43: Unnecessary soft hyphens are here. I'm not sure why the tests sometimes failed on this locale, but also they shouldn't be present in this scenario. I was using a slightly different version of ICU with these tests.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
```diff
@@ -0,0 +1,542 @@
...
+            } else {
+                numStr = String.valueOf(d);
+            }
+        } else {
+            numStr = String.valueOf(number.longValue());
+        }
+        return type + ";" + ruleSetName + ";" + numStr + ";" + formatted;
+    }
+
+    public static void main(String[] args) {
```
grhoten: If you need to seed the conformance tests, just run this class, and it will automatically generate some sample conformance tests.

common/rbnf/ky.xml
```diff
@@ -77,7 +77,7 @@ x.x: << бүтүн >%%z-spellout-fraction>;
 1000: миң;
 1001: << миң[>%%and>];
 2000: << миң[>%%and>];
-100000: << миң[>%%commas>];
+100000/1000: << миң[>%%commas>];
```
grhoten, line 80: Oops! The original syntax probably wasn't intended, since the word is the same as the thousand. Now 1000 and 100000 don't conflict on the value.

common/rbnf/ky.xml
```diff
@@ -125,34 +104,34 @@ x.x: =#,##0.#=;
...
-2000000000000: <%spellout-numbering< триллион>%%yncy>;
+1000000000001: <%spellout-numbering< триллион[ >>|унчу];
+2000000000000: <%spellout-numbering< триллион[ >>|унчу];
 1000000000000000: квадриллионунчу;
-1000000000000001: <%spellout-numbering< квадриллион>%%yncy>;
-2000000000000000: <%spellout-numbering< квадриллион>%%yncy>;
-1000000000000000000: =#,##0='inci;
+1000000000000001: <%spellout-numbering< квадриллион[ >>|унчу];
+2000000000000000: <%spellout-numbering< квадриллион[ >>|унчу];
+1000000000000000000: =#,##0='инчи;
```
grhoten, line 134: Oops! Wrong script. It now matches the digits-ordinal ruleset. Also modernizing the syntax magically fixed the roundtripping issue.

common/rbnf/su.xml
```diff
@@ -35,14 +35,15 @@ x.x: << titik >>;
...
+1000: sarebu[ >>];
+2000: << rebu[ >>];
+1000000: sajuta[ >>];
+2000000: << juta[ >>];
+1000000000: samiliar[ >>];
+2000000000: << miliar[ >>];
 1000000000000: =#,##0=;
-%%spellout-cardinal-large:
-1: sa;
-2: =%spellout-cardinal= ;
```
grhoten, line 45: The extra space at the end was confusing the ICU parsing for some reason. Making it structured more like the other languages fixed the issue.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
```diff
@@ -0,0 +1,567 @@
...
+            return;
+        }
+        for (String xmlFile : xmlFiles) {
+            checkRoundtrip(xmlFile);
+        }
+    }
+
+    private static final Set<String> KNOWN_BROKEN_LOCALES =
+            new TreeSet<>(Arrays.asList("ga", "lt"));
+    private static final Set<String> ALIASES = new TreeSet<>(Arrays.asList("nb", "en_001"));
```
grhoten: The known broken locales have roundtrip issues. We will just warn about them for now instead of failing on them.

We skip over the known aliases where we are acknowledging their existence, but we don't really care that they're empty.

common/rbnf/hr.xml
```diff
@@ -177,30 +150,21 @@ x.x: =#,##0.#=;
...
-900: devetst[ >>];
+100: stot[ >>];
+200: dvestot[ >>];
+300: tristot[ >>];
+400: četristot[ >>];
+500: petstot[ >>];
+600: šestot[ >>];
+700: sedamstot[ >>];
+800: osamstot[ >>];
+900: devetstot[ >>];
```
grhoten, line 161: This is a big ol' truncation typo. Number Format Tester confirms that these were wrong.

This source reference confirms the correct way to spell 3 examples.
https://www.lets-learn.eu/croatian/guide/ordinal-numbers-in-croatian

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
macchiati: SGTM. We just want a log known issue so we can track.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
macchiati: We'll want to restructure this so that [GenerateTestData.java](https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateTestData.java) can call it to generate the test data for every locale (that has RBNF data).

Cf, [GenerateTestData.md](https://github.com/unicode-org/cldr/blob/main/docs/site/development/GenerateTestData.md)

macchiati: **Note for the future.** Ideally we would be able to find out the set of rules that are **not** triggered by a set of numbers. That way we could add additional test cases to ensure that that set is empty.

@richgillam Any ideas how this could be done easily? I can imagine a brute-force approach, which would be to 
1. from set of rules (as strings), generate the outcomes for the set of test numbers.
2. in a loop, make a new set of rules by deleting one rule, and generate the outcomes
3. If the outcomes are different, then we know that the deleted rule was triggered by the test set of numbers.
4. Output the set of rules that didn't make a difference.

richgillam: I'm not completely sure I understand what you're talking about here.  The idea is to figure out if any rules in a proposed ruleset are superfluous?  Or not covered by the current test set?  Or both?

I think what you're suggesting would work, but it'll be slow and complicated.  For a simple ruleset that doesn't call out to any other rulesets, you'd really just be looking for all the "special" rules (negative, fractional part, etc.) ad to make sure there's a rule for zero-- all higher values would automatically be covered.  I guess you'd also want to check that no two rules have the same base value (and you'd have to do the checks on a compiled ruleset to account for any rules that use the `[]` notation).

The more interesting check would be if all the rules are exercised by the set of test input values, but I'd think you could do that by iterating over the rules and making sure there's a value in the test set that's in each rule's operating range (if any rules use the `[]` notation, they expand to two rules under the covers, but if you're operating on a compiled ruleset, you'd know that.

This all gets a lot more complicated if you have rulesets calling other rulesets, because then for each ruleset you'd have to derive the set of values that'd cause that ruleset to be called in the first place.  Anything in the ruleset that operates only on values outside that range would be superfluous.  I'm not _excatly_ sure how you'd derive that set, but it seems possible...

grhoten: > For a simple ruleset that doesn't call out to any other rulesets, you'd really just be looking for all the "special" rules (negative, fractional part, etc.) ad to make sure there's a rule for zero-- all higher values would automatically be covered. I guess you'd also want to check that no two rules have the same base value (and you'd have to do the checks on a compiled ruleset to account for any rules that use the `[]` notation).

The currently generated tests just picked a few common values. If you want me to deduce the test values from the rules, that would not be too hard. Though if I do that, what's the time budget for doing that? Right now it runs over the tests in a couple of seconds.

Is the request to roundtrip each value that has a rule?

> The more interesting check would be if all the rules are exercised by the set of test input values, but I'd think you could do that by iterating over the rules and making sure there's a value in the test set that's in each rule's operating range (if any rules use the `[]` notation, they expand to two rules under the covers, but if you're operating on a compiled ruleset, you'd know that.

This is a good idea. It just takes extra time to implement. I was just trying to get something implemented in a day, and fix the simple stuff that arised.

The cardinal and ordinal rules would need to pull in the plural data, which would be a little harder to reverse, unless there is sample data to leverage instead.

> This all gets a lot more complicated if you have rulesets calling other rulesets, because then for each ruleset you'd have to derive the set of values that'd cause that ruleset to be called in the first place. Anything in the ruleset that operates only on values outside that range would be superfluous. I'm not _excatly_ sure how you'd derive that set, but it seems possible...

With the additions for the cardinal/ordinal syntax, and the or-else pipe operator in the brackets, calling out to private rules have been reduced. I recommend a shallow inspection. For any twists and turns, it's probably better to depend on manually adding such test cases to the conformance files.

If the request for tests gets too complicated, I'm likely to just punt that work. I'd prefer to keep it simple for now.

common/rbnf/el.xml
```diff
@@ -248,13 +248,14 @@ x.x: =#,##0.#=;
 9000: εννεάκις χιλιοστή[ >>];
 10000/1000: δεκάκις χιλιοστή[ >>];
 11000/1000: <%spellout-cardinal-neuter< χιλιοστή[ >>];
-1000000: <%spellout-cardinal-neuter< εκατομμυριοστή [ >>];
+1000000: <%spellout-cardinal-neuter< εκατομμυριοστή[ >>];
```
grhoten, line 251: Whitespace typo. It's doubling up the whitespace, which doesn't seem right.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
grhoten: I have moved the test file generation code to GenerateRBNFTestData.java. It also now parses the rules for the base numbers to test, and it also tests the numbers ±1 around that base number up to 2^53-1.  After that number, you start hitting rounding errors when using IEEE754.  9 quadrillion is generally good enough. All of the rules tend to stop at 1 quintillion anyway.

Testing around a base value does tend to hit the "other" plural rule, and it helps to find gaps in the rules.

I regenerated all of the files.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestRBNF.java
srl295, line 1: General Note: Please don't add an additional old-style test.  Completely new tests should be JUnit, see https://github.com/unicode-org/cldr/blob/main/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestIso3166Data.java for example ( and put it in that package, not under unittest).

Most of what this involves is using `@Test` before any test function and using modern assert functions.

grhoten: How do you log warnings or other known issues? It's stuff like disabled tests, or known issues, but it shouldn't block the merge process.

Do I have to derive from TestWithKnownIssues, and use logKnownIssue?

srl295: > Do I have to derive from TestWithKnownIssues, and use logKnownIssue?

yes

we can move it to JUnit after this goes in, just trying not to add more testfmwk tests

srl295: you can use parameterized tests, and use `assumeTrue()` where if the assumption is false that subtest gets skipped. See TestCLDRLinks for an example (this one skips subtests due to configuration, but similar idea)

sffc (commented):
Thanks for the rbnf testdata! This can be added to the conformance dashboard.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateRBNFTestData.java
```diff
@@ -0,0 +1,324 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import org.unicode.cldr.util.CLDRPaths;
+import org.unicode.cldr.util.Pair;
+import org.unicode.cldr.util.XMLFileReader;
+import org.unicode.cldr.util.XPathParts;
+
+public class GenerateRBNFTestData {
```
srl295, line 24: ideally add a `@CLDRTool` annotation so we can find this in the list of tools

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
```diff
@@ -0,0 +1,392 @@
...
+        if (ssvFiles == null || ssvFiles.length == 0) {
+            errln("No .ssv test files found in " + TEST_DATA_DIR);
+            return;
+        }
+        for (String ssvFile : ssvFiles) {
+            checkLocaleFile(ssvFile);
+        }
+    }
+
+    private void checkLocaleFile(String ssvFile) {
```
srl295: Doesn't this make this a 'golden' test against the ssv files? What is the value add since they are generated?  I'd expect such a test in ICU instead perhaps.

srl295 (dismissed):
tests seem OK. On the conformance tests against the generated files seems like it really testing that the generation machinery works.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestRBNF.java
grhoten: It was part of the request for this ticket. Also the roundtrip test just just blindly roundtrips the value. The conformance tests allow you to manually review the results, if you so desire. It has found a few oddities that the roundtrip tests didn't find. By generating the files with default data, you get to spend more time reviewing the file instead of painstakingly write the file by hand.

If you had an alternative to ICU's RBNF that could use these rules, these files can be used for ensuring the behavior is the same as ICU.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateRBNFTestData.java
grhoten: I have made this change.

common/rbnf/lt.xml
```diff
@@ -391,11 +391,19 @@ x.x: =#,##0.#=;
 10000: =#,##0=;
 %spellout-ordinal-feminine:
 0: =%%spellout-ordinal-base=a;
+100: šimt[as >>|a];
+200: <%spellout-cardinal-masculine<šimt[ai >>|a];
+1000: tūkstant[is >>|a];
+2000: <%spellout-cardinal-masculine<tūkstant[ai >>|a];
```
grhoten, line 397: Fixes a roundtripping issue.

grhoten:
@macchiati The tests roundtrip now. I had to merge on top of the other pull requests to get them to pass. There is one French pull request left. I can update the relevant ssv files to match that anticipated behavior, instead of the current behavior, if you want me to do that.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestRBNF.java
```diff
@@ -0,0 +1,405 @@
...
+                    Double.NaN
+                });
+        ROUNDTRIP_VALUES.put("OrdinalRules", ROUNDTRIP_LONG_VALUES);
+        ROUNDTRIP_VALUES.put("NumberingSystemRules", ROUNDTRIP_LONG_VALUES);
+    }
+
+    private void checkRoundtrip(String xmlFile, List<String> errors) {
+        String localeId = xmlFile.substring(0, xmlFile.length() - 4);
+        if (KNOWN_BROKEN_LOCALES.contains(localeId)) {
+            System.err.println("WARNING: Skipping known broken locale: " + localeId);
```
srl295, line 292: Leave this as a TODO in the ticket to move this to a logknown issue (I can do it)

srl295 (approved):
Looks good, i think we should get this in and then update the test to use logKnownIssues.

PR #5463, opened 2026-03-14
CLDR-19275 move webdriver to be a proper subproject

srl295:
- change webdriver main to be a main() instead of a test
- allow webdriver to inherit versions from parent

The PR #5403 pointed out that we have two copies of the spotless machinery. This starts to reduce that.

CLDR-19275

ALLOW_MANY_COMMITS=true


srl295:
builds, doesn't work yet.

btangmu:
> builds, doesn't work yet.

How does it fail? Should I investigate?

srl295:
@btangmu you can see the failure https://github.com/unicode-org/cldr/actions/runs/23076569629/job/67038129961#step:11:26 - i think I just didn't wire up the test properly yet.

THe point of this is to normalize the webdriver's build so it's part of the other projects, rather than separate (as if it were still in a separate repo) so that it can benefit from dependency updates etc.  It had its own copy of the spotless config for example.

.github/workflows/maven.yml
```diff
@@ -315,13 +321,10 @@ jobs:
...
       run: docker compose up -d cldr-apps selenium
       working-directory: tools/cldr-apps
-    - name: Build webdriver
-      run: docker compose run -v ~/.m2/repository:/root/.m2/repository:rw --rm webdriver mvn -B test-compile
-      working-directory: tools/cldr-apps
     - name: Test with Webdriver
       # See tools/cldr-apps/README.md
       run: >
-        docker compose run -v ~/.m2/repository:/root/.m2/repository:rw --rm webdriver || ( docker compose exec -it cldr-apps tail /logs/messages.log ; true )
+        docker compose run --rm webdriver || ( docker compose exec -it cldr-apps tail /logs/messages.log ; true )
```
btangmu: main failure occurs here?

btangmu: org.openqa.selenium.SessionNotCreatedException "Could not start a new session. Possible causes are invalid address of the remote server or browser start-up failure." caused by java.nio.channels.ClosedChannelException "I/O operation on a channel that has already been closed"

btangmu:
There are error messages from setup-config.sh involving /tmp/cldr-code.jar -- not clear whether these are central to the main failure.

setup-config.sh runs in cldr-apps/Dockerfile, not cldr-apps-webdriver/Dockerfile. So it's cldr-apps, not cldr-apps-webdriver, that fails here. That suggests the error is "invalid address of the remote server" rather than "browser start-up failure".

This command has two parts:

`docker compose run --rm webdriver || ( docker compose exec -it cldr-apps tail /logs/messages.log ; true )`

Seems like it's the second part (`docker compose exec -it cldr-apps...`) that's failing if the setup-config.sh errors are the cause. That would imply the first part (`docker compose run --rm webdriver`) succeeds, in which case I guess the browser starts up OK...

btangmu:
Why is it that webdriver runs before cldr-apps? Could cldr-apps start first instead?

srl295:
> Why is it that webdriver runs before cldr-apps? Could cldr-apps start first instead?

Yes, or at least it can be set to wait until cldr-apps starts.


BTW i found that _not_ having this, means that intellisense is broken for webdriver— because it can't see the disconnected sub-project.

PR #5492, opened 2026-03-19
CLDR-18954 Use tags (chits) in ST value input dialog

btangmu:
-Value input dialog box has new eye icon (show hidden) for displaying the value with tags (chits) in place of special characters (except ASCII space)

-The new features are suppressed if the path uses UnicodeSet (exemplars, parseLenient, ...)

-Special characters are defined by cldrEscaper.mjs

-Support four modes for tags: no tags; basic tags; tags with menus; and tags with editing

-Only tags with menus are available through the obvious visible interface. The other modes are only accessible for experimental testing if TEST_ALTERNATIVE_MODES is true and the user Shift-clicks on the Cancel button, bringing up a menu with radio buttons to switch modes

-Tags with menus mode enables the user to click on any tag to change it to a different special character chosen from a menu

-Tags with menus mode also enables the user to click on the new ⎀ (insert) button to insert (at the insertion point) a special character chosen from a menu (⎀ = U+2380 INSERTION SYMBOL, might be replaced by different icon in the future)

-Basic tags mode (currently hidden behind TEST_ALTERNATIVE_MODES) displays each character (special or not) as a tag, with a hover to display its Unicode Scalar Value and name

-Tags with editing mode (currently hidden behind TEST_ALTERNATIVE_MODES) is like basic tags mode, but in addition each tag has an x for deleting it, and between each pair of adjacent tags there is a + for inserting a new tag. Also, clicking in a tag enables changing to a different character, either by entering the character directly, or by typing its Unicode Scalar Value as U+xxxx

-The alternative modes may in the future be useful for special kinds of path, such as exemplars and annotations

-Some special characters may be appropriate only for certain locales (e.g., ALM = Arabic letter mark), so a future enhancement may be to include them in menus only if appropriate for the current locale

CLDR-18954

ALLOW_MANY_COMMITS=true


btangmu:
This screenshot shows the add-value dialog with "tags" checked:

image

When pointing to a tag, there is a hover describing the tag:

image


Clicking on the tag brings up a menu for changing its value:

image


There is also an "I" (insert) button, for inserting a new special character at the insertion point:

image


For long values, such as exemplar characters, the tags wrap:

image

(Note that changing SP to anything else would not generally be useful for exemplars, and a different tag mode focused on the character values rather than the spaces between them, could be more useful.)

AEApple:
Where are we on resolving this ticket?

btangmu:
> Where are we on resolving this ticket?

I'm still working on making the menus work the way Mark requested, with a single click. It mostly works but there are visual artifacts I'm trying to get rid of. The Ant Vue library doesn't seem to have a documented way to accomplish it, so there's a lot of trial and error so far.

macchiati:
I had made some comments in slack and elsewhere. Repeating here:

1. Never show plain space with a chit. 
    * So we won't see the wild wrapping as one of your images
2. Replace the checkbox + tags label with an 'eye' icon / 'slashed eye' icon.
    * Use the non-slashed for showing, and slashed version for hiding.
    * The hover text should be "Show Hidden"
    * Well-understood in lots of interfaces for showing/hiding fields, such as passwords. Like https://www.shutterstock.com/search/show-hide-icon (but licensed for our usage). 
    * Takes less room
4. The button for "insert special character at insertion point" 
    * should have hover text "Insert special characters"
    * should just be a pulldown menu,*not* bring up a special box.
    * The menu would be cleaner, if possible, by showing just the abbreviation, with the code point and description being in 'hover' text. 

Here's a rough sketch of the result. The icons & positioning could vary. (This doesn't show all the recommended hover text.)

## Plain box. Shows with "hidden" by default unless the text contains a chitable character.

Screenshot 2026-04-14 at 13 17 08

## With Show hidden turned on

Screenshot 2026-04-14 at 13 17 37

## With the Insert Character menu pulled down

Screenshot 2026-04-14 at 13 19 12

macchiati:
Put this in a doc also: https://docs.google.com/document/d/13xjlcwZgjyMAkQ5XD66reeSW8ssd9EJ3XSuola3E2oE/edit?tab=t.0

btangmu:
My earlier comments and screenshots are outdated. Here is a new screenshot:

image

btangmu:
I think I've covered what's in the two design docs. I could post more screenshots, however I think the interactivity is what really needs testing. Shall I push this to the cldr-staging.unicode.org server for testing? Alternatively we could merge the PR and fix or disable the new features if problems are discovered.

tools/cldr-apps/js/src/esm/cldrXpathUtils.mjs
```diff
@@ -36,4 +36,37 @@ function extraPathAllowsNullValue(path) {
   return false;
 }
 
-export { extraPathAllowsNullValue };
+function pathUsesUnicodeSet(path) {
```
srl295, line 39: This really should come from java.  maybe as part of the escaper's payload.

Also these are partial paths so could potentially have false positives.

OK for now but i would restructure it later, this will very easily get out of sync.  In fact, usesUnicodeSet could be part of the row payload.

btangmu, line 39: @macchiati this array is copied from CheckForExemplars.java; is it appropriate for this purpose?

tools/cldr-apps/js/src/views/AddValueCharMenu.vue
```diff
@@ -0,0 +1,175 @@
...
+  if (DEBUG) {
+    console.log("handleDropdownVisibleChange: isVisible = " + isVisible);
+  }
+}
+</script>
+
+<style>
+/* Not scoped, for overriding ant styles */
+
+/* The div with "ant-select-selector" normally contains the menu arrow and placeholder.
```
srl295, line 148: These should be a known issue, check with the latest ant, and have an issue filed upstream.   Otherwise this is very fragile because it presumes on internals of ant.

srl295 (commented):
Code looks fine to me with comments.

Getting it on staging does sound like a good path forward, if review of it interactively is the next part of the review.

macchiati (approved):
If Stephen has checked it over, then I agree with him. We should just go ahead and merge it and then have follow-up PRS to address any lingering issues

tools/cldr-apps/js/src/views/AddValueTags.vue
```diff
@@ -0,0 +1,153 @@
...
+   Even if showArrow=false and placeholder=null, an empty div is displayed as an
+   obnoxious little rectangle overlapping the tag to which the menu is attached,
+   unless this style is overridden, and it can't be "style scoped". (Note that
+  "display: none" would hide the entire menu.) */
+.ant-select-selector {
+  width: 0 !important;
+  height: 0 !important;
+  border: 0 !important;
+}
+</style>
```
srl295, line 142: this breaks the "Add User" dialog. Why are we overriding ant's css without scope?

btangmu: With "style scoped" it would be ignored. Overriding appears only to be possible without scope.

The reason for overriding is explained in the comment.

That's bad news about "Add User". I did express concerns about trying to make Ant Vue do things it wasn't designed to do. For now I think the priority should be to revert this change in order to fix "Add User", and then to find a compromise so that the tag menus work, maybe requiring a container around the menu and an additional click to open it.

PR #5537, opened 2026-04-01
CLDR-19359 Release notes for 48.2

btangmu:
CLDR-19359

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -6,9 +6,9 @@ title: CLDR 48 Release Note
 
 | No. |    Date    | Rel. Note |  Data  |  Charts  | Spec |   Delta  | GitHub Tag | Delta DTD | CLDR JSON |
 |:---:|:----------:|:---------:|:------:|:--------:|:------------:|:---:|:----------:|:---------:|:---------:|
-|  48 | 2025-10-29 | [v48] | [CLDR48] | [Charts48]| [LDML48] | [Δ48] | [release-48] | [ΔDtd48] | [48.0.0] |
+|  48 | 2025-10-29 | [v48] | [CLDR48] | n/a | [LDML48] | [Δ48] | [release-48] | [ΔDtd48] | [48.0.0] |
 |  48.1 | 2026-01-08 | [v48.1] | [CLDR48.1] | n/a | [LDML48.1] | [Δ48.1] | [release-48-1] | n/a | [48.1.0] |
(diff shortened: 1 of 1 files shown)


btangmu:
Please see my notes in the ticket. I'm making this as a *draft* PR for discussion. It's not clear whether this is the right approach as far as dot-releases are concerned...

AEApple:
I thought we had agreed that we would keep the initial CLDR 48 charts (major version) and then whatever the most recent dot release was (minor version)?

srl295:
> I thought we had agreed that we would keep the initial CLDR 48 charts (major version) and then whatever the most recent dot release was (minor version)?

OK so basically `/48` (.0) plus `/48.2` in this case?

btangmu:
> > I thought we had agreed that we would keep the initial CLDR 48 charts (major version) and then whatever the most recent dot release was (minor version)?
> 
> OK so basically `/48` (.0) plus `/48.2` in this case?

If so, we need to reconsider these questions I asked in the ticket:

- What should be the URL for 48.2 charts? Should it end in “48” or “48.2”?
- What should be the branch for 48.2 charts? Should it end in “48” or “48.2”?

Based on our earlier discussion my understanding was "48" for both, but I don't think we can use the same URL or branch for distinct data.

srl295:
> > > I thought we had agreed that we would keep the initial CLDR 48 charts (major version) and then whatever the most recent dot release was (minor version)?
> 
> > 
> 
> > OK so basically `/48` (.0) plus `/48.2` in this case?
> 
> 
> 
> If so, we need to reconsider these questions I asked in the ticket:
> 
> 
> 
> - What should be the URL for 48.2 charts? Should it end in “48” or “48.2”?
> 
> - What should be the branch for 48.2 charts? Should it end in “48” or “48.2”?
> 
> 
> 
> Based on our earlier discussion my understanding was "48" for both, but I don't think we can use the same URL or branch for distinct data.

Thanks for remembering. 

The branch, url, and directory name must be all the same, that's how the deploy works.

macchiati:
We can't use the same URL. The question is whether we need to show
different charts for dot versions, or just maintain a single chart
that shows the latest dot version (which may be 0, of course).

We should definitely discuss this in the TC. It is not, however, a pre-sub
or pre-sub issue, so let's focus on the other issues until we hit
submission.

docs/site/downloads/cldr-48.md
srl295, line 376: I know this is out of scope but the links to the spec are broken here.  should probably be https://unicode.org/reports/tr35/48/tr35-general.html#nested-bracket-replacement etc

PR #5550, opened 2026-04-04
CLDR-19372 Remove %%lenient-parse rules from RBNF

grhoten:
CLDR-19372

ICU-23358 will be removing the unnecessary %%lenient-parse rules. Removing them will reduce the size of the rules, and improve the lenient parsing speed without using a collator.

CLDR occasionally defines this rule, but it’s inconsistently defined. All of the ones in CLDR are used to define ignorable characters for lenient parsing. We should remove these rules once https://github.com/unicode-org/icu/pull/3926 is merged into ICU.

ALLOW_MANY_COMMITS=true


richgillam (approved):
Re-approving.

PR #5562, opened 2026-04-09
CLDR-18759 Error candidate: restore red background

btangmu:
-The line with setDivClass was accidentally removed in a PR for a different ticket

-This does not fix the main bug for this ticket; some error items should not be present at all

CLDR-18759

ALLOW_MANY_COMMITS=true


srl295 (approved):

srl295:
> -The line with setDivClass was accidentally removed in a PR for a different ticket
> 
> 
> 
> -This does not fix the main bug for this ticket; some error items should not be present at all

Right. Not because they are errors per se but because they are inappropriate candidate options. 


> 
> 
> 
> CLDR-18759
> 
> 
> 
> - [ ] This PR completes the ticket.
> 
> 
> 
> 
> 
> 
> 
> ALLOW_MANY_COMMITS=true
> 
>

PR #5567, opened 2026-04-10
CLDR-18759 Do not use non-winning inherited empty exemplar set as a candidate item

btangmu:
-Skip inherited item if it has error missingMainExemplars and is not winning

-New method DataRow.testsExcludeCandidateItem

-Fix front end to display selected item even if displayValue is empty/falsy

-Encapsulate brackets as UnicodeSetPrettyPrinter.SET_OPEN/SET_CLOSE

-Make UnicodeSetPrettyPrinter.target final (fixing compiler warning)

CLDR-18759

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -804,7 +804,11 @@ public class DataPage {
                     new Output<>(); // may be used to construct inheritedLocale
             inheritedValue =
                     ourSrc.getBaileyValue(xpath, inheritancePathWhereFound, localeWhereFound);
-
+            if (UnicodeSetPrettyPrinter.EMPTY_SET.equals(inheritedValue)
+                    && xpath.contains("exemplarCharacters")) {
+                // Do not treat inherited empty exemplar set "[]" as a candidate item
+                inheritedValue = null;
+            }
             if (TRACE_TIME) {
                 System.err.println("@@1:" + (System.currentTimeMillis() - lastTime));
             }
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
@@ -804,11 +804,6 @@ public class DataPage {
                     new Output<>(); // may be used to construct inheritedLocale
             inheritedValue =
                     ourSrc.getBaileyValue(xpath, inheritancePathWhereFound, localeWhereFound);
-            if (UnicodeSetPrettyPrinter.EMPTY_SET.equals(inheritedValue)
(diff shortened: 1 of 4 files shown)


tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
```diff
@@ -804,7 +804,11 @@ private void updateInheritedValue(CLDRFile ourSrc, TestResultBundle checkCldr) {
                     new Output<>(); // may be used to construct inheritedLocale
             inheritedValue =
                     ourSrc.getBaileyValue(xpath, inheritancePathWhereFound, localeWhereFound);
-
+            if (UnicodeSetPrettyPrinter.EMPTY_SET.equals(inheritedValue)
+                    && xpath.contains("exemplarCharacters")) {
+                // Do not treat inherited empty exemplar set "[]" as a candidate item
+                inheritedValue = null;
```
srl295: I think this is the wrong layer.  There's no reason to mutate the value at this level, from a data integrity perspective. It really is a candidate, and it is the fallback.

On the client side, I would just skip showing candidate items that have this specific error,  IF there is any other winning value.

If for some reason the XML had [] as winning then we ought to get this error.

I think the requirement is just don't want to show this value as an "other" with that error. Not that [] should be suppressed everywhere.

srl295 (changes requested):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/DataPage.java
srl295: The error info is in the ticket, namely

```
cause: "CheckExemplars"
​​​​entireLocale: false
​​​​​message: "Exemplar set (main) must not be empty -- that would imply that this language uses no letters!"
​​​​​​subtype: "missingMainExemplars"
​​​​​​type: "Error"
```

I think missingMainExemplars should be excluded from the others column - that's all.  There may be other subtypes in this category in the future.

macchiati: I agree with Steven

btangmu: OK, the 2nd commit changes the code to exclude the item only if it has `CheckCLDR.CheckStatus.Subtype.missingMainExemplars` and is not winning. Also, it fixes the front end so that the selected candidate item is described in the Info Panel even if its `displayValue` is empty; otherwise the error message wasn't shown even if such an item was winning.

We might consider future improvements to the display when a candidate item (or its `displayValue`) is an empty string. As it stands, with this PR, the empty display value (corresponding to underlying value "[]") is displayed as a thin vertical red line:

image

srl295 (approved):
Looks great. Yeah we may have more in this category so I like the refactoring.

PR #5571, opened 2026-04-10
CLDR-14139 Remove ideographs from non-ideographic script languages

macchiati:
CLDR-14139

As in the ticket, this is low-hanging fruit for lowering the number of (correct) warnings.

ALLOW_MANY_COMMITS=true


AEApple:
Question before I approve this:

I didn't see it in [the spec](https://www.unicode.org/reports/tr35/tr35-general.html#annotations-and-labels), but would it make sense is we updated the spec to recommend including the Decomposition_Mapping for search keywords? e.g. ココ <U+30B3, U+30B3> for 🈁 , and then update the info hub for linguist so they don't try adding it back?

For me I would expect that I could get the emoji for a Japanese character by copy-pasting it or otherwise entering it into the search field, and it would be weird that it wouldn't work. I'm concerned that it may not cover this in the spec right now, or am I missing that in the explanation?

cc @eggrobin

AEApple:
also do we need to worry about any of the DDL languages which are currently already open if they have this character in their votes?

macchiati:
I was the one who originally added the plaintext version of various CJK emoji.

However, in hindsight that was overkill, because a very small percentage of people outside of CJK languages know those characters and would think to use them to search for the character. Many vetters have removed them, but many have kept them — I think out of inertia. They'll see the old value (with a blue star) in case those characters are really used in their languages.

The end goal is data quality, cutting down the number of spurious warnings we have by removing characters that are not in the locale's script (with some exceptions), so that vetters can focus. We also want them to propose adding characters that are used in their language, and should be in either main or aux.

macchiati:
Part of the goal is to exclude paths from the ST checks that can legitimately have characters outside of the script of the locale. For example, some non-Latin-script locales use K for compact decimals.

I also filed a followup ticket at https://unicode-org.atlassian.net/browse/CLDR-19380 to add characters to CJK locale exemplars.

macchiati:
> also do we need to worry about any of the DDL languages which are currently already open if they have this character in their votes?

These are all good questions, but I don't think we need to worry too much about the DDLs. We can take a later pass with them if needed.

Note: we used to do  https://unicode-org.atlassian.net/browse/CLDR-19380 on basically every release, but have lapsed; I think we should add that step back into the BRS.

AEApple:
> https://unicode-org.atlassian.net/browse/CLDR-19380

Maybe this could actually be an automated run?

AEApple (approved):

PR #5573, opened 2026-04-11
CLDR-19129 Better dev documentation for updating dtds

macchiati:
CLDR-19129

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/development/updating-dtds.md
+++ b/docs/site/development/updating-dtds.md
@@ -12,7 +12,9 @@ Sounds simple, right? But it isn't quite that easy.
 
 ## Summary
 
-In summary, when you add an element, attribute, or new kind of attribute value, there are some important steps you must also take. Note that running our unit tests and ConsoleCheck will catch most of these, but you should understand what is going on. Make sure that you don't break any of the invariants below (read through once to make sure you get them)! There is more detailed information further down on the page.
+In summary, when you add an element, attribute, or new kind of attribute value, there are some important steps you must also take. Note that running our unit tests and ConsoleCheck will catch most of these, but you should understand what is going on. 
+
+Make sure that you don't break any of the invariants below (read through once to make sure you get them)! There is more detailed information further down on the page.
 
 ### New Alt Values
 
@@ -28,8 +30,6 @@ We augment the DTD structure in various ways.
     - \<!--@ORDERED--> to indicate that an element's children are ordered.
     - \<!--@DEPRECATED--> to indicate that an attribute or element is deprecated.
     - \<!--@DEPRECATED:attribute-value--> to indicate that an attribute value is deprecated.
-2. attributeValueValidity.xml
-    - For additional validity checks
 3. Check\* tests and unit tests
     - There are many consistency tests that are performed on the data that can't be expressed with the above.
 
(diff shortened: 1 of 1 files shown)


macchiati:
I want to clean up the "updating DTDs" page to make sure we cover all the important information, since there are lots of details. I see this as the following steps:

1. Capture important information
2. Circulate for addition questions that people would have.
3. Add some more PR example, preferably in the sections they are relevant for.
4. Then, once the info is complete, restructure for clarity.

srl295 (approved):

docs/site/development/updating-dtds.md
AEApple, line 21: Is this still true when adding language specific alts? or only when adding an alt at the root level?

Also this is a circulat link, can we please fix it?

docs/site/development/updating-dtds.md
```diff
@@ -47,19 +47,70 @@ We augment the DTD structure in various ways.
     1. add @ORDERED if it is must be ordered.
     2. read more details below.
```
AEApple, line 48: Which more details below should I read about this?

docs/site/development/updating-dtds.md
```diff
@@ -47,19 +47,70 @@ We augment the DTD structure in various ways.
...
 2. For each attribute
-    1. add @VALUE or @METADATA to an !ATTLIST if the attribute is non-distinguishing. (See the spec for what this means)
+    1. If the attribute is on a new element, it can be either #REQUIRED or #IMPLIED (#IMPLIED is XML jargon for OPTIONAL).
+However, **all** new attributes on preexisting elements (those that were in a release) _must_ be #IMPLIED, because making them #REQUIRED would break past versions. (That is, all versions of CLDR need to be readable using all later DTDs.)
```
AEApple, line 51: What does break past versions mean here? Just if you try reading an earlier DTD with the latest version of CLDR, it would fail?

docs/site/development/updating-dtds.md
```diff
@@ -47,19 +47,70 @@ We augment the DTD structure in various ways.
...
+    2. add @VALUE or @METADATA to an !ATTLIST if the attribute is non-distinguishing. (See the spec for what this means)
```
AEApple, line 52: Can we link "the spec" to the right part of the spec to make it easy to find and access if you're reading these instructions?

docs/site/development/updating-dtds.md
```diff
@@ -47,19 +47,70 @@ We augment the DTD structure in various ways.
...
         1. **@VALUE should never occur except on leaf nodes!** (There are some cases before we realized this was a mistake.)
     2. If the attribute values are a closed set, you can add them explicitly, like:
         - \<!ATTLIST version draft (approved | contributed | provisional | unconfirmed) #IMPLIED>
     3. Otherwise
         1. Make it NMTOKEN where only single values are allowed, or NMTOKENS otherwise (CDATA in rare cases, but clear with the committee first)
```
AEApple, line 57: Should the use of CDATA require a CLDR/ICU+ Design WG recommendation to be considered?

docs/site/development/updating-dtds.md
```diff
@@ -47,19 +47,70 @@ We augment the DTD structure in various ways.
...
+
+As an other example, the `pattern` element can have very different contents depending on whether it is in 
+
+* `<pattern type="atLeast">{0}+</pattern>` or in 
+* `<pattern>EEEE, MMMM d, r(U)</pattern>`
+
+#### Ordering of attributes
+
+When you add an element or attribute, it is _vital_ to add them in the right place in the DTD.
+The right ordering among attributes is especially important, because that determines the ordering of attributes in an XPath that we create.
```
AEApple, line 86: Is there a document which describes what an XPath is and how we use it in CLDR?

docs/site/development/updating-dtds.md
AEApple: Wait, couldn't we have a test that would fail if you do this by accident? Do we have a test right now?

PR #5574, opened 2026-04-11
CLDR-16859 docker: webdriver test summary, first few sanity checks

srl295:
CLDR-16859

- Supports first few items of Sanity Check automatically
- for now, disabled the existing webdriver tests (may bring them back by default)
- Summary file [see example](https://github.com/unicode-org/cldr/actions/runs/24298408899#summary-70947450242) , uploaded screenshots, uploaded server logs— all in the Summary page of the Maven build
- updated docs

click-warning

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/AuthSurveyDriver.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/AuthSurveyDriver.java
@@ -5,6 +5,7 @@ import java.util.regex.Pattern;
 import org.unicode.cldr.util.CLDRConfig;
 import org.unicode.cldr.util.Organization;
 import org.unicode.cldr.util.StandardCodes;
+import org.unicode.cldr.util.VoteResolver.Level;
 import org.unicode.cldr.web.UserRegistry.User;
 
 public class AuthSurveyDriver {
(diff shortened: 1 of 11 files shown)


srl295:
think this is ready ~to start getting some review…~ i have more tests to add, but those could be done in a followup.

AEApple:
Report looks really nice! I tried scanning the tests but I'm not able to focus well right now. Hopefully someone else will be able to review more thoroughly since it will be exciting to see these running!

srl295:
> Report looks really nice! I tried scanning the tests but I'm not able to focus well right now. Hopefully someone else will be able to review more thoroughly since it will be exciting to see these running!

Thanks. Yeah, i will make followup PRs that complete the rest of the sanity check.

If we get reviews of the general approach being OK, i'd like to merge this and tweak what it tests separately.

macchiati (approved):
Looks good to me. As you say, we can refine later. 

That being said, I'm not as familiar with the internals for this, so Tom should weigh in.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/AuthSurveyDriver.java
```diff
@@ -94,13 +95,28 @@ private static User addTestUser(String password, String email, int userIndex) {
         u.name = EMAIL_PREFIX + userIndex;
         Organization[] orgArray = Organization.values();
         u.org = orgArray[userIndex % orgArray.length].name();
+        // Special users
+        if (userIndex == 1000) {
```
btangmu, line 99: Better to define a constant along with INVALID_USER_INDEX and  FIRST_VALID_USER_INDEX, could be in a follow-up PR

btangmu (approved):
Amazing!

tools/cldr-apps/src/main/java/org/unicode/cldr/web/AuthSurveyDriver.java
srl295: Or even some shared configuration, or a query parameter to actually request specific orgs… 

Good point and i'll take a note in the ticket. I'd like to finish the sanity check (once presub priorities are in) and let that check inform any further requirements here.

PR #5575, opened 2026-04-12
CLDR-14273 Suppress plurals in Survey Tool for Vietnamese

macchiati:
CLDR-14273

ALLOW_MANY_COMMITS=true


srl295 (approved):

srl295:
Looks fine, i think it might be best to have a TODO in referencing https://unicode-org.atlassian.net/browse/CLDR-15634 - but i know there's a limited time before we need to use this. SO hitting approve.

PR #5576, opened 2026-04-13
CLDR-9980 Add numeric date and numeric time separators

macchiati:
Adds two new paths for numeric date and numeric time separators, customizable per locale and per calendar.

These are based by analysing the date for regularities in the existing patterns. It is provisional, so vetters can adjust.

Will need to follow up with all of the Update DTD actions, so just draft for now. That will include a warning when the numeric date/time patterns use a different separator.

Fixed small problem in CLDRModify, triggered when new_value is a single /.

CLDR-9980

ALLOW_MANY_COMMITS=true


macchiati:
CC: @sffc and @AEApple

macchiati:
I'd like to get review for this today if possible.

sffc (commented):
This is mainly to help vetters improve data quality? This item isn't intended for use by ICU?

macchiati:
It is intended also to be useful for cases where implementations want to
allow people to change numeric separators in the existing patterns. That
is, we want to transition to using a symbol in place of the numeric
date/time separators, so that that can happen smoothing. There are a few
ways to do that, but the first step is to gather and verify the data.

sffc:
The use case of users switching their datetime separators is one that needs a lot more design; it's not clear to me that a single coarse data item is sufficient for that.

macchiati:
The single coarse option is a common one; eg Apple's UI. While we could look at something fancier later on, the perfect should not be the enemy of the good.

sffc:
Date separators go hand-in-hand with date field ordering. To me `/` means MDY, `.` means DMY, and `-` means YMD. I'm not the only person who has these expectations. Other people may have other expectatoins.

We've had this discussion in DWG before and I thought the course of action was "more design needed".

macchiati:
This ticket **was** approved:

"Discussed in design meeting 2025-07-21, approved as the following in the CLDR TC meeting 2025-07-23"

Typically for a given locale, the date order for numeric dates is consistent within a calendar. The data was derived from the patterns that are already in CLDR. 

> / means MDY, . means DMY, and - means YMD.

That is not relevant to this case, which is the primary format within each locale. We are **not** trying to handle multiple orders with different separators marking the different orders. Moreover, at least in some countries, there is pretty free variation between use of / and -, without regard to order: I have seen both "meet on 2/14" and "meet on 2-14" referring to Feb 14.

The data was derived from the current CLDR data. It does have multiple purposes — but either one is sufficient for doing the ticket.
1. Get the vetters to use consistent terminology, by alerting them to when items deviate. It's up to them to decide that the default value is wrong, or the particular inconsistent usage is ok.
2. Allow clients of CLDR to know what the predominant separator is, to allow users to change the separator. This is being hacked by implementations right now; this ticket is a step towards making that cleaner.

sffc:
ok.

macchiati:
I need a reviewer

common/dtd/ldml.dtd
```diff
@@ -1516,14 +1516,32 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
 <!ATTLIST dateTimeFormats draft (approved | contributed | provisional | unconfirmed | true | false) #IMPLIED >
     <!--@METADATA-->
     <!--@DEPRECATED-->
 <!ATTLIST dateTimeFormats validSubLocales CDATA #IMPLIED >
     <!--@VALUE-->
     <!--@DEPRECATED-->
 
+<!ELEMENT numericSeparators ( alias | ( default*, numericDateSeparator*, numericTimeSeparator*, special*)) >
+
+<!ELEMENT numericDateSeparator ( #PCDATA ) >
```
AEApple, line 1529: is there a convention about having a space between the parenthesis sign and the text, or that i should know about? Currently line 1527 looks inconsistent as compared to line 1529. I would like to know if there is a standard practice that I should follow for DTDs?

common/dtd/ldml.dtd
```diff
@@ -1516,14 +1516,32 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+<!ATTLIST numericDateSeparator alt NMTOKENS #IMPLIED >
+    <!--@MATCH:literal/variant-->
+<!ATTLIST numericDateSeparator draft (approved | contributed | provisional | unconfirmed) #IMPLIED >
+    <!--@METADATA-->
+<!ATTLIST numericDateSeparator references CDATA #IMPLIED >
```
AEApple, line 1534: Per [your documentation](https://cldr.unicode.org/development/updating-dtds#adding-structure-elements-attributes-attribute-values), it says we need to ask the TC to approve using CDATA?

Why do we want to use it here?

AEApple (commented):
We're not doing the interval separators as part of this right?

common/dtd/ldml.dtd
macchiati: right. There is a separate ticket for the interval formats, and I think its more appropriate there.

common/dtd/ldml.dtd
macchiati: That's the element value (CDATA means freeform). I'll have to check the documentation, because it is ok for element values, but should take TC approval for attribute values.

macchiati: That is, CDATA lets the user enter any punctuation mark. However, that will raise an error in the Survey tool if they add, say, "a".

common/dtd/ldml.dtd
macchiati: Sorry for the confusion; that response was for a different comment.

We don't need to worry about the formats for the DTDs. That is normalized by a tool during the release phase, (

common/dtd/ldml.dtd
macchiati: Ah, the references are always CDATA; that's just using the same format as other references.

AEApple: ah, it says it in the 'attributes' section of "Adding structure (elements, attributes, attribute-values)" but the elements section right above it says '2. read more details below.' which was ambiguous to me. The element section doesn't say whether CDATA is acceptable or not? I'll propose adding a link to the see below based on this discussion.

macchiati: sounds good, thanks. Can you approve?

AEApple (approved):
Thank you for answering all my questions. I spot checked the data updates and the provisional data LGTM!

PR #5583, opened 2026-04-15
CLDR-18541 Add missing skeletons

macchiati:
CLDR-18541

ALLOW_MANY_COMMITS=true


srl295 (approved):

AEApple:
Should we update [date-time-symbols.md](https://cldr.unicode.org/translation/date-time/date-time-symbols) and translation.md?

Right now it says: "Don’t change v to z or vice versa; just leave either z or v as in English."; however, it doesn't explain what a capital Z means from what I've seen?

macchiati:
Yes, adding more explanation would be good. Probably a different ticket.

PR #5584, opened 2026-04-15
CLDR-10480 Singapore Population Numbers

conradarcturus:
CLDR-10480

As requested by the task, this adds a lot more population numbers for Singapore. When possible, the data comes from the official Singapore census, otherwise it comes from J LeClerc's website with numerous trustworthy datapoints.

I had to make a few overrides to the Likely Subtags to avoid making incorrect changes to them since our population data for PR China is sparse. Additionally, I added annotations for a number of Chinese languages for their default script -- they all have the biggest worldly population in PR China so the default script should be Hans.

ALLOW_MANY_COMMITS=true


macchiati (approved):
This is easier for us to read (and compare) if the items are or

srl295:
> This is easier for us to read (and compare) if the items are or

I think something was truncated…

PR #5585, opened 2026-04-15
CLDR-15993 Add translator notes

robertbastian:
CLDR-15993

ALLOW_MANY_COMMITS=true


AEApple (approved):
TY!

PR #5586, opened 2026-04-15
CLDR-19386 Remove sr-Cyrl-ME collator fallback

robertbastian:
CLDR-19386

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/downloads/cldr-48.md
+++ b/docs/site/downloads/cldr-48.md
@@ -399,6 +399,7 @@ and feedback that the general practice for exact dates is to use Gregorian for p
 - Segmentation format changed, spec updated in CLDR 48.1 to remove obsolete note. [CLDR-19140]
 - There are a few collisions in the emoji short names in ba, fr, rm, vi. Data fixed in CLDR 48.2. [CLDR-19189]
 - The Lazuli locale (lzz.xml) uses the wrong language tag (laz instead of lzz) to identify the name of the language in the language itself. Data will be corrected in CLDR 49. [CLDR-19316]
+- The collation-specific fallback from `sr-Cyrl-ME` to `sr-ME` is incorrect
 
 ## Acknowledgments
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
@@ -189,8 +189,7 @@ public class ConvertLanguageData {
                     } else {
                         // These exceptions are OK, because these locales by default use the
                         // non-default script
-                        Set<String> OKExceptions =
-                                ImmutableSet.of("sr_Cyrl_ME", "zh_Hans_HK", "zh_Hans_MO");
+                        Set<String> OKExceptions = ImmutableSet.of("zh_Hans_HK", "zh_Hans_MO");
                         if (OKExceptions.contains(locale)) {
                             continue;
                         }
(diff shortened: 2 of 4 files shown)


srl295 (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
```diff
@@ -4,10 +4,7 @@
 import com.google.common.collect.ImmutableSet;
 import com.google.common.math.DoubleMath;
 import com.ibm.icu.impl.Relation;
-import com.ibm.icu.impl.Row;
 import com.ibm.icu.impl.Row.R2;
-import com.ibm.icu.text.NumberFormat;
-import com.ibm.icu.text.UTF16;
```
srl295, line 10: why were these removed?

srl295:
lgtm once building

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
robertbastian: ugh my IDE keeps deleting imports

srl295:
> TestParentLocales: org.unicode.cldr.util.TestSupplementalDataInfo
> collations/sr_Cyrl_ME ==> expected: <sr_ME> but was: <sr_Cyrl>

got to love those specific tests..

srl295 (dismissed):

srl295:
@robertbastian curious, why all the force pushes?  >1 commit is OK here in CLDR. It does make it harder to see what's changing.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestSupplementalDataInfo.java
```diff
@@ -11,7 +11,6 @@
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.junit.jupiter.api.Test;
```
srl295, line 14: I think this is breaking the build

tools/cldr-code/src/main/java/org/unicode/cldr/tool/ConvertLanguageData.java
macchiati: You should run spotless before commits, to catch problems like this.

macchiati:
We have a link to the ticket in the description. That appears to have been deleted. Best to leave the description template alone, except fixing _____ to be the ticket number, and of course adding the real description

srl295:
> We have a link to the ticket in the description. That appears to have been deleted. Best to leave the description template alone, except fixing _____ to be the ticket number, and of course adding the real description

I addded the link

srl295 (approved):

docs/site/downloads/cldr-48.md
```diff
@@ -399,6 +399,7 @@ and feedback that the general practice for exact dates is to use Gregorian for p
 - Segmentation format changed, spec updated in CLDR 48.1 to remove obsolete note. [CLDR-19140]
 - There are a few collisions in the emoji short names in ba, fr, rm, vi. Data fixed in CLDR 48.2. [CLDR-19189]
 - The Lazuli locale (lzz.xml) uses the wrong language tag (laz instead of lzz) to identify the name of the language in the language itself. Data will be corrected in CLDR 49. [CLDR-19316]
+- The collation-specific fallback from `sr-Cyrl-ME` to `sr-ME` is incorrect
```
AEApple, line 402: In the future can you add the issue number when adding a known issue so that people can find the fix to cherrypick if they need to in the future?

PR #4577, opened 2025-04-07
CLDR-16665 key/type core values

macchiati:
CLDR-16665

Add core values for key/type

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #5591, opened 2026-04-15
CLDR-19252 move core types to comprehensive if the key has a digit

srl295:
example `//ldml/localeDisplayNames/types/type[@key="d0"][@type="hwidth"][@scope="core"]`

CLDR-19252

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
@@ -735,6 +735,10 @@ public class TestCoverageLevel extends TestFmwkPlus {
                 }
             } else if (xpp.contains("posix")) {
                 continue;
+            } else if (path.equals(
+                            "//ldml/localeDisplayNames/types/type[@key=\"t0\"][@type=\"und\"][@scope=\"core\"]")
+                    && logKnownIssue("CLDR-19252", "comprehensive path: " + path)) {
+                continue;
             }
 
             errln("Comprehensive & no exception for path =>\t" + path);
--- a/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestExtraPaths.java
+++ b/tools/cldr-code/src/test/java/org/unicode/cldr/util/TestExtraPaths.java
@@ -194,7 +194,8 @@ public class TestExtraPaths {
             final String path = i.next();
             final Level l = cov.getLevel(path);
             if (path.endsWith("[@scope=\"core\"]")) {
-                assertEquals(Level.MODERN, l, () -> localeId + ":" + path);
+                // should be modern or lower. Some were already lower.
+                assertTrue(Level.MODERN.isAtLeast(l), () -> l + "=" + localeId + ":" + path);
             }
         }
     }
@@ -247,6 +248,13 @@ public class TestExtraPaths {
                             Level.COMPREHENSIVE,
                             l,
                             () -> localeId + ":" + path + " - expect comprehensive");
+                }
+            } else {
+                if (l48 != null && l48 != Level.COMPREHENSIVE) {
+                    // present in v48
+                    assertEquals(l48, l, () -> localeId + ": vs v48 " + path);
+                } else {
+                    assertEquals(Level.MODERN, l, () -> localeId + ": (not in v48)" + path);
                 }
             }
         }
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -287,11 +287,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 
 		<coverageLevel	value="basic"	 	match="localeDisplayNames/measurementSystemNames/measurementSystemName[@type='(metric|UK|US)']"/>
 
-		<!-- all types with a digit in the key need to be comprehensive-->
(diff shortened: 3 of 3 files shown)


macchiati:
That works for now. 

However, I would rather have them be not visible at all; we do not vet the comprehensive data, which means that garbage can leak in.

srl295:
> That works for now.
> 
> However, I would rather have them be not visible at all; we do not vet the comprehensive data, which means that garbage can leak in.

OK, i will update it.

the original wording said "should be level=comprehensive".

srl295:
@macchiati the following was in modern coverage per your #4577 (per `<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='t0'][@type='%t0Types80'][@scope='%A']"/>` 

`//ldml/localeDisplayNames/types/type[@key="t0"][@type="und"][@scope="core"]`

.. but fails with "Comprehensive & no exception for path" in TestCoverageLevel.TestCoverageCompleteness — what is that test testing for?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
```diff
@@ -735,6 +735,10 @@ public void TestCoverageCompleteness() {
                 }
             } else if (xpp.contains("posix")) {
                 continue;
+            } else if (path.equals(
+                            "//ldml/localeDisplayNames/types/type[@key=\"t0\"][@type=\"und\"][@scope=\"core\"]")
```
srl295, line 739: could not repro this locally

srl295:
Ready to go.  
CLDR-19253 (removing other core items) will be separate.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestExtraPaths.java
```diff
@@ -194,7 +194,8 @@ public void testTypeCoverageCore(final String localeId) {
             final String path = i.next();
             final Level l = cov.getLevel(path);
             if (path.endsWith("[@scope=\"core\"]")) {
-                assertEquals(Level.MODERN, l, () -> localeId + ":" + path);
+                // should be modern or lower. Some were already lower.
+                assertTrue(Level.MODERN.isAtLeast(l), () -> l + "=" + localeId + ":" + path);
```
btangmu, line 198: The word "lower" is confusing. By "modern or lower" I believe you mean "modern or comprehensive", but numerically for the enum, MODERATE < MODERN < COMPREHENSIVE. So I think the code is correct but the comment could be confusing.

btangmu (approved):

macchiati:
> @macchiati the following was in modern coverage per your #4577 (per `<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='t0'][@type='%t0Types80'][@scope='%A']"/>`
> 
> `//ldml/localeDisplayNames/types/type[@key="t0"][@type="und"][@scope="core"]`
> 
> .. but fails with "Comprehensive & no exception for path" in TestCoverageLevel.TestCoverageCompleteness — what is that test testing for?

As I recall; There is a test to make sure that we don't miss anything inadvertently. So if something is comprehensive deliberately. It needs to be in an exception list someplace

AEApple:
> > That works for now.
> > However, I would rather have them be not visible at all; we do not vet the comprehensive data, which means that garbage can leak in.
> 
> OK, i will update it.
> 
> the original wording said "should be level=comprehensive".

I thought the discussion was we were going to make them hidden? I added a comment to the wrong PR by accident. It said hidden [at the top of the comment from the meeting discussion](https://unicode-org.atlassian.net/browse/CLDR-19334?focusedCommentId=205389) I hadn't edited the contents of the comment copied from the walkthrough doc.

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestExtraPaths.java
macchiati: The scope=core should be modern or lower (that is: Core..Modern), except if for a key/type where the key has a digit.

I'm not sure this does it. 

BTW, I think it is much clearer to say the following (the set exists already)

assertTrue(Level.CORE_TO_MODERN.contains(l))

And if we need other constant sets we can add them.

macchiati (approved):
Approving for now, just so we can get a better hold on the number of items for people to translate. I agree that we should look at HIDE or not generate in ExtraPaths, but that can come in a subsequent PR.

srl295:
> > @macchiati the following was in modern coverage per your #4577 (per `<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='t0'][@type='%t0Types80'][@scope='%A']"/>`
> > `//ldml/localeDisplayNames/types/type[@key="t0"][@type="und"][@scope="core"]`
> > .. but fails with "Comprehensive & no exception for path" in TestCoverageLevel.TestCoverageCompleteness — what is that test testing for?
> 
> As I recall; There is a test to make sure that we don't miss anything inadvertently. So if something is comprehensive deliberately. It needs to be in an exception list someplace

Do you think this specific path should be in coverage, comprehensive, or hidden? (It was in v48's coverage.xml)

tools/cldr-code/src/test/java/org/unicode/cldr/util/TestExtraPaths.java
srl295: I mean Core…Modern which isAtLeast does.

PR #5593, opened 2026-04-16
CLDR-14121 Add UTC alt for gmtFormat

robertbastian:
CLDR-14121

ALLOW_MANY_COMMITS=true


common/main/en.xml
```diff
@@ -4207,12 +4207,9 @@ annotations.
 			</field>
 		</fields>
 		<timeZoneNames>
-			<hourFormat>+HH:mm;-HH:mm</hourFormat>
-			<gmtFormat>GMT{0}</gmtFormat>
```
srl295, line 4211: An item being repeated as in the parent is sometimes used to confirm that a locale specific value has been entered.  I'm not sure why they are removed here. If anything they could be updated to this, which is common in many locales.

```xml
			<hourFormat>↑↑↑</hourFormat>
			<gmtFormat>↑↑↑</gmtFormat>
```

common/main/en.xml
```diff
@@ -4207,12 +4207,9 @@ annotations.
...
 			<regionFormat>{0} Time</regionFormat>
 			<regionFormat type="daylight">{0} Daylight Time</regionFormat>
 			<regionFormat type="standard">{0} Standard Time</regionFormat>
-			<fallbackFormat>{1} ({0})</fallbackFormat>
```
srl295, line 4215: Same here

```xml
			<fallbackFormat>↑↑↑</fallbackFormat>
```

common/main/en.xml
macchiati: Yes, although there is a CLDR modify step that replaces the ↑↑↑ to prevent misalignment. (I'll have to grab that ticket.)

robertbastian: I don't think explicit ↑↑↑ are useful for English, as it's not a survey tool locale. 

I removed these because the `gmtUnknownFormat` item was also not here, and I don't want to duplicate data.

macchiati: We'll discover thqt in the tests, though. So we can let this go through if it passes the tests.

macchiati (dismissed):

common/main/en.xml
srl295: > I don't think explicit ↑↑↑ are useful for English, as it's not a survey tool locale.

other tooling reads it, and, SurveyTool English is used even if there's no voting there.

srl295: Perhaps it would be best to not remove unrelated fields in this PR? it seems like unnecessary churn, this pr removed the element, it was added later, etc.

robertbastian:
What does "literal/variant" mean?

```
Error:  (TestPaths.java:634)  Error: main/root.xml/gmtFormat@alt, expected match to: ⟪literal/variant⟫ actual: «utc»
```

srl295:
> What does "literal/variant" mean?
> 
> 
> 
> ```
> 
> Error:  (TestPaths.java:634)  Error: main/root.xml/gmtFormat@alt, expected match to: ⟪literal/variant⟫ actual: «utc»
> 
> ```

The DTD expected `<gmtFormat alt="variant">` but found `<gmtFOrmat alt="utc">`

Look for a `@MATCH` line in the DTD.

robertbastian:
Removing that doesn't seem to do anything.

I also don't know why the example generator is failing for `gmtUnknownFormat[@alt="utc"]`, I've treated it the exact same way as `gmtUnknownFormat`.

srl295:
> Removing that doesn't seem to do anything.

Remove it or fix it,docs at https://cldr.unicode.org/development/dtd_annotations

> I also don't know why the example generator is failing for `gmtUnknownFormat[@alt="utc"]`, I've treated it the exact same way as `gmtUnknownFormat`.

Not sure. I will take a look in a bit.

macchiati:
I only have my phone right now, but what can happen is that any tests to check on particular paths can fail if they're unexpected alternates. So sometimes you have to adjust the matching.

macchiati:
The tests are failing. Click on the "cldr-mvn / build (pull_request)Failing after 25m" ... , then View details.

Best I find is to search for "Error summary:". You'll find:

Error summary:
CLDR/TestAttributeValues/TestValid
CLDR/TestExampleGenerator/TestAllPaths
CLDR/TestExampleGenerator/TestForMissing
CLDR/TestPaths/TestNonLdml

You can then search for the individual error messages, like:

```
  TestPaths {
    TestCalendarAvailablePaths (0.027s) Passed
    TestNonLdml
Error:  (TestPaths.java:634)  Error: main/root.xml/gmtFormat@alt, expected match to: ⟪literal/variant⟫ actual: «utc»
```

In that case, the problem is that you have fixes for some items, but not all:
```
<!ELEMENT gmtFormat ( #PCDATA ) >
<!ATTLIST gmtFormat alt NMTOKENS #IMPLIED >
<!--@MATCH:literal/variant-->
```

Then locally you can work on getting those tests to pass. For example, run TestAttributeValues on your system, and if you want, pass in TestValid as an argument to just focus on that.

common/dtd/ldml.dtd
```diff
@@ -1813,17 +1789,15 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@METADATA-->
 
 <!ELEMENT gmtZeroFormat ( #PCDATA ) >
-<!ATTLIST gmtZeroFormat alt NMTOKENS #IMPLIED >
-    <!--@MATCH:literal/variant-->
+<!ATTLIST gmtZeroFormat alt (utc) #IMPLIED >
```
macchiati, line 1792: We prefer the @MATCH for items that can grow over time. The explicit values (like your "(utc)") are not as flexible.  So
``` suggestion
<!ATTLIST gmtZeroFormat alt NMTOKENS #IMPLIED >
    <!--@MATCH:literal/variant,utc-->
```
Similar elsewhere.

macchiati (dismissed):

common/dtd/ldml.dtd
macchiati: It appears that you forgot to add the `utc` here also.

```suggestion
<!ELEMENT gmtFormat ( #PCDATA ) >
<!ATTLIST gmtFormat alt NMTOKENS #IMPLIED >
<!--@MATCH:literal/variant, utc-->
```

macchiati:
Still failing the following tests:
```
CLDR/TestAttributeValues/TestValid
CLDR/TestExampleGenerator/TestAllPaths
CLDR/TestExampleGenerator/TestForMissing
CLDR/TestPaths/TestNonLdml
```
Please consult https://cldr.unicode.org/development/updating-dtds. 

Also, https://github.com/unicode-org/cldr/pull/5608 gives an example of the kinds of changes needed. That is more complicated than this case, because it is adding elements instead of just adding an alt variant. But you can see the kinds of changes that are involved.

srl295:
> The tests are failing. Click on the "cldr-mvn / build (pull_request)Failing after 25m" ... , then View details.
> 
> Best I find is to search for "Error summary:". You'll find:

also in the Summary of the job it lists them for you! <https://github.com/unicode-org/cldr/actions/runs/24661368327#summary-72107938694>

image

srl295 (approved):

docs/site/translation/time-zones-and-city-names.md
```diff
@@ -32,8 +32,10 @@ The following special patterns are used in formatting timezones.
...
 |---|---|---|---|---|
-| gmtFormat | GMT{0}<br /> *or*<br /> GMT{ HOURS_FROM_GMT } | GMT -2:00 | **GMT Pattern.** Modify this field if the format for GMT time uses different letters, such as  HUA+0200  for  GMT+02:00 , or if the letters GMT occur after the time. Make sure you include the  {0} ; that is where the actual time value will go! |  |
+| gmtFormat | GMT{0}<br /> *or*<br /> GMT{ HOURS_FROM_UTC } | GMT-2:00 | **Offset Pattern.** Modify this field if the format for time zone offsets uses different letters, such as  HUA+02:00 or UTC+02:00 instead of GMT+02:00 , or if the letters occur after the time. Make sure you include the  {0} ; that is where the actual time value will go! |  |
+| gmtFormat alt="utc" | UTC{0}<br /> *or*<br /> UTC{ HOURS_FROM_UTC } | UTC-2:00 | **UTC Pattern.** Similar to gmtFormat, but this is *required* to use the localized abbreviation for "Coordinated Universal Time". Also set this if gmtFormat already uses the localized abbreviation for "Coordinated Universal Time". Note that this is usually "UTC", which was explicitly chosen as the international designation for "Coordinated Universal Time".  |  |
```
macchiati, line 36:
```suggestion
| gmtFormat alt="utc" | UTC{0}<br /> *or*<br /> UTC{ HOURS_FROM_UTC } | UTC-2:00 | **UTC Pattern.** Similar to gmtFormat, but this is *required* to use the localized abbreviation for "Coordinated Universal Time". For Latin-script languages, this typically uses the abbreviation "UTC". Set this even if gmtFormat (without the 'alt') already uses the localized abbreviation for "Coordinated Universal Time".  |  |
```

macchiati (approved):
I have one further comment to restate what is in docs/site/translation/time-zones-and-city-names.md. However, in the interests of time I'm going to merge; that wording can be fixed in a later PR.

PR #5594, opened 2026-04-16
CLDR-18913 Revise locale setting for Add User

btangmu:
-If organization for new user covers all locales, filter to exclude readonly/scratch; new method cldrLocales.getAllVotable

-Fix call to makeApiUrl for locales/normalize; it was missing module name cldrAjax

-Make the All Locales option available for manager and stronger, not for vetter and weaker; it was the other way around

-In AddUser.java, getNewUserLocales, disallow default-content, and do not substitute LocaleNames.MUL for a set that would otherwise be empty

CLDR-18913

ALLOW_MANY_COMMITS=true


tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/AddUser.java
```diff
@@ -126,11 +125,7 @@ private AddUserResponse doAddNewUser(AddUserRequest request, CookieSession sessi
     }
 
     private String getNewUserLocales(String requestLocales) {
-        requestLocales = LocaleNormalizer.normalizeQuietly(requestLocales);
-        if (requestLocales.isEmpty()) {
-            return LocaleNames.MUL;
```
srl295, line 131: 🎉

srl295 (approved):

PR #5596, opened 2026-04-16
CLDR-19352 remove core key/types for now as they aren't in English

srl295:
CLDR-19352

Remove most key/types which are core from extraPaths

These aren't in English now, so should at least be in English before re-adding.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
@@ -227,6 +227,29 @@ public class ExtraPaths {
                         "timezone" // ldml/dates/timeZoneNames
                         );
 
+        // skip these core 'keys' entirely from extraPaths
+        private static final Set<String> skipCoreKeys =
+                ImmutableSet.of(
+                        "fw",
+                        "numbers",
+                        "t0",
+                        "colAlternate",
+                        "colBackwards",
+                        "colCaseFirst",
+                        "colCaseLevel",
+                        "colNormalization",
+                        "colNumeric",
+                        "colReorder",
+                        "colStrength",
+                        "d0",
+                        "k0",
+                        "h0",
+                        "i0",
+                        "m0",
+                        "mu",
+                        "numbers",
+                        "s0");
+
         /** add BCP47 keys */
         private void addBcp47Keys() {
             final Relation<String, String> bcp47Keys = supplementalData.getBcp47Keys();
@@ -346,8 +369,10 @@ public class ExtraPaths {
                     final String typeKeyTypePath =
                             typeKeyPath + String.format("[@type=\"%s\"]", type);
                     pathsTemp.add(typeKeyTypePath);
-                    // add all of these with [@scope="core"]
-                    pathsTemp.add(typeKeyTypePath + SCOPE_CORE);
+                    // add some of these with [@scope="core"]
+                    if (!skipCoreKeys.contains(k)) {
+                        pathsTemp.add(typeKeyTypePath + SCOPE_CORE);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


srl295:
@macchiati Mark I thought you had said you had a list of suggested English for the core types? Otherwise, what would you think of this to take them out of extra paths for now?

macchiati:
It's in https://docs.google.com/spreadsheets/d/1mHXc0rZyBUAU3-55ulyL7eLogSlYNwZLbQYSv0YNyek/edit?gid=0#gid=0 

See tabs: English Values and Number (nu) values

Just made some tweaks and fixed the non-script code names.

macchiati:
And don't remove them... We need them for vetters to reference

srl295:
> And don't remove them... We need them for vetters to reference

Removing *from extra paths *

srl295:
> It's in https://docs.google.com/spreadsheets/d/1mHXc0rZyBUAU3-55ulyL7eLogSlYNwZLbQYSv0YNyek/edit?gid=0#gid=0 
> 
> 
> 
> See tabs: English Values and Number (nu) values
> 
> 
> 
> Just made some tweaks and fixed the non-script code names.

Ok. I haven't seen this before. Will do

macchiati:
Huh, suddenly https://cldr-smoke.unicode.org/cldr-apps/v#/de// is missing all of the key/type core lines in the Dashboard.  Was that from a different PR?

AEApple:
> Huh, suddenly https://cldr-smoke.unicode.org/cldr-apps/v#/de// is missing all of the key/type core lines in the Dashboard. Was that from a different PR?

That seems like a blocking issue for starting shakedown?

srl295:
> Huh, suddenly https://cldr-smoke.unicode.org/cldr-apps/v#/de// is missing all of the key/type core lines in the Dashboard.  Was that from a different PR?

It's not from this unmerged draft PR, but i don't mind discussing here..

I see them in dashboard and in the list, do you see something different?

image

srl295:
> > Huh, suddenly https://cldr-smoke.unicode.org/cldr-apps/v#/de// is missing all of the key/type core lines in the Dashboard. Was that from a different PR?
> 
> That seems like a blocking issue for starting shakedown?

I wasn't able to repro …

macchiati:
I got the same results when refreshing the page; only 18 Missing. But then I logged out, and logged back in. At that point it was back to missing 218.

So, cosmic rays? Anyway, we can just be on the lookout for recurrence

srl295:
> I got the same results when refreshing the page; only 18 Missing. But then I logged out, and logged back in. At that point it was back to missing 218.
> 
> So, cosmic rays? Anyway, we can just be on the lookout for recurrence

could the coverage level has been changed?

srl295:
> I got the same results when refreshing the page; only 18 Missing. But then I logged out, and logged back in. At that point it was back to missing 218.
> 
> So, cosmic rays? Anyway, we can just be on the lookout for recurrence

I get 18 missing if i switch to Moderate, right now, on https://cldr-smoke.unicode.org/cldr-apps/v#/de/Keys/

Could you have created a smoketest account for an org that defaults `de` to moderate?  for example

```
wikimedia ; de ; moderate ;
```

macchiati:
Ah, that's it. Because smoketest is mainly for our usage, we should make it
predictable, with the default be the most useful one, a TC org with *
locales. Someone who needs something out of the ordinary of course can
adjust.

srl295:
> Ah, that's it. Because smoketest is mainly for our usage, we should make it predictable, with the default be the most useful one, a TC org with * locales. Someone who needs something out of the ordinary of course can 

it tells you what the org and locales it will create are

it also predates when locales.txt actually affected vote privileges

anyway, we could also easily just allow:  <https://cldr-smoke.unicode.org/cldr-apps/v?vap=≈≈≈≈≈≈≈≈≈&org=google&level=tc#createAndLogin///> then clicking that will always create a google tc.   If you want to file a ticket for that its a couple lines of code and then we could have several links for creating different types of users, tc org by default etc

srl295:
OK, is this PR itself an improvement? Do not generate (in ExtraPaths) any core types not already in English, but lets us bring them back selectively.

srl295:
folded into #5601

PR #5599, opened 2026-04-17
CLDR-19258 Update datetime test data after icu4j version bump

sffc:
~Follow-up to #5598~

~The previous commit (9ffad5d0b7) disabled stock patterns by default in DateTimeFormats. This change updates the generated test data to reflect the new behavior, where expected strings are synthesized from skeletons without falling back to stock patterns.~

EDIT: The culprit is https://github.com/unicode-org/cldr/commit/09e74533a67c3a316f9ffb4d78f56f2b49d0b972. The datetime test data was not updated when icu4j was updated.

CLDR-19258

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/testData/datetime/datetime.json
+++ b/common/testData/datetime/datetime.json
@@ -201,7 +201,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2000-01-01T00:00Z[Etc/GMT]",
-    "expected": "Saturday, January 1, 2000"
+    "expected": "2000 January 1, Saturday"
   },
   {
     "semanticSkeleton": "YMDE",
@@ -210,7 +210,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2024-07-01T08:50:07Z[Etc/GMT]",
-    "expected": "Monday, July 1, 2024"
+    "expected": "2024 July 1, Monday"
   },
   {
     "semanticSkeleton": "YMDE",
@@ -219,7 +219,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2014-07-15T12:00Z[Etc/GMT]",
-    "expected": "Tuesday, July 15, 2014"
+    "expected": "2014 July 15, Tuesday"
   },
   {
     "semanticSkeleton": "YMDE",
@@ -289,7 +289,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2000-01-01T00:00Z[Etc/GMT]",
-    "expected": "Saturday, January 1, 2000 AD"
+    "expected": "AD 2000 January 1, Saturday"
   },
   {
     "semanticSkeleton": "YMDE",
@@ -299,7 +299,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2024-07-01T08:50:07Z[Etc/GMT]",
-    "expected": "Monday, July 1, 2024 AD"
+    "expected": "AD 2024 July 1, Monday"
   },
   {
     "semanticSkeleton": "YMDE",
@@ -309,7 +309,7 @@
     "calendar": "gregorian",
     "locale": "en",
     "input": "2014-07-15T12:00Z[Etc/GMT]",
-    "expected": "Tuesday, July 15, 2014 AD"
+    "expected": "AD 2014 July 15, Tuesday"
   },
   {
     "semanticSkeleton": "Z",


sffc:
Actually these diffs are pre-existing. I'm looking for the culprit commit.

sffc:
The culprit is 09e74533a6. The datetime test data was not updated when icu4j was updated.

macchiati:
ugg

sffc:
this is a bit messy; I thought we verified cldr testdata consistency with the tool output in CI but I guess not?

macchiati (approved):

macchiati:
Right now, I don't think we test the test data each time. Any time we change data, dtd, or code, it is possible for the test data to change. In the past, we've just regenerated the test data at major milestones. But if we want to make sure the build is always clean, we could have implement one of the two options below.

- a test for each test data file, OR
- a test that regenerates the test data in a temp folder using GenerateTestData, and compares it to the contents of the common/testData folder. It fails if they are not identical.

In either case, it is up to the PR writer to rerun GenerateTestData (if the PR is blocked) and include the changed files. 

The second of these has the great advantage of being more turn-key and robust. I timed it on my laptop, and it only takes 45sec to run. We could also run it in parallel with the unit tests. We'd need a ticket for this.

cc: @AEApple @srl295 @robertbastian

macchiati:
Shane, I reviewed this too quickly (I was focused on the process). The following is wrong:
```
    "semanticSkeleton": "YMDE",
    "semanticSkeletonLength": "long",
    "classicalSkeleton": "yMMMMdEEEE",
    "calendar": "gregorian",
    "locale": "en",
    "input": "2000-01-01T00:00Z[Etc/GMT]",
-    "expected": "Saturday, January 1, 2000"
+    "expected": "2000 January 1, Saturday"
```

The English gregorian value should **not** be "2000 January 1, Saturday"; this PR needs to be reverted.

srl295:
> Right now, I don't think we test the test data each time. Any time we change data, dtd, or code, it is possible for the test data to change. In the past, we've just regenerated the test data at major milestones. But if we want to make sure the build is always clean, we could have implement one of the two options below.
> 
> 
> 
> - a test for each test data file, OR
> 
> - a test that regenerates the test data in a temp folder using GenerateTestData, and compares it to the contents of the common/testData folder. It fails if they are not identical.
> 
> 
> 
> In either case, it is up to the PR writer to rerun GenerateTestData (if the PR is blocked) and include the changed files. 
> 
> 
> 
> The second of these has the great advantage of being more turn-key and robust. I timed it on my laptop, and it only takes 45sec to run. We could also run it in parallel with the unit tests. We'd need a ticket for this.
> 

I'd recommend running it like (or even with) the automatic CLDRModify we already run. 

I.e.. it could tag onto the CLDRModify PRs or be a new series of PRs 

Then the original author wouldn't need to do anything. If they forgot it would be fixed in a later pr. 

> 
> 
> cc: @AEApple @srl295 @robertbastian

sffc:
I already made https://github.com/unicode-org/cldr/pull/5605 to add a GitHub Action for this.

The test data should reflect the source of truth in the repo, even if the source of truth is bad, because otherwise we end up with an inconsistent mess of test cases.

macchiati:
> Shane, I reviewed this too quickly (I was focused on the process). The following is wrong:
> 
> ```
>     "semanticSkeleton": "YMDE",
>     "semanticSkeletonLength": "long",
>     "classicalSkeleton": "yMMMMdEEEE",
>     "calendar": "gregorian",
>     "locale": "en",
>     "input": "2000-01-01T00:00Z[Etc/GMT]",
> -    "expected": "Saturday, January 1, 2000"
> +    "expected": "2000 January 1, Saturday"
> ```
> 
> The English gregorian value should **not** be "2000 January 1, Saturday"; this PR needs to be reverted.

@sffc , I don't want this to get lost — the test data is very wrong as it stands for English. 

Was there a reason you didn't revert it?

sffc:
@macchiati, the test data reflects the behavior on the main branch. I double-checked and the skeleton `yMMMMdEEEE` is matching the following data in root.xml

```xml
<dateFormatItem id="yMMMMEd">y MMMM d, E</dateFormatItem>
<dateFormatItem id="yMMMMEddd">y MMMM ddd, E</dateFormatItem>
```

There is no override of that skeleton in en.xml, so we match the one in root.xml

sffc:
Note: The PR that added that skeleton is #5583, not the ICU4J version bump like my AI agent supposedly found.

sffc:
I opened https://unicode-org.atlassian.net/browse/CLDR-19399 to track this.

PR #5600, opened 2026-04-17
CLDR-7408 Ordinal dates Part2

macchiati:
CLDR-7408

Continuing work started in https://github.com/unicode-org/cldr/pull/5590

- add date patterns: instances of MMME+d were duplicated with ddd in the id and pattern (root and English)
- fixed problems Rich noted: only format, only abbreviated
- adjusted for coverage problems
- for English, added comma for the fallback dow format. 
- added many aliases in root, so the calendars inherit dayOfMonths
- documented in Info Hub (I expected it will take further enhancement, I'd like to do in subsequent PR).
- added/refined ExampleGenerator for the new paths.
    - For the samples for date ordinals, confined to numbers from 1 to 32
    - Added 2 hacks to support "ddd" in examples until ICU supports it (`formatWithOrdinalHack`). @sffc, along with your 'cleanroom' DatetimePatternGenerator, we might want to do something similar for formatting, so that we have it for both examples and for generating test data.
    - Created a CldrPathUtilities.java with utilities for dealing with paths (right now we have these scattered over different files)
 - Fixed a couple of tests

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -1613,7 +1613,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT dateFormatItem ( #PCDATA ) >
 <!ATTLIST dateFormatItem id CDATA #REQUIRED >
-    <!--@MATCH:regex/((E|EEEE)?(H|HH|h|hh|Bh)(m|ms|mm|mmss)?(v|vvvv)?)|(ms|mmss)|((G|GGGGG)?(y|yy|yyyy)((M{1,4}((E|EEEE|cccc)?(d|dd))?)|(w|Q|QQQ|QQQQ))?)|(U(M|MMM)d?)|(M{1,4}(((E|EEEE|cccc)?(d|dd))|W)?)|((E|EEEE)?d)|(E|EEEE)-->
+    <!--@MATCH:regex/((E|EEEE)?(H|HH|h|hh|Bh)(m|ms|mm|mmss)?(v|vvvv)?)|(ms|mmss)|((G|GGGGG)?(y|yy|yyyy)((M{1,4}((E|EEEE|cccc)?(d{1,3}))?)|(w|Q|QQQ|QQQQ))?)|(U(M|MMM)d?)|(M{1,4}(((E|EEEE|cccc)?(d{1,3}))|W)?)|((E|EEEE)?d)|(E|EEEE)-->
 <!ATTLIST dateFormatItem count (zero | one | two | few | many | other) #IMPLIED >
 <!ATTLIST dateFormatItem alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/ascii, variant-->
--- a/common/dtd/ldml.dtd
+++ b/common/dtd/ldml.dtd
@@ -1006,12 +1006,12 @@ CLDR data files are interpreted according to the LDML specification (http://unic
     <!--@MATCH:literal/variant-->
 
 <!ELEMENT dayOfMonthContext ( alias | ( default*, dayOfMonthWidth*, special* ) ) >
-<!ATTLIST dayOfMonthContext type (format | stand-alone) #REQUIRED >
+<!ATTLIST dayOfMonthContext type (format) #REQUIRED >
 <!ATTLIST dayOfMonthContext alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
 
 <!ELEMENT dayOfMonthWidth ( alias | ( dayOfMonth*, special* ) ) >
-<!ATTLIST dayOfMonthWidth type (abbreviated | narrow | short | wide) #REQUIRED >
+<!ATTLIST dayOfMonthWidth type (abbreviated | wide) #REQUIRED >
 <!ATTLIST dayOfMonthWidth alt NMTOKENS #IMPLIED >
     <!--@MATCH:literal/variant-->
 
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -95,6 +95,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				<months>
 					<alias source="locale" path="../../calendar[@type='gregorian']/months"/>
 				</months>
+				<dayOfMonths>
+					<alias source="locale" path="../../calendar[@type='gregorian']/dayOfMonths"/>
+				</dayOfMonths>
 				<days>
 					<alias source="locale" path="../../calendar[@type='gregorian']/days"/>
 				</days>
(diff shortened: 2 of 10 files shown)


macchiati:
Just squashed a bad commit message; tests should all pass. I'd like to get this committed by tomorrow if it looks ok. @richgillam , I think I fixed the items you noted, but please take a look.

macchiati:
I'd like to get a review if possible this morning. I can follow up with further refinements in a subsequent PR.

common/dtd/ldml.dtd
```diff
@@ -1613,7 +1613,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!ELEMENT dateFormatItem ( #PCDATA ) >
 <!ATTLIST dateFormatItem id CDATA #REQUIRED >
-    <!--@MATCH:regex/((E|EEEE)?(H|HH|h|hh|Bh)(m|ms|mm|mmss)?(v|vvvv)?)|(ms|mmss)|((G|GGGGG)?(y|yy|yyyy)((M{1,4}((E|EEEE|cccc)?(d|dd))?)|(w|Q|QQQ|QQQQ))?)|(U(M|MMM)d?)|(M{1,4}(((E|EEEE|cccc)?(d|dd))|W)?)|((E|EEEE)?d)|(E|EEEE)-->
+    <!--@MATCH:regex/((E|EEEE)?(H|HH|h|hh|Bh)(m|ms|mm|mmss)?(v|vvvv)?)|(ms|mmss)|((G|GGGGG)?(y|yy|yyyy)((M{1,4}((E|EEEE|cccc)?(d{1,3}))?)|(w|Q|QQQ|QQQQ))?)|(U(M|MMM)d?)|(M{1,4}(((E|EEEE|cccc)?(d{1,3}))|W)?)|((E|EEEE)?d)|(E|EEEE)-->
```
richgillam, line 1616: At least to support the Hindu calendatrs, I think we're going to want `d` to go to 4.  On the other hand, I guess we could make that change when we actually have the Hindu-calendar stuff ready to go.

docs/site/translation.md
```diff
@@ -58,7 +58,39 @@ We are reviewing new locale requests for inclusion in CLDR 49. See [how to add a
...
+        * For example, suppose that your locale uses a "er" suffix just on `one` in dates. 
+        * In that case, the `one` form would be {0}er, but all other forms would have just {0}.
+* If your locale _never_ uses ordinals in dates, then:
+    * Set all the dayOfMonth patterns (`one`, `other`, …) to a constant {0}.
+* If your locale _always_ uses ordinals in certain kinds of dates (with non-numeric months), then make sure the patterns where they are used have `ddd` in them (instead of `d` or `dd`).
+    * For example, suppose that a form like "March 3, 2026" is not acceptable, just a form like "March 3rd, 2026".
+In that case, for the code `yMMMMd` you would change its pattern to have `ddd` in it to force the use of ordinals, something like: "MMMM ddd, y"
+
+
+### New emojii
```
srl295, line 93:
```suggestion
### New emoji
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -2818,6 +2862,36 @@ private void handleDateFormatItem(
         }
     }
 
+    private String formatWithOrdinalHack(SimpleDateFormat sdf, String calendar, Date date) {
```
srl295, line 2865: Should have a TODO to fold into the icu service builder

srl295 (approved):
typo and TODOs should have ticket links, but approving for now..

macchiati:
Thanks!

Rich is going to look this over also after his 11 o’clock meeting...

macchiati:
@AEApple We can expect about
- 1-6 items for ordinal dayOfMonth
    -  depends on locale, many (perhaps most) only have 1.
    - See  https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html#Ordinal-Integer)
- ~8 items per modern calendar

common/main/root.xml
```diff
@@ -1232,11 +1254,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						</monthWidth>
 					</monthContext>
 				</months>
-				<dayOfMonths>
```
richgillam, line 1235: So everything in `root` aliases over to `gregorian`, but now `gregorian` doesn't have a `dayOfMonths` resource at all?  That seems wrong.  I'm also surprised not to see an alias from `wide` back to `abbreviated`.  I'd think you'd want something like

```
<dayOfMonths>
    <dayOfMonthContext type="format">
        <dayOfMonthWidth type="abbreviated">
            <dayOfMonth ordinal="other"{0}</dayOfMonth>
        </dayOfMonthWidth>
        <dayOfMonthWidth type="wide">
            <alias source="locale" path="../dayOfMonthWidth[@type='abbreviated']"/>
        </dayOfMonthWidth>
    </dayOfMonthContext>
</dayOfMonths>
```

richgillam (commented):
Generally looks good, but I had one big question.

I didn't really look too deeply at the code sections, since I'm not familiar with those tools, but with a superficial look, it all looked fine.

common/dtd/ldml.dtd
macchiati: Yes, I was leaving that alone until we got to adding the Hindu calendar, in later PR

common/main/root.xml
macchiati: Those are added in ExtraPaths.java

That is a mechanism for adding paths, without adding default values — we do that for case where
- we don't want vetters to blindly approve a root value without thinking about it
- there isn't necessarily a good fallback, or code is expected to take a different approach (like use the numeric form).

NOTE: we can change to have an explicit fallback in root at any later time, like after the vetting for this version is done.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Will file ticket.

We also need to file tickets on ICU for all new features that it should support.

docs/site/translation.md
macchiati: That was just a placeholder for the other changes; we will need to fix all of those (and add missing cases) before Submission.

https://unicode-org.atlassian.net/browse/CLDR-15370

macchiati:
I'm planning to merge these changes in an hour or so; I think the review items have been addressed, but I want to give people a bit of time to respond.

common/main/root.xml
richgillam: Two questions: 1) It seems beneficial to have the inheritance from `wide` to `abbreviated` be there.  You don't want that as one of the defaults?  2) Personally, I'm not a fan of having the default behavior just be in code.  In this case, we can fake just falling back to the numeric form with `ordinal="other">{0}`, and that doesn't seem to hurt anything.  Am I missing something?

macchiati: Right now, it just supports ddd; so we don't have the wide at all. I'm presuming we add that when we get to adding the Hindu calendar. I'm also presuming that we wouldn't need vetters to be involved for the Hindu calendar, because you'd be supplying the data... so there is no rush to get that in right now; we'll just want to get it in by around July 1.

macchiati:
I went ahead and merged, but if anything else comes up I can do another PR.

richgillam:
> I went ahead and merged, but if anything else comes up I can do another PR.

Yeah, fine with me.  Belated ship-it.

PR #5603, opened 2026-04-17
CLDR-19252 site: add PathDescription link to core names

srl295:
CLDR-19252

image

The link goes here: 
> This is the [Core Name](https://srl295-cldr-19252-site-doc-core-names-cldr-next.unicode.workers.dev/translation/displaynames/locale-option-names-key#core-names).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/site/translation/displaynames/locale-option-names-key.md
+++ b/docs/site/translation/displaynames/locale-option-names-key.md
@@ -2,11 +2,11 @@
 title: 'Locale Option Names'
 ---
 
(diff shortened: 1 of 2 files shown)


srl295:
Looks like there were some path description for core before but they weren't working.

macchiati (approved):

srl295:
This only adds the one, but I'll get it in so the link target can be checked. 

I found that there were already entries for core from months ago, but they were broken.

PR #5605, opened 2026-04-18
CLDR-19258 Generate test data in cldr-modify task in GitHub Actions

sffc:
This commit introduces a new 'data-check' job in the Maven workflow that executes 'GenerateTestData' and verifies that there are no differences in the 'common/testData' directory. This ensures that generated test data remains in sync with the tool logic.

Not sure which ticket to use so I used CLDR-19258. I can file a new one if you prefer

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateTestData.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateTestData.java
@@ -9,5 +9,6 @@ public class GenerateTestData {
         GeneratePersonNameTestData.main(args);
         GenerateUnitTestData.main(args);
         GenerateDateTimeTestData.main(args);
+        GenerateRBNFTestData.main(args);
     }
 }
--- a/.github/workflows/cldr-modify.yml
+++ b/.github/workflows/cldr-modify.yml
@@ -69,6 +69,12 @@ jobs:
         env:
           CLDR_DIR: ${{ github.workspace }}
           LANG: en_US.UTF-8
+      - name: Run GenerateTestData
+        run: >
+          mvn -s .github/workflows/mvn-settings.xml -B --file tools/pom.xml -pl cldr-code
+          exec:java -Dexec.mainClass=org.unicode.cldr.tool.GenerateTestData
+          -Dexec.cleanupDaemonThreads=false
+          -DCLDR_DIR=$(pwd)
       - name: Verify git Changes
         id: gitstatus
         run: |
@@ -86,6 +92,7 @@ jobs:
           git config user.name 'cldr-modify.yml [bot]'
           git config user.email 'unicode-org@users.noreply.github.com'
           git add common/main
+          git add common/testData
           bash .github/workflows/cldr-modify-commit.sh cldr-modify.yml CLDRModify
           git checkout -b ${AUTO_BRANCH}
           # reset origin using token


tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateTestData.java
```diff
@@ -9,5 +9,6 @@ public static void main(String[] args) throws IOException {
         GeneratePersonNameTestData.main(args);
         GenerateUnitTestData.main(args);
         GenerateDateTimeTestData.main(args);
+        GenerateRBNFTestData.main(args);
```
grhoten, line 12: This only regenerates the data for the new locales. This is under the assumption that someone may have manually written test cases in the .ssv files that should not be overwritten. I hope that’s OK.

The unit tests also validate roundtripping a little differently without these files. So all new rules are tested with TestRBNF.

grhoten (dismissed):

macchiati (dismissed):

.github/workflows/maven.yml
```diff
@@ -359,6 +394,7 @@ jobs:
     needs:
       - build
       - check
+      - data-check
```
srl295, line 397: Test data doesn't affect the server, why the dependency ?

This will only have the effect of blocking smoketest deploy from main, exactly when nobody can do anything about it. 

I'd remove this.

srl295:
Thanks for doing this, this is much needed and well help the test data not be out of date! Let's fine-tune how and where this runs

.github/workflows/maven.yml
macchiati: Good point.

srl295:
That is, can you look at the [cldr-modify.yml](https://github.com/unicode-org/cldr/actions/workflows/cldr-modify.yml) job? Let's actually create a PR if there's a problem rather than just failing.

sffc:
@srl295 I pushed a change to cldr-modify.yml. Is this what you were thinking? Let me know how to proceed.

Currently the PR contains both the check and the thing that adds the testdata changes to cldr-modify. I am okay with either approach. In ICU4X we do everything with checks so that you can see the impact of changes in each individual PR, but I understand if CLDR prefers the ease of separating out "datagen" changes from code changes. Just let me know. Thanks

srl295:
> @srl295 I pushed a change to cldr-modify.yml. Is this what you were thinking? Let me know how to proceed.
> 
> 
> 
> Currently the PR contains both the check and the thing that adds the testdata changes to cldr-modify. I am okay with either approach. In ICU4X we do everything with checks so that you can see the impact of changes in each individual PR, but I understand if CLDR prefers the ease of separating out "datagen" changes from code changes. Just let me know. Thanks

We do not want all data-changing PRs to _require_ running tooling.  That raises the bar to be too high.

.github/workflows/maven.yml
```diff
@@ -275,6 +275,41 @@ jobs:
           -Dexec.cleanupDaemonThreads=false
           -Dexec.args="--schema=keyboards/dtd/ldmlKeyboardTest.xsd keyboards/test"
           || (echo 'Schema failed, rerun GenerateDtd' ; exit 0)
+  data-check:
```
srl295: I would drop this and only run on merge (in the other workflow).  I think it will otherwise just add spurious failures.

keeping test data more up to date is a good improvement (which your change to cldr-modify.yml does).

srl295 (changes requested):
Remove from maven.yml

srl295 (approved):

sffc:
It Works: https://github.com/unicode-org/cldr/pull/5629

PR #5608, opened 2026-04-19
CLDR-8260 Add range patterns for intervals

macchiati:
CLDR-8260

This is the only part of CLDR-8260 that needs to be done before the start of submission. It is adding 3 new range symbols. The types are:
- `numeric`: for two numeric fields of the same type (eg, days) that form a range, eg in "March 3–9"
- `non-numeric`: for two non-numeric fields of the same type (eg Months) that form a range, eg in "March-May 2026"
- `mixed`: for two parts that are complex, such as in "March 23 – May 17". (The range symbol doesn't separate two fields of the same type).

The design doc just added one, but it is useful to add all 3 instead of overloading other fields.
- The numeric may be different for different calendar systems
- We need to split the mixed from non-numeric
- Having the patterns in a common element lets us inherit everything from 'gregorian'.

Here is the structure:
```
<intervalFormatRanges>
	<intervalFormatRange type="mixed">{0} – {1}</intervalFormatRange>
	<intervalFormatRange type="numeric">{0}–{1}</intervalFormatRange>
	<intervalFormatRange type="non-numeric">{0}–{1}</intervalFormatRange>
</intervalFormatRanges>
```
For English, mixed has thin spaces around an en-dash, and numeric & non-numeric are identical (as per Chicago Manual of Style).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -502,6 +502,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<appendItem request="Year">{1} {0}</appendItem>
 					</appendItems>
 					<intervalFormats>
+						<intervalFormatRanges>
+							<alias source="locale" path="../../calendar[@type='gregorian']/dateTimeFormats/intervalFormats/intervalFormatRanges"/>
+						</intervalFormatRanges>				
 						<intervalFormatFallback>{0} – {1}</intervalFormatFallback>
 						<intervalFormatItem id="Bh">
 							<greatestDifference id="B">h B – h B</greatestDifference>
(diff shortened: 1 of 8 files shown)


macchiati:
Note: once this lands, we can add a link to it from Updating DTDs as an example, since it provides a relatively simple example of changes to files that need fixing.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -1777,35 +1777,65 @@ private void handleMiscPatterns(XPathParts parts, String value, List<String> exa
...
-        if (parts.getElement(6).equals("intervalFormatFallback")) {
-            SimpleDateFormat dateFormat = new SimpleDateFormat();
-            String fallbackFormat = invertBackground(setBackground(value));
-            examples.add(
-                    format(
-                            fallbackFormat,
-                            dateFormat.format(FIRST_INTERVAL),
-                            dateFormat.format(SECOND_INTERVAL.get("y"))));
-            return;
+        switch (parts.getElement(6)) {
```
macchiati, line 1780: Note that I had to modify this code to have code for a 3rd kind of path under "intervalFormats". (And that itself needed three different examples for clarity.)

common/main/root.xml
```diff
@@ -502,6 +502,9 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<appendItem request="Year">{1} {0}</appendItem>
 					</appendItems>
 					<intervalFormats>
+						<intervalFormatRanges>
```
macchiati, line 505: By grouping under the intervalFormatRanges, we can inherit across all calendars.

macchiati:
Please review; I'd like to get this in this morning if possible.

srl295 (approved):

PR #5612, opened 2026-04-21
CLDR-18954 Tags: revise visual appearance

btangmu:
-Order tag menu items the same as CodePointEscaper enum; stop at ALM; extend CharInfo.java and CodePointEscaper.java accordingly

-Color/style the tag area like similar escapes in the main vetting table

-Reduce individual tag padding/margin to 1pt

-Add data and methods to cldrEscaper.mjs; move exports to bottom

-Collapse the tag area if the value is empty

-Enclose insert menu options in div with width 100 percent to enlarge hover area

-Extend CharInfo.java to deliver the menu labels

-Fix minor Java compiler warnings

-Comments, including TODO for long shortName comment

CLDR-18954

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/CharInfo.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/CharInfo.java
@@ -52,14 +52,29 @@ public class CharInfo {
     }
 
     public static final class EscapedCharInfo {
+        @Schema(
+                description = "A regex indicating the front end should escape matching values",
+                required = true)
         public final String forceEscapeRegex =
                 CodePointEscaper.regexPattern(CodePointEscaper.ESCAPE_IN_SURVEYTOOL, "\\u{", "}");
+
+        @Schema(
+                description =
+                        "A map from characters to character data, ordered by Unicode Scalar Value",
+                required = true)
         public final Map<String, EscapedCharEntry> names = new HashMap<>();
+
+        @Schema(
+                description =
+                        "An array of character names, filtered and ordered suitably for an input menu",
+                required = true)
+        public final String[] namesForMenu;
 
         EscapedCharInfo() {
             for (final CodePointEscaper c : CodePointEscaper.values()) {
                 names.put(c.getString(), new EscapedCharEntry(c));
             }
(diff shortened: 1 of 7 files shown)


btangmu:
image

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/CharInfo.java
```diff
@@ -52,14 +52,29 @@ public EscapedCharEntry(final CodePointEscaper c) {
...
+                description =
+                        "A map from characters to character data, ordered by Unicode Scalar Value",
+                required = true)
         public final Map<String, EscapedCharEntry> names = new HashMap<>();
 
+        @Schema(
+                description =
+                        "An array of character names, filtered and ordered suitably for an input menu",
+                required = true)
+        public final String[] namesForMenu;
```
btangmu, line 71: I had difficulty getting `namesForMenu` to be included in the json. OpenApi/Swagger seems hard to please in relation to `String[]`

srl295: what was the error?

srl295 (approved):

tools/cldr-apps/src/main/java/org/unicode/cldr/web/api/CharInfo.java
btangmu: The error was, `namesForMenu` was simply missing from the http response json. I'm not sure if it's worth trying to reproduce. I'm not sure if the solution was `@Schema` or `required = true` or some difference in how/where the array was populated. Stepping through the code I could see that it was populated, but omitted from the json. Have you noticed problems like that?

srl295: maybe because it's final? 
i have had trouble where `getXxxx()` automatically creates a (possibly unwanted) `xxxx` data member. I don't see where the problem would be otherwise, but adding the schema isnt' bad

actually, if it's final then it HAS to be present (required) because you couldn't initialize the object without it?  So maybe that's it?

PR #5617, opened 2026-04-22
CLDR-19096 Remove unused Apia metazone names

btangmu:
CLDR-19096

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ab.xml
+++ b/common/main/ab.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2025 Unicode, Inc.
+<!-- Copyright © 1991-2026 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-3.0
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
@@ -3058,8 +3058,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<metazone type="Apia">
 				<long>
 					<generic draft="unconfirmed">Апиа</generic>
-					<standard draft="unconfirmed">Апиа, астандартә аамҭа</standard>
-					<daylight draft="unconfirmed">Апиа, аԥхынтәи аамҭа</daylight>
 				</long>
 			</metazone>
 			<metazone type="Aqtau">
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -4015,8 +4015,6 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<metazone type="Apia">
 				<long>
 					<generic>Apia-tyd</generic>
-					<standard>Apia-standaardtyd</standard>
-					<daylight>Apia-dagligtyd</daylight>
 				</long>
 			</metazone>
 			<metazone type="Arabian">
--- a/common/main/ak.xml
+++ b/common/main/ak.xml
@@ -3515,8 +3515,6 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<metazone type="Apia">
 				<long>
 					<generic>Apia Berɛ</generic>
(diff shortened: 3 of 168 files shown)


btangmu:
I was confused by the instructions about -k and -fk options. I ended up temporarily modifying this file --

cldr/tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt

-- to contain only the following:

```
locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Apia"]/long/standard
locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Apia"]/long/daylight
```

Then I ran
`mvn --file=tools/pom.xml clean install package -DskipTests=true`
and
`java -DCLDR_DIR=$(pwd) -jar tools/cldr-code/target/cldr-code.jar modify -fk`

To undo changes to en.xml, I did this:
`rm ../Generated/cldr/cldrModify/en.xml`

Then I replaced the xml files in common/main:
`mv ../Generated/cldr/cldrModify/*.xml common/main`

I reverted the changes to the config file:
`git checkout -- tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt`

AEApple:
It looks good to me, I do like including the config files though just so that we know what the intention of the large scale change was. I'm going to update the instructions based on your report.

AEApple (approved):

AEApple:
Are we ready to merge this?

robertbastian (changes requested):
Zones that don't use DST use the *standard* name. You should delete the generic and daylight names

robertbastian:
https://github.com/unicode-org/cldr/blob/7776faf5896ae8d461c9776d76d38c577f2c2977/common/main/en.xml#L4630-L4639

robertbastian:
It looks like we probably want to move the generic value into standard.

AEApple:
> It looks like we probably want to move the generic value into standard.

If there is only one version is that the right approach? I don't fully understand which order items get used based on their availability.

robertbastian:
Well looking at other standard-only metazones, they have only standard names

AEApple:
Okay so you're proposing this instead?

locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Apia"]/long/generic
locale=/.*/; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Apia"]/long/daylight

robertbastian:
Well, the current generic name should be the future standard name. I guess for ST locales it doesn't matter, as it'll be changed anyway, but you added the English "West Samoa Time" as the generic name, we want to move that to standard before deleting generic.

AEApple:
> Well, the current generic name should be the future standard name. I guess for ST locales it doesn't matter, as it'll be changed anyway, but you added the English "West Samoa Time" as the generic name, we want to move that to standard before deleting generic.

I'd rather just update it after this change merges.

robertbastian:
If we're ok with losing the generic translations in all other languages then you can do that.

AEApple:
No, I was proposing that we ask Tom, to replace this change with instead of the original CLDR Modify with config run:

```
locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Apia"]/long/generic
locale=/.*/; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Apia"]/long/daylight
```

Then once that is done I will update the English.

robertbastian:
Yes, and that will nuke all generic translations. You will then update English, but the generic translations for all other language will be lost. Which is probably fine.

btangmu:
> ask Tom, to replace this change

Sorry I wasn't able to get to it today. I should be able to by tomorrow.

AEApple:
Actually we should also update the status to provisional if we are zeroing out the votes. I'll update this comment with the new logic if I'm able to find it.

AEApple:
Okay the page doesn't say how to change status in [CLDRModify using Config file](https://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-using-config-file), but I know we have before. I'll make a point of finding out before next release and documenting it.

btangmu:
See https://github.com/unicode-org/cldr/pull/5631

PR #5619, opened 2026-04-23
CLDR-19406 Make 5 Append Date-Time Items visible

macchiati:
CLDR-19406

From the ticket:

- Add Time-Day-of-Week (we added Date-Timezone for 49, but missed this one).
- Change root so that other calendars alias to gregorian. (This makes them available in islamic, etc)
- Surface the 4 missing ones in coverage. 

This would add 8 items per calendar that the locale cares about. Defaults would be as current.

Would also need to add to InfoHub, but not doing here because of possible conflicts.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -2335,6 +2335,11 @@ annotations.
 						<appendItem request="Year">{0} {1}</appendItem>
 					</appendItems>
 					<intervalFormats>
+						<intervalFormatRanges>
+							<intervalFormatRange type="mixed">{0} – {1}</intervalFormatRange>
+							<intervalFormatRange type="non-numeric">{0}–{1}</intervalFormatRange>
+							<intervalFormatRange type="numeric">{0}–{1}</intervalFormatRange>
+						</intervalFormatRanges>
 						<intervalFormatFallback>{0} – {1}</intervalFormatFallback>
 						<intervalFormatItem id="Bh">
 							<greatestDifference id="B">h B – h B</greatestDifference>
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -2321,7 +2321,8 @@ annotations.
 					</availableFormats>
 					<appendItems>
 						<appendItem request="Day">{0} ({2}: {1})</appendItem>
-						<appendItem request="Day-Of-Week">{0}, {1}</appendItem>
+						<appendItem request="Day-Of-Week">{1}, {0}</appendItem>
+						<appendItem request="Time-Day-Of-Week">{1}, {0}</appendItem>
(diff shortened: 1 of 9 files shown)


macchiati:
First cut; will see what fails in unit tests. It should only affect gregorian, generic (and chinese in locales that have that calendar)

Not blocker for Shakedown!

macchiati:
Reviewers: feel free to merge if you approve.

common/main/en.xml
```diff
@@ -2321,18 +2321,25 @@ annotations.
 					</availableFormats>
 					<appendItems>
 						<appendItem request="Day">{0} ({2}: {1})</appendItem>
-						<appendItem request="Day-Of-Week">{0}, {1}</appendItem>
+						<appendItem request="Day-Of-Week">{1}, {0}</appendItem>
+						<appendItem request="Time-Day-Of-Week">{1}, {0}</appendItem>
```
AEApple, line 2325:
```suggestion
						<appendItem request="Time-Day-Of-Week">{1} {0}</appendItem>
```

I don't think there is a comma in this. No matter how I would write it:

- Tuesday 5pm
- 5pm Tuesday

I wouldn't put a comma in it? Do other people write this with commas?

"Time-Day-Of-Week — for adding a Day-Of-Week to a time (where there is no date)" per the ticket

common/main/en.xml
```diff
@@ -2972,14 +2979,16 @@ annotations.
...
```
AEApple, line 2983:
```suggestion
						<appendItem request="Time-Day-Of-Week">{1} {0}</appendItem>
```

See comment above about commas in Day of the Week + Time expressions

common/main/en.xml
AEApple: @stenshamn - WDYT?

AEApple (changes requested):

common/main/en.xml
stenshamn: Honestly, I would _probably_ use a comma, but I could go either way. 
"Tuesday, 5pm" seems fine to me.
I probably wouldn't write "5pm, Tuesday", but rather "5pm on Tuesday", but that may be asking too much.

AEApple: Okay, I'm going to ask several UX writers to write it tomorrow and see what they do.

AEApple: You're right. It seems like it's common to write with a comma. I still wouldn't write it that way, but my English is muddled at best.

AEApple (approved):

sffc (commented):
LGTM, these are the most important append items and adding Time-Day-Of-Week is a good call

PR #5627, opened 2026-04-24
CLDR-18541 Add new date format items in root.xml to en.xml

AEApple:
CLDR-18541

Add en.xml values for the items in root.xml which were added to the Gregorian calendar in [PR#5583](https://github.com/unicode-org/cldr/pull/5583/changes)

ALLOW_MANY_COMMITS=true


macchiati (approved):

sffc:
Yay, cldr-modify worked: https://github.com/unicode-org/cldr/pull/5629

PR #5628, opened 2026-04-24
CLDR-19129 Developer info on debugging new structure

macchiati:
CLDR-19129

The goal is to provide more information to help debug problem with adding new structure. 
- I am starting with some of the useful tools, but plan to provide more text about methods and pitfalls.
- Feedback & suggestions welcome

ALLOW_MANY_COMMITS=true


development/updating/debugging.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Debugging Structure
+---
+
+# Debugging Structure
+# _DRAFT!!_
```
srl295:
```suggestion
> _DRAFT!!_
```

Page shouldn't have two titles.

Actually, the `# Debugging Structure` is also unneeded.

srl295:
There's actually *yet more* pages that weren't brought over from the old site.

For example SearchXML had its own page.

I wrote up guidance on how to document them [here](https://cldr.unicode.org/development/coding-cldr-tools/documenting-cldr-tools) and documented some of them.

image

All to say, I think this is valuable content. but I would suggest that we have a page for SearchXML (etc) and fill that out with documentaiton.

srl295:
> Did you mean to put this on the site?

Do you mean this page? Then yes.

macchiati:
> There's actually _yet more_ pages that weren't brought over from the old site.
> 
> For example SearchXML had its own page.
> 
> I wrote up guidance on how to document them [here](https://cldr.unicode.org/development/coding-cldr-tools/documenting-cldr-tools) and documented some of them.
...
> All to say, I think this is valuable content. but I would suggest that we have a page for SearchXML (etc) and fill that out with documentaiton.

Do we have a ticket for bringing over the missing pages? That would be great.

I have a different target for this, which is to flesh it out specifically targeted to debugging the addition of new structure. I want to have short glosses for the tools how to apply options for this purpose. Where we have dedicated pages for a tool, that gloss should include a link to that dedicated page for the tool's details.

srl295:
> > There's actually _yet more_ pages that weren't brought over from the old site.
> > For example SearchXML had its own page.
> > I wrote up guidance on how to document them [here](https://cldr.unicode.org/development/coding-cldr-tools/documenting-cldr-tools) and documented some of them.
> > ...
> > All to say, I think this is valuable content. but I would suggest that we have a page for SearchXML (etc) and fill that out with documentaiton.
> 
> Do we have a ticket for bringing over the missing pages? That would be great.

I will check but I think we've "thought we were done" a number of times. 

> I have a different target for this, which is to flesh it out specifically targeted to debugging the addition of new structure. I want to have short glosses for the tools how to apply options for this purpose. Where we have dedicated pages for a tool, that gloss should include a link to that dedicated page for the tool's details.

Makes sense. Should it be under docs/site/development instead of under a new root /development directory though?  Or at least docs/development?

macchiati:
> Makes sense. Should it be under docs/site/development instead of under a new root /development directory though? Or at least docs/development?

Can do, my mistake.. (Our file organization is semi-orthongonal to the organization of https://cldr.unicode.org/ so that we can fix the latter organization without having to futz with URLs. A good thing but makes it not obvious where to put files.)

srl295:
> > Makes sense. Should it be under docs/site/development instead of under a new root /development directory though? Or at least docs/development?
> 
> Can do, my mistake.. (Our file organization is semi-orthongonal to the organization of https://cldr.unicode.org/ so that we can fix the latter organization without having to futz with URLs. A good thing but makes it not obvious where to put files.)

yes, semi enough to cause confusion!  no problem.  and 1.000‰ on board with documenting this better…

docs/development/debugging.md
```diff
@@ -5,15 +5,43 @@ title: Debugging Structure
...
 > _DRAFT!!_
 
-Updating structure in CLDR can be challenging. 
-Often that takes many steps, as documented on [Updating DTDs](docs/site/development/updating-dtds.md): 
-DTDs, path headers, path descriptions, coverage, examples, and so on.
-The tests help to catch problems, but it is often tricky to find out why a test is failing.
+Updating structure changes in CLDR can be challenging. The goal of this document is to help you understand what is going wrong, and how to fix the problems.
 
-There are some tools that can help.
+Adding the structure can take many steps, as documented on [Updating DTDs](docs/site/development/updating-dtds.md):
```
srl295, line 10: not sure where you're wanting this link to go… 

you put the file under `docs` but should it be under `docs/site` ?

sffc (commented):
Seems good. The best way for me to review this is to hit these issues during development and then see if the proposed solutions work.

It's good to document these in the repo since AI tools can also see this.

srl295:
One other thing that needs to be updated
- Placeholders.txt and, TestCLDRFile.testExtraPaths to indicate that placeholders are NOT expected.

PR #5630, opened 2026-04-24
CLDR-19353 Tags for Show Hidden in main vetting table

btangmu:
-Display special characters in main table similar to the Add Value dialog

-New ValueTags.vue, similar to AddValueTags.vue but read-only, ⓘ, other differences

-Move shared code to cldrAddValue.mjs

-Remove obsolete code

CLDR-19353

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrAddValue.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrAddValue.mjs
@@ -1,7 +1,8 @@
 /*
- * cldrAddValue: enable submitting a new value for a path
+ * cldrAddValue: enable displaying a value (for a CLDR path), or submitting (adding) a new value
  */
 import * as cldrChar from "./cldrChar.mjs";
+import * as cldrEscaper from "./cldrEscaper.mjs";
 import * as cldrNotify from "./cldrNotify.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
 import * as cldrTable from "./cldrTable.mjs";
@@ -10,6 +11,12 @@ import * as cldrVue from "./cldrVue.mjs";
 import * as cldrXpathUtils from "./cldrXpathUtils.mjs";
 
 import AddValue from "../views/AddValue.vue";
+import ValueTags from "../views/ValueTags.vue";
+
+/**
+ * Should ASCII space (U+0020) be displayed as a tag?
+ */
+const SPACES_HAVE_TAGS = false;
 
 /**
  * Is the "Add Value" form currently visible?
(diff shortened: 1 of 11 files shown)


btangmu:
There's a test failure. I'll look at it shortly. I'm in the middle of another ticket (CLDR-19096)

btangmu:
> test failure

fixed

macchiati:
Just started workflow for staging; for these UI changes that's handy to have.

macchiati (changes requested):
The circled-i goes to https://cldr.unicode.org/translation/getting-started/guide#special-characters. We should update that as well. I'd suggest

OLD

> These will be shown underneath the regular value in an additional box. For example:
> 
> h:mm:ss a z
> 
> h:mm:ss❰NBTSP❱a z
> 
> This indicates that there is a special character that was not visible in the upper line, in this case a no-break thin space, U+202F.

NEW

These will be shown underneath the regular value in an additional box. For example:

h:mm:ss a z

[Image of new show-hidden box]

This indicates that there is a special character that was not visible in the upper line, in this case a no-break thin space, U+202F.

NOTE: The Alphabetic Characters section shows an older format, with bold angle-brack characters, such as:

h:mm:ss❰NBTSP❱a z


NOTE: I'm assuming that a different PR will make the ShowHidden box in both locations be non-selectable. But that can be a different PR.

tools/cldr-apps/js/src/esm/cldrAddValue.mjs
```diff
@@ -10,6 +11,12 @@ import * as cldrVue from "./cldrVue.mjs";
 import * as cldrXpathUtils from "./cldrXpathUtils.mjs";
 
 import AddValue from "../views/AddValue.vue";
+import ValueTags from "../views/ValueTags.vue";
+
+/**
+ * Should ASCII space (U+0020) be displayed as a tag?
+ */
+const SPACES_HAVE_TAGS = false;
```
srl295, line 19: Really want to get this Server-to-client mechanism in place. So we don't have to have these or DEBUG all over..

tools/cldr-apps/js/src/views/ValueTags.vue
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.right-side {
+  color: black;
+}
+
+/* Make the ⓘ (right side) more salient when hovering on the tag area, less salient otherwise */
+.tag-area:hover > .right-side {
+  opacity: 1;
+}
```
srl295, line 86: very nice!

srl295 (approved):
I'm going to +1 the code here, you can coordinate w/ Mark on the behavior.

macchiati (approved):
I'm ok with getting in the other changes in a later PR.

PR #5631, opened 2026-04-24
CLDR-19096 Remove Samoan time zone names

btangmu:
-In all locales except en, remove metazone long/generic and long/daylight paths with type Apia

CLDR-19096

ALLOW_MANY_COMMITS=true


btangmu:
This deletes generic and daylight, where the earlier (closed) https://github.com/unicode-org/cldr/pull/5617 would have deleted standard and daylight

AEApple (approved):

PR #5634, opened 2026-04-24
CLDR-19227 Prevent digit-digit merges

btangmu:
-New element placeholderBoundarySpacing, with attributes type, scope

-Coverage is moderate for lineBreakLetterLanguages (new category), comprehensive for other languages

-Update CldrVersion and ToolConstants to include 48.2

-Update outdated.data and outdatedEnglish.data by running GenerateBirth

CLDR-19227

ALLOW_MANY_COMMITS=true


btangmu:
This now passes tests and ST runs OK locally. Screenshot:
image

@macchiati I looked at https://cldr.unicode.org/development/updating-dtds#updating-dtds and have some comments/questions.

Comments:
- The new attributes are `#REQUIRED`, and they are distinguishing
- The attribute values are closed sets (only `digit-digit` for `type`, and only `datetime` for `scope`)
- I imagine the Info Panel text in PathDescriptions.md could use improvement. For now I made it: `The character or short string used as a separator. For example, a space may be used to separate timezone "GMT+5" and year "2026" to produce "GMT+5 2026..." rather than "GMT+52026...".`

Questions:
- Should Placeholders.txt be revised?
- Should we run GenerateBirth? Note that CldrVersion.java currently goes up to v48_1; should it be v48_2?
- For now the new path is in `TestExampleGenerator.TEMPORARY_EXCLUDED_EXAMPLES`. I'm not sure how to construct a suitable example in ExampleGenerator.java; any suggestions, or suitable model to imitate?

common/dtd/ldml.dtd
```diff
@@ -648,6 +648,12 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 
 <!-- ######################################################### -->
 
+<!ELEMENT placeholderBoundarySpacing ( alias | ( parseLenient*, special* ) ) >
```
macchiati: It won't have any children, so it can't have `alias | ( parseLenient*, special* )`. Take a look at similar cases that are leaf nodes.

common/dtd/ldml.dtd
```diff
@@ -648,6 +648,12 @@ CLDR data files are interpreted according to the LDML specification (http://unic
...
+<!ATTLIST placeholderBoundarySpacing type (digit-digit) #REQUIRED >
```
macchiati, line 652: These values have to allow spaces, so best to check in the Survey tool (staging or locale) to check that you can change the space to another character (like *) and then back.

common/supplemental/coverageLevels.xml
```diff
@@ -1045,6 +1046,9 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel value="moderate" match="characters/moreInformation"/>
 		<coverageLevel value="moderate" match="characters/parseLenients[@scope='%anyAttribute'][@level='%anyAttribute']/parseLenient[@sample='%anyAttribute']"/>
 
+		<coverageLevel inLanguage="%lineBreakLetterLanguages" value="moderate" match="characters/placeholderBoundarySpacing[@type='digit-digit'][@scopes='datetime']"/>
```
macchiati: It is safest to use type = %A, scopes=%A; that way we don't miss something if we add a new attribute value

macchiati:
I hadn't been looking at this because it was still marked Draft.

> The new attributes are #REQUIRED, and they are distinguishing

That's ok for _new_ elements.

> The attribute values are closed sets (only digit-digit for type, and only datetime for scope)

Slightly safer to use %A instead for coverage; that way a new path automatically shows up.

> I imagine the Info Panel text in PathDescriptions.md could use improvement. For now I made it: The character or short string used as a separator. For example, a space may be used to separate timezone "GMT+5" and year "2026" to produce "GMT+5 2026..." rather than "GMT+52026...". 

Add "See [preventing-digit-digit-concatenations](https://cldr.unicode.org/translation#preventing-digit-digit-concatenations)."

And use the example there, which is probably more realistic.

> Should Placeholders.txt be revised?

Yes. The Code value will be waaay long. 
Probably best is Spacing for $2; separating $1

> Should we run GenerateBirth? Note that CldrVersion.java currently goes up to v48_1; should it be v48_2?

Yes to both
@AEApple After Shakedown, we should make sure we run and check in GenerateBirth after any change in data in en.xml

> For now the new path is in TestExampleGenerator.TEMPORARY_EXCLUDED_EXAMPLES. I'm not sure how to construct a suitable example in ExampleGenerator.java; any suggestions, or suitable model to imitate?

For now that's ok. The problem is that an example with real locale data will only make sense if some of the available formats in the locale together with appendItems results in a pattern with two adjacent digits that belong to different fields. (as per the example on the Info Hub). 

We need to update the code before showing an example. @sffc will be working on that, but we don't need it for starting Shakedown.

macchiati (changes requested):

macchiati:
Just a few small fixes needed, ow looks ready to go.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Placeholders.txt
```diff
@@ -268,3 +268,5 @@
 ^//ldml/numbers/rationalFormats/rationalPattern ; {0}=NUMERATOR integer or negative integer; {1}=DENOMINATOR integer
 ^//ldml/numbers/rationalFormats/integerAndRationalPattern ; {0}=INTEGER integer or negative integer; {1}=RATIONAL formatted positive rational
 ^//ldml/numbers/rationalFormats/integerAndRationalPattern\[@alt="%A"] ; {0}=INTEGER integer or negative integer; {1}=RATIONAL formatted positive rational
+
+# ^//ldml/characters/placeholderBoundarySpacing\[@type="digit-digit"]\[@scopes="datetime"] ; {0}=FOO bar; {1}=FOO bar
```
btangmu: What should replace FOO and bar here?

btangmu: @macchiati I've addressed most of your comments. Here I'm still perplexed. You wrote "Probably best is Spacing for $2; separating $1", I think related to what's needed in Placeholders.txt. Note that the line I've added is commented out; tests failed otherwise.

macchiati:
I remember what the space issue is. Where we don't want spaces trimmed, we add @xml:space.

An example is //ldml/personNames/foreignSpaceReplacement

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Placeholders.txt
macchiati: Ah, I missed this. The values for this path have no placeholders, so just delete this line.

That is, people can never put in something like {0} into the value: it is just a constant. See however the note about space.

```suggestion
```

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDisplayAndInputProcessor.java
```diff
@@ -775,6 +775,7 @@ public void TestFSR() {
         DisplayAndInputProcessor daip = new DisplayAndInputProcessor(info.getEnglish(), false);
         checkPathAllowsEmpty(daip, DisplayAndInputProcessor.FSR_START_PATH);
         checkPathAllowsEmpty(daip, DisplayAndInputProcessor.NOL_START_PATH);
+        checkPathAllowsEmpty(daip, DisplayAndInputProcessor.PBS_START_PATH);
```
macchiati, line 778: good! We need to add this to the 'how to debug' page

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
```diff
@@ -525,6 +525,15 @@ public void TestCoverageCompleteness() {
                 continue;
             }
             Level lvl = sdi.getCoverageLevel(path, "en");
+            if (path.equals(
```
macchiati, line 528: Why did you need to add this? The coverage should be Modern.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
```diff
@@ -566,7 +566,9 @@ public CheckCLDR handleCheck(
 
         // check for spaces
 
-        if (!value.equals(value.trim()) && !path.contains("SpaceReplacement")) {
+        if (!value.equals(value.trim())
```
macchiati, line 569: For later; we should have a common resource that can be used for both this Check and for DAIP.

macchiati (changes requested):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCoverageLevel.java
macchiati: I take this back; we are using Moderate, and only for languages that don't need spaces.

macchiati (approved):
I think this is now good. We can delete the PathDescription line later, so approving

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Placeholders.txt
macchiati: Spacing for $2; separating $1" was for the PathHeader, not the placeholders. Sorry if I put it in the wrong place.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -54,6 +54,7 @@
 //ldml/delimiters/alternateQuotation%E                      ; Core Data ; Alphabetic Information ; Quotation Marks ; embedded-$1
 //ldml/characters/moreInformation                           ; Core Data ; Alphabetic Information ; Symbols ; More Information
 //ldml/characters/nestedBracketReplacement[@bracket="%A"]   ; Core Data ; Alphabetic Information ; Symbols ; Nested Bracked Replacement: $1
+//ldml/characters/placeholderBoundarySpacing[@type="%A"][@scopes="%A"]   ; Core Data ; Alphabetic Information ; Symbols ; Placeholder Boundary Spacing $1-$2
```
macchiati: Spacing for $2; separating $1
```suggestion
//ldml/characters/placeholderBoundarySpacing[@type="%A"][@scopes="%A"]   ; Core Data ; Alphabetic Information ; Symbols ; Spacing for $2, separating $1
```

macchiati (approved):
Whoops, missed this earlier.

macchiati (changes requested):
We need to fix the PathHeader, sorry for the confusion (hurrying too much!)

sffc (dismissed):
Design looks about right. I can implement this in the CldrDateTimePatternGenerator once it lands.

btangmu:
@macchiati With the 6th commit I think I've taken all your comments into account. I added `@xml:space` in ldml.dtd only; I'm not sure if that's sufficient, but tests pass, and locally running ST I could change the value from space to asterisk and back to space again.

common/dtd/ldml.dtd
btangmu: I've done that.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/Placeholders.txt
btangmu: Got it, thanks!

macchiati (approved):

PR #5636, opened 2026-04-25
CLDR-19415 Add Mara (mrh) keyboard layout version 3.0

Laitei40:
CLDR-19415

ALLOW_MANY_COMMITS=true


srl295:
Keyboard looks good so far (will review in detail).

However, exemplar changes should not be in this same PR.  Please open a PR targeting https://unicode-org.atlassian.net/browse/CLDR-18760 with just the exemplar change (adding a letter)? Assuming @conradarcturus is OK with reopening it.

The exemplar spacing changes are superfluous

Laitei40:
> Keyboard looks good so far (will review in detail).
> 
> However, exemplar changes should not be in this same PR. Please open a PR targeting https://unicode-org.atlassian.net/browse/CLDR-18760 with just the exemplar change (adding a letter)? Assuming @conradarcturus is OK with reopening it.
> 
> The exemplar spacing changes are superfluous

@srl295 Thank you for the review and guidance!

I have reverted the exemplar and spacing changes in this branch, so this PR is now strictly focused on the Mara keyboard layout version 3.0.

I am preparing a separate PR targeting CLDR-18760 for just the exemplar letter addition (without the spacing changes), and I will tag @conradarcturus there to ask about reopening the ticket.

Please let me know if the keyboard layout looks good to go or if you need any further adjustments!

srl295:
> > Keyboard looks good so far (will review in detail).
> 
> > 
> 
> > However, exemplar changes should not be in this same PR. Please open a PR targeting https://unicode-org.atlassian.net/browse/CLDR-18760 with just the exemplar change (adding a letter)? Assuming @conradarcturus is OK with reopening it.
> 
> > 
> 
> > The exemplar spacing changes are superfluous
> 
> 
> 
> @srl295 Thank you for the review and guidance!
> 
> 
> 
> I have reverted the exemplar and spacing changes in this branch, so this PR is now strictly focused on the Mara keyboard layout version 3.0.
> 
> 
> 
> I am preparing a separate PR targeting CLDR-18760 for just the exemplar letter addition (without the spacing changes), and I will tag @conradarcturus there to ask about reopening the ticket.
> 
> 
> 
> Please let me know if the keyboard layout looks good to go or if you need any further adjustments!

Thanks. 

Do you have a survey tool login? You can request exemplar changas there without needing a PR.

Laitei40:
> > > Keyboard looks good so far (will review in detail).
> > 
> > 
> > > 
> > 
> > 
> > > However, exemplar changes should not be in this same PR. Please open a PR targeting https://unicode-org.atlassian.net/browse/CLDR-18760 with just the exemplar change (adding a letter)? Assuming @conradarcturus is OK with reopening it.
> > 
> > 
> > > 
> > 
> > 
> > > The exemplar spacing changes are superfluous
> > 
> > 
> > @srl295 Thank you for the review and guidance!
> > I have reverted the exemplar and spacing changes in this branch, so this PR is now strictly focused on the Mara keyboard layout version 3.0.
> > I am preparing a separate PR targeting [CLDR-18760](https://unicode-org.atlassian.net/browse/CLDR-18760) for just the exemplar letter addition (without the spacing changes), and I will tag @conradarcturus there to ask about reopening the ticket.
> > Please let me know if the keyboard layout looks good to go or if you need any further adjustments!
> 
> Thanks.
> 
> Do you have a survey tool login? You can request exemplar changas there without needing a PR.

@srl295 Yes, I do have an account. I have successfully submitted the exemplar change requests via the Survey Tool. With those changes handled separately, please let me know if the keyboard layout in this PR is approved or if you require any additional updates. Thanks again for the guidance.

[CLDR-18760]: https://unicode-org.atlassian.net/browse/CLDR-18760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE keyboard3 SYSTEM "../dtd/ldmlKeyboard3.dtd">
+<keyboard3 locale="mrh" conformsTo="47">
+    <info name="Mara" author="Mara Language Preservation" layout="QWERTY"/>
+    
+    <keys>
+        <import base="cldr" path="47/keys-Zyyy-punctuation.xml"/>
+        <import base="cldr" path="47/keys-Zyyy-currency.xml"/>
+
+        <key id="q" output="â"/>
```
mcdurdin, line 10: I would suggest having a method of typing q, x, and j, so that users can type domain names, etc, without having to switch keyboards. altr (Right alt) might be the easiest method.

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,86 @@
...
+            <row keys="space"/>
+        </layer>
+
+        <layer modifiers="shift">
+            <row keys="tilde bang at hash dollar percent caret amp asterisk open-paren close-paren underscore plus"/>
+            <row keys="Q W E R T Y U I O P open-curly close-curly pipe"/>
+            <row keys="A S D F G H J K L colon double-quote"/>
+            <row keys="Z X C V B N M open-angle close-angle question"/>
+            <row keys="space"/>
+        </layer>
```
mcdurdin, line 93: This keyboard does not have a Caps or Caps Shift layer defined -- do you want those too?

mcdurdin: You would need a separate layer definition because Caps affects the alphabetic keys but not the punctuation or numeric keys.

Laitei40: Thank you very much for the review and the helpful feedback.

I appreciate the suggestions regarding the dictionary-based approach, Caps/Caps Shift layers, and the layer definitions for proper capitalization behavior. I will review these issues carefully, make the necessary corrections, and update the keyboard accordingly.

Thank you again for your time, guidance, and support. Your feedback is very helpful in improving the Mara keyboard.

Laitei40 (commented):
Thank you for the review and the helpful feedback! I have updated the layout to fully support standard Caps Lock behavior. 

**Key updates in this revision:**

* Added `caps` and `caps shift` layers to ensure alphabetic keys correctly invert case while numbers and punctuation remain unshifted.
* Corrected the `j` key outputs to accurately handle `ch`, `Ch` (Shift), and `CH` (Caps Lock).
* Added AltGr (`altR`) fallback layers so users can still type the original English characters (`q`, `x`, `j`).
* Verified the build and corrected the language tag to the standard `mrh`.

Please let me know if everything looks good or if any further adjustments are needed!

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,124 @@
...
+
+        <layer modifiers="altR shift">
+            <row keys="tilde bang at hash dollar percent caret amp asterisk open-paren close-paren underscore plus"/>
+            <row keys="Q-orig W E R T Y U I O P open-curly close-curly pipe"/>
+            <row keys="A S D F G H J-orig K L colon double-quote"/>
+            <row keys="Z X-orig C V B N M open-angle close-angle question"/>
+            <row keys="space"/>
+        </layer>
```
mcdurdin, line 127: It might be better to make the unused keys into `gap` keys, rather than have them emit the same as the unmodified keystroke. You can then also omit keys to the right of the overridden key, for examepl with the zxcvbnm row:

```
            <row keys="gap x-orig"/>
```

`gap` is a predefined key identifier that means the key will have no effect.

mcdurdin (commented):
Your changes are looking good. One suggestion.

Laitei40 (commented):
**Key updates:**
* Used the `gap` key identifier for the `altR` and `altR shift` layers to properly target only the overridden keys (`q`, `x`, `j`).
* Omitted trailing unused keys to rely on standard LDML fallback behavior.
* Applied this same sparse logic to the `caps` and `caps shift` layers, completely removing the redundant number and punctuation definitions.

keyboards/3.0/mrh.xml
Laitei40: Great suggestion, thank you! I have updated the XML to utilize sparse layers as recommended.

**Key updates:**
* Used the `gap` key identifier for the `altR` and `altR shift` layers to properly target only the overridden keys (`q`, `x`, `j`).
* Omitted trailing unused keys to rely on standard LDML fallback behavior.
* Applied this same sparse logic to the `caps` and `caps shift` layers, completely removing the redundant number and punctuation definitions.

The updated layout is now much cleaner and builds perfectly. Please let me know if it looks good to go!

mcdurdin:
> * Applied this same sparse logic to the `caps` and `caps shift` layers, completely removing the redundant number and punctuation definitions.

This is not correct -- you need to define these keys on the `caps` and `caps shift` layers, or they will have no output; there is no default output.

Laitei40:
> > * Applied this same sparse logic to the `caps` and `caps shift` layers, completely removing the redundant number and punctuation definitions.
> 
> This is not correct -- you need to define these keys on the `caps` and `caps shift` layers, or they will have no output; there is no default output.

Ah, understood! I didn't realize `caps` layers lacked default fallbacks. Thank you for catching that; otherwise, typing numbers with Caps Lock on would have been broken!

I have reverted the `caps` and `caps shift` layers to fully define the number and punctuation rows. I kept the sparse `gap` logic exclusively for the `altR` layers where it functions properly as an override. 

Just pushed the fix!

tools/cldr-apps/.settings/org.eclipse.jdt.core.prefs
srl295, line 1: Ideally remove this file from the PR.  Yes, it's annoying.

tools/cldr-rdf/.settings/org.eclipse.jdt.core.prefs
srl295, line 1: Ideally remove this file from the PR.  Yes, it's annoying.

srl295 (commented):
This seems to be looking reasonable.. please remove the .settings files.

srl295:
This will need to merge fairly soon to make it into CLDR 49

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE keyboard3 SYSTEM "../dtd/ldmlKeyboard3.dtd">
+<keyboard3 locale="mrh" conformsTo="46">
+    <info name="Mara" author="Mara Language Preservation" layout="QWERTY"/>
+    
+    <keys>
+        <import base="cldr" path="46/keys-Zyyy-punctuation.xml"/>
+        <import base="cldr" path="46/keys-Zyyy-currency.xml"/>
```
miloush, line 8: check the imports are actually used

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,129 @@
...
+        <!-- Sparse Right Alt Layer: Keeps the gap optimizations -->
+        <layer modifiers="altR">
+            <row keys="gap"/>
+            <row keys="q-orig"/>
+            <row keys="gap gap gap gap gap gap j-orig"/>
+            <row keys="gap x-orig"/>
+        </layer>
+
+        <!-- Sparse Right Alt + Shift Layer: Keeps the gap optimizations -->
+        <layer modifiers="altR shift">
```
miloush, line 122: altR and altR shift seem to be casing pairs of each other, no caps for altR?

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,129 @@
...
+            <row keys="tilde bang at hash dollar percent caret amp asterisk open-paren close-paren underscore plus"/>
+            <row keys="q w e r t y u i o p open-curly close-curly pipe"/>
+            <row keys="a s d f g h j k l colon double-quote"/>
+            <row keys="z x c v b n m open-angle close-angle question"/>
+            <row keys="space"/>
+        </layer>
+
+        <!-- Sparse Right Alt Layer: Keeps the gap optimizations -->
+        <layer modifiers="altR">
+            <row keys="gap"/>
```
miloush, line 115: presumably <row /> would work?

keyboards/3.0/mrh.xml
```diff
@@ -0,0 +1,129 @@
...
+        <key id="X" output="Ô"/>
+        <key id="C" output="C"/> 
+        <key id="V" output="V"/>
+        <key id="B" output="B"/>
+        <key id="N" output="N"/>
+        <key id="M" output="M"/>
+
+        <!-- Keys for Caps Lock and AltGr (Right Alt) fallback mappings -->
+        <key id="J-caps" output="CH"/>
+        <key id="q-orig" output="q"/>
```
miloush, line 70: Why do we have `q` and `q-orig` keys rather than `â` and `q` or `a-circumflex` and `q`? The current way people cannot easily tell what the layers output in the keys. It feels like an abuse, you could basically import generic QWERTY layer and redefine output of all keys.

PR #5639, opened 2026-04-25
CLDR-19416 Fix broken Add User dialog and Add Value dialog

btangmu:
-Remove ant style overrides in AddValueCharMenu.vue and AddValueTags.vue

-Use :deep and style scoped for ant style overrides in AddValueCharMenu.vue and AddValueTags.vue

CLDR-19416

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/AddValueCharMenu.vue
+++ b/tools/cldr-apps/js/src/views/AddValueCharMenu.vue
@@ -138,32 +138,39 @@ function handleDropdownVisibleChange(isVisible) {
 }
 </script>
 
-<style>
-/* Not scoped, for overriding ant styles */
-
+<style scoped>
 /* The div with "ant-select-selector" normally contains the menu arrow and placeholder.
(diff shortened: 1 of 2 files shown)


srl295 (dismissed):

btangmu:
@srl295 I've just learned about `:deep(...)` as a way to do an override within `style scoped`. Are you likely to be available in a little while to review again if I make that change?

srl295:
> @srl295 I've just learned about `:deep(...)` as a way to do an override within `style scoped`. Are you likely to be available in a little while to review again if I make that change?

At least Within a couple of hrs ? I can also merge after review

btangmu:
The second commit, with `:deep()`, seems to fix everything!

btangmu:
> I can also merge after review

That would be great, thanks!!

srl295 (approved):

srl295:
Perfect. 💯％ 

image

image

PR #5642, opened 2026-04-25
CLDR-19417 Update ordering of DayOfMonth ordinal forms in ST

AEApple:
CLDR-19417

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
@@ -98,7 +98,7 @@
 //ldml/dates/calendars/calendar[@type="%A"]/months/monthContext[@type="%A"]/monthWidth[@type="%A"]/month[@type="%A"][@yeartype="%A"]  ; Special  ; Suppress ;  &calendar($1) ; &calField(Months:$3:$2)-M$4 (leap) ; HIDE
 //ldml/dates/calendars/calendar[@type="%A"]/months/monthContext[@type="%A"]/monthWidth[@type="%A"]/month[@type="%A"]            ; Special  ; Suppress ;  &calendar($1) ; &calField(Months:$3:$2)-M$4 ; HIDE
 
-//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; &unitCount($1-$3-$4)
+//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; &pluralNumber($1)
 //ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@type="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; $4
 //ldml/dates/calendars/calendar[@type="%A"]/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; Special  ; Suppress ; &calendar($1) ; &calField(Days:$3:$2)-&day($4) ; HIDE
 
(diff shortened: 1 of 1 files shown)


AEApple:
I'm not sure I understand the ordering logic in PathHeader.txt. It says how to order alphabetically but not by plural forms.

srl295 (dismissed):

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -98,7 +98,7 @@
 //ldml/dates/calendars/calendar[@type="%A"]/months/monthContext[@type="%A"]/monthWidth[@type="%A"]/month[@type="%A"][@yeartype="%A"]  ; Special  ; Suppress ;  &calendar($1) ; &calField(Months:$3:$2)-M$4 (leap) ; HIDE
 //ldml/dates/calendars/calendar[@type="%A"]/months/monthContext[@type="%A"]/monthWidth[@type="%A"]/month[@type="%A"]            ; Special  ; Suppress ;  &calendar($1) ; &calField(Months:$3:$2)-M$4 ; HIDE
 
-//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; $4
+//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; &pluralNumber($1)
```
srl295:
```suggestion
//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; &pluralNumber($4)
```

the test failure was "no plural type `generic`" which tells me it was using capture group 1 ( the first `%` above, aka $1 ) calendar type=generic .. so, seems like this should fix

also, the 'old' side had just a bare $4 - i.e. sorting by ordinal alphabetically
also the next line 102 had $4 being `dayOfMonth[@type="$4"]` so by extension `dayOfMonth[@ordinal="$4"]`

srl295:
@AEApple sorry delayed post on my review

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
AEApple: Okay, so &pluralNumber($4) would work? Is what 1-5 defined somewhere or do they change based on PathHeader function?

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -98,7 +98,7 @@
 //ldml/dates/calendars/calendar[@type="%A"]/months/monthContext[@type="%A"]/monthWidth[@type="%A"]/month[@type="%A"][@yeartype="%A"]  ; Special  ; Suppress ;  &calendar($1) ; &calField(Months:$3:$2)-M$4 (leap) ; HIDE
 //ldml/dates/calendars/calendar[@type="%A"]/months/monthContext[@type="%A"]/monthWidth[@type="%A"]/month[@type="%A"]            ; Special  ; Suppress ;  &calendar($1) ; &calField(Months:$3:$2)-M$4 ; HIDE
 
-//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; $4
+//ldml/dates/calendars/calendar[@type="%A"]/dayOfMonths/dayOfMonthContext[@type="%A"]/dayOfMonthWidth[@type="%A"]/dayOfMonth[@ordinal="%A"]                    ; DateTime ; &calendar($1) ; &calField(DayOfMonth:$3:$2) ; &count($2-$3)
```
srl295: i think the way you had it before was closer, see my comment

AEApple: I asked a question about if the usage of the numbers is consistent across the functions. I didn't see it written up anywhere and I didn't read through all the functions. Both currency which uses the &count() function and plurals which use the &pluralNumber() function are sorted by plural in the Survey Tool currently (as are the units, but units have more grammar that they sort by)

srl295: it's the [capture group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Capturing_group) number. If there are no other parenthesis in the pattern, in this case, it's which `%A` is being referred to.  The first %A is $1, the second $2 etc.

%A is shorthand for `(…)` or something, so each %A makes a new capture group.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
srl295: `&pluralNumber($4)` should work.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
AEApple: Thank you for the explanation! So I take it by listing only $4, it will sort by the ordinal values only using the plural logic.

srl295 (approved):
LGTM assuming tests ok

macchiati:
Thanks, good catch!

PR #5644, opened 2026-04-26
CLDR-19129 Add unit test for root aliases

macchiati:
CLDR-19129

I got suspicious while debugging; the alias paths are definitely tricky. So I wrote a unit test, and it turns out that some of them were bad. The new unit test will also help to prevent that in the future.

ALLOW_MANY_COMMITS=true


macchiati:
Resolved conflicts.

macchiati:
Resolved the merge conflicts

btangmu (approved):
Awesome!

PR #5649, opened 2026-04-27
CLDR-14086 Handle more raw quotes

robertbastian:
CLDR-14086

* Apply the Hebrew treatment to Yiddish as well
* Use the correct "prime" and "double prime" for units `length-inch`, `angle-arc-second`, `pressure-inch-ofhg`, `foot`, `angle-arc-minute`, `*-per-inch`
* Don't exclude unit display names from apostrophe handling
* Don't exclude number symbols from apostrophe handling (affects Swiss thousand separator)
* Also handle double apostrophes in `normalizeForDisplay`
* ~Use the modifier apostrophe in names of languages that use modifier apostrophes across all languages~
  * ~Previously `mic` would use ASCII apostrophes, while the others would use typographic apostrophes~

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -293,7 +293,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="mfe">Morisjen</language>
 			<language type="mg">Malgassies</language>
 			<language type="mgh">Makhuwa-Meetto</language>
-			<language type="mgo">Meta’</language>
+			<language type="mgo">Metaʼ</language>
 			<language type="mh">Marshallees</language>
 			<language type="mi">Maori</language>
 			<language type="mic">Micmac</language>
@@ -368,7 +368,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="pt_BR">↑↑↑</language>
 			<language type="pt_PT">↑↑↑</language>
 			<language type="qu">Quechua</language>
-			<language type="quc">K’iche’</language>
+			<language type="quc">Kʼicheʼ</language>
 			<language type="raj">Rajasthani</language>
 			<language type="rap">Rapanui</language>
 			<language type="rar">Rarotongaans</language>
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -293,7 +293,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="mfe">Morisjen</language>
 			<language type="mg">Malgassies</language>
 			<language type="mgh">Makhuwa-Meetto</language>
-			<language type="mgo">Metaʼ</language>
+			<language type="mgo">Meta’</language>
 			<language type="mh">Marshallees</language>
 			<language type="mi">Maori</language>
 			<language type="mic">Micmac</language>
@@ -368,7 +368,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="pt_BR">↑↑↑</language>
 			<language type="pt_PT">↑↑↑</language>
 			<language type="qu">Quechua</language>
-			<language type="quc">Kʼicheʼ</language>
+			<language type="quc">K’iche’</language>
 			<language type="raj">Rajasthani</language>
 			<language type="rap">Rapanui</language>
 			<language type="rar">Rarotongaans</language>
--- a/common/main/bn.xml
+++ b/common/main/bn.xml
@@ -2099,7 +2099,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					</eraNames>
 					<eraAbbr>
 						<era type="0">খ্রিস্টপূর্ব</era>
-						<era type="0" alt="variant">&quot;খ্রিঃপূঃ&quot;</era>
+						<era type="0" alt="variant">“খ্রিঃপূঃ“</era>
 						<era type="1">খৃষ্টাব্দ</era>
 						<era type="1" alt="variant">খ্রিষ্টাব্দ</era>
 					</eraAbbr>
--- a/common/main/bn.xml
+++ b/common/main/bn.xml
@@ -2099,7 +2099,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					</eraNames>
 					<eraAbbr>
 						<era type="0">খ্রিস্টপূর্ব</era>
-						<era type="0" alt="variant">“খ্রিঃপূঃ“</era>
+						<era type="0" alt="variant">খ্রিঃপূঃ</era>
 						<era type="1">খৃষ্টাব্দ</era>
 						<era type="1" alt="variant">খ্রিষ্টাব্দ</era>
 					</eraAbbr>
--- a/common/main/kok.xml
+++ b/common/main/kok.xml
@@ -902,7 +902,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<type key="calendar" type="japanese" scope="core">जपानी</type>
 			<type key="calendar" type="persian">पर्शियन दिनदर्शिका</type>
 			<type key="calendar" type="persian" scope="core">पर्शियन</type>
-			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका (अणकाराची कुरू: जाका चिनी दिनदर्शिकेचें प्रजासत्ताक”, “रिपब्लिकन दिनदर्शिका“)</type>
+			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका</type>
 			<type key="calendar" type="roc" scope="core">मिंगुआ</type>
 			<type key="cf" type="account">लेखा चलन स्वरूप</type>
 			<type key="cf" type="account" scope="core">लेखा चलन</type>
--- a/common/main/kok.xml
+++ b/common/main/kok.xml
@@ -902,7 +902,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<type key="calendar" type="japanese" scope="core">जपानी</type>
 			<type key="calendar" type="persian">पर्शियन दिनदर्शिका</type>
 			<type key="calendar" type="persian" scope="core">पर्शियन</type>
-			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका (अणकाराची कुरू: जाका चिनी दिनदर्शिकेचें प्रजासत्ताक&quot;, &quot;रिपब्लिकन दिनदर्शिका&quot;)</type>
+			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका (अणकाराची कुरू: जाका चिनी दिनदर्शिकेचें प्रजासत्ताक”, “रिपब्लिकन दिनदर्शिका“)</type>
 			<type key="calendar" type="roc" scope="core">मिंगुआ</type>
 			<type key="cf" type="account">लेखा चलन स्वरूप</type>
 			<type key="cf" type="account" scope="core">लेखा चलन</type>
--- a/common/main/oc.xml
+++ b/common/main/oc.xml
@@ -4189,7 +4189,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<perMille draft="unconfirmed">↑↑↑</perMille>
 			<infinity draft="unconfirmed">↑↑↑</infinity>
 			<nan draft="unconfirmed">↑↑↑</nan>
-			<timeSeparator draft="unconfirmed">‘h’</timeSeparator>
+			<timeSeparator draft="unconfirmed">h</timeSeparator>
 		</symbols>
 		<symbols numberSystem="lepc">
 			<decimal draft="unconfirmed">↑↑↑</decimal>
--- a/common/main/oc.xml
+++ b/common/main/oc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2025 Unicode, Inc.
+<!-- Copyright © 1991-2026 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-3.0
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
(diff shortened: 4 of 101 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -532,9 +533,16 @@ private String processInputMore(String path, String value) {
...
+        if (path.startsWith("//ldml/units")) {
+            if (path.contains("length-foot") || path.contains("angle-arc-minute")) {
+                // Normalize single apostrophes
+                value = value.replace("'", "′");
+            } else if (path.contains("length-inch")
+                    || path.contains("angle-arc-second")
+                    || path.contains("pressure-inch-ofhg")) {
+                // Normalize double apostrophes
+                value = value.replace("''", "″").replace("\"", "″");
+            }
```
srl295, line 549: why is this so specific?

srl295 (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
robertbastian: as opposed to?

srl295: applying to every unit not just specific ones?

robertbastian: we only want to do this particular replacement for units where it makes sense, for other units these should be treated like for all other values (which is to add a warning in the future)

common/main/bn.xml
```diff
@@ -2099,7 +2099,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 					</eraNames>
 					<eraAbbr>
 						<era type="0">খ্রিস্টপূর্ব</era>
-						<era type="0" alt="variant">&quot;খ্রিঃপূঃ&quot;</era>
+						<era type="0" alt="variant">“খ্রিঃপূঃ“</era>
```
robertbastian: probably want to change this?

```suggestion
						<era type="0" alt="variant">খ্রিঃপূঃ</era>
```

common/main/oc.xml
```diff
@@ -4189,7 +4189,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<perMille draft="unconfirmed">↑↑↑</perMille>
 			<infinity draft="unconfirmed">↑↑↑</infinity>
 			<nan draft="unconfirmed">↑↑↑</nan>
-			<timeSeparator draft="unconfirmed">'h'</timeSeparator>
+			<timeSeparator draft="unconfirmed">‘h’</timeSeparator>
```
robertbastian: and this. this is not a pattern, so literals don't have to be quoted
```suggestion
			<timeSeparator draft="unconfirmed">h</timeSeparator>
```

common/main/kok.xml
```diff
@@ -902,7 +902,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<type key="calendar" type="japanese" scope="core">जपानी</type>
 			<type key="calendar" type="persian">पर्शियन दिनदर्शिका</type>
 			<type key="calendar" type="persian" scope="core">पर्शियन</type>
-			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका (अणकाराची कुरू: जाका चिनी दिनदर्शिकेचें प्रजासत्ताक&quot;, &quot;रिपब्लिकन दिनदर्शिका&quot;)</type>
+			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका (अणकाराची कुरू: जाका चिनी दिनदर्शिकेचें प्रजासत्ताक”, “रिपब्लिकन दिनदर्शिका“)</type>
```
robertbastian: according to Google Translate, this is 

`Mingua Calendar (translation key: Jaka Republic of Chinese Calendar”, “Republican Calendar“)`

so the quotes aren't matching, and it contains unnecessary data. I propose we trim this


```suggestion
			<type key="calendar" type="roc">मिंगुआ दिनदर्शिका</type>
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
robertbastian: *which is to rewrite to typographic apostrophes/quotes, and maybe add a warning for quotes in the future

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
```diff
@@ -535,8 +532,12 @@ private String processInputMore(String path, String value) {
 
         if (path.startsWith("//ldml/units")) {
             if (path.contains("length-foot") || path.contains("angle-arc-minute")) {
-                // Normalize single apostrophes
-                value = value.replace("'", "′");
+                // Normalize single apostrophes at the end of the string.
+                // This avoids interfering with actual apostrophes ("d'arc"), which
+                // are replaced later.
+                if (value.indexOf("'") == value.length() - 1) {
+                    value = value.replace("'", "′");
```
macchiati: // Convert single apostrophes at the end of the string to primes for certain units.
...

tools/cldr-code/src/main/java/org/unicode/cldr/test/DisplayAndInputProcessor.java
macchiati: Certain units have common abbreviations as single or double primes (and should be neither ASCII ' nor apostrophe [apostrophe](https://util.unicode.org/UnicodeJsps/character.jsp?a=2019))

robertbastian:
@macchiati can you respond to my comments?

common/main/af.xml
```diff
@@ -293,7 +293,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<language type="mfe">Morisjen</language>
 			<language type="mg">Malgassies</language>
 			<language type="mgh">Makhuwa-Meetto</language>
-			<language type="mgo">Meta’</language>
+			<language type="mgo">Metaʼ</language>
```
macchiati: Why are we changing

 ’ 	U+2019	RIGHT SINGLE QUOTATION MARK - which is the apostrophe *punctuation mark*

to 

ʼ 	U+02BC	MODIFIER LETTER APOSTROPHE - which is an apostrophe *letter*

This should only happen if _Afrikaans_ favors the apostrophe letter instead of the punctuation mark. That is we go by the preferences of the locale, not of the language it is giving a name to.

macchiati (changes requested):

common/main/af.xml
robertbastian: `mgo` favours the apostrophe letter instead of the punctuation mark, so that should be respected in its names. This was previously done for only one of the apostrophe-letter-using languages (`mic`), but I think we should do it for all or none.

macchiati:
> Don't exclude number symbols from apostrophe handling (affects Swiss thousand separator)

We do want to be careful with these, since some systems use the format for parsing. Probably best to leave this for a new, separate ticket.

common/main/af.xml
robertbastian: actually, this wasn't done previously, but the display names for `mic` were explicitly excluded from quote replacement, and were using U+0027 APOSTROPHE.

common/main/bn.xml
macchiati: Hmmm. This really looks like the quotes were just a mistake by vetters. I think we should remove them completely, and make a forum post on that item that the quotes were removed.

common/main/oc.xml
macchiati: Agreed.

common/main/kok.xml
macchiati: Right call!

common/main/shn.xml
```diff
@@ -16963,7 +16963,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			</unit>
 			<unit type="pressure-hectopascal">
 				<displayName>ႁႅၵ်ႇတူဝ်ႇပၢတ်ႉသၵႄးလ်</displayName>
-				<unitPattern count="other">{0} ႁႅၵ်ႇတူဝ်ႇပၢတ်ႉသၵႄးလ်'</unitPattern>
+				<unitPattern count="other">{0} ႁႅၵ်ႇတူဝ်ႇပၢတ်ႉသၵႄးလ်‘</unitPattern>
```
robertbastian, line 16966: not sure this is correct

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
```diff
@@ -584,9 +584,64 @@ public CheckCLDR handleCheck(
...
     }
 
+    private void checkQuotes(String value, List<CheckStatus> result) {
+        int a = 0;
+        int b = 0;
+        int c = 0;
+        int d = 0;
+        int e = 0;
+        for (var i = 0; i < value.length(); i++) {
+            int cp = value.codePointAt(i);
```
macchiati:
```suggestion
            int cp = value.codePointAt(i);
            // this is ok because none of the characters are >FFFFf
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
```diff
@@ -584,9 +584,64 @@ public CheckCLDR handleCheck(
...
+            if (cp == '\u201c') {
```
macchiati: This looks nice, but won't quite work. That is, some languages have odd behavior for quotation marks. For examples, see https://www.unicode.org/cldr/charts/48/by_type/core_data.alphabetic_information.html#Quotation_Marks

What should be done is to get the quotation characters for the locale being tested, start and end, and embedded start and end.

We don't have to test for overlap, eg start...embedded start...end...embedded end. So you can just have 2 booleans 

startMain
startEmbedded

So just for the start/end case (the embedded start/end are parallel) you can have a simple loop, something like:

if  start==end, just count and make sure you have an even number.

Otherwise:
- If you hit a 'start'
    - if startMain, stop with error (
    - ow startMain=true
- If you hit an 'end'
    - if !startMain, stop with error
    - ow set startMain=false
If you fall out of the loop, and startMain, you have an error

I think you'd also have to use a slightly different loop if start/end contains one or more of embedded start/end

macchiati (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
robertbastian: what's wrong with testing that all types of quotation marks are balanced? if the locale doesn't use a type, 0 and 0 is balanced as well.

this already discovered an issue in Hebrew where a quotation mark was used instead of a gershayim. I don't know if this would have been caught if we only looked at hebrew quotation marks.

the current approach doesn't have false positives. sure it has some false negatives, but I don't think that's worth blocking this PR on.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
robertbastian: aren't they? how do I do this correctly in Java?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
robertbastian: ok I've just removed this check because I don't want to block this PR on it

macchiati: Removing is good for now. The check will be valuable, but we can do under another PR, later.

For that time:

This is independent from the handling of ASCII ". For that, we can have a different check that there is just an even number of those. 

For all of these tests, I think we already exclude certain paths, like exemplars, that need to allow unpaired items.

macchiati (changes requested):
I could have sworn I made a comment to the following effect, but I can't find it.

We should only be changing ' to the _modifier letter apostrophe_ in locales that have that in their exemplar characters (main or auxiliary). That is a small number of locales, and does not include many of the locales you've made changes in. In other locales, they should change to the _right curly apostrophe_ (right single quote).

Those locales are: blo, br, bss, cad, cch, cic, gn, ha, ie, kek, lkt, mgo, mic, moh, mus, nnh, qu, quc, trv, uk, uz. The DAIP code, however, should get the exemplar sets and set a flag for whether the  _modifier letter apostrophe_ is in them, then convert on that basis.

robertbastian:
> We should only be changing ' to the modifier letter apostrophe in locales that have that in their exemplar characters (main or auxiliary). That is a small number of locales, and does not include many of the locales you've made changes in. In other locales, they should change to the right curly apostrophe (right single quote).

I don't think that is correct. Even in other locales, the apostrophe still represents a letter, not punctuation, and `U+2019 RIGHT SINGLE QUOTATION MARK` will not behave correctly under e.g. linkification.

I think it's too strict to apply exemplar characters to *names* in this form. If you want to disallow `U+02BC MODIFIER LETTER APOSTROPHE` in English, the name for Miʼkmaq should be Mikmaq – we should not insert punctuation. However, I don't see the issue of using the correct Unicode letter in English.

There's an issue CLDR-11219 in which someone asks whether to use `U+02BC MODIFIER LETTER APOSTROPHE` for the English name of Miʼkmaq. I don't know why the issue was closed, it doesn't seem to have been resolved.

I reverted all changes to language names to not block on this discussion.

robertbastian:
In fact, for `en`, we use `U+02BC MODIFIER LETTER APOSTROPHE` for language names of these languages, even though `en` does not have that letter in its exemplar set:

https://github.com/unicode-org/cldr/blob/fb0b4f0cb809cac10e8539dcba669c1d27d8e70c/common/main/en.xml#L256

https://github.com/unicode-org/cldr/blob/fb0b4f0cb809cac10e8539dcba669c1d27d8e70c/common/main/en.xml#L318

https://github.com/unicode-org/cldr/blob/fb0b4f0cb809cac10e8539dcba669c1d27d8e70c/common/main/en.xml#L359

https://github.com/unicode-org/cldr/blob/fb0b4f0cb809cac10e8539dcba669c1d27d8e70c/common/main/en.xml#L414

https://github.com/unicode-org/cldr/blob/fb0b4f0cb809cac10e8539dcba669c1d27d8e70c/common/main/en.xml#L520

My code would have extended this to other locales as well.

macchiati (approved):
Ok, we can discuss further in the design meeting as to a recommendation for the policy for the modifier letter apostrophe.

PR #5650, opened 2026-04-28
CLDR-19421 reduce key types

srl295:
CLDR-19421
- Removes items from ExtraPaths as requested.
- Note that numbers/thai (noncore) for example are in en, so will show up. 

image

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
@@ -346,13 +346,24 @@ public class ExtraPaths {
 
                     final String typeKeyTypePath =
                             typeKeyPath + String.format("[@type=\"%s\"]", type);
-                    pathsTemp.add(typeKeyTypePath);
+                    // add unless skipped
+                    if (!skipNoncoreType(originalKey, t)) {
+                        pathsTemp.add(typeKeyTypePath);
+                    }
                     // add some of these with [@scope="core"]
                     if (!skipCoreType(k, t)) {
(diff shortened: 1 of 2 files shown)


AEApple:
I'd like to do a push to production. Should I wait for this change?

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
```diff
@@ -382,7 +394,31 @@ private void addBcp47Keys() {
                         "s0",
                         "t0");
 
-        static final Pattern MATCH_LOWERCASE_SCRIPT_CODE = PatternCache.get("^[a-z]{4}$");
+        @SuppressWarnings("null")
+        private static final Set<Pair<String, String>> skipNoncoreKeyType =
```
macchiati, line 398: Very minor. This would be a bit cleaner as a multimap, and avoiding the Pair creation

Containment is determined by calling [containsEntry](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimap.html#containsEntry-java.lang.Object-java.lang.Object-)

That being said, this works fine, so approving.

macchiati (approved):

macchiati:
I'll go ahead and merge so you can push in a bit, @AEApple

tools/cldr-code/src/main/java/org/unicode/cldr/util/ExtraPaths.java
srl295: Yes, I thought about that, also remember this is only used once, so I went for a little more simple than optimal

PR #5652, opened 2026-04-28
CLDR-19419 report ordinals

srl295:
- refactor the format function

CLDR-19419

ALLOW_MANY_COMMITS=true

image


macchiati:
Steven, looks good. But can you add an English screenshot, so that we can see it with working data?

macchiati (approved):

srl295:
> Steven, looks good. But can you add an English screenshot, so that we can see it with working data?

English is the left column. but

image

PR #5653, opened 2026-04-28
CLDR-14921 ui: welcome page

srl295:
- add WelcomeWidget loading from welcome.md
- add Welcome to main menu (= locales list)
- click on overlay always hides it
- added some documentation on internal cldrLoad functions

CLDR-14921

ALLOW_MANY_COMMITS=true

#### First Load

image

#### Click Show Locales button

margin makes the welcome still visible (clicking on the gray hides it)

image

#### Using the menu to get back to Welcome

image

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrLocales.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrLocales.mjs
@@ -1,16 +1,21 @@
 /*
  * cldrLocales: encapsulate functions concerning locales for Survey Tool
+ * It's also the "special" named "locales", which happens to be the default page
+ * users see when they login.
  */
 import * as cldrAjax from "./cldrAjax.mjs";
+import * as cldrDom from "./cldrDom.mjs";
 import * as cldrEvent from "./cldrEvent.mjs";
 import * as cldrLoad from "./cldrLoad.mjs";
 import { LocaleMap } from "./cldrLocaleMap.mjs";
-import * as cldrMenu from "./cldrMenu.mjs";
 import * as cldrNotify from "./cldrNotify.mjs";
 import * as cldrRetry from "./cldrRetry.mjs";
 import * as cldrStatus from "./cldrStatus.mjs";
 import * as cldrSurvey from "./cldrSurvey.mjs";
 import * as cldrText from "./cldrText.mjs";
+import * as cldrVue from "./cldrVue.mjs";
+
+import WelcomeWidget from "../views/WelcomeWidget.vue";
 
 // The current locale may be "USER" temporarily, meaning the back end should choose an appropriate locale
 // for the current user. The real locale ID should be set when a server response contains it.
@@ -47,6 +52,9 @@ function setMap(json) {
   }
 }
 
+/** App of the previous app - so we can unmount it */
+let welcomePageApp = null;
+
 // called as special.load
 function load() {
   cldrSurvey.hideLoader();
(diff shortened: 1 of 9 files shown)


tools/cldr-apps/js/src/md/welcome.md
srl295, line 1: Yes, it's just markdown.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
```diff
@@ -47,32 +52,33 @@ function setMap(json) {
...
-    const topLoc = locmap.locmap.topLocales[n];
-    cldrMenu.addTopLocale(topLoc, theDiv);
+  if (welcomePageApp != null) {
+    try {
+      welcomePageApp.unmount();
+    } catch (e) {
+      // ignore unmount error
+    }
   }
-  cldrLoad.flipToOtherDiv(null);
```
srl295, line 65: @btangmu The comment may be a misunderstanding here. 

The code here actually creates a "main page" that looks like this  (it's the "very old" way of navigating to a locale, before the sidebar ws added.) 

Image

… then, `flipToOtherDiv(null)` ignores all of that and shows a blank page instead.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
```diff
@@ -47,32 +52,33 @@ function setMap(json) {
...
   cldrEvent.filterAllLocale(); // filter for init data
   cldrEvent.forceSidebar();
   cldrStatus.setCurrentLocale(null);
-
-  // When clicking on the locale name in the header of the main Page view,
-  // the OtherSection div may be non-empty and needs to be hidden here
-  const otherSection = document.getElementById("OtherSection");
-  if (otherSection) {
-    otherSection.style.display = "none";
-  }
```
srl295, line 75: we actually want to show this now, because it's the welcome page.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
srl295: But, this code didn't do any productive work related to cldrMenu.mjs - it was for an entirely different homepage.

addTopLocale etc is now dead code.

macchiati:
images look like what we want, thanks!

macchiati:
Tom should review the js code

tools/cldr-apps/js/src/views/WelcomeWidget.vue
```diff
@@ -0,0 +1,32 @@
...
+    <hr />
+    <a-button @click="showSidebar">Show Locales…</a-button>
+  </div>
+</template>
+
+<script setup>
+// Note: SignCla.vue uses a similar pattern.
+// We could consider a common component for both.
+import { marked } from "../esm/cldrMarked.mjs";
+import * as cldrEvent from "../esm/cldrEvent.mjs";
```
btangmu, line 14: Minor: I'd rather see these alphabetized, and `import *` for both. It makes it easier to see where/how the module is used by searching for `cldrMarked`, and consistency in general makes code easier to parse both visually and with tools.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
btangmu: Is the dead code removed in this PR?

btangmu (approved):
If we had more time, I'd like to see this in action on cldr-staging or locally, since there might be unintended side effects. Given the time constraint, maybe it's best just to test this as thoroughly as possible on cldr-smoke before pushing to production. Ideally, all the places where special.load is called should be tested... which might mean for all "specials"

tools/cldr-apps/js/src/views/WelcomeWidget.vue
srl295: Thanks. I'll reorder in a followup.

tools/cldr-apps/js/src/esm/cldrLocales.mjs
srl295: no, it may not actually be dead (but it should be).

It was a dead store before.

btangmu: It looks like the answer is no, since addTopLocale is in cldrMenu.mjs which isn't changed in this PR; removal could be in a follow-up

srl295: right. it's actually called in another place during menu loading. Probably also to no useful side effect, but should be checked.

srl295:
> If we had more time, I'd like to see this in action on cldr-staging or locally, since there might be unintended side effects. Given the time constraint, maybe it's best just to test this as thoroughly as possible on cldr-smoke before pushing to production. Ideally, all the places where special.load is called should be tested... which might mean for all "specials"

I'm not sure what effect you mean though, because I didn't change special.load, just added documentation.

btangmu:
> didn't change special.load

Well, there is a change to `cldrLocales.load`, which I think is called as `special.load`, but I was confused/wrong in thinking it could affect specials other than cldrLocale. Good!

srl295:
> > didn't change special.load
> 
> Well, there is a change to `cldrLocales.load`, which I think is called as `special.load`, but I was confused/wrong in thinking it could affect specials other than cldrLocale. Good!

Right.

PR #5654, opened 2026-04-28
CLDR-19362 Make gmtZeroFormat match gmtFormat

robertbastian:
CLDR-19362

ALLOW_MANY_COMMITS=true


srl295 (dismissed):
LGTM.

macchiati:
gmtZeroFormat needs to have the same structure as it had in v47. 

Any code that used that needs to be able to use it as is, _without supplying a parameter._ That won't happen if the pattern is not constant.

So I think the actual solution might be simpler; and focus on the known problem. 

* Drop this PR.
* Make sure gmtZeroFormat does not exist in any locale (which might take a different PR)
* Document that for compatibility, implementations should use the gmtZeroFormat for offset 0 _IFF_ it exists, and otherwise use the regular gmtFormat.
* Fix ICU/ICU4X to do that.
* Document if implementations need to restore the old behavior for English or other locales, they can patch CLDR to _add_ the v47 value back for any locales they are concerned about. (and copy or point to that in ICU/ICU4X documentation)

The other alternative would be to generate parameter-less gmtZeroFormat values for all the locales, by taking the gmtFormat with the parameter 0 to generate constant values for each locale, of the form "GTM+0" or "UTC+0" or ... with the default number format for the locale. That has the downside we still need to do the above. But that is 

1. More work to get right
2. Has the disadvantage that in locales that need to allow switching number systems (like Arabic), the constant gmtZeroFormat value wouldn't align.

macchiati (changes requested):
See other comment

robertbastian:
so if I call it `gmtZeroFormat2` that's ok?



> * Document that for compatibility, implementations should use the gmtZeroFormat for offset 0 _IFF_ it exists, and otherwise use the regular gmtFormat.

Can we easily do "IFF it exists"? It will fall back, no?

> The other alternative would be to generate parameter-less gmtZeroFormat values for all the locales, by taking the gmtFormat with the parameter 0 to generate constant values for each locale, of the form "GTM+0" or "UTC+0" or ... with the default number format for the locale. That has the downside we still need to do the above. But that is

That doesn't work because the pattern is used for both the long and short format

macchiati:
Let's discuss in person (well, in video)

macchiati:
This isn't a blocker for Submission, so we have a bit of time. Let's add to the design meeting agenda.

srl295 (dismissed):

macchiati:
There is a recommendation from the Design WG for this; we should get approval Wednesday morning before landing.

macchiati:
Looks good to me, except for the build failure.

BTW, I was thinking about the tooling. It is very easy to add a CLDRModify option to copy one path to another; that could be the simplest option. I can walk you through that if you want.

robertbastian:
This doesn't seem to be needed after all, see the ticket

PR #5657, opened 2026-04-28
CLDR-19394 off/on: stub out data, dtd, spec for off/on

srl295:
CLDR-19394

I'd like to bring in provisional seed data separately, so that we can try the UI on smoketest.

image

ALLOW_MANY_COMMITS=true

image

image

### DIFF
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -1572,6 +1572,9 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='calendar'][@type='%calendarType80'][@scope='%A']"/>
 		<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='calendar'][@type='%calendarType80']"/>
 
+		<!-- type values-->
+		<coverageLevel value="modern" match="localeDisplayNames/typeValues/typeValue[@type='%yesNo']"/>
+
 		<coverageLevel value="modern" match="characterLabels/characterLabel[@type='%anyAttribute']"/>
 		<coverageLevel value="modern" match="characterLabels/characterLabelPattern[@type='%anyAttribute']"/>
 		<coverageLevel value="modern" match="characterLabels/characterLabelPattern[@type='%anyAttribute'][@count='%allPlurals']"/>
--- a/docs/ldml/tr35-general.md
+++ b/docs/ldml/tr35-general.md
@@ -497,6 +497,20 @@ Note that the subdivision names are in separate files, in the subdivisions/ dire
 
 See also **Part 6** _[Subdivision Containment](tr35-info.md#Subdivision_Containment)_.
 
+### Type Values
+
+Many BCP47 types have boolean values, such as `ka` (`colAlternate`).
+Rather than track translations for each of these separately, the `<typeValues>` element provides a centralized translation for this particular purpose.
+
+```xml
+<localeDisplayNames>
+    <typeValues>
+        <typeValue type="no">Off</typeValue>
+        <typeValue type="yes">On</typeValue>
+    </typeValues>
+</localeDisplayNames>
+```
+
 ## <a name="Layout_Elements" href="#Layout_Elements">Layout Elements</a>
 
 ```xml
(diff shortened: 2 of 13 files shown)


common/supplemental/coverageLevels.xml
```diff
@@ -1572,6 +1572,9 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='calendar'][@type='%calendarType80'][@scope='%A']"/>
 		<coverageLevel value="modern" match="localeDisplayNames/types/type[@key='calendar'][@type='%calendarType80']"/>
 
+		<!-- type values-->
+		<coverageLevel value="modern" match="localeDisplayNames/typeValues/typeValue[@type='%yesNo']"/>
```
macchiati, line 1576: For the future, %A is better. I'll capture this on the how-to page.

docs/ldml/tr35-general.md
```diff
@@ -497,6 +497,28 @@ Note that the subdivision names are in separate files, in the subdivisions/ dire
...
+</localeDisplayNames>
+```
+
+These are intended to be used with key type names such as:
+
+* Ignore Symbols Sorting: **On**
+* Reversed Accent Sorting: **Off**
+* Uppercase/Lowercase Ordering: **On**
+
+These two strings are not inflected.
```
macchiati, line 520:
```suggestion
#### Guidelines

Choose the best terms that indicate that an option is active or not in a menu.
The values are not inflected according to the option, so choose the form that is most generally applicable.
Look at the "ss" values for comparison.
```

docs/site/translation/displaynames/locale-option-names-key.md
```diff
@@ -43,7 +43,22 @@ Here are examples of names of Options to be translated.
...
+
+For more details, see [On Off].
+
+Rather than translate many different similar phrases , that is **Do Ignore Symbols Sorting / Don’t Ignore Symbols Sorting**, there are simply values for **On** and **Off** which are used together with these. Examples of how these are used:
+
+* Ignore Symbols Sorting: **On**
+* Reversed Accent Sorting: **Off**
+* Uppercase/Lowercase Ordering: **On**
+
+These two string are not inflected.
```
macchiati, line 58: Reword like above.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -3837,6 +3840,32 @@ private void handleDisplayNames(
...
             }
+        } else if (parts.getElement(-1).equals("typeValue")) {
+            final String COL_ALTERNATE =
+                    "//ldml/localeDisplayNames/keys/key[@type=\"colAlternate\"]";
+            String type = parts.getAttributeValue(-1, "type");
+            String exampleValue = getCldrFile().getStringValue(COL_ALTERNATE);
+            if (exampleValue == null || exampleValue.isEmpty()) {
+                exampleValue = englishFile.getStringValue(COL_ALTERNATE);
+            }
+            if (YES_NO.contains(type)) {
```
macchiati, line 3851: We could add examples for each of the applicable keys; that would help people to pick the most general form.

macchiati (approved):
There are a couple of small items here. Approving in the interests of time, but if merged, we should have a small followup PR.

common/supplemental/coverageLevels.xml
srl295: Why is %A better?

docs/ldml/tr35-general.md
srl295: is this more applicable for the infohub vs. the spec?

docs/site/translation/displaynames/locale-option-names-key.md
srl295:
```diff
+#### Guidelines
+
+Choose the best terms that indicate that an option is active or not in a menu.
+The values are not inflected according to the option, so choose the form that is most generally applicable.
+Look at the "ss" values for comparison.
```

Once 'ss' is copied over, that won't be applicable (in fact we may hide "ss")

PR #5660, opened 2026-04-29
CLDR-19226 Improve spec for append items and implement in CldrDTPG

sffc:
CLDR-19226

CLDR-19406

There is no testdata diff, but I am adding more testdata in #5660 which surfaces a diff across the new append items.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/ldml/tr35-dates.md
+++ b/docs/ldml/tr35-dates.md
@@ -834,9 +834,15 @@ Finally: If the requested skeleton included both seconds and fractional seconds 
 If a client-requested set of fields includes both date and time fields, and if the `availableFormats` data does not include a `dateFormatItem` whose skeleton matches the same set of fields, then the request should be handled as follows:
 
 1. Divide the request into a date fields part and a time fields part.
+    * Date fields are: [year](#dfst-year), [month](#dfst-month), [day](#dfst-day), [era](#dfst-era), [week](#dfst-week), [quarter](#dfst-quarter), and [week day](#dfst-weekday).
+    * Time fields are: [hour](#dfst-hour), [minute](#dfst-minute), [second](#dfst-second), [period](#dfst-period), and [zone](#dfst-zone).
 2. For each part, find the matching `dateFormatItem`, and expand the pattern as above.
-3. Combine the patterns for the two `dateFormatItem`s using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
-   * If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
+    * If there is still no `dateFormatItem` whose skeleton matches the same set of fields, select the one with the greatest number of matching fields (but no extra fields), then use `appendItems` to append any missing fields (see below).
+    * If multiple `dateFormatItem`s with missing fields have the same distance, rank them by their matching fields in the order listed in step 1. For example, if the request is for "HBv", and the locale has `dateFormatItem`s for only "HB" and "Hv", select the "HB" pattern, because "B" has a higher weight than "v", and then use the `appendItem` for "v" (time zone).
+3. Combine the patterns for the two `dateFormatItem`s using the appropriate glue pattern, determined as follows from the requested date fields:
+   * If the date fields part contains *only* a weekday, use `<appendItem request="Time-Day-Of-Week">`.
+   * Otherwise, if the time fields part contains *only* a time zone, use `<appendItem request="Date-Timezone">`.
+   * Otherwise, if the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
    * Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
    * Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`
    * Otherwise use `<dateTimeFormatLength type="short">`
@@ -848,6 +854,8 @@ If a client-requested set of fields includes both date and time fields, and if t
 ```
 
 In case the best match does not include all the requested calendar fields, the `appendItems` element describes how to append needed fields to one of the existing formats. Each `appendItem` element covers a single calendar field. In the pattern, {0} represents the format string, {1} the data content of the field, and {2} the display name of the field (see [Calendar Fields](#Calendar_Fields)).
+
+Note: as described above `appendItems` for date fields should be appended to the date, and `appendItems` for time fields should be appended to the time, _before_ combining them with the `dateTimeFormat`.
 
 #### <a name="intervalFormats" href="#intervalFormats">Element intervalFormats</a>
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
@@ -432,6 +432,14 @@ public class CldrDateTimePatternGenerator {
         String timeSkeleton = getTimeSkeleton(canonicalSkeleton);
 
         if (dateSkeleton.length() > 0 && timeSkeleton.length() > 0) {
+            if (isOnlyDayOfWeek(dateSkeleton)) {
+                String timePattern = getBestPattern(timeSkeleton, log);
+                return appendFieldWithCustomName(timePattern, dateSkeleton, "Time-Day-Of-Week");
+            }
+            if (isOnlyZone(timeSkeleton)) {
+                String datePattern = getBestPattern(dateSkeleton, log);
+                return appendFieldWithCustomName(datePattern, timeSkeleton, "Date-Timezone");
+            }
             return combineDateAndTime(dateSkeleton, timeSkeleton, log);
         }
 
@@ -564,6 +572,11 @@ public class CldrDateTimePatternGenerator {
     private String appendField(String pattern, String field) {
         char firstChar = field.charAt(0);
         String requestName = getAppendRequestName(firstChar);
+        return appendFieldWithCustomName(pattern, field, requestName);
+    }
+
+    private String appendFieldWithCustomName(String pattern, String field, String requestName) {
+        char firstChar = field.charAt(0);
         String appendFormat = appendItems.get(requestName);
         if (appendFormat == null) {
             appendFormat = "{0} \u251c{2}: {1}\u2524";
@@ -577,6 +590,22 @@ public class CldrDateTimePatternGenerator {
                 .replace("{0}", pattern)
                 .replace("{1}", firstFieldPattern)
                 .replace("{2}", quote(fieldDisplayName));
+    }
+
+    private boolean isOnlyZone(String skeleton) {
+        for (int i = 0; i < skeleton.length(); i++) {
+            char c = skeleton.charAt(i);
+            if ("zZOvVXx".indexOf(c) < 0) return false;
+        }
+        return true;
+    }
+
+    private boolean isOnlyDayOfWeek(String skeleton) {
+        for (int i = 0; i < skeleton.length(); i++) {
+            char c = skeleton.charAt(i);
+            if ("Eec".indexOf(c) < 0) return false;
+        }
+        return true;
     }
 
     private String quote(String s) {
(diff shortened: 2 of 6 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
```diff
@@ -432,6 +432,14 @@ private String getBestPatternInternal(String skeleton, List<String> log) {
...
 
         if (dateSkeleton.length() > 0 && timeSkeleton.length() > 0) {
+            if (isOnlyZone(timeSkeleton)) {
+                String datePattern = getBestPattern(dateSkeleton, log);
+                return appendFieldWithCustomName(datePattern, timeSkeleton, "Date-Timezone");
+            }
+            if (isOnlyDayOfWeek(dateSkeleton)) {
+                String timePattern = getBestPattern(timeSkeleton, log);
+                return appendFieldWithCustomName(timePattern, dateSkeleton, "Time-Day-Of-Week");
+            }
```
sffc: Note: There is an order of operations requirement here. If the skeleton is "Ev", do we use Date-Timezone or Time-Day-Of-Week? It should be in the spec I guess?

docs/ldml/tr35-dates.md
```diff
@@ -834,12 +834,17 @@ Finally: If the requested skeleton included both seconds and fractional seconds
 If a client-requested set of fields includes both date and time fields, and if the `availableFormats` data does not include a `dateFormatItem` whose skeleton matches the same set of fields, then the request should be handled as follows:
 
 1. Divide the request into a date fields part and a time fields part.
+    * Date fields are: [year](#dfst-year), [month](#dfst-month), [day](#dfst-day), [era](#dfst-era), [week](#dfst-week), [quarter](#dfst-quarter), and [week day](#dfst-weekday).
+    * Time fields are: [hour](#dfst-hour), [minute](#dfst-minute), [second](#dfst-second), [period](#dfst-period), and [zone](#dfst-zone).
```
macchiati, line 838: However, if there are no date fields other than the [week day](#dfst-weekday), then the weekday is included with the time fields.

docs/ldml/tr35-dates.md
```diff
@@ -834,12 +834,17 @@ Finally: If the requested skeleton included both seconds and fractional seconds
 If a client-requested set of fields includes both date and time fields, and if the `availableFormats` data does not include a `dateFormatItem` whose skeleton matches the same set of fields, then the request should be handled as follows:
 
 1. Divide the request into a date fields part and a time fields part.
+    * Date fields are: [year](#dfst-year), [month](#dfst-month), [day](#dfst-day), [era](#dfst-era), [week](#dfst-week), [quarter](#dfst-quarter), and [week day](#dfst-weekday).
```
macchiati, line 837: However, if there are no time fields other than the [zone](#dfst-zone), then the zone is included with the date fields.

docs/ldml/tr35-dates.md
```diff
@@ -834,12 +834,17 @@ Finally: If the requested skeleton included both seconds and fractional seconds
...
 3. Combine the patterns for the two `dateFormatItem`s using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
    * If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
    * Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
    * Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`
    * Otherwise use `<dateTimeFormatLength type="short">`
+   * If the time fields part contains *only* a time zone, use the `Date-Timezone` append item instead of splitting the parts and combining with the `dateTimeFormat`. Similar with `Time-Day-Of-Week` (if the date fields part contains only a day-of-week).
```
macchiati: I think this needs some rework.  The day-of-week can and does occur with either date or time. And (in theory) also the timezone.

Here is my sketch for the order of operations. (roughly speaking, need to flesh out any exceptional cases).

1. lookup the date (+ timezone if there are no other time fields) and time (+ dow if there are no other date fields.
2. If either is not successful, use the appendItems to included missing fields where necessary.
3. Only then do we combine as needed.
* If there is only the date pattern, that is the result
* If there is only the time pattern, then that is the result
* Otherwise the result is the combination of the two using `dateFormatItem`s.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
macchiati: Definitely an edge case! But we should not be ambiguous. I think the Date-Timezone is the best.

macchiati: I'm not sure your code reflects the ordering I was looking at above.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
```diff
@@ -860,35 +889,20 @@ private List<String> splitSkeleton(String skel) {
...
-                    int[] lengths2 = lengthsMap.get(e2.getKey());
-
-                    int limit = DateTimePatternGenerator.TYPE_LIMIT;
-                    for (int i = 0; i < limit; i++) {
-                        if (chars1[i] != chars2[i]) {
-                            return chars2[i] - chars1[i];
-                        }
-                        if (lengths1[i] != lengths2[i]) {
-                            return lengths2[i] - lengths1[i];
+                    String k1 = e1.getKey();
```
macchiati: The goal is to pick the id with the least distance from the desired skeleton, filtering to just ids that contain exactly the same fields (but with perhaps different lengths). Now, if we want to optimize, we would produce a multimap from a "field set" to skeletons that have exactly those fields. Then we'd only have to scan a small subset of all fields.

I'm not completely sure what is going on here. (don't have time right now for a full look). But I don't know why we are sorting the available formats, because we are dealing with multiple dimensions. For example, I can't find the closest state capital city to X location by pre-sorting the capital state capitals by any function.

macchiati (changes requested):

docs/ldml/tr35-dates.md
sffc: No. We still need to split off the weekday and assemble a well-formed time pattern before we append the weekday to it.

docs/ldml/tr35-dates.md
sffc: If there is any skeleton where the fields match, _we don't get to this section at all_. We _only_ reach this section if we are forced to use either a ~dateTimePattern~ dateTimeFormatLength or an appendItem.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
sffc: This edit is, I believe, a no-op change that makes the code simpler, because it is more explicit that we are sorting the skeletons by the canonical field order.

docs/ldml/tr35-dates.md
sffc: No. We still need to split off the zone and assemble a well-formed date pattern before we append the zone to it.

docs/ldml/tr35-dates.md
macchiati: Good point. #3 should be just  

\3. The result is the combination of the two using dateFormatItems.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
macchiati: That only has an effect on the choice where there are 2+ ids with equal distances to the desiredSkeleton, right?

sffc:
I tweaked Step 3 to make it more clear that `appendItem` here is used like a glue pattern.

I was going to make the spec more clear that this section is only reached after normal matching has failed, but I find it to already be clear.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
sffc: Right, this ordering function helps disambiguate cases like

Requested: `yMMMdE`
Available: `yMMMMdE` and `yMMMdEEEE`

In both cases, the distance is the same (a long field vs an abbreviated field), so the ordering here breaks ties.

docs/ldml/tr35-dates.md
macchiati: I'll have to get to the review of this later; looks like I need to review the whole context.

docs/ldml/tr35-dates.md
sffc: I think we had considered encoding the Date+Zone and Time+Weekday patterns as `dateTimeFormatLength`, but I asked for them to be encoded as `appendItem` because that is more flexible, but for spec purposes, I find it simpler and easier to follow if we treat them like glue patterns.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
sffc: I've written the spec to reflect the code, which is that Date-Timezone is preferred over Time-Day-Of-Week.

docs/ldml/tr35-dates.md
macchiati: I do not understand. We have patterns like 

`<dateFormatItem id="GyMMMEd">G y MMM d, E</dateFormatItem>`
and like:
`<dateFormatItem id="Ehm">E h:mm a</dateFormatItem>`

So:

- If there is a desired skeleton with GyMMMEd, we do not want to split off the E, format the rest, then append the E.
- 
- If there is a desired skeleton with Ehm, we do not want to split off the E, format the rest, then append the E.

So at a high level I think the breakdown would be the following:

- See if there is a pattern P for the desired skeleton S. If so, return it.
- If there are both date and time fields (excluding E), and there is an E
    - Try splitting into day fields and time fields, with any E going with the **date fields**. If those work, then return the combined resulting pattern
    - Try splitting into day fields and time fields, with any E going with the **time fields**. If those work, then combined the result 
- Otherwise split into day fields and time fields. ...

sffc: GyMMMEd and Ehm are reachable in the normal "Matching Skeletons" section. This PR only touches "Missing Skeleton Fields" which only happens if there is no skeleton that has the correct fields.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
macchiati: Yes, definitely. This brings up a point. It is probably better to favor months over others. Eg given a choice between 

Desired A. yMMMMdEEEE B. yMMMdEEE

1 yMMMdEEEE
2 yMMMMdEEE

We should match the closest month width for A and for B

sffc: Right, I believe that is what this code is implementing.

tools/cldr-code/src/main/java/org/unicode/cldr/util/CldrDateTimePatternGenerator.java
macchiati: I  think _3:00 Tuesday_ is common. And _Jan 10, Pacific Time_ is extremely rare. So given the choice, we should absolutely favor the former.

sffc: Sure, makes sense. I switched the order.

docs/ldml/tr35-dates.md
macchiati: ok

macchiati:
I think we can go ahead and merge this, once you amend the commit with the missing "CLDR-19226 " prefix.

macchiati (dismissed):

sffc:
I split the sorting change to https://github.com/unicode-org/cldr/pull/5809.

macchiati (approved):

PR #5664, opened 2026-04-29
CLDR-19394 on/off: copy from "ss", also improve CLDRModify

srl295:
CLDR-19394

What's in the box:
- Copying from `ss` On/Off to `typeValues` (as unconfirmed)

**Note** CLDRModify stuff schlepped off to another PR.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/ba.xml
+++ b/common/main/ba.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2025 Unicode, Inc.
+<!-- Copyright © 1991-2026 Unicode, Inc.
 For terms of use, see http://www.unicode.org/copyright.html
 SPDX-License-Identifier: Unicode-3.0
 CLDR data files are interpreted according to the LDML specification (http://unicode.org/reports/tr35/)
@@ -1151,6 +1151,10 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 			<type key="ss" type="none" scope="core">һүндерелгән</type>
 			<type key="ss" type="standard" scope="core">ҡабыҙылған</type>
 		</types>
+		<typeValues>
+			<typeValue type="no" draft="unconfirmed">һүндерелгән</typeValue>
+			<typeValue type="yes" draft="unconfirmed">ҡабыҙылған</typeValue>
+		</typeValues>
 		<measurementSystemNames>
 			<measurementSystemName type="metric">метрик</measurementSystemName>
 			<measurementSystemName type="UK">Бөйөк Британия</measurementSystemName>
--- a/common/main/af.xml
+++ b/common/main/af.xml
@@ -1071,6 +1071,10 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<type key="ss" type="none" scope="core">Af</type>
 			<type key="ss" type="standard" scope="core">Aan</type>
 		</types>
+		<typeValues>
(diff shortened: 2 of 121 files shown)


tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
```diff
@@ -1,3 +1,6 @@
 #locale;	action	path	new_value
-locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Samoa"]/long/generic
-locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Samoa"]/long/daylight
+# See: https://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-using-config-file
```
srl295: the link in CLDRModify's own docs was broken. Let's leave this here so we can find it…

common/main/ba.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!DOCTYPE ldml SYSTEM "../../common/dtd/ldml.dtd">
-<!-- Copyright © 1991-2025 Unicode, Inc.
+<!-- Copyright © 1991-2026 Unicode, Inc.
```
macchiati, line 3: BTW, at the start of each year, we should do a pass on all the data files to update the copyright date; that way we don't get date differences showing up in diffs.

srl295: Hm.  This only updates if the file actually changes. That does keep the file's copyright accurate.

docs/site/development/cldr-big-red-switch/cldrmodify-using-config-file.md
```diff
@@ -72,6 +74,18 @@ Same as add, but only for paths that don't have values in the original value.
 
 Replace \<path, value\> with \<newPath, newValue\>
 
+**action=copy**
```
macchiati: This option looks useful _for the case where regex replacement handles the changes_. However, I wish this had been in a separate PR, because I would have to look it over to make sure that there are not any subtle problems, and we want to get the data changes in.

I'm going to approve for now, but I think we want to disable the new options right after (or warn devs against using them) until we can do a full review.

macchiati (dismissed):

macchiati:
Hmmm. Needs a change anyway because of the test failure. I think there is an exception list for items like:

`es_419 [Latin American Spanish]	error	▸Locale_Display_Names|Keys|Type_Values|no◂	〈Off〉	【】	〈no〉	«=»	【】	⁅same as code⁆	❮Error: The value is the same as the 'code'❯	https://st.unicode.org/cldr-apps/v#/es_419//2f3689b454e651ff
Error:  Error: The value is the same as the 'code'`

docs/site/development/cldr-big-red-switch/cldrmodify-using-config-file.md
srl295: The regex replacement does handle the changes..

Anyway, I'll split the PR.

srl295:
> Hmmm. Needs a change anyway because of the test failure. I think there is an exception list for items like:
> 
> 
> 
> `es_419 [Latin American Spanish]	error	▸Locale_Display_Names|Keys|Type_Values|no◂	〈Off〉	【】	〈no〉	«=»	【】	⁅same as code⁆	❮Error: The value is the same as the 'code'❯	https://st.unicode.org/cldr-apps/v#/es_419//2f3689b454e651ff
> 
> Error:  Error: The value is the same as the 'code'`

Yeah, I'm testing a fix to that locally..

srl295:
Please Review:   Data Change + 1 test fix

macchiati (approved):

PR #5669, opened 2026-04-30
CLDR-19640 update CLDRModify and docs (add 'copy')

srl295:
CLDR-19640

- This adds `copy` and `copyNew` features to CLDRModify's config file mode (`-fk`)
- also, split out the config file mode of CLDRModify to a separate file (CLDRModify.java is nearly 4,000 lines)
- also, add `-V` option to CLDRModify

This was used to generate the changes in these PRs:
- #5888
- #5664

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRConfigFileFilter.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRConfigFileFilter.java
@@ -1,0 +1,394 @@
+package org.unicode.cldr.tool;
+
+import com.google.common.base.Splitter;
+import com.ibm.icu.text.UnicodeSet;
+import java.util.EnumMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import org.unicode.cldr.util.CLDRFile;
+import org.unicode.cldr.util.FileProcessor;
+import org.unicode.cldr.util.Joiners;
+import org.unicode.cldr.util.PatternCache;
+import org.unicode.cldr.util.RegexUtilities;
+import org.unicode.cldr.util.StringId;
+
+/** implementation for option -fk */
+class CLDRConfigFileFilter extends CLDRModify.CLDRFilter {
+
+    // pattern for a hex string id
+    static final UnicodeSet HEX = new UnicodeSet("[a-fA-F0-9]").freeze();
+
+    enum ConfigKeys {
+        action,
+        locale,
+        path,
+        value,
+        new_path,
+        new_value,
+        draft
+    }
+
+    enum ConfigAction {
+        /** Remove a path */
+        delete,
+        /** Add a path/value */
+        add,
+        /** Replace a path/value. Equals 'add' but tests selected paths */
+        replace,
+        /** Add a a path/value. Equals 'add' but tests that path did NOT exist */
+        addNew,
+        /** Copy a path to new_path */
+        copy,
+        /** Equals 'copy' but tsts that path did NOT exist. */
+        copyNew
+    }
+
+    static final class ConfigMatch {
+        final String exactMatch;
+        final Matcher regexMatch; // doesn't have to be thread safe
+        final ConfigAction action;
+        final boolean hexPath;
+
+        static UnicodeSet SUSPICIOUS_NON_REGEX = new UnicodeSet("[*|]").freeze();
+
+        public ConfigMatch(ConfigKeys key, String match) {
+            if (key == ConfigKeys.action) {
+                exactMatch = null;
+                regexMatch = null;
+                action = ConfigAction.valueOf(match);
+                hexPath = false;
+            } else if (match.length() > 1 && match.startsWith("/") && match.endsWith("/")) {
+                if (key != ConfigKeys.locale && key != ConfigKeys.path && key != ConfigKeys.value) {
+                    throw new IllegalArgumentException(
+                            "Regex only allowed for locale=, path=, or value'.");
+                }
+                exactMatch = null;
+                // for convenience, we automatically change [@attr="something"] to
+                // \[@attr="something"]
+                regexMatch =
+                        PatternCache.get(
+                                        match.substring(1, match.length() - 1)
+                                                .replace("[@", "\\[@"))
+                                .matcher("");
+                action = null;
+                hexPath = false;
+            } else {
+                exactMatch = match;
+                regexMatch = null;
+                action = null;
+                hexPath =
+                        (key == ConfigKeys.new_path || key == ConfigKeys.path)
+                                && HEX.containsAll(match);
+                if (key == ConfigKeys.locale || key == ConfigKeys.path) {
+                    if (SUSPICIOUS_NON_REGEX.containsSome(match)) {
+                        System.out.println(
+                                Joiners.ES.join(
+                                        "The value ",
+                                        match,
+                                        " is being matched literally, but contains regex charcters. Did you mean /",
+                                        match,
+                                        "/ ?"));
+                    }
+                }
+            }
+        }
+
+        public boolean matches(String other) {
+            if (exactMatch == null) {
+                return regexMatch.reset(other).find();
+            } else if (hexPath) {
+                // convert path to id for comparison
+                return exactMatch.equals(StringId.getHexId(other));
+            } else {
+                return exactMatch.equals(other);
+            }
+        }
+
+        @Override
+        public String toString() {
+            return action != null
+                    ? action.toString()
+                    : exactMatch == null
+                            ? regexMatch.toString()
+                            : hexPath ? "*" + exactMatch + "*" : exactMatch;
+        }
+
+        public String getPath(CLDRFile cldrFileToFilter) {
+            if (!hexPath) {
+                return exactMatch;
+            }
+            // ensure that we have all the possible paths cached
+            String path = StringId.getStringFromHexId(exactMatch);
+            if (path == null) {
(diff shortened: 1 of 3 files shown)


srl295:
@macchiati split out from #5664

tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt
```diff
@@ -1,3 +1,6 @@
 #locale;	action	path	new_value
-locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Samoa"]/long/generic
-locale=/.*/ ; action=delete ; path=//ldml/dates/timeZoneNames/metazone[@type="Samoa"]/long/daylight
+# See: https://cldr.unicode.org/development/cldr-big-red-switch/cldrmodify-using-config-file
+
+# CLDR-19394 copy Off/On from "ss" to the new typeValues (at unconfirmed)
+locale=/^.*/ ; action=copyNew ; path=//ldml/localeDisplayNames/types/type[@key="ss"][@type="none"][@scope="core"] ; new_path=//ldml/localeDisplayNames/typeValues/typeValue[@type="no"] ; draft=unconfirmed
+locale=/^.*/ ; action=copyNew ; path=//ldml/localeDisplayNames/types/type[@key="ss"][@type="standard"][@scope="core"] ; new_path=//ldml/localeDisplayNames/typeValues/typeValue[@type="yes"] ; draft=unconfirmed
```
srl295: @macchiati  wrote:

> This option looks useful for the case where regex replacement handles the changes. 

This is copying from one xpath to another.

I put in a todo to support `new_value` such as:

`path=/a/b ; value=/^(.*)ubber$/ ; new_value=fl$1 ; new_path=/c/d`

could then yield

`/a/b "rubber" ->  /c/d "flubber"`


but _as it is_ it's useful for copying values around as above. 

(it's a copy. Delete could be a separate step)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2416,24 +2424,38 @@ public void handleStart() {
...
-                                                newPathString,
-                                                newPathString,
-                                                newValue.exactMatch,
-                                                "config");
+                                    break;
+                                case copy:
+                                case copyNew:
+                                    {
+                                        // just check
+                                        if (pathMatch == null && newPath == null) {
```
macchiati: It seems the goal is to copy old <path,value> to <new_path, new_value> (where the new path and new_value may be derived from the <path,value> by regex replacement), then the new path AND old path must not be null.

```suggestion
                                        if (pathMatch == null || newPath == null) {
```

Or am I misreading the goal? I was thinking you were trying for something like the following, where the new path could have variables and new constant parts.

path: xxx(ABC)yyy
newpath: xxx$1zzz

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2550,26 +2572,62 @@ public void handlePath(String xpath) {
...
-                                    if (loc >= 0) {
-                                        int loc2 = fullpath.indexOf(']', loc + 7);
-                                        draft = fullpath.substring(loc, loc2 + 1);
-                                    }
+                                    {
+                                        ConfigMatch newPath = entry.get(ConfigKeys.new_path);
+                                        ConfigMatch newValue = entry.get(ConfigKeys.new_value);
+
+                                        String fullpath = cldrFileToFilter.getFullXPath(xpath);
+                                        String draft = "";
```
macchiati: Old path could have [@draft=xxx] vs no draft

If I pass in a 'draft' parameter, like draft: provisional vs no draft parameter, what do I expect to happen? There are 4 different combinations. Can you outline them?

macchiati: Also the old draft value could be either > or < than the draft parameter; does that make a difference?

I want to get the goals clear.

macchiati (dismissed):
Not really requesting changes so much as explanation as to goals for various combinations, so I can understand whether the code meets those goals.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
srl295: > It seems the goal is to copy old <path,value> to <new_path, new_value> (where the new path and new_value may be derived from the <path,value> by regex replacement), then the new path AND old path must not be null.
> 
> 
> Or am I misreading the goal?


First, you are correct that this should be `||`, will fix.

> I was thinking you were trying for something like the following, where the new path could have variables and new constant parts.
> 
> path: xxx(ABC)yyy
> newpath: xxx$1zzz

This is a TODO, yes as a later goal.

But it is already useful with a _constant_ path / new_path -- that is, copying from one path to another. That's how I used it in the linked PR that copies from `ss` to `<typeValue>`

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
srl295: > Old path could have [@draft=xxx] vs no draft

Yes, it could.  At present the full path is not used for the old path, so it doesn't check.

> If I pass in a 'draft' parameter, like draft: provisional vs no draft parameter, what do I expect to happen? There are 4 different combinations. Can you outline them?

Two combinations (In this codebase, can augment of course.)

- If you don't pass in a draft parameter, new path will have no draft (approved).
- If you pass in a draft parameter, it will be used unconditionally.

srl295: > Also the old draft value could be either > or < than the draft parameter; does that make a difference?
> 
> I want to get the goals clear.

It doesn't make a difference currently. (For my copy, I copied it as `unconfirmed`, the lowest level, so neither the old nor the new path could have anything at all lower level.)

I thought about having code that would take the maximum of the old-path's draft level and the _draft=_ parameter, and using that.

srl295: I hadn't used SearchXml before and I wasn't quite sure what the steps were.  I think you were suggesting something that would select the paths and then produce:

```
…
locale=de_CH;	action=add;	new_path=//ldml/localeDisplayNames/types/type[@key="ss"][@type="none"][@scope="core"][@draft="contributed"];	new_value=Aus
locale=de_CH;	action=add;	new_path=//ldml/localeDisplayNames/types/type[@key="ss"][@type="standard"][@scope="core"][@draft="contributed"];	new_value=Ein
…
```

… and then modify *that* to produce a new config file, to then run.

`copyNew` simplifies that, by allowing an A to B copy directly.

srl295:
I think I have docs (and this was already useful, it was used for the on/off mapping).  If this looks OK i can rebase

macchiati:
Will check it out

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -2416,24 +2424,38 @@ public void handleStart() {
...
-                                            || valueMatch != null
-                                            || newPath == null
-                                            || newValue == null) {
-                                        throw new IllegalArgumentException(
-                                                action.action
-                                                        + ": must have no path nor value = null AND new_path or new_value:\n\t"
-                                                        + entry);
+                                    {
+                                        if (pathMatch != null
+                                                || valueMatch != null
```
macchiati: Ah, this would be the place where you could allow valueMatch to be non-null.

srl295:
> tools/cldr-code/src/main/resources/org/unicode/cldr/tool/modify_config.txt

Changing this file makes this PR unmergeable. I'll remove it from the PR.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRConfigFileFilter.java
```diff
@@ -0,0 +1,394 @@
...
+                            // the copyNew action skips if it's already here
+                            break;
+                        }
+                        // TOOD: Allow skipping inheritance marker?
+                        replace(
+                                oldNewPathString,
+                                newPathString,
+                                value, // TODO: allow new_value to override with
+                                // match
+                                "config");
```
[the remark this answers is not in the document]
srl295: @copilot is 100% wrong here.

Line 377 calls `isHere()` in just this situation.

Also, oldNewPathString is appropriate for test, because existence uses the Dpath not the full path.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRModify.java
```diff
@@ -342,7 +212,13 @@ public static String getModified(
 
     static String sourceInput;
 
+    /** print out if in verbose mode */
+    public static final void verboseln(String fmtString, Object... opts) {
+        if (isVerbose) System.err.println("CLDRModify: " + String.format(fmtString, opts));
+    }
+
     static CLDRTreeWriter treeWriter = null;
+    private static boolean isVerbose = true;
```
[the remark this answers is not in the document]
srl295: Fixed

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CLDRConfigFileFilter.java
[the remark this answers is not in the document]
robertbastian: thanks for checking

robertbastian (approved):
without reviewing this line by line, this seems to do what it should

PR #5672, opened 2026-04-30
CLDR-19035 Update ExampleDependencies

btangmu:
CLDR-19035

ALLOW_MANY_COMMITS=true


btangmu:
I suspect there's a mistake here. The file size decreased to 37% of the old size! I'm making this draft PR to aid in investigation.

btangmu:
In the 1st commit there are 148 lines added and 1,195 lines removed. Maybe some clues will emerge from picking at random one of the 1,195 lines removed and investigating why it's no longer generated...

macchiati:
Yes, that seems very surprising; if anything, we have added more examples,
which should increase the number of dependencies.

btangmu:
> Yes, that seems very surprising; if anything, we have added more examples, which should increase the number of dependencies.

As agreed in the infrastructure meeting, the 2nd commit combines the existing dependencies with the newly-generated dependencies, as a temporary work-around. It does so in the simplest way, by copy-pasting. The resulting duplications in the file may have a performance penalty, only when ST starts, probably not noticeable or worth worrying about in the short run. Another performance penalty is that some examples may get regenerated unnecessarily, in those cases where the reduction in dependencies is correct. In the short run, inefficiency is better than failing to update examples where needed.

AEApple:
When will this be ready to review? It would be great if we can include this fix in the push to production this afternoon.

btangmu:
> When will this be ready to review?

I haven't had time today to test it, sorry! There's a chance I can test it late tonight, or tomorrow. Testing should at least verify that the mechanism is working to update the example for a path when a value for a different path, on which the example depends, changes.

AEApple:
> > When will this be ready to review?
> 
> I haven't had time today to test it, sorry! There's a chance I can test it late tonight, or tomorrow. Testing should at least verify that the mechanism is working to update the example for a path when a value for a different path, on which the example depends, changes.

SG! I'll go ahead with the push to production now and then we can do another one tomorrow if needed.

btangmu:
I've tested locally and confirmed that a new dependency works that didn't work previously.

Change value for http://localhost:9080/cldr-apps/v#/fr/Keys/605933a80afb8993 and the example for http://localhost:9080/cldr-apps/v#/fr/Keys/2f3689b454e651ff follows suit.

The corresponding URLs for cldr-smoke are 

https://cldr-smoke.unicode.org/cldr-apps/v#/fr/Keys/605933a80afb8993 and https://cldr-smoke.unicode.org/cldr-apps/v#/fr/Keys/2f3689b454e651ff

Note that even locally it was necessary to reload the page in order for the example to update. Possibly waiting longer would make it happen automatically. Anyway, for the present purpose, the point is that the dependency works with this PR where it didn't work previously.

I also tested that a dependency that worked previously still works with this PR.

https://cldr-smoke.unicode.org/cldr-apps/v#/fr/T_Africa/7bb1ecb728990415

"Afrique du Sud" -- change to "Afrique du Sud 123"

https://cldr-smoke.unicode.org/cldr-apps/v#/fr/Languages_A_D/97554ebf534323c

Example includes 123, good. Tested the same thing locally with changes from this PR, and it still works.

Note that both on cldr-smoke and locally an error occured. I made a new ticket for that: https://unicode-org.atlassian.net/browse/CLDR-19458 . In spite of the error, the example did update correctly according to the dependency.

btangmu:
I will be out of office for the rest of the day, so if you approve this you may also want to merge it.

macchiati (approved):
We really need to turn this into an ordered  text file to make it reviewable. But approving in the interests of time

PR #5680, opened 2026-05-01
CLDR-19353 Tags: prevent select/copy/paste

btangmu:
CLDR-19353

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/AddValueTags.vue
+++ b/tools/cldr-apps/js/src/views/AddValueTags.vue
@@ -110,6 +110,7 @@ function handleChooseCharacter() {
 }
 
 .tag-array {
+  user-select: none;
   font-weight: bold;
   color: black;
   margin: 1px;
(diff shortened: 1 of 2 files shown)


srl295 (approved):

macchiati (approved):
Looks good. Like all the UI stuff, should quickly check on staging before merging.

btangmu:
> Looks good. Like all the UI stuff, should quickly check on staging before merging.

I've tested it locally and it works. I think it would be harmless even if it didn't work under some circumstances (such as different browsers). Would you like me to push it to cldr-staging before merging this PR? Alternatively we could test on cldr-smoke before pushing to production.

macchiati:
I kicked off a staging run: https://github.com/unicode-org/cldr/actions/runs/25230861689

srl295:
it passed the sanity check. It's just a CSS update to that section. i can check it locally but i expect it to be a low risk item.

srl295:
From the test run linked here, i can see the input box is working (it doesn't test tags _yet_) 
image

srl295:
Works for me. On smoketest i can select and copy as shown here, with this change i can't.

Smoketest:
image

This PR: (note selection skips the chit area)

image

macchiati:
Merging

macchiati:
I checked also; works great.

PR #5682, opened 2026-05-02
CLDR-19433 Other should not overlap Abstained

srl295:
CLDR-19433

image
- 

ALLOW_MANY_COMMITS=true


macchiati (approved):

AEApple:
Happy to be reviewer on the ticket in Jira after this merges and can be checked on Smoke Test.

PR #5686, opened 2026-05-05
CLDR-19394 on/off: add collisions and logical groups

srl295:
- move YES_NO to LogicalGrouping
- attempt (!!) to document some mysterious code

CLDR-19394

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
@@ -91,7 +91,7 @@ public class CheckDisplayCollisions extends FactoryCheckCLDR {
         TYPOGRAPHIC_AXIS("//ldml/typographicNames/axisName", MatchType.PREFIX),
         TYPOGRAPHIC_FEATURE("//ldml/typographicNames/featureName", MatchType.PREFIX),
         TYPOGRAPHIC_STYLE("//ldml/typographicNames/styleName", MatchType.PREFIX),
-        ;
+        TYPE_VALUE("//ldml/localeDisplayNames/typeValues/typeValue", MatchType.PREFIX);
 
         private MatchType matchType;
         private String basePrefix;
@@ -611,7 +611,12 @@ public class CheckDisplayCollisions extends FactoryCheckCLDR {
 
         // removeMatches(myType);
         // check again on size
-        if (paths.size() <= 1) {
+        if (myType == Type.TYPE_VALUE) {
+            // TODO: CLDR-19394 typeValue collides with just 1 item, why "<=1" below?
+            if (paths.isEmpty()) {
+                return this;
+            }
+        } else if (paths.size() <= 1) {
             return this;
         }
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -1,9 +1,10 @@
 package org.unicode.cldr.test;
 
+import static org.unicode.cldr.util.LogicalGrouping.YES_NO;
+
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import com.ibm.icu.impl.Row.R3;
 import com.ibm.icu.impl.Utility;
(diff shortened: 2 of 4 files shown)


srl295 (commented):
## Examples:

### With Off vs Off:

```
en [English]    error   ▸Locale_Display_Names|Keys|Type_Values|no◂      〈Off〉 【】    〈Off〉 «=»     【】    ⁅display collision⁆      ❮Error: Can't have same translation as [Type Values: yes]. Please change either this name or the other one.❯     https://st.unicode.org/cldr-apps/v#/en//2f3689b454e651ff
/Users/srl295/src/cldr/common/main/en.xml:1786:41: error
en [English]    error   ▸Locale_Display_Names|Keys|Type_Values|yes◂     〈Off〉 【】    〈Off〉 «=»     【】    ⁅display collision⁆      ❮Error: Can't have same translation as [Type Values: no]. Please change either this name or the other one.❯      https://st.unicode.org/cldr-apps/v#/en//acf3f6d6ab8b155
```

### With "Off" but no "On"


```
en [English]    error   ▸Locale_Display_Names|Keys|Type_Values|yes◂     〈null〉        【】    〈null〉        «»【】    ⁅incomplete logical group⁆      ❮Error: Incomplete logical group - missing values for: [yes]; level=modern❯     https://st.unicode.org/cldr-apps/v#/en//acf3f6d6ab8b155  code-fallback
```

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
```diff
@@ -611,7 +611,12 @@ public CheckCLDR handleCheck(
 
         // removeMatches(myType);
         // check again on size
-        if (paths.size() <= 1) {
+        if (myType == Type.TYPE_VALUE) {
+            // TODO: CLDR-19394 typeValue collides with just 1 item, why "<=1" below?
+            if (paths.isEmpty()) {
+                return this;
+            }
+        } else if (paths.size() <= 1) {
```
srl295, line 619: @macchiati I can't figure out why this is `<=1` and not `<1` - any ideas?

macchiati: I think the paths variable might at this point include the path being checked.

srl295: The path cleaning logic deletes the path under investigation

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
```diff
@@ -611,7 +611,12 @@ public CheckCLDR handleCheck(
 
         // removeMatches(myType);
         // check again on size
-        if (paths.size() <= 1) {
+        if (myType == Type.TYPE_VALUE) {
+            // TODO: CLDR-19394 typeValue collides with just 1 item, why "<=1" below?
```
macchiati, line 615: Which paths are colliding with which others? Without knowing that, tricky to say what is happening. Also compare to results when you have other collisions, like "Jan" for February (just manually make a termporary change in data)

srl295:
Looks like there's some coverage issue still

macchiati (commented):
Looks good, except that we should figure out what's happening with the question you raise.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
srl295: OK, so in this situation, you would expect two paths in the collision set?

macchiati: I think so. But you should verify

srl295: > Which paths are colliding with which others? Without knowing that, tricky to say what is happening. Also compare to results when you have other collisions, like "Jan" for February (just manually make a termporary change in data)

"Jan" is an unrelated test, `CheckDates`.  It triggers with only one path difference, does not require 2.

srl295:
@macchiati I already have this marked with "TODO: CLDR-19394" in the code.  Can we merge this and come back to what's happening with CheckDisplayCollisions? I can't figure out what the code is trying to do.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
macchiati: Hmmm.

macchiati (commented):
We can merge for now, but let's look into this oddity when the dust settles.

macchiati (approved):

PR #5688, opened 2026-05-05
CLDR-9980 Examples/checks for numericDateSeparator, timeSeparator

macchiati:
CLDR-9980

The code for the test 
1. pulls in 'core' values: the short date and yMd patterns, and the short time, Hms, and hms patterns
2. Extracts numeric separators, looking for them in the patterns between certain numeric fields (eg only considers M or MM, not MMM or MMMM). 
    * yM, My, Md, dM, yd, dy for dates
    * hm, ms, Hm for times (those are monotonic-decreasing in all our data.
4. this is done with some updates to the DatetimeUtilities (a step towards modernization)
5. checks to see that the value for the corresponding numeric separator (date or time) is in the set of extracted separators
6. Emits an Check error if not.
    * Special exception for fr_CA time; that is a warning not an errorl.
7. This is run using ConsoleCheck, extracting the errors, and using those (with some modification) to create a modify_config file:
    * https://docs.google.com/spreadsheets/d/194TD8od1y7DZdnkNMESfkVT_L0bHSVu401Sg_seLeK0/edit?gid=409625540#gid=409625540
9. That is used to make changes in the data (still provisional) so that vetters don't have to deal with a batch of errors.
    * https://docs.google.com/spreadsheets/d/194TD8od1y7DZdnkNMESfkVT_L0bHSVu401Sg_seLeK0/edit?gid=2026585233#gid=2026585233

Note the error message provide an abbreviated reference to the source paths and their values, like
    * Warning: Mismatch with [h, min], see: [**Time Formats | short** → «HH 'h' mm», **24 Hour Time Formats | Hms** → «HH 'h' mm 'min' ss 's'», **12 Hour Time Formats | hms** → «h 'h' mm 'min' ss 's' a»]

I'd like to put html links on those, but have forgotten whether we can do HTML that in the generated messages or not.

Note: next step after this PR will be to modify ExampleGenerator to add examples. My idea is to use those same patterns, and substitute a character for the separator to show what effect a user option could have. Maybe ➖, to make it obvious.

I'm also thinking that we can use some of this code to check short numeric dates and times use consistent separators (maybe a warning on them).

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/bs.xml
+++ b/common/main/bs.xml
@@ -5238,7 +5238,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</timeFormats>
 				<dateTimeFormats>
 					<numericSeparators>
-						<numericDateSeparator draft="provisional">-</numericDateSeparator>
+						<numericDateSeparator draft="provisional">.</numericDateSeparator>
 						<numericTimeSeparator draft="provisional">:</numericTimeSeparator>
 					</numericSeparators>
 					<dateTimeFormatLength type="full">
--- a/common/main/am.xml
+++ b/common/main/am.xml
@@ -2399,7 +2399,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</timeFormats>
 				<dateTimeFormats>
 					<numericSeparators>
-						<numericDateSeparator draft="provisional">-</numericDateSeparator>
+						<numericDateSeparator draft="provisional">/</numericDateSeparator>
 						<numericTimeSeparator draft="provisional">:</numericTimeSeparator>
 					</numericSeparators>
 					<dateTimeFormatLength type="full">
--- a/common/main/ar.xml
+++ b/common/main/ar.xml
@@ -2088,7 +2088,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</timeFormats>
 				<dateTimeFormats>
 					<numericSeparators>
-						<numericDateSeparator draft="provisional">-</numericDateSeparator>
+						<numericDateSeparator draft="provisional">‏/</numericDateSeparator>
 						<numericTimeSeparator draft="provisional">:</numericTimeSeparator>
 					</numericSeparators>
 					<dateTimeFormatLength type="full">
--- a/common/main/as.xml
+++ b/common/main/as.xml
@@ -1602,7 +1602,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</timeFormats>
 				<dateTimeFormats>
 					<numericSeparators>
-						<numericDateSeparator draft="provisional">.</numericDateSeparator>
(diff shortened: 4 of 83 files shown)


common/main/bs.xml
```diff
@@ -5238,7 +5238,7 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</timeFormats>
 				<dateTimeFormats>
 					<numericSeparators>
-						<numericDateSeparator draft="provisional">-</numericDateSeparator>
+						<numericDateSeparator draft="provisional">.</numericDateSeparator>
```
AEApple, line 5241:
```suggestion
						<numericDateSeparator draft="provisional">. </numericDateSeparator>
```

AEApple (commented):
I checked up through es_MX and then from ur through the end.

I had a questions since it looks like there are some locales like bs where the separator is "`. `" for dates.

Do we need to do anything to handle where white space is an expected part of the abbreviation since it can also appear after the final date unit is formats like ymd see example https://st.unicode.org/cldr-apps/v#/bs/Gregorian/4af189b39e4e8ddf

common/main/bs.xml
macchiati: I looked at that. Right now I'm trimming those (eg, dropping trailing spaces). We do have the ability to change them to full strings, and then replace those strings. 

However, it's a bit tricky. "d. M. y." is an interesting example (and that is the example shared across the locales that have it), because the ". " (that is, <dot, space>) is not a really a **separator**; the dot is a **terminator** (for each field) and the **separator** is a space. We see that in a related pattern on the page: "d. M. y. G"

I think the desired result for users is that if they want to change the separator to be a "/" (for example), the result would be the following:

"d. M. y." → "d/M/y" — Not "d/M/y/" 
"d. M. y. G" → "d/M/y G" — Not "d/M/y/G" 

1. We could specify that that is the behavior when the separator is ". ". It would be something like: with Sep = X+space, and the replacement is Y, the behavior is:
* Replace Sep between yMd fields by Y, and following the last yMd field replace X by "".

2. We could also specify that when replacing a separator of X by Y, if we find X+space between yMd fields, we replace that by Y, and then also replace a trailing X.

Both of those could produce the above result. The first approach might be the cleaner one.

I propose merging this for now, and then discuss how to best deal with X+space offline and in the Design WG. That way we have a fix for the situation for the other locales as soon as possible.

The alternative would be to go ahead and space for the 6ish locales that had ". " I'm ok with that as well (might be able to do that before tomorrow's meeting).

Thoughts?

AEApple: We should go ahead and merge, and then push to production tomorrow. We also need to add instructions in the meantime for vetters for any locales that do this tomorrow so that they don't change in an unexpected way.

Agree that we would need to discuss patterns where the locales use a . as a terminator in the Design WG next week.

AEApple (approved):

common/main/bs.xml
macchiati: SGTM

PR #5696, opened 2026-05-07
CLDR-19457 search: be able to search by hexid or url

srl295:
- also add a test for stringid length
- allow hex id or pasted URL that looks like a ST URL

### pasting a Survey Tool URL

Example `https://cldr-smoke.unicode.org/cldr-apps/v#/es/Languages_E_J/2ca9655fb4378956`

Note: if you are on the main page, it will use the locale from the URL (such as 'es').
Otherwise, it will just search within your current locale.

image

### Pasting a hex hash

image

CLDR-19457

ALLOW_MANY_COMMITS=true


macchiati:
An xpath or segment of it world be really useful for us

srl295:
> An xpath or segment of it world be really useful for us

- exact xpath ( as in `//ldml…` ) already works
- this adds searching for a hex xpath id
- I think xpath substring is there, not sure. That could be added.

tools/cldr-apps/js/src/views/SearchPanel.vue
```diff
@@ -2,7 +2,7 @@
   <div>
     <a-input-search
       v-model:value="searchText"
-      placeholder="Search…"
+      placeholder="Words, XPath, Hex…"
```
macchiati: "Words" is misleading. It appears to be searching:
* Values (only whole)
* Codes (whole or part)
* Xpaths (only whole)**
* ST urls (whole or hex; only items*) // new in this release

\* just row URLs, not whole or part of:
* Page: https://st.unicode.org/cldr-apps/v#/de/Alphabetic_Information/
* Header: https://st.unicode.org/cldr-apps/v#/de/Alphabetic_Information/header_Characters_in_Use

\** But it says "Partial XPath" in the search box.

----
For the future, we should also increase the size of the search box

And fix the confusing top (dotted underline followed by close box). 

Image

Compare a browser Find in Page box.

Image

It's going to be different, because it needs the search results in the box. But the unusual dotted line, etc. can be collapsed; we don't need the magnifying glass because we just clicked on one to get there, and the box has a gray 'Search..." anyway. So it can all be on the top line and look like similar search boxes to that extent.

macchiati (changes requested):
Ah, I wondered why you hadn't responded — but my mistake, I had a pending comment for a review but didn't actually hit the button.

I suggest at least fixing the messages to not be misleading.

tools/cldr-apps/js/src/views/SearchPanel.vue
srl295: Sounds good.

srl295: It does search partial xpaths, such as `//ldml/dates/calendars/calendar[@type="gregorian"]/days/dayContext[@type="stand-alone"]/dayWid`
It does only search whole values

ST URLs need to be whole, as do hex values.

macchiati: I searched for an element (I think `gmtFormat`) and got nothing.

srl295: Thanks for the example, that's helpful.  By partial XPath, I should have said a _prefix_, so `//ldml/…`

srl295:
> Ah, I wondered why you hadn't responded — but my mistake, I had a pending comment for a review but didn't actually hit the button.
> 
> I suggest at least fixing the messages to not be misleading.

I think it just says Search... now

macchiati:
That's probably best. In the Survey tool guide, it should list the kinds of things it can (and can't) search for.

srl295:
> That's probably best. In the Survey tool guide, it should list the kinds of things it can (and can't) search for.

The UI now looks like this 

image

and links to https://cldr.unicode.org/translation#searching-in-the-survey-tool

Could we merge this change and then iterate on the documentation?

macchiati:
The docs definitely need fixing, but ok to do in separate PR.

macchiati (approved):

PR #5697, opened 2026-05-07
CLDR-19394 Off/On: suppress 'ss/core'

srl295:
- The old Off/On should now be suppressed in SurveyTool
- also, update examples for off/on as follows. (Did not add every type, so it's not unwieldy. All of the other sorting types end with Sorting.) 
  - 〖Ignore Symbols Sorting: On〗〖Sentence Break After Abbr.: On〗
  - 〖Ignore Symbols Sorting: ❬Off❭〗〖Sentence Break After Abbr.: ❬Off❭〗

CLDR-19394

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -175,7 +175,10 @@ For terms of use, see http://www.unicode.org/copyright.html
 
 		<coverageVariable key="%variantTypes" value="([A-Z0-9]++)"/>
 		<coverageVariable key="%wideAbbr" value="(wide|abbreviated)"/>
+		<!-- for typeValue -->
 		<coverageVariable key="%yesNo" value="(yes|no)"/>
+		<!-- keys that are applied to yesNo for core values. TODO CLDR-19394 should come from structure to be spec’d -->
+		<coverageVariable key="%yesNoKeys" value="(colAlternate|colBackwards|colCaseLevel|colNormalization|colNumeric|ss)" />
 
 		<coverageVariable key="%anyAttribute" value='([^\x{22}]++)'/>
 		<coverageVariable key="%A" value='([^\x{22}]++)'/>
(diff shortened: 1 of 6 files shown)


macchiati (dismissed):

srl295:
@macchiati thanks for the early review. ready to merge once tests pass.

common/supplemental/coverageLevels.xml
```diff
@@ -175,7 +175,10 @@ For terms of use, see http://www.unicode.org/copyright.html
 
 		<coverageVariable key="%variantTypes" value="([A-Z0-9]++)"/>
 		<coverageVariable key="%wideAbbr" value="(wide|abbreviated)"/>
+		<!-- for typeValue -->
 		<coverageVariable key="%yesNo" value="(yes|no)"/>
+		<!-- keys that are applied to yesNo for core values. TODO CLDR-19394 should come from structure to be spec’d -->
+		<coverageVariable key="%yesNoKeys" value="(colAlternate|colBackwards|colCaseLevel|colNormalization|colNumeric|ss)" />
```
macchiati, line 181: These values should be dropped from ExtraPaths, so that they never occur in the first place. If we then remove the data, they will never show up.

However, this would be a good first step.

(Having paths we never need is just a burden for all processing.)

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -4180,4 +4191,64 @@ public static String simplify(String exampleHtml, boolean internal) {
...
+     * @param path Example: "//ldml/%s"
+     * @param args
+     * @return result or null
+     */
+    final String getStringValue(final String path, final String... args) {
+        return getStringValue(getCldrFile(), String.format(path, args));
+    }
+
+    /**
+     * Convenience function for getting a formatted path from the current file, fallign back to
```
macchiati, line 4227: Where is this being used? We don't want to show "half english" examples.

srl295: So you can format On/Off without the underlying types? Otherwise it's just blank. 
Should it instead be:  missing: On, missing: off etc?

common/supplemental/coverageLevels.xml
srl295: they are dropped from extraPaths

srl295: dropped from ExtraPaths in https://github.com/unicode-org/cldr/pull/5697/changes#diff-2dea279cfe11acf813bb0fe7cc195da8117539f92b31d79cff262b5ef49a1761R388

srl295:
@macchiati think this is ready to go then?

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
macchiati: Hmmm. Let's use the english value, but within ⟨…⟩ to let them know the item is missing and is only temporarily English. 
We can add that to the InfoPage.

srl295: image

srl295: image

srl295:
image

macchiati (approved):

PR #5699, opened 2026-05-08
CLDR-19440 Keyboards: use UnicodeSet notation for test repertoire

mcdurdin:
- Test repertoire.chars, unlike other elements and attributes, use UnicodeSet notation, per draft spec.
- Update README.md to highlight difference in Unicode character escapes.

CLDR-19440

ALLOW_MANY_COMMITS=true


srl295 (approved):

AEApple:
@mcdurdin - The ticket was approved by the CLDR TC, you should be able to merge the PR now.

PR #5705, opened 2026-05-09
CLDR-9980 check for datetime numeric separators

macchiati:
CLDR-9980

The previous PR checked that the numeric separators matched the source patterns (from stock-date/time, yMd, Hms, hms).
This PR checks that the _other_ patterns match the numeric separators. If both are done, then everything will match cleanly.

1. I found that the previous PR wasn't checking non-gregorian. That led to fixes in the common/main files
    * root.xml needed two fixes; the calendars don't inherit everything from one another, so 2 needed fixes to inherit properly
2. In CheckCLDR fixed some enums to be more accurate
3. In CheckDates there was a big dispatch table. I cleaned up part of it to have a switch(dateTypePatternType), with chunks of code refactored into function calls.
    * checkNumericSeparators(parts, value, result);
    * checkForNumericSeparator(parts, value, result);
        *  this one is new: it checks a pattern to see if it has any numeric separators, and if so checks them against the define ones. 
        * If they don't match, emits a message like: ❮Error: [-] separator doesn’t match “‏/” (default numeric date separator for this locale+calendar, from date-short and yMd).❯
        * For now this is a warning. There are ~280 instances across all TC locales. Most are probably real errors — inconsistencies that shouldn't be there: but some are ok (such as Swedish 〈d/M〉), and we don't want to be errors. So it takes some more investigation to see which we can enforce changes for.
    * checkStockAvailableInterval(path, value, dateTypePatternType, result);
    *  checkDateTimeSkeleton(path, value, result);
    * checkHourFormat(value, result);
    * One change was to refine the criterion for when to consider characters between numeric date fields to be separators, excluding letters, for example.
5. In ConsoleCheckCLDR a minor change to stop printing blank lines unless a locale has issues.
6. In DateUtilities, added FieldKind and some more information to the FieldType, and some utilities
7. TestCheckCLDR just had updated enums.

ALLOW_MANY_COMMITS=true


macchiati:
Tests now pass

macchiati:
I would like to get this merged. Can someone review please? (The only test failure is Docker, which is irrelevant)

srl295:
@macchiati sorry did not get to look at this one

bin/.gitignore
```diff
@@ -1 +1,2 @@
+/fixedcandidates/
 /org/
```
srl295: @macchiati why is this here? This directory shouldn't even exist unless there is an error.

srl295 (dismissed):

srl295 (dismissed):

srl295 (dismissed):

macchiati:
Sorry, can you stamp again? My removal of the .gitignore had two unrelated files in the commit, so I reverted, and then did another commit to remove it.

----
Side issue

I think what happens with the .gitignore is that 
* eclipse creates the bin directory for class files whenever I build
* I've used eclipse settings to avoid checking those in
* Eclipse creates a bin/.gitignore to do that
* and that file keeps getting created.

srl295 (approved):

srl295:
> Sorry, can you stamp again? My removal of the .gitignore had two unrelated files in the commit, so I reverted, and then did another commit to remove it.
> 
> Side issue
> 
> I think what happens with the .gitignore is that
> 
> * eclipse creates the bin directory for class files whenever I build
> * I've used eclipse settings to avoid checking those in
> * Eclipse creates a bin/.gitignore to do that
> * and that file keeps getting created.

fixed in https://github.com/unicode-org/cldr/pull/5713

PR #5709, opened 2026-05-11
CLDR-19466 Generate Solid Blueprint for Decimal and Compact Decimal Format Testing Based on ICU4J

younies:
CLDR-19466

## Summary
Implementint a solid Blueprint data generation and verification tool for decimal and compact decimal formats in CLDR, based on **ICU4J**.

## Key Changes
- **GenerateDecimalFormatTestData.java**: Added a new tool to generate expected formatted numbers across various locales and dimensions (input decimal, compactness ... etc.).

## Next Steps
- Add the needed dimensions to test both decimal and compact decimal numbers.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/testData/decimal/decimals_extended_numbers.tsv
+++ b/common/testData/decimal/decimals_extended_numbers.tsv
@@ -1,0 +1,6301 @@
+locale	number_format	format_length	input	expected
+ar	decimal		-5.0E12	‎-5,000,000,000,000
+ar	decimal		-1.5E12	‎-1,500,000,000,000
+ar	decimal		-1.0E12	‎-1,000,000,000,000
+ar	decimal		-5.0E11	‎-500,000,000,000
+ar	decimal		-1.5E11	‎-150,000,000,000
+ar	decimal		-1.0E11	‎-100,000,000,000
+ar	decimal		-5.0E10	‎-50,000,000,000
+ar	decimal		-1.5E10	‎-15,000,000,000
+ar	decimal		-1.0E10	‎-10,000,000,000
+ar	decimal		-5.0E9	‎-5,000,000,000
+ar	decimal		-1.5E9	‎-1,500,000,000
+ar	decimal		-1.0E9	‎-1,000,000,000
+ar	decimal		-5.0E8	‎-500,000,000
+ar	decimal		-1.5E8	‎-150,000,000
+ar	decimal		-1.0E8	‎-100,000,000
+ar	decimal		-5.0E7	‎-50,000,000
+ar	decimal		-1.5E7	‎-15,000,000
+ar	decimal		-1.0E7	‎-10,000,000
+ar	decimal		-5000000.0	‎-5,000,000
+ar	decimal		-1500000.0	‎-1,500,000
+ar	decimal		-1234567.0	‎-1,234,567
+ar	decimal		-1234565.0	‎-1,234,565
+ar	decimal		-1000000.0	‎-1,000,000
+ar	decimal		-999999.9	‎-999,999.9
+ar	decimal		-500000.0	‎-500,000
+ar	decimal		-150000.0	‎-150,000
+ar	decimal		-100000.0	‎-100,000
+ar	decimal		-50000.0	‎-50,000
+ar	decimal		-15000.0	‎-15,000
+ar	decimal		-10000.0	‎-10,000
+ar	decimal		-5000.0	‎-5,000
+ar	decimal		-1500.0	‎-1,500
+ar	decimal		-1234.56	‎-1,234.56
+ar	decimal		-1000.0	‎-1,000
+ar	decimal		-999.9	‎-999.9
+ar	decimal		-500.0	‎-500
+ar	decimal		-150.0	‎-150
+ar	decimal		-123.0	‎-123
+ar	decimal		-100.0	‎-100
+ar	decimal		-50.0	‎-50
+ar	decimal		-15.0	‎-15
+ar	decimal		-12.0	‎-12
+ar	decimal		-10.0	‎-10
+ar	decimal		-5.0	‎-5
+ar	decimal		-3.5	‎-3.5
+ar	decimal		-2.5	‎-2.5
+ar	decimal		-1.5	‎-1.5
+ar	decimal		-1.2	‎-1.2
+ar	decimal		-1.0	‎-1
+ar	decimal		-0.5	‎-0.5
+ar	decimal		-0.15000000000000002	‎-0.15
+ar	decimal		-0.135	‎-0.135
+ar	decimal		-0.125	‎-0.125
+ar	decimal		-0.1	‎-0.1
+ar	decimal		-0.05	‎-0.05
+ar	decimal		-0.015	‎-0.015
+ar	decimal		-0.01	‎-0.01
+ar	decimal		-0.00831765	‎-0.008318
+ar	decimal		-0.005	‎-0.005
+ar	decimal		-0.0015	‎-0.0015
+ar	decimal		-0.001	‎-0.001
+ar	decimal		-5.0E-4	‎-0.0005
+ar	decimal		-1.5000000000000001E-4	‎-0.00015
+ar	decimal		-1.23E-4	‎-0.000123
+ar	decimal		-1.0E-4	‎-0.0001
+ar	decimal		-5.0E-5	‎-0.00005
+ar	decimal		-1.5000000000000002E-5	‎-0.000015
+ar	decimal		-1.0E-5	‎-0.00001
+ar	decimal		-4.9999999999999996E-6	‎-0.000005
+ar	decimal		-1.5E-6	‎-0.000002
+ar	decimal		-1.0E-6	‎-0.000001
+ar	decimal		-0.0	‎-0
+ar	decimal		1.0E-6	0.000001
+ar	decimal		1.5E-6	0.000002
+ar	decimal		4.9999999999999996E-6	0.000005
+ar	decimal		1.0E-5	0.00001
+ar	decimal		1.5000000000000002E-5	0.000015
+ar	decimal		5.0E-5	0.00005
+ar	decimal		1.0E-4	0.0001
+ar	decimal		1.23E-4	0.000123
+ar	decimal		1.5000000000000001E-4	0.00015
+ar	decimal		5.0E-4	0.0005
+ar	decimal		0.001	0.001
+ar	decimal		0.0015	0.0015
+ar	decimal		0.005	0.005
+ar	decimal		0.01	0.01
+ar	decimal		0.015	0.015
+ar	decimal		0.05	0.05
+ar	decimal		0.1	0.1
+ar	decimal		0.125	0.125
+ar	decimal		0.135	0.135
+ar	decimal		0.15000000000000002	0.15
+ar	decimal		0.5	0.5
+ar	decimal		1.0	1
+ar	decimal		1.5	1.5
+ar	decimal		2.5	2.5
+ar	decimal		3.5	3.5
+ar	decimal		5.0	5
+ar	decimal		10.0	10
+ar	decimal		12.0	12
+ar	decimal		15.0	15
+ar	decimal		50.0	50
+ar	decimal		100.0	100
+ar	decimal		123.0	123
+ar	decimal		150.0	150
+ar	decimal		500.0	500
+ar	decimal		999.9	999.9
+ar	decimal		1000.0	1,000
+ar	decimal		1234.56	1,234.56
+ar	decimal		1500.0	1,500
+ar	decimal		5000.0	5,000
+ar	decimal		10000.0	10,000
+ar	decimal		15000.0	15,000
+ar	decimal		50000.0	50,000
+ar	decimal		100000.0	100,000
+ar	decimal		150000.0	150,000
+ar	decimal		500000.0	500,000
+ar	decimal		999999.9	999,999.9
+ar	decimal		1000000.0	1,000,000
+ar	decimal		1234567.0	1,234,567
+ar	decimal		1500000.0	1,500,000
+ar	decimal		5000000.0	5,000,000
+ar	decimal		1.0E7	10,000,000
+ar	decimal		1.5E7	15,000,000
+ar	decimal		5.0E7	50,000,000
+ar	decimal		1.0E8	100,000,000
+ar	decimal		1.5E8	150,000,000
+ar	decimal		5.0E8	500,000,000
+ar	decimal		1.0E9	1,000,000,000
+ar	decimal		1.5E9	1,500,000,000
+ar	decimal		5.0E9	5,000,000,000
+ar	decimal		1.0E10	10,000,000,000
+ar	decimal		1.5E10	15,000,000,000
+ar	decimal		5.0E10	50,000,000,000
+ar	decimal		1.0E11	100,000,000,000
+ar	decimal		1.5E11	150,000,000,000
+ar	decimal		5.0E11	500,000,000,000
+ar	decimal		1.0E12	1,000,000,000,000
+ar	decimal		1.5E12	1,500,000,000,000
+ar	decimal		5.0E12	5,000,000,000,000
+ar	percent		-5.0E12	‎-500,000,000,000,000‎%‎
+ar	percent		-1.5E12	‎-150,000,000,000,000‎%‎
+ar	percent		-1.0E12	‎-100,000,000,000,000‎%‎
+ar	percent		-5.0E11	‎-50,000,000,000,000‎%‎
+ar	percent		-1.5E11	‎-15,000,000,000,000‎%‎
+ar	percent		-1.0E11	‎-10,000,000,000,000‎%‎
+ar	percent		-5.0E10	‎-5,000,000,000,000‎%‎
+ar	percent		-1.5E10	‎-1,500,000,000,000‎%‎
+ar	percent		-1.0E10	‎-1,000,000,000,000‎%‎
+ar	percent		-5.0E9	‎-500,000,000,000‎%‎
+ar	percent		-1.5E9	‎-150,000,000,000‎%‎
+ar	percent		-1.0E9	‎-100,000,000,000‎%‎
+ar	percent		-5.0E8	‎-50,000,000,000‎%‎
+ar	percent		-1.5E8	‎-15,000,000,000‎%‎
+ar	percent		-1.0E8	‎-10,000,000,000‎%‎
+ar	percent		-5.0E7	‎-5,000,000,000‎%‎
+ar	percent		-1.5E7	‎-1,500,000,000‎%‎
+ar	percent		-1.0E7	‎-1,000,000,000‎%‎
+ar	percent		-5000000.0	‎-500,000,000‎%‎
+ar	percent		-1500000.0	‎-150,000,000‎%‎
+ar	percent		-1234567.0	‎-123,456,700‎%‎
+ar	percent		-1234565.0	‎-123,456,500‎%‎
+ar	percent		-1000000.0	‎-100,000,000‎%‎
+ar	percent		-999999.9	‎-99,999,990‎%‎
+ar	percent		-500000.0	‎-50,000,000‎%‎
+ar	percent		-150000.0	‎-15,000,000‎%‎
+ar	percent		-100000.0	‎-10,000,000‎%‎
+ar	percent		-50000.0	‎-5,000,000‎%‎
+ar	percent		-15000.0	‎-1,500,000‎%‎
+ar	percent		-10000.0	‎-1,000,000‎%‎
+ar	percent		-5000.0	‎-500,000‎%‎
+ar	percent		-1500.0	‎-150,000‎%‎
+ar	percent		-1234.56	‎-123,456‎%‎
+ar	percent		-1000.0	‎-100,000‎%‎
+ar	percent		-999.9	‎-99,990‎%‎
+ar	percent		-500.0	‎-50,000‎%‎
+ar	percent		-150.0	‎-15,000‎%‎
+ar	percent		-123.0	‎-12,300‎%‎
+ar	percent		-100.0	‎-10,000‎%‎
+ar	percent		-50.0	‎-5,000‎%‎
+ar	percent		-15.0	‎-1,500‎%‎
+ar	percent		-12.0	‎-1,200‎%‎
+ar	percent		-10.0	‎-1,000‎%‎
+ar	percent		-5.0	‎-500‎%‎
+ar	percent		-3.5	‎-350‎%‎
+ar	percent		-2.5	‎-250‎%‎
+ar	percent		-1.5	‎-150‎%‎
+ar	percent		-1.2	‎-120‎%‎
+ar	percent		-1.0	‎-100‎%‎
+ar	percent		-0.5	‎-50‎%‎
+ar	percent		-0.15000000000000002	‎-15‎%‎
+ar	percent		-0.135	‎-13.5‎%‎
+ar	percent		-0.125	‎-12.5‎%‎
+ar	percent		-0.1	‎-10‎%‎
+ar	percent		-0.05	‎-5‎%‎
+ar	percent		-0.015	‎-1.5‎%‎
+ar	percent		-0.01	‎-1‎%‎
+ar	percent		-0.00831765	‎-0.831765‎%‎
+ar	percent		-0.005	‎-0.5‎%‎
+ar	percent		-0.0015	‎-0.15‎%‎
+ar	percent		-0.001	‎-0.1‎%‎
+ar	percent		-5.0E-4	‎-0.05‎%‎
+ar	percent		-1.5000000000000001E-4	‎-0.015‎%‎
+ar	percent		-1.23E-4	‎-0.0123‎%‎
+ar	percent		-1.0E-4	‎-0.01‎%‎
+ar	percent		-5.0E-5	‎-0.005‎%‎
+ar	percent		-1.5000000000000002E-5	‎-0.0015‎%‎
+ar	percent		-1.0E-5	‎-0.001‎%‎
+ar	percent		-4.9999999999999996E-6	‎-0.0005‎%‎
+ar	percent		-1.5E-6	‎-0.00015‎%‎
+ar	percent		-1.0E-6	‎-0.0001‎%‎
+ar	percent		-0.0	‎-0‎%‎
+ar	percent		1.0E-6	0.0001‎%‎
+ar	percent		1.5E-6	0.00015‎%‎
+ar	percent		4.9999999999999996E-6	0.0005‎%‎
+ar	percent		1.0E-5	0.001‎%‎
+ar	percent		1.5000000000000002E-5	0.0015‎%‎
+ar	percent		5.0E-5	0.005‎%‎
+ar	percent		1.0E-4	0.01‎%‎
+ar	percent		1.23E-4	0.0123‎%‎
+ar	percent		1.5000000000000001E-4	0.015‎%‎
+ar	percent		5.0E-4	0.05‎%‎
+ar	percent		0.001	0.1‎%‎
+ar	percent		0.0015	0.15‎%‎
+ar	percent		0.005	0.5‎%‎
+ar	percent		0.01	1‎%‎
+ar	percent		0.015	1.5‎%‎
+ar	percent		0.05	5‎%‎
+ar	percent		0.1	10‎%‎
+ar	percent		0.125	12.5‎%‎
+ar	percent		0.135	13.5‎%‎
+ar	percent		0.15000000000000002	15‎%‎
+ar	percent		0.5	50‎%‎
+ar	percent		1.0	100‎%‎
+ar	percent		1.5	150‎%‎
+ar	percent		2.5	250‎%‎
+ar	percent		3.5	350‎%‎
+ar	percent		5.0	500‎%‎
+ar	percent		10.0	1,000‎%‎
+ar	percent		12.0	1,200‎%‎
+ar	percent		15.0	1,500‎%‎
+ar	percent		50.0	5,000‎%‎
+ar	percent		100.0	10,000‎%‎
+ar	percent		123.0	12,300‎%‎
+ar	percent		150.0	15,000‎%‎
+ar	percent		500.0	50,000‎%‎
+ar	percent		999.9	99,990‎%‎
+ar	percent		1000.0	100,000‎%‎
+ar	percent		1234.56	123,456‎%‎
+ar	percent		1500.0	150,000‎%‎
+ar	percent		5000.0	500,000‎%‎
+ar	percent		10000.0	1,000,000‎%‎
+ar	percent		15000.0	1,500,000‎%‎
+ar	percent		50000.0	5,000,000‎%‎
+ar	percent		100000.0	10,000,000‎%‎
+ar	percent		150000.0	15,000,000‎%‎
+ar	percent		500000.0	50,000,000‎%‎
+ar	percent		999999.9	99,999,990‎%‎
+ar	percent		1000000.0	100,000,000‎%‎
+ar	percent		1234567.0	123,456,700‎%‎
+ar	percent		1500000.0	150,000,000‎%‎
+ar	percent		5000000.0	500,000,000‎%‎
+ar	percent		1.0E7	1,000,000,000‎%‎
+ar	percent		1.5E7	1,500,000,000‎%‎
+ar	percent		5.0E7	5,000,000,000‎%‎
+ar	percent		1.0E8	10,000,000,000‎%‎
+ar	percent		1.5E8	15,000,000,000‎%‎
+ar	percent		5.0E8	50,000,000,000‎%‎
+ar	percent		1.0E9	100,000,000,000‎%‎
+ar	percent		1.5E9	150,000,000,000‎%‎
+ar	percent		5.0E9	500,000,000,000‎%‎
+ar	percent		1.0E10	1,000,000,000,000‎%‎
+ar	percent		1.5E10	1,500,000,000,000‎%‎
+ar	percent		5.0E10	5,000,000,000,000‎%‎
+ar	percent		1.0E11	10,000,000,000,000‎%‎
+ar	percent		1.5E11	15,000,000,000,000‎%‎
+ar	percent		5.0E11	50,000,000,000,000‎%‎
+ar	percent		1.0E12	100,000,000,000,000‎%‎
+ar	percent		1.5E12	150,000,000,000,000‎%‎
+ar	percent		5.0E12	500,000,000,000,000‎%‎
+ar	scientific		-5.0E12	‎-5E12
+ar	scientific		-1.5E12	‎-1.5E12
+ar	scientific		-1.0E12	‎-1E12
+ar	scientific		-5.0E11	‎-5E11
+ar	scientific		-1.5E11	‎-1.5E11
+ar	scientific		-1.0E11	‎-1E11
+ar	scientific		-5.0E10	‎-5E10
+ar	scientific		-1.5E10	‎-1.5E10
+ar	scientific		-1.0E10	‎-1E10
+ar	scientific		-5.0E9	‎-5E9
+ar	scientific		-1.5E9	‎-1.5E9
+ar	scientific		-1.0E9	‎-1E9
+ar	scientific		-5.0E8	‎-5E8
+ar	scientific		-1.5E8	‎-1.5E8
+ar	scientific		-1.0E8	‎-1E8
+ar	scientific		-5.0E7	‎-5E7
+ar	scientific		-1.5E7	‎-1.5E7
+ar	scientific		-1.0E7	‎-1E7
+ar	scientific		-5000000.0	‎-5E6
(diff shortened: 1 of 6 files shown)


common/testData/decimal/all_locales.tsv
```diff
@@ -0,0 +1,451 @@
+locale	style	input	expected
+en_US	decimal	0.0	0
```
sffc: Your format should have more expressive number formatting options. See:

https://unicode-org.github.io/icu/userguide/format_parse/numbers/skeletons.html

https://github.com/unicode-org/icu/blob/main/icu4c/source/test/testdata/numberpermutationtest.txt

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,230 @@
...
+            this.style = style;
+            this.number = number;
+        }
+    }
+
+    private static String format(ULocale locale, Style style, double number) {
+        NumberFormat nf;
+        switch (style) {
+            case DECIMAL:
+                nf = NumberFormat.getInstance(locale);
```
sffc: You should probably move this to use NumberFormatter, and then of course to your CLDR implementation.

younies: done

common/testData/decimal/all_locales.tsv
younies: I add this comment as a subtask at the ticket [CLDR-19466](https://unicode-org.atlassian.net/browse/CLDR-19466)

sffc: Figuring out the shape of the dimensions, even if you don't have all the details right now, is important to landing this PR, because it impacts the output syntax, among other things.

sffc (changes requested):

common/testData/decimal/all_locales.tsv
younies: Okay, I’m working on it right now, and I’ll update the PR soon.

Just to clarify: by “the shape of the dimensions,” do you mean how we should represent the input for each dimension individually, or how we should determine and model all dimensions?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/numberformatter/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,189 @@
+package org.unicode.cldr.tool.numberformatter;
```
sffc: Nit: Don't create a new package when there are just a couple small files. In general this is good style but please stick with the CLDR convention of putting things in one big file. You can put `Dimensions` as a static class member of `GenerateDecimalFormatTestData`.

sffc (changes requested):
This is still not including a thoroughly explored dimension for number formatting options.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/numberformatter/GenerateDecimalFormatTestData.java
younies: I want to have Dimensions.java a class that shared with many other conformance testing, for example, when we are testing `currencies`, we would use most of the dimensions that needed for `decimal` and `units` too.

common/testData/decimal/all_locales.tsv
sffc: What I mean is that you should have a mechanism to support all number formatting options and encode them in the output testdata files.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,435 @@
...
+
+public class GenerateDecimalFormatTestData {
+
+    /**
+     * This class holds all the dimensions that we are testing against for the numberformatter (this
+     * concept can expand to another testing in the future)
+     *
+     * <p>For example, <b>locale identifiers</b> are one dimension for specs such as currency formatting
+     * and unit formatting. Test generators for those areas should take locales (and any other relevant
+     * dimensions) as inputs.
```
younies:
```suggestion
    /**
     * This class holds all the dimensions that we are testing against for the numberformatter
     *
     * <p>For example, <b>locale identifiers</b> are one dimension for decimal formatting, currency formatting
     * and unit formatting. Test generators for those areas should apply locales (and any other relevant
     * dimensions) as inputs.
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,435 @@
...
+                    }
+                }
+                return null;
+            }
+        }
+
+        // Fifth dimension: numbers
+        /**
+         * Core subset of test numbers for focused, high-signal testing.
+         */
```
younies:
```suggestion
        // TODO(younies): revisit these dimension values to ensure we are testing against a comprehensive set.
        // Fifth dimension: numbers
        /**
         * Core subset of test numbers for focused, high-signal testing.
         */
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,435 @@
...
+        public static final ImmutableSet<Double> CORE_NUMBERS = ImmutableSet.of(
+            0.0, 1.0, 1.2, 12.0, 123.0, 1234.56, 1234567.0, -1234.56, 0.000123, -0.0);
+
+        /**
+         * Returns the complete list of test numbers (powers of 10, their multiples, minimal numbers, etc.).
+         *
+         * @return a sorted set of complete test numbers
+         */
```
younies, line 186:
```suggestion
         // TODO(younies): revisit these dimension values to ensure we are testing against a comprehensive set.
        /**
         * Returns the complete list of test numbers (powers of 10, their multiples, minimal numbers, etc.).
         *
         * @return a sorted set of complete test numbers
         */
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,511 @@
...
+                        combo -> true);
+
+        // 2. Complete Numbers, Core Locales (Complete)
+        List<TestCase> allNumbersCases =
+                generateTestCases(
+                        Dimensions.getCoreLocales(), allStyles, completeNumbers, combo -> true);
+
+        // Write to JSON
+        writeJson(allLocalesCases, "all_locales.json");
+        writeJson(allNumbersCases, "all_numbers.json");
```
sffc: Issue: please include all 5 files (the blue cube, the three red rectangular prisms, and 2% of the white cube)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,511 @@
...
+        List<TestCase> coreCases =
+                generateTestCases(
+                        Dimensions.getCoreLocales(),
+                        allStyles,
+                        Dimensions.CORE_NUMBERS,
+                        combo -> true);
+
+        // Write Core to TSV (for showing)
+        writeTsv(coreCases, "all_locales.tsv");
+        writeTsv(coreCases, "all_numbers.tsv");
```
sffc: Issue: I don't like duplicating the testdata in both JSON and TSV. I would rather just come up with a good syntax that is both readable and expressive.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,511 @@
...
+                    throw new IllegalArgumentException("Unknown type: " + type);
+            }
+        }
+        return lnf.format(number).toString();
+    }
+
+    public static class TestCase {
+        public String locale;
+        public String number_type;
+        public String compactness;
```
sffc: Issue: These are not the right dimensions. For example, compactness is a subset of number type. ECMA-402 calls number type "notation".

sffc (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,511 @@
...
+        /**
+         * Core number format type dimension (standard decimal, percent, or scientific notation).
+         *
+         * @see <a href="https://www.unicode.org/reports/tr35/tr35-numbers.html#Number_Formats">LDML
+         *     Number Formats</a>
+         */
+        public enum NumberType {
+            DECIMAL("decimal"),
+            PERCENT("percent"),
+            SCIENTIFIC("scientific");
```
younies, line 137: Add compact to the numbertype

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,511 @@
...
+        }
+
+        // Second dimension: format type
+        /**
+         * Core number format type dimension (standard decimal, percent, or scientific notation).
+         *
+         * @see <a href="https://www.unicode.org/reports/tr35/tr35-numbers.html#Number_Formats">LDML
+         *     Number Formats</a>
+         */
+        public enum NumberType {
```
younies: NumberFormats

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,511 @@
...
+                    if (c.label.equals(label)) {
+                        return c;
+                    }
+                }
+                return null;
+            }
+        }
+
+        // Fourth dimension: length
+        /**
```
younies, line 151: fix comment

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,454 @@
...
+        // Third dimension: length
+        /**
+         * Format length dimension (short vs long patterns, or empty for standard).
+         *
+         * @see <a
+         *     href="https://www.unicode.org/reports/tr35/tr35-numbers.html#Compact_Number_Formats">LDML
+         *     Compact Number Formats</a>
+         */
+        public enum FormatLength {
+            EMPTY(""),
```
younies, line 159: @sffc , shall we give it a value instead of empty like `default` or `empty`, I do not like that the value in the .json or .tsv looks like nothing , what do you think ?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: There is a small amount of duplication between the files. The .json file contains the full dataset, while the .tsv file is a subset used to display a sample of the output in a table.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,454 @@
...
+
+            NumberFormat(String label) {
+                this.label = label;
+            }
+
+            public String getLabel() {
+                return label;
+            }
+
+            public static NumberFormat fromLabel(String label) {
```
macchiati: Why are you doing this? All enums have .valueOf(String source)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,454 @@
...
+            ULocale locale,
+            Dimensions.NumberFormat format,
+            Dimensions.FormatLength length,
+            double number) {
+        // TODO(younies): Check this logic against the TR35 spec to ensure we are applying the
+        // correct dimension checks.
+        com.ibm.icu.number.LocalizedNumberFormatter lnf;
+        switch (format) {
+            case DECIMAL:
+                if (length == Dimensions.FormatLength.SHORT) {
```
macchiati: Use embedded switch instead of if statements

common/testData/decimal/extended_numbers.tsv
```diff
@@ -0,0 +1,1201 @@
+locale	number_format	format_length	input	expected
```
macchiati: Hmmm. If all lines have the same number of tabs, github should display this as a table.

macchiati (changes requested):

common/testData/decimal/extended_numbers.tsv
sffc: It displays as a table in preview mode, which appears to not be available currently in the PR view.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: I think an empty cell in the TSV is abundantly clear.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: I see. I think I still don't like having both formats. I think we are pretty close to having a good TSV format so I think we should stick with that. It is also less data size on disk.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: (this is not done yet)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
macchiati: I agree; the tsv format is simpler to read and more compact also.

common/testData/decimal/extended_numbers.tsv
macchiati: ah

younies: You can also see it by clicking the three dots and choosing View file.
Screenshot 2026-05-27 at 08 02 06

sffc (changes requested):
.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: because we will need to use `.valueOf(source. toUpperCase())` and also handle empty,

but it is okay, I have removed it.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: okay

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: as we have discussed online, that the other dimension are too small to have a core subset, and I am postponing the 2% for now until completing all the test cases

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: okay, I have removed the .json files and kept all the generated tests in the .tsv files

younies (commented):
okay

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: Unfortunately, now when I go to the code section in the PR, the big .tsv files make my browser crash!

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: Sorry, you are not understanding my comment. I am expecting this PR to contain 3 nonoverlapping TSV files.

Suggestion: call them `dimensionless.tsv`, `dimensionless_all_locales.tsv`, and `dimensionless_all_numbers.tsv`. Then when we add currency and unit, we can create files specific to those features.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: If the TSV makes your browser crash, then we could split it to more files.

sffc: It looks like the TSV is too big for the preview to work, either. To remediate this:

1. The core set of locales is too big. I think it should be 4-8 diverse locales.
2. The core set of numbers is too big. I think it should be 3-6 diverse numbers.
3. If the TSV is still too big to view, then split it into multiple partitions, such as `locales_aa_ez` for locales from aa to ez, etc.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,392 @@
...
+        // 2. Complete Numbers, Core Locales (Complete) -> Write to TSV
+        List<TestCase> allNumbersCases =
+                generateTestCases(
+                        Dimensions.getCoreLocales(), allStyles, completeNumbers, combo -> true);
+        writeTsv(allNumbersCases, "all_numbers.tsv");
+
+        System.out.println("Decimal format test data generation completed.");
+    }
+
+    public static class DecimalFormatStyleParser {
```
sffc: Issue: This is not a very good name for this class. What is a "style parser"?

I still prefer representing this in code as a single dimension like `NumberFormatWithLength`, but splitting it to two columns when generating the TSV.

sffc (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: do you mean by `dimensionleass.tsv` the core data ?

This means, we will have three files:
1. `core.tsv`
2. `extended_all_locales.tsv` or `exteneded_locales.tsv`
3. `extended_all_numbers` or `extended_numbers.tsv`

right ?

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: I think the core set is very reasonable because it cover variety of numbers and locales that  you can review, now you can open `core.tsv` and also in view mode, you can see the full table

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: I suggested `dimensionless` because then when we add currency and unit, we add them as different files. This file contains all the unitless / dimensionless data. (except for percent, which is dimensionless but not unitless)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,409 @@
...
+     */
+    public static final class Dimensions {
+        private static final CLDRConfig CLDR_CONFIG = CLDRConfig.getInstance();
+        private static final Factory CLDR_FACTORY = CLDR_CONFIG.getCldrFactory();
+
+        /**
+         * Locales used for focused tests: major languages, several scripts, RTL locales, and
+         * regional shapes.
+         */
+        private static final ImmutableSet<String> CORE_LOCALES =
```
sffc, line 57: Sorry but this is simply not a good set of locales for decimal formatting.

For example, you don't include any that use a non-latin numbering system by default.

I think a list should contain at least:

- bn-BG because it uses non-latin by default
- de-CH because it has interesting separators
- Some Arabic locale, maybe ar-EG, because it uses RTL control characters
- pr-PT because it does interesting things with currencies (not necessarily needed for decimal)

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,409 @@
...
+                return label;
+            }
+        }
+
+        // TODO(younies): Revisit these dimension values to ensure comprehensive test coverage.
+        // Fifth dimension: numbers
+        /** Core subset of test numbers for focused, high-signal testing. */
+        public static final ImmutableSet<Double> CORE_NUMBERS =
+                ImmutableSet.of(
+                        0.0, 1.0, 1.2, 12.0, 123.0, 1234.56, 1234567.0, -1234.56, 0.000123, -0.0);
```
sffc: This set of numbers is too big in my opinion. You should have zero, something small, something large, something negative. You don't need 10 numbers in the core set to cover a broad cross-section. You can and should have more numbers in the bigger set. That's the whole point of how we partition these.

sffc (changes requested):

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: The TSV files are still too big to view in preview mode.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: `numbers`  is still a dimension, 

we can say in the future, `currency_core.tsv`, `currency_extended.tsv` and so on

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,409 @@
...
+     *   <li>Currencies
+     *   <li>Units
+     *   <li>Others as needed
+     * </ul>
+     *
+     * <p><b>Note:</b> Each CLDR number formatting area or spec only needs a <b>subset</b> of these
+     * dimensions. For instance, decimal formatting does not need a currency dimension; currency
+     * formatting does.
+     */
+    public static final class Dimensions {
```
sffc, line 49: Nit: Don't declare the inner classes `public` unless you actually need them across files.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: de_CH  and ar_EG are already there

I have added bn and pr-PT

sffc: Please add comments justifying why each locale is included in the core set.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: Dimensions: This class must remain public. It is referenced in TestDecimalFormat.java (which is in a different package, org.unicode.cldr.unittest). 

The test uses Dimensions.NumberFormat and Dimensions.FormatLength to parse the TSV test data and call the formatting method.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: ^ not done yet. If you cannot justify a locale, then please delete it from the core set

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: ok. I question whether that test class is necessary. You are generating the test data, so obviously the test data is tested.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: ^ please verify that the TSVs render as tables in preview mode. If they don't then they are too big and we need to make them smaller.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: It seems nice to have

| Core | All Locales | All Numbers |
|---|---|---|
| `dimensionless.tsv` | `dimensionless_all_locales.tsv` | `dimensionless_all_numbers.tsv` |
| `currency.tsv` | `currency_all_locales.tsv` | `currency_all_numbers.tsv` |
| `units.tsv` | `units_all_locales.tsv` | `units_all_numbers.tsv` |

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,403 @@
...
+            public String getLabel() {
+                return label;
+            }
+        }
+
+        // TODO(younies): Revisit these dimension values to ensure comprehensive test coverage.
+        // Fifth dimension: numbers
+        /** Core subset of test numbers for focused, high-signal testing. */
+        public static final ImmutableSet<Double> CORE_NUMBERS =
+                ImmutableSet.of(0.0, 1.2, 1234567.0, -1234.56);
```
sffc: Please include at least one number that is only a fraction, like 0.00831765

sffc (commented):
GitHub isn't showing all of the comments

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: Also, please make the last digit of the other two numbers be 5, since that is more interesting for rounding. For example, 1234565.0 and -1230.05 (interior zeros are interesting)

macchiati:
For the bigger test files, we just have 1 file per locale. That makes it easier to manage, and you should do that here

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: if you insist on the name dimensionless_core.tsv instead of just core.tsv, I will change the names

younies:
> For the bigger test files, we just have 1 file per locale. That makes it easier to manage, and you should do that here 

@macchiati done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: Yes and it is just extra layer that the test data in the right order , in the right place, for example, if there is a slight shuffle happen in any column, it should be caught here .

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
sffc: My main worry is that you don't have a good scheme for how we extend the file names when we support the other feature sets like currency and units.

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,423 @@
...
+                        "ja",
+                        // Portuguese (Portugal): Represents European Portuguese compact formats and
+                        // space separator (differing from pt_BR).
+                        "pt_PT",
+                        // Russian (Russia): Represents Cyrillic script, space separator, and
+                        // complex Slavic plural rules affecting compact formats.
+                        "ru",
+                        // Simplified Chinese (China): Represents CJK myriad grouping system with
+                        // Simplified Chinese characters.
+                        "zh_Hans_CN");
```
younies:
```suggestion
                ImmutableSet.of(
                        // Arabic (Base): Represents RTL locales that default to Latin digits (latn)
                        // in CLDR, providing a contrast with ar_EG.
                        "ar",
                        // Arabic (Egypt): Represents RTL locales with default non-latin numbering
                        // system (Eastern Arabic digits).
                        "ar_EG",
                        // Bengali (Bangladesh): Represents LTR locales with default non-latin
                        // numbering system (Bengali digits) and Indian grouping system
                        // (lakh/crore).
                        "bn",
                        // German (Germany): Represents standard European formatting using comma for
                        // decimal and dot for thousands.
                        "de",
                        // German (Switzerland): High-signal for custom separators, specifically
                        // using apostrophe (') for thousands.
                        "de_CH",
                        // English (US): Baseline LTR locale, representing standard dot for decimal
                        // and comma for thousands, and standard English compact formats.
                        "en_US",
                        // Japanese (Japan): Represents CJK locales using the myriad grouping system
                        // (4-digit) for compact formats.
                        "ja",
                        // Portuguese (Portugal): Represents European Portuguese compact formats and
                        // space separator (differing from pt_BR).
                        "pt_PT",
                        // Russian (Russia): Represents Cyrillic script, space separator, and
                        // complex Slavic plural rules affecting compact formats.
                        "ru");
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,423 @@
...
+                return label;
+            }
+        }
+
+        // TODO(younies): Revisit these dimension values to ensure comprehensive test coverage.
+        // Fifth dimension: numbers
+        /** Core subset of test numbers for focused, high-signal testing. */
+        public static final ImmutableSet<Double> CORE_NUMBERS =
+                ImmutableSet.of(0.0, 1.2, 0.00831765, 1234567.0, 1234565.0, -1234.56, -1230.05);
+
```
younies, line 179:
```suggestion
        public static final ImmutableSet<Double> CORE_NUMBERS =
                ImmutableSet.of(0.0, 1.2, 0.00831765, 1234565.0, -1230.05);

```

younies: 1- change files name decimals.tsv , decimals_modern_locales.tsv .. etc
2- reduce the core locales and numbers
3- do not split the modern_locales.tsv files

sffc (commented):
I went over this on a call with @younies.

I want the testdata to be streamlined enough that we _don't_ need to fall back into splitting by locale. What we do in datetime is to include only the modern locales; I think we can do that here, too. It will make a single TSV file with ~5000 lines, which is well within the limits of good performance on GitHub.

younies:
Now, as we are using just the modern locales and the files size are manageable, the extended test data is going to be in one file.

sffc (dismissed):

younies:
@macchiati : could you have another look please ?

common/testData/decimal/decimals_extended_numbers.tsv
```diff
@@ -0,0 +1,6301 @@
+locale	number_format	format_length	input	expected
```
macchiati, line 1: The file name is misleading. 

1. It is not "all numbers" — and that would be forbiddingly large.
2. It is also limited to modern_locales

I suggest decimals_modern_locales_more_numbers.tsv

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,423 @@
...
+                        "bn",
+                        // German (Germany): Represents standard European formatting using comma for
+                        // decimal and dot for thousands.
+                        "de",
+                        // German (Switzerland): High-signal for custom separators, specifically
+                        // using apostrophe (') for thousands.
+                        "de_CH",
+                        // English (US): Baseline LTR locale, representing standard dot for decimal
+                        // and comma for thousands, and standard English compact formats.
+                        "en_US",
```
macchiati: The canonical form is just "en"
```suggestion
                        "en", // en_US
```

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
```diff
@@ -0,0 +1,423 @@
...
+                ImmutableSet.of(0.0, 1.2, 0.00831765, 1234565.0, -1230.05);
+
+        // TODO(younies): Revisit these dimension values to ensure comprehensive test coverage.
+        /**
+         * Returns the complete list of test numbers (powers of 10, their multiples, minimal
+         * numbers, etc.).
+         *
+         * @return a sorted set of complete test numbers
+         */
+        public static Set<Double> getCompleteNumbers() {
```
macchiati: It is a larger set, not a complete set.

younies: done

tools/cldr-code/src/main/java/org/unicode/cldr/tool/GenerateDecimalFormatTestData.java
younies: done

common/testData/decimal/decimals_extended_numbers.tsv
younies: okay, I will rename it to extended_numbers

the file does not contain the modern locales

sffc (approved):

macchiati:
The folder needs a readme that documents what is in each of the files, especially since "modern-locales" is false — it is modern locales **minus** the _core_ locales that are in decimals.tsv

However, this can be in a later PR on the same ticket.

macchiati (approved):

PR #5712, opened 2026-05-12
CLDR-19464 docker: improve cache further, fix version check

srl295:
CLDR-19464

Lower priority, but improvements.
Actually caches so that we can hopefully remove the maven workaround in the future.

(Docker build goes from 1.5 to 0.5 minutes on a rebuild and doesn't hit Maven at all) 

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -301,14 +301,6 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-servertest-
           servertest-
-    - name: Cache local npm repository
-      uses: actions/cache@v4
-      with:
-        path: ~/.npm
-        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-        restore-keys: |
-          ${{ runner.os }}-node-
-          node-
     - name: Setup Docker
       uses: docker/setup-buildx-action@v4.0.0
     - name: Pull Containers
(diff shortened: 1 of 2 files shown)


tools/cldr-apps/docker/setup-config.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Trying to determine CLDR version"
 java -jar /tmp/cldr-code.jar > /dev/null || exit 1
-CLDR_VERSION=$($(java -jar /tmp/cldr-code.jar) | fgrep GEN_VERSION | cut -d= -f2 | cut -d. -f1)
+CLDR_VERSION=$(java -jar /tmp/cldr-code.jar | fgrep GEN_VERSION | cut -d= -f2 | cut -d. -f1)
```
srl295, line 5: This was an earlier issue, but it ought to make the configured CLDR version be correct.

.github/workflows/maven.yml
```diff
@@ -309,13 +309,29 @@ jobs:
...
-      working-directory: tools/cldr-apps
+      uses: docker/bake-action@v7
+      with:
+        push: false
+        source: tools/cldr-apps
+        set: |
+          *.tags=unicode-org/cldr-apps-cache:latest
+    # - name: Build Containers (should be already cached)
+    #   run: docker compose build
+    #   working-directory: tools/cldr-apps
```
srl295, line 326: this builds (ooh, unintended pun) on the earlier PR, to actually setup in a way that's cached.

In GHA testing, build time for containers went from 1.5 down to .5 minutes 

This also prevents the maven traffic that caused the original issue.

macchiati (approved):

PR #5714, opened 2026-05-12
CLDR-19451 refine language core+extension collisions

srl295:
- [X] This depends on #5731 merging first.

- add a new special check against duplicate (core,extension) tuples

CLDR-19451

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestAll.java
```diff
@@ -200,6 +200,7 @@ public TestAll() {
                     "org.unicode.cldr.unittest.TestCasingInfo",
                     "org.unicode.cldr.unittest.TestCheckAltOnly",
                     "org.unicode.cldr.unittest.TestCheckCLDR",
+                    "org.unicode.cldr.unittest.TestCheckDisplayCollisions",
```
srl295: oops

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
```diff
@@ -336,4 +371,23 @@ public void TestDurationPersonVariants() {
         TestFactory factory = makeFakeCldrFile("de", pathValuePairs);
         checkDisplayCollisions("de", pathValuePairs, factory);
     }
+
+    public void TestCollisionOfTwo() {
```
srl295, line 385: Fails (as there's no collision)

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckDisplayCollisions.java
```diff
@@ -336,4 +371,23 @@ public void TestDurationPersonVariants() {
...
+        Map<String, String> pathValuePairs =
+                ImmutableMap.of(
+                        ukRegion, "GGGGG",
+                        gmRegion, "GGGGG");
+        TestFactory factory = makeFakeCldrFile("de", pathValuePairs);
+        expectDisplayCollisionsAmong("de", pathValuePairs, factory);
+    }
+
+    public void TestCollisionOfThree() {
```
srl295, line 394: passes (collides)

macchiati: Good, thanks. Now a matter of fixing the code.

srl295: i know which line will fix it, but I think more importantly is why was `==0` changed to `<=1`, there was surely a reason and we wouldn't want to introduce a different problem.

macchiati: Just capturing the discussion: we agreed to revert to `==0`, and then check that any new errors were in fact errors. And if so, we could go ahead with the change.

srl295:
well, what do you know..

> `be [Belarusian] error   ▸Locale_Display_Names|Languages_(K-N)|K|Kurdish_►_ku-core◂      〈Kurdish〉     【】    〈курдская〉    «=»     【】    ⁅display collision⁆  ❮Error: Can't have same translation as [🔗K: Kurdish ► ku🔗]. Please change either this name or the other one.❯` https://st.unicode.org/cldr-apps/v#/be//46688feaf0df4da4

and many others

macchiati:
Yes, the test needs to be changed for the core language name, because many different languages can share the same core. The same is true for the extension {but much more rare}. What is unique is the core plus extension.

macchiati:
I have some time after the meeting today. If you want to brainstorm about fixes

srl295:
> Yes, the test needs to be changed for the core language name, because many different languages can share the same core. The same is true for the extension {but much more rare}. What is unique is the core plus extension. 
> 
> 
> 
> 

To be precise, the _check_ needs to be changed.

I was going to try to run the unit test here through the debugger and try to understand the code. Yes i can discuss for a little bit after the meeting 

**Update** Meeting happened, see notes [here](https://unicode-org.atlassian.net/browse/CLDR-19451?focusedCommentId=210141)

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
```diff
@@ -143,7 +143,13 @@ public static Type getType(String path) {
 
     static final boolean SKIP_TYPE_CHECK = true;
 
-    private final Matcher exclusions = PatternCache.get("=\"narrow\"]").matcher(""); // no matches
+    private final Matcher exclusions =
```
AEApple, line 149: Wouldn't we still need core to have collisions against other core patterns, or at least when they are assembled they can't collide instead of just being excluded?

AEApple: oh, nevermind. I see you have it as a todo for a future test.

macchiati:
How many new errors do we get as a result of this?
That isn't a blocker, just interested.

srl295:
> How many new errors do we get as a result of this? That isn't a blocker, just interested.

ok, none in -z BUILD (in `be` anyway).  They show up in FINAL_TESTING

```
# Locale        Status  ▸PPath◂ 〈Eng.Value〉   【Eng.Ex.】     〈Loc.Value〉   «fill-in»       【Loc.Ex】      ⁅error/warning type⁆   ❮Error/Warning Msg❯      Full Path       AliasedSource/Path?
/Users/srl295/src/cldr/common/main/be.xml:4645:58: error
be [Belarusian] error   ▸Timezones|North_America|Hawaii|standard-long◂  〈Hawaii-Aleutian Standard Time〉       【】    〈Гавайска-Алеуцкі стандартны час〉     «=»     【】    ⁅display collision⁆     ❮Error: Can't have same translation as [🔗Hawaii_Aleutian: standard-long🔗]. Please change either this name or the other one.❯  https://st.unicode.org/cldr-apps/v#/be//3da2b2fe3b2ac678
/Users/srl295/src/cldr/common/main/be.xml:4651:58: error
be [Belarusian] error   ▸Timezones|North_America|Hawaii_Aleutian|standard-long◂ 〈Hawaii-Aleutian Standard Time〉       【】    〈Гавайска-Алеуцкі стандартны час〉     «=»     【】    ⁅display collision⁆     ❮Error: Can't have same translation as [🔗Hawaii: standard-long🔗]. Please change either this name or the other one.❯  https://st.unicode.org/cldr-apps/v#/be//35a7e4c8c7610ba4
/Users/srl295/src/cldr/common/main/be.xml:8411:48: error
be [Belarusian] error   ▸Units|Compound_Units|10p-30|short◂     〈q{0}〉        【】    〈к{0}〉        «=»     【】    ⁅display collision⁆     ❮Error: Can't have same translation as [🔗10p3: short🔗]. Please change either this name or the other one.❯     https://st.unicode.org/cldr-apps/v#/be//2f7e5838f09af551
/Users/srl295/src/cldr/common/main/be.xml:8420:48: error
be [Belarusian] error   ▸Units|Compound_Units|10p3|short◂       〈k{0}〉        【】    〈к{0}〉        «=»     【】    ⁅display collision⁆     ❮Error: Can't have same translation as [🔗10p-30: short🔗]. Please change either this name or the other one.❯   https://st.unicode.org/cldr-apps/v#/be//38a84515d33671ef
# be [Belarusian]       Summary modern  Subtotal error: 4
```

srl295:
I think this is ready, and i'll open a separate PR for the key/type test.

No failures in BUILD (will be in FINAL_CHECK)

srl295:
@macchiati I did the core+extension test since this one was still awaiting review.

Given ckb-core == ku-core and ckb-extension == ku-extension, you get four errors, one on each path.

Not totally happy with the ergonomics here, but hopefully it's enough information to fix.
We COULD drop the errors on menu="core" since presumably that's the intended duplication. 



- `//ldml/localeDisplayNames/languages/language[@type="ku"][@menu="extension"]` Can't have same translation as <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_A_D/427ca55a261bb61c">Central Kurdish ► ckb-extension</a>. Please change either this name or the other one. (with core <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_K_N/46688feaf0df4da4">Kurdish ► ku-core</a>)
- `//ldml/localeDisplayNames/languages/language[@type="ckb"][@menu="extension"]` Can't have same translation as <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_K_N/6f4adf42ec01e056">Kurdish ► ku-extension</a>. Please change either this name or the other one. (with core <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_A_D/286c882981a7bb46">Central Kurdish ► ckb-core</a>)
- `//ldml/localeDisplayNames/languages/language[@type="ku"][@menu="core"]` Can't have same translation as <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_A_D/286c882981a7bb46">Central Kurdish ► ckb-core</a>. Please change either this name or the other one. (with extension <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_K_N/6f4adf42ec01e056">Kurdish ► ku-extension</a>)
- `//ldml/localeDisplayNames/languages/language[@type="ckb"][@menu="core"]` Can't have same translation as <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_K_N/46688feaf0df4da4">Kurdish ► ku-core</a>. Please change either this name or the other one. (with extension <a class="pathReference" href="https://st.unicode.org/cldr-apps/v#/be/Languages_A_D/427ca55a261bb61c">Central Kurdish ► ckb-extension</a>)

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
```diff
@@ -734,6 +741,99 @@ public CheckCLDR handleCheck(
...
+            String path,
+            String fullPath,
+            String value,
+            Options options,
+            List<CheckStatus> result,
+            final String menuType) {
+
+        // Perf Note: At present all languages have at most 'ku', 'ckb', 'sdh' - three pairs. So
+        // this is not too expensive.
+        // May revisit if @menu becomes more popular.
```
srl295: however, it is a pretty specific search. I expect it to be pretty quick, it only searches on the exact paths which could collide and ignores the rest.

macchiati (changes requested):
I was thinking more about this.
* It is always ok for core to collide with core, or with unmarked
* It is also ok for extension to collide with core, or unmarked
* Every core must have an extension, and vice versa
* Extension must not collide with extension IFF they have the same core.

So what that means to me is, we need to:
* make core, extension, and unmarked be in _separate_ collision groups.
* raise an error if a core doesn't have an extension or vice versa
* raise an error if extensions collide, BUT filter to only those with the same core.

macchiati:
For the last bullet, it means we can use the same processing as for other types of patterns.

It got me to thinking about how we can handle derived annotations also. And maybe optimise the value lookup. But those are a separate issues: can discuss later.

srl295:
Please take a look I think I do all of this except

> Every core must have an extension, and vice versa

And also I do: (but could drop this)

- raise an error if cores collide, BUT filter to only those with the same extension

macchiati:
What I was proposing was changing just this part of the method

```
        if (myType == Type.LANGUAGE) {
            // We remove anything with @menu from the collision set.
            // TBD
            // If menu=core + "|" + menu=extension is equal for two DIFFERENT language codes,
            // then there is a collision.
            // But we have to add a special mechanism to detect that.
            Iterator<String> iterator = paths.iterator();
            while (iterator.hasNext()) {
                String curPath = iterator.next();
                if (curPath.contains("[@menu")) {
                    iterator.remove();
                }
            }
        }
```
In this, we check the path being tested for its menu_value (simplest with XPathParts). In pseudocode:
```
before the loop
let testing_menu_value = attribute value of menu (null, core, or extension)
if core, return from method
if extension, let testing_unmarked_value be the value of the path after removing @menu. If that is null, raise error and return from method

in the loop
switch(menu_value)
core -> remove
null -> remove if testing_menu_value != null
extension -> remove if testing_menu_value != extension 
  let unmarked_value = value of loop path after removing @menu
  remove if unmarked_value != testing_unmarked_value
```

macchiati:
I think this does what I recommended:
```
if (myType == Type.LANGUAGE) {
    // Ignore all paths with menu="core"
    //    if that is the one being tested, we can exist immediately
    //    if that is the possible duplicate, remove it
    // otherwise we want to ignore all paths that don't have the same menu value
    // and for menu="extensions", ignore unless noMenut path value is identical
    String checkingMenuValue = parts.getAttributeValue(-1, "menu");
    String checkingNoMenuPathValue = null;
    if (checkingMenuValue != null) { // no action if no menu
        switch (checkingMenuValue) {
            case "core":
                return this;
            case "extensions":
                String noMenuPath =
                        parts.cloneAsThawed().removeAttribute(-1, "menu").toString();
                checkingNoMenuPathValue =
                        getResolvedCldrFileToCheck().getStringValue(noMenuPath);
        }
    }

    Iterator<String> iterator = paths.iterator();
    while (iterator.hasNext()) {
        String curPath = iterator.next();
        XPathParts curParts = XPathParts.getFrozenInstance(path);
        String curMenuValue = curParts.getAttributeValue(-1, "menu");

        if (checkingMenuValue != null) {
            switch (checkingMenuValue) {
                case "core":
                    iterator.remove();
                    break;
                case "extensions":
                    String curNoMenuPath =
                            curParts.cloneAsThawed().removeAttribute(-1, "menu").toString();
                    String curNoMenuPathValue =
                            getResolvedCldrFileToCheck().getStringValue(curNoMenuPath);
                    if (!Objects.equal(curNoMenuPathValue, checkingNoMenuPathValue)) {
                        iterator.remove();
                    }
                    break;
            }
        } else if (curMenuValue != null) {
            iterator.remove();
        }
    }
}
```

srl295:
> I think this does what I recommended: …

Thanks,
I'm working on getting unit tests improved. And I think core+extension may need to be a logical group, per type. 

What would you think of having this PR address just the 2-collision issue (so we can try it on smoketest, perhaps with redacted data), and then picking up the proposed algorithm in a subsequent PR or two?

Or even what's in the PR as it stands, which I think is a performant check very similar to what you have.

srl295:
@macchiati I created https://github.com/unicode-org/cldr/pull/5731 - can we merge that first and get it tested, and then come back to the core-extension in this PR?

I'm keeping this one open because it has the review comments about core-extension

macchiati:
> Or even what's in the PR as it stands, which I think is a performant check very similar to what you have.

I'm not sure about that which is why I coded it up. 

I do like the logical group idea. I've also been thinking that there might be a fairly straightforward way to make logical groups simpler and more performant. The basic idea is that all of the paths in group X require all of the paths in group Y to exist. There's a lot of machinery to make that happen, machinery that might not be necessary.

srl295:
@macchiati algo didn't work, i've updated it (note last 2 commits)

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDisplayCollisions.java
```diff
@@ -147,7 +150,7 @@ public static Type getType(String path) {
             PatternCache.get(
                             "(?:"
                                     + "=\"narrow\"]|"
-                                    + "^//ldml/localeDisplayNames/languages.*\\[@menu=\"(?:core|extension)\"].*$"
+                                    + "^//ldml/localeDisplayNames/languages.*\\[@menu=\"(core)\"].*$"
```
srl295, line 153: @macchiati I exclude core here

macchiati: ok

macchiati (dismissed):

AEApple (approved):
Rubberstamping following the merges from main since a couple of PRs have merged since the PR was originally created.

PR #5715, opened 2026-05-12
CLDR-19466 Implement Kotlin-based decimal format test data generator and test classes

younies:
CLDR-19466

Implement Kotlin-based decimal format test data generator and test classes

Add the Kotlin-based implementation of the decimal and compact decimal format conformance test generator and the corresponding verification tests, along with their generated TSV outputs.

## Key Changes:
- Added `GenerateDecimalFormatTestDataKt.kt` to generate expected decimal formatted outputs.
- Added `TestDecimalFormatKt.kt` to verify the Kotlin-generated test data.
- Generated expected test data files: `decimal_kt.tsv`, `decimal_all_locales_kt.tsv`, `decimal_all_numbers_kt.tsv`, and `decimal_random_5percent_kt.tsv` under common/testData/decimal/.

###Note:
Build system configurations (pom.xml changes to enable Kotlin) and Java test suite registrations must be committed in a separate step to fully enable these tests.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/kotlin/org/unicode/cldr/tool/GenerateDecimalFormatTestDataKt.kt
+++ b/tools/cldr-code/src/main/kotlin/org/unicode/cldr/tool/GenerateDecimalFormatTestDataKt.kt
@@ -1,0 +1,182 @@
+package org.unicode.cldr.tool
+
+import com.google.common.collect.ImmutableSet
+import com.ibm.icu.number.NumberFormatter
+import com.ibm.icu.util.ULocale
+import java.io.IOException
+import java.util.TreeSet
+import java.util.function.Predicate
+import org.unicode.cldr.util.CLDRConfig
+import org.unicode.cldr.util.CLDRPaths
+import org.unicode.cldr.util.CalculatedCoverageLevels
+import org.unicode.cldr.util.Factory
+import org.unicode.cldr.util.Level
+import org.unicode.cldr.util.TempPrintWriter
+import kotlin.math.pow
+
+object GenerateDecimalFormatTestDataKt {
+
+    private const val OUTPUT_SUBDIR = "decimal"
+
+    private val CLDR_CONFIG = CLDRConfig.getInstance()
+    private val CLDR_FACTORY = CLDR_CONFIG.cldrFactory
+
+    // Minimal lists
+    private val MINIMAL_LOCALES = setOf("en_US", "fr", "de_CH", "ar", "hi", "bn", "zh", "ru", "ja")
+
+    private val MINIMAL_NUMBERS = setOf(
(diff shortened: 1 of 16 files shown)


macchiati (changes requested):
Adding kotlin is a significant step, and needs justification. Also, what is the difference between the .kt.TSV and the .TSV files? Why would an application using those care about the programming language used to generate them

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDecimalFormat.java
srl295, line 1: Please don't add more TestFmwk tests.

Add a new JUnit test (not under the unittest package but under the package of the file under test.)

tools/cldr-code/src/test/kotlin/org/unicode/cldr/unittest/TestDecimalFormatKt.kt
srl295, line 1: can you make a JUnit test in Kotlin?

srl295 (commented):
structure generally OK but please use JUnit

tools/cldr-code/src/main/kotlin/org/unicode/cldr/tool/GenerateDecimalFormatTestDataKt.kt
```diff
@@ -0,0 +1,182 @@
...
+                        val expected = format(locale, style, number)
+                        pw.println("$localeStr\t${style.label}\t$number\t$expected")
+                    }
+                }
+            }
+        }
+    }
+
+    @JvmStatic
+    fun main(args: Array<String>) {
```
mihnita, line 120: This is very unfriendly for testing.

Hard to trigger, easy to forget about it, completely undocumented.

Can we make this a regular test class that runs tests?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDecimalFormat.java
macchiati: > Please don't add more TestFmwk tests.

That has not been decided; please add this to the CLDR Ops Agenda

tools/cldr-code/src/test/kotlin/org/unicode/cldr/unittest/TestDecimalFormatKt.kt
macchiati: That's moot for now, since there has been no proposal to add Kotlin at all.

srl295:
> Adding kotlin is a significant step, and needs justification. 

CLDR-19469

PR #5708, opened 2026-05-11
CLDR-19272 New Locale: Kaitag [xdq]

conradarcturus:
CLDR-19272

The usual, adding a new locale, updating supplemental files, and rebuilding them with  `mvn package -DskipTests=true &&  java -jar tools/cldr-code/target/cldr-code.jar ConvertLanguageData &&  java -jar tools/cldr-code/target/cldr-code.jar GenerateLikelySubtags`.

- [x] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #5718, opened 2026-05-12
CLDR-19272 CheckExemplars should detect and report empty string {}

srl295:
CLDR-19272 (but can add a new one if needed)

#5708 failed exemplar checks with a crash (stack trace). 

This changes that to:

> `xdq [Kaitag]    error   ▸Core_Data|Alphabetic_Information|Characters_in_Use|Others:_punctuation◂        〈[\- ‐‑ – — , ; \: ! ? . … '‘’ "“” ( ) \[ \] § @ * / \& # † ‡ ′ ″]〉       【】    〈[\- – — , ; \: ! ? . … ' ‚ " „ « » ( ) \[ \] { } @ * / \& # №]〉  «=»     【】    ⁅illegal characters in exemplars⁆       ❮Error: Exemplar has an empty string: {}. Ensure that any curly braces are quoted: \{ \}❯   https://st.unicode.org/cldr-apps/v#/xdq//6cf943e652b01478`

Note that the issue was: `{ }` which normalized to `{}` which should instead have been `\{ \}`.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
@@ -12,6 +12,7 @@ import java.util.BitSet;
 import java.util.Comparator;
 import java.util.List;
 import org.unicode.cldr.test.CheckCLDR.CheckStatus.Subtype;
+import org.unicode.cldr.test.CheckCLDR.CheckStatus.Type;
 import org.unicode.cldr.util.CLDRFile;
 import org.unicode.cldr.util.ComparatorUtilities;
 import org.unicode.cldr.util.ExemplarSets;
@@ -277,6 +278,16 @@ public class CheckExemplars extends FactoryCheckCLDR {
     private void checkMixedScripts(String title, UnicodeSet set, List<CheckStatus> result) {
         BitSet s = new BitSet();
         for (String item : set) {
+            if (item.isEmpty()) {
+                result.add(
+                        new CheckStatus()
+                                .setCause(this)
+                                .setMainType(Type.Error)
+                                .setSubtype(Subtype.illegalCharactersInExemplars)
+                                .setMessage(
+                                        "Exemplar has an empty string: {}. Ensure that any curly braces are quoted: \\{ \\}"));
+                continue; // skip this string
+            }
             int script = UScript.getScript(item.codePointAt(0));
             if (script != UScript.COMMON && script != UScript.INHERITED) {
                 s.set(script);


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
```diff
@@ -277,6 +278,16 @@ private void checkParse(String path, String value, List<CheckStatus> result) {
...
         BitSet s = new BitSet();
         for (String item : set) {
+            if (item.isEmpty()) {
+                result.add(
+                        new CheckStatus()
+                                .setCause(this)
+                                .setMainType(Type.Error)
+                                .setSubtype(Subtype.illegalCharactersInExemplars)
+                                .setMessage(
+                                        "Exemplar has an empty string: {}. Ensure that any curly braces are quoted: \\{ \\}"));
```
btangmu, line 288: Is this explanation always true in this context? I mean, does `item.isEmpty()` always mean the ST user would see curly braces in the value?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckExemplars.java
srl295: Yes, `{}` is the required syntax to produce this.

The user presumably intended `{` and `}` in the punctuation exemplars.

macchiati:
Note: this illustrates why using bulk upload can be problematic. Unless people understand UnicodeSet syntax, they are likely to get it wrong.

@conradarcturus , was the xdq.xml generated from the 'Core data needed for new CLDR locale' form? If so, we need to modify the generation to not use UnicodeSet syntax, but rather to use the syntax that is used in the survey tool.

srl295:
> Note: this illustrates why using bulk upload can be problematic. Unless people understand UnicodeSet syntax, they are likely to get it wrong.

This isn't a bulk upload, though. 

> @conradarcturus , was the xdq.xml generated from the 'Core data needed for new CLDR locale' form? If so, we need to modify the generation to not use UnicodeSet syntax, but rather to use the syntax that is used in the survey tool.

The form was filled out with: `, ; : ! ? . «» ( ) [ ] * /` So perhaps Conrad added the curly braces.

The "generation" is to XML, so I would think it should use UnicodeSet syntax as that's what's in the XML, such as `\[ \]` which were correctly escaped.

**Edit** Perhaps instead we should take punctation exemplars **out** of core, that would mean people would use the SurveyTool to add them.

macchiati:
Let me be clearer.
1. Whatever is checked into XML has to be with UnicodeSet syntax. That is a hard requirement.
2. The people filling out the form should **not** be expected to know the UnicodeSet syntax; it is too unfamiliar to non-devs; and too easy to make mistakes.
3. Instead, they should use a simple syntax based on https://cldr.unicode.org/translation/core-data/exemplars, with spaces separating each grapheme cluster. That is, 
    * a b c č ć d dž đ e f g h i j k l lj m n nj o p r s š t u v z ž [ ] { }
    * _and **not**_
    * [a b c č ć d {dž} đ e f g h i j k l {lj} m n {nj} o p r s š t u v z ž \\[ \\] \\{ \\}]
4. **We** need to convert that into value UnicodeSet syntax. (The simplest way is to paste it into a vote on smoketest, and then pickup the "internal format" in the examples).
5. The vast majority of locales will not need the 'escape' characters ➖, ❰, and ❱ at all. 
    * Only really big sets like CJK really need to use the ➖
    * Only invisibles need the ❰ and ❱

srl295:
> Let me be clearer.
> 
> 1. Whatever is checked into XML has to be with UnicodeSet syntax. That is a hard requirement.

Correct. That was done by Conrad (I presume), with some minor errors. The error message was unhelpful before, with a stack trace if there was a 0-length string included. With this PR it explains what the problem is.

> 2. The people filling out the form should **not** be expected to know the UnicodeSet syntax; it is too unfamiliar to non-devs; and too easy to make mistakes.

They didn't fill out UnicodeSet format. They filled out `, ; : ! ? . «» ( ) [ ] * /`. 

> 3. Instead, they should use a simple syntax based on https://cldr.unicode.org/translation/core-data/exemplars, with spaces separating each grapheme cluster. That is,
>    
>    * a b c č ć d dž đ e f g h i j k l lj m n nj o p r s š t u v z ž [ ] { }
>    * _and **not**_
>    * [a b c č ć d {dž} đ e f g h i j k l {lj} m n {nj} o p r s š t u v z ž \[ \] \{ \}]
> 4. **We** need to convert that into value UnicodeSet syntax. (The simplest way is to paste it into a vote on smoketest, and then pickup the "internal format" in the examples).

We do convert it, that's where the original PR came from.

Recommending a simpler syntax is a great idea. 

> 5. The vast majority of locales will not need the 'escape' characters ➖, ❰, and ❱ at all.
>    
>    * Only really big sets like CJK really need to use the ➖
>    * Only invisibles need the ❰ and ❱

I filed https://unicode-org.atlassian.net/browse/CLDR-19470

What would you think of having the core form only ask for the main and aux exemplars? And let the punctuation be entered in the SurveyTool?

macchiati:
> What would you think of having the core form only ask for the main and aux exemplars? And let the punctuation be entered in the SurveyTool?

That sounds reasonable; less work on our part and fewer chances for mistakes.

conradarcturus:
oh wow! Thanks for figuring this out and fixing it.

Yea I have to manually convert inputted values to UnicodeSet syntax -- I'd love to understand the rules on what needs to be escape and what doesn't so I don't have to manually convert them and hit this bugs :p for some submissions it really slows me down.

srl295:
> oh wow! Thanks for figuring this out and fixing it.
> 
> 
> 
> Yea I have to manually convert inputted values to UnicodeSet syntax -- I'd love to understand the rules on what needs to be escape and what doesn't so I don't have to manually convert them and hit this bugs :p for some submissions it really slows me down.

well, the rules are linked here- but not HAVING to convert them is even better..

PR #5719, opened 2026-05-12
CLDR-13465 TestAll should run all tests

srl295:
- add a test comparing TestAll's list with what's on disk
- use ClassGraph for introspection
- for now, still use the old list

CLDR-13465

ALLOW_MANY_COMMITS=true

Analysis:

### Wasn't in TestAll, but had to mark with `@Disabled` because it broke. 

These are remanded to https://unicode-org.atlassian.net/browse/CLDR-19472

- TestAlt
- TestBcp47Numbers
- TestCheckDisplayCollisions
- TestCldrFileWrite
- TestCldrResolver (didn't fail just super slow)
- TestCollators “See Ticket CLDR-8288”
- TestCompatibility
- TestCoverage
- TestDateOrder
- TestFmwkPlus (framework tests)
- TestIcuPersonNames
- TestLanguageGroup
- TestLsrvCanonicalizer
- TestPerf
- TestPluralRuleGeneration
- TestUExtension
- TestURLs
- TestXMLSource
- UnicodeSetPrettyPrinterTest
- XLocaleDistanceTest
- XLocaleMatcherTest

### Not in TestAll, so previously not tested.

- TestBcp47Transforms
- TestCheckWidths
- TestMetadataEquals
- TestPathsModule
- TestUnicodeProperty
- TestVettingViewerCategories

### In TestAll's list, but didn't exist (ignored)

- org.unicode.cldr.unittest.TestKeyboardModifierSet
- org.unicode.cldr.unittest.TestLdml2ICU
- org.unicode.cldr.unittest.TestLocaleCanonicalizer
- org.unicode.cldr.unittest.TestMetadata
- org.unicode.cldr.unittest.TestNumbers

### Shouldn't be in TestAll, run separately (separate group)

- org.unicode.cldr.api.AllTests

### DIFF
--- a/tools/cldr-code/pom.xml
+++ b/tools/cldr-code/pom.xml
@@ -38,6 +38,12 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- this is for tests, but we want cldr-apps to use it too. -->
+		<dependency>
+			<groupId>io.github.classgraph</groupId>
+			<artifactId>classgraph</artifactId>
+		</dependency>
+
 		<dependency>
 			<groupId>com.google.code.gson</groupId>
 			<artifactId>gson</artifactId>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/ShimmedMain.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/ShimmedMain.java
@@ -1,10 +1,16 @@
 package org.unicode.cldr.util;
 
+import io.github.classgraph.ClassGraph;
+import io.github.classgraph.ClassInfo;
+import io.github.classgraph.ScanResult;
 import java.io.PrintWriter;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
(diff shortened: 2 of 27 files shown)


macchiati (commented):
Looks good to me

macchiati:
BTW, I think there are some subclasses of TestFwrk; does your method handle those?

srl295:
> BTW, I think there are some subclasses of TestFwrk; does your method handle those?

Yes, it will only exclude direct TestGroup subclasses  (such as TestAll itself).

srl295:
I ended up having to mark many of these `@Disabled` - we an re-enable them separately.

macchiati (approved):
It looks like 16 tests are now covered, but disabled; which is approximately half of the tests that were ignored.

I'm approving, but can you file an infrastructure ticket that lists those, and is to investigate them to see which we should fix or should delete.

srl295:
> It looks like 16 tests are now covered, but disabled; which is approximately half of the tests that were ignored.
> 
> I'm approving, but can you file an infrastructure ticket that lists those, and is to investigate them to see which we should fix or should delete.

I'll make it a subtask 

It also took 44 minutes for them all to fail.

PR #5729, opened 2026-05-15
CLDR-19455 Dashboard/Abstain test with minimal change

btangmu:
-In BallotBoxXMLSource.setValueFromResolver, skip vote resolution if xpd is null or empty, even if voteLoadingContext is not ORDINARY_LOAD_VOTES

-This is expected to fail TestSTFactory.TestSparseVote which expects INHERITANCE_MARKER but gets null

CLDR-19455

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/BallotBoxXMLSource.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/BallotBoxXMLSource.java
@@ -61,8 +61,7 @@ public class BallotBoxXMLSource<T> extends DelegateXMLSource {
          * TODO: move this into the caller somehow. Maybe just use a special resolver that just
          * calls into diskData.
          */
-        if (voteLoadingContext == STFactory.VoteLoadingContext.ORDINARY_LOAD_VOTES
-                && (xpd == null || xpd.isEmpty())) {
+        if (xpd == null || xpd.isEmpty()) {
             /*
              * Skip vote resolution
              */


btangmu:
@srl295 The test failure is what I expected from running locally:

Error:    TestSTFactory.TestSparseVote:416->expect:111 ERR: (TestSTFactory.java:416): Expected '↑↑↑': fr_BE://ldml/localeDisplayNames/keys/key[@type="calendar"] ='<NULL>', ( did NOT VOTE) {bailey: “calendrier” baseline: {null, missing}, {orgToVotes: , totals: {}, conflicted: []}, sameVotes: [], O: null, N: null, totals: {}, winning: {↑↑↑, missing}} ==> expected: <↑↑↑> but was: <<NULL>>

Note that VoteResolver.getWinningValue returns '↑↑↑', however, VoteResolver.getWinningStatus returns 'missing'.

Possible solutions:

1. This 'missing' status could be taken into account by BallotBoxXMLSource.setValueFromResolver, causing it to skip delegate.putValueAtPath just as if VoteResolver.getWinningValue returned null
2. VoteResolver.getWinningValue could return null instead of '↑↑↑' when there are no votes

I'm not sure which is best; I'll investigate the consequences. I'm also adding code to identify Abstain explicitly in the http request and on the back end. There are far too many implicit assumptions that a null, empty, or blank value (possibly changed from blank to empty by DAIP) is an abstention. As we discussed, "abstention" also has two meanings: (1) the event and handling when a user clicks the abstain button; and (2) the absence of any vote. They should be equivalent in their effects -- voting then abstaining should result in the same state as never voting at all -- but as it stands, currently that's not entirely true, hence this ticket!

srl295:
> @srl295 The test failure is what I expected from running locally:
> 
> 
> 
> Error:    TestSTFactory.TestSparseVote:416->expect:111 ERR: (TestSTFactory.java:416): Expected '↑↑↑': fr_BE://ldml/localeDisplayNames/keys/key[@type="calendar"] ='<NULL>', ( did NOT VOTE) {bailey: “calendrier” baseline: {null, missing}, {orgToVotes: , totals: {}, conflicted: []}, sameVotes: [], O: null, N: null, totals: {}, winning: {↑↑↑, missing}} ==> expected: <↑↑↑> but was: <<NULL>>
> 
> 
> 
> Note that VoteResolver.getWinningValue returns '↑↑↑', however, VoteResolver.getWinningStatus returns 'missing'.
> 
> 
> 
> Possible solutions:
> 
> 
> 
> 1. This 'missing' status could be taken into account by BallotBoxXMLSource.setValueFromResolver, causing it to skip delegate.putValueAtPath just as if VoteResolver.getWinningValue returned null
> 
> 2. VoteResolver.getWinningValue could return null instead of '↑↑↑' when there are no votes
> 
> 
> 
> I'm not sure which is best; I'll investigate the consequences. I'm also adding code to identify Abstain explicitly in the http request and on the back end. There are far too many implicit assumptions that a null, empty, or blank value (possibly changed from blank to empty by DAIP) is an abstention. As we discussed, "abstention" also has two meanings: (1) the event and handling when a user clicks the abstain button; and (2) the absence of any vote. They should be equivalent in their effects -- voting then abstaining should result in the same state as never voting at all -- but as it stands, currently that's not entirely true, hence this ticket!

Does the test test before and after voting and abstaining? 
Or option 3 Maybe the test case is just wrong and should let no value be the right answer here.

btangmu:
> Does the test test before and after voting and abstaining?

Yes, and it expects the current behavior: null before voting, ^^^ after voting and abstaining. I think the test is wrong and it should expect the same before and after.

> Or option 3 Maybe the test case is just wrong and should let no value be the right answer here.

My sentiments exactly!

srl295:
> > Does the test test before and after voting and abstaining?
> 
> Yes, and it expects the current behavior: null before voting, ^^^ after voting and abstaining. I think the test is wrong and it should expect the same before and after.
> 
> > Or option 3 Maybe the test case is just wrong and should let no value be the right answer here.
> 
> My sentiments exactly!

why did you close it? maybe we just need to fix the test case?

btangmu:
> why did you close it? maybe we just need to fix the test case?

I'll make a PR shortly that fixes the test along with other improvements.

PR #5730, opened 2026-05-15
CLDR-19455 Make vote plus abstain equivalent to not voting

btangmu:
-Revise BallotBoxXMLSource.setValueFromResolver to recognize Status.missing

-Fix TestSTFactory.TestSparseVote

-Add isAbstain to http request, making Abstain explicitly distinct from Vote/Submit and not only implied by null or empty value

-The value of isAbstain does not yet change the behavior of Survey Tool; it serves as a basis for debugging and future improvement

-Empty values are legal for some paths; how vetters (should) input them in ST needs clarification, which should be facilitated by making explicit distinctions in the code

-New value for STFactory.VoteLoadingContext: SINGLE_ABSTENTION, distinct from SINGLE_VOTE and ORDINARY_LOAD_VOTES

-Fix a compiler warning for expectString != ANY

-Comments, including TODO

CLDR-19455

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/src/main/java/org/unicode/cldr/web/BallotBoxXMLSource.java
+++ b/tools/cldr-apps/src/main/java/org/unicode/cldr/web/BallotBoxXMLSource.java
@@ -3,6 +3,7 @@ package org.unicode.cldr.web;
 import java.util.Date;
 import java.util.logging.Logger;
 import org.unicode.cldr.util.CLDRFile.DraftStatus;
+import org.unicode.cldr.util.CldrUtility;
 import org.unicode.cldr.util.VoteResolver;
 import org.unicode.cldr.util.VoteResolver.Status;
 import org.unicode.cldr.util.XMLSource;
@@ -57,7 +58,7 @@ public class BallotBoxXMLSource<T> extends DelegateXMLSource {
          * result in "no votes", "skip vote resolution", failure to get the right winning value, possibly inherited.
          */
 
-        /**
+        /*
          * TODO: move this into the caller somehow. Maybe just use a special resolver that just
          * calls into diskData.
          */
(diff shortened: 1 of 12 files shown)


btangmu:
This doesn't complete the ticket. There are TODO comments. Also, there is a different possible way to fix BallotBoxXMLSource.setValueFromResolver, illustrated in the previous draft PR https://github.com/unicode-org/cldr/pull/5729 . The fix here for that function is more conservative, being more limited to fixing what is clearly a bug, with potentially fewer side-effects (for better or for worse) than the other possibility.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/BallotBoxXMLSource.java
```diff
@@ -71,6 +72,16 @@ VoteResolver<String> setValueFromResolver(
...
             value = resolver.getWinningValue();
+            Status status = resolver.getWinningStatus();
+            // If there are no votes, the winning value may be INHERITANCE_MARKER and the status may
+            // be missing. This should not lead to delegate.putValueAtPath(INHERITANCE_MARKER),
+            // since that would be inconsistent with the initial state when there are no votes, vote
+            // resolution is skipped, and diskData.getValueAtDPath returns null (see above). For
+            // example, a vote followed by an abstention by the same user should result in no
+            // change, just as if the user had never voted.
+            if (value == CldrUtility.INHERITANCE_MARKER && status == Status.missing) {
+                value = null;
```
srl295, line 83: Ok.

srl295 (approved):

srl295:
I'm approving because I think the clarity and defensive design is worthwhile. It seems complex but I'm ok with it.

tools/cldr-apps/src/main/java/org/unicode/cldr/web/BallotBoxXMLSource.java
```diff
@@ -57,7 +58,7 @@ VoteResolver<String> setValueFromResolver(
          * result in "no votes", "skip vote resolution", failure to get the right winning value, possibly inherited.
```
btangmu, line 58: The comment above ("Do not skip...") is old, and looks like it could have been written by me more than three years ago. It may no longer be applicable, but it suggests we should be cautious about skipping vote resolution, although that is another possible solution worth investigating, as an alternative to taking `resolver.getWinningStatus` into account.

btangmu:
> I'm approving because I think the clarity and defensive design is worthwhile. It seems complex but I'm ok with it.

Thanks! Let's give Mark a chance to review as well...

macchiati (commented):
I'm not as familiar with this code. 

If you both think it works, then please describe a case we can use in smoketest to verify that it works (something that fails in production).

btangmu:
> describe a case we can use in smoketest to verify that it works

It's the bug described in the ticket: look at Missing count, then vote, then abstain. The Missing count should be back to what it was. This PR only fixes half of the bug, that is the calculation. It's still necessary to refresh the dashboard manually after abstain to see the Missing count get updated; a subsequent PR will need to fix the refresh bug. I'll go ahead and merge this. Thanks!

PR #5739, opened 2026-05-19
CLDR-19455 Refresh Dashboard after Abstain: work-around

btangmu:
-Reload the entire Dashboard after user Abstains

-Wait 2 seconds for reload, to avoid conflict with cldrDashContext.updatePath

-This is a crude work-around, pending row-specific update that should be possible with cldrDashContext.updatePath

-Comments, including TODO

CLDR-19455

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrDashData.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDashData.mjs
@@ -340,6 +340,10 @@ function convertData(json, locale) {
  *               containing notifications for a single path (old format)
  */
 function updatePath(dashData, json) {
+  if (!dashData) {
+    console.warn("dashData is null in cldrDashDarta.updatePath; skipping.");
+    return;
+  }
   try {
     if (json.xpstrid in dashData.pathIndex) {
       // We already have an entry for this path
--- a/tools/cldr-apps/js/src/esm/cldrDashContext.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrDashContext.mjs
@@ -145,11 +145,28 @@ function updateWithCoverage(newLevel) {
   }
 }
 
+function updateAfterAbstain() {
+  if (dashVisible) {
+    // Crude work-around: reload entire dashboard.
+    // Delay is to avoid interference with single-row dash update (cldrDashContext.updatePath).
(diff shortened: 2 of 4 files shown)


tools/cldr-apps/js/src/esm/cldrDashData.mjs
```diff
@@ -340,6 +340,10 @@ function convertData(json, locale) {
  *               containing notifications for a single path (old format)
  */
 function updatePath(dashData, json) {
+  if (!dashData) {
+    console.warn("dashData is null in cldrDashDarta.updatePath; skipping.");
```
btangmu, line 344: This should not happen any more, but it's best to return early and warn rather than crash on null if it does happen. It happened when cldrDashContext.updateAfterAbstain didn't have the 2-second delay to avoid interfering with cldrDashContext.updatePath

AEApple:
Should I add a known issue if we are reloading the entire dashboard when someone abstains so that the vetters know that it may take longer when they vote to abstain?

btangmu:
> Should I add a known issue if we are reloading the entire dashboard when someone abstains so that the vetters know that it may take longer when they vote to abstain?

Maybe; I'm not sure if it's worth your trouble.

AEApple:
> > Should I add a known issue if we are reloading the entire dashboard when someone abstains so that the vetters know that it may take longer when they vote to abstain?
> 
> Maybe; I'm not sure if it's worth your trouble.

Does that mean it won't be noticeable to vetters?

btangmu:
> Does that mean it won't be noticeable to vetters?

Two seconds after they abstain, the Dashboard will reload and show the updated Missing count. They might notice it, but I doubt they would be alarmed or greatly inconvenienced by it.

AEApple:
> > Does that mean it won't be noticeable to vetters?
> 
> Two seconds after they abstain, the Dashboard will reload and show the updated Missing count. They might notice it, but I doubt they would be alarmed or greatly inconvenienced by it.

SG! I just wasn;t sure if it would affect other parts of the ST performance (e.g. voting)

srl295 (approved):
Good update for now.

PR #5740, opened 2026-05-19
CLDR-19451 progress meter on console check

srl295:
CLDR-19451

just adds a progress bar to the console check
could be reused for other long running process such as charts

```
# ConsoleCheckCLDR: 000% [----------] 0/8
# ConsoleCheckCLDR: 088% [========--] 7/8
# ConsoleCheckCLDR: 100% [==========] 8/8
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
@@ -61,6 +61,7 @@ import org.unicode.cldr.util.PathHeader;
 import org.unicode.cldr.util.PathHeader.SectionId;
 import org.unicode.cldr.util.PathUtilities;
 import org.unicode.cldr.util.PatternCache;
+import org.unicode.cldr.util.ProgressTracker;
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
```diff
@@ -542,9 +543,13 @@ public static void main(String[] args) throws Throwable {
             stream = locales.parallelStream();
         }
 
+        final ProgressTracker progress =
```
macchiati, line 546: Do you need a try (ProgressTracker progress =...) {<code>}

like opening files?

Not a blocker

macchiati (approved):
Looks reasonable

tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
srl295: It could be - that would re-indent the entire file.  But it can be used that way, yes.

PR #5741, opened 2026-05-19
CLDR-13465 fix ShimmedMain.findAllTests to include subclasses

srl295:
- subclasses (such as TestFmwkPlus) were apparently actually skipped
- also, make it an error to run a TestFmwk run but no actual tests are run
- also, update TestFmwk printouts to skip utility functions in the stack trace

CLDR-13465

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
@@ -636,9 +636,16 @@ public class TestFmwk extends AbstractTestLog {
         if (errorCount > 0) {
             writeStepSummary("- :x: " + errorCount + " TEST(S) FAILED");
             localParams.log.println("\n<< " + errorCount + " TEST(S) FAILED >>");
-        } else {
-            localParams.log.println("\n<< ALL TESTS PASSED >>");
-            writeStepSummary("- :white_check_mark: ALL TESTS PASSED");
+        } else if (params.testCount > 0) {
+            localParams.log.println("\n<< ALL " + params.testCount + " TESTS PASSED >>");
+            writeStepSummary("- :white_check_mark: ALL \" + params.testCount + \" TESTS PASSED");
+        } else if (params.listlevel == 0) {
+            // unless in list mode
+            final String ALL_TESTS_SKIPPED =
+                    "No tests were run: either nothing matched the filters, all tests were skipped, or they were never written. This is an error.";
+            writeStepSummary("- :x: " + ALL_TESTS_SKIPPED);
+            localParams.log.println("\n ERROR: " + ALL_TESTS_SKIPPED);
+            return 1;
         }
 
         if (prompt) {
@@ -658,12 +665,14 @@ public class TestFmwk extends AbstractTestLog {
 
     public int runTests(TestParams _params, String[] tests) {
         int ec = 0;
+        // int runCount = 0;
 
         StringBuffer summary = null;
         try {
             if (tests.length == 0 || tests[0] == null) { // no args
                 _params.init();
                 resolveTarget(_params).run();
+                // runCount++;
                 ec = _params.errorCount;
             } else {
                 for (int i = 0; i < tests.length; ++i) {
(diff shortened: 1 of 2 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/ShimmedMain.java
```diff
@@ -91,20 +91,33 @@ public static String[] findAllTests(final Package inPackage) {
...
             return allTestClasses.toArray(new String[allTestClasses.size()]);
         }
     }
+
+    private static void processScanResult(
+            final Package inPackage,
+            final Set<String> allTestClasses,
+            ScanResult scanResult,
+            final Class<?> clazz) {
+        for (ClassInfo info : scanResult.getClassInfo(clazz.getCanonicalName()).getSubclasses()) {
```
srl295, line 107: This is the main fix.  Class.getSubclasses gets *all* subclasses.  `scanResult.getSubclasses(TestFmwk.class)`  only returned direct subclasses.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
```diff
@@ -636,9 +636,16 @@ public int run(String[] args, PrintWriter log) {
         if (errorCount > 0) {
             writeStepSummary("- :x: " + errorCount + " TEST(S) FAILED");
             localParams.log.println("\n<< " + errorCount + " TEST(S) FAILED >>");
-        } else {
-            localParams.log.println("\n<< ALL TESTS PASSED >>");
-            writeStepSummary("- :white_check_mark: ALL TESTS PASSED");
+        } else if (params.testCount > 0) {
+            localParams.log.println("\n<< ALL " + params.testCount + " TESTS PASSED >>");
```
btangmu, line 640: For the future, it would be ideal if the category of test were indicated in this message. "ALL" doesn't really mean "all", it means "all within a certain category". Typically when we really run all tests, we see "ALL TESTS PASSED" preceded or followed by a failure for a test in a different category.

tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
```diff
@@ -694,6 +703,11 @@ public int runTests(TestParams _params, String[] tests) {
             _params.log.println("\nencountered a test failure, exiting\n" + e);
             e.printStackTrace(_params.log);
         }
+        // if (runCount == 0) {
+        //     ec++;
+        //     writeStepSummary("- :x: No tests were run. This is an error.");
+        //     _params.log.println("\nNo tests were run. This is an error.\n");
+        // }
```
btangmu, line 710: don't need the commented-out code?

btangmu (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
srl295: Will remove next.

tools/cldr-code/src/main/java/org/unicode/cldr/icu/dev/test/TestFmwk.java
srl295: > Typically when we really run all tests, we see "ALL TESTS PASSED" preceded or followed by a failure for a test in a different category.

I'm not quite sure what you mean by a 'category'.


Are you talking about ALL TESTS PASSED in cldr-code but having some tests in the cldr-apps project fail? 

Could you give an example?

btangmu: Yes, your example is what I meant: cldr-code and cldr-apps are the "categories"

srl295: Ok. Those are separate modules and never run together. 
But we can de emphasize the ALL

PR #5743, opened 2026-05-20
CLDR-19496 Prevent NO_WINNING_VALUE from showing as candidate item

btangmu:
-Skip NO_WINNING_VALUE in DataPage.DataRow.addItem, CheckCLDR.check, ExampleGenerator.getExampleHtmlExtended, and cldrInfo.getValueAndClass

-Comments

CLDR-19496

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrInfo.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrInfo.mjs
@@ -401,8 +401,8 @@ function getValueAndClass(theRow, item) {
     item?.value === cldrSurvey.INHERITANCE_MARKER
       ? theRow?.inheritedDisplayValue
       : item?.value;
-  if (!displayValue) {
-    displayValue = ""; // not "undefined"
+  if (!displayValue || cldrTable.NO_WINNING_VALUE === displayValue) {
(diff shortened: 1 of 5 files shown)


srl295 (approved):
OK for now. I think could go back a bit and revisit whether NO_WINNING_VALUE needs to be there, since as you noted there's status=missing. Then we can go back to just returning null, and checking for null.

I'm not sure what value propagating NO_WINNING_VALUE to more code sites gets us ultimately. `CLDRFile.getStringValue()` for example can return null which is what will happen when a path with no winning values gets integrated and shipped.

btangmu:
> revisit whether NO_WINNING_VALUE needs to be there

I think it's valuable to distinguish NO_WINNING_VALUE, which derives only from VoteResolver, from null values that can come from practically anywhere, including bugs. There are still unsolved bugs that have existed for years (such as "there is INHERITANCE_MARKER without inheritedValue", and DataPage.populateFromThisXpath setting isExtraPath = true when it sees a null value, regardless of whether the path is really an extra path). It's too easy to misinterpret null as meaning one thing when it really means something else.

PR #5744, opened 2026-05-20
CLDR-19489 add New category

srl295:
CLDR-19489

image_720

- refactor CheckoutArchive so we can run it on server
- run CheckoutArchive at startup
- track a factory for the last-vote-value files

- Note, does NOT support unhiding the 'New' category. Once you have voted for it, it's no longer New.

Also, while touching the loading code:

- make the abstracts  (The Wikipedia descriptions of languages etc) only load in production. they are often broken in smoketest (there’s another issue [CLDR-18518](https://unicode-org.atlassian.net/browse/CLDR-18518) for that)

Prints the warning: `CLDR_LOAD_ABSTRACTCACHE=false and we aren't on PRODUCTION, skipping abstract cache`

You can set CLDR_LOAD_ABSTRACTCACHE=true to override and load them locally

Also, since this affects server startup:
- Fail-fast when the webdriver can't login.  More improvements will be in CLDR-16859

ALLOW_MANY_COMMITS=true

[CLDR-18518]: https://unicode-org.atlassian.net/browse/CLDR-18518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
+++ b/tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
@@ -1413,20 +1413,25 @@ public class SurveyDriver {
 
         result = sanityLogin(SurveyDriverCredentials.SPECIAL_USER_TC) && result;
 
-        /// ----- OK. now start the steps
-        /// 10. CLA
-        result = testCla() && result;
+        // Exit quickly if we can't even login.
+        if (result) {
+            /// ----- OK. now start the steps
+            /// 10. CLA
+            result = testCla() && result;
 
-        /// 11.- Voting
-        result = testVoting() && result;
+            /// 11.- Voting
+            result = testVoting() && result;
 
-        /// 24.- Forum
-        result = testForum() && result;
+            /// 24.- Forum
+            result = testForum() && result;
 
-        /// 28.- Reports
-        result = testReports() && result;
+            /// 28.- Reports
+            result = testReports() && result;
 
-        /// 29 BRS - n/a
+            /// 29 BRS - n/a
+        } else {
+            printlnSummary("- Can’t proceed, since login has failed.");
+        }
 
         printSection("Overall Sanity Test: " + getResult(result));
         return result;
--- a/tools/cldr-apps/js/src/esm/cldrText.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrText.mjs
@@ -569,6 +569,8 @@ const strings = {
 
   mail_noMail: "No simulated notification emails.",
 
+
+  // NB: see DashboardWidget.vue#describe forthe notification_category replacement values
   notification_category_abstained:
     "You have abstained, or not yet voted for any value.",
   notification_category_changed:
@@ -588,6 +590,8 @@ const strings = {
   notification_category_missing:
     "Your current coverage level requires the item to be present. " +
     "(During the vetting phase, this is informational: you can’t add new values.)",
+  notification_category_new:
+    "Item did not have a winning value in ${lastVoteVersion}, and you have not voted for it.",
   notification_category_provisional:
     "There are not enough votes for this item to be approved (and used).",
   notification_category_warning:
--- a/tools/cldr-apps/js/src/esm/cldrText.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrText.mjs
@@ -569,8 +569,7 @@ const strings = {
 
   mail_noMail: "No simulated notification emails.",
 
-
-  // NB: see DashboardWidget.vue#describe forthe notification_category replacement values
+  // NB: see DashboardWidget.vue#describe for the notification_category replacement values
   notification_category_abstained:
     "You have abstained, or not yet voted for any value.",
   notification_category_changed:
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
@@ -91,6 +91,8 @@ public class CheckoutArchive {
         for (final String ver : ToolConstants.CLDR_VERSIONS) {
             if (onlyVersion != null && !onlyVersion.equals(ver)) continue;
             final Path dirName = archiveDir.resolve("cldr-" + ver);
+            // 48.1 -> release-48-1
+            // 49.0 -> release-49
             final String tag = "release-" + ver.replaceAll("\\.", "-").replaceAll("-0$", "");
             if (dirName.toFile().isDirectory()) {
                 skip++;
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
@@ -7,6 +7,7 @@ import java.nio.file.Path;
 import java.util.Set;
 import org.unicode.cldr.tool.Option.Options;
 import org.unicode.cldr.util.CLDRPaths;
+import org.unicode.cldr.util.CLDRURLS;
 import org.unicode.cldr.util.CLDRTool;
 
 @CLDRTool(
@@ -16,6 +17,7 @@ import org.unicode.cldr.util.CLDRTool;
 public class CheckoutArchive {
     enum MyOptions {
         prune("Perform a 'git prune' first"),
+        clone("Perform a clone instead of creating a worktree"),
         echo("Only show commands, don't run them. (Dry run)"),
         ;
 
@@ -47,15 +49,19 @@ public class CheckoutArchive {
         MyOptions.parse(args, true);
 
         final boolean doPrune = MyOptions.prune.option.doesOccur();
-        final String onlyVersion = null;
+        final boolean doClone = MyOptions.clone.option.doesOccur();
 
-        doCheckout(doPrune, onlyVersion);
+        doCheckout(doPrune, null, doClone);
     }
 
     /**
+     * Perform the checkout.
+     * @param doPrune perform a git worktree prune first
+     * @param onlyVersion only checkout this version. See {@link ToolConstants#formatVersion(String)} and {@link ToolConstants#haveVersion(String)}
+     * @param doClone use a git clone instead of a worktree
      * @returns number of created directories
      */
-    public static int doCheckout(final boolean doPrune, final String onlyVersion)
+    public static int doCheckout(final boolean doPrune, final String onlyVersion, final boolean doClone)
             throws IOException, InterruptedException {
 
         Path archiveDir = new File(CLDRPaths.ARCHIVE_DIRECTORY).toPath();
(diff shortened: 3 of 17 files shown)


tools/cldr-apps/js/src/esm/cldrText.mjs
```diff
@@ -569,6 +569,8 @@ const strings = {
 
   mail_noMail: "No simulated notification emails.",
 
+
+  // NB: see DashboardWidget.vue#describe forthe notification_category replacement values
```
btangmu: typo "forthe", needs space

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
```diff
@@ -68,22 +89,46 @@ public static void main(String args[]) throws IOException, InterruptedException
         }
 
         for (final String ver : ToolConstants.CLDR_VERSIONS) {
+            if (onlyVersion != null && !onlyVersion.equals(ver)) continue;
             final Path dirName = archiveDir.resolve("cldr-" + ver);
+            final String tag = "release-" + ver.replaceAll("\\.", "-").replaceAll("-0$", "");
```
btangmu, line 96: If ver = "48.1", then tag = "release-48-1"; if ver = "49.0", then tag = "release-49" -- is that right? A comment saying that here would be nice

tools/cldr-code/src/main/java/org/unicode/cldr/util/NotificationCategory.java
```diff
@@ -17,6 +17,17 @@ public enum NotificationCategory {
...
+    /**
+     * New since CLDR_LASTVOTEVERSION
+     *
+     * @see {@link VettingParameters#setFiles(CLDRLocale, Factory, Factory, Factory)}
+     * @see {@link VettingViewer#setOldVoteFactory(Factory)}
+     */
+    newSinceLastVote(
+            'N',
+            "New",
+            "Item did not have a winning value (contributed+) in the previous cycle, and you have not voted for it."),
```
btangmu: "at least contributed" might be easier for vetters to understand than "contributed+"

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -270,6 +275,14 @@ public VettingViewer(
         reasonsToPaths = Relation.of(new HashMap<>(), HashSet.class);
     }
 
+    /**
+     * Set the prior old vote factory. Needed for category {@link
```
btangmu: I'm not sure what "prior old vote factory" means -- is "prior" redundant?

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -584,6 +599,9 @@ private void updateVotedOrAbstained(String path) {
             if (userVoteStatus.userDidVote(voterId, cldrLocale, path)) {
                 VoteType voteType = userVoteStatus.getUserVoteType(voterId, cldrLocale, path);
                 voterProgress.incrementVotedPathCount(voteType);
+                if (problems.remove(NotificationCategory.newSinceLastVote)) {
+                    vc.problemCounter.decrement(NotificationCategory.newSinceLastVote);
```
btangmu, line 602: If we want to enable a user to un-hide and review New items even though they've voted on them, this might not work. A New item would need to stay New, but get hidden automatically as though the user had checked the Hide checkbox. The New count wouldn't change (unless it meant "New and not voted")...

That could be revised later.

The ticket does say "Category includes all items which are new since the last submission period at the current coverage level" and "Once a vetter votes for a New item it will be hidden from the Dashboard unless they later revert their vote by abstaining" -- it doesn't say an item stops being New when the user votes.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
```diff
@@ -613,6 +631,14 @@ private MissingStatus recordMissingChangedEtc(
             } else {
                 missingStatus = MissingStatus.PRESENT;
             }
+            if (choices.contains(NotificationCategory.newSinceLastVote)
+                            && oldVoteFile == null // completely new file
+                    || (oldVoteFile.getStringValue(path) == null
+                            || oldVoteFile.getDraftStatus(path).compareTo(DraftStatus.contributed)
+                                    < 0)) {
```
btangmu: `a && b || (c || d)` == `(a && b) || (c || d)`, right? But is the intention `a && (b || c || d)`?

I would add parentheses to make sure it matches the intention -- even if redundant, but I think not redundant in this case?

btangmu (changes requested):
I might just be confused but please check the boolean with `&&` and `||`

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: As it stands, I think the notification may be added even when `choices.contains(NotificationCategory.newSinceLastVote)` is false

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
srl295: That would be fine, it should be using the ToolConstants helper function for this calculation.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: Probably.

Note SurveyTool uses "oldVote" to refer to the previous version where there was voting.  Whereas VettingViewer uses "old" to refer to the on-disk (actually current) value.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: Yes, I guess this doesn't do the hide/unhide behavior there.

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: Actually I had another commit that fixes this that wasn't pushed - take a look

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
btangmu: We sometimes use "baseline" to mean "on-disk (actually current) value" -- that would be clearer than "old"

btangmu (dismissed):
You fixed the most important problem (the boolean condition); other changes could be in follow-up

srl295:
> You fixed the most important problem (the boolean condition); other changes could be in follow-up

Thanks, would like to get this out in front of TC and will start a followon

tools/cldr-code/src/main/java/org/unicode/cldr/util/NotificationCategory.java
srl295: Done in [9f05ce5](https://github.com/unicode-org/cldr/pull/5744/commits/9f05ce5d6694b6da6e18ea39e5bf1eec4f58728f)

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: Fixed in [9f05ce5](https://github.com/unicode-org/cldr/pull/5744/commits/9f05ce5d6694b6da6e18ea39e5bf1eec4f58728f)

tools/cldr-code/src/main/java/org/unicode/cldr/util/VettingViewer.java
srl295: Noted in description

tools/cldr-code/src/main/java/org/unicode/cldr/tool/CheckoutArchive.java
srl295: Added comment in [9f05ce5](https://github.com/unicode-org/cldr/pull/5744/commits/9f05ce5d6694b6da6e18ea39e5bf1eec4f58728f)

tools/cldr-apps/js/src/esm/cldrText.mjs
srl295: Fixed in n[9f05ce5](https://github.com/unicode-org/cldr/pull/5744/commits/9f05ce5d6694b6da6e18ea39e5bf1eec4f58728f)

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
```diff
@@ -1413,20 +1413,25 @@ boolean testSanity() {
...
         result = sanityLogin(SurveyDriverCredentials.SPECIAL_USER_TC) && result;
 
-        /// ----- OK. now start the steps
-        /// 10. CLA
-        result = testCla() && result;
+        // Exit quickly if we can't even login.
+        if (result) {
+            /// ----- OK. now start the steps
+            /// 10. CLA
+            result = testCla() && result;
```
btangmu, line 1420: `&& result` is redundant here

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
btangmu, line 1464: This is missing `&& result` so it could get true even if sanityLogin or earlier test returned false.

Is there a reason to keep testing after a failure? The logic would be simpler if each test were called like`if (result) result = testFoo();`

btangmu, line 1480: TODO should have ticket reference

btangmu (changes requested):

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
srl295: this is pre-existing, and has a separate PR, can i come back to it separately?

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
srl295: Only in this case (but trying to keep the pattern).  It means, run the subtest, and change result to false only if it fails.

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
srl295: This isn't part of the change, can we return to this separately? 

This PR only makes a smaller change so that the if exits quickly, so the docker test fails in a couple minutes instead of retrying for 20 minutes.

srl295: CLDR-16859

srl295: CLDR-16859

btangmu: I was wrong. I see below:
```
    private boolean testVotingWinningValueWithError(
            boolean result, String locale, String localeName) {
        return result;
    }
```
Still I don't understand the rationale for chaining the tests this way; and there should be a TODO comment for the work in progress

btangmu (approved):

tools/cldr-apps-webdriver/src/test/java/org/unicode/cldr/surveydriver/SurveyDriver.java
srl295: There is a TODO, it's two lines prior.

srl295: The other PR was just a branch, now a PR https://github.com/unicode-org/cldr/pull/5752 

I had stubbed out some of the testing (hence the if false and TODO) so that this test could be useful before start of shakedown, so that the rest of the PR work could continue at a later time.

macchiati:
> Note, does NOT support unhiding the 'New' category. Once you have voted for it, it's no longer New.

I think it is right to leave that in place; that is, I _don't_ think we want unhide to show New's that the user has voted on.

AEApple:
> > Note, does NOT support unhiding the 'New' category. Once you have voted for it, it's no longer New.
> 
> I think it is right to leave that in place; that is, I _don't_ think we want unhide to show New's that the user has voted on.

I think there will be some vetters who may want to revisit their votes on the new items, but I'm okay waiting for them to ask for that feature if it would take significant time to add.

macchiati:
I think having that under the Hide button will be confusing; if I unhide, that will change the number of New items I see, but not change the number in the label. Will I be able to hide items I haven't voted on? etc.

AEApple:
> I think having that under the Hide button will be confusing; if I unhide, that will change the number of New items I see, but not change the number in the label. Will I be able to hide items I haven't voted on? etc.

Agree that we'll have to be careful with the UX experience when a user wants to unhide new since they still need to be able to hide warnings separately.

PR #5745, opened 2026-05-20
CLDR-19499 Align Ethiopic eras, remove `ethiopic-amete-alem` data

robertbastian:
CLDR-19499

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
+++ b/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
@@ -27,6 +27,8 @@
 %O = (chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|persian)
 %P = (future|past)
 %R = (gregorian|buddhist|coptic|ethiopic|hebrew|indian|islamic|japanese|persian|roc)
+# Calendars in the data, but not in SurveyTool
+%hiddenCalendars = (ethiopic-amete-alem|islamic-civil|islamic-tbla|islamic-umalqura|islamic-rgsa)
 %S = ([^/]*+)
 %V = (speed|acceleration)
 %W = (temperature|pressure)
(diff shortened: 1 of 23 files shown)


srl295 (dismissed):
Unless it's added to HIDE it's possible these paths may still be in comprehensive (and could come back).. approving for now

srl295:
Try editing PathHeader.txt

I'll check soon

srl295:
PTAL.  Also @macchiati i filed https://unicode-org.atlassian.net/browse/CLDR-19505 based on pain points here.

tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/PathHeader.txt
```diff
@@ -26,7 +26,9 @@
 # all others use M## form months.
 %O = (chinese|coptic|dangi|ethiopic|hebrew|indian|islamic|persian)
 %P = (future|past)
-%R = (gregorian|buddhist|coptic|ethiopic|ethiopic-amete-alem|hebrew|indian|islamic|japanese|persian|roc)
+%R = (gregorian|buddhist|coptic|ethiopic|hebrew|indian|islamic|japanese|persian|roc)
+# Calendars in the data, but not in SurveyTool
+%hiddenCalendars = (ethiopic-amete-alem|islamic-civil|islamic-tbla|islamic-umalqura|islamic-rgsa)
```
macchiati, line 31: We should use this technique more often, especially when only a few items need to be excluded.

We can then replace the %R by %A : that is, let everything else go through.

macchiati (approved):
I assume the islamic variants are in a separate PR, eh?

srl295:
> I assume the islamic variants are in a separate PR, eh?

did not confirm.

robertbastian:
The islamic variants are in this PR

macchiati (commented):
Ah, I was looking for changes in in some of the data files but it looks like you didn't need any any except 4 root.

PR #5748, opened 2026-05-20
CLDR-18314 Add language names for apw, mid, nnp, rej, sjd, sje, sju 

DraganBesevic:
CLDR-18314

Adding 7 language names to en.xml that are needed for keyboard name localization. These languages have keyboards in Apple platforms but their names were missing from CLDR, causing them to appear unlocalized.

Languages added: Western Apache (apw), Mandaic (mid), Wancho (nnp), Rejang (rej), Kildin Sami (sjd), Pite Sami (sje), Ume Sami (sju).

Also added these codes to attributeValueValidity.xml so they pass CLDR validation.

Chickasaw (cic) and Osage (osa) were already present. Lushootseed (lut) is deferred pending TC discussion since its locale doesn't exist in CLDR yet.

Per TC decision 2025-05-06: approved at comprehensive coverage level.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/en.xml
+++ b/common/main/en.xml
@@ -45,6 +45,7 @@ annotations.
 			<language type="ang">Old English</language>
 			<language type="ann">Obolo</language>
 			<language type="anp">Angika</language>
+			<language type="apw">Western Apache</language>
 			<language type="ar">Arabic</language>
 			<language type="ar_001">Modern Standard Arabic</language>
 			<language type="arc">Aramaic</language>
@@ -415,6 +416,7 @@ annotations.
 			<language type="mh">Marshallese</language>
 			<language type="mi">Māori</language>
 			<language type="mic">Mi'kmaw</language>
+			<language type="mid">Mandaic</language>
 			<language type="min">Minangkabau</language>
 			<language type="mk">Macedonian</language>
 			<language type="ml">Malayalam</language>
@@ -460,6 +462,7 @@ annotations.
 			<language type="nmg">Kwasio</language>
 			<language type="nn">Norwegian Nynorsk</language>
 			<language type="nnh">Ngiemboon</language>
+			<language type="nnp">Wancho</language>
 			<language type="no">Norwegian</language>
 			<language type="nog">Nogai</language>
 			<language type="non">Old Norse</language>
@@ -522,6 +525,7 @@ annotations.
 			<language type="raj">Rajasthani</language>
 			<language type="rap">Rapanui</language>
 			<language type="rar">Rarotongan</language>
+			<language type="rej">Rejang</language>
 			<language type="rgn">Romagnol</language>
 			<language type="rhg">Rohingya</language>
 			<language type="rif">Riffian</language>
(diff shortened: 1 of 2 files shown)


common/main/en.xml
```diff
@@ -573,6 +577,9 @@ annotations.
 			<language type="shu">Chadian Arabic</language>
 			<language type="si">Sinhala</language>
 			<language type="sid">Sidamo</language>
+			<language type="sjd">Kildin Sami</language>
```
AEApple, line 580: Shouldn't the Sami languages have the `á` instead of an `a` without the diacritic?

https://en.wikipedia.org/wiki/S%C3%A1mi_languages

AEApple (changes requested):

common/main/en.xml
DraganBesevic: I followed the existing convention for current Sami entries. All of Northern Sami, Southern Sami, Lule Sami, Inari Sami, Skolt Sami use "Sami" without the diacritic. If we want to change these to "Sámi" I'll be glad too, but it will look inconsistent.

common/supplemental/attributeValueValidity.xml
srl295, line 1: Yeah, I thought we got rid of this.

common/main/en.xml
srl295: I think we've used Sami (Which is an English variant given in that article).

srl295 (approved):

common/main/en.xml
srl295: It could be an alt variant in English if we want?

It also probably should have core/extension values but that could be a follow on (especially if these are in comprehensive)

AEApple: SG! Makes sense to be consistent with what's already in English. It just differed from what was in the ticket.

AEApple (approved):

stenshamn:
"Sami" seems to be generally accepted, but the Samis themselves seem to be consistently using Sámi or Saami. Minor pet peeve of mine, but I'm kind of in the camp of using the diacritic. My 2 cents...

srl295:
> "Sami" seems to be generally accepted, but the Samis themselves seem to be consistently using Sámi or Saami. Minor pet peeve of mine, but I'm kind of in the camp of using the diacritic. My 2 cents... 

Maybe we should put it in as is and revisit during this cycle?

AEApple:
> > "Sami" seems to be generally accepted, but the Samis themselves seem to be consistently using Sámi or Saami. Minor pet peeve of mine, but I'm kind of in the camp of using the diacritic. My 2 cents...
> 
> Maybe we should put it in as is and revisit during this cycle?

Yes, I was suggesting that merge this, and assign the ticket for review by Riitta with the ask to research how the Sami community would write their language name when writing it in English.

common/supplemental/attributeValueValidity.xml
stenshamn: Should we? I'm curious what the purpose is as well.

stenshamn (approved):

macchiati:
> On common/supplemental/attributeValueValidity.xml
<https://github.com/unicode-org/cldr/pull/5748#discussion_r3282302982>:
> Should we? I'm curious what the purpose is as well.

We need to document this better; I added a draft explanation of this in
https://unicode-org.atlassian.net/browse/CLDR-19129?focusedCommentId=210559

common/supplemental/attributeValueValidity.xml
macchiati: > On [common/supplemental/attributeValueValidity.xml](https://github.com/unicode-org/cldr/pull/5748#discussion_r3282302982):
> Should we? I'm curious what the purpose is as well.

We need to document this better; I added a draft explanation of this in https://unicode-org.atlassian.net/browse/CLDR-19129?focusedCommentId=210559

PR #5755, opened 2026-05-22
CLDR-19489 Add information about new Dashboard category

AEApple:
CLDR-19489

ALLOW_MANY_COMMITS=true


stenshamn:
Why is the diff view the whole document on this one?

AEApple:
> Why is the diff view the whole document on this one?

No idea, I think it has something to do with when we modify the placement of a table. Actually the rich diff displays it a bit more clearly I think.

srl295 (approved):

PR #5757, opened 2026-05-24
CLDR-18980 Constructed interval formats

macchiati:
CLDR-18980

The end goal is to have better consistency between the available formats and the interval formats (and within the interval formats). Note: for 49, I think it best to just focus on gregorian; that's probably the high-runner case for datetime intervals.

This PR adds a Check for for conflicts between the interval separators and their base formats. It also improves the tooling and utilities, including code for the automatic construction of interval formats, and a test that they work. The test has a flag that allows output of a comparison of results, to help fine-tune the algorithm. 

### Next steps

* Add links to the warning message for conflicts.
* Add a Check of the interval formats that will issue a warning when the value conflicts with the constructed value. 

### Later

* If feasible, add the constructed interval separator to the Others column, to make it easy for vetters to pick.
* Maybe, rework the charts so people can more easily see the available formats next to their corresponding interval formats.

## For reviewers

* common/main adds provisional values that don't conflict with the the base patterns. That will give the vetters a leg up on supplying items, and reduce warning clutter from the new CheckDates.java checkIntervalSeparators.
* CheckDates.java adds checkIntervalSeparators. This is the main feature.
* CldrIntervalFormat.java lots of cleanup, plus IntervalDiff
* CldrPathUtilities.java adds more convenience methods for XPaths
* DateTimeCanonicalizer.java adds INTERVAL_SEPARATOR (and changes SEPARATOR to NUMERIC_SEPARATOR.
* DatetimeUtilities.java Many additions/refactorings of utilities
* NestedMap.java Made Iterable for convenience
* PathHeader.txt Changed case of Fallback for consistency
* TestCheckCLDR.java minor
* TestCoverageLevel.java added test for interval separators
* TestDateOrder.java added testIntervalConstructor, testDatetimeUtilities, testAddSeparators, testNormalized
    * testNormalized is for later, to fix denormalized skeletons. We can't change until after XML

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
```diff
@@ -635,6 +641,43 @@ && isTooMuchWiderThan(value, abbrValue)
...
+                            new CheckStatus()
+                                    .setCause(this)
+                                    .setMainType(CheckStatus.warningType)
+                                    .setSubtype(Subtype.patternDatetimeMismatchWithSeparator)
+                                    .setMessage(
+                                            "Conflict between separator «{0}» and interval patterns like «{1}» (Code {2}/{3}). Fix one or the other.",
+                                            value,
+                                            intervalPattern,
+                                            separatorType.id,
+                                            separatorType.subId));
```
srl295, line 668: do we want this to be a link?

common/main/am.xml
```diff
@@ -2495,6 +2495,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 						<appendItem request="Timezone">↑↑↑</appendItem>
 					</appendItems>
 					<intervalFormats>
+						<intervalFormatRanges>
+							<intervalFormatRange type="mixed" draft="provisional">{0} – {1}</intervalFormatRange>
+
```
btangmu, line 2500: Odd to see a blank line here

srl295 (approved):
looks ok. not sure why the same value is in each locale rather than in root?

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
macchiati: Re: in root
It is easier just to generate all the values, and let the survey tool convert them into ^^^ where reasonable.

Yes, we do; I'm working on a followup PR

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
```diff
@@ -536,4 +565,315 @@ public void testSkeletons() {
...
+
+        Set<PatternElement> pe1 = DatetimeUtilities.getSkeletonElements("Gy");
+        Set<PatternElement> pe2 = DatetimeUtilities.getSkeletonElements("yG");
+        Set<PatternElement> pe3 = DatetimeUtilities.getSkeletonElements("y");
+        Set<PatternElement> pe4 = DatetimeUtilities.getSkeletonElements("G");
+        assertEquals(
+                "PATTERN_COMPARATOR", 0, DatetimeUtilities.PATTERN_COMPARATOR.compare("Gy", "y"));
+        String actual = Joiners.N.join(SkeletonField.getDatecombos());
+        assertEquals("getDatecombos", "", actual);
+    }
```
srl295, line 726: @macchiati 
I've been scratching my head trying to figure out what's going on in #5769 

You added this test above, but it's in a `@Disabled` class and fails.

common/main/cy.xml
```diff
@@ -8414,12 +8420,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
...
-			<pluralMinimalPairs count="two">{0} gi, {0} gath</pluralMinimalPairs>
-			<pluralMinimalPairs count="few">{0} chi, {0} cath</pluralMinimalPairs>
-			<pluralMinimalPairs count="many">{0} chi, {0} chath</pluralMinimalPairs>
-			<pluralMinimalPairs count="other">{0} ci, {0} cath</pluralMinimalPairs>
+
+
+
+
+
+
```
srl295, line 8428: Missed this before…

PR #5761, opened 2026-05-25
CLDR-19035 Fix generation of example dependencies

btangmu:
-Ensure that the ExampleGenerator used by GenerateExampleDependencies creates its ICUServiceBuilder instance using the provided RecordingCLDRFile, not an ordinary CLDRFile

-Add a test TestCLDRPaths.testGenerateExampleDependencies that would fail without the fix in ExampleGenerator

-Rename locIsFake to locIsExceptional, to avoid implying that locales like mul are fake

-Comments

CLDR-19035

ALLOW_MANY_COMMITS=true


btangmu:
This should fix GenerateExampleDependencies. However, that takes a few hours to run and produce ExampleDependencies.java, so I'm making this a draft PR for now. If that goes well as expected, then this PR can be made ready for review, and the revised ExampleDependencies.java can either be included in another commit to this PR, or in another PR.

macchiati (approved):

btangmu:
The generated file looks good; I'll add that in a separate PR shortly.

PR #5766, opened 2026-05-26
CLDR-19035 Update ExampleDependencies.java by running GenerateExample…

btangmu:
…Dependencies

-Generation locally took 161 minutes (2 hours, 41 minutes)

CLDR-19035

ALLOW_MANY_COMMITS=true


macchiati:
Does this fix the problem you found earlier (that you noticed based on size)?

btangmu:
> Does this fix the problem you found earlier (that you noticed based on size)?

Yes; actually the just-merged PR fixed that. Clearly this is what made the file too small: the ICUServiceBuilder wasn't using a RecordingCLDRFile.

srl295 (approved):

PR #5529, opened 2026-03-31
CLDR-19359 Chart-generation-related changes for 48.2

btangmu:
CLDR-19359

- [ ] This PR completes the ticket.

ALLOW_MANY_COMMITS=true

PR #5767, opened 2026-05-26
CLDR-19430 GenerateBirth

btangmu:
-Ran as follows: java -jar tools/cldr-code/target/cldr-code.jar GenerateBirth -o46

CLDR-19430

ALLOW_MANY_COMMITS=true


btangmu:
The two binary data files are opaque (to me at least), so I guess we just trust they're OK if tests pass?

btangmu:
Two test failures:

build: tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java#L698
(TestCheckCLDR.java:698) Error: : CheckNew el, //ldml/units/unitLength[@type="long"]/unit[@type="mass-grain"]/displayName: expected "In CLDR 40.0 the English value for this field changed from “grain” to “grains”, but the corresponding value for your locale didn't change.", got ""

build: tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java#L698
(TestCheckCLDR.java:698) Error: : CheckNew de, //ldml/localeDisplayNames/territories/territory[@type="001"]: expected "In CLDR 39.0 the English value for this field changed from “World” to “world”, but the corresponding value for your locale didn't change.", got ""

btangmu:
I imagine this test (TestCheckCLDR.TestCheckNew) needs updating:

```
/**
 * Check that at least one path in a locale is outdated and one path is not. That may change
 * each time. This needs to be a <locale,path> that is currently outdated (birth older than
 * English's) if the test fails with "no failure message" run GenerateBirths (if you haven't
 * done so) look at readable results in the log file in
 * https://github.com/unicode-org/cldr-staging/blob/main/births/41.0/fr.txt (for the current
 * version, not nec. 41.0) for a reasonable locale ( may change locale to something other than
 * fr) find a path that is outdated. To work on both limited and full submissions, choose one
 * with English = trunk Sometimes the English change is suppressed in a limited release if the
 * change is small. Pick another in that case. check the data files to ensure that it is in fact
 * outdated. change the path to that value the 3rd parameter is the message displayed to the
 * user, or "" if not 'English Changed' So the first group of tests are for items that should
 * not be outdated And the second group is ones that should be outdated.
 */
public void TestCheckNew() {
    // Not outdated
    checkCheckNew("de", "//ldml/localeDisplayNames/languages/language[@type=\"en\"]", "");

    // Outdated
    checkCheckNew(
            "de",
            "//ldml/localeDisplayNames/territories/territory[@type=\"001\"]",
            "In CLDR 39.0 the English value for this field changed from “World” to “world”, but the corresponding value for your locale didn't change.");
    checkCheckNew(
            "el",
            "//ldml/units/unitLength[@type=\"long\"]/unit[@type=\"mass-grain\"]/displayName",
            "In CLDR 40.0 the English value for this field changed from “grain” to “grains”, but the corresponding value for your locale didn't change.");
}
```

Locally I find (only) two files matching files "el.txt" and "de.txt" (except for files in directories named personNameTest which I assume are unrelated). Both are in the directory ../cldr-staging/births/49.0. There are no other directories inside ../cldr-staging/births.

The comment references a URL. Browsing to https://github.com/unicode-org/cldr-staging/tree/main/births I see directories up to 44.0, dated 3 years ago. Presumably there's an overdue task that would populate newer directories.

srl295:
> I imagine this test (TestCheckCLDR.TestCheckNew) needs updating:

yeah - you could find another value, or maybe just make it a logknown issue for now?

> 
> Locally I find (only) two files matching files "el.txt" and "de.txt" (except for files in directories named personNameTest which I assume are unrelated). Both are in the directory ../cldr-staging/births/49.0. There are no other directories inside ../cldr-staging/births.
> 
> The comment references a URL. Browsing to https://github.com/unicode-org/cldr-staging/tree/main/births I see directories up to 44.0, dated 3 years ago. Presumably there's an overdue task that would populate newer directories.

they are the output logs from the generator. i don't know if there's a need to check them in after generation.

srl295 (dismissed):
RSLGTM

btangmu:
> they are the output logs from the generator. i don't know if there's a need to check them in after generation.

The comment on TestCheckCLDR.TestCheckNew seems to imply those logs would provide clues about paths/locales to use for updating the test. Otherwise, I don't see how to proceed, other than disabling the test...

macchiati:
There are output text files that match the binaries, and those can be used. Otherwise, one could use the delta charts (https://www.unicode.org/cldr/charts/48/delta/index.html). But you'd have to find an item that changed in English but not in some language.

I'll look more tomorrow.

Also, I noticed that we are not generating the charts correctly. v48 has CLDR v48.2 – 48.1. It should have CLDR v48.2 – 47. That is, it should be the latest dot version of each release. https://unicode.org/cldr/charts/47/delta/index.html has the correct versions.

btangmu:
> output text files

Where might I find those text files?

srl295:
> Also, I noticed that we are not generating the charts correctly. v48 has CLDR v48.2 – 48.1. It should have CLDR v48.2 – 47. That is, it should be the latest dot version of each release. https://unicode.org/cldr/charts/47/delta/index.html has the correct versions.

Is this part of the dot-release-charts discussion we haven't had yet? https://github.com/unicode-org/cldr/pull/5529

btangmu:
I'm in the middle of some detective work. It looks like the secret is to run TestOutdatedPaths with -v...

btangmu:
Here's the formula:
```
mvn test --file tools/pom.xml -pl cldr-code -Dtest=org.unicode.cldr.unittest.TestShim -Dorg.unicode.cldr.unittest.testArgs='-e10 -v -n -filter:TestOutdatedPaths'
```
Notice the `-v` (verbose), which is essential. The output will include something like this:
```
    TestShow {
      (TestOutdatedPaths.java:69) de total outdated:	266
      (TestOutdatedPaths.java:78) English (48.0):\t«Saurashtra»	⇒\t«Sourashtra»	Native: «Saurashtra»	Path: Locale Display Names	Languages (O-S)	S	Sourashtra ► saz	XML-Path: //ldml/localeDisplayNames/languages/language[@type="saz"]
...
      (TestOutdatedPaths.java:78) English (47.0):\t«↑↑↑»	⇒\t« »	Native: « »	Path: Miscellaneous	Person Name Formats	AuxiliaryItems	foreignSpaceReplacement	XML-Path: //ldml/personNames/foreignSpaceReplacement
```

Copy those paths from XML-Path into `TestCheckCLDR.TestCheckNew`, replacing the "Outdated" paths for which it failed. Run:
```
mvn test --file tools/pom.xml -pl cldr-code -Dtest=org.unicode.cldr.unittest.TestShim -Dorg.unicode.cldr.unittest.testArgs='-e10 -n -filter:TestCheckNew'
```
It will fail, since the messages need updating. Copy the messages it produced (like "In CLDR 48.0 the English value for this field changed from “Saurashtra” to “Sourashtra”, but the corresponding value for your locale didn't change.") into `TestCheckCLDR.TestCheckNew`. Tests should now pass.

srl295:
> Here's the formula:
> 
> ```
> mvn test --file tools/pom.xml -pl cldr-code -Dtest=org.unicode.cldr.unittest.TestShim -Dorg.unicode.cldr.unittest.testArgs='-e10 -v -n -filter:TestOutdatedPaths'
> ```
> 
> Notice the `-v` (verbose), which is essential. The output will include something like this:
> 
> ```
>     TestShow {
>       (TestOutdatedPaths.java:69) de total outdated:	266
>       (TestOutdatedPaths.java:78) English (48.0):\t«Saurashtra»	⇒\t«Sourashtra»	Native: «Saurashtra»	Path: Locale Display Names	Languages (O-S)	S	Sourashtra ► saz	XML-Path: //ldml/localeDisplayNames/languages/language[@type="saz"]
> ...
>       (TestOutdatedPaths.java:78) English (47.0):\t«↑↑↑»	⇒\t« »	Native: « »	Path: Miscellaneous	Person Name Formats	AuxiliaryItems	foreignSpaceReplacement	XML-Path: //ldml/personNames/foreignSpaceReplacement
> ```
> 
> Copy those paths from XML-Path into `TestCheckCLDR.TestCheckNew`, replacing the "Outdated" paths for which it failed. Run:
> 
> ```
> mvn test --file tools/pom.xml -pl cldr-code -Dtest=org.unicode.cldr.unittest.TestShim -Dorg.unicode.cldr.unittest.testArgs='-e10 -n -filter:TestCheckNew'
> ```
> 
> It will fail, since the messages need updating. Copy the messages it produced (like "In CLDR 48.0 the English value for this field changed from “Saurashtra” to “Sourashtra”, but the corresponding value for your locale didn't change.") into `TestCheckCLDR.TestCheckNew`. Tests should now pass.

and @macchiati 

would it make more sense to just keep one additional static version of the outdated paths, and have the unit tests run against that? otherwise this seems like a lot of busywork each release with not much value. 

That or, cause TestOutDatedPaths to automatically recommend one changed path and write that to a (human readable and reviewable!) data file, to be tested against.   If there were really no changes (hey it could happen) that gets written to the data file and the test is skipped.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
```diff
@@ -654,6 +654,34 @@ public void TestCheckNames() {
      * outdated. change the path to that value the 3rd parameter is the message displayed to the
      * user, or "" if not 'English Changed' So the first group of tests are for items that should
      * not be outdated And the second group is ones that should be outdated.
+     *
+     * <p>Alternative instructions: the above may not work if
```
srl295, line 658: these should go in the site under the BRS pages, maybe linked from here

srl295 (approved):

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestCheckCLDR.java
btangmu: Could we make that a follow-up PR?

btangmu: I've un-checked "This PR completes the ticket"

srl295: 100% follow up

btangmu: @srl295 by "BRS pages", do you mean docs/site/development/cldr-big-red-switch.md or a new file in docs/site/development/cldr-big-red-switch or ...?

btangmu: Or https://cldr.unicode.org/development/cldr-development-site/updating-englishroot ?? That's referenced in the BRS task "GenerateBirths"

AEApple: Could we have a SBRS page under https://github.com/unicode-org/cldr/tree/main/docs/site/development/cldr-big-red-switch ?

I would like a few items included there that are recurring tasks including the build example dependencies, generate births (happy to check the BRS and list out the tasks if you'd like?) Then we could just link from the BRS sheet to the header for each recurring task and more easily be able to look up the history of why tasks have changed. WDYT?

macchiati:
Definitely!

PR #5769, opened 2026-05-27
CLDR-13465 Re-enabled some Disabled tests, add function-level Disabled

srl295:
CLDR-13465

- re-enabled some tests by using `@Disabled` selectively on functions that still failed
- we now support `@Disabled` (yes, from JUnit) on TestFmwk classes and functions

ALLOW_MANY_COMMITS=true


tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
```diff
@@ -147,6 +117,7 @@ public void TestDateImportance() {
     static final String intervalFormatPathPrefix =
             "//ldml/dates/calendars/calendar[@type=\"gregorian\"]/dateTimeFormats/intervalFormats/";
 
+    @Disabled
     public void TestIso8601() {
```
srl295, line 150: This makes it really quick to keep a test compiling, but skip it.

AEApple:
How do we want to track resolving the individual tests that had to be disabled longer term?

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestDateOrder.java
macchiati: Nice 👍

macchiati (dismissed):

srl295:
> How do we want to track resolving the individual tests that had to be disabled longer term?

they are marked as such when the tests run - can search for `@Disabled` 

Actually we could perhaps do

```java
@Disabled("CLDR-00000 skipped because")
TestSomething() {
…
}
```

.. and pick those up as logknownissues

srl295:
For now i need a rubber stamp to approve because a failing disabled test was merged in the meantime

btangmu (approved):
RSLGTM

PR #5770, opened 2026-05-27
CLDR-19076 For redacting db on cldr-smoke, use /mnt

btangmu:
-Revise ansible: fetch_db_sh.j2, redact_db_sh.j2; define cldr_tmp_drive in main.yml

CLDR-19076

ALLOW_MANY_COMMITS=true


srl295 (approved):
good. i should make sure /mnt is writable but i can do that separately

PR #5774, opened 2026-05-27
CLDR-19511 add New to the VettingParticipation table

srl295:
CLDR-19511

- [X]  This PR completes the ticket.

Also improved some of the messaging

loading

image

html preview
image

spreadsheet
image

Also, only TC Orgs can do Vetting Participation!

Others get this due to CLDR-18868
> Please log in as a user with sufficient permissions. (Your organization is not enabled for Vetting Participation.)
 

ALLOW_MANY_COMMITS=true


AEApple:
I still think that missing and provisional should also count as 'new' since they usually won't have winning values in the prior release, but I'm okay just using a formula. I'm just a little concerned that vetters may be confused if their number of 'new' items grow when a vetter from another organization votes for the items before they do.

macchiati:
I thought the formula was: there was no draft=contributed+ & coverage=basic..org-coverage value in last release, and the vetter hasn't yet voted on it.

macchiati:
> I still think that missing and provisional should also count as 'new' since they usually won't have winning values in the prior release, but I'm okay just using a formula. I'm just a little concerned that vetters may be confused if their number of 'new' items grow when a vetter from another organization votes for the items before they do.

I think the disconnect could be the following.

* I vote on X, but don't have enough votes to get it approved. New goes down by 1, but missing & provisional stay the same.
* I don't vote on X, but there are enough votes to convert from Missing to Provisional, or Missing or Provisional to approved.

This is the way we want it to work, since New focuses on me, while Missing focuses on "enough vetters". But means that Missing could be 0 and New be 100, or vice versa.

macchiati (dismissed):
I'm approving @srl295, but wait until @AEApple has a chance to look at the comments.

srl295:
> I still think that missing and provisional should also count as 'new' 


That's a separate concern from this PR which is just the export to VP

macchiati:
Right, that's not a blocker for this. So let's merge (once the build failure is fixed)

srl295:
Failure fixed

macchiati (approved):
Y

PR #5780, opened 2026-05-30
CLDR-19409 fix ownership and lifecycle of ICUServiceBuilder

srl295:
CLDR-19409

Completely get rid of `ICUServiceBuilder.forLocale()`

Instead, `ICUServiceFactory` contains the cache, and, every `Factory` ownes one. 

Perf should be the same or better.

Also fixes CLDR-19518  and CLDR-19166

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
@@ -497,7 +497,7 @@ public class ConsoleCheckCLDR {
         english = backCldrFactory.make("en", true);
 
         CheckCLDR.setDisplayInformation(english);
-        setExampleGenerator(new ExampleGenerator(english, english));
+        setExampleGenerator(new ExampleGenerator(english, backCldrFactory));
         PathShower pathShower = new PathShower();
 
         // call on the files
@@ -681,12 +681,12 @@ public class ConsoleCheckCLDR {
                     UnicodeSet missingExemplars = new UnicodeSet();
                     UnicodeSet missingCurrencyExemplars = new UnicodeSet();
                     FlexibleDateFromCLDR fset =
-                            checkFlexibleDates ? new FlexibleDateFromCLDR(file) : null;
+                            checkFlexibleDates ? new FlexibleDateFromCLDR(cldrFactory, file) : null;
                     pathShower.set(localeID);
 
                     // only create if we are going to use it
                     final ExampleGenerator exampleGenerator =
-                            SHOW_EXAMPLES ? new ExampleGenerator(file, englishFile) : null;
+                            SHOW_EXAMPLES ? new ExampleGenerator(file, cldrFactory) : null;
 
                     int pathCount = 0;
                     Status otherPath = new Status();
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -68,6 +68,7 @@ import org.unicode.cldr.util.DayPeriodInfo;
 import org.unicode.cldr.util.DayPeriodInfo.DayPeriod;
 import org.unicode.cldr.util.EmojiConstants;
 import org.unicode.cldr.util.ExemplarSets.ExemplarType;
+import org.unicode.cldr.util.Factory;
 import org.unicode.cldr.util.GrammarInfo;
 import org.unicode.cldr.util.GrammarInfo.GrammaticalFeature;
 import org.unicode.cldr.util.GrammarInfo.GrammaticalScope;
@@ -419,21 +420,13 @@ public class ExampleGenerator {
      * Create an Example Generator. If this is shared across threads, it must be synchronized.
      *
      * @param resolvedCldrFile
-     */
-    public ExampleGenerator(CLDRFile resolvedCldrFile) {
-        this(resolvedCldrFile, CLDRConfig.getInstance().getEnglish());
-    }
-
-    /**
-     * Create an Example Generator. If this is shared across threads, it must be synchronized.
-     *
-     * @param resolvedCldrFile
      * @param englishFile
      */
-    public ExampleGenerator(CLDRFile resolvedCldrFile, CLDRFile englishFile) {
+    public ExampleGenerator(CLDRFile resolvedCldrFile, Factory cldrFactory) {
         if (!resolvedCldrFile.isResolved()) {
             throw new IllegalArgumentException("CLDRFile must be resolved");
         }
+        final CLDRFile englishFile = cldrFactory.make("en", true);
         if (!englishFile.isResolved()) {
             throw new IllegalArgumentException("English CLDRFile must be resolved");
         }
@@ -450,10 +443,7 @@ public class ExampleGenerator {
         // GenerateExampleDependencies needs the ICUServiceBuilder instance to use the
         // RecordingCLDRFile if one is provided, rather than a pre-existing ordinary CLDRFile.
         boolean fileIsRecording = cldrFile.getClass() == RecordingCLDRFile.class;
-        this.icuServiceBuilder =
-                (locIsExceptional || fileIsRecording)
-                        ? new ICUServiceBuilder(resolvedCldrFile, locIsExceptional)
-                        : ICUServiceBuilder.forLocale(CLDRLocale.getInstance(localeId));
+        this.icuServiceBuilder = cldrFactory.getICUServiceBuilder(CLDRLocale.getInstance(localeId));
 
         bestMinimalPairSamples = new BestMinimalPairSamples(cldrFile, icuServiceBuilder, false);
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
@@ -75,7 +75,6 @@ import org.unicode.cldr.util.GrammarInfo.GrammaticalScope;
 import org.unicode.cldr.util.GrammarInfo.GrammaticalTarget;
 import org.unicode.cldr.util.ICUServiceBuilder;
 import org.unicode.cldr.util.LanguageTagParser;
-import org.unicode.cldr.util.LocaleNames;
 import org.unicode.cldr.util.NameGetter;
 import org.unicode.cldr.util.NameType;
 import org.unicode.cldr.util.Pair;
@@ -438,12 +437,13 @@ public class ExampleGenerator {
                 supplementalDataInfo.getGrammarInfo(localeId); // getGrammarInfo can return null
         this.englishFile = englishFile;
         this.typeIsEnglish = (resolvedCldrFile == englishFile);
-        boolean locIsExceptional =
-                LocaleNames.XX_TEST.equals(localeId) || LocaleNames.MUL.equals(localeId);
         // GenerateExampleDependencies needs the ICUServiceBuilder instance to use the
         // RecordingCLDRFile if one is provided, rather than a pre-existing ordinary CLDRFile.
         boolean fileIsRecording = cldrFile.getClass() == RecordingCLDRFile.class;
-        this.icuServiceBuilder = cldrFactory.getICUServiceBuilder(CLDRLocale.getInstance(localeId));
+        this.icuServiceBuilder =
+                fileIsRecording
+                        ? ICUServiceBuilder.inefficientSingletonServiceBuilder(cldrFile)
+                        : cldrFactory.getICUServiceBuilder(CLDRLocale.getInstance(localeId));
 
         bestMinimalPairSamples = new BestMinimalPairSamples(cldrFile, icuServiceBuilder, false);
 
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/Factory.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/Factory.java
@@ -12,6 +12,7 @@ import java.util.function.Supplier;
 import org.unicode.cldr.test.TestCache;
 import org.unicode.cldr.util.CLDRFile.DraftStatus;
 import org.unicode.cldr.util.CLDRLocale.SublocaleProvider;
+import org.unicode.cldr.util.ICUServiceBuilder.ICUServiceFactory;
 import org.unicode.cldr.util.SupplementalDataInfo.ParentLocaleComponent;
 import org.unicode.cldr.util.XMLSource.ResolvingSource;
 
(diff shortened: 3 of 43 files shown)


srl295:
and it works 
image


and reports update **instantly**

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
```diff
@@ -512,7 +512,7 @@ private String getTZName(String tzcode) {
         if (tzcode.length() == 4 && !tzcode.equals("gaza")) {
             return longid;
         }
-        TimezoneFormatter tzf = new TimezoneFormatter(cldrFile);
+        TimezoneFormatter tzf = new TimezoneFormatter(null, cldrFile);
```
srl295, line 515: CLDRFile constructor initializes a NameGetter which pulls in TimezoneFormatter which pulls in ICUServiceBuilder…… !!!! 

I kind of think we should move NameGetter into ICUServiceBuilder. Then we could get rid of this.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -23,16 +23,46 @@
...
+        public ICUServiceFactory(Factory f) {
+            cldrFactory = f;
+        }
+    }
+
+    @Deprecated
+    public static final ICUServiceBuilder inefficientSingletonServiceBuilder(
+            final CLDRFile resolvedFile) {
+        return new ICUServiceBuilder(resolvedFile, defaultCollationFactory);
+    }
```
srl295, line 61: see https://github.com/unicode-org/cldr/pull/5780/changes#r3329379795

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -123,26 +140,6 @@ public CLDRFile getCldrFile() {
         return cldrFile;
     }
 
-    public static ICUServiceBuilder forLocale(CLDRLocale locale) {
```
srl295, line 126: note, this is a really old function.

I suspect even outside of the current issues, there may be latent bugs this was hiding.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
macchiati: I think that is reasonable. We might think about whether or not it makes sense to incrementally load ICUBuilder. For example, if we are building a number formatter, maybe we just build that. If the ICU builder gets flushed in the meantime, we haven't spent time building something we don't need.

If we are building a date formatter, then that would need a number formatter. It would call a getter, which would cause the number formatter data to be loaded.

I don't know whether this would be productive or not; it would depend on how much time it takes to build an ICUServiceBuilder.

srl295: I thought about that, splitting up functionality. 

This PR makes the ownership clear. So the builder could simply initialize its own functionality only as needed.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
srl295, line 119: Test data looks suspicious here.

Now fails with:

```
    testCurrency {
      Error: (TestExampleGenerator.java:134) : en-one-BMD: expected "〖❬1 ❭Bermudan dollar〗", got "〖❬1 ❭value-one〗"
      Error: (TestExampleGenerator.java:139) : en-one-pat: expected "〖❬1❭ ❬US dollar❭〗〖❬1❭ ❬euro❭〗", got "〖❬1❭_❬US dollar❭〗〖❬1❭_❬euro❭〗"
      Error: (TestExampleGenerator.java:134) : en-other-BMD: expected "〖❬1.23 ❭Bermudan dollars〗〖❬0.00 ❭Bermudan dollars〗", got "〖❬1.23 ❭value-other〗〖❬0.00 ❭value-other〗"
      Error: (TestExampleGenerator.java:139) : en-other-pat: expected "〖❬1.23❭ ❬US dollars❭〗〖❬1.23❭ ❬euros❭〗〖❬0.00❭ ❬US dollars❭〗〖❬0.00❭ ❬euros❭〗", got "〖❬1.23❭_❬US dollars❭〗〖❬1.23❭_❬euros❭〗〖❬0.00❭_❬US dollars❭〗〖❬0.00❭_❬euros❭〗"
    } (0.001s) FAILED (4 failure(s))
```

But the value we're using is "value-one". If anything it looks like it was using the wrong results.

macchiati: I don't understand. the following looks reasonable to me. `other` means plural, all all the examples are plural.
```
                "en",
                "other",
                "〖❬1.23 ❭Bermudan dollars〗〖❬0.00 ❭Bermudan dollars〗",
                "〖❬1.23❭ ❬US dollars❭〗〖❬1.23❭ ❬euros❭〗〖❬0.00❭ ❬US dollars❭〗〖❬0.00❭ ❬euros❭〗"
```

srl295: But this is the scenario the test sets up: the English value has been changed to `value-one` and `value-other` respectively. Just as with French.

I simulated the situation with `en_CA` locally:

image

Why does the test expect `Bermudan dollars` to be the value? I think it was passing because of ownership errors in the ExampleGenerator.  Compare to the French test in this same unit test, where the values are value-one and value-other.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -450,10 +443,7 @@ public ExampleGenerator(CLDRFile resolvedCldrFile, CLDRFile englishFile) {
         // GenerateExampleDependencies needs the ICUServiceBuilder instance to use the
         // RecordingCLDRFile if one is provided, rather than a pre-existing ordinary CLDRFile.
         boolean fileIsRecording = cldrFile.getClass() == RecordingCLDRFile.class;
-        this.icuServiceBuilder =
-                (locIsExceptional || fileIsRecording)
-                        ? new ICUServiceBuilder(resolvedCldrFile, locIsExceptional)
-                        : ICUServiceBuilder.forLocale(CLDRLocale.getInstance(localeId));
+        this.icuServiceBuilder = cldrFactory.getICUServiceBuilder(CLDRLocale.getInstance(localeId));
```
btangmu: This fails TestCLDRPaths.testGenerateExampleDependencies I recently added in https://github.com/unicode-org/cldr/pull/5761 . The value of boolean `fileIsRecording` can't be ignored; see the comment a few lines above. If an ExampleGenerator has a RecordingCLDRFile, then its ICUServiceBuilder needs to use that same RecordingCLDRFile.

The boolean `locIsExceptional` is also getting ignored with this change. It's not clear to me whether that's OK; maybe it is if the proposed `cldrFactory.getICUServiceBuilder` handles exceptional locales correctly, unlike `ICUServiceBuilder.forLocale`. If `locIsExceptional` isn't needed anymore it should be removed.

btangmu (dismissed):

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
srl295: I'm looking at that now. Yes exceptional should be removed, no more exceptional locales.

srl295: @btangmu take a look at the last update. This preserves the recording CLDRFile.

Even cleaner would be to have a recording **Factory**, that would work correctly (and would even cache the service builder by returning the same recording file…). Perhaps a future update.

btangmu: I'm not sure what you mean by "the last update". I see 4 commits; is there a 5th commit pending that I don't see yet?

srl295: ah yes, commited but not pushed. pushed it now.

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
srl295: > I don't understand. the following looks reasonable to me. `other` means plural, all all the examples are plural.
> 
> ```
>                 "en",
>                 "other",
>                 "〖❬1.23 ❭Bermudan dollars〗〖❬0.00 ❭Bermudan dollars〗",
>                 "〖❬1.23❭ ❬US dollars❭〗〖❬1.23❭ ❬euros❭〗〖❬0.00❭ ❬US dollars❭〗〖❬0.00❭ ❬euros❭〗"
> ```

More context:

```java
            String value = "value-" + row[1];

            String path = sampleCurrencyPrefix + row[1] + sampleTemplateSuffix;
            String result =
                    ExampleGenerator.simplify(exampleGenerator.getExampleHtml(path, value), false);
```

So the example is called with a value of `value-other` etc.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
```diff
@@ -684,12 +684,12 @@ public static void main(String[] args) throws Throwable {
...
                     UnicodeSet missingCurrencyExemplars = new UnicodeSet();
                     FlexibleDateFromCLDR fset =
-                            checkFlexibleDates ? new FlexibleDateFromCLDR(file) : null;
+                            checkFlexibleDates ? new FlexibleDateFromCLDR(cldrFactory, file) : null;
                     pathShower.set(localeID);
 
                     // only create if we are going to use it
                     final ExampleGenerator exampleGenerator =
-                            SHOW_EXAMPLES ? new ExampleGenerator(file, englishFile) : null;
+                            SHOW_EXAMPLES ? new ExampleGenerator(file, cldrFactory) : null;
```
macchiati, line 692: Maybe I'll understand more as I continue reviewing, but how does the generator know you want English in this case?

tools/cldr-code/src/main/java/org/unicode/cldr/util/Factory.java
```diff
@@ -351,17 +352,31 @@ public Set<CLDRLocale> calculateSubLocalesOf(CLDRLocale locale, Set<CLDRLocale>
...
 
+    /** thread-safe lazy load of TestCache */
+    private Supplier<TestCache> testCache = Suppliers.memoize(() -> new TestCache(this));
+
     /** get the TestCache owned by this Factory */
     public final TestCache getTestCache() {
         return testCache.get();
     }
 
+    /** thread-safe lazy load of ICUServiceFactory */
```
macchiati, line 366: A question: when a cldrFile changes on in the SurveyTool, how do you flush its icuServiceBuilder from the cache?

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -44,36 +74,23 @@ public class ICUServiceBuilder {
...
-        if (locIsExceptional) {
-            this.collationFile = null;
-        } else {
-            this.collationFile =
-                    Factory.make(CLDRPaths.COLLATION_DIRECTORY, ".*")
-                            .makeWithFallback(cldrFile.getLocaleID());
-        }
+        this.collationFactory = collationFactory;
+
+        this.collationFile = makeCollationFile(cldrFile.getLocaleID());
```
macchiati, line 89: The collation files could be built when needed (and they never change in the ST). Not a big deal, except that I think the big CJK collators take non-trivial time to build. But doesn't have to be part of this PR.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
macchiati: Yup, we need to revisit many of our old-code caches.

tools/cldr-code/src/main/java/org/unicode/cldr/util/NameGetter.java
macchiati: yup

tools/cldr-code/src/test/java/org/unicode/cldr/unittest/TestExampleGenerator.java
macchiati: Ah!

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
```diff
@@ -415,25 +415,17 @@ public void setVerboseErrors(boolean verbosity) {
...
-    }
-
     /**
      * Create an Example Generator. If this is shared across threads, it must be synchronized.
      *
      * @param resolvedCldrFile
      * @param englishFile
      */
-    public ExampleGenerator(CLDRFile resolvedCldrFile, CLDRFile englishFile) {
+    public ExampleGenerator(CLDRFile resolvedCldrFile, Factory cldrFactory) {
```
macchiati, line 424: I believe that the example generator could just use the config's english file, since that file never changes over runs of the survey tool, which would mean we could just dispense with the 2nd argument. But we don't need to look at that for this PR.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
```diff
@@ -23,16 +23,46 @@
...
+        private final Map<CLDRLocale, ICUServiceBuilder> ISBMap = new ConcurrentHashMap<>();
+
+        // TODO CLDR-19409: separate interface for collation?
+
+        public ICUServiceBuilder forLocale(final CLDRLocale loc) {
+            return ISBMap.computeIfAbsent(
+                    loc,
+                    newLoc ->
+                            new ICUServiceBuilder(
+                                    cldrFactory.make(newLoc.getBaseName(), true),
```
macchiati, line 48: Side note: I was wondering why you would make a locale from the name of a locale, but realized that name == id. I was also curious as to why CLDRLocale has two fields, basename and fullname. It appears from the constructor that these are always identical.

macchiati (commented):
I had a few questions

(and made a few side comments just because some of the code got me thinking)

tools/cldr-code/src/main/java/org/unicode/cldr/test/ExampleGenerator.java
srl295: That factory is needed to build the ICUServiceBuilder (and in fact it owns the ICUServiceBuilder). 
English can be loaded from that factory.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
srl295: (Sorry,i keep hitting resolve instead of reply)

Full name includes -u- extensions which aren't part of the base name. 

Constructing a CLDRLocale means we have a singleton, with parentage, and an object to hash on.

tools/cldr-code/src/main/java/org/unicode/cldr/test/ConsoleCheckCLDR.java
srl295: It will call factory.make("en") when needed.

tools/cldr-code/src/main/java/org/unicode/cldr/util/Factory.java
srl295: It should be listening for notifications of changes, but i can check this.  The factory it receives is a STFactory which is updated.

In practice at least for quarter names, i found I could vote for new names,and the examples and reports using `QQQ` were updated instantly.

tools/cldr-code/src/main/java/org/unicode/cldr/util/ICUServiceBuilder.java
srl295: Right. In fact, in general, there's probably more lazy processing that could be done.

macchiati (approved):
Ok, thanks for answering my questions

PR #5781, opened 2026-06-01
CLDR-13435 Clean up coverage regexes

robertbastian:
CLDR-13435

Improves readability.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/supplemental/coverageLevels.xml
+++ b/common/supplemental/coverageLevels.xml
@@ -69,8 +69,8 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageVariable key="%currency40" value="(BRL|CNY|EUR|GBP|INR|JPY|RUB|USD)"/>
 		<coverageVariable key="%currency60" value="(AUD|CAD|CHF|DKK|HKD|IDR|KRW|MXN|NOK|PLN|SAR|SEK|THB|TRY|TWD|ZAR)"/>
 		<coverageVariable key="%currency60_EU" value="(CZK|HUF)"/>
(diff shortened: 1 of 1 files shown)


srl295 (approved):
+10E6

never liked the compact regexes

common/supplemental/coverageLevels.xml
```diff
@@ -93,7 +93,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 		<coverageVariable key="%intervalFormatDateItems" value="(d|G?y|(G?y)?(M(MMM?)?)(E?d)?|GGGGGyM(E?d)?)"/>
 		<coverageVariable key="%intervalFormatTimeItems" value="((h|H)m?v?)"/>
 		<coverageVariable key="%intervalFormatTimeItemsDayPer" value="(Bhm?v?)"/>
-		<coverageVariable key="%intervalFormatGDiff" value="([GyMdaBHhm])"/>
+		<coverageVariable key="%intervalFormatGDiff" value="(G|y|M|d|a|B|H|h|m)"/>
```
macchiati, line 96: I agree with the above change, but this doesn't seem particularly more readable.

macchiati (changes requested):
Everything looks all more readable, except for the single character classes. I assume you tested the regular expressions to make sure that they capture exactly the same results.

common/supplemental/coverageLevels.xml
```diff
@@ -134,23 +134,23 @@ For terms of use, see http://www.unicode.org/copyright.html
...
+		<coverageVariable key="%script100" value="(Afak|Aghb|Ahom|Armi|Avst|Bali|Bamu|Bass|Batk|Berf|Blis|Brah|Bugi|Buhd|Cari|Cham|Chis|Chrs|Cirt|Copt|Cpmn|Cprt|Cyrs|Diak|Dogr|Dsrt|Dupl|Egyd|Egyh|Egyp|Elba|Elym|Gara|Geok|Glag|Gong|Gonm|Goth|Gran|Gukh|Hano|Hatr|Hluw|Hmng|Hmnp|Hrkt|Hung|Inds|Ital|Java|Jurc|Kali|Kawi|Khar|Khoj|Kits|Kpel|Krai|Kthi|Lana|Latf|Latg|Lepc|Limb|Lina|Linb|Lisu|Loma|Lyci|Lydi|Mahj|Maka|Mand|Mani|Maya|Medf|Mend|Merc|Mero|Modi|Moon|Mroo|Mult|Nagm|Nand|Narb|Nbat|Nkgb|Nshu|Ogam|Onao|Orkh|Osma|Ougr|Palm|Pauc|Pcun|Perm|Phag|Phli|Phlp|Phlv|Phnx|Plrd|Prti|Rjng|Roro|Runr|Samr|Sara|Sarb|Saur|Seal|Sgnw|Shaw|Shrd|Sidd|Sidt|Sind|Sogd|Sogo|Sora|Soyo|Sunu|Sylo|Syrc|Syre|Syrj|Syrn|Tagb|Takr|Tale|Talu|Tang|Tavt|Tayo|Teng|Tglg|Tirh|Tnsa|Todr|Tols|Toto|Tutg|Ugar|Visp|Vith|Wara|Wcho|Wole|Xpeo|Xsux|Yezi|Zanb|Zinh|Zmth)"/>
```
srl295, line 153: presumably this is "all scripts" anyway, so perhaps doesn't even need to be a regex. (probably a separate discussion)

macchiati:
I think in some of the coverage tickets this is discussed.


   - We almost never need something at the comprehensive coverage level
   except to *exclude* things.
   - And a whole lot of matches should be %A, because they match all of the
   possible items anyway.

I don't think that should stand in the way of this PR, because incremental
progress is better than none!

robertbastian:
> I assume you tested the regular expressions to make sure that they capture exactly the same results.

I generated them from the old ones, except for trivial changes (e.g. `language60_EU`)

common/supplemental/coverageLevels.xml
robertbastian: not using character classes is more consistent with the other regexes

macchiati (approved):

PR #5783, opened 2026-06-01
CLDR-19382 Model Europe/{Dublin,London} as metazones

robertbastian:
CLDR-19382

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -2958,6 +2958,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="Pacific/Wallis">
 				<exemplarCity>Wallis &amp; Futuna</exemplarCity>
 			</zone>
+			<metazone type="British">
+				<alias source="locale" path="../metazone[@type='GMT']"/>
+			</metazone>
+			<metazone type="Irish">
+				<alias source="locale" path="../metazone[@type='GMT']"/>
+			</metazone>
 		</timeZoneNames>
 	</dates>
 	<numbers>
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/LogicalGrouping.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/LogicalGrouping.java
@@ -49,6 +49,7 @@ public class LogicalGrouping {
                     "Azores",
                     "Bangladesh",
                     "Brasilia",
+                    "British",
                     "Cape_Verde",
                     "Chatham",
                     "Chile",
@@ -70,6 +71,7 @@ public class LogicalGrouping {
                     "Hong_Kong",
                     "Hovd",
                     "Iran",
+                    "Irish",
                     "Irkutsk",
                     "Israel",
                     "Japan",
@@ -108,6 +110,9 @@ public class LogicalGrouping {
                     "Volgograd",
                     "Yakutsk",
                     "Yekaterinburg");
+
+    public static final ImmutableSet<String> metazonesGmtInheritanceSet =
+            ImmutableSet.of("British", "Irish");
 
     public static final ImmutableList<String> days =
             ImmutableList.of("sun", "mon", "tue", "wed", "thu", "fri", "sat");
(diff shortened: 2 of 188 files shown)


common/main/root.xml
```diff
@@ -2958,6 +2958,12 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 			<zone type="Pacific/Wallis">
 				<exemplarCity>Wallis &amp; Futuna</exemplarCity>
 			</zone>
+			<metazone type="British">
+				<alias source="locale" path="../metazone[@type='GMT']"/>
```
robertbastian, line 2962: this doesn't seem to work, as a test is telling me this is null. ideas?

macchiati: Check the other aliases; it might be that you don't need the first ../

macchiati: I take that back, we have
```
			<unit type="duration-year">
				<displayName>yr</displayName>
				<unitPattern count="other">{0} y</unitPattern>
				<perUnitPattern>{0}/y</perUnitPattern>
			</unit>
			<unit type="duration-year-person">
				<alias source="locale" path="../unit[@type='duration-year']"/>
			</unit>
```

macchiati: We don't have a good console-based tool to inspect the inheritance chain.

A workaround might be to put your branch up on staging, and use the inheritance tool (the down arrow next to the "Info Panel" title.

Eg https://st.unicode.org/cldr-apps/v#/de/Duration/54e36e844ec917df
@srl295 is there a simpler console tool to do this?

srl295: There's not a CLI version of the inheritance tool, but there could be…

the API is on a CLDRFile:

`public List<LocaleInheritanceInfo> CLDRFile.getPathsWhereFound(String xpath)`

you can call toString() on the LocaleInheritanceInfo items in that list

macchiati (commented):
Looks good, just needs the tests to pass.

AEApple:
Looks like it's still failing. I'll go ahead with pushing the other fixes and I'll check back to see if this passes tests and is merged tomorrow for a follow up push to production.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/LogicalGrouping.java
```diff
@@ -349,6 +354,11 @@ void addPaths(Set<String> set, CLDRFile cldrFile, String path, XPathParts parts)
                 String metazoneName = parts.getAttributeValue(3, "type");
                 if (metazonesDSTSet.contains(metazoneName)) {
                     for (String str : ImmutableSet.of("generic", "standard", "daylight")) {
+                        if (metazonesGmtInheritanceSet.contains(metazoneName)
+                                && str.equals("standard")) {
+                            // Skip standard path for metazones that inherit GMT standard name
+                            continue;
```
srl295, line 360: @robertbastian this produces an unusable XPath for Irish/British.

Your check seems to assume aliases are already resolved, they're not…

`//ldml/dates/timeZoneNames/metazone[@type="British"]/alias[@source="locale"][@path="../metazone[@type='GMT']"]` »  `//ldml/dates/timeZoneNames/metazone[@type="British"]/alias[@source="locale"][@path="../generic`

srl295: it was half right. they are resolved, AND /alias is present. Bug.
https://github.com/unicode-org/cldr/pull/5948

robertbastian: can we just remove the alias when we resolve it, instead of ignoring it later? this is not the last time this will be confusing

srl295: > can we just remove the alias when we resolve it, instead of ignoring it later? this is not the last time this will be confusing

That's exactly why i filed https://unicode-org.atlassian.net/browse/CLDR-19671 (I think it will bring confusion again) — I propose that _by default_ the /alias is not included with the iterator. If some tool really needs to know about aliases, there could be a `.fullIterableWithAliases()` function.

robertbastian: Ah I thought https://github.com/unicode-org/cldr/pull/5948 was the end of it.

PR #5792, opened 2026-06-02
CLDR-18953 ddl: load Noto Sunuwar font

srl295:
CLDR-18953

image

ALLOW_MANY_COMMITS=true


conradarcturus (approved):
Cool! Today I learned.

PR #5793, opened 2026-06-02
CLDR-19434 Make cursor visible after using insert menu

btangmu:
-Call both setRangeText and setSelectionRange to ensure insertion point is updated correctly

-If a range is highlighted before choosing from the menu, replace that range

-Encapsulate/confirm Ant Vue input element assumption in new method getInputElement

CLDR-19434

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/views/AddValue.vue
+++ b/tools/cldr-apps/js/src/views/AddValue.vue
@@ -19,6 +19,7 @@
     <div>
       <a-config-provider :direction="dir">
         <span class="show-as-text">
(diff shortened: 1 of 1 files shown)


btangmu:
This took me a lot longer than expected. One mistake was calling `setSelectionRange` on `inputToFocus.value` when it should have been on `inputToFocus.value.input`. Another problem was that even though `setValue` was sufficient to update the text correctly, the insertion point always wrongly moved to the very end of the text; this was solved by adding a call to `setRangeText`.

srl295 (approved):

PR #5795, opened 2026-06-02
CLDR-14549 remove Zawgyi fix from DAIP

srl295:
CLDR-14549

ALLOW_MANY_COMMITS=true


AEApple:
Is there a reason we're just not removing the old test instead of disabling it if we believe that we shouldn't use it going forward? What would still use it in CLDR?

srl295:
> Is there a reason we're just not removing the old test instead of disabling it if we believe that we shouldn't use it going forward? What would still use it in CLDR?

I want to use it against the warning, then i'll remove it.

macchiati (approved):

PR #5796, opened 2026-06-02
CLDR-19353 Revise display of tags (chits)

btangmu:
-Make the borders thicker and darker

-Show some additional characters (quotes/apostrophes/primes) as tags but do not include them in the Insert menu

-Add documentation for the additional characters in getting-started/guide.md

-Minor related refactoring and comments

-(Unrelated:) Suppress distracting console warnings for intentionally invalid locale IDs in TestCldrLocales

CLDR-19353

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-apps/js/src/esm/cldrChar.mjs
+++ b/tools/cldr-apps/js/src/esm/cldrChar.mjs
@@ -6,6 +6,32 @@ import { unicodeName } from "unicode-name";
 
 import * as cldrEscaper from "./cldrEscaper.mjs";
 
+/*
+  For clarity, these characters are displayed as tags, but they are not included
+  in the Insert menu. They are not invisibles, so the characters themselves, rather
+  than their names, are displayed in the tags. Users can see their names by hovering.
+
+  ◦	" U+0022 QUOTATION MARK
+  ◦	“ U+201C LEFT DOUBLE QUOTATION MARK
+  ◦	” U+201D RIGHT DOUBLE QUOTATION MARK
+  ◦	″ U+2033 DOUBLE PRIME
+  ◦	ʼ U+02BC MODIFIER LETTER APOSTROPHE
+  ◦	‘ U+2018 LEFT SINGLE QUOTATION MARK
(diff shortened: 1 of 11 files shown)


macchiati (changes requested):
Looks good. I think we want the border to be a bit more eye catching , like dark blue. Ow, ok.

btangmu:
> we want the border to be a bit more eye catching , like dark blue

The 2nd commit changes it from black to midnightblue. Per web search, "MidnightBlue (#191970) is actually lighter and less saturated than DarkBlue (#00008B)" and on a cyan background "Its muted 64% saturation structure stops the layout from looking too jarringly blocky."

btangmu:
image

tools/cldr-apps/js/src/esm/cldrChar.mjs
```diff
@@ -6,6 +6,32 @@ import { unicodeName } from "unicode-name";
 
 import * as cldrEscaper from "./cldrEscaper.mjs";
 
+/*
+  For clarity, these characters are displayed as tags, but they are not included
+  in the Insert menu. They are not invisibles, so the characters themselves, rather
+  than their names, are displayed in the tags. Users can see their names by hovering.
+
+  ◦	" U+0022 QUOTATION MARK
```
macchiati, line 14: Either we should have both or neither of these:

- U+0022 QUOTATION MARK
- U+0027 APOSTROPHE

I'd lean towards neither.
(This was probably my fault in supplying the list.)

macchiati (changes requested):
The rest looks good, though my JS is very weak. @srl295 should review that if available. Otherwise I could stamp once this is resolved.

srl295 (approved):
JS looks fine. We should use a logging framework ideally rather than lots of different debug/quiet options.

macchiati:
Looks good!

macchiati (approved):

tools/cldr-apps/js/src/esm/cldrChar.mjs
btangmu: I'll remove U+0022 in a follow-up.

PR #5797, opened 2026-06-02
CLDR-18980 fix error messages

macchiati:
CLDR-18980

ALLOW_MANY_COMMITS=true


common/main/root.xml
AEApple, line 400: Should this gregorian also be generic for consistent behavior?

srl295 (approved):

PR #5800, opened 2026-06-03
CLDR-18980 Turn some more errors into warnings

macchiati:
CLDR-18980

The wording in the TC agenda was: "Turn errors into warnings only if one or more of the items in the new date separator test has a value at contributed or higher if it is not in their coverage level, or it’s a CLDR locale with a target level of moderate or higher"

* I renamed some items for clarity (to basePath and baseValue)
* Reorganized the code from an inline if statement to make it clearer.
* Added two new helper functions, also for clarity.
* Changed the position of the Numeric Separators on the calendar page to be more logical (after the lists of eras, quarters, months, etc. and just before the standard formats (full, long ...)

The wording in the agreement was a bit tricky (we were editing quickly), so this is my take on what we meant. I'd asked for the CLDR locale clause to be added, but I don't think that is necessary. 

| Locale | **TC** coverage target | NS Path | Draft Status | Result |
| -- | -- | -- | -- | -- |
| French | Modern | Comprehensive |  provisional |  **warning** |
| French | Modern | Not Comp. |  provisional |  error |
| French | Modern | Comprehensive |  approved |  error |
| French | Modern | Not Comp. |  approved |  error |

Once I get that right I can incorporate the logic into the check.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/common/main/root.xml
+++ b/common/main/root.xml
@@ -2005,7 +2005,8 @@ Warnings: All cp values have U+FE0F characters removed. See /annotationsDerived/
 				</timeFormats>
 				<dateTimeFormats>
 					<numericSeparators>
-						<alias source="locale" path="../../../calendar[@type='gregorian']/dateTimeFormats/numericSeparators"/>
+						<numericDateSeparator>-</numericDateSeparator>
+						<numericTimeSeparator>:</numericTimeSeparator>
 					</numericSeparators>
 					<dateTimeFormatLength type="full">
 						<dateTimeFormat>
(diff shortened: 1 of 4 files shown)


AEApple:
Looks great on staging! Agree it would be nice to update the header so that the separators appear in a more logical part of the page.

AEApple (dismissed):

macchiati:
Hi, I'd like a restamp for the change in the path header to move it down the page.

AEApple:
Why is Docker failing? It wasn't failing before.

macchiati:
It seems like half the time docker fails randomly. Not a blocker so you can approve

AEApple (approved):

macchiati:
Merged. So we can push now.

PR #5801, opened 2026-06-03
CLDR-14549 add a warning on misencoded zawgyi

srl295:
CLDR-14549

Message is like this:

> ⚠️ Value appears to contain Zawgyi (97% confidence)

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
@@ -7,9 +7,11 @@
 package org.unicode.cldr.test;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.Multiset;
 import com.google.common.collect.Multiset.Entry;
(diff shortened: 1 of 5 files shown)


macchiati (dismissed):
LGTM

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
```diff
@@ -586,6 +599,33 @@ public CheckCLDR handleCheck(
...
+    private void checkEncodingMyanmar(String value, List<CheckStatus> result) {
+        double zawgyiProbability = detector.get().getZawgyiProbability(value);
+        if (zawgyiProbability > ZAWGYI_LIKELY) {
+            result.add(
+                    new CheckStatus()
+                            .setCause(this)
+                            .setMainType(CheckStatus.warningType)
+                            .setSubtype(Subtype.misencodedZawgyi)
+                            .setMessage(
+                                    "Value appears to contain Zawgyi with {0,number,percent} confidence."));
```
srl295: message isn't quite right, let me try again

macchiati (approved):

PR #5804, opened 2026-06-05
CLDR-19430 GenerateBirth, docs and clean-up

btangmu:
-Update the binary files generated by running GenerateBirth again; I don't know why they are changed

-Update updating-englishroot.md, for clarity, and with instructions for handling TestCheckCLDR.TestCheckNew failure

-Make messages clearer in GenerateBirth.java

-Fix double-slash in path for englishDataFile

-Change comments in TestCheckCLDR.TestCheckNew to reference updating-englishroot.md

-Fix compiler warnings

-Comments

CLDR-19430

ALLOW_MANY_COMMITS=true


docs/site/development/cldr-development-site/updating-englishroot.md
```diff
@@ -4,36 +4,54 @@ title: Updating English/Root
...
-**cldr (set with \-t \<target\>, default\=CldrUtility.BASE\_DIRECTORY, set with environment variable \-DCLDR\_DIR)**
+1. An archive directory (corresponding to the cldr-archive repository), containing a set of sources from all previous major CLDR releases, which defaults to ../cldr-archive or CLDRPaths.ARCHIVE_DIRECTORY, and may be set with the environment variable CLDR_ARCHIVE
+2. Trunk (corresponding to the cldr repository), which defaults to the current directory or may be set with the environment variable CLDR\_DIR
+3. An output directory for the binary data files it generates, which defaults to tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/births (replacing the previous versions), and may be set by passing an argument for \-t to specify the output directory
+4. An output directory for the readable .txt files it generates (corresponding to the cldr-staging repository; for example, ../cldr-staging/births/49.0)
 
-... common/ ... tools/ java/ (apps such as GenerateBirth are run from here) ...
+The archive directory should have the latest version of every major and minor version (where versions before 21\.0 have the major version split across the top two fields).
```
srl295: This should be a note under line 29 - since the CheckoutArchive tool will automatically do exactly this. In this order, it sounds like it's a prerequisite to have the archive directory up to date before proceeding.

srl295 (dismissed):
Fine with a comment.

And, should we really check in the 'human readable' files? Maybe we should just skip them, they are debugging output to the generator and can be reproduced at any time.

Also since they are in a separate repo, it would hinder automatic updates of the data.

btangmu:
> can be reproduced at any time

That's not clear to me, for whom all this is very new and confusing. How would we reproduce the .txt files for version 45 now? Would suitable versions of the three repos -- cldr, cldr-archive, and cldr-staging, all need to be checked out at the same time? What if the old code wouldn't compile with our current dev tools? What if the files generated now are different from what would have been generated a few years ago, and we could never know that without having saved the old versions? The binary files are opaque, and if there are no corresponding text files, we're just trusting that they're good. Maybe it's not important, I don't know, but either way we should be clear about what is expected for completion of a BRS task.

I've made a new ticket: https://unicode-org.atlassian.net/browse/CLDR-19536

docs/site/development/cldr-development-site/updating-englishroot.md
btangmu: I've addressed this in the 2nd commit.

srl295 (approved):

docs/site/development/cldr-development-site/updating-englishroot.md
```diff
@@ -56,23 +74,31 @@ In a limited release, the file **SubmissionLocales.java** is set up to allow jus
...
+Here, **cldr\-staging** corresponds to a git repo distinct from both **cldr** and **cldr\-archive**. Online this data is found in https://github.com/unicode-org/cldr-staging/tree/main/births .
 
-Eg https://github.com/unicode-org/cldr/pull/243
+The readable data is easiest to read if pasted into a spreadsheet.
 
-## Debugging
+**Check the readable data into the cldr\-staging repo, for comparison over time.**
 
-It also generates readable log files for double checking. These will be in {workspace}/cldr\-aux/births/\<version\>/, that is: CLDRPaths.AUX\_DIRECTORY \+ "births/" \+ trunkVersion. Examples: https://unicode.org/repos/cldr-aux/births/35.0/en.txt, https://unicode.org/repos/cldr-aux/births/35.0/fr.txt.
+TODO: The readable data has not been checked into the cldr-staging repo since version 44. Reference: https://unicode-org.atlassian.net/browse/CLDR-19536
```
btangmu, line 99: Note the TODO and new ticket reference

srl295:
> > can be reproduced at any time
> 
> That's not clear to me, for whom all this is very new and confusing. How would we reproduce the .txt files for version 45 now? Would suitable versions of the three repos -- cldr, cldr-archive, and cldr-staging, all need to be checked out at the same time? What if the old code wouldn't compile with our current dev tools? What if the files generated now are different from what would have been generated a few years ago, and we could never know that without having saved the old versions? The binary files are opaque, and if there are no corresponding text files, we're just trusting that they're good. Maybe it's not important, I don't know, but either way we should be clear about what is expected for completion of a BRS task.

if i run GenerateBirths from CLDR 45.0, it works and outputs cldr-staging/births/45.0/en.txt etc.   

The files wouldn't be different from what was generated a few years ago - the inputs are the same.

> I've made a new ticket: https://unicode-org.atlassian.net/browse/CLDR-19536

I commented on it and linked to a bug in CldrVersion

docs/site/development/cldr-development-site/updating-englishroot.md
```diff
@@ -4,36 +4,54 @@ title: Updating English/Root
...
 
-The tool generates (among other things) the following two binary files (among others) in the output directory specified with \-t:
+To specify the oldest version to go back to, pass an argument for \-o (such as \-o36.1 for version 36.1).
+
+Example command line:
+```
+java -jar tools/cldr-code/target/cldr-code.jar GenerateBirth -o46
+```
+
+GenerateBirth takes a few minutes to run, and may require setting Java with enough memory.
```
AEApple, line 45: How much memory is 'enough memory'? Is there a recommended minimum amount?

btangmu:
> if i run GenerateBirths from CLDR 45.0, it works and outputs cldr-staging/births/45.0/en.txt etc.

How did you run it from CLDR 45.0?

> The files wouldn't be different from what was generated a few years ago - the inputs are the same.

Sorry I'm dense, but what are the inputs? Three repos are potentially involved, and maybe some are only outputs, not inputs, but that's not obvious. If bug CLDR-19537 "prevents a cldr-archive from the future to be used with a prior CLDR version" doesn't that sort of confirm my suspicion that the task is non-trivial and not guaranteed to succeed if postponed?

docs/site/development/cldr-development-site/updating-englishroot.md
btangmu: I don't know; I have the environment variable `JAVA_OPTIONS=-Xmx3G` set locally, which seems to suffice on my computer -- I just made slight changes to the sentence; it was "make sure you have set Java with enough memory". Maybe it's not a helpful message, or maybe we should specify `JAVA_OPTIONS=-Xmx3G`

srl295:
> > if i run GenerateBirths from CLDR 45.0, it works and outputs cldr-staging/births/45.0/en.txt etc.
> 
> How did you run it from CLDR 45.0?
> 
> > The files wouldn't be different from what was generated a few years ago - the inputs are the same.
> 
> Sorry I'm dense, but what are the inputs? Three repos are potentially involved, and maybe some are only outputs, not inputs, but that's not obvious. 

Only one repo is involved, to do the task.  

The input is the current CLDR version, plus the prior versions in cldr-archive.
The output is to update the births files in the current CLDR version, plus there's an output of text files in the cldr-staging directory. (debug logs)

It could be done minimally as follows:

1. checkout CLDR, let's call it `/cldr` for simplicity
2. switch to tag `release-45`
3. `mkdir ../cldr-archive`
4. run the `CheckoutArchive` tool. It will only checkout cldr-archive through v44 (since nothing later had been released)
5. run the `GenerateBirth` tool

That's it. It will output in-repo, and create `../cldr-staging/births/45.0/en.txt`

I basically did the above, but I also added `-DARCHIVE=${HOME}/src/cldr-archive-45/` to keep a separate cldr-archive directory.

> If bug [CLDR-19537](https://unicode-org.atlassian.net/browse/CLDR-19537) "prevents a cldr-archive from the future to be used with a prior CLDR version" doesn't that sort of confirm my suspicion that the task is non-trivial and not guaranteed to succeed if postponed?

It adds challenge. but I think it's just a simple bug that should be fixed.

Since the task only depends on the current+prior CLDR versions, I don't see why we need to keep its logfile around (especially if we change to running it frequently in an automated way)

[CLDR-19537]: https://unicode-org.atlassian.net/browse/CLDR-19537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

btangmu:
Thanks, that's helpful. Seems like the doc could be improved by clarifying the roles of ../cldr, ../cldr-archive, and ../cldr-staging -- what's input and what's output, and how to use a tag like release-45.

About whether to keep the log (.txt) files in the cldr-staging repo, that's a policy that should be decided on one way or the other. Note that the last paragraph of uploading-englishroot.md (moved from TestCheckNew) now reads as follows; if anyone still wants to do it that way, they'll have to go through an extra step of generating those files, and they won't be able to get them from the cldr-staging repo.

> Another way to choose appropriate paths for TestCheckNew, if cldr-staging/births is up to date, is to look in the log file such as in https://github.com/unicode-org/cldr-staging/blob/main/births/41.0/fr.txt (for the current version, not necessarily 41.0) for a reasonable locale (not necessarily fr). Find a path that is outdated. To work on both limited and full submissions, choose one with English = trunk. Sometimes the English change is suppressed in a limited release if the change is small. Pick another in that case. Check the data files to ensure that it is in fact outdated. Change the path to that value. The 3rd parameter is the message displayed to the user, or "" if not 'English Changed'. So the first group of tests are for items that should not be outdated, and the second group is ones that should be outdated.

PR #5810, opened 2026-06-09
CLDR-17854 Show multiple numbering systems in date report

btangmu:
-Avoid methods that implicitly assume defaultNumberingSystem if numberingSystem is omitted

-New ICUServiceBuilder.NUMBERING_SYSTEM_DEFAULT = null serves as explicit marker for using defaultNumberingSystem

-Move some of the logic from SurveyAjax.generateDateTimesReport to new DateTimeFormats.generateReport

-Refactoring: prefer existing constants like CLDRFile.NumberingSystem.defaultSystem.path in place of repeating string literals like //ldml/numbers/defaultNumberingSystem

-TODO comments for extending beyond date report to show multiple numbering systems also in charts and examples

-Fix compiler warnings; remove dead code

-Comments

CLDR-17854

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
@@ -427,10 +427,10 @@ class LdmlConvertRules {
             Pattern.compile(
                     "//ldml/numbers/(symbols|miscPatterns|(decimal|percent|scientific|currency|rational)Formats)\\[@numberSystem=\"([^\"]++)\"\\]/.*");
     public static final String[] ACTIVE_NUMBERING_SYSTEM_XPATHS = {
-        "//ldml/numbers/defaultNumberingSystem",
-        "//ldml/numbers/otherNumberingSystems/native",
-        "//ldml/numbers/otherNumberingSystems/traditional",
-        "//ldml/numbers/otherNumberingSystems/finance"
+        CLDRFile.NumberingSystem.defaultSystem.path,
+        CLDRFile.NumberingSystem.nativeSystem.path,
+        CLDRFile.NumberingSystem.traditional.path,
+        CLDRFile.NumberingSystem.finance.path,
     };
 
     /** resolved identity should be discarded if inherited, known issue CLDR-17790 */
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
@@ -16,14 +16,14 @@ import com.ibm.icu.util.GregorianCalendar;
 import com.ibm.icu.util.ICUUncheckedIOException;
 import com.ibm.icu.util.Output;
 import com.ibm.icu.util.TimeZone;
-import com.ibm.icu.util.ULocale;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
-import java.util.EnumSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Locale;
 import java.util.Map;
@@ -33,6 +33,7 @@ import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.unicode.cldr.draft.FileUtilities;
+import org.unicode.cldr.icu.LDMLConstants;
 import org.unicode.cldr.tool.ChartDelta;
 import org.unicode.cldr.tool.FormattedFileWriter;
 import org.unicode.cldr.tool.Option;
@@ -72,7 +73,6 @@ public class DateTimeFormats {
     private static final String TIMES_24H_TITLE = "Times 24h";
     private static final boolean DEBUG = false;
     private static final String DEBUG_SKELETON = "y";
-    private static final ULocale DEBUG_LIST_PATTERNS = ULocale.JAPANESE; // or null;
 
     private static final String FIELDS_TITLE = "Fields";
 
@@ -200,7 +200,7 @@ public class DateTimeFormats {
      * @return the request name, e.g., "Era", or null if not found
      */
     public static String getAppendRequestName(char fieldChar) {
-        int field = -1;
+        int field;
         try {
             field =
                     new com.ibm.icu.text.DateTimePatternGenerator.VariableField(
@@ -223,7 +223,7 @@ public class DateTimeFormats {
      * @return the field name key, e.g., "era", or null if not found
      */
     public static String getFieldDisplayNameKey(char fieldChar) {
-        int field = -1;
+        int field;
         try {
             field =
                     new com.ibm.icu.text.DateTimePatternGenerator.VariableField(
@@ -248,8 +248,6 @@ public class DateTimeFormats {
     };
     private static final Date SAMPLE_DATE = new Date(2012 - 1900, 0, 13, 14, 45, 59);
 
-    private static final String SAMPLE_DATE_STRING = CldrUtility.isoFormat(SAMPLE_DATE);
-
     private static final Map<String, Date> SAMPLE_DATE_END =
             ImmutableMap.<String, Date>builder()
                     .put("G", SAMPLE_DATE_DEFAULT_END)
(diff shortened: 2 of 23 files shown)


btangmu:
[Fixed -- was a false alarm] This isn't right yet. Testing locally, even for a locale like "fr" with only one numbering system, the report has "mars 1missing" under "Ordinal Dates", where on cldr-smoke it's "mars 1e"

btangmu:
The "missing" comes from ICUServiceBuilder.formatWithOrdinalHack:

```
    public static String formatWithOrdinalHack(
            SimpleDateFormat sdf, String calendar, Date date, CLDRFile cldrFile) {
        String pattern = sdf.toPattern();
        if (!pattern.contains("ddd")) {
            return sdf.format(date);
        }
        PluralRules ordinalRules =
                supplementalData.getPluralRules(
                        cldrFile.getLocaleID(), PluralRules.PluralType.ORDINAL);
        if (ordinalRules == null) {
            return sdf.format(date);
        }
        int dayOfMonth = date.getDate();
        String keyword = ordinalRules.select(dayOfMonth, 0, 0);
        String path =
                CldrPathUtilities.dayOfMonthPath(
                        Count.valueOf(keyword), calendar, "format", "abbreviated");
        String ordinalPattern = cldrFile.getStringValueWithBailey(path);
        if (ordinalPattern == null) {
            ordinalPattern = "{0}missing";
```

Locally, path = `//ldml/dates/calendars/calendar[@type="gregorian"]/dayOfMonths/dayOfMonthContext[@type="format"]/dayOfMonthWidth[@type="abbreviated"]/dayOfMonth[@ordinal="one"]`

and if I look it up with XPath Calculator, it's not found. But on cldr-smoke, XPath Calculator does find it as 5aff1f2a84ad62f0. So the difference is in the database?

https://cldr-smoke.unicode.org/cldr-apps/v#/fr/Gregorian/5aff1f2a84ad62f0 -- winning value = "{0}e"

https://st.unicode.org/cldr-apps/v#/fr/Gregorian/5aff1f2a84ad62f0 -- winning value = "{0}er"

http://localhost:9080/cldr-apps/v#/fr/Gregorian/5aff1f2a84ad62f0 -- no winning value

I'll try updating my local database...

btangmu:
After db update, the report looks good for both ar and fr. I'm making this ready for review.

btangmu:
I see now that even on production, the report does have "missing" for some locales, such as "aa":

https://st.unicode.org/cldr-apps/v#r_datetime/aa//

...
YW 2012 Qunxa Garablu 13missing, Gum
...

BTW that's different from other examples that have "ⓜⓘⓢⓢⓘⓝⓖ"

srl295:
> I see now that even on production, the report does have "missing" for some locales, such as "aa":
> 
> https://st.unicode.org/cldr-apps/v#r_datetime/aa//
> 
> ... YW 2012 Qunxa Garablu 13missing, Gum ...
> 
> BTW that's different from other examples that have "ⓜⓘⓢⓢⓘⓝⓖ"

these are ordinal dates, and 'missing' is coming from the ICUServiceBuilder. CLDR-19419

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -1175,9 +1155,13 @@ public static void main(String[] args) throws IOException {
                             + "</h1>"
                             + "<p><a href='index.html'>Index</a></p>\n"
                             + "<p>The following chart shows typical usage of date and time formatting with the Gregorian calendar and default number system.</p>\n");
-            formats.addTable(english, out);
-            formats.addDateTable(englishFile, out);
-            formats.addDayPeriods(englishFile, out);
+            // TODO: loop through numbering systems for this locale
+            // Reference: https://unicode-org.atlassian.net/browse/CLDR-17854
+            String numberingSystem =
+                    cldrFile.getStringValue("//ldml/numbers/defaultNumberingSystem");
```
srl295, line 1161: ?

```suggestion
                    cldrFile.getStringValue(CLDRFile.NumberingSystem.defaultSystem.path);
```

srl295 (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
```diff
@@ -427,10 +427,10 @@ public static boolean attrValueAsArraySet(final String nodeName, final String ke
             Pattern.compile(
                     "//ldml/numbers/(symbols|miscPatterns|(decimal|percent|scientific|currency|rational)Formats)\\[@numberSystem=\"([^\"]++)\"\\]/.*");
     public static final String[] ACTIVE_NUMBERING_SYSTEM_XPATHS = {
-        "//ldml/numbers/defaultNumberingSystem",
-        "//ldml/numbers/otherNumberingSystems/native",
-        "//ldml/numbers/otherNumberingSystems/traditional",
-        "//ldml/numbers/otherNumberingSystems/finance"
+        CLDRFile.NumberingSystem.defaultSystem.path,
```
macchiati, line 430: Better to use List.of(...)

Arrays in Java are mutable.

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -870,10 +855,7 @@ private void showRow(
         output.append("<tr>");
         switch (rowStyle) {
             case separator:
-                String link = name.replace(' ', '_');
-                output.append("<th colSpan='3' class='dtf-sep'>")
-                        .append(hackDoubleLinked(link, name))
-                        .append("</th>");
+                output.append("<th colSpan='3' class='dtf-sep'>").append(name).append("</th>");
```
macchiati, line 858: In a few places you seem to be dropping links. I'd like to understand why

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -1003,10 +981,10 @@ public void addDateTable(CLDRFile english, Appendable output) {
                     "thu",
                     "fri",
                     "sat");
-            output.append("<h2>" + hackDoubleLinked("Ordinal Dates") + "</h2>\n");
-            addOrdinalDateSubtable(output);
+            output.append("<h2>Ordinal Dates</h2>\n");
+            addOrdinalDateSubtable(output, numberingSystem);
```
macchiati, line 985: A thought for the future. We might make an enum NumSystem, for type safety, instead of passing strings. It would have the BCP values plus DEFAULT

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -1397,32 +1395,6 @@ private String getCleanValue(String evalue, Width width, String fallback) {
...
-    //        } else {
-    //            pattern = pattern.replace('a', '\uE000');
-    //        }
-    //        SimpleDateFormat df = icuServiceBuilder.getDateFormat("gregorian", pattern);
-    //        String formatted = df.format(timeInDay);
-    //        String result = formatted.replace("\uE000", evalue);
-    //        return result;
-    //    }
-
-    private String hackDoubleLinked(String link, String name) {
```
macchiati, line 1418: Ah, may have been a noop

macchiati (dismissed):
Nice work, Tom!

tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
btangmu: OK, it was already an array, but I can change it to a list

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
btangmu: `hackDoubleLinked` was a no-op, just returning the name. I don't know the history of that, maybe it was a feature that got broken or was never implemented...

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
btangmu: Good idea!

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
btangmu: Yes

tools/cldr-code/src/main/java/org/unicode/cldr/json/LdmlConvertRules.java
macchiati: Right. We have a lot of old code that uses arrays when we should be using immutable lists. It's most important when they are static and presumed to be constant.

btangmu:
@macchiati and @srl295 your suggested refactoring improvements are all good. I would be inclined to do them in a follow up PR than make changes to this PR, since it already has 23 changed files...

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
btangmu: Good point. I can fix that along with the TODO for charts

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
srl295: A wild guess is that it had the purpose of having:

`<a href="https://cldr.unicode.org/…/Ordinal_Dates" name="Ordinal_Dates">Ordinal Dates</a>` where 'double linked means it's both a link and an anchor.

PR #5812, opened 2026-06-10
CLDR-19545 exemplar aux/main an date/time separator conflicts should be warning in submission

srl295:
CLDR-19545

image

image

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckDates.java
@@ -820,7 +820,7 @@ public class CheckDates extends FactoryCheckCLDR {
             Set<String> separatorFromBase = extractNumericSeparator(basePath, baseValue);
             if (!separatorFromBase.contains(value)) {
                 String separatorsFromBase = Joiners.COMMA_SP.join(separatorFromBase);
-                CheckStatus.Type errorType = getErrorTypeButWarningInBuild();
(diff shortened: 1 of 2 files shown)


AEApple (approved):

AEApple:
Can we document the different error types available in CheckCLDR.java in the dev docs and when to use them to make it easier for people to decide which error type to use for new tests on data in the future?

macchiati (approved):

btangmu (approved):

macchiati:
> Can we document the different error types available in CheckCLDR.java in the dev docs and when to use them to make it easier for people to decide which error type to use for new tests on data in the future?

Formally there are just two error types (error and warning). But we have various special cases where we turn errors into warnings that we handle via code.

There are a host of different error subtypes. Those mostly affect the message provided to users.

So I think what you are asking for is some documentation and consistency among the "when errors become warnings".

srl295:
> > Can we document the different error types available in CheckCLDR.java in the dev docs and when to use them to make it easier for people to decide which error type to use for new tests on data in the future?
> 
> Formally there are just two error types (error and warning). But we have various special cases where we turn errors into warnings that we handle via code.
> 
> There are a host of different error subtypes. Those mostly affect the message provided to users.
> 
> So I think what you are asking for is some documentation and consistency among the "when errors become warnings".

I captured this at https://unicode-org.atlassian.net/browse/CLDR-15430?focusedCommentId=210956 where I am investigating phase sensitivity.

PR #5814, opened 2026-06-10
CLDR-17854 Show multiple numbering systems in charts

btangmu:
-Loop through numbering systems for charts as was already done for reports

-In CodeFallback, fix the ordering of type and key in //ldml/localeDisplayNames/types/type/... (it got normalized anyway, but better to generate correctly in the first place)

-Refactor ACTIVE_NUMBERING_SYSTEM_XPATHS by making it a List (since Array is mutable)

-Add comment in ICUServiceBuilder about numberingSystem attributes containing equals signs

CLDR-17854

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/CodeFallback.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/CodeFallback.java
@@ -223,11 +223,11 @@ public class CodeFallback {
         for (String[] typeDisplayName : typeDisplayNames) {
             constructedItems.putValueAtPath(
                     "//ldml/localeDisplayNames/types/type"
-                            + "[@key=\""
-                            + typeDisplayName[1]
-                            + "\"]"
                             + "[@type=\""
                             + typeDisplayName[0]
+                            + "\"]"
+                            + "[@key=\""
+                            + typeDisplayName[1]
                             + "\"]",
                     typeDisplayName[0]);
         }
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
@@ -1155,13 +1155,7 @@ public class DateTimeFormats {
                             + "</h1>"
                             + "<p><a href='index.html'>Index</a></p>\n"
                             + "<p>The following chart shows typical usage of date and time formatting with the Gregorian calendar and default number system.</p>\n");
-            // TODO: loop through numbering systems for this locale
-            // Reference: https://unicode-org.atlassian.net/browse/CLDR-17854
-            String numberingSystem =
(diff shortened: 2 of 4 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/util/CodeFallback.java
```diff
@@ -223,11 +223,11 @@ public static XMLSource getConstructedItems() {
...
             constructedItems.putValueAtPath(
                     "//ldml/localeDisplayNames/types/type"
-                            + "[@key=\""
-                            + typeDisplayName[1]
-                            + "\"]"
                             + "[@type=\""
                             + typeDisplayName[0]
+                            + "\"]"
+                            + "[@key=\""
+                            + typeDisplayName[1]
```
btangmu, line 230: While looking at how to get a list of numbering systems here (for possible future enum), I noticed the normalized paths are like this:
```
"//ldml/localeDisplayNames/types/type[@type=\"arab\"][@key=\"numbers\"]"
```

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
```diff
@@ -1505,7 +1510,15 @@ private static void writeNumberingSystemLinks(
             String link = numberingSystemLinkId(numberingSystem);
             String onClick = "document.getElementById('" + link + "').scrollIntoView()";
             String kind = numberingSystems.get(numberingSystem);
-            out.write("<a onclick=\"" + onClick + "\">" + numberingSystem + " (" + kind + ")</a>");
+            // For some reason, in the chart, this link lacks underline unless added here explicitly
+            out.write(
+                    "<a style=\"text-decoration: underline;\" onclick=\""
```
macchiati, line 1515: In general, putting it into the CSS is cleaner. Not a blocker though.

macchiati (approved):

tools/cldr-code/src/main/java/org/unicode/cldr/util/DateTimeFormats.java
btangmu: Yes. I looked at ../cldr-staging/docs/charts/49/index.css trying to understand why the links lack underlines, but without success. Also I see code like this:
```
FileCopier.copy(GenerateAllCharts.class, "index.css", CLDRPaths.CHART_DIRECTORY);
```
And I noticed some slight differences like this:
```
$ diff ../cldr-staging/docs/charts/index.css  ../cldr-staging/docs/charts/49/index.css
12c12,13
< 	text-align: left
---
> 	text-align: left;
> 	padding: 0.25em
```
Yet the source file (?) seems not to have changed in 6 years:

https://github.com/unicode-org/cldr-staging/blob/main/docs/charts/index.css

I can follow up by editing that file but I'd like to understand it better to avoid unintended consequences

PR #5815, opened 2026-06-10
CLDR-19547 LOCALES_ALLOW_ADDING_IN_VETTING

srl295:
CLDR-19547

image

doesn't add anything to UI, except that the + is available in VETTING
(similar locale not listed was not included)

activate with

```properties
CLDR_ADD_ITEMS_IN_VETTING_LOCALES=ta
```

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
@@ -284,9 +284,9 @@ public abstract class CheckCLDR implements CheckAccessor {
                         : StatusAction.ALLOW_VOTING_AND_TICKET;
             }
 
-            // locales in CLDR_ADD_ITEMS_IN_VETTING_LOCALES may continue to be added
+            // locales in LOCALES_ALLOW_ADDING_IN_VETTING may continue to be added
             if (pathValueInfo != null
-                    && SubmissionLocales.CLDR_ADD_ITEMS_IN_VETTING_LOCALES.contains(
+                    && SubmissionLocales.LOCALES_ALLOW_ADDING_IN_VETTING.contains(
                             pathValueInfo.getLocale().toString())) {
                 return StatusAction.ALLOW;
             }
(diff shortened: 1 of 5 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckCLDR.java
```diff
@@ -284,6 +284,13 @@ public StatusAction getShowRowAction(
                         : StatusAction.ALLOW_VOTING_AND_TICKET;
             }
 
+            // locales in CLDR_ADD_ITEMS_IN_VETTING_LOCALES may continue to be added
+            if (pathValueInfo != null
+                    && SubmissionLocales.CLDR_ADD_ITEMS_IN_VETTING_LOCALES.contains(
```
macchiati:
```suggestion
                    && SubmissionLocales.LOCALES_ALLOW_ADDING_IN_VETTING.contains(
```

macchiati: clearer name

macchiati (changes requested):
Looks good, except that the name is confusing; it is a set of locales that allow...

tools/cldr-code/src/main/java/org/unicode/cldr/test/SubmissionLocales.java
```diff
@@ -64,13 +64,25 @@ public final class SubmissionLocales {
     /** Space-separated list of TC locales to extend submission */
     public static final String DEFAULT_EXTENDED_SUBMISSION = "";
 
-    /** Additional TC locales which have extended submission. Do not add non-tc locales here. */
+    /**
+     * Additional TC locales which have extended submission. Do not add non-tc locales here, which
+     * are already included by default.
+     */
     public static final Set<String> ADDITIONAL_EXTENDED_SUBMISSION =
```
macchiati: Not really part of the goal, but this could also be improved: LOCALES_ALLOWING_EXTENDED_SUBMISSION.

srl295: this one is already used in production, in documentation, etc. I can rename the variable to match the property, though?

srl295 (commented):
Updated per comments

macchiati (approved):

srl295:
@macchiati sorry, i meant to fix the name of the PR (and thus the merge commit) before merging!

PR #5816, opened 2026-06-10
CLDR-19273 Fix remaining links to tsvs

macchiati:
CLDR-19273

This fixes the two remaining TSV files, and also puts the printfs under an environment flag, to reduce the chaff when generating the charts.

I chose a simple way to reduce the chaff, just prefixing System.out.print... with if (DEBUG) .
We can look at a better way later, but it would be easy to change at that point.

ALLOW_MANY_COMMITS=true


macchiati:
I'd like to get this reviewed.

srl295 (approved):

PR #5817, opened 2026-06-11
CLDR-16983 use punctuation, number exemplars in CheckForExemplars

srl295:
CLDR-16983

- also cleanup javadocs and add more TODO references to CLDR-19115
- factor the `are not in the exemplar characters` message into a single constant

ALLOW_MANY_COMMITS=true

### DIFF
--- a/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
+++ b/tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Date;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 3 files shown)


tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
```diff
@@ -218,6 +223,14 @@ public final CLDRLocale getCLDRLocale() {
         return cldrLocale;
     }
 
+    private static final EnumSet<ExemplarType> EXEMPLARS_TO_ADD_TO_DEFAULT =
```
macchiati, line 226: In new code always best to use ImmutableSet.of or ImmutableSortedSet.of for statics since EnumSet is mutable.

not blocker.

macchiati (approved):
I'll add a separate ticket for the item I observed.

tools/cldr-code/src/main/java/org/unicode/cldr/test/CheckForExemplars.java
srl295: Good point, thanks.

